PR #7, opened 2025-03-21
Remove Fingerprint Generation Capabilities

decaffeinatedio:
This PR removes the most problematic fingerprint generation from synthetic.rs while keeping the overall workflow intact (and tests still passing). 

image

Even in a prototype, I would strongly recommend against creating a fingerprint like this. 

Why?
- In its default form, it cannot be effectively controlled or observed by an end user. In its default form, an incognito session and a standard browser session could be linked together by any platforms downstream. Does that meet user expectations? 
- Running this out of the box is likely unlawful in many regions of the world. Shipping this without a big warning label on the side (much less implementing some simple consent checks) sends a poor message about what smart defaults look like in 2025. 
- While I don't expect activists and browser makers to be enthused about this project (or much adtech does at all), the signal of integrating a super cookie fingerprint as a v.0 priority is a great way to reinforce the existing convictions of that group and spur more drastic actions.

### DIFF
--- a/src/synthetic.rs
+++ b/src/synthetic.rs
@@ -12,9 +12,6 @@ type HmacSha256 = Hmac<Sha256>;
 
 /// Generates a fresh synthetic_id based on request parameters
 pub fn generate_synthetic_id(settings: &Settings, req: &Request) -> String {
-    let user_agent = req
-        .get_header(header::USER_AGENT)
-        .map(|h| h.to_str().unwrap_or("Unknown"));
     let first_party_id = handle_request_cookies(req).and_then(|jar| {
         jar.get("pub_userid")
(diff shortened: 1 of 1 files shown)


iabshailley:
We will not take any action on this PR for now. The project will be governed by the Trusted Server task force that is being constituted and it will make decisions on features and functions of the trusted server. The current code base is a proof of concept to demonstrate certain capabilities.

sulleh:
Wasn't the request to us to specifically submit a PR?

jevansnyc:
Thanks. We [deployed this as a templating option](https://github.com/IABTechLab/trusted-server/blob/main/potsi.toml#L19) so the publisher can choose if they would like the 1st party synthetic ID. Closing this request out.

PR #21, opened 2025-05-27
updated README for better description of TS

jevansnyc:
updated README to better describe what Trusted Server does and the goal


aram356 (approved):
👍 Looks good

PR #23, opened 2025-06-26
GAM module, gam testing endpoints, working locally

jevansnyc:
New functionality to call GAM server side accomplishing the following: 

1. Declaring intent to google 
2. Initiates the ad request and handles the response (as well as error handling) 

... we aren't trying to render yet 

More context and details: 

The response is multi-part, with a JSON metadata header for each ad slot followed by a full HTML payload.

1. Response Format: JSON Header + HTML Body

For each ad request, we get a JSON array that contains critical metadata. 

Important components would be: 

"/37xx/trustedserver/homepage": The ad unit this response is for.

"html": The format of the creative.

90, 728: The ad's height and width (needed for rendering)

[138210129xxx]: The Creative ID.

[4410806xxx]: The Line Item ID.

2. Creative Payload: Self-Contained HTML Document

The core of the response is a full <!doctype html>...</html> document. This document should have everything needed for the ad to function in a sandboxed environment:

The creative asset itself: An  tag pointing to a tpc.googlesyndication.com URL.

The click-through URL: The <a> tag's href points to a googleads.g.doubleclick.net/pcs/click tracker.

3. Rendering Instructions: Implicitly an Iframe

The payload is a full HTML document, implicitly telling us: "You must render this inside an iframe."

Simply injecting this HTML into our main page would break the DOM. The iframe is necessary to isolate the ad's environment from the publisher's page.

4. Tracking & Lifecycle Beacons: The vu() function

Inside the creative's HTML, there is a critical JavaScript call: vu("https://securepubads.g.doubleclick.net/pcs/view?xai=...").

This vu() (View URL) function fires the impression tracking pixel. This is a mandatory action. We cannot consider the ad delivered until this URL has been requested.


aram356 (approved):
Looks good

PR #27, opened 2025-07-03
Changed to use logger in place of print statement

aram356:


jevansnyc (approved):
Approved, but lets drop the "Fastly Jason POP" header next :)

PR #29, opened 2025-07-03
Added publisher config

aram356:


jevansnyc (approved):
Couple of things to ask on call around plan for populating partner URLs but LGTM

PR #40, opened 2025-08-05
Feature/proxy to prebidserver

aram356:


crates/common/src/html_processor.rs
```diff
@@ -1,6 +1,8 @@
 //! Simplified HTML processor that combines URL replacement and Prebid injection
 //!
 //! This module provides a StreamProcessor implementation for HTML content.
+use std::cell::RefCell;
+use std::rc::Rc;
```
jevansnyc, line 5: test comment for slack integration ...

PR #43, opened 2025-09-04
Caching optimizations

jevansnyc:
- Added static asset endpoints with 24-hour caching for CSS/JS files
- Updated main page caching to 10 minutes with ETag versioning
- Enhanced no-cache headers for all ad content to ensure freshness
- Implemented cache-busting timestamps for ad requests only
- Optimized asset loading order with async/defer JavaScript 

Things to test: 
- Verify main page loads quickly from cache with proper cache headers
- Static assets return 304 Not Modified after initial load, 
- Ad content is not cached and loading new assets with each reload


aram356:
Closing to reimplemented at different stage of the project

PR #66, opened 2025-10-23
Change x-synthetic-trusted-server header to x-psid-ts

### LINKED ISSUE #52: Rename headers for publisher IDs
Headers that should be renamed: 

x-synthetic-trusted-server should become x-psid-ts (x-publisher specific trusted server)

ChristianPavilonis:
Closes #52


ChristianPavilonis:
Based on the issue description it's ambiguous if we want to literally change the header names.
or do we mean to do something like this: #67 to make it configurable?

aram356 (approved):
🔧 Please resolve conflict and formatting

PR #84, opened 2025-11-06
removes robot tag on every request adds setting for custom response headers

### LINKED ISSUE #77: As a publisher I need a setting to be able to toggle the x-robots-tag header to control crawlers and llms
We now have the response header that injects the following into every response: 

x-robots-tag: noindex, nofollow

found in this PR: 

https://github.com/IABTechLab/trusted-server/pull/74

However, obviously some publishers will want their sites indexed and crawled so we just need a simple setting to turn this on (for testing, blocking LLMs, etc) or off (for prod, open web traffic)

ChristianPavilonis:
Closes #77

To add any custom response header add them to the trusted-server.toml
```toml
[response_headers]
X-Robots-Tag = "noindex, nofollow"
```

### DIFF
--- a/trusted-server.toml
+++ b/trusted-server.toml
@@ -28,3 +28,7 @@ secret_key = "trusted-server"
 #   - "publisher_domain"
 #   - "accept_language"
 template = "{{ client_ip }}:{{ user_agent }}:{{ first_party_id }}:{{ auth_user_id }}:{{ publisher_domain }}:{{ accept_language }}"
+
+# Custom headers to be included in every response
+[response_headers]
+X-Custom-Header = "custom header value"


crates/fastly/src/main.rs
```diff
@@ -84,8 +84,9 @@ async fn route_request(settings: Settings, req: Request) -> Result<Response, Err
     // Convert any errors to HTTP error responses
     let mut response = result.unwrap_or_else(to_error_response);
 
-    // Add X-Robots-Tag header to prevent crawlers and indexers
-    response.set_header("X-Robots-Tag", "noindex, nofollow");
+    for (key, value) in &settings.response_headers {
```
aram356: ⛏️ Can you add example headers to trusted-server.toml

aram356 (approved):
👍 Looks good

PR #113, opened 2025-11-17
Proposal to standardize integrations

### LINKED ISSUE #109: As publisher I want to selective build Javascript modules

### LINKED ISSUE #110: As vendor I want to contribute modules in standardized way to support proxying, configuring and shimming JS

aram356:
Closes #109 
Closes #110


ChristianPavilonis:
I like it.

Take a look at #114 for the javascript. It bundles only the packages listed in an env var at build time into one javascript file, might be easier to integrate by loading all shims into one file and.
Also if there were a lot of integrations we could have potentially lot of js files to build into the binary.

On the rust side if a lot of proxies do the exact same thing we could create a derive macro or something.

PR #124, opened 2025-11-21
fix: prebid detection for nextjs

### LINKED ISSUE #106: For Next.JS application prebid detection fails

ChristianPavilonis:
Closes #106

### DIFF
--- a/crates/common/src/integrations/nextjs.rs
+++ b/crates/common/src/integrations/nextjs.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 
 use regex::{escape, Regex};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


trusted-server.toml
```diff
@@ -42,7 +42,7 @@ auto_configure = false
 debug = false
 
 [integrations.nextjs]
-enabled = false
+enabled = true
```
aram356: 🔧 Should not be enabled by default. Can be enabled via env variable:

`TRUSTED_SERVER__INTEGRATIONS__NEXTJS__ENABLED`

aram356 (commented):
♻️ Please close. Preferred implementation #128

PR #131, opened 2025-12-01
Add didomi integration

### LINKED ISSUE #65: As publisher I want to configure and use Didome as CMP in TS

aram356:
Closes #65


crates/common/src/integrations/didomi.rs
```diff
@@ -0,0 +1,279 @@
...
+    Some(
+        IntegrationRegistration::builder(DIDOMI_INTEGRATION_ID)
+            .with_proxy(integration)
+            .build(),
+    )
+}
+
+#[async_trait(?Send)]
+impl IntegrationProxy for DidomiIntegration {
+    fn routes(&self) -> Vec<IntegrationEndpoint> {
```
ChristianPavilonis: 🔧 Use new route helpers

PR #123, opened 2025-11-20
Lockr Integration

ChristianPavilonis:
closes #121 

Would be nice to have a test account to confirm it's working correctly. (#120)

PR #108, opened 2025-11-14
Permutive proxy

ChristianPavilonis:
Closes #97
Closes #98

PR #136, opened 2025-12-02
Support wildcard route matching for integrations

ChristianPavilonis:
Allows integrations to easily proxy any request matching a wildcard pattern to downstream apis.

This is needed for #108 and #123 to work and pass their api calls through trusted server without needing exact matches for every call.

- Uses matchit crate for fast route matching.
- adds helper methods for defining endpoints under a namespace
  - but still allow integrations like prebid to customize the route


aram356 (changes requested):
🔧 Please look into eliminating loop

Option 1: Integration should register a prefix that would match to
Option 2: Maybe we can take advantage of https://docs.rs/matchit/latest/matchit/

aram356 (approved):
👍 Look good.

PR #142, opened 2025-12-11
experiment/tree sitter

ChristianPavilonis:


aram356:
Closing since parsing JS is not needed at the moment

PR #147, opened 2025-12-16
Auction orchestration

### LINKED ISSUE #139: As developer I want to register my auction integration to be in the middle of auction process

ChristianPavilonis:
Closes #139

### DIFF
--- a/crates/common/src/integrations/prebid.rs
+++ b/crates/common/src/integrations/prebid.rs
@@ -145,6 +145,7 @@ impl PrebidIntegration {
             .with_body(body))
     }
 
+    #[allow(dead_code)]
     async fn handle_first_party_ad(
         &self,
         settings: &Settings,
@@ -682,6 +683,7 @@ fn append_query_params(url: &str, params: &str) -> String {
 ///
 /// Searches through the OpenRTB seatbid/bid structure for a bid whose `impid`
 /// matches `slot` and returns its `adm` (ad markup) value.
+#[allow(dead_code)]
 fn extract_adm_for_slot(response: &Json, slot: &str) -> Option<String> {
     let seatbids = response.get("seatbid")?.as_array()?;
     for seatbid in seatbids {


crates/common/src/auction/README.md
```diff
@@ -0,0 +1,233 @@
...
+### 1. Parallel + Mediation (Recommended)
+**Use case:** Header bidding with GAM as primary ad server
+
+```toml
+[auction]
+enabled = true
+strategy = "parallel_mediation"
+bidders = ["prebid", "aps"]
+mediator = "gam"
+timeout_ms = 2000
```
jevansnyc: timeout meaning the time for PBS, APS etc to come back to TS right? This does not include any GAM RT's or think time?

crates/common/src/auction/README.md
```diff
@@ -0,0 +1,233 @@
...
+[auction]
+enabled = true
+strategy = "parallel_only"
+bidders = ["prebid", "aps"]
+timeout_ms = 2000
+```
+
+**Flow:**
+1. All bidders run in parallel
+2. Highest bid wins
```
jevansnyc: This may be addressed elsewhere, but AFAIK APS doesn't return a clear text creative bid, so we'll need to understand what the decryption of the TAM value looks like

docs/APS_QUICKSTART.md
```diff
@@ -0,0 +1,188 @@
+# APS Integration Quick Start
```
aram356: ⛏️ Should be lower case file and be inside guide

trusted-server.toml
```diff
@@ -81,3 +81,41 @@ rewrite_sdk = true
...
+
+# Real APS Configuration (disabled by default - enable when real integration is ready)
+[integrations.aps]
+enabled = false
+pub_id = "your-aps-publisher-id"
+endpoint = "https://aax.amazon-adsystem.com/e/dtb/bid"
+timeout_ms = 800
+
+# Mock GAM Configuration (for testing)
+[integrations.gam_mock]
```
aram356: 🔧 Why we need this because mocking is separated into different service

crates/common/src/auction/mod.rs
```diff
@@ -0,0 +1,484 @@
...
+                .store(creative_key.clone(), rewritten_creative)
+                .change_context(TrustedServerError::Auction {
+                    message: format!(
+                        "Failed to store creative for auction {} slot {}",
+                        auction_id, slot_id
+                    ),
+                })?;
+
+            // Generate proxy URL for the creative
+            let url = format!(
```
aram356: 🔧 Add token to decode on request to prevent fraud.

aram356 (changes requested):
👍 Good start

1. Please address tests and formatting 
2. I would like to re-review

crates/common/src/integrations/aps.rs
```diff
@@ -0,0 +1,908 @@
...
+
+    /// Decode base64-encoded price from mock APS response.
+    /// Note: Real APS uses proprietary encoding that only Amazon/GAM can decode.
+    /// Our mock uses simple base64 encoding for testing purposes only.
+    fn decode_aps_price(encoded: &str) -> f64 {
+        use base64::{engine::general_purpose::STANDARD, Engine as _};
+
+        let decoded = match STANDARD.decode(encoded) {
+            Ok(bytes) => bytes,
+            Err(_) => return 2.50,
```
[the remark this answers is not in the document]
ChristianPavilonis: actually because eventually we're sending the encoded price to gam shouldn't I move any decode logic to the mock service anyway?

crates/common/src/auction/orchestrator.rs
```diff
@@ -0,0 +1,680 @@
...
+            let mediator = self.get_provider(mediator_name)?;
+
+            log::info!(
+                "Sending {} provider responses to mediator: {}",
+                provider_responses.len(),
+                mediator.provider_name()
+            );
+
+            // Create a modified request with all bids attached
+            let mut mediation_request = request.clone();
```
aram356: ♻️ Reconsider cloning with `Cow<'_, AuctionRequest> `

crates/common/src/integrations/aps.rs
```diff
@@ -0,0 +1,906 @@
+//! Amazon Publisher Services (APS/TAM) integration.
```
aram356: ❓ We decide not mock here. Is there reason why it is still here?

ChristianPavilonis: This module for now is sending "aps" bid requests to the configured mock server, to test orchestrator and simulate how aps bids flow to an ad server, but perhaps it shouldn't be called aps.

PR #163, opened 2026-01-08
Proposal for attestation

### LINKED ISSUE #162: Propose attestation framework and tooling

aram356:
Closes #162


ChristianPavilonis (approved):
Looks good.

We could probably add more to the cli also:
- init fastly compute / cf worker
- initialize config and secret stores for request signing.

prk-Jr (approved):

Issue #184, opened 2026-01-21
Standardize logging: Remove manual prefixes, add module target to logger format


aram356:
## Summary

Remove manual prefixes like `[Gzip]`, `[HtmlRewriter]`, `proxy:` from log messages and update the logger format to automatically include the module name via `record.target()`.

## Current State

Log messages use inconsistent manual prefixes:

```rust
// streaming_processor.rs - manual prefix
log::info!("[Gzip] Decompressed size: {} bytes", decompressed.len());

// proxy.rs - different manual prefix  
log::debug!("proxy: no synthetic_id to forward to origin");

// backend.rs - no prefix
log::info!("enable ssl for backend: {}", backend_name);
```

The logger format in `crates/fastly/src/main.rs:132-138` currently outputs:
```
2026-01-21T10:00:00.000Z INFO message here
```

It does **not** include `record.target()` which would provide the module path automatically.

## Target State

Update logger to include module name automatically, then remove manual prefixes:

```rust
// Logger format includes module automatically
// streaming_processor.rs
log::info!("Decompressed size: {} bytes", decompressed.len());
// Output: 2026-01-21T10:00:00.000Z INFO [streaming_processor] Decompressed size: 1024 bytes

// proxy.rs
log::debug!("no synthetic_id to forward to origin");
// Output: 2026-01-21T10:00:00.000Z DEBUG [proxy] no synthetic_id to forward to origin
```

### 1. Update logger format in `crates/fastly/src/main.rs:132-138`

```rust
.format(|out, message, record| {
    out.finish(format_args!(
        "{} {} [{}] {}",
        chrono::Local::now().to_rfc3339_opts(chrono::SecondsFormat::Millis, true),
        record.level(),
        record.target().split("::").last().unwrap_or(record.target()),
        message
    ))
})
```

### 2. Remove manual prefixes from `crates/common/src/streaming_processor.rs`

| Line | Current | New |
|------|---------|-----|
| 188 | `"[Gzip] Decompressed size: {} bytes"` | `"Decompressed size: {} bytes"` |
| 198 | `"[Gzip] Processed size: {} bytes"` | `"Processed size: {} bytes"` |
| 231-232 | `"[Gzip->None] Decompressed size: {} bytes"` | `"Gzip->None decompressed size: {} bytes"` |
| 244 | `"[Gzip->None] Processed size: {} bytes"` | `"Gzip->None processed size: {} bytes"` |
[...]

PR #186, opened 2026-01-21
Standardize logging: Remove manual prefixes

prk-Jr:
Standardize logging: Remove manual prefixes, add module target to logger format

Uses record.target() in the logger to automatically include module names in brackets, eliminating inconsistent manual prefixes that caused duplicate context like "[proxy] proxy: message".

closes https://github.com/IABTechLab/trusted-server/issues/184


aram356 (approved):
👍 Looks good.

Issue #190, opened 2026-01-21
Standardize integration trait naming: remove IntegrationConfigTrait alias


aram356:
## Summary

Standardize the import and usage of the `IntegrationConfig` trait. Currently some modules rename it to `IntegrationConfigTrait` via import alias, creating inconsistency.

## Current State

The trait is defined in `crates/common/src/settings.rs:75`:
```rust
pub trait IntegrationConfig: DeserializeOwned + Validate {
    fn enabled(&self) -> bool;
}
```

**Inconsistent imports across integrations:**

| File | Import Style | Usage |
|------|--------------|-------|
| `integrations/prebid.rs:25` | `use crate::settings::{IntegrationConfig, ...}` | `impl IntegrationConfig for PrebidIntegrationConfig` |
| `integrations/didomi.rs:14` | `use crate::settings::{IntegrationConfig, ...}` | `impl IntegrationConfig for DidomiIntegrationConfig` |
| `integrations/nextjs/mod.rs:7` | `use crate::settings::{IntegrationConfig, ...}` | `impl IntegrationConfig for NextJsIntegrationConfig` |
| `integrations/permutive.rs:21` | `use crate::settings::{IntegrationConfig as IntegrationConfigTrait, ...}` | `impl IntegrationConfigTrait for PermutiveConfig` |
| `integrations/lockr.rs:32` | `use crate::settings::{IntegrationConfig as IntegrationConfigTrait, ...}` | `impl IntegrationConfigTrait for LockrConfig` |
| `integrations/testlight.rs:18` | `use crate::settings::{IntegrationConfig as IntegrationConfigTrait, ...}` | `impl IntegrationConfigTrait for TestlightConfig` |

## Target State

All integrations should use the trait name directly without aliasing:

```rust
// Before (inconsistent)
use crate::settings::{IntegrationConfig as IntegrationConfigTrait, Settings};
impl IntegrationConfigTrait for PermutiveConfig { ... }

// After (consistent)
use crate::settings::{IntegrationConfig, Settings};
impl IntegrationConfig for PermutiveConfig { ... }
```

## Rationale

- **Consistency** - Same pattern across all integration modules
- **Clarity** - No need to remember alias names
- **Searchability** - `impl IntegrationConfig` finds all implementations
- **Rust convention** - Trait aliasing is typically only used to avoid name collisions (none exist here)

### 1. Update `crates/common/src/integrations/permutive.rs`

```diff
[...]

PR #196, opened 2026-01-22
Standardize IntegrationConfig trait imports

prk-Jr:
Remove the IntegrationConfigTrait alias from permutive, lockr, and testlight integration modules. Use IntegrationConfig directly without aliasing, matching the pattern in other integrations and improving code consistency and searchability.

Trait aliasing is typically reserved for avoiding name collisions, which don't exist here.

Resolves: #190


aram356 (approved):
👍 Looks good

PR #213, opened 2026-01-29
Fix rust build cache github actions

ChristianPavilonis:
First run: Rust tests ~5min run
Second run: Rust tests ~1min run


aram356 (changes requested):
It seems we need rustup for wasm

Issue #192, opened 2026-01-21
Replace unwrap() with expect() or ? operator in production code


aram356:
## Summary

Replace `unwrap()` calls in production code paths with proper error handling using `expect()` (with context message) or the `?` operator. `unwrap()` causes panics without context, making debugging difficult.

## Current State

Multiple `unwrap()` calls exist in production code:

| File | Line(s) | Example |
|------|---------|---------|
| `crates/common/src/proxy.rs` | ~18 instances | Various `.unwrap()` calls |
| `crates/common/src/http_util.rs` | 64 | `.expect("encryption failure")` (good) but others use `.unwrap()` |
| `crates/common/src/http_util.rs` | 125 | `.unwrap_or_default()` - silent error recovery |
| `crates/common/src/synthetic.rs` | 36, 51, 55-60 | `.unwrap_or()` chains |

## Target State

- **Production code**: Use `?` operator with proper error types, or `expect("clear reason")` for truly impossible cases
- **Test code**: `unwrap()` is acceptable, but `expect("test context")` is preferred
- **Optional values**: Use `unwrap_or_default()` or `unwrap_or()` only when the default is semantically correct

```rust
// Before - no context on panic
let value = some_option.unwrap();

// After - context on panic (for truly impossible cases)
let value = some_option.expect("value should exist because X was validated");

// Better - proper error handling
let value = some_option.ok_or_else(|| 
    Report::new(TrustedServerError::InvalidState { 
        message: "value missing".into() 
    })
)?;
```

### 1. Audit `crates/common/src/proxy.rs`

This file has the most `unwrap()` usage. For each instance:
- Determine if the unwrap can actually fail
- If yes: convert to `?` with proper error
- If no (impossible case): convert to `expect("reason why this can't fail")`

### 2. Audit `crates/common/src/http_util.rs`

Review encryption/decryption code for proper error handling.

### 3. Audit `crates/common/src/synthetic.rs`

Review `unwrap_or()` chains - ensure defaults are semantically correct.

### 4. Run clippy with `unwrap_used` lint

```bash
cargo clippy -- -W clippy::unwrap_used
```

This will identify all remaining `unwrap()` calls.

## Decision Guide

```
Can this unwrap actually fail in production?
├── Yes → Use ? operator with proper error type
└── No (impossible by construction)
    └── Use expect("reason") documenting why it can't fail
```

## Examples of Good Patterns

```rust
// Impossible case - document why
[...]

PR #214, opened 2026-01-29
Replace unwrap with expect

prk-Jr:
Replace unwrap() with expect() or ? operator in production code

Resolves: #192


ChristianPavilonis (changes requested):
The issue #192 has 5 files to look at.

If they didn't require changes please state why those `unwrap()`, `unwrap_or_default()`, etc. are not an issue.

The unwraps in `proxy.rs` are in the tests but could probably use expects in case those tests fail for better visibility as to why they failed.

PR #238, opened 2026-02-05
Add configurable TLS certificate verification and fix port handling

### LINKED ISSUE #246: Add configurable TLS certificate verification and fix port handling
## Description

Add configurable TLS certificate verification for backend requests and fix non-standard port handling in Host headers.

## Changes

- Add `Proxy.certificate_check` setting (defaults to `true` for secure production use)
- Add `compute_host_header()` to properly format Host header with non-standard ports
- Extend `ensure_origin_backend()` with `certificate_check` parameter
- Include cert setting in backend name to avoid reusing backends with different TLS settings
- Add comprehensive tests for port preservation

## Why

This fixes an issue where backends behind reverse proxies (like Caddy) would generate URLs without the port when the Host header didn't include it.

## Related

- PR: #238
- Parent: #179

aram356:
## Summary

- Adds `Proxy.certificate_check` setting to control TLS certificate verification (defaults to `true` for secure production use, set to `false` for local development with self-signed certs)
- Adds `compute_host_header()` function to properly format Host header with non-standard ports, fixing an issue where backends behind reverse proxies would generate URLs without the port
- Extends `ensure_origin_backend()` API with `certificate_check` parameter
- Includes cert setting in backend name to avoid reusing backends with different TLS settings

## Test plan

- [x] Verify `cargo check` passes
- [x] Verify `cargo test -p trusted-server-common` passes
- [x] Test proxy requests to origins with non-standard ports (e.g., `:9443`) preserve the port in signed URLs
- [x] Test HTML rewriting preserves non-standard ports in sub-resource URLs
- [x] Test with `certificate_check = false` in local dev with self-signed certs

Closes #246
Related to #179

### DIFF
--- a/crates/common/src/auction/provider.rs
+++ b/crates/common/src/auction/provider.rs
@@ -62,7 +62,7 @@ pub trait AuctionProvider: Send + Sync {
     ///
     /// This is used by the orchestrator to correlate responses with providers
     /// when using `select()` to wait for multiple concurrent requests.
-    /// The backend name should match what `ensure_backend_from_url()` returns
+    /// The backend name should match what `BackendConfig::from_url()` returns
     /// for this provider's endpoint.
     fn backend_name(&self) -> Option<String> {
         None
--- a/crates/common/src/backend.rs
+++ b/crates/common/src/backend.rs
@@ -33,128 +33,167 @@ fn compute_host_header(scheme: &str, host: &str, port: u16) -> String {
     }
 }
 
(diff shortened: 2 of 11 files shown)


prk-Jr (approved):
Looks good.

ChristianPavilonis (approved):
👍 looks good.

But, maybe we should consider using a struct with builder pattern instead of a function for dynamic backend creation, especially if we wrap any more options we'd have to do less code changes with a struct with defaults

PR #255, opened 2026-02-06
Set x-synthetic-id header on integration routes for response from Trusted Server

### LINKED ISSUE #205: Synthetic ID Header [x-synthetic-id] is not being set on /integration re-hosted into first party publisher domain
For example here: 

Image

We should properly set the Synthetic ID header on all URL paths under the pub domain, provided the user hasn't opted out via the CMP.

prk-Jr:
Integration proxy responses were missing the `x-synthetic-id` header because `handle_proxy` dispatched to integrations without adding it. This caused identity tracking to break for first-party re-hosted integrations like Permutive's secure-signal endpoint.

Centralizing the header logic in `handle_proxy` ensures all current and future integrations automatically include the synthetic ID, rather than requiring each integration to implement it manually.

## Test Plan

Manually verified locally via `fastly compute serve` against Prospect A production config:

- `GET /integrations/permutive/secure-signals` → `x-synthetic-id` header and `Set-Cookie: synthetic_id=...` present
- `GET /integrations/lockr/api` → same
- `GET /` (publisher proxy) → same
- Confirmed `x-synthetic-id` is **not** forwarded to downstream third-party origins (filtered by `INTERNAL_HEADERS` via `copy_custom_headers`)
- All 336 tests pass

Fixes #205

### DIFF
--- a/crates/common/src/integrations/registry.rs
+++ b/crates/common/src/integrations/registry.rs
@@ -9,6 +9,7 @@ use fastly::{Request, Response};
 use matchit::Router;
 
 use crate::constants::HEADER_X_SYNTHETIC_ID;
+use crate::cookies::set_synthetic_cookie;
 use crate::error::TrustedServerError;
 use crate::settings::Settings;
 use crate::synthetic::get_or_generate_synthetic_id;
@@ -613,19 +614,31 @@ impl IntegrationRegistry {
         method: &Method,
         path: &str,
         settings: &Settings,
-        req: Request,
+        mut req: Request,
     ) -> Option<Result<Response, Report<TrustedServerError>>> {
         if let Some((proxy, _)) = self.find_route(method, path) {
             // Generate synthetic ID before consuming request
             let synthetic_id_result = get_or_generate_synthetic_id(settings, &req);
 
+            // Set synthetic ID header on the request so integrations can read it
+            if let Ok(ref synthetic_id) = synthetic_id_result {
+                req.set_header(HEADER_X_SYNTHETIC_ID, synthetic_id.as_str());
+            }
+
             let mut result = proxy.handle(settings, req).await;
 
             // Set synthetic ID header on successful responses
             if let Ok(ref mut response) = result {
-                if let Ok(ref synthetic_id) = synthetic_id_result {
-                    response.set_header(HEADER_X_SYNTHETIC_ID, synthetic_id.as_str());
-                    crate::cookies::set_synthetic_cookie(settings, response, synthetic_id.as_str());
+                match synthetic_id_result {
+                    Ok(ref synthetic_id) => {
+                        response.set_header(HEADER_X_SYNTHETIC_ID, synthetic_id.as_str());
+                        set_synthetic_cookie(settings, response, synthetic_id.as_str());
+                    }
+                    Err(ref err) => {
+                        log::warn!(
+                            "Failed to generate synthetic ID for integration response: {err:?}"
+                        );
+                    }
                 }
             }
             Some(result)
--- a/crates/common/src/integrations/registry.rs
+++ b/crates/common/src/integrations/registry.rs
@@ -630,9 +630,9 @@ impl IntegrationRegistry {
             // Set synthetic ID header on successful responses
             if let Ok(ref mut response) = result {
                 if let Ok(ref synthetic_id) = synthetic_id_result {
-                    response.set_header(HEADER_X_SYNTHETIC_ID, synthetic_id.as_str());
+                    response.append_header(HEADER_X_SYNTHETIC_ID, synthetic_id.as_str());
                     if !has_synthetic_cookie {
-                        response.set_header(
+                        response.append_header(
                             header::SET_COOKIE,
                             create_synthetic_cookie(settings, synthetic_id.as_str()),
                         );
--- a/crates/common/src/integrations/registry.rs
+++ b/crates/common/src/integrations/registry.rs
@@ -4,12 +4,11 @@ use std::sync::{Arc, Mutex};
 
 use async_trait::async_trait;
 use error_stack::Report;
-use fastly::http::{header, Method};
+use fastly::http::Method;
 use fastly::{Request, Response};
 use matchit::Router;
 
-use crate::constants::{COOKIE_SYNTHETIC_ID, HEADER_X_SYNTHETIC_ID};
-use crate::cookies::{create_synthetic_cookie, handle_request_cookies};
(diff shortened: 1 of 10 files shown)


crates/common/src/integrations/registry.rs
```diff
@@ -611,7 +617,35 @@ impl IntegrationRegistry {
         req: Request,
     ) -> Option<Result<Response, Report<TrustedServerError>>> {
         if let Some((proxy, _)) = self.find_route(method, path) {
-            Some(proxy.handle(settings, req).await)
+            // Generate synthetic ID before consuming request
```
aram356: 🔧 Please use functions in cookies.rs We already have utils for this kind operations

aram356 (changes requested):
Good start but should use funcitons in cookies.rs

Please don't make the feature branch have this long name. Please stick to `feature/short-name`

trusted-server.toml
```diff
@@ -21,11 +21,6 @@ secret_key = "trusted-server"
 #   - "random_uuid"
 template = "{{ client_ip }}:{{ user_agent }}:{{ accept_language }}:{{ accept_encoding }}"
 
-# Custom headers to be included in every response
```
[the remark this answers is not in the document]
prk-Jr: Sorry, my bad. Let me add that again.

ChristianPavilonis (dismissed):

crates/common/src/integrations/registry.rs
```diff
@@ -611,7 +617,29 @@ impl IntegrationRegistry {
...
+                .unwrap_or(false);
+
+            let mut result = proxy.handle(settings, req).await;
+
+            // Set synthetic ID header on successful responses
+            if let Ok(ref mut response) = result {
+                if let Ok(ref synthetic_id) = synthetic_id_result {
+                    response.set_header(HEADER_X_SYNTHETIC_ID, synthetic_id.as_str());
+                    if !has_synthetic_cookie {
+                        response.set_header(
```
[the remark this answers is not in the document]
ChristianPavilonis: Sorry I just ment for the cookie, not the x-synthetic-id.

crates/common/src/integrations/registry.rs
```diff
@@ -1042,4 +1070,208 @@ mod tests {
         assert!(!registry.has_route(&Method::GET, "/integrations/test/users"));
         assert!(!registry.has_route(&Method::POST, "/integrations/test/users"));
     }
+
```
aram356: 🔧 Duplicate test

These three `cookie_jar_*` tests are testing `parse_cookies_to_jar` from `cookies.rs` — they should live there, not in the registry module. `cookies.rs` already has equivalent coverage (`test_parse_cookies_to_jar`, `test_parse_cookies_to_jar_emtpy`, etc.), so these are duplicates and can be removed.

aram356 (changes requested):
🔧 We need a more thorough fix:

##
publisher.rs:201-211 uses manual string splitting to check for an existing synthetic cookie:

```
let has_synthetic_cookie = req
    .get_header(header::COOKIE)
    .and_then(|h| h.to_str().ok())
    .map(|cookies| {
        cookies.split(';').any(|cookie| {
            cookie.trim_start().starts_with(&format!("{}=", COOKIE_SYNTHETIC_ID))
        })
    })
    .unwrap_or(false);
```   
    
While registry.rs:622-626 uses handle_request_cookies with CookieJar:

```
let has_synthetic_cookie = handle_request_cookies(&req)
    .ok()
    .flatten()
    .and_then(|jar| jar.get(COOKIE_SYNTHETIC_ID).map(|_| true))
    .unwrap_or(false);
```

These should use the same approach. The CookieJar version is more robust (proper cookie parsing). 

##

Since we already have cookie utilities in cookies.rs, suggest adding a `set_synthetic_cookie(settings: &Settings, response: &mut Response, synthetic_id: &str)` helper there that handles the append_header call. That way both publisher.rs and registry.rs call the same function, and the append_header fix and cookie creation logic live in one place.

##

We don't need `has_synthetic_cookie` check we was primarily for logging we should always set `COOKIE_SYNTHETIC_ID`

aram356 (changes requested):

crates/common/src/publisher.rs
```diff
@@ -198,23 +198,8 @@ pub fn handle_publisher_request(
...
-            })
-        })
-        .unwrap_or(false);
 
-    log::debug!(
-        "Proxy synthetic IDs - trusted: {}, has_cookie: {}",
-        synthetic_id,
-        has_synthetic_cookie
-    );
+    log::debug!("Proxy synthetic IDs - trusted: {}", synthetic_id,);
```
aram356: ⛏️ Trailing comma in log statement

aram356 (changes requested):

crates/common/src/integrations/registry.rs
```diff
@@ -611,7 +616,19 @@ impl IntegrationRegistry {
...
         if let Some((proxy, _)) = self.find_route(method, path) {
-            Some(proxy.handle(settings, req).await)
+            // Generate synthetic ID before consuming request
+            let synthetic_id_result = get_or_generate_synthetic_id(settings, &req);
+
+            let mut result = proxy.handle(settings, req).await;
+
+            // Set synthetic ID header on successful responses
+            if let Ok(ref mut response) = result {
+                if let Ok(ref synthetic_id) = synthetic_id_result {
```
aram356: 🔧 Silent swallowing error

aram356 (changes requested):

aram356 (changes requested):
🔧 We inadvertently now passing `HEADER_X_SYNTHETIC_ID` header to downstream integrations. We should not because synthetic header should stay within TS.

For example: `permutive.rs :502`

Please audit al integrations not to pass `HEADER_X_SYNTHETIC_ID` header

ChristianPavilonis (commented):
It's looking pretty good to me now

aram356 (approved):
Looks good. I would like to see test plan. Please confirm manually

prk-Jr:
> Looks good. I would like to see test plan. Please confirm manually

Manually verified locally via `fastly compute serve` against Prospect A production config:

- `GET /integrations/permutive/secure-signals` → `x-synthetic-id` and `Set-Cookie: synthetic_id=...` present
- `GET /integrations/lockr/api` → same
- `GET /` (publisher proxy) → same
- `x-synthetic-id` is not forwarded to downstream origins
- All 336 tests pass


> Looks good. I would like to see test plan. Please confirm manually

aram356 (approved):
Ship it!

PR #262, opened 2026-02-09
Add Google Tag Manager first-party integration

### LINKED ISSUE #224: As publisher I want to host GTM in publisher domain

prk-Jr:
Scripts like GTM and GA4 are often blocked by ad blockers or privacy extensions when loaded from third-party domains, leading to data loss. Third-party cookie deprecation further limits tracking durability.

This change transparently proxies GTM/GA4 scripts and analytics beacons through the Trusted Server, establishing a first-party context. It automatically rewrites HTML tags (including `<link rel="preload">`) and script content to point to local proxy endpoints, bypassing blockers and extending cookie life.

Includes:

- Proxy endpoints for `gtm.js`, `gtag/js`, `/collect`, and `/g/collect` with configurable caching and strict validation
- Server-side HTML rewriting for `src` and `href` attributes targeting GTM/GA domains
- Server-side script content rewriting to redirect internal GTM/GA calls through the proxy
- Client-side script guard (DOM interception) for dynamically inserted scripts on Next.js and SPA sites
- Privacy enhancement: client IP addresses are not forwarded to Google (Google sees only the edge server IP)
- Regex hardened against subdomain spoofing (e.g., `www.googletagmanager.com.evil.com`)
- Comprehensive testing: 18 unit/integration tests covering configuration, URL rewriting, HTML pipeline, header validation, and IP stripping

### Manual Test Plan

**Prerequisites**: Configure `.env` with GTM enabled and a valid container ID, then start the local server.

**1. Script proxy — gtm.js rewritten**

    curl -s 'http://127.0.0.1:7676/integrations/google_tag_manager/gtm.js?id=GTM-522ZT3X6' \
      | grep -c '/integrations/google_tag_manager'

Expected: count > 0

**2. Script proxy — gtag/js returns 200**

    curl -s -o /dev/null -w '%{http_code}' \
      'http://127.0.0.1:7676/integrations/google_tag_manager/gtag/js?id=G-DQMZGMPHXN'

Expected: `200`

**3. gtag/js content rewritten (no Google domains)**

    curl -s 'http://127.0.0.1:7676/integrations/google_tag_manager/gtag/js?id=G-DQMZGMPHXN' \
      | grep -c 'www.google-analytics.com'

Expected: `0`

**4. Beacon proxy — POST /g/collect**

    curl -s -o /dev/null -w '%{http_code}' -X POST \
      'http://127.0.0.1:7676/integrations/google_tag_manager/g/collect?v=2&tid=G-TEST&cid=123&en=page_view' \
      -d 'v=2&tid=G-TEST'

Expected: `204` or `200`

**5. Beacon proxy — GET /collect**

    curl -s -o /dev/null -w '%{http_code}' \
      'http://127.0.0.1:7676/integrations/google_tag_manager/collect?v=2&tid=G-TEST'

Expected: `204` or `200`

**6. Cache headers present**

    curl -sI 'http://127.0.0.1:7676/integrations/google_tag_manager/gtm.js?id=GTM-522ZT3X6' \
      | grep -i cache-control

Expected: `cache-control: public, max-age=900`

Resolves: #224


aram356 (changes requested):
🔧 Please make sure checks pass before assigning to review.

crates/common/src/integrations/google_tag_manager.rs
```diff
@@ -0,0 +1,740 @@
...
+    pub upstream_url: String,
+}
+
+impl IntegrationConfig for GoogleTagManagerConfig {
+    fn is_enabled(&self) -> bool {
+        self.enabled
+    }
+}
+
+fn default_enabled() -> bool {
```
aram356: 🔧 default_enabled() returns true. Other integrations default to false (or require explicit enabling)

crates/common/src/integrations/google_tag_manager.rs
```diff
@@ -0,0 +1,740 @@
...
+}
+
+#[must_use]
+pub fn build(settings: &Settings) -> Option<Arc<GoogleTagManagerIntegration>> {
+    let config = settings
+        .integration_config::<GoogleTagManagerConfig>(GTM_INTEGRATION_ID)
+        .ok()
+        .flatten()?;
+
+    if !config.enabled {
```
aram356: 🔧 Redundant enabled check in build()
The if !config.enabled check is redundant. `IntegrationConfig::is_enabled()` is already checked by the settings infrastructure when integration_config() returns Ok(None) for disabled integrations. DataDome and Permutive don't have this extra check.

aram356 (changes requested):
Good start. Need to address specific items and the following.

##
Duplicated rewrite logic across three places

The GTM URL rewriting logic exists in three separate methods: rewrite_gtm_script(), IntegrationAttributeRewriter::rewrite(), and IntegrationScriptRewriter::rewrite(). Each handles a slightly different set of patterns. This is error-prone — a new URL pattern needs to be added in multiple places.

crates/common/src/integrations/google_tag_manager.rs
```diff
@@ -0,0 +1,767 @@
...
+    }
+}
+
+impl IntegrationScriptRewriter for GoogleTagManagerIntegration {
+    fn integration_id(&self) -> &'static str {
+        GTM_INTEGRATION_ID
+    }
+
+    fn selector(&self) -> &'static str {
+        "script" // Match all scripts to find inline GTM snippets
```
aram356: ❓ Why this selector we already register "src" at line 240?

crates/common/src/integrations/google_tag_manager.rs
```diff
@@ -0,0 +1,767 @@
...
+            log::debug!("Rewriting GTM script content");
+            let body_str = response.take_body_str();
+            let rewritten_body = Self::rewrite_gtm_urls(&body_str);
+
+            response = Response::from_body(rewritten_body)
+                .with_header(
+                    fastly::http::header::CONTENT_TYPE,
+                    "application/javascript; charset=utf-8",
+                )
+                .with_header(fastly::http::header::CACHE_CONTROL, "public, max-age=3600");
```
aram356: ⛏️ Use cache max age setting

#[serde(default = "default_cache_max_age")]
pub cache_max_age: u32,

fn default_cache_max_age() -> u32 {
    900 // Match Google's default
}

crates/common/src/integrations/google_tag_manager.rs
prk-Jr: Good question! Yes, your example is exactly the case we're handling. There are two distinct rewriting mechanisms at play, each targeting a different way GTM can be embedded:

1.` IntegrationAttributeRewriter` (L253: handles_attribute("src" | "href")) — rewrites HTML attributes:

```html
<!-- External script tag — rewrite the "src" attribute value -->
<script src="https://www.googletagmanager.com/gtm.js?id=GTM-XXXX"></script>
<!-- Preload link — rewrite the "href" attribute value -->
<link rel="preload" href="https://www.googletagmanager.com/gtm.js?id=GTM-XXXX" as="script">
```

2. `IntegrationScriptRewriter` (L276: selector("script")) — rewrites inline script content, which is the standard GTM snippet that Google provides:


```html
<script>
  (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
  new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
  j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;
  j.src='https://www.googletagmanager.com/gtm.js?id='+i+dl;
  f.parentNode.insertBefore(j,f);
  })(window,document,'script','dataLayer','GTM-XXXX');
</script>
```
In this inline snippet, the URL is constructed dynamically inside JavaScript — there's no src attribute on the <script> tag itself to rewrite. Without the script content rewriter, this snippet would still load gtm.js directly from Google, bypassing our proxy entirely.

Both rewriters share the same consolidated regex (GTM_URL_PATTERN) so the replacement logic is defined once.

crates/common/src/integrations/google_tag_manager.rs
```diff
@@ -0,0 +1,786 @@
...
+            // Analytics beacons (GA4/UA)
+            // The GTM script is rewritten to point these beacons to our proxy.
+            self.get("/collect"),
+            self.post("/collect"),
+            self.get("/g/collect"),
+            self.post("/g/collect"),
+        ]
+    }
+
+    async fn handle(
```
ChristianPavilonis: POST analytics payloads are dropped while proxying GTM

Take a look at the failing tests I added here.

https://github.com/IABTechLab/trusted-server/compare/feat/gtm-integration...review/gtm-integration

ChristianPavilonis (changes requested):

ChristianPavilonis (changes requested):
For the re-writes to work with nextjs sites we need to use the scriptGuard pattern. Next sites can load the scripts dynamically after the fact so we catch and rewrite the script tags client side also.

see: https://github.com/IABTechLab/trusted-server/blob/main/crates/js/lib/src/integrations/permutive/script_guard.ts

aram356 (approved):
Look good overall. I would like to see test plan and address script guard.

ChristianPavilonis (approved):

aram356 (changes requested):
@prk-Jr Please deploy to staging. I did not see in the plan.

aram356:
from @jevansnyc 

There are some differences. For example the "Request Payload" data that shows up on production and not staging

Screenshot 2026-02-21 at 8 45 15 AM

prk-Jr:
> from @jevansnyc
> 
> There are some differences. For example the "Request Payload" data that shows up on oring and not on edge
> 
> Screenshot 2026-02-21 at 8 45 15 AM


I looked into it and I found something like this. @jevansnyc can you please point me to the request.

Screenshot 2026-02-23 at 2 41 49 PM

aram356 (approved):
👍 Looks good. Ship it!

PR #263, opened 2026-02-09
send permutive segments to adserver

### LINKED ISSUE #132: As publisher I want to see Permutive segment propagate to ad server
Should not send segments to Prebid server

ChristianPavilonis:
Passes Permutive audience segments from the browser through to the ad server mediation endpoint, enabling segment-targeted ad decisioning.
- Introduces a context provider registry on the JS side (context.ts) — a generic mechanism for integrations to contribute data to auction requests without core needing integration-specific knowledge.
- The Permutive integration registers a context provider that reads cohort segments from localStorage (permutive-app key) and injects them as permutive_segments in the request config.
- On the Rust side, convert_tsjs_to_auction_request now forwards all config entries into the AuctionRequest.context map instead of discarding them.
- The AdServer Mock provider reads permutive_segments from the context and appends them as a ?permutive=seg1,seg2,... query parameter on the mediation endpoint URL.
Changes

| Layer | File | What |
|-------|------|------|
| JS core | core/context.ts (new) | Context provider registry (registerContextProvider, collectContext) |
| JS core | core/request.ts | Calls collectContext() to build config payload |
| JS integration | integrations/permutive/segments.ts (new) | Reads segments from localStorage (core.cohorts.all primary, eventUpload fallback) |
| JS integration | integrations/permutive/index.ts | Registers context provider with Permutive segments |
| Rust common | auction/formats.rs | Forwards config entries into AuctionRequest.context |
| Rust common | integrations/adserver_mock.rs | build_endpoint_url() appends ?permutive= query param |
| Tests | context.test.ts, segments.test.ts, Rust unit tests | Full coverage for new functionality |

Closes #132

---

How to Manually Test
1. Verify Permutive segment extraction (JS)
1. Open a publisher page that has Permutive active (or any page where you can set localStorage manually).
2. Open DevTools Console and confirm localStorage.getItem('permutive-app') has data — look for core.cohorts.all containing segment IDs.
3. If testing without a real Permutive install, seed it manually:
```
      localStorage.setItem('permutive-app', JSON.stringify({
     core: { cohorts: { all: ['10000001', '10000003', 'adv', 'bhgp'] } }
   }));
```
   5. Load the page with the tsjs tag. In the DevTools Console, filter for context — you should see a log like:
      requestAds: payload { units: N, contextKeys: ["permutive_segments"] }
   
2. Verify segments are sent to the server
1. Open DevTools Network tab and filter for /auction.
2. Trigger an ad request.
3. Inspect the /auction POST request body — the config object should contain:
```
      {
     config: {
       permutive_segments: [10000001, 10000003, adv, bhgp]
     }
   }
```

### DIFF
--- a/crates/common/src/auction/context.rs
+++ b/crates/common/src/auction/context.rs
@@ -4,7 +4,25 @@
 //! forward integration-supplied data (e.g. audience segments) as URL query
 //! parameters without hard-coding integration-specific knowledge.
 
-use std::collections::HashMap;
+use serde::{Deserialize, Serialize};
+use std::collections::{BTreeMap, HashMap};
+
+/// A strongly-typed context value forwarded from the JS client payload.
+///
+/// Replaces raw `serde_json::Value` so that consumers get compile-time
+/// exhaustiveness checks. The `#[serde(untagged)]` attribute preserves
+/// wire-format compatibility — the JS client sends plain JSON arrays, strings,
+/// or numbers which serde maps to the matching variant in declaration order.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(untagged)]
+pub enum ContextValue {
+    /// A list of string values (e.g. audience segment IDs).
+    StringList(Vec<String>),
+    /// A single string value.
+    Text(String),
+    /// A numeric value.
+    Number(f64),
+}
 
 /// Mapping from auction-request context keys to query-parameter names.
 ///
@@ -17,7 +35,7 @@ use std::collections::HashMap;
 /// permutive_segments = "permutive"
 /// lockr_ids          = "lockr"
 /// ```
-pub type ContextQueryParams = HashMap<String, String>;
+pub type ContextQueryParams = BTreeMap<String, String>;
 
 /// Build a URL by appending context values as query parameters according to the
 /// provided mapping.
@@ -34,7 +52,7 @@ pub type ContextQueryParams = HashMap<String, String>;
 #[must_use]
 pub fn build_url_with_context_params(
     base_url: &str,
-    context: &HashMap<String, serde_json::Value>,
+    context: &HashMap<String, ContextValue>,
     mapping: &ContextQueryParams,
 ) -> String {
     let Ok(mut url) = url::Url::parse(base_url) else {
@@ -64,38 +82,28 @@ pub fn build_url_with_context_params(
     url.to_string()
 }
 
-/// Serialise a single [`serde_json::Value`] into a string suitable for a query
-/// parameter value.  Arrays are joined with commas; strings and numbers are
-/// returned directly; anything else yields an empty string (skipped).
-fn serialize_context_value(value: &serde_json::Value) -> String {
+/// Serialise a single [`ContextValue`] into a string suitable for a query
+/// parameter value.  String lists are joined with commas; strings and numbers
+/// are returned directly.
+fn serialize_context_value(value: &ContextValue) -> String {
     match value {
-        serde_json::Value::Array(arr) => arr
-            .iter()
-            .filter_map(|v| match v {
-                serde_json::Value::String(s) => Some(s.clone()),
-                serde_json::Value::Number(n) => Some(n.to_string()),
-                _ => None,
-            })
-            .collect::<Vec<_>>()
-            .join(","),
-        serde_json::Value::String(s) => s.clone(),
-        serde_json::Value::Number(n) => n.to_string(),
-        _ => String::new(),
+        ContextValue::StringList(items) => items.join(","),
+        ContextValue::Text(s) => s.clone(),
+        ContextValue::Number(n) => n.to_string(),
     }
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
-    use serde_json::json;
 
     #[test]
     fn test_build_url_with_context_params_appends_array() {
         let context = HashMap::from([(
             "permutive_segments".to_string(),
-            json!(["10000001", "10000003", "adv"]),
+            ContextValue::StringList(vec!["10000001".into(), "10000003".into(), "adv".into()]),
         )]);
-        let mapping = HashMap::from([("permutive_segments".to_string(), "permutive".to_string())]);
+        let mapping = BTreeMap::from([("permutive_segments".to_string(), "permutive".to_string())]);
 
         let url = build_url_with_context_params(
             "http://localhost:6767/adserver/mediate",
@@ -110,8 +118,11 @@ mod tests {
 
     #[test]
     fn test_build_url_with_context_params_preserves_existing_query() {
-        let context = HashMap::from([("permutive_segments".to_string(), json!(["123", "adv"]))]);
-        let mapping = HashMap::from([("permutive_segments".to_string(), "permutive".to_string())]);
+        let context = HashMap::from([(
+            "permutive_segments".to_string(),
+            ContextValue::StringList(vec!["123".into(), "adv".into()]),
+        )]);
+        let mapping = BTreeMap::from([("permutive_segments".to_string(), "permutive".to_string())]);
 
         let url = build_url_with_context_params(
             "http://localhost:6767/adserver/mediate?debug=true",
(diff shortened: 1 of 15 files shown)


aram356 (changes requested):

crates/js/lib/src/core/context.ts
```diff
@@ -0,0 +1,41 @@
...
+// without core needing integration-specific knowledge.
+import { log } from './log';
+
+/**
+ * A context provider returns key-value pairs to merge into the auction
+ * request's `config` payload, or `undefined` to contribute nothing.
+ */
+export type ContextProvider = () => Record<string, unknown> | undefined;
+
+const providers: ContextProvider[] = [];
```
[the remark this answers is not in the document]
ChristianPavilonis: Fixed — switched `providers` from `ContextProvider[]` to `Map<string, ContextProvider>`. `registerContextProvider` now takes an `(id, provider)` signature so re-registration replaces rather than appends. Updated the Permutive caller to pass `'permutive'` as the ID. Added a dedup test.

crates/common/src/auction/formats.rs
```diff
@@ -135,6 +134,26 @@ pub fn convert_tsjs_to_auction_request(
...
     });
 
+    // Forward all config entries from the JS request into the context map.
+    // Each integration's context provider contributes its own keys (e.g.
+    // permutive_segments, lockr_ids, …) — we pass them all through so
+    // auction providers can read whatever they need.
+    let mut context = HashMap::new();
+    if let Some(ref config) = body.config {
+        if let Some(obj) = config.as_object() {
+            for (key, value) in obj {
```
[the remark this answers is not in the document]
ChristianPavilonis: Fixed — added `allowed_context_keys: Vec<String>` to `AuctionConfig` (defaults to `["permutive_segments"]`). The context forwarding loop in `formats.rs` now only inserts keys present in this allowlist; everything else is dropped with a debug log. Configurable in `trusted-server.toml` under `[auction]`. Empty list = block all keys.

crates/common/src/auction/formats.rs
```diff
@@ -135,6 +134,31 @@ pub fn convert_tsjs_to_auction_request(
...
     });
 
+    // Forward allowed config entries from the JS request into the context map.
+    // Only keys listed in `auction.allowed_context_keys` are accepted;
+    // unrecognised keys are silently dropped to prevent injection of
+    // arbitrary data by a malicious client payload.
+    let allowed = &settings.auction.allowed_context_keys;
+    let mut context = HashMap::new();
+    if let Some(ref config) = body.config {
+        if let Some(obj) = config.as_object() {
```
aram356: 🔧 Use HashSet in place of loop

crates/common/src/auction/context.rs
```diff
@@ -0,0 +1,205 @@
...
+/// - Other JSON types are skipped.
+///
+/// The [`url::Url`] crate is used for construction so all values are
+/// percent-encoded, preventing query-parameter injection.
+///
+/// Returns the original `base_url` unchanged when no parameters are appended.
+#[must_use]
+pub fn build_url_with_context_params(
+    base_url: &str,
+    context: &HashMap<String, serde_json::Value>,
```
aram356: ❓ Should we use BTreeMap to preserve order

aram356 (changes requested):
Thanks for making changes. Few more things introduced with recent changes that needs addressing.

crates/common/src/auction/context.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    #[test]
+    fn test_serialize_context_value_number() {
+        assert_eq!(serialize_context_value(&ContextValue::Number(99.0)), "99");
+    }
+
+    #[test]
+    fn test_context_value_deserialize_array() {
+        let v: ContextValue = serde_json::from_str(r#"["a","b"]"#).unwrap();
```
aram356: **P1 — `unwrap()` in tests**: These three deserialization tests (lines 233, 239, 245) use `.unwrap()` instead of `.expect("should ...")` per project conventions.

**Fix**:
```rust
let v: ContextValue = serde_json::from_str(r#"["a","b"]"#).expect("should deserialize string list");
// ...
let v: ContextValue = serde_json::from_str(r#""hello""#).expect("should deserialize text");
// ...
let v: ContextValue = serde_json::from_str("42").expect("should deserialize number");
```

crates/common/src/auction/formats.rs
```diff
@@ -135,6 +135,46 @@ pub fn convert_tsjs_to_auction_request(
         geo: GeoInfo::from_request(req),
     });
 
+    // Forward allowed config entries from the JS request into the context map.
+    // Only keys listed in `auction.allowed_context_keys` are accepted;
+    // unrecognised keys are silently dropped to prevent injection of
+    // arbitrary data by a malicious client payload.
+    let allowed: HashSet<&str> = settings
+        .auction
+        .allowed_context_keys
```
aram356: **P2 — `HashSet` rebuilt per request**: `allowed_context_keys` is converted from `Vec<String>` to `HashSet<&str>` on every auction request. With a small allowlist this is negligible, but could be avoided by storing the allowlist as a `HashSet<String>` in `AuctionConfig` at deserialization time.

crates/common/src/integrations/adserver_mock.rs
```diff
@@ -256,8 +283,11 @@ impl AuctionProvider for AdServerMockProvider {
 
         log::debug!("AdServer Mock: mediation request: {:?}", mediation_req);
 
+        // Build endpoint URL with optional Permutive segments query string
```
aram356: **P2 — Comment names Permutive but code is generic**: The `context_query_params` mechanism is integration-agnostic, but this comment says "Permutive segments query string". Future integrations using the same mechanism would make this misleading.

**Fix**:
```rust
// Build endpoint URL with context-driven query parameters
```

crates/js/lib/src/integrations/permutive/segments.ts
```diff
@@ -0,0 +1,62 @@
...
+            count: segments.length,
+          });
+          return segments
+            .filter((s: unknown) => typeof s === 'string' || typeof s === 'number')
+            .map(String)
+            .slice(0, MAX_SEGMENTS);
+        }
+      }
+    }
+  } catch {
```
aram356: **P2 — Bare `catch` swallows all errors**: This catches everything (JSON parse failures, property access errors, out-of-memory) and logs "failed to read from localStorage" which isn't accurate for all error types.

**Fix**: Consider logging the error object for debuggability:
```ts
} catch (err) {
  log.debug('getPermutiveSegments: failed to read segments', err);
}
```

aram356 (commented):
## Summary

👍 Well-designed feature that forwards Permutive audience segments from the client through the auction pipeline to the ad server as URL query parameters. The architecture is clean — a generic context provider registry on the JS side and a config-driven `allowed_context_keys` allowlist on the Rust side prevent tight coupling between integrations and the auction system. Good security posture with server-side allowlisting.

## Findings

### 🔧 High (P1)

- **`unwrap()` in tests**: `context.rs:233,239,245` — three deserialization tests use `.unwrap()` instead of `.expect("should ...")`. (inline comment)

### 🤔 Medium (P2)

- **`ContextValue::Number(f64)` precision**: `context.rs:24` — works for current string-based segments, but `f64` loses precision for integers above 2^53. Low risk now but worth noting.
- **`HashSet` rebuilt per request**: `formats.rs:144` — `allowed_context_keys` is converted from `Vec` to `HashSet` on every auction request. Negligible with a small list, but could be stored as a `HashSet` in `AuctionConfig`. (inline comment)
- **Comment names Permutive but code is generic**: `adserver_mock.rs:286` — the `context_query_params` mechanism is integration-agnostic, but the comment says "Permutive segments". (inline comment)
- **Bare `catch` swallows all errors**: `segments.ts:58` — catches everything including unexpected errors, logs a misleading message. (inline comment)

### ⛏ Low (P3)

- **Test names use `test_` prefix**: `context.rs`, `adserver_mock.rs`, `settings.rs` — consistent with existing codebase style.
- **No end-to-end test for context provider registration**: `segments.test.ts` covers `getPermutiveSegments()` but the `registerContextProvider` call in `index.ts` isn't tested e2e (registry itself is tested in `context.test.ts`).

## CI Status

- cargo fmt: PASS
- cargo test: PASS
- vitest: PASS
- CodeQL: PASS

aram356 (approved):
👍 Good to merge. Need to resolve conflicts.

prk-Jr (approved):
👍 Looks good

Issue #130, opened 2025-11-25
As publisher I want to use Kargo SSP through Prebid server


aram356:
Make sure that all PrebidJS parameters pass through to TS to Prebid Server

PR #276, opened 2026-02-10
fix: forward browser user-agent in OpenRTB `device.ua` field

### LINKED ISSUE #264: Prebid OpenRTB requests missing `device.ua`, causing bidders to return no-bid
Bidders (confirmed with Kargo) are returning 204 No Content for all OpenRTB requests sent through Prebid Server. The root cause is that the `device.ua` field is never populated in the OpenRTB bid request payload.

 When `device.ua` is absent, Prebid Server falls back to the `User-Agent` HTTP header of the server-to-server request itself (e.g. `curl/8.7.1` or the Fastly compute runtime UA), which bidders reject as non-browser traffic.

 ### Reproduction

 Sending the same OpenRTB request directly to Kargo with a real browser UA returns 200 (bid). With a non-browser UA it returns 204 (no bid). All other fields (banner formats, buyeruid, geo) had no effect.

 | Test | Change | Result |
 |------|--------|--------|
 | Baseline (as sent by Prebid) | None | 204 |
 | + 10 banner formats | Formats only | 204 |
 | + Real Chrome UA | UA only | **200** |
 | + buyeruid | buyeruid only | 204 |

 ### Root Cause

 Two issues:

1. The OpenRTB `Device` struct has no `ua` field, so user-agent is never serialized into the bid request JSON.
2. `DeviceInfo` construction in `formats.rs` is gated on geo lookup succeeding -- if `GeoInfo::from_request` returns `None`, the entire `DeviceInfo` (including user-agent) is dropped.

### Expected Behavior

The original browser `User-Agent` header should be forwarded as `device.ua` in every OpenRTB bid request, regardless of whether geo information is available.

ChristianPavilonis:
closes #264, #130 

### Summary

- Add `ua` field to the OpenRTB `Device` struct so user-agent is serialized into bid requests
 - Update `to_openrtb` in the Prebid integration to map `DeviceInfo.user_agent` into `device.ua`
- Fix `DeviceInfo` construction to always include user-agent and IP, not just when geo lookup succeeds

### Context

 Bidders like Kargo filter out requests with non-browser user-agents. Because `device.ua` was never set, Prebid Server would fall back to the HTTP UA of the server-to-server call (e.g. the Fastly runtime), causing bidders to return 204 No Content on every request. Verified via isolated curl testing that UA is the sole deciding factor.

 ### Changes

 | File | Change |
 |------|--------|
 | `crates/common/src/openrtb.rs` | Added `ua: Option<String>` to `Device` |
 | `crates/common/src/integrations/prebid.rs` | Map `user_agent` into `device.ua`; use `.map()` instead of `.and_then()` so device is present even without geo |
 | `crates/common/src/auction/formats.rs` | Always construct `Some(DeviceInfo { .. })` with geo as optional inside, instead of gating on geo lookup |

### Refactoring

| File | Change |
|------|--------|
| `crates/common/src/streaming_processor.rs` | Extract shared `decompress_and_process` helper to deduplicate identical decompress→process→write logic across gzip, deflate, and brotli `*_to_none` methods |
| `crates/common/src/proxy.rs` | Extract `finalize_response` helper to consolidate repeated `stream_passthrough` branching (4 call sites) |
| `crates/common/src/integrations/prebid.rs` | `vec![]` → `[]` for static `cdn_patterns` array |
| `crates/common/src/auction/formats.rs` | Use `HashMap` directly instead of `std::collections::HashMap` |

### DIFF
--- a/crates/common/src/auction/formats.rs
+++ b/crates/common/src/auction/formats.rs
@@ -108,7 +108,7 @@ pub fn convert_tsjs_to_auction_request(
                 }
 
                 // Extract bidder params from the bids array
-                let mut bidders = std::collections::HashMap::new();
+                let mut bidders = HashMap::new();
                 if let Some(bids) = &unit.bids {
                     for bid in bids {
                         bidders.insert(bid.bidder.clone(), bid.params.clone());
@@ -119,20 +119,20 @@ pub fn convert_tsjs_to_auction_request(
                     id: unit.code.clone(),
                     formats,
                     floor_price: None,
-                    targeting: std::collections::HashMap::new(),
+                    targeting: HashMap::new(),
                     bidders,
                 });
             }
         }
     }
 
-    // Get geo info if available
-    let device = GeoInfo::from_request(req).map(|geo| DeviceInfo {
+    // Build device info with user-agent (always) and geo (if available)
+    let device = Some(DeviceInfo {
         user_agent: req
             .get_header_str("user-agent")
             .map(std::string::ToString::to_string),
         ip: req.get_client_ip_addr().map(|ip| ip.to_string()),
-        geo: Some(geo),
+        geo: GeoInfo::from_request(req),
     });
 
     Ok(AuctionRequest {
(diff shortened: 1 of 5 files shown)


aram356 (changes requested):

aram356:
@jevansnyc Per https://www.iab.com/wp-content/uploads/2016/03/OpenRTB-API-Specification-Version-2-5-FINAL.pdf

We need to think about passing some of the additional parameters. Can we document this in another ticket?

aram356 (approved):
👍 Looks good

PR #321, opened 2026-02-18
perf: stream HTML chunks through lol_html without full-document buffering

### LINKED ISSUE #320: Enable Streaming Chunks for responses to improve TTFB on TS
What it does: Flush <head> to the browser immediately. Run the auction concurrently. Stream the body as origin HTML arrives, inject auction results when they're ready.

What changes in TS: The process_response_with_pipeline function in [proxy.rs:141-169] currently buffers the entire body before returning. This needs to become a true streaming pipeline, parse and forward HTML in chunks, holding only the ad-config injection point. Fastly's Rust SDK supports streaming response bodies via Body streaming APIs so we should be good. 

Browser caching changes: With streaming, the response headers go out immediately, so something like this:

Cache-Control: max-age=60, stale-while-revalidate=300
Surrogate-Control: max-age=300, stale-while-revalidate=3600
Browser caches for 60s, serves stale (instantly) for 300s while TS revalidates in background. On Fastly's side, CDN holds it 300s with 1hr SWR. Real users see TTFB of edge cache latency (~5-30ms) on virtually every visit.

jevansnyc:
## Summary

- HtmlRewriterAdapter now uses lol_html's incremental streaming API instead of accumulating the entire response body before processing. The HtmlRewriter is created eagerly in new() with a shared RcVecSink output buffer; each process_chunk call writes directly to the live rewriter and drains whatever lol_html has ready, so the prefix of the document (everything up to the first matched element) flows downstream immediately rather than waiting for the last byte.
- process_gzip_to_gzip now delegates to the existing process_through_compression helper (same as deflate and brotli already did), eliminating the read_to_end that buffered the entire decompressed body into a Vec before processing.
- Tests updated: intermediate chunks may now carry data, so assertions collect output across all chunks rather than asserting intermediates are always empty.

## What this does not change yet (follow-up)

The full TTFB improvement requires two more changes tracked for a follow-up PR:
- publisher.rs: let mut output = Vec::new() still collects the complete processed body before Fastly sends response headers downstream
- publisher.rs: req.send() is still synchronous; need send_async() + StreamingBody to push bytes to the client as they arrive from origin

## Test plan

- [ ] cargo check -p trusted-server-common passes (confirmed locally)
- [ ] CI passes
- [ ] Manual: deploy to staging and confirm TTFB metrics improve for HTML responses vs baseline

### DIFF
--- a/crates/common/src/streaming_processor.rs
+++ b/crates/common/src/streaming_processor.rs
@@ -7,7 +7,9 @@
 //! - UTF-8 boundary handling
 
 use error_stack::{Report, ResultExt};
+use std::cell::RefCell;
 use std::io::{self, Read, Write};
+use std::rc::Rc;
 
 use crate::error::TrustedServerError;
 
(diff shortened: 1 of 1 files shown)


aram356:
@jevansnyc ⚠️ This not my no as trivial because

this will break the RSC Next.js integration for HTML responses that contain self.__next_f.push() scripts with origin URLs. The post-processing phase that rewrites RSC payloads will receive only the tail of the document instead of the full output, causing placeholder strings to leak into the response and breaking React hydration.

Possible fix

The HtmlWithPostProcessing wrapper needs to accumulate all output across chunks when post-processors are registered, then run post-processing on the complete document at is_last=true. Alternatively, the post-processing could be restructured to work incrementally — but that's a larger change given the cross-script T-chunk combining logic.

PR #350, opened 2026-02-20
Optimize Settings::from_toml() call

prk-Jr:
## Summary

Optimize `Settings::from_toml()` by removing redundant `config` crate env var scanning at runtime. WASM guest profiling confirmed: **~4.8% → ~2.7% CPU per request**.

### Why this is safe

`build.rs` already resolves all configuration at **compile time**:
1. Reads `trusted-server.toml`
2. Merges `TRUSTED_SERVER__*` env vars via the `config` crate
3. Writes the fully-resolved result to `target/trusted-server-out.toml`
4. Registers `cargo:rerun-if-env-changed` for all possible env vars

At runtime, `get_settings()` loads this pre-resolved TOML via `include_bytes!`. The old `from_toml()` was re-running the entire `config` crate pipeline (env source scanning, config merging, deserialization) on every request — completely redundant since the embedded TOML already contains the final values. On Fastly Compute (WASM sandbox), there are no environment variables available at runtime anyway.

### What changed

- **`from_toml()`** now uses direct `toml::from_str()` deserialization — no `config` crate, no env var scanning
- **`build.rs::merge_toml()`** now owns the env var merging logic inline using the `config` crate directly, instead of delegating to `from_toml()`
- **`config` crate import** in `settings.rs` is now `#[cfg(test)]` — not compiled into the production binary

### Why some tests use `from_toml_with_env()`

`build.rs` runs once before tests start, producing a single static `target/trusted-server-out.toml`. But 7 tests need to verify that different `TRUSTED_SERVER__*` env vars correctly override different TOML values per test (e.g., overriding `publisher.origin_url` in one test, `integrations.prebid.bidders` in another). They can't rely on `build.rs` for this because:

- `build.rs` runs once, not per-test
- Each test sets different env vars via `temp_env` that `build.rs` never sees
- The resolved output file can't reflect per-test env var combinations

So these 7 tests call `from_toml_with_env()` directly — a `#[cfg(test)]` helper that preserves the original env var merging behavior. This lets them test the merging contract in isolation without going through `build.rs`.

All other tests (20+) use the new fast `from_toml()` and continue passing — they only test TOML parsing, not env var overrides.

### Files changed

| File | Change |
|------|--------|
| `settings.rs` | `from_toml()` → `toml::from_str`; added `#[cfg(test)] from_toml_with_env()`; `config` import → `#[cfg(test)]` |
| `build.rs` | Env var merging logic moved here (was in `from_toml()`) |
| `Cargo.toml` | Added `toml` to `[dependencies]` |

## Test plan

- [x] `cargo test` — 352 tests pass
- [x] WASM build succeeds
- [x] WASM guest profiling confirms ~4.8% → ~2.7% CPU for `get_settings()`
- [ ] Deploy to staging, run `BENCH_URL=<staging> ./scripts/benchmark.sh` and compare against baseline
- [ ] Team review: verify `build.rs` env var merging matches original `from_toml()` behavior

### DIFF
--- a/crates/common/Cargo.toml
+++ b/crates/common/Cargo.toml
@@ -41,6 +41,7 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 sha2 = { workspace = true }
 tokio = { workspace = true }
+toml = { workspace = true }
 trusted-server-js = { path = "../js" }
 url = { workspace = true }
 urlencoding = { workspace = true }
--- a/crates/common/build.rs
+++ b/crates/common/build.rs
@@ -43,6 +43,8 @@ fn rerun_if_changed() {
 }
 
 fn merge_toml() {
+    use config::{Config, Environment, File, FileFormat};
+
     // Get the OUT_DIR where we'll copy the config file
     let dest_path = Path::new(TRUSTED_SERVER_OUTPUT_CONFIG_PATH);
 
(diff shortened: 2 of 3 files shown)


aram356:
@prk-Jr Thanks for these findings. I also found bunch of this issue in config parsing so opened PR with of this plus bug fixed #362

PR #351, opened 2026-02-20
Feat/optimize html streaming

### LINKED ISSUE #320: Enable Streaming Chunks for responses to improve TTFB on TS

prk-Jr:
## Summary

This PR combines the core publisher-proxy streaming optimization with the
Next.js RSC follow-up work.

At the platform level, Trusted Server moves from a fully buffered proxy model
to chunked streaming using Fastly `stream_to_client()`, enabling early header
flush and incremental HTML delivery to reduce TTFB and improve subresource
discovery.

On top of that foundation, the HTML pipeline now supports RSC-aware lazy
accumulation: non-RSC content continues to stream immediately, while only RSC
content that requires post-processing is buffered and rewritten safely. This
preserves correctness for fragmented/cross-script RSC payloads while restoring
meaningful streaming behavior.

---

## Key Changes

* **`stream_to_client()` Integration (`publisher.rs`)**  
  Replaced fully buffered response collection with `stream_to_client()` to
  enable immediate header dispatch and incremental chunk streaming.

* **`lol_html` Output Pipeline (`streaming_processor.rs`)**  
  Refactored the `HtmlRewriter` adapter to implement the `OutputSink` trait
  with a shared `Rc<RefCell<Vec<u8>>>`, enabling true incremental streaming.

* **Buffer Pre-allocation**  
  Replaced `std::mem::take` with `Vec::with_capacity` and
  `std::mem::replace` to reduce reallocation churn during chunk processing.

* **WASM Hostcall Batching**  
  Wrapped `StreamingBody` output in an 8KB `std::io::BufWriter` to reduce
  WASM-to-host boundary crossings.

* **RSC Lazy Accumulation (`html_processor.rs`)**  
  Added conditional accumulation mode that starts buffering only when
  post-processing is required (for example, RSC placeholders or fragmented
  scripts). Non-RSC pages continue streaming instead of being fully buffered.

* **RSC Post-processing Triggers (`nextjs` integration)**  
  Added `needs_accumulation` support to integration post-processors and
  `needs_post_processing` detection in placeholder state, including fragmented
  script tracking for fallback re-parse correctness.

* **Memory Safety Guardrail**  
  Added a 10MB cap for accumulated post-processed HTML to avoid unbounded
  memory growth on large/malicious documents.

* **Routing and Header Consistency (`fastly/src/main.rs`, `publisher.rs`)**  
  Centralized route classification and standardized response-header application
  across buffered and streaming paths.

* **RSC Fixture/Test Expansion**  
  Added fixture-driven Next.js integration tests (including real Next.js output)
  plus a dedicated example app and scripts for fixture capture and live streaming
  validation.

* **Code Health**  
  Resolved associated Clippy warnings and added missing `# Errors`
  documentation in streaming-related handlers.

---

## Test Plan

* [x] **Local Unit & Workspace Tests**  
  Run:

  ```bash
  cargo test --workspace
  ```

* [x] **TypeScript Bundle Build**  
  Run:

  ```bash
  npm run build
  ```

  in `crates/js/lib` to verify successful generation of integration modules.

* [x] **Next.js RSC Integration Tests**  
  Run:

  ```bash
  cargo test --test nextjs_integration -- --nocapture
  ```

  to validate URL rewriting correctness and streaming behavior across fixture
  sets/chunk sizes.

* [x] **Local Fastly Simulation**  
  Run:

  ```bash
  fastly compute serve
  ```

  Verify:

  * Headers are correctly injected on streamed responses
  * Proxy behavior remains correct
  * Baseline TTFB improvements (for example, via `curl`)

* [x] **Staging Load Testing**  
  Execute:

  ```bash
  ./scripts/benchmark.sh
  ```

  against staging to quantify external TTFB and Time-to-Last-Byte (TTLB)
  improvements under concurrent traffic.
  
  # Closes
closes #320


prk-Jr:
# Performance Benchmark: HTML Streaming Optimization

We ran a comprehensive apples-to-apples benchmark to measure the impact of the `feat/optimize-html-streaming` branch (which introduces `lol_html` for streaming `<body>` transformations instead of buffering). 

To ensure statistical accuracy:
- We increased the sample size to **50 requests**.
- We ran a **10-request deep warmup** to eliminate cold-start WASM instantiations.
- We tested both branches on **Staging** and **Production** directly to isolate environment variables.

---

## 🚀 The Results: Production
This is the true impact on live users hitting the Fastly Edge:

| Metric | Baseline ([main](file:///Users/prk-jr/Desktop/opensource/rust/trusted-server/scripts/benchmark.sh#340-372)) | Optimization (`feat/...`) | Net Impact |
|--------|-------------|---------------|------|
| **First Byte (Median TTFB)** | 160.09 ms | 144.75 ms | 🟢 **15.34 ms Faster** |
| **First Byte (p95 Tail)** | 252.78 ms | 228.93 ms | 🟢 **23.85 ms Faster** |
| **Total Transfer Time** | 315.19 ms | 345.20 ms | 🔶 **30.01 ms Slower** |

---

## 📉 The Results: Staging 
*(Note: Total times are higher here because Staging serves a 190KB uncompressed JS bundle, whereas Prod serves a minified 28KB bundle).*

| Metric | Baseline ([main](file:///Users/prk-jr/Desktop/opensource/rust/trusted-server/scripts/benchmark.sh#340-372)) | Optimization (`feat/...`) | Net Impact |
|--------|---------------|------------------|------|
| **First Byte (Median TTFB)** | 220.44 ms | 217.38 ms | 🟢 **3.06 ms Faster** |
| **First Byte (p95 Tail)** | 478.48 ms | 277.93 ms | 🟢 **200.55 ms Faster** |
| **Total Transfer Time** | 505.28 ms | 654.58 ms | 🔶 **149.30 ms Slower** |

---

## 🎯 Conclusion
The `lol_html` streaming processor behaves exactly as architecturally intended:

1. **Massive Win for Core Web Vitals:** Because we no longer wait for the backend to buffer the entire HTML document before transmitting, the Fastly edge begins sending the `<head>` tag to the user's browser **15ms sooner on average** (and up to 200ms sooner in worst-case staging scenarios). This means the browser can start downloading CSS/JS assets much faster.
2. **Acceptable CPU Overhead:** Streaming chunk-by-chunk through the WASM boundary does consume more CPU time. On production hardware, this means the page finishes loading about **30ms later**. 

Exchanging 30ms of trailing transfer time for 15-20ms of upfront TTFB savings is a highly favorable trade for perceived performance. **This branch is safe and recommended for merge.**

PR #390, opened 2026-03-02
Add OpenRTB 2.6 proto types

ChristianPavilonis:
## Summary

- **Replace hand-written OpenRTB types with spec-compliant generated types** from the official IAB Tech Lab protobuf schema, ensuring full OpenRTB 2.6 coverage including DOOH, GPP, Qty, Refresh, and DurFloors fields.
- **Enrich Prebid bid requests** with many more OpenRTB fields derived from available auction data — bid floors, device signals, user consent, geo, regs/privacy, site referrer, and request-level defaults.

## Changes

| File | Change |
| ---- | ------ |
| `crates/openrtb/Cargo.toml` | New crate (edition 2024) with serde, serde_json — no prost runtime dep |
| `crates/openrtb/proto/openrtb.proto` | OpenRTB 2.6 proto2 schema from IAB Tech Lab upstream |
| `crates/openrtb/src/generated.rs` | Checked-in generated code (1566 lines) — no build-time protoc needed |
| `crates/openrtb/src/codegen.rs` | Postprocessing logic (strip prost derives, inject serde, add ext fields) with 7 unit tests |
| `crates/openrtb/src/lib.rs` | `bool_as_int` serde module (6 edge-case tests), `ToExt` supertrait, explicit re-exports, deserialization tests |
| `crates/openrtb/generate.sh` | Bash wrapper to regenerate from proto using the codegen crate |
| `crates/openrtb/README.md` | Crate documentation |
| `crates/openrtb-codegen/` | Standalone codegen crate (excluded from workspace) with `prost-build` dependency |
| `crates/common/src/openrtb.rs` | Replace ~180 lines of hand-written structs with type aliases; shared `to_openrtb_i32()` helper; opt-in `ToExt` impls for 4 ext types |
| `crates/common/src/geo.rs` | `GDPR_COUNTRIES` set (EU/EEA/UK), `is_gdpr_country()` function with 5 tests |
| `crates/common/src/integrations/prebid.rs` | Enrich bid requests with bid floors, device signals, consent, geo, regs, site fields; `DEFAULT_CURRENCY` and `GPC_US_PRIVACY` constants; language subtag stripping; typed `get_prebid_ext()` test helper; 17+ tests |
| `crates/common/src/auction/formats.rs` | Update response construction for generated types with `Option` wrappers and `ToExt` |
| `crates/common/Cargo.toml` | Add `trusted-server-openrtb` dependency |
| `Cargo.toml` | Add `openrtb` to workspace members, exclude `openrtb-codegen`, remove `prost-build` from workspace deps |
| `Cargo.lock` | Dependency additions (serde derives for openrtb) |
| `.github/workflows/format.yml` | Removed protoc install step (no longer needed) |
| `.github/workflows/test.yml` | Removed protoc install step (no longer needed) |

## Key design decisions

- **No prost runtime dependency**: Generated code has `prost::Enumeration` and alloc types stripped during postprocessing, so the openrtb crate only depends on serde.
- **Checked-in generated code**: Avoids requiring `protoc` in CI/dev environments. Regenerate via `./generate.sh` when the proto changes.
- **GDPR via geo check**: Uses Fastly `geo.country_code()` to determine GDPR applicability (EU/EEA/UK country list) rather than parsing TCF consent strings. Falls back to consent string presence when geo is unavailable.
- **`ToExt` as supertrait**: `ToExt: Serialize` with opt-in impls on the 4 Prebid ext types, preventing accidental misuse.

## Test plan

- [x] `cargo test --workspace` — all tests pass


aram356 (commented):
_(superseded — see latest review below)_

crates/openrtb/build.rs
```diff
@@ -0,0 +1,260 @@
...
+    "Bid",
+    "Transparency",
+];
+
+/// Post-process prost-generated Rust code.
+///
+/// Strips `::prost::Message` derives and `#[prost(...)]` field attributes (we
+/// don't need protobuf binary encoding), then adds serde derives, skip
+/// annotations, and `ext` fields.
+fn postprocess(code: &str) -> String {
```
aram356: 🔧 **Fragile text-munging codegen, no tests**

The entire `postprocess()` function is ~180 lines of string-level manipulation — matching on substrings like `trimmed.contains("::prost::Message)"]` and `trimmed.starts_with("#[derive(Clone,")`. This works today but is brittle:

- **Prost version upgrades** can change derive ordering, attribute formatting, or whitespace — silently producing broken output.
- **Brace-counting** for ext injection (lines 179-207) is fragile. A doc comment containing `{` would throw off depth tracking.
- There are no tests for `postprocess()` itself — no guardrails for bool adapter placement, ext injection placement, or derive replacement correctness.

**Suggestion:** Add at least one `#[test]` for `postprocess()` with a representative prost snippet to catch regressions. Even better, consider using `syn` to parse as AST rather than text, or commit the generated file and only re-generate on demand (eliminates the `protoc` dependency entirely).

crates/common/src/auction/formats.rs
```diff
@@ -217,6 +217,34 @@ pub fn convert_to_openrtb_response(
             })
         })?;
 
+        let width = match i32::try_from(bid.width) {
```
aram356: 🔧 **Duplicated `u32 → i32` conversion logic**

This inline `match i32::try_from(...)` is identical in structure to `to_openrtb_i32()` in `prebid.rs:468-481`. The only difference is the log message format.

**Suggestion:** Move `to_openrtb_i32` (or a variant that takes a context string) to the `openrtb` module or a shared utility, and use it from both call sites. This also reduces the line count here significantly — the width + height blocks are 26 lines that could be 2.

crates/common/src/integrations/prebid.rs
```diff
@@ -571,35 +605,110 @@ impl PrebidAuctionProvider {
...
+        // authoritative `isSubjectToGDPR` signal.
+        //
+        // us_privacy is set when Sec-GPC header signals opt-out.
+        let gdpr_from_geo = request
+            .device
+            .as_ref()
+            .and_then(|d| d.geo.as_ref())
+            .map(|geo| is_gdpr_country(&geo.country));
+
+        let gdpr = match gdpr_from_geo {
```
aram356: 🔧 **P1 — GDPR can be forced to `false` even when a consent string exists**

When geo resolves to a non-GDPR country, `gdpr_from_geo` is `Some(false)`, which overrides the conservative fallback — even if `user.consent` contains a valid TCF string. The test `to_openrtb_sets_gdpr_false_for_non_eu_country` explicitly locks in this behavior.

Risk: IP geo false-negatives (VPN, mobile carrier, IP DB drift) can mark EU traffic as non-GDPR and weaken compliance signaling. A user in Berlin on a US mobile carrier IP gets `gdpr=0` despite having a TCF consent string from a CMP.

Suggested fix — when geo says non-GDPR but a consent string is present, the consent string is the stronger signal:

```rust
let gdpr = match gdpr_from_geo {
    Some(true) => Some(true),
    Some(false) if request.user.consent.is_some() => Some(true),
    Some(false) => Some(false),
    None => request.user.consent.as_ref().map(|_| true),
};
```

crates/openrtb/src/lib.rs
```diff
@@ -0,0 +1,348 @@
...
+        }
+    }
+
+    pub fn deserialize<'de, D: Deserializer<'de>>(
+        deserializer: D,
+    ) -> Result<Option<bool>, D::Error> {
+        let value = Option::<serde_json::Value>::deserialize(deserializer)?;
+        match value {
+            Some(serde_json::Value::Bool(b)) => Ok(Some(b)),
+            Some(serde_json::Value::Number(n)) => Ok(Some(n.as_i64() != Some(0))),
```
aram356: 🤔 **P2 — `bool_as_int` deserialization misinterprets floating-point zero**

`n.as_i64() != Some(0)` silently treats `0.0` as `true`. `serde_json::Number` returns `None` from `as_i64()` for `0.0` (stored as a float), so the expression evaluates to `true`.

While unlikely in well-formed OpenRTB, malformed payloads do appear in the wild. Suggested fix:

```rust
Some(serde_json::Value::Number(n)) => {
    Ok(Some(n.as_i64().map_or_else(
        || n.as_f64().is_some_and(|f| f != 0.0),
        |i| i != 0,
    )))
}
```

crates/common/src/integrations/prebid.rs
```diff
@@ -553,11 +575,23 @@ impl PrebidAuctionProvider {
                 }
 
                 Imp {
```
aram356: 🤔 **P2 — Invalid banner imps can still be emitted after dimension overflow filtering**

When all formats for a slot have out-of-range dimensions, `formats` is empty but the `Imp` is still built with `banner: Some(Banner { format: vec![], .. })`. An impression with an empty format list is malformed per OpenRTB — some PBS adapters may reject it, reducing fill.

Consider changing `.map()` to `.filter_map()` and returning `None` when formats is empty:

```rust
.filter_map(|slot| {
    // ... format collection ...
    if formats.is_empty() {
        log::warn!("prebid: dropping imp '{}' — no valid banner formats", slot.id);
        return None;
    }
    Some(Imp { ... })
})
```

crates/openrtb/src/codegen.rs
```diff
@@ -0,0 +1,375 @@
...
+    "Uid",
+    "Data",
+    "Segment",
+    "Regs",
+    "DurFloors",
+    "BidResponse",
+    "SeatBid",
+    "Bid",
+    "Transparency",
+];
```
aram356: **minor**: `NativeRequest` and `NativeResponse` (and their nested sub-messages like `Asset`, `Title`, `ImageAsset`, etc.) are not in `EXT_STRUCTS`, so any `ext` data on native ad payloads would be silently dropped during deserialization. Not blocking since native media type isn't supported yet (`supports_media_type` only returns `true` for `Banner`), but worth a tracking note for when native support is added.

aram356 (commented):
Staff-engineer review of OpenRTB full implementation. 1 blocker, 5 significant issues, 4 minor issues -- all posted as inline comments.

aram356 (approved):
👍 Looks good

PR #442, opened 2026-03-05
Add integration testing with testcontainers and Playwright

### LINKED ISSUE #398: Integration tests: WordPress and Next.js frontends with Testcontainers
### Description

Add integration tests using Testcontainers that verify the trusted server
operates correctly against real frontend applications.

## WordPress frontend

- Spin up a WordPress container with a minimal theme
- Configure the trusted server to proxy/inject into the WordPress site
- Verify:
  - HTML response contains injected `<script>` tags
  - `/static/tsjs=` endpoint serves concatenated JS bundles
  - Integration-specific attributes are rewritten in ad markup
  - GDPR consent signals propagate correctly

## Next.js frontend

- Spin up a Next.js container with a minimal app
- Configure the trusted server to proxy/inject into the Next.js site
- Verify:
  - Server-side rendered pages receive script injection
  - Client-side navigation does not break after injection
  - Deferred script tags load correctly on route transitions
  - `/static/tsjs=` endpoint serves valid JS

## Shared concerns

- Tests should be runnable locally and in CI
- Each test suite should start, exercise, and tear down containers
  automatically
- Use Testcontainers lifecycle management for setup and teardown
- Document any required Docker images or build steps in the test README

### Done when

Running CI pipeline

### Affected area

CI / Tooling

### Additional context

_No response_

prk-Jr:
## Summary

- Add end-to-end integration testing infrastructure using testcontainers for HTTP-level tests and Playwright for browser tests against WordPress and Next.js containers
- Add Playwright browser tests verifying script injection, SPA navigation, deferred script execution, API passthrough, and form URL rewriting in a real Chromium browser
- Introduce a `/health` endpoint and CI workflow for running both test suites separately from the main CI

## Changes

| File | Change |
| ---- | ------ |
| `crates/fastly/src/main.rs` | Add `/health` GET endpoint returning 200 with "ok" body |
| `crates/integration-tests/Cargo.toml` | New crate with reqwest, testcontainers, and tokio dependencies |
| `crates/integration-tests/tests/common/` | Core test infrastructure: runtime traits, config builder, assertion helpers including `assert_form_action_rewritten` (9 unit tests) |
| `crates/integration-tests/tests/environments/` | Fastly/Viceroy runtime environment with dynamic port allocation |
| `crates/integration-tests/tests/frameworks/` | Framework abstractions for WordPress and Next.js with 7 standard + 5 custom scenarios |
| `crates/integration-tests/tests/integration.rs` | Matrix test runner (`test_all_combinations`, per-framework tests) |
| `crates/integration-tests/fixtures/configs/` | Viceroy and Fastly config templates with KV stores and secrets |
| `crates/integration-tests/fixtures/frameworks/wordpress/` | WordPress Docker image: PHP built-in server with minimal test theme and `/wp-admin/` stub |
| `crates/integration-tests/fixtures/frameworks/nextjs/` | Next.js 14 Docker image: 4 pages, API routes, forms, shared navigation with hydration signal, deferred route scripts |
| `crates/integration-tests/browser/` | Playwright test suite: global setup/teardown, Docker + Viceroy infra helpers with cleanup-on-failure |
| `crates/integration-tests/browser/tests/shared/` | Cross-framework browser tests: script injection, script bundle loading |
| `crates/integration-tests/browser/tests/nextjs/` | Next.js browser tests: 4-page SPA navigation chain, deferred script execution, API passthrough, form URL rewriting |
| `crates/integration-tests/browser/tests/wordpress/` | WordPress browser tests: admin page script injection |
| `.github/workflows/integration-tests.yml` | CI workflow with parallel HTTP-level and browser test jobs, split artifact uploads for Playwright reports and traces |
| `scripts/integration-tests.sh` | Shell script for HTTP-level test orchestration (WASM build + Docker + test run) |
| `scripts/integration-tests-browser.sh` | Shell script for browser test orchestration (WASM build + Docker + Playwright install + test run) |
| `crates/integration-tests/README.md` | Full documentation: prerequisites, quick start, test scenario tables, architecture overview |
| `INTEGRATION_TESTS_PLAN.md` | Design document with architecture decisions and progress tracking |
| `Cargo.toml` | Add integration-tests to workspace members |
| `Cargo.lock` | Updated lockfile with new dependencies |

### HTTP-level (Rust + testcontainers)

| Scenario | Frameworks |
| ---- | ------ |
| `HtmlInjection` | WordPress, Next.js |
| `ScriptServing` | WordPress, Next.js |
| `AttributeRewriting` | WordPress, Next.js |
| `ScriptServingUnknownFile404` | WordPress, Next.js |
| `WordPressAdminInjection` | WordPress |
| `NextJsRscFlight` | Next.js |
| `NextJsServerActions` | Next.js |
| `NextJsApiRoute` | Next.js |
| `NextJsFormAction` | Next.js |

### Browser-level (Playwright + Chromium)

| Spec | Frameworks |
| ---- | ------ |
| `script-injection` | WordPress, Next.js |
| `script-bundle` | WordPress, Next.js |
| `navigation` (4-page SPA chain + back button + deferred route script) | Next.js |
| `api-passthrough` | Next.js |
| `form-rewriting` | Next.js |
| `admin-injection` | WordPress |

**Totals: 12 Next.js + 5 WordPress browser tests, 0 flaky**

## Known gaps

- **GDPR consent propagation** — the consent field exists in `AuctionRequest` but is not yet populated upstream. Requires implementation before it can be tested.
- **Browser suite uses manual Docker orchestration** — HTTP-level tests use the Rust `testcontainers` crate; browser tests use `docker run`/`docker stop` via Playwright's globalSetup/globalTeardown (Node.js cannot use the Rust crate).

## Closes

Closes #398

## Test plan

- [x] `cargo test --workspace` (excluding integration-tests, which requires native target)
- [x] `cargo clippy --all-targets --all-features -- -D warnings` (excluding integration-tests)
- [x] JS tests: pre-existing ESM/CJS failure on main (not introduced by this branch)
- [x] JS format: `cd crates/js/lib && npm run format` -- passes
- [x] Docs format: `cd docs && npm run format` -- passes
- [x] Integration-tests crate compiles for native target: `cargo test -p integration-tests --target aarch64-apple-darwin --no-run`
- [x] HTTP integration tests: `./scripts/integration-tests.sh` -- all pass
- [x] Browser integration tests: `./scripts/integration-tests-browser.sh` -- 17 pass, 0 fail
- [x] Cargo clippy on integration-tests: `cargo clippy --manifest-path crates/integration-tests/Cargo.toml --tests` -- clean

## Checklist

- [x] New code has tests (9 unit tests for assertions and config, 17 browser tests)

### DIFF
--- a/crates/common/src/integrations/gpt.rs
+++ b/crates/common/src/integrations/gpt.rs
@@ -36,19 +36,20 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use error_stack::{Report, ResultExt};
-use fastly::http::{header, Method, StatusCode};
+use fastly::http::header;
 use fastly::{Request, Response};
 use serde::{Deserialize, Serialize};
 use url::Url;
 use validator::Validate;
 
-use crate::backend::BackendConfig;
+use crate::constants::{HEADER_ACCEPT, HEADER_ACCEPT_ENCODING, HEADER_ACCEPT_LANGUAGE};
 use crate::error::TrustedServerError;
 use crate::integrations::{
     AttributeRewriteAction, IntegrationAttributeContext, IntegrationAttributeRewriter,
     IntegrationEndpoint, IntegrationHeadInjector, IntegrationHtmlContext, IntegrationProxy,
     IntegrationRegistration,
 };
+use crate::proxy::{proxy_request, ProxyRequestConfig};
 use crate::settings::{IntegrationConfig, Settings};
 
 const GPT_INTEGRATION_ID: &str = "gpt";
@@ -123,6 +124,133 @@ impl GptIntegration {
         ))
     }
 
+    fn build_proxy_config<'a>(target_url: &'a str, req: &Request) -> ProxyRequestConfig<'a> {
+        let mut config = ProxyRequestConfig::new(target_url)
+            .with_streaming()
+            .without_forward_headers();
+        config.follow_redirects = false;
+        config.forward_synthetic_id = false;
+
+        Self::apply_request_header_allowlist(config, req)
+    }
+
+    fn apply_request_header_allowlist<'a>(
+        mut config: ProxyRequestConfig<'a>,
+        req: &Request,
+    ) -> ProxyRequestConfig<'a> {
+        for header_name in [
+            &HEADER_ACCEPT,
+            &HEADER_ACCEPT_LANGUAGE,
+            &HEADER_ACCEPT_ENCODING,
+        ] {
+            if let Some(value) = req.get_header(header_name).cloned() {
+                config = config.with_header(header_name.clone(), value);
+            }
+        }
+
+        config.with_header(
+            header::USER_AGENT,
+            fastly::http::HeaderValue::from_static("TrustedServer/1.0"),
+        )
+    }
+
+    fn ensure_successful_gpt_asset_response(
+        response: &Response,
+        context: &str,
+    ) -> Result<(), Report<TrustedServerError>> {
+        if response.get_status().is_success() {
+            return Ok(());
+        }
+
+        let status = response.get_status();
+        log::error!(
+            "GPT proxy upstream returned status {} for {}",
+            status,
+            context
+        );
+        Err(Report::new(Self::error(format!(
+            "{context}: upstream returned {status}"
+        ))))
+    }
+
+    fn finalize_gpt_asset_response(&self, mut response: Response) -> Response {
+        let status = response.get_status();
+        let content_type = response.get_header(header::CONTENT_TYPE).cloned();
+        let content_encoding = response.get_header(header::CONTENT_ENCODING).cloned();
+        let etag = response.get_header(header::ETAG).cloned();
+        let last_modified = response.get_header(header::LAST_MODIFIED).cloned();
+        let upstream_vary = response
+            .get_header(header::VARY)
+            .and_then(|value| value.to_str().ok())
+            .map(str::to_owned);
+        let body = response.take_body();
+
+        let mut finalized = Response::from_status(status).with_body(body);
+        finalized.set_header("X-GPT-Proxy", "true");
+
+        if let Some(content_type) = content_type {
+            finalized.set_header(header::CONTENT_TYPE, content_type);
+        }
+
+        if let Some(etag) = etag {
+            finalized.set_header(header::ETAG, etag);
+        }
+
+        if let Some(last_modified) = last_modified {
+            finalized.set_header(header::LAST_MODIFIED, last_modified);
+        }
+
+        if let Some(content_encoding) = content_encoding {
+            finalized.set_header(header::CONTENT_ENCODING, content_encoding);
+            finalized.set_header(
+                header::VARY,
+                Self::vary_with_accept_encoding(upstream_vary.as_deref()),
+            );
+        }
+
+        if status.is_success() {
+            finalized.set_header(
+                header::CACHE_CONTROL,
+                format!("public, max-age={}", self.config.cache_ttl_seconds),
+            );
+        }
+
+        finalized
+    }
+
+    fn vary_with_accept_encoding(upstream_vary: Option<&str>) -> String {
+        match upstream_vary.map(str::trim) {
+            Some("*") => "*".to_string(),
+            Some(vary) if !vary.is_empty() => {
+                if vary
+                    .split(',')
+                    .any(|header_name| header_name.trim().eq_ignore_ascii_case("accept-encoding"))
+                {
+                    vary.to_string()
+                } else {
+                    format!("{vary}, Accept-Encoding")
+                }
+            }
+            _ => "Accept-Encoding".to_string(),
+        }
+    }
+
+    async fn proxy_gpt_asset(
+        &self,
+        settings: &Settings,
+        req: Request,
+        target_url: &str,
+        context: &str,
+    ) -> Result<Response, Report<TrustedServerError>> {
+        let config = Self::build_proxy_config(target_url, &req);
+        let response = proxy_request(settings, req, config)
+            .await
+            .change_context(Self::error(context))?;
+
+        Self::ensure_successful_gpt_asset_response(&response, context)?;
+        Ok(self.finalize_gpt_asset_response(response))
+    }
+
     /// Check if a URL points at Google's GPT bootstrap script (`gpt.js`).
     ///
     /// Only matches the canonical host:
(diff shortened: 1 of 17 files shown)


scripts/integration-tests-browser.sh
```diff
@@ -0,0 +1,63 @@
...
+npx playwright install chromium
+
+# --- Export env vars for global-setup.ts ---
+export WASM_BINARY_PATH="$REPO_ROOT/target/wasm32-wasip1/release/trusted-server-fastly.wasm"
+export INTEGRATION_ORIGIN_PORT="$ORIGIN_PORT"
+export VICEROY_CONFIG_PATH="$REPO_ROOT/crates/integration-tests/fixtures/configs/viceroy-template.toml"
+
+# Cleanup trap: stop any leftover containers on failure
+cleanup() {
+    docker ps -q \
```
aram356: 🔧 Multiple `--filter ancestor=` flags in one `docker ps` call are AND-joined by Docker — a container can't descend from both images simultaneously, so this always returns zero results.

On failure paths the trap silently does nothing and can leave port-8888 containers behind.

Fix: split into two separate calls:
```bash
cleanup() {
    docker ps -q --filter ancestor=test-nextjs:latest 2>/dev/null | xargs -r docker stop 2>/dev/null || true
    docker ps -q --filter ancestor=test-wordpress:latest 2>/dev/null | xargs -r docker stop 2>/dev/null || true
}
```

crates/fastly/src/main.rs
```diff
@@ -80,6 +80,12 @@ async fn route_request(
 
     // Match known routes and handle them
     let result = match (method, path.as_str()) {
+        // Health check endpoint for integration tests and monitoring.
+        // Namespaced to avoid shadowing publisher routes.
+        (Method::GET, "/__trusted-server/health") => {
```
aram356: ❓ This adds a hard intercept for `GET /__trusted-server/health` returning `200 ok` before the catch-all publisher proxy.

Previously this path fell through to the origin. The `/__trusted-server/` namespace makes collision unlikely, but if any publisher origin serves content at this path, behavior regresses from origin content to a forced health response.

Is the intent to keep this in production, or should it be gated behind a build-time flag so it's only active during testing?

crates/integration-tests/browser/helpers/infra.ts
```diff
@@ -0,0 +1,128 @@
...
+    throw err;
+  }
+
+  return containerId;
+}
+
+/** Stop a Docker container by ID. */
+export function stopContainer(containerId: string): void {
+  try {
+    execSync(`docker stop ${containerId}`, { timeout: 10_000 });
```
aram356: 🔧 `execSync` with string interpolation passes `containerId` through the shell. If the value is malformed (e.g. from a corrupted state file), this is a command injection vector. Same pattern at lines 47–54 for `docker run`.

Fix: use `execFileSync('docker', ['stop', containerId], { timeout: 10_000 })` which bypasses the shell entirely.

crates/integration-tests/tests/integration.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+/// Test all combinations: frameworks x runtimes (matrix testing).
+///
+/// Iterates every registered runtime and framework, running all standard
+/// and custom scenarios for each combination. Uses `--test-threads=1`
+/// because all containers share the same fixed origin port.
+#[test]
+#[ignore = "requires Docker, Viceroy, and pre-built WASM binary"]
+fn test_all_combinations() {
```
aram356: 🤔 With `--include-ignored` in the CI script, this test runs alongside `test_wordpress_fastly` and `test_nextjs_fastly` below — each framework gets tested twice (once in the matrix, once individually). For Docker-based tests this roughly doubles CI time.

Consider either removing the individual test functions or not using `--include-ignored` and calling specific tests instead.

aram356 (changes requested):
Solid integration testing infrastructure. The `RuntimeEnvironment` + `FrontendFramework` matrix design is clean and extensible, the two-layer approach (Rust HTTP + Playwright browser) gives good coverage, and the assertion library having its own unit tests is a nice touch.

### Cross-cutting notes

📝 **Separate `Cargo.lock`** — The 3400-line `crates/integration-tests/Cargo.lock` is disconnected from the workspace lock. If `error-stack` (or any shared dep) is updated in the workspace but not here, they silently diverge. Consider documenting the upgrade process or adding a CI check that validates shared dependency versions stay in sync.

📝 **Viceroy config placeholders** — `viceroy-template.toml` uses `key = "placeholder"` / `data = "placeholder"` for all KV stores. This means integration tests don't exercise KV store functionality, which is fine but worth a comment in the config noting this intentional scope limitation.

⛏ **`.gitignore` missing trailing newline** — The last line lacks a POSIX-compliant trailing newline.

### What's done well

👍 `Drop` impl on `ViceroyHandle` ensures process cleanup even on panics — good defensive design.

👍 Assertion functions (`assertions.rs`) have thorough unit tests covering both passing and failing cases.

👍 Browser tests use hydration detection (`data-hydrated` attribute) before SPA navigation — avoids the classic "clicked before client router was ready" flake.

👍 Health check endpoint is namespaced (`/__trusted-server/health`) to minimize publisher route conflicts.

👍 Shell scripts detect native target dynamically via `rustc -vV` — no hardcoded platform assumptions.

👍 Smart use of PHP built-in server for WordPress fixture — avoids MySQL dependency while still testing real HTML processing behavior.

scripts/integration-tests-browser.sh
prk-Jr: Addressed on the current PR head: browser cleanup now stops matching the Next.js and WordPress test containers independently instead of combining ancestor filters.

crates/integration-tests/browser/helpers/infra.ts
prk-Jr: Addressed on the current PR head: infra.ts uses execFileSync for Docker process execution instead of shell-interpolated execSync.

crates/integration-tests/fixtures/frameworks/nextjs/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+#
+# Build:
+#   docker build -t test-nextjs:latest \
+#     crates/integration-tests/fixtures/frameworks/nextjs/
+#
+# Run manually:
+#   docker run -p 3000:3000 test-nextjs:latest
+
+# --- Build stage ---
+FROM node:20-alpine AS builder
```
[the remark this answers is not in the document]
prk-Jr: Addressed on the current PR head: the Next.js fixture build now takes NODE_VERSION from .tool-versions in both CI and the browser test script.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,179 @@
...
+            crates/integration-tests/fixtures/frameworks/nextjs/
+
+      - name: Run integration tests
+        run: >-
+          cargo test
+          --manifest-path crates/integration-tests/Cargo.toml
+          --target x86_64-unknown-linux-gnu
+          -- --include-ignored --test-threads=1
+        env:
+          INTEGRATION_ORIGIN_PORT: ${{ env.ORIGIN_PORT }}
```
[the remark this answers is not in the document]
prk-Jr: Addressed on the current PR head: the duplicated workflow setup has been extracted into the shared composite action at .github/actions/setup-integration-test-env.

aram356 (changes requested):
### Summary

All 12 CI checks pass. The architecture is solid — clean separation between runtime environments, frontend frameworks, and test scenarios; proper `error-stack` usage; good `Drop`-based cleanup; thoughtful extensibility patterns.

Three blocking issues need attention before merge, plus six non-blocking suggestions.

**Blocking:** 3 · **Non-blocking:** 6

crates/integration-tests/tests/common/runtime.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// Get the fixed origin port used for Docker container port mapping.
+///
+/// This must match the port baked into the WASM binary via
+/// `TRUSTED_SERVER__PUBLISHER__ORIGIN_URL` at build time.
+pub fn origin_port() -> u16 {
+    std::env::var("INTEGRATION_ORIGIN_PORT")
+        .ok()
+        .and_then(|s| s.parse().ok())
+        .unwrap_or(8888)
```
ChristianPavilonis: 🔧 **Silent env var parse failure** (P1): If `INTEGRATION_ORIGIN_PORT` is set to an invalid value like `"abc"`, `.parse().ok()` silently falls back to `8888`, which will cause confusing test failures downstream.

**Suggestion** — fail loudly when the var IS present but unparseable:
```rust
pub fn origin_port() -> u16 {
    match std::env::var("INTEGRATION_ORIGIN_PORT") {
        Ok(s) => s.parse().expect("should parse INTEGRATION_ORIGIN_PORT as u16"),
        Err(_) => 8888,
    }
}
```

crates/integration-tests/README.md
```diff
@@ -0,0 +1,225 @@
...
+All tests share the same fixed origin port (8888). The trusted server config is
+baked into the WASM binary at compile time with this port, so only one Docker
+container can be bound to it at a time.
+
+## CI
+
+Integration tests run in a separate workflow (`.github/workflows/integration-tests.yml`)
+triggered by:
+
+- Push to `main`
```
ChristianPavilonis: 🔧 **README/CI trigger mismatch** (P1): The README says "Push to `main`" and "PR approval" triggers, but the actual workflow (`.github/workflows/integration-tests.yml`) triggers on `pull_request: [opened, synchronize, reopened]` and `workflow_dispatch` only. There is no push-to-main trigger and no approval gate.

**Suggestion**: Update to match actual triggers:
```markdown
- Pull request opened/updated/reopened
- Manual dispatch
```

aram356 (approved):
👍 Looks good

PR #459, opened 2026-03-07
Harden inline creative rendering

### LINKED ISSUE #401: Unsanitized creative HTML injected into iframe with weakened sandbox
## Summary

Creative HTML (`bid.adm`) from the upstream bidder response is injected into an iframe via `srcdoc` with no sanitization. The iframe sandbox includes both `allow-scripts` and `allow-same-origin`, which together allow script inside the iframe to remove its own sandbox attribute and gain full access to the publisher's page context, cookies, and localStorage.

## Refs

- `crates/js/lib/src/core/request.ts` lines 77-102 — `iframe.srcdoc = buildCreativeDocument(creativeHtml)`
- `crates/js/lib/src/core/render.ts` lines 104-111 — sandbox with `allow-scripts` + `allow-same-origin`
- `crates/js/lib/src/core/render.ts` lines 133-137 — `buildCreativeDocument` does raw string replace
- `crates/js/lib/src/core/auction.ts` lines 141-172 — `parseAuctionResponse` passes `adm` straight through

## Recommendation

Either (a) remove `allow-same-origin` if creatives don't need it, (b) serve creatives from a separate origin, or (c) sanitize `adm` with DOMPurify before injection.

## Context

prk-Jr:
## Summary

- Add server-side HTML sanitization for inline ad creatives using `lol_html`: strip executable elements (`script`, `iframe`, `object`, `embed`, `base`, `meta`, `form`, `link`, `style`, `noscript`), `on*` event-handler attributes, and dangerous URI schemes (`javascript:`, `vbscript:`, unsafe `data:` URIs including `data:image/svg+xml`).
- Sanitization runs server-side before URL rewriting so the client only ever receives clean markup.
- Tighten the client-side iframe sandbox: remove `allow-scripts` and `allow-same-origin`, keeping only `allow-popups`, `allow-popups-to-escape-sandbox`, and `allow-top-navigation-by-user-activation`.
- Fail closed on oversized input and parse errors (return empty string rather than raw markup).
- Validate `bid.adm` before rendering; reject and log malformed or empty creatives with structured metadata, never logging raw creative HTML.
- Add regression coverage for sandbox permissions, dangerous URI/style payloads, malformed creatives, and accepted safe markup.

## Changes

| File | Change |
| ---- | ------ |
| `crates/common/src/creative.rs` | Add `sanitize_creative_html`: lol_html-based server-side sanitizer that strips dangerous elements and attributes. Fail closed on oversized input and parse errors. |
| `crates/common/src/auction/formats.rs` | Run `sanitize_creative_html` before `rewrite_creative_html` so URL rewriting only ever sees clean markup. |
| `crates/js/lib/src/core/render.ts` | Simplify client-side `sanitizeCreativeHtml` to type/emptiness validation only (server handles sanitization). Tighten iframe sandbox — remove `allow-scripts` and `allow-same-origin`. |
| `crates/js/lib/src/core/request.ts` | Guard against missing `bid.adm`. Clear slot only after sanitization succeeds so rejected creatives never blank existing content. Add structured rejection logging. |
| `crates/js/lib/test/core/render.test.ts` | Cover sandbox tokens, accepted safe markup, rejected type/emptiness payloads, and `buildCreativeDocument` template injection. |
| `crates/js/lib/test/core/request.test.ts` | Cover safe request-path rendering plus fail-closed behavior for dangerous, malformed, and empty creatives without logging raw HTML. |

## Closes

Closes #401

## Test plan

- [x] Other: `cd crates/js/lib && npm run build`

## Checklist

- [x] New code has tests


crates/common/src/creative.rs
```diff
@@ -303,6 +303,170 @@ pub fn rewrite_css_body(settings: &Settings, css: &str) -> String {
...
+/// On parse errors the original markup is also returned unchanged with a warning.
+#[must_use]
+pub fn sanitize_creative_html(markup: &str) -> String {
+    if markup.len() > MAX_CREATIVE_SIZE {
+        log::warn!(
+            "sanitize_creative_html: creative too large ({} bytes > {} byte limit); returning unchanged",
+            markup.len(),
+            MAX_CREATIVE_SIZE
+        );
+        return markup.to_owned();
```
aram356: 🔧 **wrench** — Fail-open: oversized input bypasses sanitization

When `markup.len() > MAX_CREATIVE_SIZE`, the raw unsanitized markup is returned unchanged. An attacker who controls creative content could exceed the 1 MiB limit, causing `<script>`, `on*` handlers, etc. to pass through to the client.

**Fix** — fail closed by returning an empty string:
```rust
if markup.len() > MAX_CREATIVE_SIZE {
    log::warn!("sanitize_creative_html: creative too large; rejecting");
    return String::new();
}
```

The test `sanitize_returns_unchanged_when_over_size_limit` would need updating to assert an empty string.

crates/common/src/creative.rs
```diff
@@ -303,6 +303,170 @@ pub fn rewrite_css_body(settings: &Settings, css: &str) -> String {
...
+            ..HtmlSettings::default()
+        },
+        |c: &[u8]| out.extend_from_slice(c),
+    );
+
+    // Short-circuit: do not call end() after a failed write(), as lol_html's
+    // rewriter is in an error state and may produce garbage output.
+    if rewriter.write(markup.as_bytes()).is_err() || rewriter.end().is_err() {
+        log::warn!("sanitize_creative_html: html parse error; returning markup unchanged");
+        return markup.to_owned();
```
aram356: 🔧 **wrench** — Fail-open: parse errors bypass sanitization

If `rewriter.write()` or `rewriter.end()` fails, the raw unsanitized markup is returned. A crafted input that triggers a lol_html parse error would skip all sanitization, and the browser's more forgiving HTML parser would execute the payload.

**Fix** — fail closed:
```rust
if rewriter.write(markup.as_bytes()).is_err() || rewriter.end().is_err() {
    log::warn!("sanitize_creative_html: html parse error; rejecting markup");
    return String::new();
}
```

crates/js/lib/src/core/render.ts
```diff
@@ -6,11 +6,74 @@ import { getUnit, getAllUnits, firstSize } from './registry';
 import NORMALIZE_CSS from './styles/normalize.css?inline';
 import IFRAME_TEMPLATE from './templates/iframe.html?raw';
 
+const CREATIVE_SANDBOX_TOKENS = [
```
ChristianPavilonis: 🤔 **P2 — `allow-forms` silently removed from sandbox**

The previous sandbox included `allow-forms`. The PR description mentions removing `allow-scripts` and `allow-same-origin`, but `allow-forms` was also removed without being called out.

Removing `allow-forms` could break ad creatives that use forms for lead generation (e.g., email signup forms within ad units). The server-side sanitizer already strips `<form>` elements, so this is consistent — but it's a notable behavioral change that should be documented.

Either:
1. Add `allow-forms` back if form submission from creatives is a valid use case for your publishers, or
2. Explicitly document the `allow-forms` removal in the PR description and confirm it's intentional (since the server strips `<form>` anyway, this is likely fine — just worth calling out)

PR #461, opened 2026-03-08
Fail fast on invalid config regexes and enabled config

### LINKED ISSUE #403: Panicking .expect() on regex compilation from user configuration
## Summary

Configuration-derived regex patterns use `.expect()` which will panic at runtime if the pattern is invalid. If `handler.path` or integration config contains invalid regex metacharacters and validation is bypassed (env override, manual TOML edit), the service crashes on first matching request.

## Refs

- `crates/common/src/settings.rs` line 255 — `Regex::new(&self.path).expect(...)`
- `crates/common/src/integrations/nextjs/script_rewriter.rs` line 125 — `Regex::new(&pattern).expect(...)`
- `crates/common/src/integrations/nextjs/script_rewriter.rs` line 141 — `Regex::new(&pattern).expect(...)`
- `crates/common/src/integrations/nextjs/shared.rs` line 83 — `Regex::new(...).expect(...)`

## Recommendation

Return `Result` from these constructors; catch at startup with a descriptive error message.

## Context

prk-Jr:
## Summary

- Fail startup instead of panicking or silently disabling when config-derived regexes or enabled integration/provider configs are invalid.
- Prepare handler and Next.js rewrite artifacts before request handling so invalid config returns descriptive `TrustedServerError` responses.
- Add regression coverage for handler overrides, disabled-config bypasses, Next.js RSC/`__NEXT_DATA__` rewrites, and publisher fallback encoding behavior.

## Changes

| File | Change |
| ---- | ------ |
| `.claude/agents/pr-creator.md` | Require PRs touching config-derived regex or pattern compilation to document startup hardening and validation coverage. |
| `.claude/agents/pr-reviewer.md` | Treat panic-prone config compilation and silent invalid-enabled-config disablement as blocking review findings. |
| `crates/common/src/auction/mod.rs` | Make auction provider construction fallible and propagate startup configuration errors. |
| `crates/common/src/auth.rs` | Make basic-auth evaluation fallible so invalid handler regexes return config errors instead of panicking. |
| `crates/common/src/integrations/adserver_mock.rs` | Return startup errors for invalid enabled provider config instead of logging and skipping it. |
| `crates/common/src/integrations/aps.rs` | Return startup errors for invalid enabled provider config instead of logging and skipping it. |
| `crates/common/src/integrations/datadome.rs` | Make enabled config registration fail fast with descriptive validation errors. |
| `crates/common/src/integrations/didomi.rs` | Make enabled config registration fail fast with descriptive validation errors. |
| `crates/common/src/integrations/google_tag_manager.rs` | Make enabled config registration fail fast with descriptive validation errors. |
| `crates/common/src/integrations/gpt.rs` | Make enabled config registration fail fast with descriptive validation errors. |
| `crates/common/src/integrations/lockr.rs` | Make enabled config registration fail fast with descriptive validation errors. |
| `crates/common/src/integrations/mod.rs` | Update shared integration builder signatures and helpers for fallible startup registration. |
| `crates/common/src/integrations/nextjs/html_post_process.rs` | Pass request host and scheme through the post-processor RSC rewrite path. |
| `crates/common/src/integrations/nextjs/mod.rs` | Build Next.js rewriters during registration and add fixture-style regressions for RSC and `__NEXT_DATA__` coverage. |
| `crates/common/src/integrations/nextjs/rsc.rs` | Reuse the new RSC rewrite helpers while preserving payload sizing and chunk handling. |
| `crates/common/src/integrations/nextjs/script_rewriter.rs` | Replace config-derived `expect()` regex compilation with fallible rewriters and add hostname, port, whitespace, and metacharacter regressions. |
| `crates/common/src/integrations/nextjs/shared.rs` | Replace origin-specific regex construction with static generic patterns and safe hostname-boundary helpers. |
| `crates/common/src/integrations/permutive.rs` | Make enabled config registration fail fast with descriptive validation errors. |
| `crates/common/src/integrations/prebid.rs` | Fail startup on invalid enabled config, including empty `server_url`, and return fallible provider builders. |
| `crates/common/src/integrations/registry.rs` | Propagate integration registration errors during startup instead of silently dropping invalid enabled integrations. |
| `crates/common/src/integrations/testlight.rs` | Make enabled config registration fail fast with descriptive validation errors. |
| `crates/common/src/publisher.rs` | Restrict publisher fallback `Accept-Encoding` to codecs the rewrite pipeline supports and add regression coverage. |
| `crates/common/src/settings.rs` | Add runtime preparation for handler regexes, short-circuit raw `enabled = false` configs, and add startup hardening regressions. |
| `crates/common/src/settings_data.rs` | Run runtime preparation when loading baked settings data so invalid handler regexes fail during startup. |
| `crates/fastly/src/main.rs` | Surface orchestrator and auth configuration errors through the existing Fastly error-response path. |

## Hardening note

Invalid `handlers[].path` regexes now fail during startup preparation and still return descriptive configuration errors if request-time code encounters unprepared settings. Enabled integrations and auction providers now fail registry/orchestrator startup instead of logging and disabling themselves, while raw `enabled = false` configs still short-circuit before validation. Regression coverage includes invalid handler TOML and env overrides, disabled-invalid config bypasses, enabled-invalid integration/provider startup failures, empty `prebid.server_url`, Next.js `__NEXT_DATA__` and RSC hostname/port/metacharacter rewrites, and the publisher fallback encoding path.

## Closes

Closes #403

## Test plan

- [x] Other: `npx vitest run` was skipped per explicit user instruction because of the unrelated repo-wide `ERR_REQUIRE_ESM` failure in `html-encoding-sniffer`.

## Checklist

- [x] No `unwrap()` in production code — use `expect(\"should ...\")`
- [x] New code has tests

### DIFF
--- a/crates/trusted-server-core/src/integrations/nextjs/html_post_process.rs
+++ b/crates/trusted-server-core/src/integrations/nextjs/html_post_process.rs
@@ -65,9 +65,13 @@ impl IntegrationHtmlPostProcessor for NextJsHtmlPostProcessor {
             })
             .unwrap_or_default();
 
+        // Single rewriter instance shared across both code paths so the compiled
+        // regex is cached and reused regardless of which branch executes.
+        let rsc_rewriter = RscUrlRewriter::new();
+
         if !payloads.is_empty() {
             // Placeholder approach: substitute placeholders with rewritten payloads
-            return self.substitute_placeholders(html, ctx, payloads);
+            return self.substitute_placeholders(html, ctx, payloads, &rsc_rewriter);
         }
 
         // Fallback: re-parse HTML to find RSC scripts that weren't captured during streaming
@@ -78,6 +82,7 @@ impl IntegrationHtmlPostProcessor for NextJsHtmlPostProcessor {
             ctx.request_host,
             ctx.request_scheme,
             self.config.max_combined_payload_bytes,
+            &rsc_rewriter,
         )
     }
 }
@@ -89,12 +94,12 @@ impl NextJsHtmlPostProcessor {
         html: &mut String,
         ctx: &IntegrationHtmlContext<'_>,
         payloads: Vec<String>,
+        rsc_rewriter: &RscUrlRewriter,
     ) -> bool {
         let payload_refs: Vec<&str> = payloads.iter().map(String::as_str).collect();
-        let rsc_rewriter = RscUrlRewriter::new();
         let mut rewritten_payloads = rewrite_rsc_scripts_combined_with_limit(
             payload_refs.as_slice(),
-            &rsc_rewriter,
+            rsc_rewriter,
             ctx.origin_host,
             ctx.request_host,
             ctx.request_scheme,
@@ -368,12 +373,14 @@ pub fn post_process_rsc_html_in_place(
     request_host: &str,
     request_scheme: &str,
 ) -> bool {
+    let rsc_rewriter = RscUrlRewriter::new();
     post_process_rsc_html_in_place_with_limit(
         html,
         origin_host,
         request_host,
         request_scheme,
         super::rsc::DEFAULT_MAX_COMBINED_PAYLOAD_BYTES,
+        &rsc_rewriter,
     )
 }
 
@@ -383,6 +390,7 @@ fn post_process_rsc_html_in_place_with_limit(
     request_host: &str,
     request_scheme: &str,
     max_combined_payload_bytes: usize,
+    rsc_rewriter: &RscUrlRewriter,
 ) -> bool {
     let mut scripts = find_rsc_push_scripts(html.as_str());
     if scripts.is_empty() {
@@ -455,11 +463,9 @@ fn post_process_rsc_html_in_place_with_limit(
             );
         }
 
-        let rewriter = RscUrlRewriter::new();
-
         let rewritten_payloads = rewrite_rsc_scripts_combined_with_limit(
             payloads.as_slice(),
-            &rewriter,
+            rsc_rewriter,
             origin_host,
             request_host,
             request_scheme,
--- a/crates/trusted-server-core/src/integrations/nextjs/shared.rs
+++ b/crates/trusted-server-core/src/integrations/nextjs/shared.rs
@@ -151,36 +151,46 @@ impl RscUrlRewriter {
             return Cow::Borrowed(input);
         }
 
-        // Retrieve a compiled origin-specific pattern from cache, or compile and store one.
+        // Check cache first. Use the regex inside the borrow scope to avoid cloning it.
         // The same origin_host is used for every payload in a request, so this avoids
         // recompiling the pattern on each call. regex::escape() guarantees validity, but
         // treat compilation failure as non-fatal rather than panic.
         let cached = self.cached_pattern.borrow();
-        let origin_pattern = if let Some((ref host, ref regex)) = *cached {
+        if let Some((ref host, ref regex)) = *cached {
             if host == origin_host {
-                Some(regex.clone())
-            } else {
-                None
+                return Self::apply_rewrite(
+                    input,
+                    regex,
+                    origin_host,
+                    request_host,
+                    request_scheme,
+                );
             }
-        } else {
-            None
-        };
+        }
         drop(cached);
-        let origin_pattern = match origin_pattern {
-            Some(p) => p,
-            None => match build_origin_url_pattern(origin_host) {
-                Ok(pattern) => {
-                    let regex = pattern.clone();
-                    *self.cached_pattern.borrow_mut() = Some((origin_host.to_string(), pattern));
-                    regex
-                }
-                Err(e) => {
-                    log::error!("Failed to compile origin URL pattern: {e}");
-                    return Cow::Borrowed(input);
-                }
-            },
-        };
 
+        // Cache miss: compile and store the pattern, then apply.
+        match build_origin_url_pattern(origin_host) {
+            Ok(pattern) => {
+                let result =
+                    Self::apply_rewrite(input, &pattern, origin_host, request_host, request_scheme);
+                *self.cached_pattern.borrow_mut() = Some((origin_host.to_string(), pattern));
+                result
+            }
+            Err(e) => {
+                log::error!("Failed to compile origin URL pattern: {e}");
+                Cow::Borrowed(input)
+            }
+        }
+    }
+
+    fn apply_rewrite<'a>(
+        input: &'a str,
+        origin_pattern: &Regex,
+        origin_host: &str,
+        request_host: &str,
+        request_scheme: &str,
+    ) -> Cow<'a, str> {
         // Phase 1: Regex-based URL pattern rewriting (handles escaped slashes, schemes, etc.)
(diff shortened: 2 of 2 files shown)


crates/common/src/publisher.rs
```diff
@@ -15,6 +15,13 @@ use crate::streaming_processor::{Compression, PipelineConfig, StreamProcessor, S
 use crate::streaming_replacer::create_url_replacer;
 use crate::synthetic::get_or_generate_synthetic_id;
 
+/// Encodings supported by the publisher response rewrite pipeline.
+const SUPPORTED_ENCODINGS: &str = "gzip, deflate, br";
```
aram356: 🤔 **thinking** — `restrict_accept_encoding` unconditionally overwrites client preferences

The fix correctly prevents the pipeline from receiving encodings it can't decode (e.g., zstd). But a client that only accepts `identity` or a strict subset will now receive content encoded with gzip/br it didn't ask for.

Consider computing the intersection of client capabilities and pipeline capabilities, falling back to `identity` if the intersection is empty.

crates/common/src/integrations/nextjs/script_rewriter.rs
```diff
@@ -62,8 +71,14 @@ impl IntegrationScriptRewriter for NextJsNextDataRewriter {
     }
 }
 
-fn rewrite_nextjs_values_with_rewriter(content: &str, rewriter: &UrlRewriter) -> Option<String> {
-    rewriter.rewrite_embedded(content)
+fn rewrite_nextjs_values_with_rewriter(
```
aram356: ♻️ **refactor** — `rewrite_nextjs_values_with_rewriter` is a pass-through wrapper

This function just delegates to `rewriter.rewrite_embedded(...)` with no additional logic. The call site in `rewrite_structured` (line 42) could call `self.rewriter.rewrite_embedded(...)` directly and this intermediary could be removed.

crates/common/src/settings.rs
```diff
@@ -411,10 +471,14 @@ fn validate_no_trailing_slash(value: &str) -> Result<(), ValidationError> {
 }
 
 fn validate_path(value: &str) -> Result<(), ValidationError> {
```
aram356: ♻️ **refactor** — `validate_path` compiles the regex a second time

`validate_path` compiles the regex during serde validation, then `prepare_runtime()` → `compiled_regex()` compiles it again into the `OnceLock`. Since `from_toml` now calls `prepare_runtime()`, the validation-time compile is redundant work.

Consider deferring regex validation to `prepare_runtime` only and removing the `validate_path` custom validator (or having `validate_path` just check non-empty, letting `prepare_runtime` handle compilation).

crates/common/src/settings.rs
```diff
@@ -136,6 +136,13 @@ impl IntegrationSettings {
         }
     }
 
+    fn is_explicitly_disabled(raw: &JsonValue) -> bool {
```
aram356: 🤔 **thinking** — `is_explicitly_disabled` edge case with non-bool `enabled`

If someone writes `"enabled": "false"` (string instead of bool), `as_bool()` returns `None`, so the integration proceeds to deserialization and may fail validation. This is correct behavior (malformed config should fail), but worth noting that the bypass only fires for the literal JSON boolean `false`.

Also worth noting: integrations that default `enabled` to `true` with no `enabled` field in the config will correctly proceed to validation, which is the desired behavior.

crates/common/src/publisher.rs
```diff
@@ -15,6 +15,83 @@ use crate::streaming_processor::{Compression, PipelineConfig, StreamProcessor, S
...
+                continue;
+            };
+            if name.trim().eq_ignore_ascii_case("q") {
+                if let Ok(parsed_qvalue) = value.trim().parse::<f32>() {
+                    qvalue = parsed_qvalue;
+                }
+            }
+        }
+
+        matched_qvalue = Some(qvalue);
```
aram356: ♻️ **refactor** — `accept_encoding_qvalue` returns the last match, not the first

If `Accept-Encoding` contains a duplicate token (e.g., `gzip;q=1, gzip;q=0`), this returns the **last** match's qvalue. Most HTTP implementations use first-match semantics.

**Suggestion**:
```rust
matched_qvalue = Some(qvalue);
break; // first match wins per RFC convention
```

In practice duplicate tokens are rare and unlikely to cause issues, but first-match is more conventional.

crates/common/src/publisher.rs
```diff
@@ -15,6 +15,83 @@ use crate::streaming_processor::{Compression, PipelineConfig, StreamProcessor, S
...
+    else {
+        return;
+    };
+    req.set_header(
+        header::ACCEPT_ENCODING,
+        select_supported_accept_encoding(Some(current)),
+    );
+}
+
+fn select_supported_accept_encoding(client_accept_encoding: Option<&str>) -> String {
```
aram356: ♻️ **refactor** — `select_supported_accept_encoding(None)` is dead code

The only caller `restrict_accept_encoding` returns early when `Accept-Encoding` is absent, so the `None` branch (returning `"identity"`) is unreachable. You could simplify the parameter to `&str` if this function isn't intended for other callers.

crates/common/src/integrations/nextjs/shared.rs
```diff
@@ -2,19 +2,31 @@
...
     Regex::new(
         r#"(?s)(?:(?:self|window)\.__next_f\.push|\(\s*(?:self|window)\.__next_f\s*=\s*(?:self|window)\.__next_f\s*\|\|\s*\[\]\s*\)\s*\.push)\(\[\s*1\s*,\s*(['"])"#,
     )
     .expect("valid RSC push call regex")
 });
 
+/// Generic URL pattern for RSC payload rewriting.
+pub(crate) static RSC_URL_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(
+        r#"(https?)?(:)?(\\\\\\\\\\\\\\\\//|\\\\\\\\//|\\/\\/|//)(?P<host>\[[^\]]+\](?::\d+)?|[^/"'\\\s?#<>,)]+)"#,
```
aram356: 🌱 **seedling** — Host capture group `[^/"'\\\s?#<>,)]+` is very permissive

This character class matches anything not in the exclusion set, which could theoretically match non-hostname content in unusual payloads. Currently safe because `strip_origin_host_with_optional_port` filters properly, but worth keeping an eye on if RSC payload formats evolve.

crates/trusted-server-core/src/settings.rs
```diff
@@ -467,13 +527,24 @@ impl Settings {
...
+    ///
+    /// Returns [`TrustedServerError::Configuration`] if any handler has an invalid path regex.
+    pub(crate) fn uncovered_admin_endpoints(
+        &self,
+    ) -> Result<Vec<&'static str>, Report<TrustedServerError>> {
+        let mut uncovered = Vec::new();
+        for &path in Self::ADMIN_ENDPOINTS {
+            let covered = self
+                .handlers
+                .iter()
```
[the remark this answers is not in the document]
prk-Jr: Replaced with a `for` loop that breaks on first match.

ChristianPavilonis (approved):
Looks good

PR #467, opened 2026-03-10
Fix weak and inconsistent secret default validation

### LINKED ISSUE #406: Weak/inconsistent secret default validation
## Summary

Only `synthetic.secret_key` is checked against the literal `"secret-key"`. `publisher.proxy_secret` defaults to `"change-me-proxy-secret"` and `synthetic.secret_key` defaults to `"trusted-server"` — neither is caught by the single check. A deployment using defaults has predictable encryption keys.

## Refs

- `trusted-server.toml` line 10 — `proxy_secret = "change-me-proxy-secret"`
- `trusted-server.toml` line 15 — `secret_key = "trusted-server"`
- `crates/common/src/settings.rs` line 197 — `Settings::validate()` — no proxy_secret check
- `crates/common/src/settings_data.rs` line 37 — only checks `== "secret-key"`

## Recommendation

Reject all known placeholder values for both secrets. Consider minimum entropy requirements.

## Context

ChristianPavilonis:
## Summary

- **Reject all known placeholder values** for `synthetic.secret_key` (`"secret-key"`, `"secret_key"`, `"trusted-server"`) and `publisher.proxy_secret` (`"change-me-proxy-secret"`) at runtime startup so misconfigured deployments fail fast.
- **Consolidate validation** into predicate methods (`is_placeholder_secret_key`, `is_placeholder_proxy_secret`) with explicit placeholder lists, replacing the inconsistent checks that missed the actual TOML defaults.
- **Improve error reporting** by renaming `InsecureSecretKey` to `InsecureDefault { field }` so the error message identifies which secret triggered rejection.

## Changes

| File | Change |
| ---- | ------ |
| `crates/common/src/error.rs` | Renamed `InsecureSecretKey` → `InsecureDefault { field: String }` with a descriptive display message |
| `crates/common/src/settings.rs` | Added `PROXY_SECRET_PLACEHOLDERS` const + `is_placeholder_proxy_secret()` on `Publisher`; replaced `validate_secret_key()` with `SECRET_KEY_PLACEHOLDERS` const + `is_placeholder_secret_key()` on `Synthetic`; added 4 unit tests |
| `crates/common/src/settings_data.rs` | Replaced single `== "secret-key"` check with calls to both predicates; returns field-specific `InsecureDefault` errors; updated test to assert placeholder rejection |
| `crates/common/build.rs` | Added comment explaining why placeholder rejection is intentionally skipped at build time |

## Closes

Closes #406

## Test plan

## Checklist

- [x] New code has tests

### DIFF
--- a/crates/common/src/settings.rs
+++ b/crates/common/src/settings.rs
@@ -723,6 +723,18 @@ mod tests {
     }
 
     #[test]
+    fn is_placeholder_secret_key_is_case_insensitive() {
+        assert!(
+            Synthetic::is_placeholder_secret_key("SECRET-KEY"),
+            "should detect case-insensitive placeholder secret_key"
+        );
+        assert!(
+            Synthetic::is_placeholder_secret_key("Trusted-Server"),
+            "should detect mixed-case placeholder secret_key"
+        );
+    }
+
+    #[test]
     fn is_placeholder_secret_key_accepts_non_placeholder() {
         assert!(
             !Synthetic::is_placeholder_secret_key("test-secret-key"),
@@ -738,6 +750,14 @@ mod tests {
                 "should detect placeholder proxy_secret '{placeholder}'"
             );
         }
+    }
+
+    #[test]
+    fn is_placeholder_proxy_secret_is_case_insensitive() {
+        assert!(
+            Publisher::is_placeholder_proxy_secret("CHANGE-ME-PROXY-SECRET"),
+            "should detect case-insensitive placeholder proxy_secret"
+        );
     }
 
     #[test]
--- a/crates/common/src/settings_data.rs
+++ b/crates/common/src/settings_data.rs
@@ -60,18 +60,21 @@ mod tests {
     /// which would cause the substitution to silently no-op.
     fn toml_with_secrets(secret_key: &str, proxy_secret: &str) -> String {
         let original = crate_test_settings_str();
-        let result = original
-            .replace(
-                r#"secret_key = "test-secret-key""#,
-                &format!(r#"secret_key = "{secret_key}""#),
-            )
-            .replace(
-                r#"proxy_secret = "unit-test-proxy-secret""#,
-                &format!(r#"proxy_secret = "{proxy_secret}""#),
-            );
+        let after_secret_key = original.replace(
+            r#"secret_key = "test-secret-key""#,
+            &format!(r#"secret_key = "{secret_key}""#),
+        );
         assert_ne!(
-            result, original,
-            "should have replaced at least one secret value"
+            after_secret_key, original,
+            "should have replaced secret_key value"
+        );
+        let result = after_secret_key.replace(
+            r#"proxy_secret = "unit-test-proxy-secret""#,
+            &format!(r#"proxy_secret = "{proxy_secret}""#),
+        );
+        assert_ne!(
+            result, after_secret_key,
+            "should have replaced proxy_secret value"
         );
         result
     }
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -244,6 +244,7 @@ TRUSTED_SERVER__PUBLISHER__PROXY_SECRET=your-secret-here
 **Security**:
 
 - Keep confidential and secure
+- Cannot be the placeholder `"change-me-proxy-secret"` (case-insensitive) — startup will fail
 - Rotate periodically (90 days recommended)
 - Use cryptographically random values (32+ bytes)
 - Never commit to version control
@@ -359,7 +360,7 @@ fastly kv-store create --name=opid_store
 **Security**:
 
 - Must be non-empty
-- Cannot be `"secret_key"` or `"secret-key"` (reserved/invalid)
+- Cannot be a known placeholder: `"secret-key"`, `"secret_key"`, or `"trusted-server"` (case-insensitive)
 - Rotate periodically for security
 - Store securely (environment variable recommended)
 
@@ -373,7 +374,7 @@ openssl rand -hex 32
 **Validation**: Application startup fails if:
 
 - Empty string
-- Exactly `"secret_key"` or `"secret-key"` (default placeholders)
+- Exactly `"secret-key"`, `"secret_key"`, or `"trusted-server"` (known placeholders, case-insensitive)
 
 #### `template`
 
@@ -922,11 +923,12 @@ Configuration is validated at startup:
 
 - All fields non-empty
 - `origin_url` is valid URL
+- `proxy_secret` ≠ known placeholder (`"change-me-proxy-secret"` — case-insensitive)
 
 **Synthetic Validation**:
 
 - `secret_key` ≥ 1 character
-- `secret_key` ≠ `"secret-key"`
+- `secret_key` ≠ known placeholders (`"secret-key"`, `"secret_key"`, `"trusted-server"` — case-insensitive)
 - `template` non-empty
 
 **Handler Validation**:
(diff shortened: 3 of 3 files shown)


crates/common/src/settings_data.rs
```diff
@@ -34,8 +34,17 @@ pub fn get_settings() -> Result<Settings, Report<TrustedServerError>> {
             message: "Failed to validate configuration".to_string(),
         })?;
 
-    if settings.synthetic.secret_key == "secret-key" {
-        return Err(Report::new(TrustedServerError::InsecureSecretKey));
+    // Reject known placeholder values for secrets that feed into cryptographic operations.
+    if Synthetic::is_placeholder_secret_key(&settings.synthetic.secret_key) {
```
aram356: **LOW — Only the first insecure field is reported**

The two placeholder checks are sequential early returns. If both `secret_key` and `proxy_secret` are placeholders (the default state from `trusted-server.toml`), the operator only sees the `synthetic.secret_key` error and must fix-and-redeploy to discover the `proxy_secret` error.

Collect all violations and report them in a single error to reduce fix/redeploy cycles.

crates/common/src/settings.rs
```diff
@@ -27,6 +27,17 @@ pub struct Publisher {
 }
 
 impl Publisher {
+    /// Known placeholder values that must not be used in production.
+    #[allow(dead_code)]
```
aram356: **NIT — Spurious `#[allow(dead_code)]` on public items**

Both the constants and the `is_placeholder_*` methods are `pub` and called from `settings_data.rs`. They are not dead code. The `#[allow(dead_code)]` suppresses nothing and misleads readers into thinking these items might be unused.

aram356 (commented):
## Summary

Centralizes placeholder-secret rejection into `Settings::reject_placeholder_secrets()`, adds `proxy_secret` validation, and makes tests deterministic. Addresses all findings from the previous review round.

## Blocking

### 🔧 wrench

- **Removed `test_get_settings` smoke test not replaced**: The original `test_get_settings` was the only test exercising the full `get_settings()` end-to-end path (embedded bytes → UTF-8 → parse → validate → placeholder check). The new tests only exercise `Settings::from_toml()` + `reject_placeholder_secrets()` directly. A regression in the glue code would go undetected. (`settings_data.rs`)

### ❓ question

- **`toml_with_secrets` replacement is fragile and can silently no-op**: `str::replace()` returns the original string unchanged if the pattern doesn't match. If `crate_test_settings_str()` changes quoting or default values, replacements silently do nothing. The `accepts_non_placeholder_secrets` test is especially vulnerable — both old and new values are non-placeholders, so it passes even if replacement fails. (`settings_data.rs:56-66`)

## Non-blocking

### 🤔 thinking

- **Case-sensitive placeholder matching**: `"Secret-Key"` or `"SECRET-KEY"` bypasses the check while still being predictable. Consider `eq_ignore_ascii_case`. (`settings.rs:205-207`)

### 🌱 seedling

- **Minimum length for crypto secrets**: `min = 1` prevents empty but `"x"` still passes validation then feeds into `Sha256::digest` for XChaCha20-Poly1305 key derivation. Consider `min = 16` as a follow-up. (`settings.rs:27, 194`)

### ⛏ nitpick

- **Double space in doc comment**: `"...offending field.  This centralises..."` — rest of codebase uses single space. (`settings.rs:388`)

## CI Status

- fmt: PASS
- clippy: PASS  
- rust tests: PASS
- js tests: PASS
- CodeQL: FAIL (appears infrastructure-related, not caused by this PR)

crates/common/src/settings_data.rs
```diff
@@ -49,22 +48,80 @@ pub fn get_settings() -> Result<Settings, Report<TrustedServerError>> {
...
+                r#"secret_key = "test-secret-key""#,
+                &format!(r#"secret_key = "{secret_key}""#),
+            )
+            .replace(
+                r#"proxy_secret = "unit-test-proxy-secret""#,
+                &format!(r#"proxy_secret = "{proxy_secret}""#),
+            )
+    }
 
     #[test]
```
aram356: 🔧 **wrench** — The original `test_get_settings` was the only test exercising the full `get_settings()` end-to-end (embedded bytes → UTF-8 → parse → validate → placeholder check). The new tests only call `Settings::from_toml()` + `reject_placeholder_secrets()` directly.

Consider adding back a smoke test:

```rust
#[test]
fn get_settings_loads_embedded_config() {
    let settings = super::get_settings();
    assert!(settings.is_ok(), "should load settings from embedded TOML");
}
```

crates/common/src/settings_data.rs
```diff
@@ -49,22 +48,80 @@ pub fn get_settings() -> Result<Settings, Report<TrustedServerError>> {
 
 #[cfg(test)]
 mod tests {
-    use super::*;
+    use crate::error::TrustedServerError;
+    use crate::settings::Settings;
+    use crate::test_support::tests::crate_test_settings_str;
+
+    /// Builds a TOML string with the given secret values swapped in.
+    fn toml_with_secrets(secret_key: &str, proxy_secret: &str) -> String {
```
aram356: ❓ **question** — `str::replace()` returns the original string unchanged if the pattern doesn't match. If `crate_test_settings_str()` ever changes quoting or default values, these replacements silently no-op.

The `accepts_non_placeholder_secrets` test is especially vulnerable: it replaces `"test-secret-key"` with `"production-secret-key"` — both non-placeholders — so even if replacement fails, the test passes.

Consider asserting the replacement actually took effect:

```rust
fn toml_with_secrets(secret_key: &str, proxy_secret: &str) -> String {
    let original = crate_test_settings_str();
    let result = original
        .replace(/* ... */)
        .replace(/* ... */);
    assert_ne!(result, original, "should have replaced at least one secret value");
    result
}
```

crates/common/src/settings.rs
```diff
@@ -180,23 +190,20 @@ impl DerefMut for IntegrationSettings {
...
-            _ => Ok(()),
-        }
+    /// Known placeholder values that must not be used in production.
+    pub const SECRET_KEY_PLACEHOLDERS: &[&str] = &["secret-key", "secret_key", "trusted-server"];
+
+    /// Returns `true` if `secret_key` matches a known placeholder value.
+    #[must_use]
+    pub fn is_placeholder_secret_key(secret_key: &str) -> bool {
+        Self::SECRET_KEY_PLACEHOLDERS.contains(&secret_key)
     }
```
aram356: 🤔 **thinking** — Placeholder matching is case-sensitive. Someone setting `secret_key = "Secret-Key"` bypasses the check while still using a predictable value. Consider case-insensitive matching:

```rust
pub fn is_placeholder_secret_key(secret_key: &str) -> bool {
    Self::SECRET_KEY_PLACEHOLDERS
        .iter()
        .any(|p| p.eq_ignore_ascii_case(secret_key))
}
```

crates/common/src/settings.rs
```diff
@@ -377,6 +384,33 @@ impl Settings {
         Ok(settings)
     }
 
+    /// Checks all secret fields for known placeholder values and returns an
+    /// error listing every offending field.  This centralises the placeholder
```
aram356: ⛏ **nitpick** — Double space after period: `"field.  This"` → `"field. This"`

crates/common/src/settings.rs
ChristianPavilonis: Done — both `is_placeholder_secret_key` and `is_placeholder_proxy_secret` now use `eq_ignore_ascii_case`.

PR #470, opened 2026-03-11
Add DD server-side validation and multi-backend routing

### LINKED ISSUE #317: Add server-side bot protection via DataDome Protection API
## Summary

Trusted Server already has a first-party proxy integration for DataDome (client-side SDK + signal collection proxied through the publisher's domain). The official [DataDome Fastly Compute module](https://docs.datadome.co/docs/fastly-compute) adds a second layer: **server-side request validation** that calls DataDome's Protection API on every incoming request to block/challenge bots before they reach the origin.

This proposal is to implement that server-side validation layer in Rust, making it **edge-provider agnostic** — not tied to Fastly's JS SDK or any other platform-specific runtime.

### Two Layers of DataDome Integration

| Layer | Status | What It Does |
|-------|--------|--------------|
| **1. First-party proxy** | Built (currently disabled) | Proxies `tags.js` + signal collection through publisher domain. Rewrites HTML/DOM references to use first-party paths. |
| **2. Server-side validation** | **Not built** | On every request, calls DataDome's Protection API with request metadata. Blocks/challenges bots before they reach the origin. |

Layer 1 makes client-side signals more reliable (bypasses ad blockers that block `datadome.co`). Layer 2 blocks bots even without client-side signals. Together they provide full coverage.

### What the Official Fastly Compute Module Does

The `@datadome/module-fastly-compute` npm package:
1. Intercepts every incoming request before it reaches the origin
2. Sends request metadata to `POST https://api.datadome.co/validate-request/` (the Protection API)
3. Blocks or challenges bots (returns 403, CAPTCHA pages, etc.)
4. Only forwards allowed requests to the origin, enriched with DataDome metadata headers
5. Fails open on timeout (allows the request through)

### New `IntegrationRequestFilter` Trait

The integration registry currently has no pre-request filtering hook. A new trait would be added:

```rust
#[async_trait(?Send)]
pub trait IntegrationRequestFilter: Send + Sync {
    fn integration_id(&self) -> &'static str;
    
    /// Return Some(Response) to block/challenge, None to allow.
    async fn filter_request(
        &self,
        settings: &Settings,
        req: &Request,
    ) -> Option<Result<Response, Report<TrustedServerError>>>;
}
```

This is a natural extension of the existing registry pattern (`IntegrationProxy`, `IntegrationAttributeRewriter`, `IntegrationScriptRewriter`, etc.) — it's the one missing lifecycle hook.

### Hook Point

Insert after `enforce_basic_auth()` in `route_request()`, before any route matching:

```rust
for filter in integration_registry.request_filters() {
    if let Some(response) = filter.filter_request(settings, &req).await {
        return response;
    }
}
```

### Request Flow

```
Client Request
  → enforce_basic_auth()
  → IntegrationRequestFilter (NEW — DataDome Protection API call)
      → Bot? → Return block/challenge response
      → Allowed? → Continue
      → Timeout? → Fail open, continue
  → Route matching (integration proxies, publisher origin, etc.)
  → Response
```

### DataDome Implementation

The DataDome `IntegrationRequestFilter` would:
1. Check if the URL matches the exclusion pattern (skip static assets like `.css`, `.js`, `.png`, etc.)
2. Extract request metadata (IP via `Fastly-Client-IP` / `X-Forwarded-For`, User-Agent, Referer, URL, method, cookies, headers)
3. `POST` to DataDome's Protection API with the server-side key
4. If DataDome says block → return the block/challenge response directly
5. If DataDome says allow → return `None`, continue normal processing
6. On timeout → fail open (allow the request)

### Configuration

```toml
[integrations.datadome]
enabled = true
# Existing first-party proxy settings (unchanged)
sdk_origin = "https://js.datadome.co"
api_origin = "https://api-js.datadome.co"
cache_ttl_seconds = 3600
rewrite_sdk = true

# New server-side validation settings
server_side_key = ""
client_side_key = ""
protection_api = "https://api.datadome.co"
enable_protection = false          # Independent toggle for Layer 2
url_pattern_exclusion = "\\.(css|js|ico|jpg|png|gif|svg|woff2?)$"
timeout_ms = 1500
```

## Why This Makes Sense

1. **Edge-agnostic** — The Protection API is plain HTTP. Rust/WASM can call it from Fastly, Cloudflare, or any edge platform. Official DataDome modules are platform-specific. This would be the first Rust implementation.
2. **Complements existing integration** — Layer 1 (proxy) + Layer 2 (validation) together provide full bot protection coverage.
3. **Ad protection** — DataDome has an "Ad Protect" product for invalid traffic (IVT). Server-side validation could filter bots before auction/bidding runs, saving bid costs.
4. **Clean architecture** — `IntegrationRequestFilter` is a natural addition to the registry trait system and could be used by other integrations in the future.

## Tradeoffs to Consider

| Concern | Detail |
|---------|--------|
| **Latency** | Extra roundtrip to DataDome API per request (~1-5ms from edge POPs). URL exclusion pattern mitigates this for static assets. |
| **Cost** | DataDome bills per validated request. Exclusion patterns for static/internal routes are important. |
| **Fail-open** | On timeout/error, requests are allowed through (standard across all DataDome modules). |

## Files Involved

- `crates/common/src/integrations/datadome.rs` — Extend with `IntegrationRequestFilter` impl
- `crates/common/src/integrations/registry.rs` — Add `IntegrationRequestFilter` trait + registry support
- `crates/fastly/src/main.rs` — Add filter hook in `route_request()`
- `trusted-server.toml` — Add new config fields

## References

- [DataDome Fastly Compute docs](https://docs.datadome.co/docs/fastly-compute)
- [DataDome Protection API](https://docs.datadome.co/reference/validate-request)
- [DataDome regional endpoints](https://docs.datadome.co/docs/api-server)

### LINKED ISSUE #389: Convert Arena VCL into Rust code in TS to not require extra CDN VCL hop and increase reponse times
### Description

We need to solve for Arena sites being able to pull directly from nextjs origin in front of WP without the need to cycle every request through a separate VCL layer. There are specifics we need to address such as: 

1. Datadome server-side integration (Compute + Rust is more efficient) 
2. Legacy redirects
3. Multi-backend strategies 
4. Domain normalizations
5. Hashmap based domain lookups
6. Staging strategy 

### Done when

- Requests to autoblog are working as normal 
- No errors for site content and widgets 
- Datadome validation is proxied and working as normal for unwanted and wanted bots 
- Site functionality is 100% of today's

### Affected area

HTML processing / JS injection

### Additional context

_No response_

jevansnyc:


crates/common/src/publisher.rs
```diff
@@ -204,20 +206,96 @@ pub fn handle_publisher_request(
...
     log::debug!(
         "Proxying to dynamic backend: {} (from {})",
         backend_name,
-        settings.publisher.origin_url
+        origin_url
     );
+
+    // DataDome server-side validation (if enabled)
+    // This validates requests at the edge before they reach the origin
+    match crate::integrations::datadome::validate_request_server_side(settings, &req) {
```
aram356: ❓ This runs on every proxied request — CSS, JS, images, fonts all trigger a DataDome HTTP call (200ms timeout). Most bot protection validates only document/navigation requests. Is this intentional? If so, worth a comment explaining why.

crates/common/src/publisher.rs
```diff
@@ -204,20 +206,96 @@ pub fn handle_publisher_request(
...
+    let router = ROUTER.get_or_init(|| {
+        if settings.backends.is_empty() {
+            return None;
+        }
+        BackendRouter::new(
+            &settings.backends,
+            settings.publisher.origin_url.clone(),
+            settings.proxy.certificate_check,
+        )
+        .ok()
```
aram356: 🤔 If `BackendRouter::new` fails (e.g. invalid regex in path patterns), the error is silently discarded and all requests fall through to default origin permanently (OnceLock). Should at least log the error.

```rust
.map_err(|e| {
    log::error!("Failed to build backend router: {:?}", e);
    e
})
.ok()
```

crates/common/src/backend_router.rs
```diff
@@ -0,0 +1,507 @@
...
+            path_prefix: pattern.path_prefix.clone(),
+            path_regex: pattern.path_regex.as_ref().map(|_| OnceLock::new()),
+            path_regex_pattern: pattern.path_regex.clone(),
+        }
+    }
+
+    fn compiled_regex(&self) -> Option<&Regex> {
+        if let (Some(regex_cell), Some(pattern)) = (&self.path_regex, &self.path_regex_pattern) {
+            Some(regex_cell.get_or_init(|| {
+                Regex::new(pattern).expect("path_regex pattern should be valid")
```
aram356: 🤔 Invalid regex in config will `expect()` panic on the first matching request, not at startup. `BackendRouter::new()` doc says it returns errors for invalid patterns but doesn't actually validate them. Consider compiling the regex eagerly in the constructor and returning the error there.

crates/common/src/backend_router.rs
```diff
@@ -0,0 +1,507 @@
...
+        default_origin: String,
+        default_certificate_check: bool,
+    ) -> Result<Self, Report<TrustedServerError>> {
+        let mut domain_index = HashMap::new();
+        let mut routes = Vec::with_capacity(backends.len());
+
+        for (idx, backend) in backends.iter().enumerate() {
+            for domain in &backend.domains {
+                let normalized = normalize_domain(domain);
+                domain_index.insert(normalized, idx);
```
aram356: 🤔 If two backends claim the same domain (or domains that normalize to the same value), the second silently overwrites the first. Could cause hard-to-debug routing. Consider detecting duplicates and logging a warning.

crates/common/src/settings.rs
```diff
@@ -294,6 +294,76 @@ impl Default for Proxy {
...
+    #[serde(default)]
+    #[validate(nested)]
+    pub path_patterns: Vec<PathPattern>,
+    /// Enable TLS certificate verification for this backend.
+    #[serde(default = "default_certificate_check")]
+    pub certificate_check: bool,
+    /// Unique identifier for logging/debugging (optional).
+    #[serde(default)]
+    pub id: Option<String>,
+    /// Backend-specific timeouts (unused - for future use).
```
aram356: ♻️ ~30 lines of code for "unused - for future use". Consider removing until actually wired into `BackendConfig`.

crates/common/src/backend_router.rs
```diff
@@ -0,0 +1,507 @@
...
+/// ```
+/// use trusted_server_common::backend_router::normalize_domain;
+///
+/// assert_eq!(normalize_domain("WWW.EXAMPLE.COM"), "example.com");
+/// assert_eq!(normalize_domain("www.example.com"), "example.com");
+/// assert_eq!(normalize_domain("example.com"), "example.com");
+/// assert_eq!(normalize_domain("sub.example.com"), "sub.example.com");
+/// ```
+#[must_use]
+pub fn normalize_domain(domain: &str) -> String {
```
aram356: ⛏ Allocates twice on every request (`to_lowercase()` then `to_string()` after trimming). Domains stored in the index are already normalized at construction time, but the incoming host is re-normalized on every call. Could use `Cow<str>` to avoid allocation when no transformation is needed.

crates/common/src/backend_router.rs
```diff
@@ -0,0 +1,507 @@
...
+    domain_index: HashMap<String, usize>,
+    default_origin: String,
+    default_certificate_check: bool,
+}
+
+#[derive(Debug, Clone)]
+pub struct BackendRoute {
+    pub origin_url: String,
+    pub certificate_check: bool,
+    #[allow(dead_code)]
```
aram356: ⛏ If the field is unused, consider removing it rather than suppressing the warning.

aram356 (changes requested):
## Summary

Adds DataDome server-side bot validation and multi-backend routing. The backend router design is solid with good test coverage. However, the build.rs changes break environment variable overrides at build time, and several runtime issues need attention.

## Findings

### 🔧 Must fix

- **build.rs uses `from_toml` instead of `from_toml_and_env`**: env var overrides silently ignored at build time (build.rs:54)
- **`rerun_if_changed` misses integration env vars**: `Settings::default()` has empty IntegrationSettings HashMap (build.rs:32)
- **Sampling hash has poor distribution**: byte-sum hash produces identical values for IPs with same bytes in different order (datadome.rs:506)
- **`validation_timeout_ms` configured but never applied**: misleading config field (datadome.rs:148)

### ❓ Questions

- **DataDome validation on every request**: static assets (CSS/JS/images) all trigger a 200ms HTTP call — intentional? (publisher.rs:276)
- **`backends.toml` in open-source repo**: ~175 production domains routing to a specific origin — should this be private?

### 🤔 Worth addressing

- **`BackendRouter::new` errors silently swallowed via `.ok()`** (publisher.rs:237)
- **Regex patterns panic at request time, not startup** (backend_router.rs:66)
- **Domain index silently overwrites on collision** (backend_router.rs:120)
- **No tests for `validate_request` logic** (~80 lines, zero coverage)

### 🌱 Future considerations

- `static OnceLock` patterns prevent test isolation (publisher.rs:227, datadome.rs:751)
- `extract_host` (datadome.rs:301) uses fragile `trim_start_matches` parsing when `url` crate is already a dependency. Also, `split('/').next()` on a non-empty string never returns `None`, so `unwrap_or("api-js.datadome.co")` is unreachable with a misleading hardcoded domain.

## CI Status

- Analyze (JS/TS): PASS
- Rust build/test/clippy/fmt: Not run

aram356 (changes requested):
## Summary

Adds server-side DataDome bot validation at the edge and a multi-backend routing system that selects origins by domain/path. Both features default to disabled with safe fail-open behavior.

## Findings

### 🔧 Blockers (P0)

- **build.rs no longer applies env var overrides**: `from_toml_and_env` was changed to `from_toml`, so env vars like `TRUSTED_SERVER__INTEGRATIONS__DATADOME__SERVER_SIDE_KEY` are never merged. The `rerun_if_changed` env var enumeration is now dead code. (crates/common/build.rs:54)

### 🔧 High (P1)

- **Sampling hash has poor distribution**: Byte-sum hash for IPv4 sampling produces collisions for IPs with same octets in different order. Use `DefaultHasher` or similar. (crates/common/src/integrations/datadome.rs:506)
- **Validation runs on every request**: Static assets (images, CSS, JS, fonts) all trigger a blocking HTTP call to DataDome. Should filter to document/navigational requests only. (crates/common/src/publisher.rs:276)
- **Duplicate domain silently overwrites**: If two backends declare the same domain, `HashMap::insert` keeps the last one with no warning. (crates/common/src/backend_router.rs:120)

### 🤔 Medium (P2)

- **`validation_timeout_ms` has no effect**: Fastly Compute doesn't support per-request timeouts, but the field is validated and documented as if it works. Misleading for operators. (crates/common/src/integrations/datadome.rs:148)
- **Hardcoded fallback in `extract_host`** (crates/common/src/integrations/datadome.rs:306): `.unwrap_or("api-js.datadome.co")` — if URL parsing fails, this silently uses a hardcoded domain instead of the configured `api_origin`. Should propagate the error or derive the fallback from config.
- **`BackendTimeouts` unused**: Declared "for future use" — adds config surface with no current benefit. (crates/common/src/settings.rs:364)
- **Inconsistent cookie trimming**: `trim_start()` in publisher.rs vs `trim()` in datadome.rs for the same pattern. (crates/common/src/publisher.rs:216)
- **No test coverage for `validate_request`**: Sampling, cookie extraction, request building, response handling, and fail-open/closed logic are all untested. (crates/common/src/integrations/datadome.rs:751)

### ⛏ Low (P3)

- **175 hardcoded domains in backends.toml**: Customer-specific data baked into the binary. Consider whether this should come from a config store. (crates/fastly/backends.toml)

## CI Status

- Analyze (JS/TS): PASS
- fmt: Not run
- clippy: Not run
- rust tests: Not run
- js tests: Not run

crates/trusted-server-core/src/backend_router.rs
```diff
@@ -0,0 +1,507 @@
...
+        (&self.default_origin, self.default_certificate_check)
+    }
+}
+
+/// Normalizes a domain by removing "www." prefix and converting to lowercase.
+///
+/// # Examples
+///
+/// ```
+/// use trusted_server_common::backend_router::normalize_domain;
```
aram356: 🔧 **wrench** — Doctest references stale crate name `trusted_server_common`. The crate was renamed to `trusted_server_core` in #616, and this is one of the two failures blocking `cargo test` in CI.

**Fix:**
```rust
use trusted_server_core::backend_router::normalize_domain;
```

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -475,14 +694,67 @@ fn build(settings: &Settings) -> Option<Arc<DataDomeIntegration>> {
...
+///
+/// # Errors
+///
+/// Only returns an error if validation fails AND `fail_open=false`.
+/// With default `fail_open=true`, errors result in `Ok(true)`.
+///
+/// # Examples
+///
+/// ```no_run
+/// use trusted_server_common::integrations::datadome;
```
aram356: 🔧 **wrench** — Doctest references stale crate name `trusted_server_common` (both `use` lines). This is the second doctest failure blocking CI.

**Fix:**
```rust
use trusted_server_core::integrations::datadome;
use trusted_server_core::settings::Settings;
```

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -381,6 +439,167 @@ impl DataDomeIntegration {
...
+                    // Bot detected - block request
+                    log::warn!("[datadome] Request blocked by DataDome (403)");
+                    Ok(false)
+                } else {
+                    // Unexpected status - apply fail-open policy
+                    log::warn!("[datadome] Unexpected validation response: {} (fail_open={})", status, self.config.fail_open);
+                    Ok(self.config.fail_open)
+                }
+            }
+            Err(e) => {
```
aram356: 🔧 **wrench** — `validation_timeout_ms` is documented and `#[validate(range(min=50,max=1000))]`-checked, but **never applied**. The comment claims Fastly Compute's Rust SDK lacks per-request timeouts — that's not accurate: `fastly::Backend::builder().connect_timeout(..).first_byte_timeout(..).between_bytes_timeout(..).finish()` exposes them. `BackendConfig::from_url` creates a dynamic backend with **default** timeouts (seconds, not ms).

Under a DataDome API stall, every publisher request will block for the default first-byte timeout (tens of seconds), defeating the "fast fail-open" design and causing an origin-wide latency spike. This is the opposite of what the feature is supposed to deliver.

**Fix:** build a dedicated `Backend` once (cache in `DataDomeIntegration`) with the configured timeouts converted from ms to `Duration`, and send the validation request through it. If you truly can't enforce it, remove the field entirely rather than shipping a silent no-op with validation annotations.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -381,6 +439,167 @@ impl DataDomeIntegration {
...
+        // Extract DataDome cookie (clientid)
+        let datadome_cookie = req
+            .get_header(header::COOKIE)
+            .and_then(|h| h.to_str().ok())
+            .and_then(|cookies| {
+                cookies.split(';').find_map(|cookie| {
+                    let trimmed = cookie.trim();
+                    trimmed.strip_prefix("datadome=")
+                })
+            })
```
aram356: 🔧 **wrench** — Biased sampling hash. `fold(0, |acc, b| acc.wrapping_add(b))` is a sum-of-bytes hash with terrible distribution: `"1.2.3.4"` and `"4.3.2.1"` collide, and sums for IPv4 strings cluster heavily in the middle. At `sample_rate < 100` you will not get a representative slice, which compromises the gradual-rollout signal this flag exists to provide.

**Fix:**
```rust
use std::hash::{DefaultHasher, Hash, Hasher};

let mut hasher = DefaultHasher::new();
client_ip.hash(&mut hasher);
let sample = (hasher.finish() % 100) as u8;
```
`DefaultHasher` works on `wasm32-wasip1`.

crates/trusted-server-core/src/backend_router.rs
```diff
@@ -0,0 +1,507 @@
...
+            path_prefix: pattern.path_prefix.clone(),
+            path_regex: pattern.path_regex.as_ref().map(|_| OnceLock::new()),
+            path_regex_pattern: pattern.path_regex.clone(),
+        }
+    }
+
+    fn compiled_regex(&self) -> Option<&Regex> {
+        if let (Some(regex_cell), Some(pattern)) = (&self.path_regex, &self.path_regex_pattern) {
+            Some(regex_cell.get_or_init(|| {
+                Regex::new(pattern).expect("path_regex pattern should be valid")
```
aram356: 🔧 **wrench** — Config-derived regex compiled lazily with `expect()` on the request path. CLAUDE.md is explicit: *"Config-derived regex/pattern compilation must not use panic-prone `expect()`/`unwrap()`; invalid enabled config should surface as startup/config errors."* A typo in `backends.toml` will panic on the **first matching live request**, not at startup.

Note: `BackendRouter::new` already returns `Result<Self, Report<TrustedServerError>>` but never errors today — the signature is a lie.

**Fix:** compile all regexes eagerly in `BackendRouter::new`, return a `Report<TrustedServerError>` with `.attach()` on the offending pattern + backend id, and store an owned `Regex` directly. Drop the `OnceLock` regex cell.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -507,6 +779,12 @@ mod tests {
             api_origin: "https://api-js.datadome.co".to_string(),
             cache_ttl_seconds: 3600,
```
aram356, line 513: 🔧 **wrench** — No tests for the new server-side validation path: `validate_request`, `validate_request_server_side`, sampling, fail-open vs fail-closed, missing-key guard, multi-cookie extraction. The existing tests only build a default config with `server_side_enabled: false`.

CLAUDE.md requires tests for new code, and the PR checklist is unchecked. Please add unit tests covering at minimum: disabled fast-path; missing-key + `fail_open=true`; missing-key + `fail_open=false`; `sample_rate=0` skip; cookie parsing from a multi-cookie `Cookie` header.

crates/trusted-server-core/src/settings.rs
```diff
@@ -327,6 +327,76 @@ impl Default for Proxy {
...
+    /// Optional path prefix to match (e.g., "/.api/").
+    pub path_prefix: Option<String>,
+    /// Optional path regex pattern to match (e.g., "^/image/upload/").
+    pub path_regex: Option<String>,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, Validate)]
+pub struct BackendRoutingConfig {
+    /// Origin URL for this backend (e.g., <https://raven-public.prod.saymedia.com>).
+    /// The actual Fastly backend will be created dynamically at request time.
```
aram356: 🔧 **wrench** — `BackendRoutingConfig.origin_url` is only validated as `length(min=1)`, not as a URL. A malformed `origin_url` will hit the `url::Url::parse(origin_url).ok().unwrap_or_else(|| origin_url.to_string())` fallback in `publisher.rs` and send the raw URL string as the `Host` header — silent misrouting rather than a config failure.

**Fix:** `#[validate(url)]` on `origin_url` (mirroring `validation_endpoint` in `DataDomeConfig`).

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -381,6 +439,167 @@ impl DataDomeIntegration {
...
+        }
+        if let Some(accept) = req.get_header(header::ACCEPT) {
+            validation_req.set_header(header::ACCEPT, accept);
+        }
+        if let Some(accept_lang) = req.get_header(header::ACCEPT_LANGUAGE) {
+            validation_req.set_header(header::ACCEPT_LANGUAGE, accept_lang);
+        }
+
+        // Note: Fastly Compute doesn't support per-request timeouts via Rust API
+        // The backend timeout is configured at service level (validation_timeout_ms documents intent)
```
aram356: ❓ **question** — Does URL-encoding the `x-datadome-params:ip` and `x-datadome-params:clientid` header values match DataDome's documented API contract? Most DataDome fastly/cloudflare module reference implementations pass these raw. URL-encoding a cookie value may double-encode characters DataDome expects verbatim and break fingerprinting. Please link the spec you're targeting so we can verify.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -381,6 +439,167 @@ impl DataDomeIntegration {
...
+
+    /// Extract client IP address from request (test-safe).
+    ///
+    /// Returns "0.0.0.0" if client IP is unavailable (e.g., in Viceroy test environment).
+    fn get_client_ip(req: &Request) -> String {
+        #[cfg(test)]
+        {
+            // In test mode, always return a default IP since Viceroy doesn't support get_client_ip_addr
+            let _ = req; // Suppress unused variable warning
+            "127.0.0.1".to_string()
```
aram356: ♻️ **refactor** — `#[cfg(test)]` fork inside a production function is a test seam leaking into production code. It also disables the real code path in tests, so you can never unit-test the actual IP extraction. Prefer injecting the IP via function parameter or a small trait so tests can pass `"127.0.0.1"` without a cfg split.

crates/trusted-server-core/src/backend_router.rs
```diff
@@ -0,0 +1,507 @@
...
+    default_origin: String,
+    default_certificate_check: bool,
+}
+
+#[derive(Debug, Clone)]
+pub struct BackendRoute {
+    pub origin_url: String,
+    pub certificate_check: bool,
+    #[allow(dead_code)]
+    domains: Vec<String>,
```
aram356: ⛏ **nitpick** — `#[allow(dead_code)] domains: Vec<String>` — if `domain_index` is the only consumer of domain membership, drop this field rather than allowing the lint.

crates/trusted-server-core/src/backend_router.rs
```diff
@@ -0,0 +1,507 @@
...
+            vec!["site-a.example.com", "site-b.example.com"],
+            "https://backend-a.example.com",
+        )];
+
+        let router = BackendRouter::new(
+            &backends,
+            "https://default-origin.com".to_string(),
+            true,
+        )
+        .unwrap();
```
aram356: ⛏ **nitpick** — CLAUDE.md applies to tests too: prefer `.expect("should build router")` over `.unwrap()` here and in the rest of this module's tests.

crates/trusted-server-adapter-fastly/backends.toml
```diff
@@ -0,0 +1,176 @@
+# Arena Group / SayMedia backend routing configuration.
+#
+# This file is merged into the embedded binary config at build time by
+# crates/common/build.rs. It is separate from trusted-server.toml to keep
```
aram356: ⛏ **nitpick** — Stale path: `crates/common/build.rs` → `crates/trusted-server-core/build.rs` after the rename.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -271,17 +273,83 @@ pub fn handle_publisher_request(
...
+    // DataDome server-side validation (if enabled)
+    // This validates requests at the edge before they reach the origin
+    match crate::integrations::datadome::validate_request_server_side(settings, &req) {
+        Ok(true) => {
+            // Request allowed - continue to origin
+            log::debug!("[datadome] Request validated - proceeding to origin");
+        }
+        Ok(false) => {
+            // Request blocked - bot detected
+            log::warn!("[datadome] Request blocked by server-side validation");
```
aram356: 🤔 **thinking** — `"Access Denied - Bot Protection"` leaks the existence of the bot-protection layer and contradicts the generic-error-body principle that just landed in #530. Return a plain `"Forbidden"` body and keep the fingerprint out of the response.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -271,17 +273,83 @@ pub fn handle_publisher_request(
...
-        &settings.publisher.origin_url,
-        settings.proxy.certificate_check,
-    )?;
-    let origin_host = settings.publisher.origin_host();
+    let request_path = req.get_path();
+
+    // Select origin URL using backend router (if configured) or fallback to default.
+    // Settings are fixed at startup, so first-caller-wins initialization is correct —
+    // the same binary serves all requests with the same config.
+    static ROUTER: OnceLock<Option<BackendRouter>> = OnceLock::new();
```
aram356: 🤔 **thinking** — Module-level `OnceLock<BackendRouter>` inside a request handler captures the *first* `&settings` it sees. It's correct for Fastly Compute today, but if a test or hot-reload path ever invokes this with a different `Settings`, the router is silently stale. Consider building the router once at `Settings` construction and storing it on `Settings` (or a handler context struct) instead.

crates/trusted-server-core/build.rs
```diff
@@ -43,17 +66,57 @@ fn main() {
...
+                Value::String(_) | Value::Number(_) | Value::Bool(_) => {
+                    // Leaf node - create environment variable
+                    let env_var = format!(
+                        "{}{}{}",
+                        settings::ENVIRONMENT_VARIABLE_PREFIX,
+                        settings::ENVIRONMENT_VARIABLE_SEPARATOR,
+                        new_path.join(settings::ENVIRONMENT_VARIABLE_SEPARATOR)
+                    );
+                    env_vars.insert(env_var);
+                }
```
aram356: ♻️ **refactor** — The recursion silently skips `Value::Array` (`_ => {}`). That means per-element `backends[*]` env overrides are impossible today. If intentional, add a `// arrays intentionally skipped` comment; otherwise handle arrays so operators can override a specific backend's origin at build time.

crates/trusted-server-core/src/settings.rs
```diff
@@ -327,6 +327,76 @@ impl Default for Proxy {
...
+    #[validate(nested)]
+    pub path_patterns: Vec<PathPattern>,
+    /// Enable TLS certificate verification for this backend.
+    #[serde(default = "default_certificate_check")]
+    pub certificate_check: bool,
+    /// Unique identifier for logging/debugging (optional).
+    #[serde(default)]
+    pub id: Option<String>,
+    /// Backend-specific timeouts (unused - for future use).
+    #[serde(default)]
```
aram356: 🤔 **thinking** — `BackendTimeouts` is committed as a schema but the comment says `// Backend-specific timeouts (unused - for future use).`. CLAUDE.md: *"Don't design for hypothetical future requirements."* Either wire this into `BackendConfig::from_url` (which would also fix the DataDome timeout issue above) or drop the field until it has a consumer.

aram356 (changes requested):
## Summary

Feature design is reasonable (validate at edge, route by host/path), but implementation has several correctness issues, two CI failures not fixed after the recent `common → core` rename, a panic-on-bad-config path, a non-functional validation timeout, a biased sampling hash, and no tests for the new DataDome server-side validation. Addressing the 🔧 items is required before merge.

## Blocking

### 🔧 wrench

- **Doctests reference stale crate name**: `trusted_server_common` → `trusted_server_core` (`backend_router.rs:183`, `datadome.rs:725-726`). These are why `cargo test` is red in CI.
- **`cargo fmt` failing**: run `cargo fmt --all` and commit. The PR checklist is unchecked — please tick the boxes you've actually run.
- **`validation_timeout_ms` is documented but never applied** (`datadome.rs:587-589`): Fastly's Rust SDK *does* support per-request backend timeouts via `Backend::builder()`. `BackendConfig::from_url` creates a dynamic backend with default timeouts (seconds). Under a DataDome API stall, every request blocks for the default first-byte timeout, defeating fast fail-open and causing an origin-wide latency spike.
- **Biased sampling hash** (`datadome.rs:522-524`): sum-of-bytes hash has terrible distribution. Use `std::hash::DefaultHasher` (works on wasm32-wasip1).
- **Regex compiled lazily with `expect()` on the request path** (`backend_router.rs:64-67`): violates CLAUDE.md's rule that config-derived regex compilation must surface as startup errors. A typo in `backends.toml` will panic on the first matching live request. Compile eagerly in `BackendRouter::new` and error out — the `Result` return type already exists but is a lie.
- **Missing tests for new DataDome server-side validation**: no tests cover `validate_request`, `validate_request_server_side`, sampling, fail-open branches, missing-key guard, or cookie extraction.
- **`BackendRoutingConfig.origin_url` not validated as URL** (`settings.rs`): currently `length(min=1)`; a malformed URL silently falls through and is sent as the `Host` header. Use `#[validate(url)]`.

### ❓ question

- **PR body claims "Synthetic cookie is only set on responses when not already present"** — I can't find that change in the diff. Is it missing, or was it split to another PR? Please update the body or push the change.
- **Does URL-encoding `x-datadome-params:ip` and `x-datadome-params:clientid` match DataDome's API contract?** (`datadome.rs:565,570). Most DataDome integration references pass these raw. Please link the spec.

## Non-blocking

### 🤔 thinking

- **`[[backends]].timeouts` is dead code** — either wire it into `BackendConfig::from_url` (which would also resolve the DataDome timeout gap) or drop the field until it has a consumer. CLAUDE.md: no speculative fields.
- **`BackendRouter::new` signature returns `Result` but never errors today.** Either eagerly compile regexes and make it honest, or simplify to `-> Self`.
- **`static OnceLock<BackendRouter>` inside the handler** captures the first `&settings` seen. Prefer building the router once during `Settings` construction and storing it on a handler context struct.
- **403 body `"Access Denied - Bot Protection"`** leaks the bot-protection layer and contradicts #530. Return a generic "Forbidden" body.
- **`503` on fail-closed DataDome error** couples your edge to DataDome's uptime. Add an ops warning to `trusted-server.toml` against `fail_open=false` without a standby.
- **Validation runs on every publisher request when enabled** — doubles edge request count on the critical latency path. Please document expected p50/p99 overhead for the Arena Group rollout.

### ♻️ refactor

- **`collect_env_vars` silently skips `Value::Array`** (`build.rs`): per-element `backends[*]` env overrides are impossible today. Intentional? Either comment or handle arrays.
- **`#[cfg(test)]` fork inside `get_client_ip`** is a production test seam and prevents unit-testing the real IP extraction. Prefer a function parameter or trait seam.

### 🌱 seedling

- **Path-pattern lookup is O(routes × patterns) per request.** Fine for current scale (one Arena route, no path patterns). Precompute an index if tenant count grows.

### 📝 note

- **`.gitignore` adds `*.vcl` globally** but no `.vcl` files exist in the repo. Consider scoping.

### ⛏ nitpick

- Stale path in `backends.toml` header: `crates/common/build.rs` → `crates/trusted-server-core/build.rs`.
- `.unwrap()` in `backend_router.rs` tests — CLAUDE.md wants `.expect("should ...")` in tests too.
- `#[allow(dead_code)] domains: Vec<String>` on `BackendRoute` — if `domain_index` is the only consumer, drop the field.
- Log prefix `[datadome]` — prefer `log::info!(target: "datadome", ...)`.
- `trusted-server.toml` schema example should document `timeouts` (or drop it per the note above).

### 👍 praise

- **Good separation of customer data from template config** — `backends.toml` under `crates/trusted-server-adapter-fastly/` merged at build time keeps tenant lists out of the shared template.
- **Fail-open default with a sampling knob** is the right posture for a new edge-gate rollout.
- **Thorough doctest `# Examples` on `normalize_domain` and `validate_request_server_side`** — once the crate name is fixed.

## CI Status

- fmt: **FAIL**
- rust tests: **FAIL** (two doctests reference unresolved `trusted_server_common` crate)
- clippy: not reported separately
- vitest / browser / format-docs / CodeQL: PASS

crates/common/build.rs
```diff
@@ -9,36 +9,99 @@ mod auction_config_types;
...
+fn merge_toml() {
     // Read init config
     let init_config_path = Path::new(TRUSTED_SERVER_INIT_CONFIG_PATH);
     let toml_content = fs::read_to_string(init_config_path)
         .unwrap_or_else(|_| panic!("Failed to read {init_config_path:?}"));
 
-    // Merge base TOML with environment variable overrides and write output
-    let settings = settings::Settings::from_toml_and_env(&toml_content)
+    // For build time: use from_toml to parse with environment variables
+    let mut settings = settings::Settings::from_toml(&toml_content)
```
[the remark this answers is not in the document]
ChristianPavilonis: The current `build.rs` at line 69 already uses `from_toml_and_env`. This comment referenced an earlier version at `crates/common/build.rs` before the crate rename.

crates/trusted-server-core/build.rs
```diff
@@ -9,36 +9,99 @@ mod auction_config_types;
...
+    rerun_if_changed();
+}
+
+fn rerun_if_changed() {
+    // Watch the root trusted-server.toml file for changes
+    println!("cargo:rerun-if-changed={}", TRUSTED_SERVER_INIT_CONFIG_PATH);
+    println!("cargo:rerun-if-changed={}", BACKENDS_CONFIG_PATH);
+
+    // Create a default Settings instance and convert to JSON to discover all fields
+    let default_settings = settings::Settings::default();
```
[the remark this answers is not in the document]
ChristianPavilonis: The `from_toml_and_env` call already handles env var overrides at build time. The `rerun_if_changed` enumeration uses `Settings::default()` which may miss dynamically-keyed HashMap fields, but this is a pre-existing limitation that doesn't affect correctness — it only risks a missed incremental rebuild if an env var changes. Not addressing in this PR.

crates/common/build.rs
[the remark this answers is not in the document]
ChristianPavilonis: The current `build.rs` at line 69 already uses `from_toml_and_env`. This comment referenced the old `crates/common/build.rs` path before the crate rename.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -381,6 +439,167 @@ impl DataDomeIntegration {
...
+            }
+        };
+
+        // Extract client IP (test-safe)
+        let client_ip = Self::get_client_ip(req);
+
+        // Sample rate check: only validate sample_rate% of requests
+        if self.config.sample_rate < 100 {
+            // Use client IP hash for deterministic sampling
+            let hash = client_ip.bytes().fold(0u32, |acc, b| acc.wrapping_add(u32::from(b)));
```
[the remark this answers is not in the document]
ChristianPavilonis: Fixed: replaced the byte-sum hash with a deterministic polynomial hash (`acc.wrapping_mul(31).wrapping_add(u32::from(b))`), extracted as `hash_client_ip`. Avoids permutation collisions and is process-stable on wasm32 (unlike `DefaultHasher`, which uses a random seed per process invocation).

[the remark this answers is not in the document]
ChristianPavilonis: Fixed: replaced byte-sum with a deterministic polynomial hash (`wrapping_mul(31)`) extracted as `hash_client_ip`. `DefaultHasher` was considered but rejected — Fastly Compute gives each request a fresh process invocation, so `DefaultHasher`'s random seed makes the hash non-deterministic per request.

crates/trusted-server-core/src/integrations/datadome.rs
ChristianPavilonis: Fixed: replaced `fold(0, |acc, b| acc.wrapping_add(b))` with a deterministic polynomial hash (`wrapping_mul(31).wrapping_add(b)`), extracted as the testable `hash_client_ip` helper.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -125,6 +125,38 @@ pub struct DataDomeConfig {
...
+    /// Validation endpoint (default: <https://api-fastly.datadome.co>)
+    #[serde(default = "default_validation_endpoint")]
+    #[validate(url)]
+    pub validation_endpoint: String,
+
+    /// Validation request timeout in milliseconds (default: 200ms)
+    /// Lower timeout = faster fail-open on `DataDome` API issues
+    #[serde(default = "default_validation_timeout_ms")]
+    #[validate(range(min = 50, max = 1000))]
+    pub validation_timeout_ms: u64,
```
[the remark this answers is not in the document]
ChristianPavilonis: Fixed: `BackendConfig::from_url_with_first_byte_timeout` is now called with `Duration::from_millis(self.config.validation_timeout_ms)`, applying the configured timeout to the DataDome backend's first-byte wait. The stale comment claiming Fastly doesn't support per-request timeouts has been removed.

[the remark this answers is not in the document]
ChristianPavilonis: Fixed: `validation_timeout_ms` is now applied via `BackendConfig::from_url_with_first_byte_timeout`.

crates/trusted-server-core/src/integrations/datadome.rs
ChristianPavilonis: Fixed: `BackendConfig::from_url_with_first_byte_timeout` is now called with `Duration::from_millis(self.config.validation_timeout_ms)`. The stale comment has been removed.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -204,20 +206,96 @@ pub fn handle_publisher_request(
...
     log::debug!(
         "Proxying to dynamic backend: {} (from {})",
         backend_name,
-        settings.publisher.origin_url
+        origin_url
     );
+
+    // DataDome server-side validation (if enabled)
+    // This validates requests at the edge before they reach the origin
+    match crate::integrations::datadome::validate_request_server_side(settings, &req) {
```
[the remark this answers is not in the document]
ChristianPavilonis: Acknowledged. Filtering to document/navigational requests only is tracked as a follow-up. The `sample_rate` config allows reducing validation overhead in the interim.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -204,20 +206,96 @@ pub fn handle_publisher_request(
...
+    let router = ROUTER.get_or_init(|| {
+        if settings.backends.is_empty() {
+            return None;
+        }
+        BackendRouter::new(
+            &settings.backends,
+            settings.publisher.origin_url.clone(),
+            settings.proxy.certificate_check,
+        )
+        .ok()
```
ChristianPavilonis: Fixed: `.map_err(|e| { log::error!("Failed to build backend router: {:?}", e); e })` is now chained before `.ok()`.

crates/trusted-server-core/src/backend_router.rs
ChristianPavilonis: Fixed: `CompiledPathPattern::new` now returns `Result<Self, Report<TrustedServerError>>` and compiles the regex eagerly via `Regex::new(s).map_err(...)`. `BackendRouter::new` propagates the error, making the `Result` return honest.

ChristianPavilonis: Fixed: eager compilation in `CompiledPathPattern::new` returning `Result`. `BackendRouter::new` now propagates the error at startup rather than panicking on the first matching request.

crates/trusted-server-core/src/backend_router.rs
```diff
@@ -0,0 +1,507 @@
...
+        default_origin: String,
+        default_certificate_check: bool,
+    ) -> Result<Self, Report<TrustedServerError>> {
+        let mut domain_index = HashMap::new();
+        let mut routes = Vec::with_capacity(backends.len());
+
+        for (idx, backend) in backends.iter().enumerate() {
+            for domain in &backend.domains {
+                let normalized = normalize_domain(domain);
+                domain_index.insert(normalized, idx);
```
ChristianPavilonis: Fixed: domain insertion now uses `if let Some(existing_idx) = domain_index.insert(normalized.clone(), idx)` and emits a `log::warn!` on collision.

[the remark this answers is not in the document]
ChristianPavilonis: Fixed: added a warning log when a domain appears in multiple backends.

crates/trusted-server-core/src/settings.rs
```diff
@@ -294,6 +294,76 @@ impl Default for Proxy {
...
+    #[validate(nested)]
+    pub path_patterns: Vec<PathPattern>,
+    /// Enable TLS certificate verification for this backend.
+    #[serde(default = "default_certificate_check")]
+    pub certificate_check: bool,
+    /// Unique identifier for logging/debugging (optional).
+    #[serde(default)]
+    pub id: Option<String>,
+    /// Backend-specific timeouts (unused - for future use).
```
ChristianPavilonis: Fixed: removed the `BackendTimeouts` struct and the `timeouts` field from `BackendRoutingConfig`.

crates/trusted-server-core/src/settings.rs
```diff
@@ -294,6 +294,76 @@ impl Default for Proxy {
...
+    #[serde(default)]
+    #[validate(nested)]
+    pub timeouts: BackendTimeouts,
```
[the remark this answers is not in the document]
ChristianPavilonis: Fixed: `BackendTimeouts` struct and its `timeouts` field removed.

crates/trusted-server-core/src/settings.rs
ChristianPavilonis: Fixed: `BackendTimeouts` struct and the `timeouts` field on `BackendRoutingConfig` have been removed.

crates/trusted-server-core/src/backend_router.rs
```diff
@@ -0,0 +1,507 @@
...
+    domain_index: HashMap<String, usize>,
+    default_origin: String,
+    default_certificate_check: bool,
+}
+
+#[derive(Debug, Clone)]
+pub struct BackendRoute {
+    pub origin_url: String,
+    pub certificate_check: bool,
+    #[allow(dead_code)]
```
ChristianPavilonis: Fixed: removed the field entirely.

crates/trusted-server-core/src/backend_router.rs
ChristianPavilonis: Fixed: `domains` field removed from `BackendRoute`.

crates/common/src/publisher.rs
```diff
@@ -204,20 +206,96 @@ pub fn handle_publisher_request(
...
     // Generate synthetic identifiers before the request body is consumed.
     let synthetic_id = get_or_generate_synthetic_id(settings, &req)?;
+    let synthetic_cookie_prefix = format!("{}=", COOKIE_SYNTHETIC_ID);
+    let has_synthetic_cookie = req
+        .get_header(header::COOKIE)
+        .and_then(|h| h.to_str().ok())
+        .map(|cookies| {
+            cookies
+                .split(';')
+                .any(|cookie| cookie.trim_start().starts_with(&synthetic_cookie_prefix))
```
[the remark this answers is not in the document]
ChristianPavilonis: The `trim_start()` referenced here was in the old `crates/common/src/publisher.rs`. The current `publisher.rs` no longer has this cookie-parsing code path. DataDome's cookie extraction now uses `cookie.trim()` consistently via the extracted `extract_datadome_cookie` helper.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -475,14 +694,67 @@ fn build(settings: &Settings) -> Option<Arc<DataDomeIntegration>> {
...
+///     }
+/// }
+/// ```
+pub fn validate_request_server_side(
+    settings: &Settings,
+    req: &Request,
+) -> Result<bool, Report<TrustedServerError>> {
+    // Cache the integration after the first call. Settings are fixed at startup,
+    // so first-caller-wins is correct — the same binary serves all requests.
+    static INTEGRATION: OnceLock<Option<Arc<DataDomeIntegration>>> = OnceLock::new();
```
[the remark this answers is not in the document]
ChristianPavilonis: Fixed: added 7 unit tests — `validation_disabled_fast_path`, `missing_key_fail_open`, `missing_key_fail_closed`, `sample_rate_zero_skips_all_requests`, `sample_rate_full_does_not_skip`, `cookie_extracted_from_multi_cookie_header`, and `deterministic_sampling_hash`.

crates/trusted-server-core/src/integrations/datadome.rs
ChristianPavilonis: Fixed: added 7 unit tests covering the fast path, fail-open/closed, sampling bounds, cookie extraction, and hash determinism.

crates/trusted-server-core/src/backend_router.rs
ChristianPavilonis: Fixed: changed `trusted_server_common` → `trusted_server_core`.

crates/trusted-server-core/src/integrations/datadome.rs
ChristianPavilonis: Fixed: both `use trusted_server_common` lines changed to `trusted_server_core`.

crates/trusted-server-core/src/settings.rs
ChristianPavilonis: Fixed: changed `#[validate(length(min = 1))]` → `#[validate(url)]`. HTTP URLs are intentionally accepted (internal backends may not use TLS).

crates/trusted-server-core/src/integrations/datadome.rs
ChristianPavilonis: Fixed: `x-datadome-params:ip` now passes the IP raw; `x-datadome-params:clientid` keeps URL encoding. Confirmed against the DataDome Fastly VCL module reference (`req.http.fastly-client-ip` raw, `urlencode(req.http.x-datadome-clientid)` encoded).

crates/trusted-server-core/src/integrations/datadome.rs
ChristianPavilonis: Fixed: extracted `validate_request_inner(&self, client_ip: &str, req: &Request)` containing all business logic. `validate_request` calls `get_client_ip` then delegates. `get_client_ip` no longer has a `#[cfg(test)]` fork.

crates/trusted-server-core/src/backend_router.rs
ChristianPavilonis: Fixed: all `.unwrap()` calls in the test module replaced with `.expect("should ...")`.

crates/trusted-server-adapter-fastly/backends.toml
ChristianPavilonis: Fixed: header comment updated to `crates/trusted-server-core/build.rs`.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Fixed: body changed to `"Forbidden"`.

crates/trusted-server-core/build.rs
ChristianPavilonis: Fixed: added a comment explaining this is intentional — per-element env var overrides for arrays aren't supported.

crates/trusted-server-core/src/backend_router.rs
```diff
@@ -0,0 +1,507 @@
...
+/// ```
+/// use trusted_server_common::backend_router::normalize_domain;
+///
+/// assert_eq!(normalize_domain("WWW.EXAMPLE.COM"), "example.com");
+/// assert_eq!(normalize_domain("www.example.com"), "example.com");
+/// assert_eq!(normalize_domain("example.com"), "example.com");
+/// assert_eq!(normalize_domain("sub.example.com"), "sub.example.com");
+/// ```
+#[must_use]
+pub fn normalize_domain(domain: &str) -> String {
```
ChristianPavilonis: Fixed: `normalize_domain` now returns `Cow<'_, str>`. Returns `Cow::Borrowed` when the host is already lowercase with no `www.` prefix (zero allocations on the hot request path), `Cow::Owned` only when lowercasing is needed. The `HashMap` lookup uses `.as_ref()` to get `&str` without any conversion.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Fixed: removed `static OnceLock<Option<BackendRouter>>` entirely. In Fastly Compute, each request runs in its own process so the lock provided no caching benefit while causing exactly the test-isolation problem you described. The router is now built directly from `settings` on each call.

crates/trusted-server-core/src/backend_router.rs
```diff
@@ -0,0 +1,524 @@
...
+            .path_regex
+            .as_deref()
+            .map(|s| {
+                Regex::new(s).map_err(|e| {
+                    Report::new(TrustedServerError::Configuration {
+                        message: format!("Invalid path_regex pattern `{s}`: {e}"),
+                    })
+                })
+            })
+            .transpose()?;
```
aram356: 🔧 **wrench** — Invalid `path_regex` is detected too late and then silently disables routing.

`validate_path_pattern` (settings.rs) only checks `path_prefix`/`path_regex` mutual exclusivity. Regex syntax is compiled here lazily, and `BackendRouter::new` is invoked per-request in `handle_publisher_request` with `.ok()` swallowing errors. A typo in a customer's `backends.toml` will:

1. Pass `build.rs` parsing
2. Pass runtime `Settings::validate()`
3. Log an error on every request
4. Silently fall through to default origin → routing broken, app keeps serving

This violates CLAUDE.md: *"config-derived regex/pattern compilation must not use panic-prone expect()/unwrap(); invalid enabled config should surface as startup/config errors"* and *"Invalid enabled integrations/providers must not be silently logged-and-disabled during startup or registration."*

**Fix**: validate regex syntax inside `validate_path_pattern` (try `Regex::new(r).map_err(...)`), and/or construct a `BackendRouter` inside `Settings::validate()` so bad config → startup failure. Then in `handle_publisher_request` the router build becomes infallible (or unwraps an already-validated router).

crates/trusted-server-core/src/publisher.rs
```diff
@@ -271,17 +272,82 @@ pub fn handle_publisher_request(
...
     log::debug!(
         "Proxying to dynamic backend: {} (from {})",
         backend_name,
-        settings.publisher.origin_url
+        origin_url
     );
+
+    // DataDome server-side validation (if enabled)
+    // This validates requests at the edge before they reach the origin
+    match crate::integrations::datadome::validate_request_server_side(settings, &req) {
```
aram356: 🔧 **wrench** — DataDome validation runs on every publisher request, including static assets.

`handle_publisher_request` is invoked for every proxied path — HTML, CSS, JS, images, fonts. With `server_side_enabled=true` every one of those enters `validate_request_server_side`. Unsampled requests still pay JSON-deserialize + validate + Arc-alloc; sampled requests pay a 200 ms round-trip to `api-fastly.datadome.co` *per asset*.

This is the same finding from the 2026-03-20 and 2026-04-14 reviews at this exact line — still not addressed.

**Fix**: filter to navigational/document requests before invoking (skip common static extensions, or use `Sec-Fetch-Dest`/`Accept: image/*`). Document the selected filter so operators know what is and isn't protected.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -475,14 +735,64 @@ fn build(settings: &Settings) -> Option<Arc<DataDomeIntegration>> {
...
+///         }
+///     }
+/// }
+/// ```
+pub fn validate_request_server_side(
+    settings: &Settings,
+    req: &Request,
+) -> Result<bool, Report<TrustedServerError>> {
+    match build(settings) {
+        Some(integration) => integration.validate_request(req),
```
aram356: 🔧 **wrench** — Per-request DataDome config deserialization + Arc allocation.

Every `validate_request_server_side` call goes through `build(settings)` → `serde_json::from_value::<DataDomeConfig>(raw.clone())` + `config.validate()` + `Arc::new(...)`. Prior `OnceLock` caching was removed (rightly, for test isolation) but nothing replaced it. Combined with the finding in publisher.rs (validation running on every asset), each asset request pays the full deserialize/validate cost.

**Fix**: build the integration once during `get_settings()` (or at handler init if you introduce a `HandlerContext`) and pass `&DataDomeIntegration` into `validate_request_server_side` instead of `&Settings`.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -465,6 +715,16 @@ fn build(settings: &Settings) -> Option<Arc<DataDomeIntegration>> {
...
+            // IntegrationSettings derefs to HashMap<String, JsonValue>, so we can inspect
+            // the raw config even when enabled=false to detect the common misconfiguration
+            // of setting server_side_enabled=true without setting enabled=true.
+            if let Some(raw) = settings.integrations.get(DATADOME_INTEGRATION_ID) {
+                if let Ok(raw_config) = serde_json::from_value::<DataDomeConfig>(raw.clone()) {
+                    if raw_config.server_side_enabled {
+                        log::warn!("[datadome] server_side_enabled = true has no effect when enabled = false; server-side validation is inactive");
+                    }
+                }
+            }
```
aram356: 🔧 **wrench** — The misconfiguration warning re-deserializes on every disabled-but-configured request.

When `enabled=false` and a raw `[integrations.datadome]` section exists, this `Ok(None)` branch still runs `serde_json::from_value::<DataDomeConfig>(raw.clone())` per request to decide whether to log a warning. Move this to startup (one-time pass in `get_settings()` or `Settings::validate()`). A warning logged once at boot is what operators want; logging it per-request floods the log *and* keeps paying deserialize cost.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -381,6 +441,196 @@ impl DataDomeIntegration {
...
+        validation_req.set_header("x-datadome-params:requestmodulename", "TrustedServerRust");
+        validation_req.set_header("x-datadome-params:moduleversion", "1.0");
+        validation_req.set_header("x-datadome-params:ip", client_ip);
+        validation_req.set_header("x-datadome-params:method", req.get_method().as_str());
+
+        if !datadome_cookie.is_empty() {
+            validation_req.set_header(
+                "x-datadome-params:clientid",
+                urlencoding::encode(datadome_cookie).as_ref(),
+            );
```
aram356: ❓ **question** — `x-datadome-params:clientid` is still URL-encoded while `ip` is raw.

Commit `01e02e2c` changed `ip` to raw per DataDome's contract, but left `clientid` wrapped in `urlencoding::encode`. Since `extract_datadome_cookie` already returns just the token value (no `;` or `,`), the encoding is likely unnecessary and could mismatch what DataDome expects.

Please link the DataDome spec line that requires encoding for `clientid` but not `ip`, or drop the `urlencoding::encode`.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -381,6 +441,196 @@ impl DataDomeIntegration {
...
+        if !self.config.server_side_enabled {
+            return Ok(true);
+        }
+
+        // Check if API key is configured
+        let api_key = match &self.config.server_side_key {
+            Some(key) if !key.is_empty() => key,
+            _ => {
+                log::warn!("[datadome] Server-side validation enabled but no API key configured");
+                return Ok(self.config.fail_open);
```
aram356: ❓ **question** — Missing API key with `fail_open=false` blocks all traffic.

With `fail_open=false`, a missing key returns `Ok(false)` → publisher returns 403 → every request is blocked. That conflates a configuration error (missing key, e.g., during credential rotation) with a validation failure (network hiccup). Is this intentional?

If the intent is *"fail-closed means never let an uncertain request through, regardless of why it's uncertain"*, the test at line 1186-1195 should document that. Otherwise: missing key → always log-and-allow; `fail_open` only governs the network-failure branch. Missing-key-blocks-everyone is a footgun during key rollouts.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -271,17 +272,82 @@ pub fn handle_publisher_request(
...
+            &settings.backends,
+            settings.publisher.origin_url.clone(),
+            settings.proxy.certificate_check,
+        )
+        .map_err(|e| {
+            log::error!("Failed to build backend router: {:?}", e);
+            e
+        })
+        .ok()
+    };
```
aram356: 🤔 **thinking** — `BackendRouter` rebuilt per request, even though `settings.backends` is immutable after `build.rs`.

Fastly Compute is typically 1 request ≈ 1 process, so "once per process" happens to ≈ "once per request" — but (a) Fastly does keep some processes warm, and (b) this pattern wastes CPU on cold starts by re-building a ~170-entry HashMap + compiling regexes every handler call. Store the router on a `HandlerContext` built once alongside `Settings`, or have `Settings` lazily expose an `Arc<BackendRouter>`.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -271,17 +272,82 @@ pub fn handle_publisher_request(
...
+
+    let origin_host = url::Url::parse(origin_url)
+        .ok()
+        .and_then(|url| {
+            url.host_str().map(|host| match url.port() {
+                Some(port) => format!("{}:{}", host, port),
+                None => host.to_string(),
+            })
+        })
+        .unwrap_or_else(|| origin_url.to_string());
```
aram356: 🤔 **thinking** — `origin_host` derivation is duplicated.

This inline URL-parse → host[:port] logic is a verbatim copy of `Publisher::origin_host()` (settings.rs:62-72). Extract a free function `fn origin_host(url: &str) -> String` and have `Publisher::origin_host()` delegate to it.

crates/trusted-server-core/src/settings.rs
```diff
@@ -351,6 +383,11 @@ pub struct Settings {
     pub consent: ConsentConfig,
     #[serde(default)]
     pub proxy: Proxy,
+    /// Optional multi-backend routing configuration.
+    /// Use `BackendRouter::new()` to create a router from this config.
+    #[serde(default)]
```
aram356: ♻️ **refactor** — `backends` is a top-level `Settings` field.

Nests awkwardly next to `[publisher]`, `[synthetic]`, `[consent]`, `[proxy]`. A `[routing]` section (`settings.routing.backends`, `settings.routing.default_certificate_check`, etc.) would group routing knobs and leave room for future additions (global timeouts, sticky-session strategies) without further root-level growth.

crates/trusted-server-core/src/backend_router.rs
```diff
@@ -0,0 +1,524 @@
...
+    default_origin: String,
+    default_certificate_check: bool,
+}
+
+#[derive(Debug, Clone)]
+pub struct BackendRoute {
+    pub origin_url: String,
+    pub certificate_check: bool,
+    path_patterns: Vec<CompiledPathPattern>,
+}
```
aram356: ♻️ **refactor** — `BackendRoute` has `pub` fields but no public constructor.

`origin_url` and `certificate_check` are `pub`, `path_patterns` is private — external callers can read but can't build one. Either `pub(crate)` the struct, or keep all fields private and expose accessors.

crates/trusted-server-core/src/backend_router.rs
```diff
@@ -0,0 +1,524 @@
...
+    if domain.bytes().any(|b| b.is_ascii_uppercase()) {
+        let mut lower = domain.to_lowercase();
+        while lower.starts_with("www.") {
+            lower.drain(..4);
+        }
+        Cow::Owned(lower)
+    } else {
+        Cow::Borrowed(domain.trim_start_matches("www."))
+    }
+}
```
aram356: 🌱 **seedling** — IDN normalization edge case.

The Cow::Borrowed fast path skips lowercasing when the input has no ASCII-uppercase. A mixed-case Unicode host (`MüNCHen.de`) would match a config entry `müchen.de` only if the config is stored pre-lowercased. Current Prospect A list is all ASCII — no impact today. If IDN domains land, force the slow path uniformly.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -381,6 +441,196 @@ impl DataDomeIntegration {
...
+                    Ok(true)
+                } else {
+                    // Fail-closed: block request on validation errors
+                    Err(Report::new(Self::error(
+                        "Validation failed and fail_open=false",
+                    )))
+                }
+            }
+        }
+    }
```
aram356: 📝 **note** — Test gap: `validate_request_inner` HTTP path has no coverage.

All existing tests exit early (disabled, missing-key, sample=0, sample=100+missing-key). None exercises the actual HTTP request/response flow: header construction, 200→allow, 403→block, error→fail-open.

Viceroy can host a mock backend for this. Without it, a regression like lowercasing `x-datadome-params:Key` or dropping the method header would ship silently.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -245,7 +304,8 @@ impl DataDomeIntegration {
             .trim_start_matches("http://")
             .split('/')
             .next()
-            .unwrap_or("api-js.datadome.co")
+            // split() on &str always yields at least one element, so this is unreachable
+            .unwrap_or(url)
```
aram356: ⛏ **nitpick** — `.unwrap_or(url)` with "unreachable" comment is misleading.

Prefer `.expect("split always yields at least one element")` — clippy/future-you wouldn't wonder why the fallback was chosen.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -271,17 +272,82 @@ pub fn handle_publisher_request(
...
+    let router = if settings.backends.is_empty() {
+        None
+    } else {
+        BackendRouter::new(
+            &settings.backends,
+            settings.publisher.origin_url.clone(),
+            settings.proxy.certificate_check,
+        )
+        .map_err(|e| {
+            log::error!("Failed to build backend router: {:?}", e);
```
aram356: ⛏ **nitpick** — `{:?}` on `Report<_>` prints the full debug backtrace; `{err}` (Display) is usually what you want in edge logs. Same on line 343. Fine if you're sinking to Fastly real-time logging with structured output — confirm before changing.

.gitignore
```diff
@@ -36,3 +36,6 @@ src/*.html
 /crates/integration-tests/browser/test-results/
 /crates/integration-tests/browser/playwright-report/
 /crates/integration-tests/browser/.browser-test-state.json
+
```
aram356: ⛏ **nitpick** — Global `*.vcl` ignore with no `.vcl` files in tree. Scope to a specific directory or drop until there's content.

aram356 (changes requested):
## Summary

DataDome server-side validation and multi-backend routing. The author addressed most of the prior reviews' P0/P1 items (stale doctests, fmt, biased sampling hash, timeout wiring, eager regex compile, origin URL validation, `extract_host` fallback). What remains: a handful of config-validation and per-request-cost issues, plus two questions about DataDome API-contract details.

## Blocking

### 🔧 wrench

- **Invalid `path_regex` in `backends.toml` is detected too late, then silently disables routing**: `validate_path_pattern` only checks mutual exclusivity; regex compilation is lazy in `BackendRouter::new`, which is invoked per-request with `.ok()` swallowing errors. A typo passes build + startup validation, logs on every request, and silently falls through to default origin. Violates CLAUDE.md's config-validation rule. (backend_router.rs:62)
- **DataDome validation runs on every publisher request**, including static assets. With `server_side_enabled=true`, every CSS/JS/image call pays either deserialize+Arc-alloc (unsampled) or a 200 ms API round-trip (sampled). Flagged in prior reviews at this line; still not addressed. (publisher.rs:328)
- **Per-request DataDome config deserialization + Arc allocation**: `build(settings)` does `serde_json::from_value` + `validate()` + `Arc::new` on every call. Build once during `get_settings()` and pass `&DataDomeIntegration` into `validate_request_server_side`. (datadome.rs:791)
- **Misconfiguration-warning branch re-deserializes on every disabled-but-configured request**: move the warning check to startup — log once at boot, not per-request. (datadome.rs:727)

### ❓ question

- **`x-datadome-params:clientid` URL-encoded while `ip` is raw** — intentional? Link the spec or drop `urlencoding::encode`. (datadome.rs:573)
- **Missing API key with `fail_open=false` blocks all traffic** — conflates config error with validation failure; footgun during key rotation. Intentional? (datadome.rs:518)

## Non-blocking

### 🤔 thinking

- **`BackendRouter` rebuilt per request** despite `settings.backends` being immutable after `build.rs`. Store on a `HandlerContext`. (publisher.rs:290)
- **`origin_host` derivation duplicated** — the inline URL-parse in `handle_publisher_request` is a verbatim copy of `Publisher::origin_host()`. (publisher.rs:318)

### ♻️ refactor

- **`backends` top-level in `Settings`** — prefer a `[routing]` section. (settings.rs:388)
- **`BackendRoute` half-public** — mix of `pub` and private fields with no public constructor. (backend_router.rs:31)

### 🌱 seedling / 📌 out of scope

- IDN normalization edge case (no impact today; Arena Group list is ASCII).
- `backends.toml` commits ~170 publisher domains to the OSS repo — flagged three times now. Either move to a Fastly config store / private overlay, or add a comment documenting that it's intentionally public.

### 📝 note

- Test gap: `validate_request_inner` HTTP path has zero coverage — all tests exit early. A mock-backend test via Viceroy would catch regressions like header-casing changes. (datadome.rs:633)

### ⛏ nitpick

- `.unwrap_or(url)` with "unreachable" comment in `extract_host` — prefer `.expect(...)`. (datadome.rs:308)
- `{:?}` on `Report<_>` prints the full debug backtrace; `{err}` (Display) is usually preferable in edge logs. (publisher.rs:286, 343)
- `.gitignore` adds global `*.vcl` with no matching files in tree.

## CI Status (local run)

- fmt: **PASS**
- clippy `--all-targets --all-features -D warnings`: **PASS**
- `cargo test -p trusted-server-core --lib`: **PASS** (731 tests)
- `cargo test --workspace --doc`: **PASS**
- GitHub Actions on PR head: only `Analyze (javascript-typescript)` ran; Rust workflow did not run on this PR head — worth investigating (required-checks gating / branch protection skew), but code-level CI passes locally.

aram356:
@ChristianPavilonis Please split into 2 prs

PR #503, opened 2026-03-13
Replace regex with DOMParser in GPT document.write rewriting

### LINKED ISSUE #407: Regex-based HTML rewriting in document.write interception can fail-open
## Summary

The GPT guard uses a regex to match and rewrite GPT domain script URLs in `document.write` calls. If the regex fails to match (escaped quotes, unusual spacing, mixed quote styles), the original unproxied URL is passed through, causing the browser to load the GPT script directly from Google's CDN instead of through the first-party proxy.

## Refs

- `crates/js/lib/src/integrations/gpt/script_guard.ts` lines 206-230 — `rewriteHtmlString` regex

## Recommendation

Use DOM-based parsing (`DOMParser`) instead of regex, or fail-closed (block unmatched URLs rather than passing through).

## Context

ChristianPavilonis:
## Summary

- Replace fragile `SCRIPT_SRC_RE` regex with `DOMParser`-based HTML parsing in the GPT `document.write` / `document.writeln` interception layer, fixing edge cases where unquoted attributes, unusual spacing, or mixed quote styles caused the regex to fail-open and bypass the first-party proxy.
- Fail closed: when `DOMParser` is unavailable or throws, return an empty string instead of passing through unproxied GPT URLs.
- Add 8 new test cases covering single-quoted attrs, extra whitespace, multiple script tags, `writeln` parity, non-GPT passthrough, protocol-relative URLs, and both fail-closed scenarios.

## Changes

| File | Change |
| ---- | ------ |
| `crates/js/lib/src/integrations/gpt/script_guard.ts` | Replaced `SCRIPT_SRC_RE` regex and regex-based `rewriteHtmlString` with `DOMParser.parseFromString` implementation; uses `getAttribute("src")` + `replaceAll()` for robust URL swapping; adds fast-path skip when GPT domain not present; fails closed on error |
| `crates/js/lib/test/integrations/gpt/script_guard.test.ts` | Added 8 edge-case tests: single-quoted src, extra whitespace, multiple scripts, writeln parity, non-GPT passthrough, protocol-relative URLs, DOMParser unavailable, DOMParser throws |

## Closes

Closes #407

## Test plan

## Checklist

- [x] New code has tests

### DIFF
--- a/.github/actions/setup-integration-test-env/action.yml
+++ b/.github/actions/setup-integration-test-env/action.yml
@@ -70,6 +70,8 @@ runs:
       shell: bash
       env:
         TRUSTED_SERVER__PUBLISHER__ORIGIN_URL: http://127.0.0.1:${{ inputs.origin-port }}
+        TRUSTED_SERVER__PUBLISHER__PROXY_SECRET: integration-test-proxy-secret
+        TRUSTED_SERVER__SYNTHETIC__SECRET_KEY: integration-test-secret-key
         TRUSTED_SERVER__PROXY__CERTIFICATE_CHECK: "false"
       run: cargo build --bin trusted-server-fastly --release --target wasm32-wasip1
 
--- a/scripts/integration-tests-browser.sh
+++ b/scripts/integration-tests-browser.sh
@@ -31,6 +31,8 @@ echo "==> Validating shared integration-test dependency versions..."
 # --- Build WASM binary ---
 echo "==> Building WASM binary (origin=http://127.0.0.1:$ORIGIN_PORT)..."
 TRUSTED_SERVER__PUBLISHER__ORIGIN_URL="http://127.0.0.1:$ORIGIN_PORT" \
+TRUSTED_SERVER__PUBLISHER__PROXY_SECRET="integration-test-proxy-secret" \
+TRUSTED_SERVER__SYNTHETIC__SECRET_KEY="integration-test-secret-key" \
 TRUSTED_SERVER__PROXY__CERTIFICATE_CHECK=false \
     cargo build --bin trusted-server-fastly --release --target wasm32-wasip1
 
--- a/scripts/integration-tests.sh
+++ b/scripts/integration-tests.sh
@@ -52,6 +52,8 @@ fi
 
 echo "==> Building WASM binary (origin=http://127.0.0.1:$ORIGIN_PORT)..."
 TRUSTED_SERVER__PUBLISHER__ORIGIN_URL="http://127.0.0.1:$ORIGIN_PORT" \
+TRUSTED_SERVER__PUBLISHER__PROXY_SECRET="integration-test-proxy-secret" \
+TRUSTED_SERVER__SYNTHETIC__SECRET_KEY="integration-test-secret-key" \
 TRUSTED_SERVER__PROXY__CERTIFICATE_CHECK=false \
     cargo build --bin trusted-server-fastly --release --target wasm32-wasip1
 


crates/js/lib/src/integrations/gpt/script_guard.ts
```diff
@@ -190,43 +190,53 @@ function rewriteLinkHref(element: HTMLLinkElement): void {
...
 function rewriteHtmlString(html: string): string {
-  SCRIPT_SRC_RE.lastIndex = 0;
-  if (!SCRIPT_SRC_RE.test(html)) return html;
-  SCRIPT_SRC_RE.lastIndex = 0;
-
-  return html.replace(SCRIPT_SRC_RE, (_match, prefix: string, url: string, suffix: string) => {
-    const { url: rewrittenUrl, didRewrite } = maybeRewrite(url);
-    if (!didRewrite) {
-      return `${prefix}${url}${suffix}`;
+  if (typeof DOMParser === 'undefined') {
```
aram356: 🔧 **wrench** — Missing fast-path guard: non-GPT HTML silently dropped when DOMParser unavailable

`rewriteHtmlString` enters the DOMParser / fail-closed code path for **every** string, including non-GPT HTML. The old regex code had a fast-path exit (`if (!SCRIPT_SRC_RE.test(html)) return html;`). Without it:

1. **Correctness**: If `DOMParser` is unavailable or throws, `document.write('<p>Hello</p>')` returns `''` — silently dropping legitimate non-GPT content. In ad-tech pages that use `document.write` for other purposes, this could break the page entirely.
2. **Performance**: Every `document.write` call parses the full HTML via `DOMParser`, even when the string has nothing to do with GPT.

**Fix** — add the fast-path at the top, before the `DOMParser` availability check:

```typescript
function rewriteHtmlString(html: string): string {
  if (!html.includes(GPT_DOMAIN)) return html;

  if (typeof DOMParser === 'undefined') {
```

crates/js/lib/src/integrations/gpt/script_guard.ts
```diff
@@ -190,43 +190,53 @@ function rewriteLinkHref(element: HTMLLinkElement): void {
...
+    for (const script of scripts) {
+      const rawSrc = script.getAttribute('src') ?? '';
+      const { url: rewrittenUrl, didRewrite } = maybeRewrite(rawSrc);
+      if (!didRewrite) continue;
+
+      log.info(`${LOG_PREFIX}: rewriting document.write script src`, {
+        original: rawSrc,
+        rewritten: rewrittenUrl,
+      });
+      result = result.replaceAll(rawSrc, rewrittenUrl);
```
aram356: 🤔 **thinking** — `replaceAll` is a blunt instrument for URL substitution

This replaces the URL **everywhere** in the HTML string — not just inside `src` attributes. If the same URL appeared in a `data-*` attribute, a comment, or text content, those occurrences would also be rewritten. In practice GPT's `document.write` output is simple `<script src="..."></script>`, so this is unlikely to cause issues today, but it's less precise than the old regex approach which was scoped to `src` attributes.

crates/js/lib/src/integrations/gpt/script_guard.ts
```diff
@@ -190,43 +190,53 @@ function rewriteLinkHref(element: HTMLLinkElement): void {
...
+ * edge-cases (unquoted attributes, unusual spacing, mixed quote styles)
+ * are handled by the browser's native parser. The raw `getAttribute`
+ * value is swapped in the original HTML string so the surrounding markup
+ * is preserved verbatim.
+ *
+ * If the GPT domain is present in the HTML but `DOMParser` is
+ * unavailable or throws, the function **fails closed** (returns an
+ * empty string) rather than passing the unproxied URL through.
  */
 function rewriteHtmlString(html: string): string {
```
prk-Jr: 🔧 **wrench** — Non-GPT `document.write` output gets dropped when `DOMParser` is unavailable or broken.

`rewriteHtmlString()` returns `''` before it knows whether this HTML actually contains a GPT URL. In the failure modes this PR adds explicit handling for, benign publisher markup and non-GPT scripts are erased instead of passing through unchanged. That is a breaking change from the previous behavior and it contradicts the docstring/log wording here.

**Fix**: keep a cheap GPT-domain fast path and only fail closed once the input is known to contain a GPT candidate.

crates/js/lib/src/integrations/gpt/script_guard.ts
```diff
@@ -190,43 +190,53 @@ function rewriteLinkHref(element: HTMLLinkElement): void {
...
+    return '';
+  }
+
+  try {
+    const doc = new DOMParser().parseFromString(html, 'text/html');
+    const scripts = doc.querySelectorAll('script[src]');
+    let result = html;
+
+    for (const script of scripts) {
+      const rawSrc = script.getAttribute('src') ?? '';
```
prk-Jr: 🔧 **wrench** — HTML-escaped query strings can bypass this rewrite and leak back to the Google host.

This reads `src` via `getAttribute()` and then does a raw `replaceAll(rawSrc, rewrittenUrl)` against the original HTML string. For serialized HTML like `src="...gpt.js?x=1&amp;y=2"`, the parsed attribute value is normalized to `&`, so the replacement misses and the original `securepubads.g.doubleclick.net` URL is emitted unchanged.

**Fix**: either mutate the parsed DOM and serialize it, or replace the exact original attribute token span instead of matching on decoded `getAttribute()` text.

crates/js/lib/test/integrations/gpt/script_guard.test.ts
```diff
@@ -155,4 +155,154 @@ describe('GPT script guard', () => {
...
+    expect(writtenHtml).toContain(window.location.host);
+    expect(writtenHtml).toContain('/integrations/gpt/pagead/managed/js/gpt/current/pubads_impl.js');
+    expect(writtenHtml).not.toContain('securepubads.g.doubleclick.net');
+  });
+
+  // -----------------------------------------------------------------------
+  // Fail-closed behaviour
+  // -----------------------------------------------------------------------
+
+  it('fails closed when DOMParser is unavailable', () => {
```
prk-Jr: 🤔 **thinking** — The new DOMParser coverage still misses the two regression cases introduced by the new rewrite strategy.

A test for HTML-escaped URLs such as `?x=1&amp;y=2` would catch the proxy-bypass case, and a test asserting non-GPT HTML still passes through unchanged when `DOMParser` is unavailable would catch the breaking behavior from returning `''` too early.

prk-Jr (changes requested):
## Summary
This change replaces the GPT `document.write` regex rewrite path with a `DOMParser`-based implementation and adds broader coverage for edge-case markup. The parser-based approach is directionally better, but the new replacement strategy introduces a behavior regression for non-GPT writes under parser failure and a proxy-bypass case for HTML-escaped URLs.

## Blocking
### 🔧 wrench
- **Non-GPT `document.write` output gets dropped when `DOMParser` is unavailable or broken**: `rewriteHtmlString()` returns an empty string before it knows whether the HTML actually contains a GPT URL, so benign writes are erased in the very failure modes this PR adds support for.
- **HTML-escaped query strings can bypass the rewrite and leak back to the Google host**: the new flow reads `src` through `getAttribute()` and then runs `replaceAll()` on the original HTML string, which misses common serialized forms such as `&amp;` and can emit the original `securepubads.g.doubleclick.net` URL unchanged.

## Non-blocking
### 🤔 thinking
- **The new test section still misses the DOMParser-specific regression cases above**: adding coverage for escaped entities like `&amp;` and for non-GPT pass-through when `DOMParser` is unavailable would catch both regressions.

## CI Status
- fmt: PASS
- rust tests: PASS
- vitest: PASS
- CodeQL: PASS

PR #505, opened 2026-03-13
Strip spoofable forwarded headers at the Fastly edge entry point

### LINKED ISSUE #409: X-Forwarded-Host / Forwarded header spoofing enables URL rewrite hijack
## Summary

`RequestInfo` trusts `Forwarded`, `X-Forwarded-Host`, and `X-Forwarded-Proto` headers without validation against trusted proxies. An attacker setting `X-Forwarded-Host: evil.com` causes the HTML rewriter to replace all origin URLs with `evil.com`.

## Refs

- `crates/common/src/http_util.rs` lines 55-75 — `extract_request_host` trusts forwarded headers

## Recommendation

Strip or validate forwarded headers at the Fastly VCL layer, or validate against a configured allowlist.

## Context

ChristianPavilonis:
## Summary

Fixes #409 — `X-Forwarded-Host` / `Forwarded` header spoofing enables URL rewrite hijack.

- Strip `Forwarded`, `X-Forwarded-Host`, `X-Forwarded-Proto`, and `Fastly-SSL` headers at the top of `route_request()` before any routing or header inspection occurs
- Add `log::debug!` tracing when a spoofable header is actually stripped (aids incident response)
- Update `RequestInfo` and `publisher.rs` doc comments to reflect the new trust boundary

## Why it's safe to remove these headers

On Fastly Compute, the service **is** the edge — there is no legitimate upstream proxy setting forwarded headers. Any `X-Forwarded-Host` or `Forwarded` values present in the request were sent directly by the client and are therefore untrusted.

After stripping, the existing fallback logic in `RequestInfo::from_request` naturally lands on trustworthy signals:

- **Host**: `extract_request_host` falls through `Forwarded` (gone) → `X-Forwarded-Host` (gone) → **`Host` header** — which Fastly validates against the service configuration
- **Scheme**: `detect_request_scheme` hits **Fastly SDK TLS detection** first (priority 1, unchanged by this PR, not a header — it's SDK-level and unspoofable) → forwarded headers (gone) → default `http`

The forwarded header parsing code in `common` remains intact for potential non-edge deployment contexts, but the Fastly entry point now enforces the trust boundary.

### Impact on integrations

| Area | Impact |
|------|--------|
| **Prebid `site.domain` / `site.page`** | None — sourced from `settings.publisher.domain`, not `RequestInfo` |
| **Prebid ad-proxy URLs** | More correct — `Host` header is the actual browser-facing domain |
| **Prebid request signing** | Self-consistent — signing and `ext.trusted_server` both use the same source |
| **APS** | None — does not use `RequestInfo` at all |
| **Permutive SDK URL** | Same as ad-proxy — `Host` is the correct domain |

### Git blame context

The forwarded header logic was introduced in commit `563c281` ("Improve forwarded header parsing for host and scheme") with no security review (PR #176 had an empty description and no review comments). It was speculative support for a hypothetical chained-proxy scenario (`Client → Proxy A → Trusted Server`) that does not apply to the Fastly Compute deployment model.

## Test plan

- 2 new tests: `sanitize_removes_all_spoofable_headers`, `sanitize_then_request_info_falls_back_to_host`
- All 475 existing tests pass (including the existing `RequestInfo` forwarded-header parsing tests, which validate the `common` crate logic independently)
- `cargo fmt`, `cargo clippy` clean

### DIFF
--- a/crates/common/src/http_util.rs
+++ b/crates/common/src/http_util.rs
@@ -26,32 +26,66 @@ pub fn copy_custom_headers(from: &Request, to: &mut Request) {
     }
 }
 
+/// Headers that clients can spoof to hijack URL rewriting.
+///
+/// On Fastly Compute the service is the edge — there is no upstream proxy that
+/// legitimately sets these. Stripping them forces [`RequestInfo::from_request`]
+/// to fall back to the trustworthy `Host` header and Fastly SDK TLS detection.
+const SPOOFABLE_FORWARDED_HEADERS: &[&str] = &[
+    "forwarded",
+    "x-forwarded-host",
+    "x-forwarded-proto",
+    "fastly-ssl",
+];
+
+/// Strip forwarded headers that clients can spoof.
+///
+/// Call this at the edge entry point (before routing) to prevent
+/// `X-Forwarded-Host: evil.com` from hijacking all URL rewriting.
+/// See <https://github.com/IABTechLab/trusted-server/issues/409>.
+pub fn sanitize_forwarded_headers(req: &mut Request) {
+    for header in SPOOFABLE_FORWARDED_HEADERS {
+        if req.get_header(*header).is_some() {
+            log::debug!("Stripped spoofable header: {}", header);
+            req.remove_header(*header);
+        }
+    }
+}
+
 /// Extracted request information for host rewriting.
 ///
-/// This struct captures the effective host and scheme from an incoming request,
-/// accounting for proxy headers like `X-Forwarded-Host` and `X-Forwarded-Proto`.
+/// This struct captures the effective host and scheme from an incoming request.
+/// The parser checks forwarded headers (`Forwarded`, `X-Forwarded-Host`,
+/// `X-Forwarded-Proto`) as fallbacks, but on the Fastly edge
+/// [`sanitize_forwarded_headers`] strips those headers before this method is
+/// called, so the `Host` header and Fastly SDK TLS detection are the effective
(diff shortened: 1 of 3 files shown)


aram356 (commented):
## Summary

Clean, well-scoped security fix. Sanitization is placed at exactly the right point — top of `route_request`, before any header inspection or routing. No blockers.

## Non-blocking

### 🏕 camp site

- **Dead debug log fields** (`crates/common/src/publisher.rs:229-236`): After edge sanitization, `x-forwarded-host` and `x-forwarded-proto` are always `None` in this debug log. These format fields now add noise without information. Suggest simplifying to:
```rust
log::debug!(
    "Request info: host={}, scheme={} (Host: {:?})",
    request_host,
    request_scheme,
    req.get_header(header::HOST),
);
```

### 🌱 seedling

- **Defense-in-depth for `RequestInfo::from_request`**: `from_request` still checks forwarded headers first in its fallback chain. The docs correctly note sanitization happens at the edge, but if a second entry point is added or `from_request` is called without sanitizing, the old surface is exposed. A future option: accept a newtype wrapper (`&SanitizedRequest`) instead of `&Request` to make skipping sanitization a compile error. Not needed now — just a thought for hardening later.

## CI Status

- cargo fmt: PASS
- cargo test: PASS
- vitest: PASS
- format-typescript: PASS
- format-docs: PASS
- CodeQL: PASS

aram356 (approved):
👍 Looks good

PR #506, opened 2026-03-16
Use constant-time comparison for token and credential verification

### LINKED ISSUE #410: Non-constant-time token/password comparison enables timing attacks
## Summary

Signature verification (`tstoken`, clear URL signatures) and basic auth use standard `==` comparison, enabling timing side-channel attacks.

## Refs

- `crates/common/src/proxy.rs` lines 1054-1058 — `expected != sig`
- `crates/common/src/http_util.rs` lines 289-291 — `sign_clear_url(...) == token`
- `crates/common/src/auth.rs` lines 17-18 — `password == handler.password`

## Recommendation

Use `subtle::ConstantTimeEq` (already in dependency tree via crypto crates).

## Context

prk-Jr:
## Summary

- Replace `==` comparisons with `subtle::ConstantTimeEq` in all three secret-verification sites (tstoken, clear-URL signature, Basic Auth), eliminating timing side-channel attacks
- Fix a secondary short-circuit oracle in Basic Auth: `&&` allowed an attacker to distinguish wrong-username from wrong-password via timing — replaced with bitwise `&` on `subtle::Choice` so both fields are always evaluated
- Add `log::warn!` on auth failure (path only, no credentials) and five new targeted tests covering each fix

## Changes

| File | Change |
| ---- | ------ |
| `Cargo.toml` | Add `subtle = "2"` to workspace dependencies |
| `crates/common/Cargo.toml` | Add `subtle` as direct dependency |
| `crates/common/src/auth.rs` | CT comparison with `&` instead of `&&`; `log::warn!` on failure; 2 new tests |
| `crates/common/src/http_util.rs` | CT comparison in `verify_clear_url_signature`; 2 new tests |
| `crates/common/src/proxy.rs` | CT comparison for tstoken in `reconstruct_and_validate_signed_target`; 1 new test |

## Closes

Closes #410

## Test plan

> JS tests are failing on `main` due to a pre-existing ESM/CJS incompatibility in `html-encoding-sniffer` node_modules — unrelated to this PR.

## Checklist

- [x] New code has tests

### DIFF
--- a/crates/trusted-server-core/src/request_signing/endpoints.rs
+++ b/crates/trusted-server-core/src/request_signing/endpoints.rs
@@ -48,7 +48,7 @@ pub fn handle_trusted_server_discovery(
 
     Ok(Response::from_status(200)
         .with_content_type(fastly::mime::APPLICATION_JSON)
-        .with_body_text_plain(&json))
+        .with_body(json))
 }
 
 #[derive(Debug, Deserialize, Serialize)]
@@ -118,7 +118,7 @@ pub fn handle_verify_signature(
 
     Ok(Response::from_status(200)
         .with_content_type(fastly::mime::APPLICATION_JSON)
-        .with_body_text_plain(&response_json))
+        .with_body(response_json))
 }
 
 #[derive(Debug, Deserialize, Serialize)]
@@ -199,7 +199,7 @@ pub fn handle_rotate_key(
 
             Ok(Response::from_status(200)
                 .with_content_type(fastly::mime::APPLICATION_JSON)
-                .with_body_text_plain(&response_json))
+                .with_body(response_json))
         }
         Err(e) => {
             let response = RotateKeyResponse {
@@ -220,7 +220,7 @@ pub fn handle_rotate_key(
 
             Ok(Response::from_status(500)
                 .with_content_type(fastly::mime::APPLICATION_JSON)
-                .with_body_text_plain(&response_json))
+                .with_body(response_json))
         }
     }
 }
@@ -305,7 +305,7 @@ pub fn handle_deactivate_key(
 
             Ok(Response::from_status(200)
                 .with_content_type(fastly::mime::APPLICATION_JSON)
-                .with_body_text_plain(&response_json))
+                .with_body(response_json))
         }
         Err(e) => {
             let response = DeactivateKeyResponse {
@@ -329,7 +329,7 @@ pub fn handle_deactivate_key(
 
             Ok(Response::from_status(500)
                 .with_content_type(fastly::mime::APPLICATION_JSON)
-                .with_body_text_plain(&response_json))
+                .with_body(response_json))
         }
     }
 }
@@ -366,6 +366,11 @@ mod tests {
         let mut resp =
             handle_verify_signature(&settings, req).expect("should handle verification request");
         assert_eq!(resp.get_status(), StatusCode::OK);
+        assert_eq!(
+            resp.get_content_type(),
+            Some(fastly::mime::APPLICATION_JSON),
+            "should return application/json content type"
+        );
 
         // Parse response
         let resp_body = resp.take_body_str();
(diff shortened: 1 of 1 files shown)


aram356 (commented):
## Summary

Deep re-review after the `Unauthorized` → `Forbidden` rename was addressed. This PR is well-executed: constant-time comparisons via `subtle::ConstantTimeEq` at all three secret-verification sites (tstoken, clear-URL signature, Basic Auth), bitwise `&` on `subtle::Choice` to eliminate the short-circuit oracle in Basic Auth, `Redacted<T>` `PartialEq` removal to prevent regressions, and good test coverage for each fix.

No blocking findings remain.

## Non-blocking

### 🤔 thinking

- **`ct_str_eq` is `pub` but has a length-leak invariant**: The `&&` short-circuits on length mismatch, which is safe for current call-sites (fixed-length base64url SHA-256 digests, always 43 bytes). But the function is `pub` and a future caller with variable-length secrets could misuse it. Consider `pub(crate)` to limit exposure or keep the thorough `# Security` doc section (which is already excellent). (http_util.rs:347)

- **Status code change from 502 → 403 for invalid tstoken**: Previously invalid tstoken returned `TrustedServerError::Proxy` → 502 Bad Gateway. Now returns `Forbidden` → 403 Forbidden. The new code is semantically correct (a failed signature check is a client auth error, not an upstream failure). Just flagging for awareness in case any monitoring or client retry logic depends on the 502. (proxy.rs:1080)

### 🌱 seedling

- **Consider HMAC-SHA-256 for URL signing**: `sign_clear_url` uses `SHA-256(prefix || secret || url)` — a keyed hash with domain separation. While not practically exploitable here (the prefix prevents length-extension on the secret, and the output is compared not forwarded), HMAC-SHA-256 is the standard construct for keyed message authentication and would be a stronger primitive. Worth a follow-up issue, not this PR.

## CI Status

- cargo fmt: PASS
- cargo clippy: PASS
- cargo test: PASS
- vitest: PASS
- integration tests: PASS
- CodeQL: PASS

aram356 (approved):
## Summary

Solid security hardening — replaces timing-vulnerable `==` comparisons with `subtle::ConstantTimeEq` across all three secret-verification sites (tstoken, clear-URL signature, Basic Auth). The bitwise `&` fix for the username/password oracle and SHA-256 length normalization are both well-implemented.

## Non-blocking

### 👍 praise

- **Excellent `ct_str_eq` documentation** (`http_util.rs`) — clearly explains when the length leak is safe and when this function should NOT be used. Exactly the kind of security doc that prevents future misuse.
- **SHA-256 length normalization in auth** — hashing credentials before comparison to normalize to fixed-size digests is the right approach, and the comment explaining why is valuable.
- **Bitwise `&` instead of `&&`** for credential matching — eliminates the username-oracle timing channel cleanly.
- **Removing `PartialEq` from `Redacted<T>`** — defense-in-depth: makes it impossible to accidentally `==`-compare `Redacted` values at compile time.
- **WASM-caveat comment** — honestly acknowledges the inherent limitation of constant-time guarantees on WASM/JIT targets.

### 🤔 thinking

- **`Forbidden` vs `Proxy` for invalid tstoken** — semantically correct (403 vs 502), but this is a behavioral change for clients. Worth confirming no monitoring/alerting keys off 502 rates for token validation failures.
- **`log::warn!` includes request path** — good that credentials aren't logged. If protected paths contain sensitive info, logging the full path reveals which endpoints are under attack. Probably fine for operational visibility.

### ⛏ nitpick

- PR body checklist says "Uses `tracing` macros" but the project uses `log` macros per CLAUDE.md. The actual code correctly uses `log::warn!` — just a copy-paste inaccuracy in the description.

## CI Status

- fmt: PASS
- clippy: PASS
- cargo test: PASS
- vitest: PASS
- CodeQL: PASS
- integration tests: PASS

PR #507, opened 2026-03-16
Add HttpOnly flag to synthetic ID cookie

### LINKED ISSUE #411: Synthetic ID cookie missing HttpOnly flag
## Summary

The `synthetic_id` cookie is set with `Secure; SameSite=Lax` but no `HttpOnly`. Any XSS on the publisher's page can exfiltrate this tracking identifier via `document.cookie`.

## Refs

- `crates/common/src/cookies.rs` lines 67-72 — `create_synthetic_cookie` format string

## Recommendation

Add `HttpOnly` if client-side JS doesn't need to read this cookie directly (it already gets the value via the `x-synthetic-id` header).

## Context

prk-Jr:
## Summary

- The `synthetic_id` cookie was missing `HttpOnly`, allowing XSS on a publisher page to exfiltrate the tracking identifier via `document.cookie`.
- `HttpOnly` is safe to add because no client-side JS reads this cookie — integrations receive the synthetic ID via the `x-synthetic-id` response header instead.
- Adds `debug_assert!` guards against cookie metacharacter injection in both the `synthetic_id` value and `cookie_domain`, and documents the rationale for each security attribute in the function's doc comment.

## Changes

| File | Change |
| ---- | ------ |
| `crates/common/src/cookies.rs` | Add `HttpOnly` to cookie format string; add `debug_assert!` metacharacter guards; update doc comment to enumerate all security attributes with rationale |

## Closes

Closes #411

## Test plan

- [ ] JS tests: `cd crates/js/lib && npx vitest run` — pre-existing ESM/CJS failure on `main`, unrelated to this change

## Checklist

- [x] New code has tests

### DIFF
--- a/docs/internal/ssc-prd.md
+++ b/docs/internal/ssc-prd.md
@@ -1,0 +1,976 @@
+# Product Requirements: Edge Cookie (EC)
+
+**Status:** Draft
+**Author:** Trusted Server Product
+**Last updated:** 2026-03-12
+
+---
+
+## Table of Contents
+
+1. [Overview](#1-overview)
+2. [Problem Statement](#2-problem-statement)
+3. [Goals and Non-Goals](#3-goals-and-non-goals)
+4. [Target Customers](#4-target-customers)
+5. [TS Lite Deployment Mode](#5-ts-lite-deployment-mode)
+6. [EC Identity and Cookie Structure](#6-ec-identity-and-cookie-structure)
+7. [Consent Lifecycle](#7-consent-lifecycle)
+8. [KV Store Identity Graph](#8-kv-store-identity-graph)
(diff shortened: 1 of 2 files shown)


crates/common/src/cookies.rs
```diff
@@ -61,13 +61,38 @@ pub fn handle_request_cookies(
 
 /// Creates a synthetic ID cookie string.
```
ChristianPavilonis: 🟢 **P3 — Missing `# Examples` doc section**

Per CLAUDE.md line 198: "Add `# Examples` sections for public API functions." The doc comment is already excellent — thorough security rationale, `# Panics` section — but lacks `# Examples`. Minor, and other functions in this file don't have them either, so this is a nit rather than a blocker.

PR #508, opened 2026-03-16
Validate synthetic ID format on inbound header and cookie values

### LINKED ISSUE #412: No synthetic ID format validation on inbound values
## Summary

The synthetic ID from cookies or headers is accepted without format validation. An attacker can inject arbitrary strings (very long, special characters, newlines) which are then set as response headers, cookies, and forwarded to third-party APIs.

## Refs

- `crates/common/src/synthetic.rs` lines 129-153 — `get_synthetic_id` accepts any string
- `crates/common/src/publisher.rs` line 336 — set as response header
- `crates/common/src/proxy.rs` line 442 — forwarded as query parameter

## Recommendation

Validate against the expected format (64 hex + dot + 6 alphanumeric) in production code, not just tests.

## Context

prk-Jr:
## Summary

- Inbound synthetic IDs from the `x-synthetic-id` header and `synthetic_id` cookie were accepted without validation, allowing injection of arbitrary strings into response headers, cookies, and third-party API calls
- Adds a private `is_valid_synthetic_id()` validator (64 lowercase hex + `.` + 6 alphanumeric) with an O(1) length check first to bound all downstream work; invalid values are silently discarded and a fresh ID is generated in their place
- Hardens logging by ensuring the raw ID value never appears in warn logs and demoting synthetic ID values from `INFO` to `DEBUG` to avoid recording pseudonymous identifiers in production log pipelines

## Changes

| File | Change |
| ---- | ------ |
| `crates/common/src/synthetic.rs` | Add `is_valid_synthetic_id()` production validator; validate in `get_synthetic_id()`; add `debug_assert!` in generator; demote ID values to `debug!`; new rejection + fallthrough tests |
| `crates/common/src/test_support.rs` | Add shared `VALID_SYNTHETIC_ID` constant |
| `crates/common/src/proxy.rs` | Update test fixture to use `VALID_SYNTHETIC_ID` |
| `crates/common/src/integrations/registry.rs` | Update test fixture to use valid-format cookie value |
| `docs/guide/synthetic-ids.md` | Document inbound validation behaviour and clarify HMAC determinism |
| `CHANGELOG.md` | Add `### Security` entry under `[Unreleased]` |

## Closes

Closes #412

## Test plan

## Checklist

- [x] New code has tests


aram356 (approved):
## Summary

Clean, well-designed security fix that validates inbound synthetic IDs from the `x-synthetic-id` header and `synthetic_id` cookie. The `is_valid_synthetic_id()` validator is correct, defense-in-depth layering across `synthetic.rs`, `cookies.rs`, and `publisher.rs` is sound, and test coverage is thorough.

## Non-blocking

### 📌 out of scope

- **Synthetic ID still logged at INFO in proxy.rs redirect handler**: The redirect handler at `proxy.rs:717` still logs the validated synthetic ID value at `log::info!`. Low risk since the value is already validated, but inconsistent with the PR's stated goal of demoting ID values to DEBUG. Worth a follow-up.

### 🤔 thinking

- **`split_once('.')` vs old `split('.')` is subtly correct**: The new approach naturally handles multi-dot values because the total length constraint (71 = 64 + 1 + 6) means extra dots fail the alphanumeric suffix check. The `dot_in_suffix` test covers this. Sound design choice.

### 👍 praise

- **Excellent defense-in-depth in publisher.rs revocation path** (`publisher.rs:382-399`): Always expires the cookie regardless of validity, only passes validated IDs to `delete_consent_from_kv`, logs only the length of invalid values. Prevents malicious cookie values being used as KV keys.
- **Synthetic ID values removed from log output** (`synthetic.rs`): Old code logged actual ID values at trace level; new code logs only structural information. Eliminates log-exfiltration of pseudonymous identifiers.
- **O(1) length guard before character scanning** (`synthetic.rs:42`): Constant-time length check before iteration is a nice DoS mitigation for oversized inputs.
- **Comprehensive test coverage**: Covers invalid header/cookie rejection, header precedence, fallthrough behavior, generation on invalid input, and edge cases (uppercase hex, oversized input, empty string, dot in suffix, extra segments, missing suffix, non-hex, non-alphanumeric suffix).

### ⛏ nitpick

- **CHANGELOG says "64-hex-hmac" but validator requires lowercase hex** (`CHANGELOG.md:12`): Saying "64-lowercase-hex-hmac" would be more precise. Minor since the code is authoritative.

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS
- js tests: PASS

PR #510, opened 2026-03-16
Add domain allowlist to block SSRF via first-party proxy redirects

### LINKED ISSUE #414: SSRF via first-party proxy -- no target domain allowlist
## Summary

The `/first-party/proxy` endpoint proxies to arbitrary URLs (protected only by `tstoken` signature). `proxy_with_redirects` follows up to 4 redirects with no domain or IP range restriction, allowing SSRF to internal services if a signed URL redirects.

## Refs

- `crates/common/src/proxy.rs` lines 600-621 — `handle_first_party_proxy`
- `crates/common/src/proxy.rs` lines 463-582 — `proxy_with_redirects` follows redirects

## Recommendation

Validate redirect targets against an allowlist or block private IP ranges.

## Context

prk-Jr:
## Summary

- The first-party proxy followed redirects with no restriction on destination hosts. A signed URL pointing to an attacker-controlled origin could redirect to an internal service, enabling SSRF.
- Adds `proxy.allowed_domains` — an opt-in list of permitted redirect-target hostnames supporting exact match (`"example.com"`) and subdomain wildcard (`"*.example.com"`) with dot-boundary enforcement to prevent `evil-example.com` bypass.
- When the list is empty (the default), all redirect destinations are permitted — preserving backward compatibility for existing deployments.

## Changes

| File | Change |
| ---- | ------ |
| `crates/common/src/proxy.rs` | Add `is_host_allowed` helper; enforce allowlist on each redirect hop in `proxy_with_redirects`; 13 unit tests |
| `crates/common/src/settings.rs` | Add `allowed_domains: Vec<String>` field to `Proxy` struct with serde default |
| `trusted-server.toml` | Uncomment `[proxy]` section; add `allowed_domains` key with commented vendor examples |
| `.env.dev` | Document `TRUSTED_SERVER__PROXY__ALLOWED_DOMAINS` env var override |

## Closes

Closes #414

## Test plan

> Note: JS tests fail with `ERR_REQUIRE_ESM` due to a pre-existing `jsdom@28` / `html-encoding-sniffer` incompatibility on `main` — unrelated to this PR.

## Checklist

- [x] New code has tests

### DIFF
--- a/crates/trusted-server-core/src/proxy.rs
+++ b/crates/trusted-server-core/src/proxy.rs
@@ -1403,7 +1403,8 @@ mod tests {
                 sig
             ),
         );
-        req.set_header(crate::constants::HEADER_X_SYNTHETIC_ID, "synthetic-123");
+        let valid_synthetic_id = crate::test_support::tests::VALID_SYNTHETIC_ID;
+        req.set_header(crate::constants::HEADER_X_SYNTHETIC_ID, valid_synthetic_id);
 
         let resp = handle_first_party_click(&settings, req)
             .await
@@ -1421,7 +1422,7 @@ mod tests {
         assert_eq!(pairs.remove("foo").as_deref(), Some("1"));
         assert_eq!(
             pairs.remove("synthetic_id").as_deref(),
-            Some("synthetic-123")
+            Some(valid_synthetic_id)
         );
         assert!(pairs.is_empty());
     }
--- a/crates/trusted-server-core/src/settings.rs
+++ b/crates/trusted-server-core/src/settings.rs
@@ -31,18 +31,6 @@ pub struct Publisher {
 }
 
 impl Publisher {
-    /// Known placeholder values that must not be used in production.
-    pub const PROXY_SECRET_PLACEHOLDERS: &[&str] = &["change-me-proxy-secret"];
-
-    /// Returns `true` if `proxy_secret` matches a known placeholder value
-    /// (case-insensitive).
-    #[must_use]
-    pub fn is_placeholder_proxy_secret(proxy_secret: &str) -> bool {
-        Self::PROXY_SECRET_PLACEHOLDERS
-            .iter()
-            .any(|p| p.eq_ignore_ascii_case(proxy_secret))
-    }
-
     /// Extracts the host (including port if present) from the `origin_url`.
     ///
     /// # Examples
@@ -215,24 +203,7 @@ pub struct Synthetic {
 }
 
 impl Synthetic {
-    /// Known placeholder values that must not be used in production.
-    pub const SECRET_KEY_PLACEHOLDERS: &[&str] = &["secret-key", "secret_key", "trusted-server"];
-
-    /// Returns `true` if `secret_key` matches a known placeholder value
-    /// (case-insensitive).
-    #[must_use]
-    pub fn is_placeholder_secret_key(secret_key: &str) -> bool {
-        Self::SECRET_KEY_PLACEHOLDERS
-            .iter()
-            .any(|p| p.eq_ignore_ascii_case(secret_key))
-    }
-
     /// Validates that the secret key is not empty.
-    ///
-    /// Placeholder detection is intentionally **not** performed here because
-    /// this validator runs at build time (via `from_toml_and_env`) when the
-    /// config legitimately contains placeholder values. Placeholder rejection
-    /// happens at runtime via [`Settings::reject_placeholder_secrets`].
     ///
     /// # Errors
     ///
@@ -531,33 +502,6 @@ impl Settings {
         }
 
         Ok(None)
-    }
-
-    /// Checks all secret fields for known placeholder values and returns an
-    /// error listing every offending field. This centralises the placeholder
-    /// policy so callers don't need to know which fields are secrets.
-    ///
-    /// # Errors
-    ///
-    /// Returns [`TrustedServerError::InsecureDefault`] when one or more secret
-    /// fields still contain a placeholder value.
-    pub fn reject_placeholder_secrets(&self) -> Result<(), Report<TrustedServerError>> {
-        let mut insecure_fields: Vec<&str> = Vec::new();
-
-        if Synthetic::is_placeholder_secret_key(self.synthetic.secret_key.expose()) {
-            insecure_fields.push("synthetic.secret_key");
-        }
-        if Publisher::is_placeholder_proxy_secret(self.publisher.proxy_secret.expose()) {
-            insecure_fields.push("publisher.proxy_secret");
-        }
-
-        if insecure_fields.is_empty() {
-            Ok(())
-        } else {
-            Err(Report::new(TrustedServerError::InsecureDefault {
-                field: insecure_fields.join(", "),
-            }))
-        }
     }
 
     /// Known admin endpoint paths that must be covered by a handler.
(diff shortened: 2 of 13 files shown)


docs/guide/first-party-proxy.md
```diff
@@ -425,6 +425,47 @@ proxy_secret = "your-secure-random-secret"
...
+```toml
+[proxy]
+allowed_domains = [
+  "tracker.com",           # Exact match
+  "*.adserver.com",        # Wildcard: adserver.com and all subdomains
+  "*.trusted-cdn.net",
+]
+```
+
+**Semantics**: When a proxied request receives an HTTP redirect (301/302/303/307/308), the redirect target host is checked against `allowed_domains`. If the host does not match any pattern the redirect is blocked and a 403 error is returned.
```
aram356: 🔧 **wrench** — Status code mismatch: the code returns 502 (Bad Gateway) via `TrustedServerError::Proxy`, not 403 as stated here.

**Suggested fix:**
```
...the redirect is blocked and a 502 error is returned.
```

Alternatively, if 403 is the desired behavior, add a dedicated error variant that returns 403 for allowlist violations.

crates/trusted-server-core/src/settings.rs
```diff
@@ -323,10 +334,28 @@ impl Default for Proxy {
...
 }
 
+impl Proxy {
+    /// Normalizes [`allowed_domains`] in place.
+    ///
+    /// Each entry is trimmed of surrounding whitespace and lowercased.
+    /// Empty entries (including those that were only whitespace) are removed.
+    /// This prevents blank patterns from making the list non-empty while
+    /// still never matching any host, which would silently block all redirects.
+    fn normalize(&mut self) {
```
aram356: 🤔 **thinking** — Consider adding a `log::warn!` when `allowed_domains` is empty after normalization. Operators may not realize their deployment is in open mode (all redirect destinations permitted). A startup warning would improve operational visibility.

```rust
fn normalize(&mut self) {
    self.allowed_domains = self
        .allowed_domains
        .iter()
        .map(|s| s.trim().to_ascii_lowercase())
        .filter(|s| !s.is_empty())
        .collect();
    if self.allowed_domains.is_empty() {
        log::warn!("proxy.allowed_domains is empty: all redirect destinations are permitted");
    }
}
```

crates/trusted-server-core/src/proxy.rs
```diff
@@ -479,6 +479,35 @@ fn append_synthetic_id(req: &Request, target_url_parsed: &mut url::Url) {
...
+}
+
+/// Returns `true` if `host` is permitted by `pattern`.
+///
+/// - `"example.com"` matches exactly `example.com`.
+/// - `"*.example.com"` matches `example.com` and any subdomain at any depth.
+///
+/// Comparison is case-insensitive. The wildcard check requires a dot boundary,
+/// so `"*.example.com"` does **not** match `"evil-example.com"`.
+fn is_host_allowed(host: &str, pattern: &str) -> bool {
```
aram356: 👍 **praise** — The `strip_suffix` + `ends_with('.')` approach for dot-boundary enforcement is clean and correct. It properly prevents `evil-example.com` from matching `*.example.com`, which is the most common allowlist bypass vector. Well-tested too.

crates/trusted-server-core/src/proxy.rs
```diff
@@ -479,6 +479,35 @@ fn append_synthetic_id(req: &Request, target_url_parsed: &mut url::Url) {
     }
 }
 
+/// Returns `true` when a redirect to `host` should be followed.
+///
+/// When `allowed_domains` is empty every host is permitted (open mode).
+/// When non-empty the host must match at least one pattern via [`is_host_allowed`].
+fn redirect_is_permitted(allowed_domains: &[String], host: &str) -> bool {
```
aram356: 🌱 **seedling** — Per the project's newtype convention (CLAUDE.md), `allowed_domains: Vec<String>` could become an `AllowedDomains` newtype that encapsulates normalization and matching logic (`is_host_allowed`, `redirect_is_permitted`). Not needed for this PR, but would make the API more cohesive as the feature evolves.

crates/trusted-server-core/src/proxy.rs
```diff
@@ -1834,4 +1875,177 @@ mod tests {
             body
         );
     }
+
+    // --- is_host_allowed ---
```
aram356: 🌱 **seedling** — The 18 allowlist tests are thorough for the documented cases. A few edge cases that could be added in a follow-up:

- **IP literal redirect target** (e.g., `is_host_allowed("169.254.169.254", "*.example.com")`) — the logic correctly blocks it, but an explicit test documents the guarantee
- **Empty Location header** in a redirect response

These are low risk since the current logic handles them correctly, but explicit tests would prevent regressions.

crates/trusted-server-core/src/proxy.rs
```diff
@@ -479,6 +479,35 @@ fn append_synthetic_id(req: &Request, target_url_parsed: &mut url::Url) {
...
+    allowed_domains.is_empty() || allowed_domains.iter().any(|p| is_host_allowed(host, p))
+}
+
+/// Returns `true` if `host` is permitted by `pattern`.
+///
+/// - `"example.com"` matches exactly `example.com`.
+/// - `"*.example.com"` matches `example.com` and any subdomain at any depth.
+///
+/// Comparison is case-insensitive. The wildcard check requires a dot boundary,
+/// so `"*.example.com"` does **not** match `"evil-example.com"`.
```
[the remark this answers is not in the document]
prk-Jr: Fixed in `normalize()`: bare `"*"` entries are now removed with a `log::warn!` rather than silently falling through to always-false exact match. Added two settings tests: one for mixed list (`["*", "tracker.com"]` → `["tracker.com"]`) and one for solo `["*"]` → empty (open mode).

docs/guide/first-party-proxy.md
```diff
@@ -425,6 +425,47 @@ proxy_secret = "your-secure-random-secret"
...
+
+**Wildcard matching**:
+
+| Pattern         | Matches                                             | Does not match     |
+| --------------- | --------------------------------------------------- | ------------------ |
+| `tracker.com`   | `tracker.com`                                       | `sub.tracker.com`  |
+| `*.tracker.com` | `tracker.com`, `sub.tracker.com`, `a.b.tracker.com` | `evil-tracker.com` |
+
+- The `*` prefix matches the base domain and any subdomain at any depth.
+- Matching is case-insensitive; entries are normalized to lowercase on startup.
```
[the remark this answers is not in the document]
prk-Jr: Updated. Wording is now: "This default exists solely for development convenience and **must be overridden in production**."

crates/trusted-server-core/src/proxy.rs
```diff
@@ -584,6 +623,17 @@ async fn proxy_with_redirects(
...
+                next_host
+            );
+            return Err(Report::new(TrustedServerError::AllowlistViolation {
+                host: next_host.to_string(),
+            }));
+        }
+
         log::info!(
             "following redirect {} => {} (status {})",
             current_url,
```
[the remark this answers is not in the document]
prk-Jr: Fixed. A missing-host check was added before the allowlist check for redirect hops, returning `TrustedServerError::Proxy { message: "missing host in redirect location" }` instead of falling through to `AllowlistViolation { host: "" }`.

aram356 (approved):
## Summary

Solid SSRF mitigation for the first-party proxy redirect chain. The core matching logic (`is_host_allowed`) is correct, well-tested (13+ unit tests covering boundary bypass, IP literals, case sensitivity, empty lists), and cleanly scoped to only the first-party proxy path — integration proxies correctly default to open mode since they target hardcoded URLs.

CI is all green. No blocking issues found.

## Non-blocking

### 🤔 thinking

- **Wildcard semantics differ from `Rewrite::is_excluded`** — `*.example.com` matches the apex domain in the new `is_host_allowed` but not in `Rewrite::is_excluded`. Both behaviors are defensible for their use cases, but the inconsistency could confuse operators configuring both. Worth a note in docs if it comes up.

### 🌱 seedling

- **Consider deny-by-default in a future major version** — empty list = open mode is correct for backward compat, but a future breaking release could flip to require an explicit allowlist for production safety.

### 📌 out of scope

- **`handle_first_party_click` doesn't check allowlist** — correct for SSRF (no server-side request), but the click endpoint could still redirect a browser to an attacker-controlled URL via a signed URL. This is an open-redirect concern, not SSRF, and outside this PR's scope.

## CI Status

- cargo fmt: PASS
- cargo clippy: PASS
- cargo test: PASS
- vitest: PASS
- integration tests: PASS
- CodeQL: PASS

PR #511, opened 2026-03-17
Add Edge Cookie (EC) product requirements document

aram356:
## Summary

- Adds the internal PRD for Server-Side Cookie (SSC), a consent-aware first-party identity system that replaces SyntheticID
- Covers SSC identity generation (IP + salt only), consent lifecycle (TCF/GPP enforcement with real-time withdrawal), KV Store identity graph, pixel sync endpoint (`/sync`), S2S batch API (`/api/v1/sync`), and bidstream decoration (`/identify` + `/auction`)
- Defines TS Lite deployment mode — a runtime configuration that exposes only SSC routes, enabling SSPs, DSPs, and identity providers to adopt SSC without the full Trusted Server feature surface

## Key sections

- **Problem statement**: SyntheticID signal degradation (UA reduction, IPv6 rotation), no consent enforcement, adoption blocked by full TS requirement
- **SSC identity**: HMAC-SHA256 of IP (IPv6 /64 prefix) + publisher salt, `ts-ssc` cookie at apex domain
- **Consent lifecycle**: TCF/GPP enforcement with region-based rules, real-time cookie + KV deletion on withdrawal
- **KV identity graph**: SSC hash → partner ID map with atomic read-modify-write via generation markers
- **Pixel sync** (`GET /sync`): Browser redirect-based partner ID sync with rate limiting and domain allowlist
- **S2S batch API** (`POST /api/v1/sync`): Authenticated bulk ID mapping push for DSPs and partners
- **Bidstream decoration**: Header-only mode (`/identify`) and full auction mode (`/auction`) with OpenRTB 2.6 `user.eids`
- **Configuration**: New `[ssc]` and `[features]` TOML sections, partner registry in KV store


docs/internal/ssc-prd.md
```diff
@@ -0,0 +1,634 @@
...
+| Parameter | Required | Description                                                                                       |
+| --------- | -------- | ------------------------------------------------------------------------------------------------- |
+| `partner` | Yes      | Partner ID, must match a registered partner in `partner_store` KV                                 |
+| `uid`     | Yes      | Partner's user ID for this user                                                                   |
+| `return`  | Yes      | Callback URL to redirect to after sync (must match partner's `allowed_return_domains`)            |
+| `consent` | No       | TCF or GPP string from the partner's context, used if no consent signal is present on the request |
+
+### 9.4 Flow
+
+1. Read the `ts-ssc` cookie. If absent, redirect to `return` URL immediately without writing to KV. Do not create a new SSC during a sync — a sync redirect is not an organic user visit and must not be used to bootstrap identity.
```
aram356: **Clarify: What does the user experience when sync is a no-op?**

When there is no `ts-ssc` cookie, the user is redirected to the `return` URL "immediately without writing to KV." But step 6 appends `ts_synced=1` on success. What is appended (or not) on a no-op? Should it be `ts_synced=0` so the partner knows the sync did not take effect? Without this, the partner has no way to distinguish success from silent failure.

docs/internal/ssc-prd.md
```diff
@@ -0,0 +1,634 @@
...
+  "accepted": 998,
+  "rejected": 2,
+  "errors": [
+    { "index": 45, "reason": "ssc_hash_not_found" },
+    { "index": 72, "reason": "consent_withdrawn" }
+  ]
+}
+```
+
+HTTP status `207 Multi-Status` when any mappings are rejected; `200 OK` when all are accepted.
```
aram356: **Clarify: What happens when auth is valid but all mappings fail?**

The spec says `207 Multi-Status` when "any mappings are rejected" and `200 OK` when "all are accepted." But what if all 1000 mappings are rejected (e.g., all `ssc_hash_not_found`)? Is that `207` with `accepted: 0`, or a different status? Partners need to know what "complete failure of a valid request" looks like.

docs/internal/ssc-prd.md
```diff
@@ -0,0 +1,634 @@
...
+
+Trusted Server exposes two modes for injecting SSC identity into the bidstream. Publishers choose the mode that fits their existing ad stack.
+
+### 11.2 Mode A: Header decoration (`/identify`)
+
+For publishers whose existing ad server handles auction calls, Trusted Server provides an identification-only endpoint that returns the SSC value and resolved identity signals as response headers. The publisher's ad server reads these headers and injects them into its own OpenRTB bid requests.
+
+**Endpoint:** `GET /identify`
+
+**Response:** `204 No Content` with the following headers:
```
aram356: **Clarify: Who calls `/identify` and when?**

This section says "the publisher's ad server reads these headers" — but the call context is unclear:
- Is this called per-pageview? Per-auction?
- Is it a sub-request from the same edge (e.g., Fastly VCL calling TS), or a separate HTTP call from the publisher's origin server?
- Does `/identify` read the user's cookies directly (same-origin), or does the caller pass identity via headers?

This affects whether the endpoint needs its own cookie/consent handling or relies on the caller forwarding that context.

docs/internal/ssc-prd.md
```diff
@@ -0,0 +1,634 @@
...
+Consent enforcement is a core requirement of SSC. The system must not create or maintain an SSC for users who have not given consent, and must actively revoke the SSC when consent is withdrawn.
+
+### 7.1 Consent signal sources and precedence
+
+When evaluating consent on a given request, Trusted Server checks signals in the following order. The first signal found wins:
+
+1. **`X-consent-advertising` request header** — set by the Didomi integration (or another CMP proxy) in a prior server-side decode. This is the freshest signal and takes precedence over browser-stored values.
+2. **`euconsent-v2` cookie** — the TCF v2 consent string stored by the publisher's CMP.
+3. **`gpp` cookie** — the IAB Global Privacy Platform string for US state-level consent.
+4. **Default: no consent** — if no signal is found, do not create the SSC (fail safe).
```
aram356: **Clarify: Conflict between "no signal = no consent" and "rest of world = create on first visit."**

Section 7.1 step 4 says: "Default: no consent — if no signal is found, do not create the SSC."

Section 7.2 table says: "Rest of world — None required — Create SSC on first visit."

These contradict each other for a user in (say) Brazil with no TCF or GPP signal. The precedence between 7.1 step 4 and the 7.2 region table needs to be explicit. Suggestion: 7.1 describes *how signals are read*; 7.2 describes *which signals are required per region*. If the region requires no signal, 7.1 step 4 does not apply. But this should be stated.

docs/internal/ssc-prd.md
```diff
@@ -0,0 +1,634 @@
...
+- Handlebars template (input is now fixed, not configurable)
+
+**Output format (unchanged from SyntheticID):**
+
+```
+{64-character hex HMAC-SHA256}.{6-character random alphanumeric suffix}
+```
+
+The 64-character prefix is the stable, deterministic portion used as the KV store key. The 6-character suffix is random, regenerated each time a fresh SSC is created. Once an SSC is set in a cookie, the full value (prefix + suffix) is preserved on subsequent requests.
+
```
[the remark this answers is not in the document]
jevansnyc: Ok. Should have been answered above then.

PR #531, opened 2026-03-19
Remove dead ad-proxy URL rewriting from Prebid parse_response

### LINKED ISSUE #417: Prebid response rewriting trusts host/scheme from upstream response body
## Summary

`request_host` and `request_scheme` for URL rewriting are read from the Prebid server's response JSON (`ext.trusted_server.request_host`), not from the local request context. A compromised or misconfigured bidder can inject arbitrary host/scheme values used in URL rewrites.

## Refs

- `crates/common/src/integrations/prebid.rs` line 904 — reads `request_host` from response JSON
- `crates/common/src/integrations/prebid.rs` line 911 — reads `request_scheme` from response JSON
- `crates/common/src/integrations/prebid.rs` line 922 — passes them to `transform_prebid_response`

## Recommendation

Use the local request's host/scheme from `RequestInfo` instead of the bidder's response body.

## Context

ChristianPavilonis:
## Summary

Fixes a security issue (#417) where `request_host` and `request_scheme` for URL rewriting were read from the upstream Prebid Server response JSON, allowing a compromised or misconfigured bidder to inject arbitrary values into ad markup URL rewrites.

## What we discovered

During review we found that the entire `transform_prebid_response` code path is **dead code**:

1. **Route handler removed** — `transform_prebid_response` generates `/ad-proxy/` URLs (e.g. `https://host/ad-proxy/adsrvr/...`), but the `/ad-proxy/` route handler was removed in 25084ba ("NextJS with Prebid Integration"). These URLs fall through to the publisher origin fallback, which can't handle them.

2. **Downstream rewriter already handles this** — `creative::rewrite_creative_html` (called from `formats.rs` during `convert_to_openrtb_response`) rewrites all third-party URLs in creative HTML to signed `/first-party/proxy?tsurl=...&tstoken=...` paths using `Settings`. This is the active, working rewriter.

3. **Double rewriting** — When both rewriters were active, `transform_prebid_response` first rewrote `https://cdn.adsrvr.org/pixel.png` → `https://host/ad-proxy/adsrvr/pixel.png`, then `creative::rewrite_creative_html` saw that as an absolute URL and rewrote it *again* to `/first-party/proxy?tsurl=https://host/ad-proxy/adsrvr/pixel.png&tstoken=...` — a proxy URL pointing to a non-existent route.

4. **`nurl`/`burl` rewriting was dead work** — those fields are parsed into the `Bid` struct but never used in the final `convert_to_openrtb_response` output.

## How it was fixed

Instead of patching the untrusted-data path (the original approach used `OnceLock<RequestInfo>` to cache local request info), we removed the dead code entirely:

- Deleted `transform_prebid_response`, `rewrite_ad_markup`, and `make_first_party_proxy_url` (~80 lines)
- Removed `OnceLock<RequestInfo>` field from `PrebidAuctionProvider` — it was only needed to feed the deleted rewriter
- Removed the `RequestInfo` capture in `request_bids` and the `transform_prebid_response` call in `parse_response`
- Cleaned up unused imports (`OnceLock`, `base64::BASE64`, `serde_json::json`)
- Deleted the two tests that exercised the removed functions

`PrebidAuctionProvider` returns to being a stateless config-only struct. The `request_host`/`request_scheme` fields are still sent to Prebid Server in `TrustedServerExt` for the signing protocol — that outbound path is unchanged.

## Verification

- `cargo test --workspace` — all 667 tests pass
- `cargo clippy --all-targets --all-features -- -D warnings` — clean
- Manual: confirm creatives render correctly and `/first-party/proxy` URLs appear in auction responses (not `/ad-proxy/`)

Closes #417

### DIFF
--- a/crates/trusted-server-core/src/cookies.rs
+++ b/crates/trusted-server-core/src/cookies.rs
@@ -3,6 +3,8 @@
 //! This module provides functionality for parsing and creating cookies
 //! used in the trusted server system.
 
+use std::borrow::Cow;
+
 use cookie::{Cookie, CookieJar};
 use error_stack::{Report, ResultExt};
 use fastly::http::header;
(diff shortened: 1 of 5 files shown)


aram356 (approved):
👍 Looks good

PR #553, opened 2026-03-24
Add partner registration endpoint and partner KV registry

### LINKED ISSUE #537: Partner registry and admin endpoint
## Parent epic

#532

## Description

Implement `PartnerRecord`, `PartnerStore`, and the admin registration endpoint that operators use to onboard ID sync partners.

**Scope:** `ec/partner.rs`, `ec/admin.rs`, router update

## Acceptance criteria

- `PartnerRecord` contains all fields from §13.1 including `pull_sync_allowed_domains` and `batch_rate_limit`.
- `PartnerStore::get()`, `upsert()`, `find_by_api_key_hash()` operate on `partner_store` KV. `upsert()` returns `Result<bool, ...>` (true = created, false = updated).
- Secondary indexes: `apikey:{hash} → partner_id` for batch auth, `_pull_enabled → [partner_ids]` for pull sync. Best-effort, not atomic. Old API key index deleted on key rotation.
- `pull_enabled_partners()` re-checks `pull_sync_enabled == true` on fetched records (stale index guard).
- API key stored as SHA-256 hex; plaintext never written to KV. `verify_api_key()` uses constant-time comparison.
- `POST /admin/partners/register` validates `Authorization: Bearer <token>` in-handler against `settings.ec.admin_token_hash`. Returns `401` before reading body if invalid.
- Validates: required fields, `id` format (`^[a-z0-9_-]{1,32}$`), reserved names, pull_sync_url + ts_pull_token required when pull_sync_enabled, pull_sync_url domain in allowlist.
- Returns `201 Created` on new partner, `200 OK` on update, with explicit response DTO (never serialize full PartnerRecord). `400` on validation failure, `503` on KV failure.
- `/admin/partners/register` NOT added to `Settings::ADMIN_ENDPOINTS`. Admin-route-scan test updated with exclusion list. `[[handlers]]` pattern narrowed from `^/admin` to `^/admin/keys`.

## Spec ref

`docs/internal/ssc_technical_spec.md` §13

ChristianPavilonis:
## Summary

- Add Story 4 partner onboarding so operators can register partners through `/admin/partners/register` and persist validated partner records in KV.
- Ensure partner API key auth is reliable by indexing `apikey:{sha256}` entries and failing registration when index writes fail, preventing partial-success states.
- Keep admin auth aligned with current rollout by using existing Basic Auth protection and return 503 when `ec.partner_store` is not configured.

## Changes

| File | Change |
| ---- | ------ |
| `Cargo.toml` | Add workspace dependency on `subtle` for constant-time hash comparison support. |
| `Cargo.lock` | Record lockfile updates for new dependency resolution. |
| `crates/trusted-server-core/Cargo.toml` | Add `subtle` as a core crate dependency. |
| `crates/trusted-server-core/src/ec/mod.rs` | Export new `admin` and `partner` EC modules. |
| `crates/trusted-server-core/src/ec/admin.rs` | Add partner registration request/response models and `/admin/partners/register` handler with validation/defaulting. |
| `crates/trusted-server-core/src/ec/partner.rs` | Add `PartnerRecord`, validation helpers, API key hashing, KV-backed `PartnerStore`, index lookup, and tests. |
| `crates/trusted-server-adapter-fastly/src/main.rs` | Wire POST `/admin/partners/register` route and add partner-store requirement helper. |
| `crates/trusted-server-core/src/settings.rs` | Add new admin endpoint to coverage checks and update admin endpoint validation tests. |

## Closes

Closes #537

## Test plan

- [ ] Docs format: `cd docs && npm run format` *(fails on pre-existing formatting in `docs/guide/configuration.md`, not touched in this PR)*
- [ ] Other: N/A

## Checklist

- [x] New code has tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2764,6 +2764,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2 0.10.9",
+ "subtle",
  "temp-env",
  "tokio",
  "tokio-test",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,6 +82,7 @@ regex = "1.12.3"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.149"
 sha2 = "0.10.9"
+subtle = "2.6"
 temp-env = "0.3.6"
 tokio = { version = "1.49", features = ["sync", "macros", "io-util", "rt", "time"] }
 tokio-test = "0.4"
--- a/crates/trusted-server-adapter-fastly/src/main.rs
+++ b/crates/trusted-server-adapter-fastly/src/main.rs
@@ -10,6 +10,8 @@ use trusted_server_core::constants::{
     ENV_FASTLY_IS_STAGING, ENV_FASTLY_SERVICE_VERSION, HEADER_X_GEO_INFO_AVAILABLE,
     HEADER_X_TS_ENV, HEADER_X_TS_VERSION,
 };
+use trusted_server_core::ec::admin::handle_register_partner;
+use trusted_server_core::ec::partner::PartnerStore;
 use trusted_server_core::error::TrustedServerError;
 use trusted_server_core::geo::GeoInfo;
 use trusted_server_core::http_util::sanitize_forwarded_headers;
@@ -105,10 +107,13 @@ async fn route_request(
         // Signature verification endpoint
         (Method::POST, "/verify-signature") => handle_verify_signature(settings, req),
 
-        // Key rotation admin endpoints
+        // Admin endpoints
         // Keep in sync with Settings::ADMIN_ENDPOINTS in crates/trusted-server-core/src/settings.rs
         (Method::POST, "/admin/keys/rotate") => handle_rotate_key(settings, req),
         (Method::POST, "/admin/keys/deactivate") => handle_deactivate_key(settings, req),
+        (Method::POST, "/admin/partners/register") => {
+            require_partner_store(settings).and_then(|store| handle_register_partner(&store, req))
+        }
 
         // Unified auction endpoint (returns creative HTML inline)
         (Method::POST, "/auction") => handle_auction(settings, orchestrator, req).await,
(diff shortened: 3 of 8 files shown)


aram356 (changes requested):
## Summary

Adds partner registration (`POST /admin/partners/register`) with KV-backed `PartnerStore`, API key hash indexing, and solid input validation. The write-order and rollback logic in `upsert` is well designed for a non-atomic KV store. Two security issues need fixing before merge.

## Blocking

### 🔧 wrench

- **API key collision silently hijacks auth**: `upsert` overwrites the `apikey:{hash}` index without checking if it already belongs to a different partner, silently breaking the first partner's batch sync auth (`partner.rs:278`)
- **Non-constant-time hash comparison in auth path**: `find_by_api_key_hash` uses `!=` instead of `ct_eq` for the hash verification, inconsistent with `verify_api_key` and `auth.rs` (`partner.rs:444`)

### ❓ question

- **`ts_pull_token` stored in plaintext**: API key is properly hashed, but the outbound bearer token is stored as cleartext in KV. Is this intentional? (`partner.rs:95`)

## Non-blocking

### 🤔 thinking

- **No body size limit on registration endpoint**: Unbounded `take_body_bytes()` — other handlers enforce limits. Lower risk since it's behind admin auth, but a cap would be defensive (`admin.rs:154`)
- **`find_by_api_key_hash` opens KV store twice**: Calls `open_store()` then `self.get()` which calls `open_store()` again (`partner.rs:406`)
- **Upsert write order diverges from tech spec**: Spec says primary-first, implementation does index-first. The implementation's order is arguably better (prevents success with broken index), but the spec should be updated to match

### ♻️ refactor

- **Validation helpers return `String` errors**: `validate_partner_id` and `validate_pull_sync_config` return `Result<(), String>` — could return `Result<(), Report<TrustedServerError>>` directly to avoid `.map_err(bad_request)` wrapping at each call site

### ⛏ nitpick

- **`api_key` validated with `trim()` but hashed untrimmed**: Subtly different from `name`/`source_domain` which are trimmed-then-used. A clarifying comment would help (`admin.rs:183`)
- **PR checklist says "Uses `tracing` macros"** but code correctly uses `log` macros per CLAUDE.md — just a template issue

## CI Status

- integration tests: FAIL (pre-existing on base branch)
- browser integration tests: FAIL (pre-existing on base branch)

PR #581, opened 2026-03-26
Add PlatformHttpClient and PlatformBackend traits (EdgeZero PR6)

### LINKED ISSUE #487: Backend + HTTP client traits
## Description

Implement `PlatformBackend` and `PlatformHttpClient` backed by Fastly SDK. Introduce `PlatformResponse` with `backend_name: Option<String>` for upstream correlation. Abstract `send`/`send_async` behind traits. Integration send calls deferred to PR 13.

## Done when

- `PlatformResponse` with `backend_name`
- `send`/`send_async` abstracted
- Per-PR gates pass

## Blocked by

PR 2

## References

- `EDGEZERO_MIGRATION.md` — Phase 1, PR 6
- Part of #480

prk-Jr:
## Summary

- Introduce `PlatformHttpClient` trait with `send()`, `send_async()`, and `select()` for auction fan-out — a superset of EdgeZero's `ProxyClient` that covers the async/select paths not yet in EdgeZero upstream
- Introduce `PlatformBackend` trait with `predict_name()` and `ensure()` to decouple backend registration from Fastly-specific APIs
- Thread `RuntimeServices` through all proxy-layer handlers (`IntegrationProxy::handle`, endpoint handlers, `proxy_request`) so the HTTP client and backend are reachable without global state

## Changes

| File | Change |
| ---- | ------ |
| `platform/http.rs` (new) | `PlatformHttpClient` trait + `PlatformHttpRequest`, `PlatformResponse`, `PlatformPendingRequest`, `PlatformSelectResult` types |
| `platform/backend.rs` (new) | `PlatformBackend` trait + `PlatformBackendSpec` |
| `platform/types.rs` | `RuntimeServices` extended with `http_client` and `backend` fields |
| `platform/test_support.rs` | `StubHttpClient`, `StubBackend`, `build_services_with_http_client` test helpers |
| `integrations/registry.rs` | `IntegrationProxy::handle` + `IntegrationRegistry::handle_proxy` accept `&RuntimeServices` |
| All integration impls | Pass `_services` through `IntegrationProxy::handle` |
| `proxy.rs` | Migrate send path to `services.http_client().send()`; add `ProxyRequestHeaders` struct to stay under 7-arg limit; add `proxy_request_calls_platform_http_client_send` test |
| `auction/orchestrator.rs` | Thread `services` through auction handler |
| `platform.rs` (Fastly adapter) | `FastlyPlatformHttpClient` impl; document `Body::Stream` limitation with warning log |
| `main.rs` (Fastly adapter) | Pass `runtime_services` to all route handlers |

## Closes

Closes #487

## Test plan

## Checklist

- [x] New code has tests

## Notes for reviewer

This branch carries PRs 2–6 cumulatively (crate rename, platform traits, config store, secret store, HTTP client). The focused diff for this PR is the latest commit (`571656c`) plus the PR6-specific commits. The `PlatformHttpClient::send_async` and `select` methods intentionally return `Err(Unsupported)` in the Fastly adapter for now — the fan-out path in `orchestrator.rs` still uses `fastly::http::request::select` directly and will be migrated in a follow-up once EdgeZero adds upstream fan-out support (issues #147–148).


aram356 (changes requested):
## Summary

Well-structured PR introducing a platform abstraction layer that decouples `trusted-server-core` from Fastly-specific HTTP, backend, config/secret/KV store, and geo APIs. Design quality, test coverage, and convention compliance are strong. Two issues should be addressed before merge.

## Blocking

### 🔧 wrench

- **`backend_name` lost on remaining pending requests in `select()`**: When `select()` returns, remaining requests lose their `backend_name` metadata. Violates the trait contract and will break non-Fastly adapters. (`crates/trusted-server-adapter-fastly/src/platform.rs:350`)
- **Duplicated `platform_response_to_fastly`**: Identical copies in `proxy.rs:40` and `orchestrator.rs:32` that silently drop `Body::Stream`. Extract to a shared `pub(crate)` helper now to prevent silent divergence. (`crates/trusted-server-core/src/proxy.rs:40`)

## Non-blocking

### 🤔 thinking

- **Silent empty string on missing backend name**: `unwrap_or_default()` in `select()` produces `""` when backend name is missing, causing silent correlation failure in the orchestrator. Consider logging a warning. (`crates/trusted-server-adapter-fastly/src/platform.rs:358`)
- **`services` parameter vs `ProxyRequestConfig`**: Consider folding `services` into the existing config struct to keep `proxy_request` stable as more platform services arrive. (`crates/trusted-server-core/src/proxy.rs:429`)

### 🌱 seedling

- **`Body::Stream` silently dropped**: Three conversion sites log a warning but produce empty bodies. Consider `debug_assert!` in tests. (`platform.rs:263`, `proxy.rs:44`, `orchestrator.rs:34`)
- **`StubHttpClient` missing `send_async`/`select`**: Auction orchestrator can't be unit-tested with current stub. Consider queue-based fan-out support.

### 👍 praise

- `StoreName`/`StoreId` newtypes — excellent compile-time guard
- `RuntimeServicesBuilder` with `expect("should ...")` — follows conventions perfectly
- Object safety assertions — catches trait changes early
- Control character validation in `BackendConfig` — good defensive coding
- Graceful `UnavailableKvStore` degradation
- Strong test coverage across all new code paths

crates/trusted-server-adapter-fastly/src/platform.rs
```diff
@@ -239,42 +239,175 @@ impl PlatformBackend for FastlyPlatformBackend {
...
+        // Re-attach saved backend names to the remaining pending requests.
+        // Identify which request completed by matching the response backend name
+        // to the saved names, then skip that index when rebuilding remaining.
+        let completed_name = match &result {
+            Ok(resp) => resp.get_backend_name().map(str::to_string),
+            Err(_) => None,
+        };
+        let completed_idx = completed_name
+            .as_deref()
+            .and_then(|name| saved_names.iter().position(|n| n == name));
```
aram356: 🔧 **wrench** — `select()` name correlation is fragile with duplicate backend names

`position()` returns the first match in `saved_names`. If two pending requests share the same backend name (e.g., two requests to the same origin with the same timeout), this misaligns the remaining names with their requests, causing silent misrouting.

Today the auction orchestrator enforces unique backend names per provider, but the `select()` API contract doesn't document this uniqueness requirement.

**Fix:** Document the uniqueness requirement on `PlatformHttpClient::select`, or switch to positional tracking (track which index was removed from the Fastly `select()` result rather than name-matching).

crates/trusted-server-core/src/proxy.rs
```diff
@@ -569,29 +598,61 @@ async fn proxy_with_redirects(
...
         }
-
         for (name, value) in request_headers.additional_headers {
-            proxy_req.set_header(name.clone(), value.clone());
+            builder = builder.header(name.clone(), value.clone());
         }
-
-        let beresp = proxy_req
-            .send(&backend_name)
+        let body_bytes = body.map(<[u8]>::to_vec).unwrap_or_default();
```
aram356: 🤔 **thinking** — Empty body always attached for bodyless requests

`body.map(<[u8]>::to_vec).unwrap_or_default()` creates an empty `Vec<u8>` when `body` is `None`. Previously, bodies were only set when present. For `GET`/`HEAD` requests, sending a zero-length body may trip strict intermediaries.

**Suggestion:**
```rust
let edge_body = match body {
    Some(bytes) => EdgeBody::from(bytes.to_vec()),
    None => EdgeBody::empty(),
};
```

ChristianPavilonis (commented):
## PR #581 Review: Add PlatformHttpClient and PlatformBackend traits

**Overall**: Well-structured PR with strong abstractions and thorough test coverage. The previous review round's findings were all addressed. Two new findings warrant attention before merge (header handling and the select() ordering assumption); the rest are improvements for follow-ups.

### Praise
- `PlatformPendingRequest` downcast API returning `Err(self)` to preserve backend metadata is a thoughtful design
- `!Send` rationale documentation on `PlatformPendingRequest` clearly explains the wasm32 constraint
- `StubHttpClient` fan-out test infrastructure with `StubPendingResponse` is clean and comprehensive
- `RuntimeServicesBuilder` with `expect("should ...")` messages follows project conventions perfectly

crates/trusted-server-adapter-fastly/src/platform.rs
```diff
@@ -187,45 +187,175 @@ impl PlatformBackend for FastlyPlatformBackend {
...
+        if completed_name.is_some() && completed_idx.is_none() {
+            log::warn!(
+                "select: completed backend name not found in saved names; \
+                 remaining requests will lose backend correlation"
+            );
+        }
+
+        let remaining: Vec<PlatformPendingRequest> = if let Some(idx) = completed_idx {
+            remaining_fastly
```
ChristianPavilonis: 🔧 **Backend name re-association relies on positional zip with unordered remaining list**

Fastly's `select()` docs state: *"The order of the requests in this vector is not guaranteed to match the order of the requests in the argument collection."*

The positional zip on lines 323–333 assumes `remaining_fastly` preserves input order, which isn't guaranteed. With 3+ pending requests, backend names may be misattributed to the wrong remaining handles.

**In practice this isn't a data-corruption bug today** — the orchestrator correlates completed responses via `platform_response.backend_name` (from `get_backend_name()`), and remaining requests get re-correlated when they complete. The `backend_name` on remaining `PlatformPendingRequest`s is used only for debugging/logging.

**Suggested fix**: Either document the limitation, or simplify by not attaching names to remaining requests (since they're re-derived on completion):

```rust
// Fastly's select() does not preserve input order for remaining
// requests, so positional name re-association is unreliable.
// Names are re-derived from get_backend_name() when each completes.
let remaining: Vec<PlatformPendingRequest> = remaining_fastly
    .into_iter()
    .map(PlatformPendingRequest::new)
    .collect();
```

ChristianPavilonis (commented):
Good PR — the platform abstraction layer is well-designed with thoughtful type erasure, thorough test coverage, and clean `RuntimeServices` threading. A few items to address before merge:

**P0 — Stale doc comment**: `orchestrator.rs` `run_providers_parallel` doc (line 244) still references `fastly::http::request::select()` but the implementation now uses `services.http_client().select()`. _(This line is not in the diff, so noting here instead of inline.)_

**P1 — Silent body truncation in release builds**: The `debug_assert!` + `log::warn!` pattern for `Body::Stream` in `edge_request_to_fastly`, `platform_response_to_fastly`, and `fastly_response_to_platform` is a no-op in release. If a streaming body reaches these paths, requests/responses are sent with empty bodies. Recommend upgrading to `log::error!` or returning `Result`.

**P1 — Downcast failure drops all pending requests**: In `FastlyPlatformHttpClient::select()`, if any single `PlatformPendingRequest` fails downcast, the entire function errors and all remaining in-flight requests are lost. Consider adding the backend name to the error message for diagnostics.

**P2 — Pre-existing**: `rebuild_response_with_body` still uses `set_header` (drops multi-value headers like `Set-Cookie`), while the new conversion functions correctly use `append_header`.

ChristianPavilonis (approved):

crates/trusted-server-core/src/proxy.rs
```diff
@@ -560,7 +579,7 @@ async fn proxy_with_redirects(
             }));
         }
 
-        if !redirect_is_permitted(request_headers.allowed_domains, host) {
+        if !redirect_is_permitted(&settings.proxy.allowed_domains, host) {
```
aram356: 🔧 **wrench** — `allowed_domains` enforcement regressed for integration proxies

`allowed_domains` was removed from `ProxyRequestConfig`, so `proxy_with_redirects` now reads `settings.proxy.allowed_domains` unconditionally. Previously, integration proxies (GTM, testlight) called `ProxyRequestConfig::new()` which set `allowed_domains: &[]` — **open mode** where all hosts are permitted. Only the first-party proxy overrode this with `&settings.proxy.allowed_domains`.

Now all callers share the same allowed_domains list. In production, when `settings.proxy.allowed_domains` is non-empty (e.g. `["publisher-cdn.com"]`), integration proxies will be **blocked from contacting their upstream domains** (like `googletagmanager.com` for GTM) because those domains aren't in the first-party allowed list.

The same issue applies to the redirect check on line 702.

**Fix**: Restore `allowed_domains` on `ProxyRequestHeaders` so integration proxies can opt out:

```rust
struct ProxyRequestHeaders<'a> {
    additional_headers: &'a [(header::HeaderName, HeaderValue)],
    copy_request_headers: bool,
    services: &'a RuntimeServices,
    allowed_domains: &'a [String],  // restore
}
```

Pass `&[]` from integration callers, `&settings.proxy.allowed_domains` from first-party handlers.

crates/trusted-server-core/src/proxy.rs
```diff
@@ -1911,6 +1951,115 @@ mod tests {
...
+        let result = proxy_request(
+            &settings,
+            req,
+            ProxyRequestConfig {
+                target_url: "https://example.com/resource",
+                follow_redirects: false,
+                forward_ec_id: false,
+                body: None,
+                headers: Vec::new(),
+                copy_request_headers: false,
```
aram356, line 2012: ♻️ **refactor** — Header forwarding test removed without replacement

The thorough `header_copy_copies_curated_set` test was deleted. It validated that each of the 5 curated headers was forwarded and that `Accept-Encoding` was overridden to `SUPPORTED_ENCODINGS`.

This test uses `copy_request_headers: false`, so the header forwarding path through the EdgeZero builder is **not tested at all**.

**Suggestion**: Add a test with `copy_request_headers: true` that inspects the `PlatformHttpRequest` received by the stub to verify headers were forwarded correctly. The `StubHttpClient` already records calls — extending it to also capture the request headers would make this straightforward.

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -150,9 +156,27 @@ impl AuctionOrchestrator {
                     message: format!("Mediator {} failed to launch", mediator.provider_name()),
                 })?;
 
-            let backend_response = pending.wait().change_context(TrustedServerError::Auction {
-                message: format!("Mediator {} request failed", mediator.provider_name()),
-            })?;
+            let select_result = services
+                .http_client()
+                .select(vec![PlatformPendingRequest::new(pending)])
```
aram356: 🤔 **thinking** — `select()` used for single-request wait

The mediator wait uses `select(vec![PlatformPendingRequest::new(pending)])` to wait for a single request. This involves wrapping, downcasting, and destructuring a `PlatformSelectResult` for what is semantically just "wait for this one request." Previously it was simply `pending.wait()`.

The code from here through line 177 is noticeably more complex for the same operation. A `PlatformHttpClient::wait` default method would simplify this:

```rust
async fn wait(&self, pending: PlatformPendingRequest) -> Result<PlatformResponse, Report<PlatformError>> {
    let result = self.select(vec![pending]).await?;
    result.ready
}
```

Not blocking — could be a follow-up.

aram356 (changes requested):
## Summary

Implements `PlatformHttpClient` with `send`/`send_async`/`select` and threads `RuntimeServices` through all proxy-layer handlers. Good test coverage for the new platform abstractions, but the `allowed_domains` removal from `ProxyRequestConfig` introduces a correctness regression for integration proxies.

## Blocking

### 🔧 wrench

- **Integration proxies blocked by first-party allowed_domains**: Removing `allowed_domains` from `ProxyRequestConfig` means GTM, testlight, and other integration proxies now have `settings.proxy.allowed_domains` enforced on their upstream requests. Previously they operated in open mode (`&[]`). In production this will block integrations from contacting their upstream domains (`crates/trusted-server-core/src/proxy.rs:582`, `:702`)

## Non-blocking

### ♻️ refactor

- **Header forwarding test removed without replacement**: The `header_copy_copies_curated_set` test was deleted but `proxy_request_calls_platform_http_client_send` uses `copy_request_headers: false`, leaving the header forwarding path untested (`crates/trusted-server-core/src/proxy.rs:1977`)

### 🤔 thinking

- **`select()` used for single-request mediator wait**: Wrapping a single pending request in `select(vec![...])` adds unnecessary complexity vs a dedicated `wait()` method (`crates/trusted-server-core/src/auction/orchestrator.rs:161`)

### ⛏ nitpick

- **`PROXY_FORWARD_HEADERS` const vs inline `Accept-Encoding`**: The 5 forwarded headers are in a const array but Accept-Encoding override is added separately inline — minor inconsistency vs the old unified `copy_proxy_forward_headers` function

## CI Status

- fmt: PASS
- clippy: PASS
- cargo test: PASS
- vitest: PASS
- integration tests: PASS
- CodeQL: PASS

crates/trusted-server-core/src/proxy.rs
prk-Jr: Fixed. Restored `allowed_domains` to both `ProxyRequestConfig` (public) and `ProxyRequestHeaders` (private). `ProxyRequestConfig::new()` defaults to `&[]` (open mode) so GTM, testlight, GPT, and all integration callers automatically get open mode with no changes on their side. `handle_first_party_proxy` now explicitly passes `&settings.proxy.allowed_domains`. `proxy_with_redirects` uses `request_headers.allowed_domains` at both the initial request check and the redirect check.

crates/trusted-server-core/src/proxy.rs
prk-Jr: Fixed. Extended `StubHttpClient` with a `request_headers` recorder and `recorded_request_headers()` method that captures the `(name, value)` pairs from each `send` call. Added `proxy_request_forwards_curated_headers_when_copy_request_headers_is_true`: sets User-Agent, Accept, and Accept-Language on the incoming request, calls `proxy_request` with `copy_request_headers: true`, then asserts all five curated headers are forwarded and that Accept-Encoding is overridden to `SUPPORTED_ENCODINGS`.

crates/trusted-server-core/src/auction/orchestrator.rs
prk-Jr: Fixed. Added `wait()` as a default method on `PlatformHttpClient` — no changes needed to any existing impl. The mediator callsite is now `services.http_client().wait(PlatformPendingRequest::new(pending)).await`, replacing the 8-line `select`/`select_result.ready` chain.

aram356 (commented):
## Summary

All previously-blocking findings from the prior review rounds are resolved and covered by new tests. The trait design (`async_trait(?Send)` + `Box<dyn Any>` type-erased pending requests) is sound for WASM. No correctness blockers remain — only follow-up quality items.

## Non-blocking

### 🤔 thinking

- **Orchestrator `select()` fan-out has no end-to-end stub-driven test.** `AuctionOrchestrator::run_providers_parallel` is only covered by `test_no_providers_configured`. Root cause: `AuctionProvider::request_bids` still hands back a concrete `fastly::http::request::PendingRequest`, so `StubHttpClient` can't be injected without a provider-level abstraction. Ships the deadline branch, backend-name correlation, unknown-backend warning, and remaining-drop logic without direct coverage. (`crates/trusted-server-core/src/auction/orchestrator.rs:379-460`)
- **`PlatformPendingRequest` metadata lost on `select()` rewrap.** Remaining requests are rebuilt with `PlatformPendingRequest::new(pending)` — no `.with_backend_name()` (`crates/trusted-server-adapter-fastly/src/platform.rs:308-311`). Any future caller inspecting `pending.backend_name()` on a remaining request (e.g. the "dropping N remaining" log at `orchestrator.rs:454`) will see `None`. Consider re-deriving via `pending.get_backend_name().map(str::to_string)` on the rewrap.
- **`StubHttpClient::select` is strictly FIFO.** Always pops index 0, which is a less faithful double than Fastly's unordered `select()`. Tests built on it can accidentally depend on "first pushed = first ready," masking reordering bugs. Either document explicitly on the stub or randomize which item is marked ready. (`crates/trusted-server-core/src/platform/test_support.rs:250-278`)
- **`IntegrationProxy: Send + Sync` with `?Send` futures.** Consistent for single-threaded WASM, but a one-line rationale mirroring the excellent `!Send` doc block on `PlatformPendingRequest` would save future readers time. (`crates/trusted-server-core/src/integrations/registry.rs:247-248`)

### ♻️ refactor

- **`services` parameter proliferation.** Four proxy entry points now take `(settings, req, config, services)`. As more platform services thread through in PR7+, consider absorbing `services` into `ProxyRequestConfig` or a lightweight `ProxyContext<'a>`. Next PR, not this one. (`crates/trusted-server-core/src/proxy.rs:447-451`)
- **`PROXY_FORWARD_HEADERS` const vs. inline `Accept-Encoding` override.** The 5 forwarded headers live in a const array but `Accept-Encoding` is appended separately at `proxy.rs:634-637`, splitting header curation across two sites. Pulling the override into the same helper would keep curation in one place.

### ⛏ nitpick

- **`unwrap_or("")` fallback in `select()` backend-name** is now logged (good), but the subsequent `backend_to_provider.remove("")` in the orchestrator will still silently drop the response. Preferable: return `PlatformResponse { backend_name: None, .. }` and let the orchestrator's existing unknown-backend branch handle it with its more specific log message. (`crates/trusted-server-adapter-fastly/src/platform.rs:315-321`)
- Brief doc comment on the `StreamingResponseHttpClient` helper in `test_support.rs` clarifying what it models, since it's the only `!Send`-body test surface.

### 👍 praise

- `PlatformPendingRequest::downcast` returning `Err(self)` — preserves backend metadata on downcast failure. Thoughtful API.
- Exemplary `!Send` rationale doc block on `PlatformPendingRequest` — please keep that pattern on future platform traits.
- Default `wait()` method on the trait cleanly eliminates the "single-item select" awkwardness flagged in the previous round.
- `StubHttpClient` now supports `send_async`/`select` with per-call header capture — a real fan-out test surface, not just a stub-that-errors.
- `proxy_request_returns_error_for_streaming_platform_response_body` proving the release-mode behavior of the new error path is exactly the right response to the previous `debug_assert!` concern.
- Zero new Cargo dependencies — the entire abstraction lands with no supply-chain cost.

## Resolved from prior rounds

All P0/P1 items from the prior aram356 and ChristianPavilonis rounds are verified resolved on `bbef5223`:

- `backend_name` correlation now re-derived from `fastly::Response::get_backend_name()`, positional zip removed.
- Duplicated `platform_response_to_fastly` consolidated as `pub(crate)` in `proxy.rs` and reused from `orchestrator.rs`.
- `Body::Stream` truncation via `debug_assert!` now returns `PlatformError::HttpClient` / `TrustedServerError::Proxy` in release, covered by new tests on both sides of the boundary.
- `set_header` multi-value drop replaced with `append_header` everywhere; regression test `rebuild_response_with_body_preserves_multiple_set_cookie_headers` added.
- `allowed_domains` regression fixed; field restored and threaded through `handle_first_party_proxy`.
- Curated-header forwarding test re-added as `proxy_request_forwards_curated_headers_when_copy_request_headers_is_true`.

## CI Status

- cargo fmt: PASS
- cargo clippy / Analyze (rust / wasm32-wasip1): PASS
- cargo test: PASS
- vitest: PASS
- integration + browser integration: PASS
- CodeQL: PASS
- format-docs / format-typescript: PASS

aram356 (approved):

PR #594, opened 2026-03-27
Phase 4: Stream binary pass-through responses via send_to_client()

### LINKED ISSUE #592: Phase 4, Task 16: Stream binary pass-through responses via io::copy
Parent: #563

## Summary

Non-processable responses (images, fonts, video, \`application/octet-stream\`) currently pass through \`handle_publisher_request\` unchanged via the \`Buffered\` path. This buffers the entire response body in memory before sending to the client — wasteful for large binaries that need no processing.

## Changes

- Add \`PublisherResponse::PassThrough { response, body }\` variant — signals the adapter to stream the body directly via \`io::copy\` into \`StreamingBody\` with no pipeline
- Return \`PassThrough\` when \`!should_process\` and backend returned 2xx
- Update \`main.rs\` adapter to handle the new variant with \`stream_to_client()\` + \`io::copy(body, &mut streaming_body)\`
- Keep \`Buffered\` for non-2xx responses (error pages) and \`request_host.is_empty()\`

**Files:** \`publisher.rs\`, \`main.rs\`

## Expected impact

Eliminates peak memory for large binary responses (images can be 1-10 MB). Improves DOM Complete for image-heavy pages since the browser receives image bytes earlier.

## Plan
See \`docs/superpowers/specs/2026-03-25-streaming-response-design.md\`

### LINKED ISSUE #593: Phase 4, Task 17: Binary pass-through tests and verification
Parent: #563
Depends on: Task 16

## Summary

Add tests and run full verification for binary pass-through streaming.

## Tests

- Publisher-level test: image content type returns \`PassThrough\` (not \`Buffered\`)
- Publisher-level test: 4xx image stays \`Buffered\`
- Verify \`Content-Length\` is preserved for pass-through (no chunked encoding needed since body is unmodified)
- Full verification: \`cargo test\`, clippy, fmt, WASM build

## Plan
See \`docs/superpowers/specs/2026-03-25-streaming-response-design.md\`

aram356:
## Summary

Stream non-processable 2xx responses (images, fonts, video) directly to the client instead of buffering the entire body in memory. Uses `send_to_client()` with the unmodified body to preserve `Content-Length` and avoid chunked encoding overhead.

Closes #592, closes #593.
Part of epic #563. Depends on Phase 3 (#591).

### Performance results (staging vs production, median over 8 runs, Chrome 1440x900)

| Metric | Production (v135, buffered) | Staging (v141, streaming+passthrough) | Delta |
|---|---|---|---|
| **TTFB** | 57 ms | 28 ms | **-29 ms (-51%)** |
| **First Paint** | 252 ms | 182 ms | **-70 ms (-28%)** |
| **First Contentful Paint** | 252 ms | 182 ms | **-70 ms (-28%)** |
| **DOM Content Loaded** | 322 ms | 248 ms | **-74 ms (-23%)** |
| **DOM Complete** | 678 ms | 582 ms | -96 ms (-14%) |

Measured on getpurpose.ai via Chrome DevTools Protocol with `--host-resolver-rules` to route to staging Fastly edge.

### How it works

The streaming gate in `handle_publisher_request` now has three outcomes:

```
is_success (2xx, not 204/205)
├── should_process && (!is_html || !has_post_processors) → Stream (pipeline)
├── should_process && is_html && has_post_processors     → Buffered (post-processors)
└── !should_process                                      → PassThrough (send_to_client)

204 No Content, 205 Reset Content, or !is_success
└── any content type                                     → Buffered
```

`PassThrough` reattaches the unmodified body and sends via `send_to_client()`, preserving `Content-Length`. This avoids both WASM memory buffering and chunked encoding overhead.

### What changed

| File | What |
|------|------|
| `publisher.rs` | `PassThrough` variant, gate logic, 204/205 exclusion, debug logging, tests |
| `main.rs` | Handle `PassThrough`: reattach body, return for `send_to_client()` |

### Behavioral changes worth calling out

- **Empty-host + non-processable 2xx now `PassThrough` instead of `Buffered`.** Before this PR, requests with an empty `request_host` always took the buffered path. URL rewriting is skipped either way, so this is functionally equivalent — but observable as a different code path.
- **Buffered path now sets `Content-Length` to the processed output length.** Previously the buffered path called `response.remove_header(CONTENT_LENGTH)` after processing; this PR sets it to `output.len()` instead. Browsers that previously fell back to chunked encoding for processed HTML now get an explicit length.

### Tests added

- `pass_through_gate_streams_non_processable_2xx` — images/fonts return PassThrough
- `pass_through_gate_buffers_non_processable_error` — 4xx stays Buffered
- `pass_through_gate_does_not_apply_to_processable_content` — HTML goes through Stream
- `pass_through_gate_excludes_204_no_content` — 204 stays Buffered (RFC 9110 §15.3.5)
- `pass_through_gate_excludes_205_reset_content` — 205 stays Buffered (RFC 9110 §15.3.6)
- `pass_through_gate_applies_with_empty_request_host` — empty host still passes through
- `pass_through_preserves_body_and_content_length` — byte-for-byte identity + CL preserved

### Verification

- `cargo test --workspace` — all passing (819 unit + integration)
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean
- `cargo build --release --target wasm32-wasip1` — success

## Test plan

- [x] Pass-through gate unit tests pass (7 tests)
- [x] Byte-level body + Content-Length preservation test passes
- [x] 204 No Content and 205 Reset Content exclusion tests pass
- [x] All existing streaming/buffered tests pass
- [x] Full workspace tests pass
- [x] WASM build succeeds
- [x] Staging performance verified (51% TTFB, 28% FCP, 14% DOM Complete improvement)

### DIFF
--- a/crates/trusted-server-adapter-fastly/src/main.rs
+++ b/crates/trusted-server-adapter-fastly/src/main.rs
@@ -238,10 +238,7 @@ async fn route_request(
                             // Response already sent via stream_to_client()
                             return None;
                         }
-                        Ok(PublisherResponse::PassThrough {
-                            mut response,
-                            body,
-                        }) => {
+                        Ok(PublisherResponse::PassThrough { mut response, body }) => {
                             // Binary pass-through: reattach body and send via send_to_client().
                             // This preserves Content-Length and avoids chunked encoding overhead.
                             // Fastly streams the body from its internal buffer — no WASM
--- a/crates/trusted-server-adapter-fastly/src/main.rs
+++ b/crates/trusted-server-adapter-fastly/src/main.rs
@@ -244,6 +244,23 @@ async fn route_request(
                             Err(e)
                         }
                     }
+                        Ok(PublisherResponse::PassThrough {
+                            mut response,
+                            body,
+                        }) => {
+                            // Binary pass-through: reattach body and send via send_to_client().
+                            // This preserves Content-Length and avoids chunked encoding overhead.
+                            // Fastly streams the body from its internal buffer — no WASM
+                            // memory buffering occurs.
+                            response.set_body(body);
+                            Ok(response)
+                        }
+                        Ok(PublisherResponse::Buffered(response)) => Ok(response),
+                        Err(e) => {
+                            log::error!("Failed to proxy to publisher origin: {:?}", e);
+                            Err(e)
+                        }
+                    }
                 }
                 Err(e) => Err(e),
             }
--- a/crates/trusted-server-adapter-fastly/src/main.rs
+++ b/crates/trusted-server-adapter-fastly/src/main.rs
@@ -238,12 +238,6 @@ async fn route_request(
                             // Response already sent via stream_to_client()
                             return None;
                         }
-                        Ok(PublisherResponse::Buffered(response)) => Ok(response),
-                        Err(e) => {
-                            log::error!("Failed to proxy to publisher origin: {:?}", e);
-                            Err(e)
-                        }
-                    }
                         Ok(PublisherResponse::PassThrough {
                             mut response,
                             body,
--- a/crates/trusted-server-core/src/publisher.rs
+++ b/crates/trusted-server-core/src/publisher.rs
@@ -332,12 +332,14 @@ pub fn stream_publisher_body<W: Write>(
 
 /// Proxies requests to the publisher's origin server.
 ///
-/// Returns a [`PublisherResponse`] indicating whether the response can be
-/// streamed or must be sent buffered. The streaming path is chosen when:
-/// - The backend returns a 2xx status
-/// - The response has a processable content type
-/// - The response uses a supported `Content-Encoding` (gzip, deflate, br)
-/// - No HTML post-processors are registered (the streaming gate)
+/// Returns a [`PublisherResponse`] indicating how the response should be sent:
+/// - [`PassThrough`](PublisherResponse::PassThrough) — 2xx non-processable content
+///   (images, fonts, video). Body reattached unmodified for `send_to_client()`.
+/// - [`Stream`](PublisherResponse::Stream) — 2xx processable content with supported
+///   `Content-Encoding` and no HTML post-processors. Body piped through the
+///   streaming pipeline.
+/// - [`Buffered`](PublisherResponse::Buffered) — non-2xx responses, unsupported
+///   encoding, or HTML with post-processors that need the full document.
 ///
 /// # Errors
 ///
@@ -787,6 +789,38 @@ mod tests {
         assert!(
             !should_pass_through,
             "processable content should go through Stream, not PassThrough"
+        );
+    }
+
+    #[test]
+    fn pass_through_preserves_body_and_content_length() {
+        // Simulate the PassThrough path: take body from response, io::copy to output.
+        // Verify byte-for-byte identity and that Content-Length is preserved.
+        let image_bytes: Vec<u8> = (0..=255).cycle().take(4096).collect();
+
+        let mut response = Response::from_status(StatusCode::OK);
+        response.set_header("content-type", "image/png");
+        response.set_header("content-length", image_bytes.len().to_string());
+        response.set_body(Body::from(image_bytes.clone()));
+
+        // Simulate PassThrough: take body, preserve Content-Length
+        let content_length = response
+            .get_header_str("content-length")
+            .map(str::to_string);
+        let mut body = response.take_body();
+
+        // io::copy into a Vec (simulating StreamingBody)
+        let mut output = Vec::new();
+        std::io::copy(&mut body, &mut output).expect("should copy body");
+
+        assert_eq!(
+            output, image_bytes,
+            "pass-through should preserve body byte-for-byte"
+        );
+        assert_eq!(
+            content_length.as_deref(),
+            Some("4096"),
+            "Content-Length should be preserved for pass-through"
         );
     }
 
--- a/crates/trusted-server-core/src/publisher.rs
+++ b/crates/trusted-server-core/src/publisher.rs
@@ -280,7 +280,8 @@ pub enum PublisherResponse {
     /// 3. Copy body bytes directly via `io::copy(&mut body, &mut streaming_body)`
     /// 4. Call `StreamingBody::finish()`
     ///
-    /// `Content-Length` is preserved since the body is unmodified.
+    /// `Content-Length` is removed because `stream_to_client()` uses chunked
+    /// transfer encoding. The body content is unmodified.
     PassThrough {
         /// Response with all headers set but body not yet written.
         response: Response,
@@ -461,8 +462,12 @@ pub fn handle_publisher_request(
 
         // Stream non-processable 2xx responses directly to avoid buffering
         // large binaries (images, fonts, video) in memory.
-        if response.get_status().is_success() && !should_process {
+        // Exclude 204 No Content — it must not have a message body, and
+        // stream_to_client() would add chunked Transfer-Encoding.
+        let status = response.get_status();
+        if status.is_success() && status != StatusCode::NO_CONTENT && !should_process {
             let body = response.take_body();
+            response.remove_header(header::CONTENT_LENGTH);
             return Ok(PublisherResponse::PassThrough { response, body });
         }
 
@@ -793,9 +798,41 @@ mod tests {
     }
 
     #[test]
-    fn pass_through_preserves_body_and_content_length() {
-        // Simulate the PassThrough path: take body from response, io::copy to output.
-        // Verify byte-for-byte identity and that Content-Length is preserved.
+    fn pass_through_gate_excludes_204_no_content() {
+        // 204 must not have a message body; stream_to_client would add
+        // chunked Transfer-Encoding which violates HTTP spec.
+        let status = StatusCode::NO_CONTENT;
+        let should_process = false;
+        let should_pass_through =
+            status.is_success() && status != StatusCode::NO_CONTENT && !should_process;
+        assert!(
+            !should_pass_through,
+            "204 No Content should not use PassThrough"
+        );
+    }
+
+    #[test]
+    fn pass_through_gate_applies_with_empty_request_host() {
+        // Non-processable 2xx with empty request_host still gets PassThrough.
+        // The empty-host path only blocks processing (URL rewriting needs a host);
+        // pass-through doesn't process, so the host is irrelevant.
+        let should_process = false;
+        let is_success = true;
+        let request_host_empty = true;
+        // In production: enters the `!should_process || request_host.is_empty()` block,
+        // then the PassThrough guard checks `is_success && !should_process` — host irrelevant.
+        let _enters_early_return = !should_process || request_host_empty;
+        let should_pass_through = is_success && !should_process;
+        assert!(
+            should_pass_through,
+            "non-processable 2xx with empty host should still pass-through"
+        );
+    }
+
+    #[test]
+    fn pass_through_preserves_body_and_removes_content_length() {
+        // Simulate the PassThrough path: take body, remove Content-Length,
+        // io::copy to output. Verify byte-for-byte identity.
         let image_bytes: Vec<u8> = (0..=255).cycle().take(4096).collect();
 
         let mut response = Response::from_status(StatusCode::OK);
@@ -803,11 +840,10 @@ mod tests {
         response.set_header("content-length", image_bytes.len().to_string());
         response.set_body(Body::from(image_bytes.clone()));
 
-        // Simulate PassThrough: take body, preserve Content-Length
-        let content_length = response
-            .get_header_str("content-length")
-            .map(str::to_string);
+        // Simulate PassThrough: take body, remove Content-Length
+        // (stream_to_client uses chunked encoding)
         let mut body = response.take_body();
+        response.remove_header(header::CONTENT_LENGTH);
 
         // io::copy into a Vec (simulating StreamingBody)
         let mut output = Vec::new();
@@ -817,10 +853,9 @@ mod tests {
             output, image_bytes,
             "pass-through should preserve body byte-for-byte"
         );
-        assert_eq!(
-            content_length.as_deref(),
-            Some("4096"),
-            "Content-Length should be preserved for pass-through"
+        assert!(
+            response.get_header(header::CONTENT_LENGTH).is_none(),
+            "Content-Length should be removed for streaming pass-through"
         );
     }
 
--- a/crates/trusted-server-core/src/publisher.rs
+++ b/crates/trusted-server-core/src/publisher.rs
@@ -275,8 +275,8 @@ pub enum PublisherResponse {
         params: OwnedProcessResponseParams,
     },
     /// Non-processable 2xx response (images, fonts, video). The caller must:
-    /// 1. Call `finalize_response()` on the response
-    /// 2. Reattach the body via `response.set_body(body)`
+    /// 1. Reattach the body via `response.set_body(body)`
+    /// 2. Call `finalize_response()` on the response
     /// 3. Call `response.send_to_client()`
     ///
     /// `Content-Length` is preserved — the body is unmodified. Using
@@ -819,10 +819,9 @@ mod tests {
         // pass-through doesn't process, so the host is irrelevant.
         let should_process = false;
         let is_success = true;
-        let request_host_empty = true;
-        // In production: enters the `!should_process || request_host.is_empty()` block,
-        // then the PassThrough guard checks `is_success && !should_process` — host irrelevant.
-        let _enters_early_return = !should_process || request_host_empty;
+        // In production, empty host enters the early-return block via
+        // `!should_process || request_host.is_empty()`. The PassThrough guard
+        // checks `is_success && !should_process` — host is irrelevant.
         let should_pass_through = is_success && !should_process;
         assert!(
             should_pass_through,
--- a/crates/trusted-server-core/src/publisher.rs
+++ b/crates/trusted-server-core/src/publisher.rs
@@ -276,12 +276,13 @@ pub enum PublisherResponse {
     },
     /// Non-processable 2xx response (images, fonts, video). The caller must:
     /// 1. Call `finalize_response()` on the response
-    /// 2. Call `response.stream_to_client()` to get a `StreamingBody`
-    /// 3. Copy body bytes directly via `io::copy(&mut body, &mut streaming_body)`
-    /// 4. Call `StreamingBody::finish()`
+    /// 2. Reattach the body via `response.set_body(body)`
+    /// 3. Call `response.send_to_client()`
     ///
-    /// `Content-Length` is removed because `stream_to_client()` uses chunked
-    /// transfer encoding. The body content is unmodified.
+    /// `Content-Length` is preserved — the body is unmodified. Using
+    /// `send_to_client()` instead of `stream_to_client()` avoids chunked
+    /// encoding overhead. Fastly streams the body from its internal buffer
+    /// without copying into WASM memory.
     PassThrough {
         /// Response with all headers set but body not yet written.
         response: Response,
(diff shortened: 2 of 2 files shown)


prk-Jr (dismissed):

crates/trusted-server-core/src/publisher.rs
```diff
@@ -198,6 +198,21 @@ pub enum PublisherResponse {
         /// Parameters for `process_response_streaming`.
         params: OwnedProcessResponseParams,
     },
+    /// Non-processable 2xx response (images, fonts, video). The caller must:
```
prk-Jr: 🔧 **wrench** — The doc comment prescribes the wrong operation order.

It lists: finalize → `set_body` → `send_to_client`. But the actual implementation in `main.rs` does: `set_body` → finalize (implicitly via the epilogue) → `send_to_client`. Since `finalize_response` only sets headers and never touches the body, both orders are functionally equivalent — but a future implementor reading only this comment might be confused, especially when comparing with the `Stream` variant where `finalize_response` is called explicitly before body operations.

**Suggested fix:**
```rust
/// Non-processable 2xx response (images, fonts, video). The caller must:
/// 1. Reattach the body via `response.set_body(body)`
/// 2. Call `finalize_response()` on the response
/// 3. Call `response.send_to_client()`
```

crates/trusted-server-core/src/publisher.rs
```diff
@@ -198,6 +198,21 @@ pub enum PublisherResponse {
...
+    /// 2. Reattach the body via `response.set_body(body)`
+    /// 3. Call `response.send_to_client()`
+    ///
+    /// `Content-Length` is preserved — the body is unmodified. Using
+    /// `send_to_client()` instead of `stream_to_client()` avoids chunked
+    /// encoding overhead. Fastly streams the body from its internal buffer
+    /// without copying into WASM memory.
+    PassThrough {
+        /// Response with all headers set but body not yet written.
+        response: Response,
```
prk-Jr: ♻️ **refactor** — `pub` → `pub(crate)` on the struct fields is a clean API surface reduction. The adapter crate passes this struct opaquely so the narrower visibility is correct and follows CLAUDE.md's guidance to keep visibility minimal. 👍

crates/trusted-server-core/src/publisher.rs
```diff
@@ -360,6 +377,18 @@ pub fn handle_publisher_request(
...
             request_host
         );
+
+        // Stream non-processable 2xx responses directly to avoid buffering
+        // large binaries (images, fonts, video) in memory.
+        // Content-Length is preserved — the body is unmodified, so the
+        // browser knows the exact size for progress/layout.
+        // Exclude 204 No Content — it must not have a message body.
+        let status = response.get_status();
+        if status.is_success() && status != StatusCode::NO_CONTENT && !should_process {
```
prk-Jr: 👍 **praise** — Correct RFC 9110 §15.3.5 compliance. The guard is placed before `take_body()` and the dedicated `pass_through_gate_excludes_204_no_content` test makes the intent visible.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -597,6 +626,105 @@ mod tests {
         );
     }
 
+    #[test]
+    fn pass_through_gate_streams_non_processable_2xx() {
```
prk-Jr: 🤔 **thinking** — These tests duplicate the gate boolean formula locally rather than calling a named function, so if the condition in `handle_publisher_request` changes the tests will still pass (they test their own copy). This is the exact problem that was fixed for the streaming gate in `db597e6` by extracting `can_stream_response()`. Consider the same here:

```rust
fn can_pass_through(status: StatusCode, should_process: bool) -> bool {
    status.is_success() && status != StatusCode::NO_CONTENT && !should_process
}
```

Then the tests call `can_pass_through(...)` directly instead of rebuilding the formula. Non-blocking — the gate logic itself is correct — but worth aligning with the streaming-gate pattern for consistency.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -597,6 +626,105 @@ mod tests {
...
+    fn pass_through_gate_applies_with_empty_request_host() {
+        // Non-processable 2xx with empty request_host still gets PassThrough.
+        // The empty-host path only blocks processing (URL rewriting needs a host);
+        // pass-through doesn't process, so the host is irrelevant.
+        let should_process = false;
+        let is_success = true;
+        let request_host_empty = true;
+        // In production: enters the `!should_process || request_host.is_empty()` block,
+        // then the PassThrough guard checks `is_success && !should_process` — host irrelevant.
+        let _enters_early_return = !should_process || request_host_empty;
```
prk-Jr: ⛏ **nitpick** — `let _enters_early_return = ...` is never used in an assertion; the `_` prefix suppresses the warning but the value is dead. A plain comment would communicate the same intent more clearly:

```rust
// In production this enters the early-return block: !should_process || request_host_empty
```

prk-Jr (changes requested):
## Summary

Phase 4 is mechanically correct: the `PassThrough` variant cleanly separates binary pass-through from streaming/buffered paths, `Content-Length` is preserved via `send_to_client()`, the 204 exclusion is properly handled, and the `once_cell` → `LazyLock` migration reduces the dependency surface. Two items need to be addressed before merge.

## Blocking

### 🔧 wrench

- **Stale PR title**: The title says "via `io::copy`" but the final implementation uses `response.set_body(body)` + `send_to_client()`. Commit `3aa88fe` replaced the `io::copy` approach specifically to preserve `Content-Length`. The title should read something like "Phase 4: Pass-through binary responses with Content-Length preservation via `send_to_client`".

## Non-blocking

### 🤔 thinking

- **Empty-host + non-processable 2xx behavioral change is undocumented**: Before Phase 4, empty-host requests always went `Buffered`. Now non-processable 2xx responses with an empty host take the `PassThrough` path instead. Intentional and correct (URL rewriting is skipped either way), but not mentioned anywhere in the PR description. Worth a one-liner: "Non-processable 2xx with empty request host now PassThrough instead of Buffered — URL rewriting is skipped either way."

### 📝 note

- **Content-Length fix in buffered path not mentioned**: `publisher.rs:449` — Phase 3 left the buffered path calling `response.remove_header(CONTENT_LENGTH)` after processing; this PR correctly sets it to the processed output length. The fix is right but it changes observable response headers and isn't mentioned in the description or commit messages.

### 👍 praise

- **`once_cell` → `std::sync::LazyLock` migration** (`datadome.rs`, `google_tag_manager.rs`, `rsc.rs`, `shared.rs`): Correct across all four sites, reduces the dependency surface, and all `Regex::new` `expect()` messages follow the `"should ..."` convention.
- **204 No Content exclusion**: RFC 9110 §15.3.5 prohibits a message body on 204 — the guard placement before `take_body()` and the dedicated test both make this correct and visible.

### 🌱 seedling

- **No end-to-end test for the PassThrough path**: The gate-logic tests are good but none of them call `handle_publisher_request()` with a backend returning a binary content type. An integration-level test exercising the full path would catch regressions in the gate condition itself. Follow-up suggestion — does not block merge.

## CI Status

- Integration tests: PASS
- fmt / clippy / cargo test: NOT RUN (CI gates only trigger on PRs targeting `main`; expected for phased branching strategy)

ChristianPavilonis (commented):
## Review Summary

2 findings (1× P2, 1× P3) — both are suggestions for improvement, no blockers.

crates/trusted-server-core/src/publisher.rs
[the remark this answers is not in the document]
aram356: Fixed — reordered to: set_body, finalize_response, send_to_client.

aram356: Fixed — reordered to: set_body, finalize_response, send_to_client.

crates/trusted-server-core/src/publisher.rs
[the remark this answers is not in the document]
aram356: Thanks!

aram356: Thanks!

crates/trusted-server-core/src/publisher.rs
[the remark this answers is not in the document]
aram356: Thanks!

aram356: Thanks!

crates/trusted-server-core/src/publisher.rs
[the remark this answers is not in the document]
aram356: The pass-through gate tests use named variables (is_success, should_process) matching the production code variables. The formula is simple enough that extracting a function adds indirection without preventing drift — the real gate is the match arm in handle_publisher_request, not a standalone function.

aram356: Same as above — keeping inline with named variables.

crates/trusted-server-core/src/publisher.rs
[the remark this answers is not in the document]
aram356: Fixed — removed dead variable, replaced with comment.

aram356: Fixed — removed dead variable, replaced with comment.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -183,6 +183,14 @@ async fn route_request(
                     // Response already sent via stream_to_client()
                     return None;
                 }
+                Ok(PublisherResponse::PassThrough { mut response, body }) => {
```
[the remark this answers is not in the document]
aram356: Valid point. The unit tests cover the gate logic and body reattachment. A full adapter integration test would require Viceroy serving binary content. Noting as a follow-up for the integration test suite.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -243,11 +258,13 @@ pub fn stream_publisher_body<W: Write>(
...
 /// Proxies requests to the publisher's origin server.
 ///
-/// Returns a [`PublisherResponse`] indicating whether the response can be
-/// streamed or must be sent buffered. The streaming path is chosen when:
-/// - The backend returns a 2xx status
-/// - The response has a processable content type
-/// - No HTML post-processors are registered (the streaming gate)
+/// Returns a [`PublisherResponse`] indicating how the response should be sent:
+/// - [`PassThrough`](PublisherResponse::PassThrough) — 2xx non-processable content
+///   (images, fonts, video). Body streamed directly via `io::copy`.
```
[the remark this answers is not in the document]
aram356: Fixed — function doc now describes the actual flow: body reattached unmodified for send_to_client(). io::copy reference removed.

prk-Jr (commented):
## Summary

Adds `PublisherResponse::PassThrough` for non-processable 2xx responses (images, fonts, video), streaming them directly to the client via `send_to_client()` instead of buffering in WASM memory. The implementation is correct: `Content-Length` is preserved, 204 No Content is properly excluded, and EC/consent headers are still applied. The 51% TTFB improvement is well-documented with real staging measurements.

## Non-blocking

### 🤔 thinking

- **`PassThrough` doc implies `finalize_response()` is the immediate caller's obligation** — but in `main.rs` it's called by the outer `route_request` path, not the `PassThrough` arm itself. Current code is correct; the doc could mislead a future adapter implementor. (`publisher.rs:277`)

### ♻️ refactor

- **Redundant `get_status()` call** — `is_success: bool` is captured at line 455, then `response.get_status()` is called again at line 470 to produce `status`. The inner `status.is_success()` duplicates `is_success`. Hoist `let status = response.get_status()` before the outer `if` and replace `let is_success = ...` with `let is_success = status.is_success()`. (`publisher.rs:470`)

### 📝 note

- **No debug log when returning `PassThrough`** — the block logs `"Skipping response processing..."` at the top but has no log at the `PassThrough` return. A `log::debug!("Pass-through: binary response, Content-Type: {}, status: {}")` would help with production tracing. (`publisher.rs:473`)
- **Gate tests verify boolean arithmetic, not `handle_publisher_request()` directly** — consistent with the existing `streaming_gate_*` pattern and an accepted trade-off given the Fastly runtime can't be mocked. Noted for awareness. (`publisher.rs:766`)

### ⛏ nitpick

- **205 Reset Content not excluded** — RFC 9110 §15.3.6 says 205 MUST NOT include message body content. Adding `status != StatusCode::RESET_CONTENT` alongside the 204 guard would be spec-complete, low risk. (`publisher.rs:471`)

### 👍 praise

- **`send_to_client()` over `stream_to_client()`** — correct call; `stream_to_client()` forces chunked encoding which would corrupt `Content-Length` for binary assets. (`main.rs:247`)
- **204 No Content excluded correctly** — HTTP spec prohibits a body in 204; tested and documented. (`publisher.rs:471`)
- **`apply_ec_headers` called before the PassThrough gate** — EC/consent headers are correctly applied to pass-through responses, consistent with existing buffered behaviour.
- **Staging performance results** — 51% TTFB, 28% FCP improvement with methodology documented. Exemplary.

## CI Status

- integration tests: PASS

crates/trusted-server-core/src/publisher.rs
```diff
@@ -443,6 +460,18 @@ pub fn handle_publisher_request(
...
             response.get_status(),
         );
+
+        // Stream non-processable 2xx responses directly to avoid buffering
+        // large binaries (images, fonts, video) in memory.
+        // Content-Length is preserved — the body is unmodified, so the
+        // browser knows the exact size for progress/layout.
+        // Exclude 204 No Content — it must not have a message body.
+        let status = response.get_status();
+        if status.is_success() && status != StatusCode::NO_CONTENT && !should_process {
```
prk-Jr: ♻️ **refactor** — Redundant `get_status()` call

`is_success: bool` is captured at line 455 but `response.get_status()` is called again here. The inner `status.is_success()` re-checks what `is_success` already holds.

**Suggested fix:** hoist `status` before the outer `if` block:
```rust
let status = response.get_status();
let is_success = status.is_success();

if !should_process || request_host.is_empty() || !is_success {
    // ...
    if is_success && status != StatusCode::NO_CONTENT && !should_process {
```
Eliminates the second `get_status()` call and makes the guard conditions read symmetrically.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -274,6 +274,21 @@ pub enum PublisherResponse {
         /// Parameters for `process_response_streaming`.
         params: OwnedProcessResponseParams,
     },
+    /// Non-processable 2xx response (images, fonts, video). The caller must:
```
prk-Jr: 🤔 **thinking** — Doc implies `finalize_response()` is the immediate caller's obligation, but it happens implicitly in the outer path

The variant doc lists:
```
/// 2. Call `finalize_response()` on the response
```

In `main.rs`, the `PassThrough` arm only does `response.set_body(body); Ok(response)`. `finalize_response()` is called by the general `route_request` control flow — not explicitly in the `PassThrough` branch. The current code is correct, but the doc reads as a per-arm obligation.

A future adapter (Axum, Cloudflare) implementing its own `route_request` could interpret this as calling `finalize_response()` in the match arm, then have it called again by the outer path. Consider updating to:
```
/// The adapter must call `response.set_body(body)` before sending.
/// `finalize_response()` and `send_to_client()` are the adapter's
/// responsibility at the outer response-dispatch level.
```

crates/trusted-server-core/src/publisher.rs
```diff
@@ -443,6 +460,18 @@ pub fn handle_publisher_request(
...
         );
+
+        // Stream non-processable 2xx responses directly to avoid buffering
+        // large binaries (images, fonts, video) in memory.
+        // Content-Length is preserved — the body is unmodified, so the
+        // browser knows the exact size for progress/layout.
+        // Exclude 204 No Content — it must not have a message body.
+        let status = response.get_status();
+        if status.is_success() && status != StatusCode::NO_CONTENT && !should_process {
+            let body = response.take_body();
```
prk-Jr, line 481: ⛏ **nitpick** — 205 Reset Content not excluded alongside 204

RFC 9110 §15.3.6: "A server MUST NOT generate content in a 205 response." If a misbehaving origin sends 205 with `Content-Length`, `PassThrough` would relay it. Low risk in practice, but spec-complete would be:

```rust
if is_success
    && status != StatusCode::NO_CONTENT      // 204
    && status != StatusCode::RESET_CONTENT   // 205 — RFC 9110 §15.3.6
    && !should_process
{
```

prk-Jr, line 481: 👍 **praise** — 204 No Content exclusion is correct and well-reasoned

HTTP spec prohibits a message body in 204 responses. Excluding it from `PassThrough` (which reattaches a body and calls `send_to_client()`) is correct. Tested and documented.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -443,6 +460,18 @@ pub fn handle_publisher_request(
...
+            return Ok(PublisherResponse::PassThrough { response, body });
+        }
```
prk-Jr, line 483: 📝 **note** — No debug log before returning `PassThrough`

The block logs `"Skipping response processing - should_process: ..., status: ..."` at the top, but there's no log when actually returning `PassThrough`. Production tracing would benefit from:

```rust
log::debug!(
    "Pass-through: binary response, Content-Type: {}, status: {}",
    content_type, status
);
return Ok(PublisherResponse::PassThrough { response, body });
```

crates/trusted-server-core/src/publisher.rs
```diff
@@ -733,6 +762,104 @@ mod tests {
         );
     }
 
+    #[test]
+    fn pass_through_gate_streams_non_processable_2xx() {
```
prk-Jr, line 776: 📝 **note** — Gate tests verify boolean arithmetic, not `handle_publisher_request()` directly

This matches the existing `streaming_gate_*` pattern, so it's consistent. All five `pass_through_gate_*` tests verify the formula `is_success && !should_process` in isolation — a typo in the actual gate condition would pass them all undetected.

`pass_through_preserves_body_and_content_length` is more meaningful since it exercises real Fastly API behaviour. Given the runtime can't be mocked in unit tests, this is a known trade-off — not a blocker.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -238,6 +238,14 @@ async fn route_request(
                             // Response already sent via stream_to_client()
                             return None;
                         }
+                        Ok(PublisherResponse::PassThrough { mut response, body }) => {
+                            // Binary pass-through: reattach body and send via send_to_client().
+                            // This preserves Content-Length and avoids chunked encoding overhead.
+                            // Fastly streams the body from its internal buffer — no WASM
+                            // memory buffering occurs.
+                            response.set_body(body);
+                            Ok(response)
```
prk-Jr, line 247: 👍 **praise** — Correct choice of `send_to_client()` over `stream_to_client()`

`stream_to_client()` forces chunked `Transfer-Encoding` which would conflict with the origin's `Content-Length` header, causing browsers to see a malformed response for binary assets. `send_to_client()` lets Fastly relay the body from its internal buffer with the original headers intact. Clean and correct.

prk-Jr:
⛏ **nitpick** — PR title references `io::copy` but the implementation uses `send_to_client()`

The title reads: *Phase 4: Stream binary pass-through responses via io::copy*

The first commit used `io::copy + stream_to_client()`, but that was superseded by commit `ef576a19` ("Preserve Content-Length for PassThrough by using send_to_client"). The final implementation has no `io::copy` anywhere — `send_to_client()` is the right description.

Suggested title: **Phase 4: Stream binary pass-through responses via send_to_client()**

aram356:
Addressed in 6ed2b5e0:

- **PR title** — renamed to "Phase 4: Stream binary pass-through responses via send_to_client()" (no more `io::copy` reference).
- **Redundant `get_status()` call** (publisher.rs:454/470) — hoisted `let status = response.get_status()` once and derived `is_success` from it; the inner block now reuses both.
- **PassThrough doc** (publisher.rs:277) — reworded to make clear that `finalize_response()` and `send_to_client()` are the outer dispatcher's responsibility, not the per-arm caller's.
- **205 Reset Content** (publisher.rs:471) — added `status != StatusCode::RESET_CONTENT` alongside the 204 guard with an RFC 9110 §15.3.6 reference; new `pass_through_gate_excludes_205_reset_content` test covers it.
- **Debug log** (publisher.rs:473) — added `log::debug!("Pass-through binary response - Content-Type: '{}', status: {}", ...)` before the PassThrough return.
- **PR description** — added a "Behavioral changes worth calling out" section covering (a) empty-host + non-processable 2xx now `PassThrough`, and (b) buffered path now sets `Content-Length` to processed length instead of removing it.

The end-to-end adapter integration test (🌱 seedling / P2 follow-up) is not in this PR — would need a Viceroy harness for binary content. Tracking as follow-up.

Verification: `cargo test --workspace` (819 passing), `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean, `cargo fmt --all -- --check` clean, release wasm build succeeds.

PR #605, opened 2026-04-01
JS Asset Proxy engineering spec

### LINKED ISSUE #604: Create Spec for JS Asset Proxy
### Description

Create Spec for JSAP

jevansnyc:
## Summary

Engineering spec for the JS Asset Proxy feature — brings third-party JS delivery inside the publisher's operational boundary by proxying vendor assets through Fastly Compute.

Closes #604

## Docs added

- `docs/js-asset-proxy-prd.md` — original product requirements
- `docs/superpowers/specs/2026-04-01-js-asset-proxy-design.md` — engineering spec covering config schema, KV data model, 3-phase implementation plan, caching layers, degraded behavior, and Fastly KV store provisioning

## Test plan

- [ ] Review proxy spec phases and verify scope is implementable as described
- [ ] Confirm Fastly KV store provisioning steps are accurate for our service setup
- [ ] Verify PRD open question resolutions match team understanding


docs/js-asset-proxy-prd.md
```diff
@@ -0,0 +1,356 @@
...
+slug = "vendor-static"
+path = "/raven-static/*"       # wildcard — version segment forwarded to origin
+origin_url = "https://raven-static.vendor.io/prod/*"
+ttl_sec = 3600
+inject_in_head = false
+```
+
+### 9.1 Backend configuration
+
+Each distinct origin hostname requires a declared backend in `trusted-server.toml` (Fastly or similar Compute requires backends to be declared at deploy time):
```
aram356: ♻️ **refactor** — The engineering spec explicitly resolves `[backends.*]` as "Removed" (dynamic backends via `BackendConfig::from_url()` instead). This entire section is superseded. Either strike it or annotate as superseded by the eng spec.

docs/js-asset-proxy-prd.md
```diff
@@ -0,0 +1,356 @@
...
+| `/raven/golfmain.js` | Vendor loader | `https://raven-edge.vendor.io/raven/golf-main-L1Zrx/library.js` |
+| `/raven-static/*/raven.js` | Vendor bundle (versioned) | `https://raven-static.vendor.io/prod/*/raven.js` |
+
+### 5.2 Route matching
+
+TS matches inbound paths against the asset registry in `trusted-server.toml`. Wildcard segments (`*`) in the path are forwarded verbatim to the origin URL. Paths not matching any registered asset return `404`.
+
+### 5.3 Why opaque and not content-hash
+
+Content-hash URLs (e.g., `tsjs-core.abc123.js`) are appropriate for assets TS generates itself, where the hash is computed at build time. Third-party JS has content that changes independently — using a publisher-token-derived slug decouples the URL from the content version. The ETag mechanism handles freshness.
```
aram356: ⛏ **nitpick** — The TS bundle does not actually use content-hash filenames. Current path is `/static/tsjs=tsjs-unified.min.js` with ETag-based caching via `serve_static_with_etag()`. The comparison is slightly misleading about current behavior.

docs/superpowers/specs/2026-04-01-js-asset-proxy-design.md
```diff
@@ -0,0 +1,302 @@
...
+    key = "js-asset:test-slug:test-asset"
+    data = "{\"v\":1,\"origin_url\":\"https://example.com/test.js\",\"content_type\":\"application/javascript\",\"etag\":\"\\\"test\\\"\",\"fetched_at\":0,\"expires_at\":9999999999,\"asset_slug\":\"test-slug:test-asset\"}"
+```
+
+**Files:**
+- `js-assets.toml` — new, checked in with placeholder entries
+- `fastly.toml` — add `[[local_server.kv_stores.js_asset_store]]` entry
+- `crates/trusted-server-core/src/settings.rs` — add `KvStores`, `JsAsset`, extend `Settings`
+- `crates/trusted-server-adapter-fastly/src/main.rs` — add asset path routing before publisher fallback
+- `crates/trusted-server-adapter-fastly/build.rs` — include `js-assets.toml`
```
aram356: ❓ **question** — This file does not exist. The `build.rs` that loads `trusted-server.toml` is at `crates/trusted-server-core/build.rs`. The adapter crate has no `build.rs`. Implementers will look in the wrong place.

**Fix:** `crates/trusted-server-core/build.rs`

docs/superpowers/specs/2026-04-01-js-asset-proxy-design.md
```diff
@@ -0,0 +1,302 @@
...
+- `crates/trusted-server-core/src/js_assets/mod.rs` (new module) — route matcher, KV key format, metadata types
+
+### Phase 2: Fetch + cache pipeline
+
+Core value. Requires populated `js-assets.toml` from the Auditor (see delivery order).
+
+**Scope:**
+- `handle_js_asset()` handler wired into route dispatch
+- Cold path: `BackendConfig::from_url()` → origin fetch → compress body to brotli → write compressed bytes + metadata to KV
```
aram356: ❓ **question** — Two gaps in the origin fetch strategy:

1. **`certificate_check`**: `BackendConfig::from_url()` requires a `certificate_check: bool` parameter. The existing `settings.proxy.certificate_check` applies only to publisher origin proxy. Where does the JS asset proxy get this setting?

2. **`Accept-Encoding` / compression**: The spec says "compress body to brotli" but does not address what happens when the origin already returns compressed content (`Content-Encoding: br` or `gzip`). The PRD (line 158) says "matching the origin Content-Encoding", which contradicts always storing as brotli.

**Fix:** Specify `Accept-Encoding: br` on origin fetches to get brotli directly. Document behavior when origin returns a different encoding. Add a note about TLS verification policy.

docs/superpowers/specs/2026-04-01-js-asset-proxy-design.md
```diff
@@ -0,0 +1,302 @@
...
+```
+Content-Type: application/javascript
+Content-Encoding: br
+Cache-Control: public, max-age=3600
+ETag: {stored_etag}
+```
+
+**Caching layers — all aligned at 1h:**
+- **KV store TTL (1h):** how long TS keeps its cached copy before issuing a conditional GET to the vendor origin
+- **Fastly CDN edge cache (s-maxage=3600 via `public, max-age=3600`):** Fastly serves the asset from edge POP without hitting the TS WASM at all
```
aram356: ❓ **question** — `Cache-Control: public, max-age=3600` enables Fastly HTTP edge cache in front of KV, creating a dual-cache that can diverge:

1. No `Vary` header specified — risk if the service ever serves multiple publishers
2. No `Surrogate-Key` for targeted edge purging
3. A KV metadata-only update (304 path) does not purge the Fastly edge cache, which continues serving the previous version until its own max-age expires

Is the dual-cache intentional? If KV should be the sole cache layer, consider `s-maxage=0` or `Cache-Control: private, max-age=3600`. If dual-cache is intended, document the implications and add `Surrogate-Key` for purging.

docs/superpowers/specs/2026-04-01-js-asset-proxy-design.md
```diff
@@ -0,0 +1,302 @@
...
+| Versioned wildcard KV growth | Wildcard entries default `ttl_sec = 1800` (30 min) if not set. Eviction job v2. |
+| Vendor CDN URL configurability | Vendors must make CDN URLs configurable (Option 2). Per-vendor launch dependency. String rewriting deferred. |
+| Backend declarations (PRD §9.1) | **Removed.** `BackendConfig::from_url()` in `crates/trusted-server-core/src/backend.rs` creates dynamic backends at runtime. No `[backends.*]` toml entries needed. |
+
+---
+
+## Config Contract: `js-assets.toml` (new file)
+
+Separate from `trusted-server.toml` to avoid bloat. Loaded via the same `include_bytes!` / `build.rs` pattern as the main config.
+
```
aram356: 🤔 **thinking** — The current `build.rs` is tightly coupled to a single TOML file (`trusted-server.toml` -> env merge -> `target/trusted-server-out.toml` -> `include_bytes!`). Adding `js-assets.toml` via the same pattern requires either extending `build.rs` to merge both files into one output, or a separate `include_bytes!` with its own loading path. Worth specifying which approach so Phase 1 does not start with an architectural decision.

docs/superpowers/specs/2026-04-01-js-asset-proxy-design.md
```diff
@@ -0,0 +1,302 @@
...
+Add to `crates/trusted-server-core/src/settings.rs`:
+
+```rust
+#[derive(Debug, Default, Clone, Deserialize, Serialize)]
+pub struct KvStores {
+    #[serde(default)]
+    pub js_asset_store: String,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, Validate)]
```
aram356: 🤔 **thinking** — `JsAsset` derives `Validate` but no `#[validate(...)]` attributes are shown. The `slug`, `path`, and `origin_url` fields would benefit from validation annotations (URL format, path prefix, slug pattern). Without them the `Validate` derive is a no-op and malformed config entries would only fail at runtime.

docs/superpowers/specs/2026-04-01-js-asset-auditor-design.md
```diff
@@ -0,0 +1,216 @@
...
+
+**Rationale:** Fully opaque and hash-derived — no human naming required, no ambiguity for cryptic vendor filenames. The KV metadata (`origin_url`, `content_type`, `asset_slug`) serves as the lookup table. Operators can query `js-asset:{slug}` in the KV store to retrieve full provenance. The terminal summary also prints slug → origin_url at generation time.
+
+**Important:** This algorithm must produce identical output to the Proxy's KV key derivation. Engineering should implement this as a shared utility (e.g., a small JS/TS helper in the skill, or a standalone `scripts/` utility) rather than duplicating the logic.
+
+### Wildcard detection
+
+Path segments matching either pattern are replaced with `*`:
+- Semver: `\d+\.\d+[\.\d-]*` (e.g., `1.19.8-hcskhn`)
+- Hash-like: `[a-f0-9]{6,}` or `[A-Za-z0-9]{8,}` between path separators
```
aram356: 🌱 **seedling** — The hash-like pattern `[A-Za-z0-9]{8,}` would false-positive on common vendor path segments like `analytics` (9 chars), `openrtb2` (8 chars), `prebidjs` (8 chars). Worth noting as a known limitation requiring operator review of wildcard substitutions.

aram356 (changes requested):
## Summary

Well-structured engineering spec for JS asset proxying through Fastly Compute. The phased approach, KV metadata/body separation, and Auditor-as-skill design are strong. Main concerns are cross-document inconsistencies (PRD vs eng spec) and missing details needed for correct implementation (compression strategy, cache invalidation, TLS verification).

## Blocking

### ❓ question

- **PRD and eng spec disagree on config file location**: PRD §9 defines config as new sections in `trusted-server.toml` with `[backends.*]`; eng spec uses a separate `js-assets.toml` with dynamic backends. Fundamental contradiction — update PRD to match eng spec or annotate it as superseded.
- **Slug derivation algorithm differs**: PRD (line 100) derives from publisher ID; eng spec (line 126) derives from `sha256(domain + origin_url)`. Different algorithms produce different slugs. PRD examples are inconsistent with the eng spec hash-based approach.
- **Stale-on-error inconsistency**: PRD (line 201) returns 502 for origin failures, with stale fallback only for expired KV entries. Eng spec (line 167) makes stale fallback conditional on KV existence for all origin failures, but doesn't cover non-200/304 responses. Align and make explicit for all failure cases.

See inline comments for additional blocking items on specific lines.

## Non-blocking

### 🤔 thinking

- **No timeout for origin fetches**: `BackendConfig::from_url()` uses the default 15s first-byte timeout. PRD targets p99 < 300ms. Consider documenting a configurable `fetch_timeout_ms`.

### 🌱 seedling

- **No CSP guidance**: None of the three docs mention `Content-Security-Policy` implications. Publishers with strict `script-src` directives need to allow the new asset domain. Worth documenting as a deployment consideration.

### ⛏ nitpick

- **Inconsistent terminology**: "TS", "TS Lite", "Trusted Server" used interchangeably. "TS Lite" appears in PRD but is never defined and absent from eng spec.

### 👍 praise

- **Phased approach**: Three phases with clear file lists, each independently testable, logical dependency order. Auditor delivery timed between Phase 1 and 2 is particularly thoughtful.
- **KV metadata/body separation**: Reading metadata first to avoid WASM heap cost shows excellent understanding of platform constraints.
- **Auditor as Claude Code skill**: Pragmatic choice — avoids bloating the production binary, leverages existing Chrome DevTools MCP, diff mode with commented-out TOML is a nice safety mechanism.

## CI Status

- cargo fmt: PASS
- cargo test: PASS
- vitest: PASS
- format-typescript: PASS
- format-docs: **FAIL** — new markdown files need formatting (`cd docs && npm run format:write`)
- CodeQL: PASS
- integration tests: PASS

ChristianPavilonis (changes requested):
The spec suggests that we cache compressed JS into the KV store, which sounds expensive, if not impossible on platforms with restrictions on the size of the value.

We should just simply use our existing patterns for how we proxy with proper cache headers in existing integrations.

This also mentions TS Lite, so I'm not sure that should be within the scope of this initially.

aram356:
Closing in place of #742

PR #623, opened 2026-04-08
Migrate utility layer to HTTP types (PR 11)

### LINKED ISSUE #492: Utility layer type migration + compat adapter
## Description

Migrate utility modules (cookies, GDPR, HTTP helpers, synthetic) from `fastly::Request/Response` to `http::Request/Response`. Create `compat.rs` compatibility adapter to bridge remaining Fastly helper APIs.

## Done when

- Utilities use `http` types
- `compat.rs` created
- Per-PR gates pass

## Blocked by

PRs 3-10

## References

- `docs/internal/EDGEZERO_MIGRATION.md` — Phase 2, PR 11
- Part of #480

prk-Jr:
## Summary

- Migrate the PR11 utility layer off direct `fastly::Request`/`fastly::Response` usage so core helpers can operate on `http::{Request, Response}` and `edgezero_core::Body`.
- Add a temporary `compat` bridge at Fastly boundaries so handlers and integrations can keep working while later migration PRs move the remaining call stack.
- Lock in the migration with focused compat tests and a guard test that prevents the migrated utility modules from drifting back to Fastly request/response types.

## Changes

| File | Change |
| ---- | ------ |
| `Cargo.toml` | Add the workspace `mime` dependency used by migrated HTTP response helpers. |
| `Cargo.lock` | Record the new `mime` dependency. |
| `crates/trusted-server-adapter-fastly/src/main.rs` | Route forwarded-header sanitization and basic-auth response conversion through the new compat boundary. |
| `crates/trusted-server-core/Cargo.toml` | Add the core crate's `mime` workspace dependency. |
| `crates/trusted-server-core/src/auction/endpoints.rs` | Convert auction utility calls to use the HTTP request compat bridge for EC ID and consent handling. |
| `crates/trusted-server-core/src/auth.rs` | Migrate basic-auth enforcement to `http::Request/Response` and update tests to HTTP builders. |
| `crates/trusted-server-core/src/compat.rs` | Add Fastly-to-HTTP request/response conversions plus temporary Fastly boundary shims for headers, cookies, and EC cookie handling. |
| `crates/trusted-server-core/src/consent/extraction.rs` | Migrate consent signal extraction to `http::Request<EdgeBody>`. |
| `crates/trusted-server-core/src/consent/mod.rs` | Move consent pipeline input types and tests onto HTTP request types. |
| `crates/trusted-server-core/src/cookies.rs` | Migrate cookie parsing/forwarding and EC cookie response helpers to HTTP request/response types; document and test target Cookie append semantics. |
| `crates/trusted-server-core/src/edge_cookie.rs` | Add HTTP request EC ID helpers so migrated callers can reuse an existing HTTP request without rebuilding Fastly headers. |
| `crates/trusted-server-core/src/http_util.rs` | Migrate request/response helpers to HTTP types, preserve duplicate headers, and keep request-info logic on `ClientInfo`. |
| `crates/trusted-server-core/src/integrations/lockr.rs` | Use Fastly compat shims for header/cookie forwarding at the integration boundary. |
| `crates/trusted-server-core/src/integrations/permutive.rs` | Use Fastly compat shims for custom-header forwarding at the integration boundary. |
| `crates/trusted-server-core/src/integrations/prebid.rs` | Bridge request-info construction and cookie forwarding through compat conversions. |
| `crates/trusted-server-core/src/integrations/registry.rs` | Use compat conversions for EC ID generation and EC cookie response handling. |
| `crates/trusted-server-core/src/lib.rs` | Export the compat module and add migration guard tests. |
| `crates/trusted-server-core/src/migration_guards.rs` | Add a regression test preventing migrated utility modules from reintroducing direct Fastly request/response types. |
| `crates/trusted-server-core/src/publisher.rs` | Route TSJS serving, request-info extraction, consent handling, and EC cookie writes through compat conversions. |
| `crates/trusted-server-core/src/request_signing/endpoints.rs` | Switch JSON content-type constants to `mime::APPLICATION_JSON`. |

## Closes

Closes #492

## Test plan

- [x] Other: focused Rust verification with `cargo test --package trusted-server-core compat -- --nocapture`, `cargo test --package trusted-server-core http_util -- --nocapture`, `cargo test --package trusted-server-core request_signing -- --nocapture`, and `cargo test --package trusted-server-core migration_guards -- --nocapture`
- [x] Other: local `cd crates/js/lib && npx vitest run` currently fails before test execution with `ERR_REQUIRE_ESM` in `html-encoding-sniffer` -> `@exodus/bytes/encoding-lite.js`; leaving CI to capture the current JS environment issue.

## Hardening note

This PR does not add any new config-derived regex or pattern compilation paths. Basic auth still surfaces invalid enabled handler regex configuration as an error rather than panicking, covered by `auth::tests::returns_error_for_invalid_handler_regex_without_panicking` alongside the existing settings startup validation tests.

## Checklist

- [x] Uses project logging macros (not `println!`)
- [x] New code has tests


crates/trusted-server-core/src/compat.rs
```diff
@@ -0,0 +1,512 @@
...
+}
+
+/// Convert an `http::Request<EdgeBody>` into a `fastly::Request`.
+///
+/// # PR 15 removal target
+pub fn to_fastly_request(req: http::Request<EdgeBody>) -> fastly::Request {
+    let (parts, body) = req.into_parts();
+    let mut fastly_req = fastly::Request::new(parts.method, parts.uri.to_string());
+    for (name, value) in &parts.headers {
+        fastly_req.set_header(name.as_str(), value.as_bytes());
```
ChristianPavilonis: 🔧 **wrench** — `set_header` overwrites previous values for the same header name, dropping duplicates.

If an `http::Request` has multiple values for the same header (e.g., multiple `X-Custom` values), only the last one survives the conversion. `to_fastly_response` at line 109 correctly uses `append_header` — this should match.

**Fix**:
```rust
fastly_req.append_header(name.as_str(), value.as_bytes());
```

crates/trusted-server-core/src/http_util.rs
```diff
@@ -15,15 +15,11 @@ use crate::settings::Settings;
...
+    for (header_name, value) in from.headers() {
         let name_str = header_name.as_str();
-        if (name_str.starts_with("x-") || name_str.starts_with("X-"))
-            && !INTERNAL_HEADERS.contains(&name_str)
-        {
-            if let Some(value) = from.get_header(header_name) {
-                to.set_header(header_name, value);
-            }
+        if name_str.starts_with("x-") && !INTERNAL_HEADERS.contains(&name_str) {
+            to.headers_mut().insert(header_name.clone(), value.clone());
```
ChristianPavilonis: 🔧 **wrench** — `insert` replaces any prior value for the same header name, dropping duplicates.

This is a behavioral regression from the old Fastly-based `copy_custom_headers` which preserved duplicates via `append_header`. If the source request has multiple values for the same `X-*` header, only the last survives.

**Fix**:
```rust
to.headers_mut().append(header_name.clone(), value.clone());
```

ChristianPavilonis (changes requested):
## Summary

This PR cleanly migrates utility-layer functions (`auth`, `cookies`, `synthetic`, `http_util`, `consent/extraction`, `consent/mod`) from `fastly::Request`/`fastly::Response` to `http::Request`/`http::Response` with `EdgeBody`. The compat bridge pattern is sound and well-tested.

Two duplicate-header-dropping bugs need fixing before merge — see inline comments.

## Blocking

### 🔧 wrench

- **`to_fastly_request` drops duplicate headers**: Uses `set_header` instead of `append_header`, losing multi-valued headers during conversion (compat.rs:65)
- **`copy_custom_headers` drops duplicate X-headers**: Uses `insert` instead of `append`, a behavioral regression from the old Fastly-based version (http_util.rs:22)

## Non-blocking

### 🤔 thinking

- **Migration guard `strip_line_comments` is naive about string literals**: A Rust string literal like `"fastly::Request"` in a test assertion would trigger a false positive. Currently does not happen, but the approach is fragile. Consider adding a brief comment documenting the limitation. (migration_guards.rs)

### 🌱 seedling

- **`copy_custom_headers` in http_util.rs has no callers outside tests**: The integrations (lockr, permutive) now call `compat::copy_fastly_custom_headers` instead. Worth noting for PR 15 cleanup.

### ⛏ nitpick

- **Dead `"X-"` check in `copy_fastly_custom_headers`**: The check for `name_str.starts_with("X-")` in compat.rs:144 is dead code since Fastly normalizes header names to lowercase. The migrated `copy_custom_headers` in http_util.rs only checks `"x-"`. Not a correctness issue, but the asymmetry may confuse readers.
- **Temporary compat functions well-documented with removal targets**: Every public function in `compat.rs` has a `# PR 15 removal target` annotation — great practice for tracking temporary code.

### 👍 praise

- **Thorough compat test coverage**: 11 focused tests covering round-trip conversions, duplicate header preservation, header sanitization, cookie forwarding with consent stripping, and synthetic cookie lifecycle. Excellent scaffolding for temporary bridge code.
- **Migration guard test is a clever regression barrier**: The `include_str!` approach to scan source files for banned Fastly types provides a compile-time-like guarantee without proc macros.

### 📝 note

- **`mime` dependency is appropriate**: Replaces `fastly::mime::APPLICATION_JSON` references. Well-maintained, zero-dependency crate suitable for the use case.

## CI Status

- All checks: PASS

crates/trusted-server-core/src/http_util.rs
aram356: 🔧 **wrench** — `insert` silently drops duplicate headers

`headers_mut().insert(...)` replaces any existing value for the same header name. If the source request carries duplicate `x-custom` headers (e.g. `x-custom: first` and `x-custom: second`), only the last value survives.

The compat shim `copy_fastly_custom_headers` correctly uses `append_header` — and has a test (`copy_fastly_custom_headers_preserves_duplicate_values`) proving duplicates are preserved. This migrated version lacks that guarantee.

Currently no production code calls this version (all callers use the compat shim), but this is the version that will go live when PR 15 removes the compat layer.

**Fix:**
```rust
to.headers_mut().append(header_name.clone(), value.clone());
```

Also add a test mirroring `copy_fastly_custom_headers_preserves_duplicate_values`.

crates/trusted-server-core/src/compat.rs
aram356: 🔧 **wrench** — `set_header` drops duplicate headers; should be `append_header`

`set_header` replaces rather than appends, so duplicate headers in the source `http::Request` are collapsed to a single value. `to_fastly_response` on line 109 correctly uses `append_header` — this looks like an oversight.

Currently unused outside its own test, but it's a public API that future PRs may call.

**Fix:**
```rust
fastly_req.append_header(name.as_str(), value.as_bytes());
```

crates/trusted-server-core/src/compat.rs
```diff
@@ -0,0 +1,512 @@
...
+    }
+}
+
+/// Copy `X-*` custom headers between two `fastly::Request` values.
+///
+/// # PR 15 removal target
+pub fn copy_fastly_custom_headers(from: &fastly::Request, to: &mut fastly::Request) {
+    for (name, value) in from.get_headers() {
+        let name_str = name.as_str();
+        if (name_str.starts_with("x-") || name_str.starts_with("X-"))
```
aram356: ⛏ **nitpick** — Redundant case check

```rust
if (name_str.starts_with("x-") || name_str.starts_with("X-"))
```

Fastly's `HeaderName` is already case-normalized to lowercase, so the `"X-"` branch never fires. The migrated version in `http_util.rs` correctly uses only `"x-"`. Not worth changing in temporary compat code, just noting for awareness.

aram356 (changes requested):
## Summary

Migrates the utility layer (auth, cookies, synthetic, http_util, consent) off direct `fastly::Request`/`fastly::Response` to `http::{Request, Response}<EdgeBody>` with a temporary compat bridge. Well-structured migration with good test coverage and clear lifecycle annotations. Two header-handling bugs need fixing before merge.

## Blocking

### 🔧 wrench

- **`copy_custom_headers` drops duplicate headers**: uses `insert` instead of `append` in `http_util.rs:22`. The compat shim preserves duplicates; this migrated version doesn't. Will become a production regression when the compat layer is removed in PR 15.
- **`to_fastly_request` drops duplicate headers**: uses `set_header` instead of `append_header` in `compat.rs:65`. Inconsistent with `to_fastly_response` which correctly uses `append_header` on line 109.

## Non-blocking

### 🤔 thinking

- **Migration guard doesn't handle block comments**: `strip_line_comments` in `migration_guards.rs` only strips `//` line comments. A `/* fastly::Request */` block comment would cause a false positive, and a real regression hidden inside a block comment wouldn't be caught. Acceptable for a guard test (false positives are safe), just noting the limitation.

### 🌱 seedling

- **Add a `to_fastly_request` duplicate-header round-trip test**: after fixing the `append_header` issue, a round-trip test (`http::Request` → `fastly::Request` → `http::Request`) verifying duplicate header preservation would prevent future regressions.

### ⛏ nitpick

- **Redundant case check in `copy_fastly_custom_headers`**: `compat.rs:144` checks both `"x-"` and `"X-"` but Fastly's `HeaderName` is already case-normalized to lowercase. Not worth changing in temporary code.

## CI Status

- integration tests: PASS

aram356 (approved):
## Summary

Prior blocking findings (duplicate-header drops in `to_fastly_request` and `copy_custom_headers`) are both fixed in 28177610, with round-trip regression tests added. CI is green. Approving; non-blocking observations below.

## Non-blocking

### 🤔 thinking

- **`from_fastly_request_ref` silently drops body** (`crates/trusted-server-core/src/compat.rs:54`): docs note the body is empty, but the name doesn't signal it. Footgun if a future caller passes a POST expecting body access. Consider renaming to `from_fastly_headers_ref` or adding a `debug_assert!` when the source request has a non-empty body. All current callers only read headers, so not a bug today.
- **Double conversion on the auction hot path**: `auction/endpoints.rs:50` builds `http_req` once, but `auction/formats.rs:93` (`convert_tsjs_to_auction_request`) re-converts the same underlying `fastly::Request` again via `from_fastly_request_ref`. Each conversion iterates all headers — two copies per auction request. Acceptable for a temporary bridge (PR 15 removes it), but worth a comment referencing PR 15 so the duplication doesn't get frozen.
- **`forward_cookie_header` uses `insert`, not `append`** (`crates/trusted-server-core/src/cookies.rs:167,175,184`): fine for HTTP/1 where `Cookie` is a single semicolon-joined value, but HTTP/2 (RFC 7540 §8.1.2.5) allows splitting across multiple headers and only the first would be forwarded. Matches pre-migration behavior and Fastly concatenates HTTP/2 cookie headers upstream, so likely not a live bug — worth documenting the assumption.

### 🌱 seedling

- **Migration guard scope is "utility modules only"** (`crates/trusted-server-core/src/migration_guards.rs:27-37`): guards 6 files. Handler/integration files (`publisher.rs`, `proxy.rs`, `registry.rs`, `integrations/*.rs`, `auction/*.rs`) still use `fastly::Request` by design. PR 12–14 should extend the banned-patterns list as files are migrated.

### ⛏ nitpick

- **Dead case check in `copy_fastly_custom_headers`** (`crates/trusted-server-core/src/compat.rs:144`): checks both `"x-"` and `"X-"` though `fastly::Request` normalizes to lowercase. Not worth a commit in temporary code — trim on the next touch.

## CI Status

- integration tests: PASS

ChristianPavilonis (approved):

crates/trusted-server-core/src/compat.rs
```diff
@@ -0,0 +1,534 @@
...
+use fastly::http::header;
+
+use crate::constants::INTERNAL_HEADERS;
+use crate::http_util::SPOOFABLE_FORWARDED_HEADERS;
+
+fn build_http_request(req: &fastly::Request, body: EdgeBody) -> http::Request<EdgeBody> {
+    let uri: http::Uri = req
+        .get_url_str()
+        .parse()
+        .expect("should parse fastly request URL as URI");
```
aram356: 🔧 **wrench** — New edge-wide runtime panic surface on URL parsing.

`build_http_request` calls `.parse().expect("should parse fastly request URL as URI")` on every request that passes through the compat bridge. Fastly's internal URL handling (backed by `url::Url`) is more permissive than `http::Uri` (IPv6 zone identifiers, some host-character edge cases, etc.), so a URL Fastly accepts but `http::Uri` rejects now panics the entire edge handler.

This runs on every request via `route_request` → `enforce_basic_auth` (`adapter-fastly/src/main.rs:124`) before auth can even execute. Previously `enforce_basic_auth` used `req.get_path()` which does not re-parse, so this is a net regression in robustness at the very first step of request handling. Most callers don't actually use the parsed URI (they only read headers), so the blast radius is larger than the value delivered.

**Fix options**:

1. Return a 400 from the edge instead of panicking when the URL can't be represented as `http::Uri`.
2. Build the `http::Request` with a placeholder URI (e.g., `/`) when the source URL fails to parse, and let callers that actually need the URI read it from the original `fastly::Request`.
3. At minimum, convert the `expect` into a checked path so malformed URLs produce a logged error + 400, not a panic.

```rust
let uri: http::Uri = match req.get_url_str().parse() {
    Ok(uri) => uri,
    Err(_) => http::Uri::from_static("/"),
};
```

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -1008,7 +1009,8 @@ impl AuctionProvider for PrebidAuctionProvider {
             &context.settings.request_signing
         {
             if request_signing_config.enabled {
-                let request_info = RequestInfo::from_request(context.request, context.client_info);
+                let http_req = compat::from_fastly_headers_ref(context.request);
```
aram356: 🤔 **thinking** — Redundant compat conversion on the prebid hot path.

`request_bids` calls `compat::from_fastly_headers_ref(context.request)` here, and `to_openrtb` calls it again at `prebid.rs:713` for the same request in the same auction flow. Each conversion iterates every header and re-parses the URL. Pull the conversion up once in `request_bids` and thread the `http::Request` (or the derived `RequestInfo`) into `to_openrtb` via a parameter.

**Fix**:

```rust
let http_req = compat::from_fastly_headers_ref(context.request);
let request_info = RequestInfo::from_request(&http_req, context.client_info);
// ... pass request_info into to_openrtb and drop the second conversion at line 713
```

crates/trusted-server-core/src/compat.rs
```diff
@@ -0,0 +1,534 @@
...
+}
+
+/// Convert an owned `fastly::Request` into an `http::Request<EdgeBody>`.
+///
+/// # PR 15 removal target
+///
+/// # Panics
+///
+/// Panics if the Fastly request URL cannot be parsed as an `http::Uri`.
+pub fn from_fastly_request(mut req: fastly::Request) -> http::Request<EdgeBody> {
```
aram356: ♻️ **refactor** — `from_fastly_request`, `to_fastly_request`, and `from_fastly_response` are unused in this PR.

`from_fastly_request` (line 40), `to_fastly_request` (line 61), and `from_fastly_response` (line 90) are only referenced from their own unit tests — a crate-wide grep returns zero production callers in this PR. The module doc says "scaffolding created in PR 11 and scheduled for deletion in PR 15," but `CLAUDE.md` explicitly calls out: *"Don't design for hypothetical future requirements. No half-finished implementations either."*

Both `to_fastly_request` and `to_fastly_response` also silently drop the body on `EdgeBody::Stream` (lines 74-76, 118-120) with only a `warn!` log — that's a latent bug that's easier to justify one way or another when paired with a real caller.

**Recommendation**: ship these helpers in the PR that actually uses them (PR 12/13/14). Keep only `from_fastly_headers_ref`, `to_fastly_response`, `sanitize_fastly_forwarded_headers`, `copy_fastly_custom_headers`, `forward_fastly_cookie_header`, `set_fastly_synthetic_cookie`, and `expire_fastly_synthetic_cookie` — the functions with production callers in this PR.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -89,7 +90,10 @@ pub fn convert_tsjs_to_auction_request(
     geo: Option<GeoInfo>,
 ) -> Result<AuctionRequest, Report<TrustedServerError>> {
     let synthetic_id = synthetic_id.to_owned();
-    let fresh_id = generate_synthetic_id(settings, services, req).change_context(
+    // TODO(PR 15): Remove this conversion once the auction hot path is migrated to http::Request.
+    // endpoints.rs already converts this, but we do it again here as a temporary bridge.
+    let http_req = compat::from_fastly_headers_ref(req);
```
aram356: 📌 **out of scope** — Duplicate conversion acknowledged in the TODO.

`endpoints.rs:50` already converts the same request before calling this path, but `convert_tsjs_to_auction_request` accepts `&fastly::Request` and re-converts here. Accepted cost of incremental migration — worth a follow-up when the function signature can take `&http::Request<EdgeBody>`.

crates/trusted-server-core/src/compat.rs
```diff
@@ -0,0 +1,534 @@
...
+
+    fastly_resp
+}
+
+/// Sanitize forwarded headers on a `fastly::Request`.
+///
+/// # PR 15 removal target
+pub fn sanitize_fastly_forwarded_headers(req: &mut fastly::Request) {
+    for &name in SPOOFABLE_FORWARDED_HEADERS {
+        if req.get_header(name).is_some() {
```
aram356, line 86: 🌱 **seedling** — Redundant `get_header` before `remove_header`.

`remove_header` is idempotent, so the `get_header(name).is_some()` check exists only to emit the debug log. That's fine, but worth noting in case this module is later consolidated with the http-native `sanitize_forwarded_headers` — they should behave the same way.

crates/trusted-server-core/src/cookies.rs
```diff
@@ -146,26 +148,39 @@ pub fn strip_cookies(cookie_header: &str, cookie_names: &[&str]) -> String {
...
-            if !stripped.is_empty() {
-                to.set_header(header::COOKIE, &stripped);
+        match cookie_value.to_str() {
+            Ok(s) => {
+                let stripped = strip_cookies(s, CONSENT_COOKIE_NAMES);
+                if !stripped.is_empty() {
+                    to.headers_mut().append(
+                        header::COOKIE,
+                        http::HeaderValue::from_str(&stripped)
+                            .expect("should build stripped Cookie header value"),
```
aram356: ⛏ **nitpick** — Consider replacing `expect` with a graceful drop.

The `HeaderValue::from_str(&stripped).expect(...)` panic fires only if `strip_cookies` produces an invalid header value from an already-valid one, which shouldn't happen in practice — but a `try_from` + skip-on-error keeps failure local to this forwarding function instead of taking down the request. Same applies to the panic documented in the function-level `# Panics` doc at lines 152-155.

aram356 (changes requested):
## Summary

PR introduces `compat.rs` as a temporary Fastly↔http bridge and migrates six utility modules (`auth`, `cookies`, `synthetic`, `http_util`, `consent/extraction`, `consent/mod`) off `fastly::Request`/`fastly::Response`. Prior approvals (2026-04-14) predate the latest commit `ae402ff` (2026-04-15), which renamed `from_fastly_request_ref` → `from_fastly_headers_ref`, dropped a redundant case check in `copy_fastly_custom_headers`, and switched `forward_cookie_header` to `get_all` + `append`. The round-2 fixes for duplicate-header preservation are clean, but a few concerns remain — notably a new runtime-panic surface at the edge and unused compat helpers that arrived before their callers.

## Blocking

### 🔧 wrench

- **New edge-wide panic surface on URL parsing**: `compat::build_http_request` calls `.parse().expect(...)` on every bridged request. `http::Uri` is stricter than Fastly's internal `url::Url`, so a URL Fastly accepts but `http::Uri` rejects panics the entire edge handler before auth can run. Previously `enforce_basic_auth` used `req.get_path()` with no re-parse. (`crates/trusted-server-core/src/compat.rs:14-31`)

## Non-blocking

### 🤔 thinking

- **Redundant compat conversion on the prebid hot path**: `request_bids` (`prebid.rs:1012`) and `to_openrtb` (`prebid.rs:713`) both convert the same `context.request` in the same auction flow — pull up once and thread through.

### ♻️ refactor

- **Three compat functions ship without callers**: `from_fastly_request`, `to_fastly_request`, and `from_fastly_response` are only referenced from their own tests. `CLAUDE.md` says "Don't design for hypothetical future requirements. No half-finished implementations either." Ship in the PR that uses them. (`crates/trusted-server-core/src/compat.rs:40, 61, 90`)

### 🌱 seedling / 📌 out of scope / ⛏ nitpick

- **📌 Redundant conversion at the auction boundary**: acknowledged by TODO at `auction/formats.rs:93-95`; accepted cost of incremental migration.
- **🌱 `sanitize_fastly_forwarded_headers` get-then-remove**: `remove_header` is idempotent; the `get_header` guard exists only for the debug log. (`compat.rs:129-136`)
- **⛏ `forward_cookie_header` panics on `HeaderValue::from_str`**: fires only on already-validated input, but a `try_from` + skip keeps failure local to the function. (`cookies.rs:156-186`)

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS (841/841)
- browser integration / integration / artifacts (GitHub Actions): PASS

crates/trusted-server-core/src/cookies.rs
```diff
@@ -246,24 +255,35 @@ pub fn create_ec_cookie(settings: &Settings, ec_id: &str) -> String {
 /// from injecting spurious cookie attributes via a controlled ID value.
 ///
 /// `cookie_domain` comes from operator configuration and is considered trusted.
-pub fn set_ec_cookie(settings: &Settings, response: &mut fastly::Response, ec_id: &str) {
-    if !is_safe_cookie_value(ec_id) {
+pub fn set_ec_cookie(settings: &Settings, response: &mut Response<EdgeBody>, ec_id: &str) {
```
aram356, line 279: 🔧 **wrench** — Clippy fails: `set_ec_cookie` is missing a `# Panics` section.

The migration switched this helper to `http::HeaderValue::from_str(...).expect(...)`, which clippy's `missing_panics_doc` lint flags (the pre-PR `fastly::Response::append_header` form did not panic). Running locally:

```
error: docs for function which may panic missing `# Panics` section
   --> crates/trusted-server-core/src/cookies.rs:258:1
```

The project's `format.yml` runs `cargo clippy --workspace --all-targets --all-features -- -D warnings`, but it only triggers on PRs whose base is `main` — that's why this PR's CI passed despite the failure. Once this is merged or rebased onto `main`, clippy will block.

**Fix:** the `expect` is unreachable in practice because `create_ec_cookie` sanitizes via the allowlist, so document it:

```rust
/// # Panics
///
/// Does not panic in practice — the cookie value is sanitized via the
/// allowlist in [`create_ec_cookie`] so [`HeaderValue::from_str`] always
/// succeeds. Listed here only because clippy cannot prove it.
```

crates/trusted-server-core/src/cookies.rs
```diff
@@ -246,24 +255,35 @@ pub fn create_ec_cookie(settings: &Settings, ec_id: &str) -> String {
...
 }
 
 /// Expires the EC cookie by setting `Max-Age=0`.
 ///
 /// Used when a user revokes consent — the browser will delete the cookie
 /// on receipt of this header.
-pub fn expire_ec_cookie(settings: &Settings, response: &mut fastly::Response) {
-    let cookie = format!("{}=; {}", COOKIE_TS_EC, ec_cookie_attributes(settings, 0),);
-    response.append_header(header::SET_COOKIE, cookie);
+pub fn expire_ec_cookie(settings: &Settings, response: &mut Response<EdgeBody>) {
```
aram356, line 301: 🔧 **wrench** — Clippy fails: `expire_ec_cookie` is missing a `# Panics` section.

Same root cause as `set_ec_cookie`: the migration to `http::HeaderValue::from_str(...).expect(...)` introduces a panic edge that clippy's `missing_panics_doc` lint requires to be documented. CI didn't catch it because the format/clippy workflow only runs on PRs against `main`.

```
error: docs for function which may panic missing `# Panics` section
   --> crates/trusted-server-core/src/cookies.rs:277:1
```

**Fix:** add a `# Panics` section noting the panic is unreachable because the formatted value uses only allowlisted characters.

crates/trusted-server-core/src/cookies.rs
```diff
@@ -144,28 +146,35 @@ pub fn strip_cookies(cookie_header: &str, cookie_names: &[&str]) -> String {
...
-    };
-
-    if !strip_consent {
-        to.set_header(header::COOKIE, cookie_value);
-        return;
-    }
+/// are removed before forwarding. If stripping leaves no cookies or yields an
+/// invalid header value, the stripped header is omitted. Non-UTF-8 cookie
+/// headers are forwarded unchanged.
+pub fn forward_cookie_header(
```
aram356, line 153: 🔧 **wrench** — The migrated http-typed `forward_cookie_header` has no callers and no tests in this PR.

Every caller in this PR (lockr, prebid) routes through `compat::forward_fastly_cookie_header`. The two have **subtly different semantics**:

- `compat::forward_fastly_cookie_header` reads only `from.get_header(header::COOKIE)` (first value) — preserves pre-PR behavior.
- `cookies::forward_cookie_header` reads `from.headers().get_all(header::COOKIE)` (all values) — new behavior.

When a future PR switches a caller, multi-Cookie-header semantics change silently and there is no test to catch a regression in the http-typed path. `grep forward_cookie_header crates/trusted-server-core/src/cookies.rs` returns only the definition.

**Fix:** add unit tests for `cookies::forward_cookie_header` covering: strip-consent path, strip-all-leaves-empty path, multiple Cookie headers via `append`, and non-UTF-8 cookie value. This locks in the intentional `get_all` semantics before the next migration PR.

crates/trusted-server-core/src/lib.rs
```diff
@@ -35,6 +35,7 @@ pub mod auction;
 pub mod auction_config_types;
 pub mod auth;
 pub mod backend;
+pub mod compat;
```
aram356, line 39: 🤔 **thinking** — `compat` is exposed as a public module despite being PR 15 removal scaffolding.

Every doc comment in `compat.rs` says `# PR 15 removal target`, but `pub mod compat` makes it part of `trusted_server_core`'s public API. External crates can grow dependencies on it before PR 15 deletes it.

**Suggestion:** `pub(crate) mod compat;` (or `#[doc(hidden)] pub mod compat;` if cross-crate access is required from `trusted-server-adapter-fastly`). Confirms the deletion in PR 15 won't be a breaking-API change.

crates/trusted-server-core/src/compat.rs
```diff
@@ -0,0 +1,385 @@
...
+/// Headers are copied; the body is empty.
+///
+/// # PR 15 removal target
+///
+/// # Panics
+///
+/// Panics if the Fastly request URL cannot be parsed as an `http::Uri`.
+pub fn from_fastly_headers_ref(req: &fastly::Request) -> http::Request<EdgeBody> {
+    build_http_request(req, EdgeBody::empty())
+}
```
aram356, line 51: 🤔 **thinking** — `# Panics` section is inaccurate.

Doc says "Panics if the Fastly request URL cannot be parsed as an `http::Uri`." but the implementation uses:

```rust
let uri: http::Uri = req
    .get_url_str()
    .parse()
    .unwrap_or_else(|_| http::Uri::from_static("/"));
```

So URL-parse failure does not panic. The actual panic risk is the inner `builder.body(body).expect(...)` (and the comment immediately above asserts that's unreachable). Either drop the `# Panics` section or document the real condition.

crates/trusted-server-core/src/compat.rs
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+    match body {
+        EdgeBody::Once(bytes) => {
+            if !bytes.is_empty() {
+                fastly_resp.set_body(bytes.to_vec());
+            }
+        }
+        EdgeBody::Stream(_) => {
+            log::warn!("streaming body in compat::to_fastly_response; body will be empty");
```
aram356, line 74: 🤔 **thinking** — `to_fastly_response` silently truncates streaming bodies.

The `EdgeBody::Stream` arm logs a `warn!` and emits an empty body. No current caller passes a streaming body (basic-auth `unauthorized_response` and `serve_static_with_etag` both use `EdgeBody::Once`/`empty`), so this is latent. As more callers migrate, an accidental streaming response would get silently truncated.

**Suggestion:** at minimum `debug_assert!(matches!(body, EdgeBody::Once(_)))`, or make the signature accept only `EdgeBody::Once` and add a separate streaming variant when actually needed.

crates/trusted-server-core/src/compat.rs
```diff
@@ -0,0 +1,385 @@
...
+    if !crate::cookies::synthetic_id_cookie_value_is_safe(synthetic_id) {
+        log::warn!(
+            "Rejecting EC ID for Set-Cookie: value of {} bytes contains characters illegal in a cookie value",
+            synthetic_id.len()
+        );
+        return;
+    }
+    response.append_header(
+        header::SET_COOKIE,
+        crate::cookies::create_ec_cookie(settings, synthetic_id),
```
aram356: ♻️ **refactor** — Hardcoded cookie attributes drift from `cookies::ec_cookie_attributes`.

```rust
format!(
    "{}=; Domain={}; Path=/; Secure; HttpOnly; SameSite=Lax; Max-Age=0",
    crate::constants::COOKIE_TS_EC,
    settings.publisher.cookie_domain,
)
```

This duplicates `cookies::ec_cookie_attributes` (currently `fn`, private). If the canonical attributes change (e.g., `SameSite=Strict`, add `Partitioned`, change Max-Age), the compat copy diverges silently until PR 15 deletes it. Promote `ec_cookie_attributes` to `pub(crate)` and reuse it:

```rust
response.append_header(
    header::SET_COOKIE,
    format!("{}=; {}", crate::constants::COOKIE_TS_EC,
            crate::cookies::ec_cookie_attributes(settings, 0)),
);
```

crates/trusted-server-core/src/compat.rs
```diff
@@ -0,0 +1,385 @@
...
+
+    if !strip_consent {
+        to.set_header(header::COOKIE, cookie_value);
+        return;
+    }
+
+    match cookie_value.to_str() {
+        Ok(value) => {
+            let stripped = strip_cookies(value, CONSENT_COOKIE_NAMES);
+            if !stripped.is_empty() {
```
aram356, line 127: ♻️ **refactor** — Naming uses old "synthetic" terminology.

The project recently renamed Synthetic ID → Edge Cookie (EC) (commit `f92f8def`). The migrated `cookies.rs` now uses `set_ec_cookie` / `expire_ec_cookie`, but the compat shims revert to `set_fastly_synthetic_cookie` / `expire_fastly_synthetic_cookie`. Rename to `set_fastly_ec_cookie` / `expire_fastly_ec_cookie` for consistency with the rest of the EC-renaming effort.

crates/trusted-server-core/src/compat.rs
```diff
@@ -0,0 +1,385 @@
...
+use edgezero_core::body::Body as EdgeBody;
+use fastly::http::header;
+
+use crate::constants::INTERNAL_HEADERS;
+use crate::http_util::SPOOFABLE_FORWARDED_HEADERS;
+
+fn build_http_request(req: &fastly::Request, body: EdgeBody) -> http::Request<EdgeBody> {
+    let uri: http::Uri = req
+        .get_url_str()
+        .parse()
```
aram356: ⛏ **nitpick** — URI fallback to `"/"` swallows malformed URLs silently.

If `req.get_url_str()` ever returns a value `http::Uri` rejects, callers see a request whose URI is `"/"` with no authority. Host extraction (uses headers) would still work, but path-based logic would silently break. In practice Fastly always provides a parseable URL, so this is theoretical — at minimum log a `warn!` so it's not invisible if it ever fires.

crates/trusted-server-core/src/migration_guards.rs
```diff
@@ -0,0 +1,54 @@
...
+        .filter(|line| {
+            let trimmed = line.trim_start();
+            !trimmed.starts_with("//")
+        })
+        .collect::<Vec<_>>()
+        .join("\n")
+}
+
+#[test]
+fn migrated_utility_modules_do_not_depend_on_fastly_request_response_types() {
```
aram356, line 26: 👍 **praise** — Strong regression test. Concise, well-documented limitations (line comments, string literals, block comments) with explicit reasoning that false-positives are preferable to false-negatives for guard tests. Good defensive engineering — locks in the migration so future PRs can't silently revert.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -1003,23 +1003,27 @@ impl AuctionProvider for PrebidAuctionProvider {
     ) -> Result<fastly::http::request::PendingRequest, Report<TrustedServerError>> {
         log::info!("Prebid: requesting bids for {} slots", request.slots.len());
 
+        let http_req = compat::from_fastly_headers_ref(context.request);
```
aram356, line 1386: 👍 **praise** — Lifting `RequestInfo::from_request` to `prepare_request` and threading it through `to_openrtb` deduplicates the work (previously computed both inline and inside `to_openrtb`) and produces a single source of truth for the auction. A real readability and correctness improvement, not just a mechanical type swap.

aram356 (changes requested):
## Summary

PR 11 of the EdgeZero migration: lifts the utility layer (auth, cookies, http_util, consent) off `fastly::{Request, Response}` onto `http::{Request, Response}<EdgeBody>`, with a `compat` bridge module that lets non-migrated callers keep working. The shape of the migration is sound and the new `migration_guards` test is a strong regression backstop. The principal concerns are CI (clippy fails when run against `main`) and a migrated-but-untested code path that will silently change behavior in the next PR.

## Blocking

### 🔧 wrench

- **Clippy fails: `cookies::set_ec_cookie` missing `# Panics`** (`crates/trusted-server-core/src/cookies.rs:258`)
- **Clippy fails: `cookies::expire_ec_cookie` missing `# Panics`** (`crates/trusted-server-core/src/cookies.rs:277`)
- **Migrated `cookies::forward_cookie_header` has no callers and no tests** — semantics diverge from `compat::forward_fastly_cookie_header` (`get_all` vs first-only) (`crates/trusted-server-core/src/cookies.rs:152`)

CI didn't catch the clippy regressions because `format.yml` only triggers on PRs whose base is `main`. PR 623's base is the PR 10 feature branch, so the clippy + fmt gate was bypassed entirely.

## Non-blocking

### 🤔 thinking

- **`compat` is `pub mod compat`** — exposed as public API despite being PR 15 removal scaffolding (`crates/trusted-server-core/src/lib.rs:38`)
- **`from_fastly_headers_ref` `# Panics` doc is inaccurate** — claims URL parse panics but uses `unwrap_or_else` fallback (`crates/trusted-server-core/src/compat.rs:46`)
- **`to_fastly_response` silently truncates `EdgeBody::Stream`** — latent risk as more callers migrate (`crates/trusted-server-core/src/compat.rs:65`)

### ♻️ refactor

- **`compat::expire_fastly_synthetic_cookie` hardcodes cookie attributes** instead of reusing `cookies::ec_cookie_attributes` — DRY drift on security-sensitive code (`crates/trusted-server-core/src/compat.rs:145`)
- **`compat` uses old "synthetic" naming** instead of "ec" — inconsistent with the recent EC rename (`crates/trusted-server-core/src/compat.rs:118`)

### ⛏ nitpick

- **`from_fastly_headers_ref` URI fallback to `"/"` swallows malformed URLs silently** — at minimum log a warning (`crates/trusted-server-core/src/compat.rs:17`)

### 👍 praise

- **`migration_guards.rs`** — concise regression test with well-justified limitations
- **`PrebidAuctionProvider::to_openrtb` request_info refactor** — single source of truth, real readability win

## CI Status

- clippy: **FAIL** — 2 errors (`missing_panics_doc` on `set_ec_cookie` / `expire_ec_cookie`), bypassed by base branch routing
- rust tests: PASS (`cargo test --workspace`)
- integration tests (GitHub Actions): PASS (3/3)

crates/trusted-server-core/src/cookies.rs
```diff
@@ -144,28 +146,35 @@ pub fn strip_cookies(cookie_header: &str, cookie_names: &[&str]) -> String {
...
-        Ok(s) => {
-            let stripped = strip_cookies(s, CONSENT_COOKIE_NAMES);
-            if !stripped.is_empty() {
-                to.set_header(header::COOKIE, &stripped);
+        match cookie_value.to_str() {
+            Ok(s) => {
+                let stripped = strip_cookies(s, CONSENT_COOKIE_NAMES);
+                if !stripped.is_empty() {
+                    if let Ok(value) = http::HeaderValue::from_str(&stripped) {
+                        to.headers_mut().append(header::COOKIE, value);
```
aram356, line 170: 🤔 **thinking** — `cookies::forward_cookie_header` and `compat::forward_fastly_cookie_header` diverge on **two** axes, only one of which is tested.

The prior review flagged the source-side divergence (compat reads first `Cookie` header via `get_header`; this reads `get_all`) and the new test `test_forward_cookie_header_multiple_cookie_headers_appended` locks that in — good.

The target-side divergence is unflagged: this function uses `headers_mut().append(header::COOKIE, …)` (preserves any pre-existing Cookie on `to`); the compat shim uses `to.set_header(header::COOKIE, …)` (replaces). For a `to` that already carries a Cookie header, the two functions produce different output. No current caller does that, but PR 12+ will swap callers from compat to this; if a future caller seeds `to` from a request template that already has cookies, behavior changes silently.

**Suggested fix:** either

1. Add a defensive test like `forward_cookie_header_appends_to_existing_target_cookies` that documents the chosen contract, or
2. Add a one-line doc note on the function: "existing `Cookie` headers on `to` are preserved; source values are appended after them."

Non-blocking — flagging now so PR 12/13 callers know to look.

aram356 (commented):
## Summary

PR 11 of the EdgeZero migration: lifts six utility modules (`auth`, `cookies`, `http_util`, `consent/extraction`, `consent/mod`, plus the JSON content-type in `request_signing/endpoints`) off `fastly::{Request, Response}` onto `http::{Request, Response}<EdgeBody>` with a `compat` bridge for non-migrated callers. CI is green across all 13 jobs and clippy is clean against `main` locally. All blocking findings from the prior CHANGES_REQUESTED reviews are resolved — `set_ec_cookie`/`expire_ec_cookie` panics docs added, `forward_cookie_header` now has dedicated tests for the multi-Cookie and non-UTF-8 paths, and the `*_synthetic_cookie` → `*_ec_cookie` rename keeps the EC terminology consistent. Filing non-blocking observations only; a clean approve will follow.

## Non-blocking

### 🤔 thinking

- **`forward_cookie_header` source/target divergence**: source-side `get_all` vs `get_header` is tested, but target-side `append` vs `set` is unflagged and would change behavior for any future caller that seeds `to` with a pre-existing `Cookie` header (`crates/trusted-server-core/src/cookies.rs:169`).
- **TSJS path allocates the response body twice** through the `serve_static_with_etag` → `compat::to_fastly_response` round-trip; disappears in PR 13/14 (`crates/trusted-server-core/src/publisher.rs:140`).

### 🌱 seedling

- **Double `http::Request` build per request** in `handle_publisher_request` and `handle_auction` — both build `http_req` explicitly, then pass `&req` (fastly) to `get_or_generate_ec_id`, which rebuilds it inside `get_ec_id` (`crates/trusted-server-core/src/edge_cookie.rs:120`).
- **`compat::set_fastly_ec_cookie` duplicates the validate-then-build control flow** of `cookies::set_ec_cookie`. A `cookies::try_build_ec_cookie_value` helper would dedupe both append-paths (`crates/trusted-server-core/src/compat.rs:156`).

### 📝 note

- **PR description "Changes" table is out of sync with the diff**: lists `auction/formats.rs`, `integrations/testlight.rs`, `proxy.rs`, `synthetic.rs` which aren't in `git diff main...HEAD` (folded into PR 9/10 merges). Update before merge for accuracy.

### 👍 praise

- **All 9 prior findings addressed correctly** — panics docs, EC rename, attribute reuse, `#[doc(hidden)]`, `debug_assert!` on stream truncation, URI parse warning, plus comprehensive tests for the divergent code paths.
- **`migration_guards` ban list extension** to `fastly::mime::APPLICATION_JSON` keeps the regression net tight (`crates/trusted-server-core/src/migration_guards.rs:42`).
- **`PrebidAuctionProvider::to_openrtb` request_info refactor** — lifts `RequestInfo::from_request` to `prepare_request`, eliminating a duplicate construction (`crates/trusted-server-core/src/integrations/prebid.rs:1007`).

## CI Status

- fmt: PASS
- clippy: PASS (verified locally with `--workspace --all-targets --all-features -- -D warnings`)
- rust tests: PASS
- vitest: PASS

crates/trusted-server-core/src/publisher.rs
```diff
@@ -129,12 +130,15 @@ pub fn handle_tsjs_dynamic(
...
     let filename = &path[PREFIX.len()..];
+    let http_req = compat::from_fastly_headers_ref(req);
 
     if UNIFIED_FILENAMES.contains(&filename) {
         // Serve core + immediate modules (excludes deferred like prebid)
         let module_ids = integration_registry.js_module_ids_immediate();
         let body = trusted_server_js::concatenate_modules(&module_ids);
-        let mut resp = serve_static_with_etag(&body, req, "application/javascript; charset=utf-8");
+        let http_resp =
+            serve_static_with_etag(&body, &http_req, "application/javascript; charset=utf-8");
```
aram356, line 140: 🤔 **thinking** — TSJS path now allocates the response body twice.

`serve_static_with_etag` returns `Response<EdgeBody::Once(Bytes)>`, then `compat::to_fastly_response` calls `bytes.to_vec()` ([compat.rs:70](crates/trusted-server-core/src/compat.rs#L70)) producing a fresh allocation. For the unified TSJS bundle this is a real per-request copy that didn't exist before the migration.

Not blocking — disappears when publisher migrates fully in PR 13/14 — but worth noting since `/static/tsjs=*` is one of the highest-RPS endpoints in the service.

crates/trusted-server-core/src/edge_cookie.rs
```diff
@@ -116,7 +117,8 @@ pub fn get_ec_id(req: &Request) -> Result<Option<String>, Report<TrustedServerEr
         log::warn!("Rejected EC ID from x-ts-ec header with disallowed characters");
     }
 
-    match handle_request_cookies(req)? {
+    let http_req = compat::from_fastly_headers_ref(req);
```
aram356: 🌱 **seedling** — Double `http::Request` build per request.

Callers in [`publisher.rs:460`](crates/trusted-server-core/src/publisher.rs#L460) and [`auction/endpoints.rs:50`](crates/trusted-server-core/src/auction/endpoints.rs#L50) already build `http_req = compat::from_fastly_headers_ref(&req)` before invoking `get_or_generate_ec_id`, which then transitively calls into here and builds a second copy. Each rebuild clones every header.

Resolves naturally when `get_ec_id` migrates to take `&Request<EdgeBody>` in a later PR. Worth tracking as a follow-up so the duplication doesn't linger silently after compat is gone.

crates/trusted-server-core/src/compat.rs
```diff
@@ -0,0 +1,394 @@
...
+            "Rejecting EC ID for Set-Cookie: value of {} bytes contains characters illegal in a cookie value",
+            ec_id.len()
+        );
+        return;
+    }
+    response.append_header(
+        header::SET_COOKIE,
+        crate::cookies::create_ec_cookie(settings, ec_id),
+    );
+}
```
aram356, line 148: 🌱 **seedling** — `compat::set_fastly_ec_cookie` and [`cookies::set_ec_cookie`](crates/trusted-server-core/src/cookies.rs#L265-L278) share validation (`ec_cookie_value_is_safe`) and assembly (`create_ec_cookie`), but the `if !is_safe { warn!; return; }` control flow is duplicated between them.

A `cookies::try_build_ec_cookie_value(settings, ec_id) -> Option<String>` helper would let both append-paths collapse to one `if let Some(v) = … { append(SET_COOKIE, v) }`. Acceptable scaffolding cost for now; flagging for the same PR 15 cleanup window.

crates/trusted-server-core/src/migration_guards.rs
```diff
@@ -0,0 +1,54 @@
...
+            include_str!("consent/extraction.rs"),
+        ),
+        ("consent/mod.rs", include_str!("consent/mod.rs")),
+    ];
+    let banned_patterns = [
+        "fastly::Request",
+        "fastly::Response",
+        "fastly::http::Method",
+        "fastly::http::StatusCode",
+        "fastly::mime::APPLICATION_JSON",
```
aram356, line 42: 👍 **praise** — Extending the ban list to `fastly::mime::APPLICATION_JSON` is a nice touch. The migration_guards file now backstops both the request/response migration and the JSON content-type migration in `request_signing/endpoints.rs` with a single test, keeping the regression net tight without scattering assertions across modules.

crates/trusted-server-core/src/cookies.rs
```diff
@@ -502,6 +561,111 @@ mod tests {
...
+        let all_cookies: Vec<_> = to
+            .headers()
+            .get_all(header::COOKIE)
+            .iter()
+            .filter_map(|v| v.to_str().ok())
+            .collect();
+        assert_eq!(all_cookies.len(), 2, "should append all Cookie headers");
+        assert!(all_cookies.iter().any(|v| v.contains("session=abc123")));
+        assert!(all_cookies.iter().any(|v| v.contains("theme=dark")));
+    }
```
aram356, line 702: 👍 **praise** — Test coverage on `forward_cookie_header` is now thorough: strip-on, strip-empty-result, no-strip, non-UTF-8 forwarding, and multi-Cookie append. Directly closes the prior reviewer's "behavior diverges silently" concern.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -1003,23 +1003,27 @@ impl AuctionProvider for PrebidAuctionProvider {
...
+        let request_info = RequestInfo::from_request(&http_req, context.client_info);
```
aram356, line 1387: 👍 **praise** — Lifting `RequestInfo::from_request` to `prepare_request` and threading it through `to_openrtb` eliminates the duplicate construction that previously sat inside `to_openrtb`. Net correctness/readability win, not just a mechanical type swap.

aram356 (approved):

PR #633, opened 2026-04-13
Add JS Asset Auditor plugin with Playwright CLI

### LINKED ISSUE #631: Implement JS Auditor - claude code
### Description

Implement the skill.

Skill should audit js assets, and generate a usable config

ChristianPavilonis:
## Summary

- Add JS Asset Auditor as a Claude Code plugin at `packages/js-asset-auditor/` with a standalone Playwright CLI that sweeps publisher pages for third-party JS assets
- Auto-detect integrations (GPT, GTM, Didomi, DataDome, Lockr, Permutive, Prebid, APS) from swept URLs and generate `trusted-server.toml` config with `--config` flag
- Expanded heuristic filters for Google ad rendering, ad verification, ad fraud detection, and reCAPTCHA to reduce noise
- Auto-include target URL host as first-party, make `trusted-server.toml` optional with `--domain` flag for portability
- Browser launches headed by default to avoid bot detection (captchas, DataDome, etc.)

### 1. Check out the branch

```bash
git fetch origin
git checkout feature/js-asset-auditor
```

### 2. Install dependencies

```bash
cd packages/js-asset-auditor && npm install && npx playwright install chromium
cd ../..
```

### 3. Run the CLI directly

```bash
# Basic audit — opens a browser, sweeps the page, writes js-assets.toml
node packages/js-asset-auditor/lib/audit.mjs https://www.publisher.com

# Also generate trusted-server.toml with detected integrations
node packages/js-asset-auditor/lib/audit.mjs https://www.publisher.com --config

# Custom output paths
node packages/js-asset-auditor/lib/audit.mjs https://www.publisher.com \
  --output /tmp/assets.toml --config /tmp/config.toml

# Headless mode (for CI, may be blocked by bot protection)
node packages/js-asset-auditor/lib/audit.mjs https://www.publisher.com --headless
```

### 4. Use as a Claude Code plugin

```bash
claude --plugin-dir packages/js-asset-auditor
```

Then in Claude Code:
```
/js-asset-auditor:audit-js-assets https://www.publisher.com
/js-asset-auditor:audit-js-assets https://www.publisher.com --config
```

### CLI flags

| Flag | Description |
|---|---|
| `--diff` | Compare sweep against existing `js-assets.toml` |
| `--settle <ms>` | Settle window after page load (default: 6000) |
| `--first-party <hosts>` | Additional first-party hosts (comma-separated) |
| `--domain <host>` | Publisher domain (falls back to `trusted-server.toml` or URL) |
| `--no-filter` | Bypass heuristic filtering |
| `--headless` | Run browser without UI (default is headed) |
| `--output <path>` | Output file for js-assets.toml (default: `js-assets.toml`) |
| `--config [path]` | Generate `trusted-server.toml` with detected integrations |
| `--force` | Overwrite existing config file |

## Test plan

- [ ] Run against a publisher page and verify `js-assets.toml` output
- [ ] Run with `--config` and verify detected integrations in generated `trusted-server.toml`
- [ ] Run with `--diff` against an existing `js-assets.toml` and verify confirmed/new/missing
- [ ] Run without `trusted-server.toml` (e.g., from `/tmp`) and verify domain inference
- [ ] Load as Claude Code plugin and test the skill
- [ ] Verify `--config` without `--force` errors when file already exists
- [ ] Verify bot-protected sites work in headed mode

Closes https://github.com/IABTechLab/trusted-server/issues/631


packages/js-asset-auditor/lib/detect.mjs
```diff
@@ -0,0 +1,278 @@
...
+    toml += `enabled = true\n`;
+
+    // Auto-extracted fields
+    for (const [key, value] of Object.entries(integration.extracted)) {
+      toml += `${key} = ${formatTomlValue(value)}  # auto-detected\n`;
+    }
+
+    // TODO fields
+    for (const field of integration.todos) {
+      toml += `${field} = ""  # TODO: set your ${integration.label} ${field}\n`;
```
aram356: 🔧 **wrench** — Generated `bidders = ""` produces invalid Prebid config.

Rust `PrebidConfig::bidders` is `Vec<String>` (`crates/trusted-server-core/src/integrations/prebid.rs:69`). The generator emits `bidders = ""` for every TODO field regardless of the target type, which deserializes as a string and fails when the user flips `enabled = true`. Same class of issue for any future array/numeric TODO fields.

**Fix** — keep TODO fields commented out so the file at least parses:

```js
for (const field of integration.todos) {
  toml += `# ${field} = ""  # TODO: set your ${integration.label} ${field}\n`;
}
```

Or, if you prefer them uncommented, annotate array fields in the pattern registry and emit `[]` instead of `""`.

packages/js-asset-auditor/lib/audit.mjs
```diff
@@ -0,0 +1,279 @@
...
+  for (let i = 2; i < argv.length; i++) {
+    const arg = argv[i];
+    if (arg === "--domain") {
+      args.domain = argv[++i];
+    } else if (arg === "--diff") {
+      args.diff = true;
+    } else if (arg === "--settle") {
+      args.settle = parseInt(argv[++i], 10);
+    } else if (arg === "--first-party") {
+      args.firstParty = argv[++i].split(",").filter(Boolean);
```
aram356: 🔧 **wrench** — `--first-party` without a value silently consumes the next flag.

Reproduced: `audit-js-assets https://example.com --first-party --diff` ends up with `firstParty = ["--diff"]` and `diff` never set to `true`. Same shape for `--settle`, `--output`, `--domain`, `--config`, and `--config` when a path isn't provided — any flag that consumes `argv[++i]` without validating.

**Fix**:

```js
} else if (arg === "--first-party") {
  const val = argv[++i];
  if (!val || val.startsWith("--")) {
    console.error(`--first-party requires a value`);
    process.exit(1);
  }
  args.firstParty = val.split(",").filter(Boolean);
}
```

Apply the same pattern to the other value-consuming flags.

docs/superpowers/specs/2026-04-01-js-asset-auditor-design.md
```diff
@@ -0,0 +1,366 @@
...
+| Framework CDNs      | `cdnjs.cloudflare.com`, `ajax.googleapis.com`, `cdn.jsdelivr.net`, `unpkg.com`                |
+| Error tracking      | `sentry.io`, `bugsnag.com`, `rollbar.com`                                                     |
+| Font services       | `fonts.googleapis.com`, `fonts.gstatic.com`                                                   |
+| Social embeds       | `platform.twitter.com`, `platform.x.com`, `connect.facebook.net`                              |
+| Google ad rendering | `pagead2.googlesyndication.com`, `tpc.googlesyndication.com`, `s0.2mdn.net`,                  |
+|                     | `googleads.g.doubleclick.net`, `www.googleadservices.com`                                     |
+| Ad fraud detection  | `adtrafficquality.google`                                                                     |
+| Ad verification     | `adsafeprotected.com`, `moatads.com`, `doubleverify.com`                                      |
+| reCAPTCHA           | `recaptcha.net`, `www.google.com/recaptcha/*`, `www.gstatic.com/recaptcha/*`                   |
```
aram356: 🔧 **wrench** — `format-docs` CI is failing on this file.

Prettier requires table column realignment here (heuristic filter table) and at ~lines 244–253 (detection patterns table). Reproduced locally with `cd docs && npx prettier --check superpowers/specs/2026-04-01-js-asset-auditor-design.md`.

**Fix**: `cd docs && npx prettier --write superpowers/specs/2026-04-01-js-asset-auditor-design.md`

scripts/js-asset-slug.mjs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+function generateSlug(publisherDomain, originUrl) {
+  const input = `${publisherDomain}|${originUrl}`;
+  const digest = createHash("sha256").update(input).digest();
+  const base62 = bufferToBase62(digest);
+  const publisherPrefix = base62.slice(0, 8);
+  const assetStem = extractAssetStem(originUrl);
+  return `${publisherPrefix}:${assetStem}`;
+}
```
aram356: ♻️ **refactor** — Slug algorithm duplicated across files.

`packages/js-asset-auditor/lib/slug.mjs` already imports from `process.mjs`. But this script duplicates the full base62 + stem-extraction implementation. The spec acknowledges this and requires manual sync, which will drift — and silent drift here breaks KV lookups against the Rust proxy.

**Fix** — make this a thin wrapper:

```js
import { generateSlug } from "../packages/js-asset-auditor/lib/process.mjs";
// ...existing arg parsing and console.log
```

`process.mjs` only uses `node:crypto` and `node:path`, so no extra deps are needed.

packages/js-asset-auditor/lib/audit.mjs
```diff
@@ -0,0 +1,279 @@
...
+//   node packages/js-asset-auditor/lib/audit.mjs https://www.publisher.com [options]
+//   audit-js-assets https://www.publisher.com [options]  (when plugin bin/ is in PATH)
+//
+// Options:
+//   --diff              Compare against existing js-assets.toml
+//   --settle <ms>       Settle window after page load (default: 6000)
+//   --first-party <h>   Additional first-party hosts (comma-separated)
+//   --no-filter         Bypass heuristic filtering
+//   --headed            Run browser visibly for debugging
+//   --output <path>     Output file path (default: js-assets.toml)
```
aram356, line 19: ♻️ **refactor** — Top-of-file help block is stale.

Lists `--headed` (not implemented — actual flag is `--headless`) and omits `--config`, `--force`, `--domain`. The usage error on line 111 is correct; update this block to match.

packages/js-asset-auditor/lib/audit.mjs
```diff
@@ -0,0 +1,279 @@
...
+      // No config file — infer from target URL
+      try {
+        const host = new URL(args.url).hostname;
+        domain = host.startsWith("www.") ? host.slice(4) : host;
+      } catch {
+        domain = args.url;
+      }
+      console.error(`No trusted-server.toml found, using domain: ${domain}`);
+    }
+  }
```
aram356: 🤔 **thinking** — `readPublisherDomain` swallows parse errors as "file not found".

If `trusted-server.toml` exists but the hand-rolled parser can't extract `[publisher].domain` (missing section, single-quoted value, multiline string, etc.), the catch falls through to URL inference and the user sees `No trusted-server.toml found`. Because the publisher domain feeds the slug hash, this silently produces wrong slugs.

**Fix**: distinguish `ENOENT` from parse failure and fail loudly on the latter unless `--domain` was explicit.

packages/js-asset-auditor/lib/audit.mjs
```diff
@@ -0,0 +1,279 @@
...
+
+      if (detection.integrations.length > 0) {
+        // Check if config file already exists
+        let fileExists = false;
+        try {
+          readFileSync(args.config);
+          fileExists = true;
+        } catch {
+          // File doesn't exist — safe to write
+        }
```
aram356: 🤔 **thinking** — `--config` existence check uses `readFileSync`.

Reads the whole file into memory just to test existence. Use `fs.existsSync` (or `fs.statSync`) instead:

```js
import { existsSync } from "node:fs";
// ...
const fileExists = existsSync(args.config);
```

packages/js-asset-auditor/lib/detect.mjs
```diff
@@ -0,0 +1,278 @@
...
+    },
+    todos: [],
+    category: "full",
+  },
+  {
+    id: "google_tag_manager",
+    label: "Google Tag Manager",
+    match: (url) =>
+      url.hostname === "www.googletagmanager.com" &&
+      url.pathname.includes("/gtm.js"),
```
aram356: ⛏ **nitpick** — `pathname.includes("/gtm.js")` would match `/gtm.js.bak` or `/foo/gtm.js` under `www.googletagmanager.com`. In practice GTM only serves `/gtm.js`, so this is theoretical, but `.endsWith("/gtm.js")` is a one-character change that removes the ambiguity.

packages/js-asset-auditor/lib/detect.mjs
```diff
@@ -0,0 +1,278 @@
...
+    integrations: [...detected.values()],
+  };
+}
+
+// ---------------------------------------------------------------------------
+// Config generation
+// ---------------------------------------------------------------------------
+
+function formatTomlValue(value) {
+  if (typeof value === "string") return `"${value}"`;
```
aram356: ⛏ **nitpick** — `formatTomlValue` doesn't escape `"` or `\`. All current inputs are static defaults or URL/hostname strings that can't contain `"`, so it's harmless today, but `JSON.stringify(value)` would give you free escape handling for the string branch if the pattern registry grows.

aram356 (changes requested):
## Summary

Additive Claude Code plugin (`packages/js-asset-auditor/`) with a Playwright-based CLI for sweeping publisher pages, a processing library, and integration auto-detection. No Rust changes. Review uncovered three blocking issues: the `format-docs` CI gate is red, the generated `trusted-server.toml` emits an invalid `bidders = ""` for Prebid, and the CLI arg parser silently consumes the next flag when a value is omitted. A handful of non-blocking refactor/hardening suggestions follow.

## Blocking

### 🔧 wrench

- **format-docs CI failing**: prettier wants table column realignment in `docs/superpowers/specs/2026-04-01-js-asset-auditor-design.md` (heuristic filter table and detection patterns table). Fix: `cd docs && npx prettier --write superpowers/specs/2026-04-01-js-asset-auditor-design.md`. (inline at line 100)
- **Generated `bidders = ""` is invalid for Rust Prebid config**: `PrebidConfig::bidders` is `Vec<String>` (`crates/trusted-server-core/src/integrations/prebid.rs:69`). Users who flip `enabled = true` on a generated `[integrations.prebid]` block get a deserialization failure. (inline at `detect.mjs:266`)
- **`--first-party` / `--settle` / `--output` / `--domain` consume the next flag**: passing a value-taking flag without a value silently swallows the following arg. Reproduced. (inline at `audit.mjs:84`)

## Non-blocking

### ♻️ refactor

- **Slug algorithm duplicated in `scripts/js-asset-slug.mjs`** — make it import from `packages/js-asset-auditor/lib/process.mjs`. Silent drift here breaks KV lookups against the Rust proxy. (inline at `scripts/js-asset-slug.mjs:78`)
- **Stale help block in `audit.mjs`** — references `--headed` (not implemented; actual flag is `--headless`) and omits `--config`, `--force`, `--domain`. (inline at `audit.mjs:18`)

### 🤔 thinking

- **`readPublisherDomain` conflates parse errors with missing file** — a malformed but present `trusted-server.toml` silently falls through to URL-inferred domain, producing wrong slugs (publisher prefix depends on domain). Distinguish `ENOENT` from parse failure. (inline at `audit.mjs:142`)
- **`--config` existence check uses `readFileSync`** — use `fs.existsSync`. (inline at `audit.mjs:238`)
- **No unit tests for `process.mjs` / `detect.mjs`** — both have nontrivial logic (wildcard regex, first-party boundary, heuristic filter with two match shapes, integration field extraction). Vitest already runs in `crates/js/lib`; adding fixtures there (or alongside the plugin) would catch drift, especially for the slug hash that must match the Rust proxy.

### ⛏ nitpick

- **GTM match uses `pathname.includes("/gtm.js")`** — `.endsWith("/gtm.js")` removes theoretical ambiguity. (inline at `detect.mjs:37`)
- **`formatTomlValue` doesn't escape `"` / `\`** — fine today since inputs are static, but `JSON.stringify` on the string branch is free escape handling for future patterns. (inline at `detect.mjs:226`)

### 🌱 seedling

- **Cross-language slug fixture test** — the spec claims the JS slug must produce identical output to the Rust proxy's KV key derivation, but the Rust proxy lives on a separate branch. Once it lands, a shared-fixture test (same `{domain, url}` → same slug in both JS and Rust) is the only reliable guard against silent drift. Worth a follow-up issue.

## CI Status

- cargo fmt: PASS
- cargo clippy: PASS
- rust tests: PASS
- vitest (`crates/js/lib`): PASS
- format-typescript: PASS
- **format-docs: FAIL** (see 🔧 above)

packages/js-asset-auditor/lib/audit.mjs
```diff
@@ -0,0 +1,279 @@
...
+import { readFileSync, writeFileSync } from "node:fs";
+import { resolve } from "node:path";
+import { processAssets } from "./process.mjs";
+
+// ---------------------------------------------------------------------------
+// Config reading
+// ---------------------------------------------------------------------------
+
+function readPublisherDomain(repoRoot) {
+  const content = readFileSync(
```
aram356: 🔧 The auditor reads [publisher].domain from trusted-server.toml as the default.

Lookup chain in packages/js-asset-auditor/lib/audit.mjs:127-142:

--domain flag (if passed)
readPublisherDomain(repoRoot) → reads [publisher].domain from trusted-server.toml (audit.mjs:31-56)
Infer hostname from the target URL
Since trusted-server.toml:12 has domain = "test-publisher.com", every run without --domain picks that up. Fix options: pass --domain <real-domain> at the CLI, or change line 12 to the actual publisher. Want me to add a --domain example to the usage string or change the fallback?

aram356 (changes requested):

packages/js-asset-auditor/lib/detect.mjs
ChristianPavilonis: Fixed in 66951b91. TODO fields are now emitted as commented-out placeholders, so detected Prebid config no longer generates active `bidders = ""` lines. I also added a regression test covering the parseable skeleton output.

packages/js-asset-auditor/lib/audit.mjs
ChristianPavilonis: Fixed in 66951b91. The CLI now validates all value-taking flags (`--domain`, `--settle`, `--first-party`, `--output`, explicit-path `--config`) and exits with usage text instead of consuming the next flag. Added regression coverage for the `--first-party --diff` case.

docs/superpowers/specs/2026-04-01-js-asset-auditor-design.md
ChristianPavilonis: Fixed in 66951b91. I ran Prettier on the spec file and verified `cd docs && npx prettier --check superpowers/specs/2026-04-01-js-asset-auditor-design.md` passes.

scripts/js-asset-slug.mjs
ChristianPavilonis: Fixed in 66951b91. `scripts/js-asset-slug.mjs` is now a thin wrapper around `packages/js-asset-auditor/lib/process.mjs`, and I updated the spec wording plus added a parity test so the wrapper stays aligned.

packages/js-asset-auditor/lib/audit.mjs
ChristianPavilonis: Fixed in 66951b91. The top-of-file help block now matches the implementation: `--headless` is documented and `--domain`, `--config`, and `--force` are included.

packages/js-asset-auditor/lib/audit.mjs
ChristianPavilonis: Fixed in 66951b91. Domain resolution now distinguishes `ENOENT` from parse/missing-field errors: missing config falls back to URL inference, while malformed/present config fails loudly. Added test coverage for the malformed config path.

packages/js-asset-auditor/lib/audit.mjs
ChristianPavilonis: Fixed in 66951b91. The config existence probe now uses `existsSync()` instead of `readFileSync()`.

packages/js-asset-auditor/lib/detect.mjs
ChristianPavilonis: Fixed in 66951b91. GTM detection now uses `pathname.endsWith("/gtm.js")`, with a regression test to ensure we only match the expected script path.

packages/js-asset-auditor/lib/detect.mjs
ChristianPavilonis: Fixed in 66951b91. TOML string formatting now uses escaped string serialization rather than raw interpolation, and I added coverage for quote/backslash escaping.

packages/js-asset-auditor/lib/audit.mjs
ChristianPavilonis: Addressed in 66951b91 by clarifying the CLI usage/help text and tightening domain-resolution error handling. The precedence is still `--domain` → `trusted-server.toml` → inferred URL host, but the behavior is now more explicit to operators.

packages/js-asset-auditor/lib/detect.mjs
```diff
@@ -0,0 +1,277 @@
...
+  toml += `[publisher]\n`;
+  toml += `domain = ${formatTomlString(domain)}\n`;
+  toml += `# cookie_domain = ${formatTomlString(`.${domain}`)}\n`;
+  toml += `# origin_url = ${formatTomlString(`https://origin.${domain}`)}\n`;
+  toml += `# proxy_secret = ${formatTomlString("change-me")}\n`;
+
+  for (const integration of detectionResult.integrations) {
+    toml += `\n`;
+    toml += `[integrations.${integration.id}]\n`;
+    toml += `enabled = true\n`;
```
prk-Jr: 🔧 **wrench** — This always writes `enabled = true`, even for `partial` / `detect_only` integrations where the generated section is missing required fields. That means the tool can emit a config that looks valid but causes trusted-server startup to fail as soon as the user drops it in. For example, `prebid` needs a real `server_url`, `aps` needs `pub_id`, and `lockr` needs `app_id`.

**Fix**: only auto-enable `full` integrations. For `partial` / `detect_only`, emit `enabled = false` (or comment out the section) until the required TODO fields are filled in.

packages/js-asset-auditor/lib/process.mjs
```diff
@@ -0,0 +1,411 @@
...
+  const parts = parsed.pathname.split("/").filter(Boolean);
+  const filename = parts.at(-1) || parsed.pathname;
+  return `${parsed.hostname}/.../` + filename;
+}
+
+// ---------------------------------------------------------------------------
+// Diff mode: parse existing TOML
+// ---------------------------------------------------------------------------
+
+export function parseExistingToml(content) {
```
prk-Jr, line 218: 🔧 **wrench** — `--diff` appends new assets as commented TOML, but `parseExistingToml()` only looks for live `[[js_assets]]` blocks. On the next run, the same commented suggestion is invisible to the parser, so the asset is treated as new again and appended a second time. I reproduced this locally by running the processing pipeline twice against the same file; the second run duplicated the identical `NEW` block.

**Fix**: parse previously appended commented entries as existing suggestions too, or persist pending diff entries in a machine-readable form so repeated diff runs become idempotent.

packages/js-asset-auditor/lib/audit.mjs
```diff
@@ -0,0 +1,304 @@
...
+    console.error(`Wrote ${args.output} (${count} entries)`);
+
+    if (args.config) {
+      const { detectIntegrations, generateConfig } = await import("./detect.mjs");
+      const detection = detectIntegrations(scriptUrls);
+
+      if (detection.integrations.length > 0) {
+        const fileExists = existsSync(args.config);
+
+        if (fileExists && !args.force) {
```
prk-Jr: 🔧 **wrench** — When `--config` targets an existing file without `--force`, this only logs an error and then continues to print a success summary with exit code `0`. That makes automation think the run succeeded even though the requested config write was refused, and it conflicts with the PR description/spec that says this case should error.

**Fix**: treat this as a hard failure: print the message, exit non-zero, and avoid emitting the normal success JSON summary.

prk-Jr (changes requested):
## Summary

This plugin is close, but there are three blocking behavior issues in the CLI/config pipeline that can mislead operators or produce unusable output. CI is green, but these cases need to be fixed before the tool is safe to rely on for onboarding and diff workflows.

## Blocking

### 🔧 wrench

- **Generated configs enable integrations before required fields are present** ([packages/js-asset-auditor/lib/detect.mjs:260](https://github.com/IABTechLab/trusted-server/blob/66951b915d7afb36b386d93c9120c8e65616fa4e/packages/js-asset-auditor/lib/detect.mjs#L260))
- **Diff mode keeps re-appending the same NEW assets on repeated runs** ([packages/js-asset-auditor/lib/process.mjs:214](https://github.com/IABTechLab/trusted-server/blob/66951b915d7afb36b386d93c9120c8e65616fa4e/packages/js-asset-auditor/lib/process.mjs#L214))
- **`--config` conflict only logs an error and still exits successfully** ([packages/js-asset-auditor/lib/audit.mjs:266](https://github.com/IABTechLab/trusted-server/blob/66951b915d7afb36b386d93c9120c8e65616fa4e/packages/js-asset-auditor/lib/audit.mjs#L266))

packages/js-asset-auditor/lib/detect.mjs
aram356: 🔧 **wrench** — `enabled = true` is emitted for every detected integration regardless of category, including `partial` / `detect_only` sections whose required fields are left as commented `# TODO` placeholders. The 66951b91 fix made the file parseable, but the Rust deserializer still fails at startup on a missing required field (`PrebidIntegrationConfig::server_url` is `#[validate(url)] String` with no serde default — `crates/trusted-server-core/src/integrations/prebid.rs:57`). Same class of failure for `aps.pub_id` and `lockr.app_id`. Operators dropping the generated file in place get a broken boot.

**Fix** — only auto-enable when the section is complete:

```js
const isComplete = integration.category === "full" || integration.todos.length === 0;
toml += `enabled = ${isComplete ? "true" : "false"}\n`;
```

Or comment out the whole section for `detect_only` so it won't be loaded accidentally.

packages/js-asset-auditor/lib/process.mjs
aram356, line 218: 🔧 **wrench** — Diff mode re-appends the same NEW assets on every subsequent run.

Reproduced locally: ran `processAssets()` three times against the same file with an unchanged input. `summary.new.length` was `1` on every run and the appended asset appeared **6×** in the final file (two TOML lines × three runs). Cause: the appender at L356–363 writes `# [[js_assets]]` / `# origin_url = "…"`, but `parseExistingToml` splits on the literal `[[js_assets]]` marker and anchors `/^origin_url\s*=/m` with no `#` prefix, so previously-appended commented entries are invisible to the next sweep.

**Fix** — either parse commented suggestions too (tolerate an optional `#\s*` prefix on the splitter and the `origin_url` regex), or persist pending diff entries in a separate section/file so diff runs are idempotent. Add a `process.test.mjs` regression that asserts `summary.new.length === 0` on the second consecutive diff run.

packages/js-asset-auditor/lib/audit.mjs
```diff
@@ -0,0 +1,304 @@
...
+          console.error(
+            `${args.config} already exists. Use --force to overwrite.`,
+          );
```
aram356: 🔧 **wrench** — `--config` collision logs an error but still exits 0.

When the target config exists and `--force` is not passed, this branch prints the message and falls through to `console.log(JSON.stringify(result.summary))` at L290, so the process exits `0` with the normal success summary. Automation cannot distinguish "wrote nothing" from "wrote successfully," and it contradicts the spec at `docs/superpowers/specs/2026-04-01-js-asset-auditor-design.md:290` (“the CLI errors unless `--force` is passed”).

**Fix** — make it a hard failure before the summary is printed:

```js
if (fileExists && !args.force) {
  console.error(`${args.config} already exists. Use --force to overwrite.`);
  process.exit(1);
}
```

packages/js-asset-auditor/lib/audit.mjs
```diff
@@ -0,0 +1,304 @@
...
+    } else if (arg === "--output") {
+      args.output = requireFlagValue(argv, i, "--output");
+      i += 1;
+    } else if (arg === "--config") {
+      const next = argv[i + 1];
+      if (next && !next.startsWith("--")) {
+        args.config = next;
+        i += 1;
+      } else {
+        args.config = "trusted-server.toml";
```
aram356: 🔧 **wrench** — `--config` with no value defaults to `trusted-server.toml`, which is the exact file the tool just read the publisher domain from. Combined with `--force`, this silently overwrites the real publisher config with a TODO-only skeleton — in a project where `cwd()` is the repo root, that is the primary config file. Combined with the non-zero-exit bug above, the no-`--force` case fails silently.

**Fix options (pick one):**

- Change the default path to a non-colliding name: `args.config = "trusted-server.generated.toml";`
- Require an explicit path after `--config` and drop the no-arg form.
- Refuse to overwrite the file `readPublisherDomain` just read, even with `--force`, unless a separate `--allow-overwrite-source` flag is passed.

The first option is the cheapest and aligns with the PR description (“generate `trusted-server.toml`”) without pointing at the live file by default.

aram356 (changes requested):
## Summary

Re-review against `main` after 66951b91. That commit cleanly addresses every finding from my prior review (format-docs, `bidders = ""`, arg-parser flag-swallowing, slug duplication, stale help block, `existsSync`, GTM `endsWith`, TOML escaping, ENOENT-vs-parse distinction), with focused regression tests for each.

However, the 2026-04-22 review from @prk-Jr is **not** addressed by the latest push — I confirmed all three of their blockers in the worktree — and I found two additional blocking-class issues while re-reading the 18 changed files.

## Blocking

### 🔧 wrench

- **`enabled = true` for `partial` / `detect_only` integrations** — generated config boot-fails on missing required fields (`server_url`, `pub_id`, `app_id`). `packages/js-asset-auditor/lib/detect.mjs:260` (inline).
- **Diff mode re-appends the same NEW assets on repeated runs** — reproduced: same asset appears 6× after 3 runs because `parseExistingToml` ignores commented suggestions. `packages/js-asset-auditor/lib/process.mjs:214` (inline).
- **`--config` collision exits 0 with success summary** — violates spec, breaks automation. `packages/js-asset-auditor/lib/audit.mjs:269` (inline).
- **`--config` default path collides with live `trusted-server.toml`** — with `--force`, silently overwrites the real publisher config with a TODO skeleton. `packages/js-asset-auditor/lib/audit.mjs:136` (inline).
- **Plugin unit tests are not wired into CI** — `.github/workflows/test.yml` only runs vitest in `crates/js/lib`. The 8 regression tests added in 66951b91 (the sole guard against drift on slug algorithm, integration detection, arg parsing) run locally only, so regressions will land on `main` unnoticed. Add a step that runs `cd packages/js-asset-auditor && npm ci && npm test` on PRs touching this directory, or unconditionally.

## Non-blocking

### 🤔 thinking

- **No diff-idempotence test** — `processAssets(..., { diff: true })` run twice against its own output should yield `summary.new.length === 0`. A fixture test in `packages/js-asset-auditor/test/process.test.mjs` would have caught the re-append regression and will keep catching it.
- **SKILL.md still says "headless browser"** — `packages/js-asset-auditor/skills/audit-js-assets/SKILL.md:24`. The CLI has defaulted to headed since e0c7e0c3. Misleading for sandbox users.
- **`data:` / `blob:` script URLs produce `"null"` origin in slugs** — `new URL("data:…").origin === "null"` so `applyWildcards` yields `null/<pathname>`. Rare for `<script src>`, but one protocol guard is cheap.

### ♻️ refactor

- **`readPublisherDomain` hand-rolls a TOML scan** (`lib/audit.mjs:37-61`) — rejects single-quoted strings and any non-`^domain = "…"` shape. A small TOML lib would be more robust. Non-urgent.

### ⛏ nitpick

- **APS pattern uses `pathname.includes("/apstag")`** (`lib/detect.mjs:152`) — same class of ambiguity as the GTM case tightened in 66951b91. Consider anchoring.

## CI Status

- cargo fmt: **PASS**
- cargo clippy / Analyze (rust): **PASS**
- cargo test: **PASS**
- vitest (`crates/js/lib`): **PASS**
- format-typescript: **PASS**
- format-docs: **PASS**
- browser integration tests: **PASS**
- CodeQL: **PASS**
- plugin tests (`packages/js-asset-auditor/test`): **NOT WIRED INTO CI** — see blocking #5

aram356 (changes requested):
🔧 Prints test.publisher.com in place of actual domain because it reads from toml file

🔧 Please add guide documentation

packages/js-asset-auditor/lib/detect.mjs
ChristianPavilonis: Fixed in 00ce7873. `generateConfig()` now only auto-enables complete integrations: `full` integrations stay enabled, while `partial` / `detect_only` integrations with remaining TODOs are emitted with `enabled = false`. I also added regression coverage in `packages/js-asset-auditor/test/detect.test.mjs` for `full`, `partial`, and `detect_only` cases.

packages/js-asset-auditor/lib/process.mjs
ChristianPavilonis: Fixed in 00ce7873. `parseExistingToml()` now recognizes previously appended commented `# [[js_assets]]` suggestions in addition to live `[[js_assets]]` blocks, so repeated `--diff` runs are idempotent instead of re-appending the same assets. I added `packages/js-asset-auditor/test/process.test.mjs` with both commented-entry parsing coverage and a repeated-diff regression asserting `summary.new.length === 0` on the second run.

packages/js-asset-auditor/lib/audit.mjs
ChristianPavilonis: Fixed in 00ce7873. Config-path collisions are now treated as a hard failure via `ensureConfigPathWritable()`: if the `--config` target already exists and `--force` is not passed, the CLI exits non-zero before printing the normal success JSON summary. Added regression coverage in `packages/js-asset-auditor/test/audit.test.mjs`.

packages/js-asset-auditor/lib/detect.mjs
ChristianPavilonis: Fixed in 00ce7873. Same underlying config-generation issue as the earlier thread: incomplete `partial` / `detect_only` integrations are no longer auto-enabled. `generateConfig()` now emits `enabled = false` until the required TODO fields are filled in, and the new detect tests cover this behavior explicitly.

packages/js-asset-auditor/lib/process.mjs
ChristianPavilonis: Fixed in 00ce7873. Diff mode now parses previously appended commented suggestions as existing entries, and I added `packages/js-asset-auditor/test/process.test.mjs` to prove repeated diff runs against unchanged input stop at `summary.new.length === 0` instead of duplicating blocks.

packages/js-asset-auditor/lib/audit.mjs
ChristianPavilonis: Fixed in 00ce7873. The CLI now treats an existing `--config` target without `--force` as a hard error and exits before printing the success summary. That behavior is covered by a new regression test in `packages/js-asset-auditor/test/audit.test.mjs`.

packages/js-asset-auditor/lib/audit.mjs
ChristianPavilonis: Fixed in 00ce7873. The no-arg `--config` default now writes to `trusted-server.generated.toml` instead of `trusted-server.toml`, so the generated config no longer collides with the live publisher config by default. I updated the CLI help/docs and added regression coverage for the new default path in `packages/js-asset-auditor/test/audit.test.mjs`.

ChristianPavilonis:
Addressed the remaining review feedback in 00ce7873 and replied/resolved the inline threads:

- incomplete integrations are no longer auto-enabled
- diff mode is now idempotent across repeated runs
- `--config` collisions now fail non-zero before the success summary
- no-arg `--config` now writes to `trusted-server.generated.toml`
- plugin tests are wired into CI
- added `docs/guide/js-asset-auditor.md`
- clarified domain precedence and source selection in the CLI/skill/docs

Validation run after the changes:
- `cd packages/js-asset-auditor && npm test`
- `cd docs && npm run format`
- `cargo test --workspace`

Re-requesting review from @aram356 and @prk-Jr.

packages/js-asset-auditor/lib/process.mjs
```diff
@@ -0,0 +1,432 @@
...
+
+  const filterCounts = {};
+  for (const f of heuristicFiltered) {
+    filterCounts[f.host] = (filterCounts[f.host] || 0) + 1;
+  }
+
+  const assets = [];
+  const seenOrigins = new Set();
+
+  for (const url of survivingUrls) {
```
aram356, line 338: 🔧 **wrench** — `data:` and `blob:` script URLs produce garbage TOML entries.

Reproduced locally by feeding the URLs through `processAssets()`:

```
data:text/javascript,console.log(1)  →  origin_url = "nulltext/javascript,console.log(1)"
blob:https://example.com/abc-123     →  origin_url = "https://example.comhttps://example.com/abc-123"
```

`new URL("data:…").origin === "null"` and `applyWildcards` concatenates that with `pathname`, producing entries the proxy can never fetch. The summary still reports them as `surfaced` assets, so an operator could commit them without noticing.

**Fix**: skip non-`http(s)` URLs early in the loop:

```js
for (const url of survivingUrls) {
  let parsed;
  try {
    parsed = new URL(url);
  } catch {
    continue;
  }
  if (parsed.protocol !== "http:" && parsed.protocol !== "https:") {
    continue;
  }
  // …existing applyWildcards / slug logic
}
```

A fixture test in `test/process.test.mjs` covering `data:` and `blob:` inputs would lock the contract.

docs/superpowers/specs/2026-04-01-js-asset-auditor-design.md
```diff
@@ -0,0 +1,366 @@
...
+script_url = "https://securepubads.g.doubleclick.net/tag/js/gpt.js"  # auto-detected
+# cache_ttl_seconds = 3600
+# rewrite_script = true
+
+[integrations.google_tag_manager]
+enabled = true
+container_id = "GTM-TRCJMD6"  # auto-detected
+
+[integrations.lockr]
+enabled = true
```
aram356: 🔧 **wrench** — Spec example contradicts the implementation after the partial/detect_only fix.

The spec shows:

```toml
[integrations.lockr]
enabled = true
sdk_url = "https://aim.loc.kr/identity-lockr-trust-server.js"  # auto-detected
# app_id = ""  # TODO: set your Lockr Identity app_id
```

But Lockr's `todos` always include `app_id` (`lib/detect.mjs:104`), so `isIntegrationConfigComplete` returns `false` and `generateConfig` correctly emits `enabled = false`. I confirmed by running the generator against `https://aim.loc.kr/identity-lockr-trust-server.js`:

```
[integrations.lockr]
enabled = false
sdk_url = "https://aim.loc.kr/identity-lockr-trust-server.js"  # auto-detected
# app_id = ""  # TODO: set your Lockr Identity app_id
```

Flip the spec example to `enabled = false` so the documented contract matches what the tool emits — otherwise future readers will treat the spec as the source of truth and re-introduce the prior bug.

packages/js-asset-auditor/lib/audit.mjs
```diff
@@ -0,0 +1,341 @@
...
+      inPublisher = true;
+      continue;
+    }
+    if (/^\[/.test(line)) {
+      inPublisher = false;
+      continue;
+    }
+    if (inPublisher) {
+      const match = line.match(/^domain\s*=\s*"([^"]+)"/);
+      if (match) return match[1];
```
aram356, line 56: ♻️ **refactor** — `readPublisherDomain` hand-rolls a strict TOML scan that rejects valid TOML.

The regex `^domain\s*=\s*"([^"]+)"` only matches double-quoted single-line strings. Real `trusted-server.toml` files using `domain = 'site.com'`, multi-line literals, or values containing escaped `"` will fail with the misleading "Could not find [publisher].domain in trusted-server.toml" — and the new test at `test/audit.test.mjs:84` actively locks this strict behavior in.

**Fix**: use a TOML parser. `smol-toml` is ~5 KB and dependency-free; `@iarna/toml` is the broader-compat option. Either lets you replace the whole 25-line scan with:

```js
import { parse } from "smol-toml";

export function readPublisherDomain(repoRoot, configPath = "trusted-server.toml") {
  const resolvedPath = resolve(repoRoot, configPath);
  const parsed = parse(readFileSync(resolvedPath, "utf-8"));
  const domain = parsed.publisher?.domain;
  if (typeof domain !== "string" || domain.length === 0) {
    throw new Error(`Could not find [publisher].domain in ${configPath}`);
  }
  return domain;
}
```

Then update the malformed-content test to feed truly malformed TOML rather than valid single-quoted TOML. *(Carried over from the 2026-04-20 review; still present.)*

packages/js-asset-auditor/lib/detect.mjs
```diff
@@ -0,0 +1,300 @@
...
+      return missing;
+    },
+    category: "detect_only",
+  },
+  {
+    id: "aps",
+    label: "Amazon Publisher Services",
+    match: (url) =>
+      url.hostname === "c.amazon-adsystem.com" &&
+      url.pathname.includes("/apstag"),
```
aram356: ⛏ **nitpick** — APS pattern uses `pathname.includes("/apstag")`, the same loose match shape that was tightened for GTM in 66951b91.

The spec at line 253 says `c.amazon-adsystem.com/aax2/apstag*`. Today the matcher would also fire on `/foo/apstag/bar`, `/apstag.js.bak`, etc. Anchor it:

```js
match: (url) =>
  url.hostname === "c.amazon-adsystem.com" &&
  url.pathname.startsWith("/aax2/apstag"),
```

or `pathname.endsWith("/apstag.js")` if you want to allow other path prefixes.

aram356 (changes requested):
## Summary

Re-review against `main` after 5b3481a4. Every prior blocker from the 2026-04-22 (`prk-Jr`) and 2026-04-23 (`aram356`) reviews is cleanly addressed in 00ce7873 + 5b3481a4, with focused regression tests for each fix and the plugin test job now wired into PR CI. CI is green (14 checks). I verified all 16 plugin tests pass locally and `prettier` is clean on the new docs files.

Two issues warrant changes before merge:

- **`data:` / `blob:` script URLs produce garbage TOML entries** that look like real `[[js_assets]]` blocks but reference unfetchable origins. Reproduced both cases through `processAssets()`. Inline below.
- **Spec example for Lockr now contradicts the implementation** — the doc shows `enabled = true` for an integration that the (corrected) generator now emits as `enabled = false`. Inline below.

Plus two non-blocking suggestions: replace the hand-rolled TOML scan in `readPublisherDomain` with a real parser (carried over — the new test now locks in the strict-double-quote-only limitation), and tighten the loose `pathname.includes(\"/apstag\")` APS pattern the same way GTM was tightened in 66951b91.

## Blocking

### 🔧 wrench

- **`data:` / `blob:` URLs produce garbage TOML** (`packages/js-asset-auditor/lib/process.mjs:321`, inline)
- **Spec example contradicts implementation** (`docs/superpowers/specs/2026-04-01-js-asset-auditor-design.md:284`, inline)

## Non-blocking

### ♻️ refactor

- **`readPublisherDomain` hand-rolled TOML scan** (`packages/js-asset-auditor/lib/audit.mjs:39-63`, inline)

### 🤔 thinking

- **Integration detection runs against unfiltered URLs** — `lib/audit.mjs:305` calls `detectIntegrations(scriptUrls, …)` on the raw network capture, not the post-first-party set. If a publisher self-hosts their prebid bundle at `cdn.publisher.com/prebid.js`, a `[integrations.prebid]` block lands in the generated config. Not strictly wrong (publisher *does* run prebid), but it bypasses the first-party boundary the asset pipeline carefully enforces. Consider passing the third-party-filtered URL set instead.

### ⛏ nitpick

- **APS pattern is loose** (`packages/js-asset-auditor/lib/detect.mjs:170`, inline)
- **`--first-party` accepts unvalidated values** (`packages/js-asset-auditor/lib/audit.mjs:163`) — splits on `,` with no normalization, so `--first-party https://www.example.com` silently no-ops because `stripWww` doesn't strip the protocol. Either reject non-hostname values or normalize via `new URL(value).hostname`.

## CI Status

- cargo fmt: PASS
- cargo clippy / Analyze (rust): PASS
- cargo test: PASS
- vitest (`crates/js/lib`): PASS
- format-typescript: PASS
- format-docs: PASS
- CodeQL: PASS
- js-asset-auditor tests: PASS *(newly wired into CI in this PR)*

ChristianPavilonis:
I will use the generate docs skill

packages/js-asset-auditor/lib/process.mjs
ChristianPavilonis: Fixed in 4773d13b. `processAssets()` now parses each surviving URL before wildcarding/slug generation and skips anything whose protocol is not `http:` or `https:`, so `data:` and `blob:` script URLs are no longer surfaced as TOML assets. Added `packages/js-asset-auditor/test/process.test.mjs` coverage for both schemes and verified only the valid HTTPS script is emitted.

docs/superpowers/specs/2026-04-01-js-asset-auditor-design.md
ChristianPavilonis: Fixed in 4773d13b. The Lockr example in `docs/superpowers/specs/2026-04-01-js-asset-auditor-design.md` now shows `enabled = false`, matching the generator behavior for partial integrations with required TODO fields such as `app_id`.

packages/js-asset-auditor/lib/audit.mjs
ChristianPavilonis: Leaving this one deferred for now rather than adding a new dependency in this focused review-fix pass. The existing ENOENT-vs-parse-failure behavior remains intact, so malformed or unsupported present configs fail loudly instead of silently falling back to URL inference. I agree a TOML parser would be a good follow-up hardening change if we want broader TOML syntax support.

packages/js-asset-auditor/lib/detect.mjs
ChristianPavilonis: Fixed in 4773d13b. APS detection now matches the documented prefix with `url.pathname.startsWith("/aax2/apstag")` instead of the broader `includes("/apstag")`, and `packages/js-asset-auditor/test/detect.test.mjs` covers both the positive `/aax2/apstag.js` case and a false-positive `/foo/apstag/bar.js` case.

docs/guide/js-asset-auditor.md
```diff
@@ -0,0 +1,242 @@
...
+```toml
+[[js_assets]]
+# https://web.prebidwrapper.com/golf-WnLmpLyEjL/default-v2/prebid-load.js
+slug = "aB3kR7mN:prebid-load"
+path = "/js-assets/aB3kR7mN/prebid-load.js"
+origin_url = "https://web.prebidwrapper.com/golf-WnLmpLyEjL/default-v2/prebid-load.js"
+inject_in_head = true
+```
+
+| Field | Description |
```
aram356: 🔧 **wrench** — `format-docs` CI is failing on this file.

Reproduced locally with `cd docs && npm run format`: Prettier wants column-aligned tables here (the `Field / Description` table) and again at lines ~227–238 (the `Common options` table), plus the matching tables in `docs/superpowers/specs/implemented/2026-04-01-js-asset-auditor-design.md` (asset-entry table at L122–130 and diff-mode condition table at L208–212).

This is the same class of failure that landed in the original review and was fixed in 66951b91; the doc additions in `dd859bfa` reintroduced it. CI run: https://github.com/IABTechLab/trusted-server/actions/runs/25353966443/job/74339283778

**Fix:**

```bash
cd docs && npm run format:write
```

then commit the resulting whitespace-only changes. `cd docs && npm run format` should then pass.

packages/js-asset-auditor/lib/process.mjs
```diff
@@ -0,0 +1,438 @@
...
+    }
+
+    const { wildcarded, original, hasWildcard } = applyWildcards(url);
+
+    if (seenOrigins.has(wildcarded)) continue;
+    seenOrigins.add(wildcarded);
+
+    const slug = generateSlug(args.domain, wildcarded);
+    const prefix = slug.split(":")[0];
+    const injectInHead = normalizedHead.has(url);
```
aram356: 🤔 **thinking** — `inject_in_head` is determined from the pre-wildcard URL on a first-seen-wins basis.

`injectInHead = normalizedHead.has(url)` uses the original URL, but the dedup at L335 keys on the **wildcarded** URL. If a publisher's page loads `…/prod/1.19.8/raven.js` from `<body>` and `…/prod/1.19.9/raven.js` from `<head>` in the same sweep, both collapse to the same wildcard but `inject_in_head` reflects only whichever URL the iteration saw first.

Not common in practice (a single page sweep usually loads one version), but worth flagging — when a wildcarded entry is generated, `inject_in_head` should arguably be `true` if **any** of the originals appeared in head.

**Suggested approach:**

```js
const injectInHead = [...normalizedHead].some((headUrl) => {
  const { wildcarded: headWildcarded } = applyWildcards(headUrl);
  return headWildcarded === wildcarded;
});
```

or track the OR across originals collapsed into the same wildcard during the dedup loop.

packages/js-asset-auditor/lib/detect.mjs
```diff
@@ -0,0 +1,300 @@
...
+      rewrite_sdk: true,
+    },
+    todos: [],
+    category: "full",
+  },
+  {
+    id: "lockr",
+    label: "Lockr Identity",
+    match: (url) => {
+      const href = url.href.toLowerCase();
```
aram356, line 96: ♻️ **refactor** — Lockr hostname match uses `String.prototype.includes`, which is looser than every other pattern in this file.

`url.hostname.includes("aim.loc.kr")` matches `notaim.loc.kr` and `aim.loc.kr.attacker.com`. Combined with the `identity-lockr` substring + `.js` suffix requirements (and the fact that partial integrations are emitted with `enabled = false`), the practical blast radius is limited to a noisy false positive in the generated config — but the test is still wrong, and inconsistent with how GTM / Datadome / Didomi / GPT / APS / Permutive all use `===` or `endsWith`.

**Fix:**

```js
const { hostname } = url;
const hostMatch =
  hostname === "aim.loc.kr" ||
  hostname.endsWith(".aim.loc.kr") ||
  hostname === "identity.loc.kr" ||
  hostname.endsWith(".identity.loc.kr");
if (!hostMatch) return false;
const href = url.href.toLowerCase();
return href.includes("identity-lockr") && href.endsWith(".js");
```

Worth covering with a regression test that asserts `notaim.loc.kr/identity-lockr-foo.js` does **not** match.

packages/js-asset-auditor/lib/audit.mjs
```diff
@@ -0,0 +1,359 @@
...
+const DEFAULT_GENERATED_CONFIG_PATH = "trusted-server.generated.toml";
+
+const USAGE =
+  "Usage: audit-js-assets <url> [--diff] [--domain <domain>] [--settle <ms>] [--first-party <hosts>] [--no-filter] [--headless] [--output <path>] [--config [path]] [--force]";
+
+// ---------------------------------------------------------------------------
+// Config reading
+// ---------------------------------------------------------------------------
+
+export function readPublisherDomain(repoRoot, configPath = "trusted-server.toml") {
```
aram356, line 39: ♻️ **refactor** — Carry-over from the 2026-04-29 review: `readPublisherDomain` is still a hand-rolled TOML scan that only matches `domain = "…"` (double-quoted basic strings).

Author explicitly deferred adding a TOML dependency in this pass, and the new test in `audit.test.mjs` locks in the strict-double-quote-only behavior — a single-quoted `domain = 'publisher.com'` is treated as missing and the CLI errors out, which is at least consistent. Re-flagging as non-blocking so the follow-up stays on the list: an actual TOML parser (e.g., `smol-toml`, ~6 KB) would gracefully accept literal strings and multi-line strings, and would future-proof against operators copy-pasting examples from the wider TOML ecosystem.

docs/superpowers/specs/implemented/2026-04-01-js-asset-auditor-design.md
```diff
@@ -0,0 +1,373 @@
+---
+status: implemented
+implemented_in: feature/js-asset-auditor
+last_reviewed: 2026-05-04
+verified_against_commit: 4773d13b
+---
```
aram356, line 5: ⛏ **nitpick** — Frontmatter `verified_against_commit: 4773d13b` is stale.

Latest commit on this branch is `dd859bfa` (`Update docs for JS Asset Auditor`), which itself touches this file. Either bump the value to the current head or drop the field if it's not meant to track every doc edit.

aram356 (changes requested):
## Summary

Re-review of `feature/js-asset-auditor` against `main` at head `dd859bfa`. The PR has been through three earlier review passes and Christian has cleanly addressed every prior blocker — fixes landed with focused regression tests, `parseExistingToml` now recognizes commented suggestions, `--config` collisions exit non-zero, integrations with TODO fields no longer auto-enable, `data:` / `blob:` URLs are rejected, and the plugin test job is wired into PR CI. 20/20 plugin tests pass locally.

**One blocker:** `format-docs` CI is failing on the new docs files. Same class of issue that landed in the original review (fixed in 66951b91); the doc additions in `dd859bfa` reintroduced it. Reproduced locally — see inline. Everything else is non-blocking polish.

## Blocking

### 🔧 wrench

- **`format-docs` CI is failing** — Prettier wants column-aligned tables in `docs/guide/js-asset-auditor.md` (L100–105 and L227–238) and `docs/superpowers/specs/implemented/2026-04-01-js-asset-auditor-design.md` (L122–130 and L208–212). Inline.

## Non-blocking

### 🤔 thinking

- **Init mode silently overwrites an existing `js-assets.toml`** — `audit-js-assets <url>` with no `--diff` writes `args.output` unconditionally at `packages/js-asset-auditor/lib/audit.mjs:314`. Inconsistent with `--config`, which fails loudly via `ensureConfigPathWritable()` unless `--force` is passed. If an operator re-runs init forgetting `--diff`, hand-curated entries are lost silently. Consider applying `ensureConfigPathWritable(args.output, args.force)` symmetrically (or at least when output is the default path).
- **`inject_in_head` first-seen-wins on wildcard collapse** — see inline at `packages/js-asset-auditor/lib/process.mjs:340`.

### ♻️ refactor

- **Lockr `hostname.includes` is loose** — see inline at `packages/js-asset-auditor/lib/detect.mjs:88`.
- **`readPublisherDomain` hand-rolled TOML scan** — see inline at `packages/js-asset-auditor/lib/audit.mjs:39`.

### 🌱 seedling

- **`formatTomlEntry` doesn't escape strings** — `packages/js-asset-auditor/lib/process.mjs:185-196` uses raw interpolation for `slug`, `path`, `origin_url`. detect.mjs's `formatTomlValue` does escape via `JSON.stringify`. All three values today come from URL-encoded ASCII so the inconsistency is harmless, but it's a footgun if anyone later adds a non-URL-derived field.
- **Missing direct unit tests for `applyWildcards`, `matchesHeuristicFilter`, `extractAssetStem`, `normalizeUrl`, `isFirstParty`** — All exercised only indirectly through `processAssets`. Direct tests would catch regressions in the wildcard semver/hex/mixed-hash heuristics and in the dot-boundary suffix logic (e.g., a regression where `evil-googletagmanager.com` starts matching).
- **`page.waitForTimeout` is "discouraged" in Playwright** — `packages/js-asset-auditor/lib/audit.mjs:264`. For an audit tool a fixed-window settle is reasonable, but `page.waitForLoadState('networkidle', { timeout: args.settle })` is the documented alternative.
- **Permutive omits commented `cache_ttl_seconds` / `rewrite_sdk` defaults** — `packages/js-asset-auditor/lib/detect.mjs:126-129` lists only `api_endpoint` and `secure_signals_endpoint`. The Permutive Rust struct also defines `cache_ttl_seconds` and `rewrite_sdk` with serde defaults, and GPT lists analogous defaults in the generator. Inconsistent operator visibility into available overrides.

### 📌 out of scope

- **Unrelated chrome-devtools MCP permissions in root `.claude/settings.json`** — `.claude/settings.json:27-32` adds `navigate_page`, `list_network_requests`, `close_page`. The auditor uses Playwright, not chrome-devtools-mcp. These additions look orthogonal to this PR's scope and would be cleaner in a separate "expand harness permissions" change so the diff stays focused on the auditor.

## CI Status

- cargo fmt: PASS
- cargo clippy / Analyze (rust): PASS
- cargo test: PASS
- vitest (`crates/js/lib`): PASS
- format-typescript: PASS
- **format-docs: FAIL**
- CodeQL: PASS
- js-asset-auditor tests: PASS

docs/guide/js-asset-auditor.md
ChristianPavilonis: Fixed in c8cf17fc. Ran `cd docs && npm run format:write`, committed the resulting table formatting updates for both docs files, and verified with `cd docs && npm run format`.

packages/js-asset-auditor/lib/process.mjs
ChristianPavilonis: Fixed in c8cf17fc. `processAssets()` now precomputes wildcarded head URLs and sets `inject_in_head` from the wildcarded value, so a collapsed entry is head-injected if any original URL in that wildcard group appeared in `<head>`. Added regression coverage for the head/body mixed-version case and the body-only control case.

packages/js-asset-auditor/lib/detect.mjs
ChristianPavilonis: Fixed in c8cf17fc. Lockr detection now uses exact-or-subdomain host matching for `aim.loc.kr` and `identity.loc.kr` instead of `includes()`. Added regression coverage proving valid Lockr hosts still match while `notaim.loc.kr` and `aim.loc.kr.attacker.com` do not.

packages/js-asset-auditor/lib/audit.mjs
ChristianPavilonis: Leaving this deferred as non-blocking for this PR. The current behavior still fails loudly for present-but-unsupported TOML instead of silently inferring the wrong domain, and adding a TOML parser dependency is better handled as follow-up hardening rather than in this focused review-fix commit.

docs/superpowers/specs/implemented/2026-04-01-js-asset-auditor-design.md
ChristianPavilonis: Fixed in c8cf17fc by dropping `verified_against_commit` from the spec frontmatter. Since this field was only present on this one spec and had gone stale on a docs-only commit, removing it avoids future churn from tracking every doc edit.

aram356:
Closing in favor of https://github.com/IABTechLab/trusted-server/pull/669

PR #641, opened 2026-04-15
Server-side ad templates design spec

### LINKED ISSUE #639: Create Spec for Server-Side Ad-Templates
### Description

Create Detailed Spec for Server-Side Ad-Templates

jevansnyc:
## Summary

- Adds design spec for server-side ad templates, replacing client-side SDK slot discovery (Prebid.js, Smart Slots) with edge-configured creative opportunities
- TS matches URL patterns to pre-configured slot templates, fires the full auction (PBS + APS + all providers) at request receipt in parallel with origin fetch, and injects winning bids into head before the browser receives HTML
- Estimated ~2,000ms improvement in time-to-ad-visible (from ~3,200ms to ~1,200ms)

## Test plan

- [ ] Review spec for accuracy against current auction pipeline
- [ ] Validate creative-opportunities.toml schema covers publisher slot inventory
- [ ] Confirm open questions with publisher (URL patterns, homepage slot count, PBS stored request setup)


docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+
+## 2. Goal
+
+Enable Trusted Server to:
+
+1. Match an incoming page request URL against a set of pre-configured slot templates
+2. Immediately fire the full server-side auction (all providers: PBS, APS, future wrappers) in
+   parallel with the origin HTML fetch — before the browser receives a single byte
+3. Inject GPT slot definitions into `<head>` so the client can define slots without any SDK
+4. Return pre-collected winning bids to the browser's lightweight `/auction` POST before the
```
prk-Jr: ❓ **`/auction` POST role after this ships**

This goal says _"return pre-collected winning bids to the browser's lightweight `/auction` POST."_ Sections 4.4 and 4.5 say _"no `/auction` POST needed — bids are already in the page."_ These are contradictory. The JS client currently always POSTs to `/auction`. Definitive answer needed: does `__ts_bids` replace the POST entirely, or does `/auction` remain as a fallback for URLs that don't match any slot template?

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+## 4. Architecture
+
+### 4.1 New File: `creative-opportunities.toml`
+
+A new config file at the repo root, alongside `trusted-server.toml`. It holds all slot templates:
+page pattern matching rules, ad formats, floor prices, and GAM targeting key-values. Bidder-level
+params (placement IDs, account IDs) live in Prebid Server stored requests, keyed by slot ID — not
+in this file.
+
+Loaded at build time via `include_str!()`, parsed into `Vec<CreativeOpportunitySlot>` at startup.
```
prk-Jr: ❓ **`creative-opportunities.toml`: compile-time `include_str!()` or Fastly KV at runtime?**

`include_str!()` bakes the file into the WASM binary — every slot config change requires a full rebuild + Fastly deploy (~15 min). The phrase _"ad ops can edit this file independently"_ (line 71) does not hold under that model. The `RuntimeServices` abstraction already exposes `services.kv_store()`. Which model is intended: compile-time (fast reads, deploy required per change) or runtime KV (live edits, no rebuild)?

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+config — sub-millisecond.
+
+### 4.3 Auction Trigger
+
+When slots are matched, TS immediately calls `AuctionOrchestrator::run_auction()` with the
+matched slots converted to `AdSlot` objects. This happens at request receipt time — in parallel
+with the origin fetch.
+
+The orchestrator's existing behaviour is unchanged:
+- All providers (PBS, APS, any configured wrappers) are dispatched simultaneously
```
prk-Jr: ❓ **APS in Phase 1: included or excluded, and param source**

APS bidder params currently flow from the browser's `AdRequest` POST — there is no client for a server-triggered auction. `creative-opportunities.toml` has no `[slot.providers.aps]` section and APS does not use PBS stored requests. Is APS in scope for Phase 1? If yes, where do per-slot APS params come from?

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+  }
+]
+```
+
+**Second injection — `window.__ts_bids`** — injected once auction results are available, just
+before `</head>`. Keyed by slot ID. The client reads this directly — no `/auction` POST needed:
+
+```json
+{
+  "atf_sidebar_ad": { "hb_pb": "2.50", "hb_bidder": "kargo", "hb_adid": "abc123" },
```
prk-Jr: ❓ **`hb_pb` price bucketing: granularity table and full key set**

`hb_pb` is a Prebid price bucket string (discretized CPM bin), not a raw price. No bucketing logic exists in the codebase — this is a net-new component. Which granularity table: low / medium / high / auto / dense / custom? Is it per-publisher configurable? And what is the complete `__ts_bids` key set — just `hb_pb`, `hb_bidder`, `hb_adid`, or also `hb_size`, `hb_deal`, `hb_format`?

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+Prebid.js is eliminated. The client-side ad bootstrap is replaced by a small inline script
+(~20 lines) that reads `__ts_ad_slots` and `__ts_bids` and drives GPT directly:
+
+```javascript
+window.__tsAdInit = function() {
+  var slots = window.__ts_ad_slots || [];
+  var bids = window.__ts_bids || {};
+  googletag.cmd.push(function() {
+    slots.forEach(function(slot) {
+      var gptSlot = googletag.defineSlot(slot.id, slot.formats, slot.id)
```
prk-Jr: ❓ **`slot.id` as GPT `adUnitPath`**

GPT's `defineSlot(adUnitPath, size, optDiv)` first argument must be the full GAM network path (e.g., `/21765378893/homepage-banner`). Using a short key like `atf_sidebar_ad` silently produces a non-functional slot — GAM will not serve to it. Is `slot.id` the full GAM path, or does `creative-opportunities.toml` need a separate field (e.g., `gam_unit_path`)?

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+GAM falls back to its own auction.
+
+**Auction times out with zero results** — `__ts_bids` is an empty object `{}`. All slots fire
+GAM without bid targeting. No revenue impact beyond the timeout scenario itself (same as today's
+fallback).
+
+**Origin is slow (NextJS 14, buffered)** — auction has more time; results more likely to be
+complete. No change to streaming behavior.
+
+**NextJS 16 streaming** — TS must flush `<head>` before `</head>` tag passes through. If auction
```
prk-Jr: ❓ **Streaming mode `</head>` boundary: Phase 1 or deferred?**

The current streaming pipeline (merged in #562) buffers until end-of-document when post-processors run. _"Buffer only until `</head>`, inject bids, then resume streaming immediately"_ is a new mode not currently implemented — it requires new infrastructure in the HTML processor. Is this required for Phase 1 launch, or is injecting at document end acceptable as an initial release?

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+
+## 9. Open Questions
+
+1. **URL pattern coverage** — does `/20*/` cover all article paths, or are there
+   non-date-prefixed article URLs? Publisher to confirm.
+2. **PBS stored request setup** — slot IDs in `creative-opportunities.toml` must have
+   corresponding stored requests configured in the publisher's PBS instance before this goes live.
+3. **Homepage slot count** — the example shows slots 0 and 1. Are there slots 2–5 following
+   the same pattern? Slot IDs and count to be confirmed with ad ops.
+4. **Auction timeout for server-side trigger** — current `[integrations.prebid].timeout_ms`
```
prk-Jr: ❓ **Auction timeout: new config key or reuse existing?** (Section 9, Q4)

The spec recommends 500ms for server-triggered auctions vs the current 1,000ms client-side budget. There are currently three overlapping timeout values: `[auction].timeout_ms`, `[integrations.prebid].timeout_ms`, `[integrations.aps].timeout_ms`. Does the server-triggered path get a new dedicated key (e.g., `[creative_opportunities].auction_timeout_ms`) or does it override an existing one?

prk-Jr (commented):
On review against the current codebase, 8 design questions need answers before implementation planning can begin. 7 are inline below; one cross-cutting question (win notifications) is here:

---

**❓ Win notifications (nurl/burl): fire server-side or defer?**

When GAM renders the creative, TS is out of the render path. The `nurl`/`burl` fields from PBS/APS responses are known to TS at auction time but have no path to the client or to a server-side firing mechanism. SSP win counting breaks silently for server-auctioned impressions. Should TS fire `nurl` as a background HTTP request at bid-selection time, or is this explicitly deferred to a later phase?

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
jevansnyc: Good catch. _ts_bids replaces the POST for URLs that match templates. /auction also stays as the fallback path for URLs without matching templates (preserves backward compatability for non-templated pages and for publishers who haven't adopted creative-opportunities.toml yet)

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
jevansnyc: You're correct. We should add an optional gam_unit_path field per slot, plus a top-level gam_network_id config. Default behavior: gam_unit_path = "/{network_id}/{slot.id}". Publishers can override per-slot for non-standard paths.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
jevansnyc: Fastly deploy's don't take 15 minutes first of all. That said, lets reduce dependence on KV store where we can for timing and publisher cost perspective and keep it simple in the WASM binary with include_str!()

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
jevansnyc: Yes, APS in scope. Lets add them as [slot.provider.aps] in the toml. They should complete with all the other demand unless you see something i don't?

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
jevansnyc: Lets make "dense" the default with a publisher override on the granularity setting. Keep the key set as is to match GAM standard.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
jevansnyc: Lets use an optional dedicated key with fallback that makes sense:

[creative_opportunities]
"# Optional. Defaults to [auction].timeout_ms if not set.
"# Recommended: 500ms (vs client-side 1000-1500ms) due to lower edge→PBS RTT.
auction_timeout_ms = 500"

- Optional + fallback keeps it backward compatible. Publishers not setting it inherit [auction].timeout_ms
- Per-provider config should stay untouched. [integrations.prebid].timeout_ms and [integrations.aps].timeout_ms continue to define provider-level ceilings ... the orchestrator's existing min(remaining_global_budget, provider_timeout) enforcement applies as today.
- Single source of truth for 4.6 streaming. A_deadline (the </head> buffer deadline) = T₀ + creative_opportunities.auction_timeout_ms (or fallback). The same value gates both the auction and the head-boundary buffer hold ... they're the same deadline.

jevansnyc:
> On review against the current codebase, 8 design questions need answers before implementation planning can begin. 7 are inline below; one cross-cutting question (win notifications) is here:
> 
> **❓ Win notifications (nurl/burl): fire server-side or defer?**
> 
> When GAM renders the creative, TS is out of the render path. The `nurl`/`burl` fields from PBS/APS responses are known to TS at auction time but have no path to the client or to a server-side firing mechanism. SSP win counting breaks silently for server-auctioned impressions. Should TS fire `nurl` as a background HTTP request at bid-selection time, or is this explicitly deferred to a later phase?

Yes, lets split the calls by where the truth lives:

nurl: TS orchestrator fires fire-and-forget BG HTTP req from the edge at bid-selection time. Edge→SSP RTT ~20–30ms, no auction-path latency cost. Per-provider switch (e.g., [integrations.prebid].fire_nurl_at_edge, default true) for cases where PBS deployment already handles win events and we'd otherwise double-fire. APS handled per its own notification spec.

burl: Fired client-side when the ad actually renders, using the existing GPT slotRenderEnded event. Logic lives inside the tsjs bundle (__tsAdInit registers the listener after defining slots). On render: if !event.isEmpty AND event.slot.getTargeting('hb_adid')[0] === bidData.hb_adid, fire  ... this confirms both render and that our specific Prebid bid (not a direct deal or backfill) won the GAM line item match. TS embeds burl per slot in __ts_bids so the client has it without a round-trip.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+
+**Second injection — `window.__ts_bids`** — injected once auction results are available, just
+before `</head>`. Keyed by slot ID. The client reads this directly — no `/auction` POST needed:
+
+```json
+{
+  "atf_sidebar_ad": { "hb_pb": "2.50", "hb_bidder": "kargo", "hb_adid": "abc123" },
+  "below-content-ad": { "hb_pb": "1.00", "hb_bidder": "appnexus", "hb_adid": "def456" }
+}
+```
```
aram356, line 426: 🔧 **wrench** — Per-request bid injection without a cache contract is unsafe.

`__ts_bids` is per-request, per-user data injected into a publisher HTML body that flows through Fastly. The spec never states the cache contract. If the response is cacheable (default for HTML on Compute unless explicitly suppressed), bids leak across users.

**Fix:** Add a section that mandates `Cache-Control: private, no-store` (or equivalent surrogate-control + cache-key segmentation) on any response with `__ts_bids` injected, and explain what happens when the publisher origin sends conflicting cache headers. This is a P0 omission — the entire design rests on it.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+When slots are matched, TS immediately calls `AuctionOrchestrator::run_auction()` with the
+matched slots converted to `AdSlot` objects. This happens at request receipt time — in parallel
+with the origin fetch.
+
+The orchestrator's existing behaviour is unchanged:
+- All providers (PBS, APS, any configured wrappers) are dispatched simultaneously
+- Per-provider timeout budgets are enforced from the remaining auction deadline
+- Floor price filtering, bid unification, and winning bid selection are applied as today
+- PBS resolves bidder params from its stored requests by slot ID — no bidder params travel
+  through TS or the browser
```
aram356: 🔧 **wrench** — Consent / GDPR flow is missing for the new auction trigger.

Today's `/auction` handler in `auction/endpoints.rs:66-77` builds a `ConsentContext` (TCF, GPP, US Privacy, GPC) before running an auction. The spec triggers the auction at request receipt — before the page's consent UI (Didomi) can update preferences — and never describes how consent is propagated, suppressed, or reconciled.

**Required additions:**
- Explicit consent-gating decision (when auction is suppressed entirely)
- Handling of mid-page consent revocation when bids are already in `<head>`
- EC-ID / cookie behavior at the new trigger
- Treatment in §8 Edge Cases as a first-class "no consent" case

Without this the design is a regulatory regression.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+
+**On NextJS 14 (buffered mode):** TS must buffer the full origin response before forwarding.
+This gives the auction the entire origin response time (~150–400ms typical) to run before
+any HTML is forwarded. In practice, bids are often collected before origin even responds.
+
+**On NextJS 16 (streaming mode):** TS streams HTML chunks to the browser immediately. The
+auction runs in parallel. Bid injection into `<head>` must complete before the `</head>` tag
+is forwarded. If the auction has not returned by the time `</head>` is encountered, TS waits
+up to the remaining auction budget, then flushes with whatever bids have arrived (partial
+results) or no targeting if timed out. Content after `</head>` is never held.
```
aram356: 🔧 **wrench** — `IntegrationHeadInjector` is synchronous; spec assumes it can wait on async results.

The current trait at `integrations/registry.rs:392-397` is `fn head_inserts(&self, ctx: &IntegrationHtmlContext<'_>) -> Vec<String>` and runs inside a `lol_html` element callback (`html_processor.rs:236-267`) — fully sync, no `.await`.

**Pick one approach and document it:**
1. Run the auction **before** building the `HtmlProcessorConfig` and pass pre-resolved bids in (simpler — but means buffering origin until auction completes), or
2. Introduce a new async/late-injection mechanism plus a chunk-holding state machine in the streaming pipeline (which today has no "hold and wait" primitive — see `html_processor.rs:46-128`).

Also note `lol_html` `element!("head", ...)` fires on the **opening** tag; "before `</head>`" requires `el.on_end_tag()`. Spec must call this out and pick an approach.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+
+---
+
+## 2. Goal
+
+Enable Trusted Server to:
+
+1. Match an incoming page request URL against a set of pre-configured slot templates
+2. Immediately fire the full server-side auction (all providers: PBS, APS, future wrappers) in
+   parallel with the origin HTML fetch — before the browser receives a single byte
```
aram356: 🔧 **wrench** — "Auction in parallel with origin fetch" is incompatible with the current publisher path.

`publisher.rs:527` calls `req.send(&backend_name)` synchronously inside a non-async `handle_publisher_request`. Real concurrency requires `send_async` + a `select` (or join) between the orchestrator future and the pending origin request, and propagating `async` up through publisher handling.

Spec §3 lists "no orchestrator changes" and §7 lists this as a one-line "Request handler modification" — the actual lift is a meaningful restructuring of the publisher request path. List it explicitly in §7 as a multi-step migration including the async propagation.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+```
+
+**Rust type:**
+
+```rust
+#[derive(Debug, Clone, serde::Deserialize)]
+pub struct CreativeOpportunitySlot {
+    pub id: String,
+    pub page_patterns: Vec<String>,
+    pub formats: Vec<AdFormat>,
```
aram356: 🔧 **wrench** — Schema mismatch: `formats = [{ width, height }]` won't deserialize into `AdFormat`.

`auction/types.rs:47-52` defines `AdFormat { media_type: MediaType, width: u32, height: u32 }` — `media_type` is required and `MediaType` has no default. The TOML examples and Rust struct in §4.1 omit `media_type`.

**Fix:** Either change the schema/struct to include `media_type` (or default to `Banner` via `#[serde(default)]`), or define a separate `CreativeOpportunityFormat` plus an explicit converter to `AdFormat`.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+    pub targeting: HashMap<String, serde_json::Value>,
+}
+```
+
+### 4.2 URL Pattern Matching
+
+At request time, TS matches the request path against each slot's `page_patterns`. Patterns are
+glob-style strings:
+
+- `/20*/` — matches all date-prefixed article paths (e.g., `/2024/01/my-article/`)
```
aram356: 🔧 **wrench** — Glob example is wrong against standard glob libraries.

`/20*/` is described as matching `/2024/01/my-article/`. Under common glob semantics (`globset`, `wax`, shell), `*` does not cross `/`, so `/20*/` matches `/2024/` but **not** `/2024/01/my-article/`.

**Fix:** Pin the pattern semantics — use `**/` for cross-segment, or adopt `URLPattern`, or document the dialect explicitly — and fix the example. This affects whether the auction even fires on article pages.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+```
+
+**Second injection — `window.__ts_bids`** — injected once auction results are available, just
+before `</head>`. Keyed by slot ID. The client reads this directly — no `/auction` POST needed:
+
+```json
+{
+  "atf_sidebar_ad": { "hb_pb": "2.50", "hb_bidder": "kargo", "hb_adid": "abc123" },
+  "below-content-ad": { "hb_pb": "1.00", "hb_bidder": "appnexus", "hb_adid": "def456" }
+}
```
aram356, line 425: 🔧 **wrench** — Untrusted bid strings injected as inline JSON without an escape contract.

`hb_bidder` / `hb_adid` come from external SSPs and end up serialized into an inline `<script>`. The existing prebid head injector (`integrations/prebid.rs:386`) already does `.replace("</", "<\\/")` for a reason.

**Required:** Spec must specify the escape contract for `__ts_bids`: at minimum `</`, U+2028, U+2029; ideally also reject control chars and validate that slot IDs / bidder names are `[A-Za-z0-9_-]+`. Without this the change is an XSS vector that didn't exist on the `/auction` POST path.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+
+### 4.1 New File: `creative-opportunities.toml`
+
+A new config file at the repo root, alongside `trusted-server.toml`. It holds all slot templates:
+page pattern matching rules, ad formats, floor prices, and GAM targeting key-values. Bidder-level
+params (placement IDs, account IDs) live in Prebid Server stored requests, keyed by slot ID — not
+in this file.
+
+Loaded at build time via `include_str!()`, parsed into `Vec<CreativeOpportunitySlot>` at startup.
+Ad ops can edit this file independently of server configuration.
```
aram356: 🔧 **wrench** — `include_str!()` contradicts "ad ops can edit independently."

`include_str!()` bakes the file into the WASM binary at compile time. Every change requires `cargo build --target wasm32-wasip1 --release` + `fastly compute publish`. That is the opposite of "ad ops can edit independently."

**Fix:** Either drop the claim and own that ad-ops changes are deploys, or move to a Fastly Edge Dictionary / KV-backed config with a documented hot-reload story. Pick one — the spec contradicts itself today.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
aram356: 🔧 **wrench** — APS bidder isn't keyed by PBS stored requests.

The spec says "PBS resolves bidder params from its stored requests by slot ID — no bidder params travel through TS." But APS is a separate provider in this codebase (`integrations/aps.rs`) dispatched from TS, not via PBS — APS slot IDs and parameters live in TS configuration.

**Fix:** Specify how APS slot IDs are resolved from `creative-opportunities.toml` (and how that maps to APS `slotID`/`slotName`/`mediaType` entries). Otherwise APS silently no-bids on every slot.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+
+1. **URL pattern coverage** — does `/20*/` cover all article paths, or are there
+   non-date-prefixed article URLs? Publisher to confirm.
+2. **PBS stored request setup** — slot IDs in `creative-opportunities.toml` must have
+   corresponding stored requests configured in the publisher's PBS instance before this goes live.
+3. **Homepage slot count** — the example shows slots 0 and 1. Are there slots 2–5 following
+   the same pattern? Slot IDs and count to be confirmed with ad ops.
+4. **Auction timeout for server-side trigger** — current `[integrations.prebid].timeout_ms`
+   is 1,000ms. Recommend reducing to 500ms for server-side triggered auctions given the
+   lower edge→PBS RTT. Separate config key or override on the new trigger path?
```
aram356: 🔧 **wrench** — Spec references the wrong timeout config key.

Spec mentions `auction_timeout_ms` (§8) and `[integrations.prebid].timeout_ms` (§9.4). The current orchestrator-wide budget lives at `settings.auction.timeout_ms` (`auction/endpoints.rs:95`); the prebid timeout key is the *per-provider* timeout used by `PrebidAuctionProvider`. These are different knobs that govern different things.

**Fix:** Pin which knob is authoritative for the new server-triggered auction and fix the references in §4.3, §8, and §9.4.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+
+**Origin is slow (NextJS 14, buffered)** — auction has more time; results more likely to be
+complete. No change to streaming behavior.
+
+**NextJS 16 streaming** — TS must flush `<head>` before `</head>` tag passes through. If auction
+not yet complete, TS waits up to `auction_timeout_ms` from the config, then flushes. Content
+streaming resumes immediately after `</head>` regardless of bid state.
+
+**`creative-opportunities.toml` missing or malformed** — startup fails with a clear error.
+No silent degradation.
```
aram356: ❓ **question** — How is the new auction trigger gated when `creative-opportunities.toml` is empty or absent?

Spec says startup fails on missing/malformed config. Is this opt-in per environment? Per integration registration? How do publishers without server-side templates still run today's `/auction` flow?

Add a kill-switch / rollback flag (e.g., `creative_opportunities.enabled = false`) that disables all new behavior and falls back to client-side, and describe how it is exercised in production (env var, feature gate, ramp).

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+
+Enable Trusted Server to:
+
+1. Match an incoming page request URL against a set of pre-configured slot templates
+2. Immediately fire the full server-side auction (all providers: PBS, APS, future wrappers) in
+   parallel with the origin HTML fetch — before the browser receives a single byte
+3. Inject GPT slot definitions into `<head>` so the client can define slots without any SDK
+4. Return pre-collected winning bids to the browser's lightweight `/auction` POST before the
+   browser would have even finished parsing Prebid.js
+5. Eliminate Prebid.js from the client entirely
```
aram356: ❓ **question** — Who fires `nurl` / `burl` and what replaces Prebid analytics adapters?

Eliminating Prebid.js drops every Prebid analytics adapter publishers may depend on for revenue reporting. Where does win/billing notification (`nurl`/`burl` in `auction/types.rs:149-152`) get fired in the new flow — server-side from the edge after GAM impression, browser-side beacon, or PBS? And how do publishers reconstruct the analytics surface they relied on?

Either commit to a replacement (server beacon at the edge, separate `tsjs` analytics module) or note this in §3 Non-Goals.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+## 6. Performance Summary
+
+| Stage | Client-side today | With TS templates | Saving |
+|---|---|---|---|
+| Script load chain | ~700ms | ~40ms (tsjs only) | -660ms |
+| Script parse/JIT | ~280ms | ~10ms | -270ms |
+| Sequential SDK hops | ~200ms | 0 | -200ms |
+| Auction window | ~1,500ms | ~500ms | -1,000ms |
+| GAM + creative | ~570ms | ~570ms | — |
+| **Total** | **~3,250ms** | **~1,200ms** | **~2,000ms** |
```
aram356: 🤔 **thinking** — Latency numbers are modeled, not measured.

The "1,200ms" is built from optimistic point estimates. Real SSP p99 frequently exceeds 800ms; creatives with multiple sub-resources push past 1,500ms. Tag the table as "modeled with budget X" or cite RUM data; otherwise readers will use it as a commitment.

Also acknowledge the buffered-mode TTFB regression: today's buffered NextJS 14 path can stream at ~150ms; §5 has TTFB at t=502ms.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+At request time, TS matches the request path against each slot's `page_patterns`. Patterns are
+glob-style strings:
+
+- `/20*/` — matches all date-prefixed article paths (e.g., `/2024/01/my-article/`)
+- `/` — matches the homepage exactly
+- `/index.html` — exact match
+
+Multiple slots can match a single URL. All matching slots are collected and fed into a single
+auction as separate impressions. Pattern matching is purely in-memory against the pre-parsed
+config — sub-millisecond.
```
aram356: 🤔 **thinking** — Pattern matching cost grows with inventory.

"Sub-millisecond" holds at small N, but per-request matching is O(slots × patterns).

**Suggestion:** Use `globset::GlobSetBuilder` (single combined matcher) or a `matchit::Router` with normalized prefixes. Cheap to add now; expensive to retrofit when there are thousands of slots.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+
+### New
+
+- `creative-opportunities.toml` — slot template config file
+- `crates/trusted-server-core/src/creative_opportunities.rs` — config types, TOML parsing,
+  URL pattern matching, slot-to-`AdSlot` conversion
+- `build.rs` update — `include_str!()` for `creative-opportunities.toml`
+- Request handler modification — match slots at request receipt, trigger orchestrator immediately,
+  hold result for head injection
+- `tsjs-gpt` integration update — `__tsAdInit` bootstrap replaces Prebid.js ad unit setup
```
aram356: 🤔 **thinking** — No telemetry contract for the new trigger.

Today's `handle_auction` logs "Auction completed: N providers, M winning bids, Tms total" (`auction/endpoints.rs:107-112`). The new trigger needs equivalent observability: per-trigger timing, "would-have-bid-but-timed-out" counters, head-injection success/skip, the cache-control branch taken, etc.

Add a Telemetry section to §7.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+
+```rust
+#[derive(Debug, Clone, serde::Deserialize)]
+pub struct CreativeOpportunitySlot {
+    pub id: String,
+    pub page_patterns: Vec<String>,
+    pub formats: Vec<AdFormat>,
+    pub floor_price: Option<f64>,
+    pub targeting: HashMap<String, serde_json::Value>,
+}
```
aram356, line 223: ♻️ **refactor** — Add validation for slot IDs at startup.

Define the legal alphabet (e.g., `^[a-zA-Z0-9_-]{1,64}$`) and validate at TOML parse time. Prevents both XSS and ambiguous cross-references with PBS stored-request keys.

Surface invalid IDs as startup failures per `CLAUDE.md` ("Invalid enabled integrations/providers must not be silently logged-and-disabled").

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
aram356: ♻️ **refactor** — `googletag.defineSlot(slot.id, ..., slot.id)` is wrong.

`defineSlot(adUnitPath, sizes, divId)` — `adUnitPath` is the GAM ad-unit path (e.g., `/12345/network/article-atf`) and `divId` is the page DOM container ID. Reusing `slot.id` for both will break GAM trafficking.

**Fix:** Add separate `gam_ad_unit_path` and `div_id` fields to the slot config.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+  URL pattern matching, slot-to-`AdSlot` conversion
+- `build.rs` update — `include_str!()` for `creative-opportunities.toml`
+- Request handler modification — match slots at request receipt, trigger orchestrator immediately,
+  hold result for head injection
+- `tsjs-gpt` integration update — `__tsAdInit` bootstrap replaces Prebid.js ad unit setup
+
+### Modified
+
+- `crates/trusted-server-core/src/integrations/prebid.rs` head injector — emit
+  `window.__ts_ad_slots` from matched slots
```
aram356: ♻️ **refactor** — `__ts_ad_slots` belongs in a dedicated head injector, not stuffed into `prebid.rs`.

The §7 plan modifies `integrations/prebid.rs` to emit `__ts_ad_slots`. But the spec also says "Prebid.js is eliminated." Prebid integration shouldn't own the new GPT-facing global.

**Fix:** Put the head injection in a new `creative_opportunities` integration (or extend the existing `gpt` integration at `integrations/gpt.rs`) and decouple it from prebid.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
aram356: 🌱 **seedling** — Plan for Edge Dictionary / KV-backed config swap.

Even if Phase 1 ships with `include_str!()`, sketch the migration path to dictionary/KV-backed config (read at request time, hot-reload, validation contract). Prevents painting yourselves into a corner when ad ops actually does want to make changes between deploys.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
...
+5. Eliminate Prebid.js from the client entirely
+
+**Target time to ad visible: ~1,200ms. Net saving: ~2,000ms.**
+
+---
+
+## 3. Non-Goals
+
+- Eliminating client-side GPT / Google Ad Manager — GAM remains in the rendering pipeline
+  for Phase 1. The GAM call (`securepubads.g.doubleclick.net`) moves server-side in a future phase.
```
aram356: 🌱 **seedling** — Server-side GAM is the real win.

Phase 1 keeps GAM in the browser, capping the savings ceiling. Briefly outline the Phase 2 server-side-GAM approach (`securepubads` proxy, creative bundling) so reviewers can evaluate whether the Phase 1 architecture is shape-compatible with the eventual end state.

docs/superpowers/specs/2026-04-15-server-side-ad-templates-design.md
```diff
@@ -0,0 +1,363 @@
+# Server-Side Ad Templates Design
+
+*April 2026*
```
aram356: ⛏ **nitpick** — Italics convention around the date.

Other specs in `docs/superpowers/specs/` use a header like `_Author · YYYY-MM-DD_` or omit the date entirely. Pick a convention so spec metadata is grep-able.

aram356 (changes requested):
## Summary

Design spec for server-side ad templates. The diagnosis (browser is the slowest place to run an auction; edge→PBS RTT is much shorter; slot inventory is URL-derivable) is sound, but the spec materially understates the architectural lift and skips several correctness concerns. Cross-referenced against `auction/orchestrator.rs`, `auction/types.rs`, `auction/endpoints.rs`, `integrations/registry.rs`, `integrations/prebid.rs`, `html_processor.rs`, `streaming_processor.rs`, and `publisher.rs`.

## Blocking

### 🔧 wrench

- **CI is red on this file**: `cd docs && npx prettier --check superpowers/specs/2026-04-15-server-side-ad-templates-design.md` fails (`format-docs` job). Fix: `cd docs && npm run format:write` and commit.
- **Cache contract missing for `__ts_bids`** — see inline.
- **Consent / GDPR flow missing for the new trigger** — see inline.
- **`IntegrationHeadInjector` is sync; spec assumes async wait** — see inline.
- **"Auction in parallel with origin" incompatible with current sync `handle_publisher_request`** — see inline.
- **Schema mismatch with `AdFormat` (missing `media_type`)** — see inline.
- **Glob example is wrong against standard glob libraries** — see inline.
- **Untrusted bid strings injected as inline JSON without escape contract (XSS)** — see inline.
- **`include_str!()` contradicts "ad ops can edit independently"** — see inline.
- **APS bidder isn't keyed by PBS stored requests** — see inline.
- **Spec references the wrong timeout config key** — see inline.

### ❓ question

- **Kill-switch / rollback when config is empty or absent?** — see inline.
- **Who fires `nurl`/`burl`; what replaces Prebid analytics adapters?** — see inline.

## Non-blocking

### 🤔 thinking

- **Latency numbers are modeled, not measured** — see inline.
- **Pattern-matching cost is O(slots × patterns)** — see inline.
- **No telemetry contract for the new trigger** — see inline.

### ♻️ refactor

- **Validate slot IDs at startup (XSS + cross-reference safety)** — see inline.
- **`googletag.defineSlot(slot.id, ..., slot.id)` is wrong (ad-unit path ≠ div ID)** — see inline.
- **`__ts_ad_slots` belongs in a dedicated head injector, not `prebid.rs`** — see inline.

### 🌱 seedling

- **Plan for Edge Dictionary / KV-backed config swap** — see inline.
- **Sketch the Phase 2 server-side-GAM approach** — see inline.

### 📌 out of scope

- **Browser integration test coverage**: the existing browser-integration tests today exercise the client-side ad pipeline. Implementation PRs need parallel coverage — a fixture page with `creative-opportunities.toml` matched, a stubbed PBS, and asserts that `__ts_bids` lands and `googletag.setTargeting` is invoked. Worth flagging now so it lands with the implementation, not after.

### ⛏ nitpick

- **Italics convention around the date** — see inline.

## CI Status

- format-docs: **FAIL** (prettier delta on this file; 45+/32− to fix)
- cargo fmt: PASS
- cargo clippy: PASS
- cargo test: PASS
- vitest: PASS
- CodeQL / Analyze (rust, javascript-typescript, actions): PASS

aram356:
Closing since implementation PR contains updated spec https://github.com/IABTechLab/trusted-server/pull/680

PR #609, opened 2026-04-02
Wire request signing to RuntimeServices store primitives (PR 9)

prk-Jr:
## Summary

- Replaces direct `FastlyConfigStore`/`FastlySecretStore` construction in `request_signing/` with `RuntimeServices` platform traits, making the signing subsystem platform-agnostic
- Moves management API write operations (config item / secret CRUD) out of `trusted-server-core` into an adapter-only `FastlyManagementApiClient`, enforcing the EdgeZero layering rule that only the adapter may call Fastly-specific APIs
- Threads `RuntimeServices` into `AuctionContext` so `PrebidAuctionProvider` (and future providers) can call `RequestSigner::from_services` without a deprecated `FastlyConfigStore` shim

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-adapter-fastly/src/management_api.rs` | **New** — `FastlyManagementApiClient` with `update_config_item`, `delete_config_item`, `create_secret`, `delete_secret` via Fastly management API |
| `crates/trusted-server-adapter-fastly/src/platform.rs` | `FastlyPlatformConfigStore::put/delete` and `FastlyPlatformSecretStore::create/delete` delegate to `FastlyManagementApiClient` |
| `crates/trusted-server-adapter-fastly/src/main.rs` | Wire `services` into `/verify-signature` handler; add `mod management_api` |
| `crates/trusted-server-core/src/request_signing/signing.rs` | Rewrite — `RequestSigner::from_services(&RuntimeServices)` replaces `from_config()`; `verify_signature` accepts `&RuntimeServices`; deprecated shim removed |
| `crates/trusted-server-core/src/request_signing/rotation.rs` | Rewrite — `KeyRotationManager::new()` now infallible; all methods accept `services: &RuntimeServices` |
| `crates/trusted-server-core/src/request_signing/endpoints.rs` | All handlers accept `services: &RuntimeServices`; tests use in-memory stub stores |
| `crates/trusted-server-core/src/auction/types.rs` | `AuctionContext` gains `pub services: &'a RuntimeServices` |
| `crates/trusted-server-core/src/auction/orchestrator.rs` | Pass `services` through derived `AuctionContext` construction sites |
| `crates/trusted-server-core/src/auction/endpoints.rs` | Pass `services` when constructing `AuctionContext` |
| `crates/trusted-server-core/src/integrations/prebid.rs` | Use `RequestSigner::from_services(context.services)` — no more `#[allow(deprecated)]` |
[...]

PR #642, opened 2026-04-15
Support Sourcepoint GPP consent for EC generation

### LINKED ISSUE #640: Support Sourcepoint localStorage consent for Edge Cookie generation via GPP/US  consent bridging
### Summary

 Edge Cookie generation currently fails for sites using Sourcepoint when consent
 is stored only in localStorage and not surfaced via the standard cookies Trusted
 Server reads.

 In our site, Sourcepoint stores US consent in localStorage under a key like:

 - _sp_user_consent_36026

 Example payload includes:

 - full GPP string: DBABLA~BVQqAAAAAgA.QA
 - applicable section IDs: [7] (usnat)
 - consent granted state:
     - consentedToAll: true
     - consentedToAny: true
     - granularStatus.gpcStatus: false

 Today, Trusted Server only extracts consent from:

 - euconsent-v2
 - __gpp
 - __gpp_sid
 - us_privacy
 - Sec-GPC

 As a result, when Sourcepoint consent exists only in localStorage, the server
 sees no consent signal and blocks EC creation in regulated US states.

 ### Current behavior

 Given a request with cookies like:

 - _aren_ab
 - ArenaGeo=... → US / TN
 - datadome=...

 and no euconsent-v2, __gpp, __gpp_sid, or us_privacy cookie, EC generation is
 denied.

 This is expected under current code because:

 1. ArenaGeo only influences jurisdiction detection
 2. Tennessee is treated as a regulated US state
 3. US-state EC gating currently allows EC only when one of these is present:
     - TCF Purpose 1 consent
     - us_privacy that does not opt out
     - and no Sec-GPC: 1

 Even if we mirror Sourcepoint’s localStorage value into __gpp / __gpp_sid,
 current US-state EC gating does not appear to use GPP US sections as an allow
 signal for EC creation.

 ### Relevant code

 Consent extraction:
 - crates/trusted-server-core/src/consent/extraction.rs
 - crates/trusted-server-core/src/constants.rs

 Consent gating:
 - crates/trusted-server-core/src/consent/mod.rs (allows_ec_creation)

 EC generation/finalization:
 - crates/trusted-server-core/src/ec/mod.rs
 - crates/trusted-server-core/src/ec/finalize.rs

 Docs:
 - docs/guide/edge-cookies.md
 - docs/guide/ec-setup-guide.md

 ### Problem

 Trusted Server assumes consent will arrive via standard cookies/headers, but
 Sourcepoint on this site stores consent in localStorage only.

 This creates two gaps:

 1. Transport gap
 The server cannot read browser localStorage, so no consent reaches the backend
 unless client code mirrors it into cookies/headers.
 2. Semantics gap
 Even if Sourcepoint’s GPP data is mirrored into __gpp / __gpp_sid, current
 US-state EC gating does not use GPP US sections to allow EC creation.

 ### Proposed solution

 #### Phase 1: client-side mirroring

 Add client-side logic to read Sourcepoint localStorage and mirror consent into
 standard transport cookies:

 - __gpp=<full gpp string>
 - __gpp_sid=<comma-separated section ids>

 For the example payload:

 - __gpp=DBABLA~BVQqAAAAAgA.QA
 - __gpp_sid=7

 Optional temporary compatibility bridge:
 - derive and write us_privacy=<legacy value> for current server compatibility

 #### Phase 2: server-side support for GPP US sections

 Extend EC gating for Jurisdiction::UsState(_) so decoded GPP US sections can be
 used as a valid consent signal for EC creation, instead of relying only on:

 - TCF Purpose 1
 - us_privacy
 - Sec-GPC

 ### Expected behavior

 After a user accepts Sourcepoint consent:

 1. client reads _sp_user_consent_* from localStorage
 2. client writes __gpp and __gpp_sid
 3. a subsequent Trusted Server-managed request includes those cookies
 4. EC consent evaluation recognizes the US GPP signal
 5. ts-ec / x-ts-ec can be generated when consent allows it

 ### Acceptance criteria

 #### Transport

 - Sourcepoint localStorage consent can be mirrored into standard cookies
 - __gpp and __gpp_sid are populated from _sp_user_consent_*

 #### Server behavior

 - Trusted Server recognizes GPP US sections as valid US-state consent input for
 EC gating
 - allows_ec_creation() no longer requires us_privacy when equivalent GPP US
 consent is present
 - Sec-GPC: 1 still overrides and blocks EC creation
 - Existing TCF and us_privacy behavior remains unchanged

 #### Tests

 - Add tests for US-state EC allowance with GPP US section consent
 - Add tests that GPC still blocks EC when GPP is otherwise permissive
 - Add tests for deny/fail-closed behavior when no usable consent signal is
 present

 ### Notes

 This should likely be implemented as a standards-first path:

 - prefer __gpp / __gpp_sid
 - keep us_privacy only as backward-compatible fallback

 This will make TS work better not only with Sourcepoint, but with other CMPs
 that expose US consent through GPP rather than legacy us_privacy.

ChristianPavilonis:
## Summary

- Add client-side Sourcepoint JS integration that auto-discovers `_sp_user_consent_*` from localStorage and mirrors GPP consent into `__gpp` / `__gpp_sid` cookies
- Extend server-side GPP decoding to extract `sale_opt_out` from US GPP sections (IDs 7–23)
- Add GPP US consent branch to `allows_ec_creation()` between existing TCF and `us_privacy` checks

Closes #640

## Test plan

- [ ] Rust tests pass (`cargo test --workspace` — 992 tests including 8 new)
- [ ] JS tests pass (`npx vitest run` — 288 tests including 6 new)
- [ ] Clippy clean
- [ ] Verify EC generation succeeds for a Sourcepoint-only site in a regulated US state with GPP consent present
- [ ] Verify GPC still blocks EC when GPP US section is permissive
- [ ] Verify existing TCF and us_privacy EC gating behavior unchanged


crates/trusted-server-core/src/cookies.rs
```diff
@@ -145,7 +145,7 @@ mod tests {
     }
 
     #[test]
-    fn test_parse_cookies_to_jar_empty() {
+    fn test_parse_cookies_to_jar_emtpy() {
```
aram356: 🔧 **wrench** — Test-name typo regression. `test_parse_cookies_to_jar_empty` was renamed to `test_parse_cookies_to_jar_emtpy` ("emtpy") in this PR. Looks accidental — revert the rename.

crates/js/lib/src/integrations/sourcepoint/index.ts
```diff
@@ -0,0 +1,93 @@
...
+      log.debug('sourcepoint: failed to parse localStorage value', { key });
+    }
+  }
+  return null;
+}
+
+const GPP_COOKIE_MAX_AGE = 86400;
+
+function writeCookie(name: string, value: string): void {
+  document.cookie = `${name}=${value}; path=/; Secure; SameSite=Lax; Max-Age=${GPP_COOKIE_MAX_AGE}`;
```
aram356: ❓ **question** — Spec vs implementation mismatch on cookie TTL.

The design spec (`docs/superpowers/specs/2026-04-15-sourcepoint-gpp-consent-design.md:50-52`) says:

> Cookies are session-scoped (no `max-age` / `expires`) since the source of truth stays in `localStorage` and we re-mirror on each page load.

The code uses `Max-Age=86400`. Commit `d423ce6c` added the TTL with the rationale "stale consent state shouldn't outlast session," which contradicts the spec. Which is authoritative — should the spec be updated, or should `Max-Age` be dropped to match the spec?

crates/js/lib/src/integrations/sourcepoint/index.ts
```diff
@@ -0,0 +1,93 @@
...
+    try {
+      const payload = JSON.parse(raw) as SourcepointConsentPayload;
+      if (payload.gppData?.gppString) {
+        return payload;
+      }
+    } catch {
+      log.debug('sourcepoint: failed to parse localStorage value', { key });
+    }
+  }
+  return null;
```
aram356, line 132: 🤔 **thinking** — First-match behavior is position-sensitive. The loop iterates `localStorage.key(i)` in insertion order and returns the first `_sp_user_consent_*` with a non-empty `gppData.gppString`. For publishers who migrate between Sourcepoint property IDs (or for any scenario producing multiple entries), a stale entry could shadow the current one indefinitely.

Consider: scan all matching keys and prefer the most recent (if the payload carries a timestamp), or document the single-property assumption in a comment so future readers don't assume order is safe.

crates/js/lib/src/integrations/sourcepoint/index.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const GPP_COOKIE_MAX_AGE = 86400;
+
+function writeCookie(name: string, value: string): void {
+  document.cookie = `${name}=${value}; path=/; Secure; SameSite=Lax; Max-Age=${GPP_COOKIE_MAX_AGE}`;
+}
+
+function clearCookie(name: string): void {
+  document.cookie = `${name}=; path=/; Secure; SameSite=Lax; Max-Age=0`;
+}
```
aram356, line 151: 🤔 **thinking** — `clearCookie` omits the `Domain` attribute. If `__gpp` / `__gpp_sid` is ever written with `Domain=<parent>` (e.g., by a Sourcepoint bootstrap script or a parent-domain operator), `Max-Age=0` on the exact origin won't clear the parent-domain cookie and stale consent will linger.

If this integration is guaranteed to be the only writer, worth a one-line comment documenting the assumption; otherwise hedge by including a `Domain` attribute symmetric to the write.

crates/trusted-server-core/src/consent/gpp.rs
```diff
@@ -100,6 +103,83 @@ fn decode_tcf_from_gpp(parsed: &iab_gpp::v1::GPPString) -> Option<TcfConsent> {
...
+            }
+        }
+    }
+
+    if saw_not_opted_out {
+        Some(false)
+    } else {
+        None
+    }
+}
```
aram356, line 145: 🤔 **thinking** — Aggregation semantics for multiple US sections. Any decodable US section with `OptedOut` short-circuits to `Some(true)`; any decodable section saying "not opted out" (with no opt-outs seen) yields `Some(false)`. This is the privacy-safe conservative choice (OR on opt-out), but worth a short comment making that intent explicit — otherwise a future reader fixing "inconsistency" across multiple states may unknowingly relax the gating.

crates/trusted-server-core/src/consent/gpp.rs
```diff
@@ -100,6 +103,83 @@ fn decode_tcf_from_gpp(parsed: &iab_gpp::v1::GPPString) -> Option<TcfConsent> {
...
+        Section::UsNe(s) => &s.core.sale_opt_out,
+        Section::UsNh(s) => &s.core.sale_opt_out,
+        Section::UsNj(s) => &s.core.sale_opt_out,
+        Section::UsTn(s) => &s.core.sale_opt_out,
+        Section::UsMn(s) => &s.core.sale_opt_out,
+        _ => return None,
+    };
+
+    Some(*sale_opt_out == OptOut::OptedOut)
+}
```
aram356, line 178: ♻️ **refactor** — Manual enumeration of 16 US state variants will silently drop new states. When `iab_gpp` adds sections for newly passed state privacy laws (which happens every legislative session), this match returns `None` for them and the caller treats "no signal," which may incorrectly fall through to other signals.

Options:
- Add `// TODO: update when iab_gpp adds new US states` linked to a tracking issue.
- If `iab_gpp::sections::us_common` exposes a unified accessor, route through it instead of the per-state match.

crates/js/lib/test/integrations/sourcepoint/index.test.ts
```diff
@@ -0,0 +1,151 @@
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+
+import { mirrorSourcepointConsent } from '../../../src/integrations/sourcepoint';
+
+describe('integrations/sourcepoint', () => {
+  function clearAllCookies(): void {
+    document.cookie.split(';').forEach((c) => {
+      const name = c.split('=')[0].trim();
+      if (name) document.cookie = `${name}=; expires=Thu, 01 Jan 1970 00:00:00 GMT; path=/`;
```
aram356: ⛏ **nitpick** — `clearAllCookies` uses the legacy `expires=Thu, 01 Jan 1970…` pattern while production code clears via `Max-Age=0`. Purely cosmetic, but aligning the test helper with the production convention reduces cognitive overhead.

aram356 (changes requested):
## Summary

Sourcepoint GPP consent is a small, well-motivated slice — the GPP-US decoder, the `allows_ec_creation` gating branch, and their tests are solid. But as shipped, the feature is non-functional (the JS module is never served to browsers), the PR fails CI, and it bundles several large unrelated changes that should be separate PRs.

## Blocking

### 🔧 wrench

- **Sourcepoint JS module is built but never served.** `crates/trusted-server-core/src/integrations/registry.rs:790-792` hardcodes `JS_ALWAYS = &["creative"]`. The design spec (`docs/superpowers/specs/2026-04-15-sourcepoint-gpp-consent-design.md:38`) says the integration follows the same "JS-only, no Rust registration" pattern as `creative`, but that pattern requires `JS_ALWAYS` to list the module. `integrations/mod.rs` has no `sourcepoint::register` and `JS_ALWAYS` has no `"sourcepoint"` entry, so `IntegrationRegistry::js_module_ids()` will never include it — `/static/tsjs=…` will never concatenate `tsjs-sourcepoint.js` into a served bundle. End-to-end, no browser will ever execute `mirrorSourcepointConsent()`. Fix: add `"sourcepoint"` to `JS_ALWAYS` and a test asserting it ships in `js_module_ids_immediate()`.

- **Orphaned dead-code file `crates/trusted-server-core/src/ec/admin.rs`.** 380 lines implementing `POST /_ts/admin/partners/register`, but `ec/mod.rs` contains no `pub mod admin;` declaration (`grep -rn "mod admin\|ec::admin::" crates/` returns zero hits). The base branch (`feature/edge-cookies-final`) replaced the KV-backed partner registry with config-based partners in commit `049a501e`; that change dropped `admin.rs` but the rebase on this branch reinstated it unreferenced. Fix: delete `crates/trusted-server-core/src/ec/admin.rs`.

- **Clippy `-D dead_code` failures block CI.** Three dead symbols introduced by the same rebase and carrying no callers:
  - `crates/trusted-server-core/src/platform/test_support.rs:174` — `recorded_request_headers` method
  - `crates/trusted-server-core/src/platform/test_support.rs:336` — `build_services_with_http_client` function
  - `crates/integration-tests/tests/common/runtime.rs:56` — `PartnerRegistrationFailed` variant (this is the actual failing CI log message)

  Confirmed locally: `cargo clippy --workspace --all-targets --all-features -- -D warnings` fails in the PR's worktree. Fix: delete them all — nothing references any of the three.

- **Undisclosed scope: creative iframe sandbox lockdown.** `crates/js/lib/src/core/render.ts:7-18` removes `allow-scripts`, `allow-same-origin`, and `allow-forms` from the creative iframe `sandbox=` attribute, and `crates/trusted-server-core/src/creative.rs:361-367` strips `<iframe>` elements entirely during HTML sanitization. Together, any creative relying on JavaScript (viewability pixels, click tracking, rich media, VPAID) or on third-party ad-tag iframes will stop rendering. This is a major ad-tech behavioral change wholly unrelated to Sourcepoint GPP consent, and it is not mentioned in the PR body. It needs its own PR with a threat model, product sign-off, and a rollout plan. Fix: extract these two changes to a dedicated PR.

### ❓ question

- **Why is Prebid User ID Module support shipping inside this PR?** The title and body are "Support Sourcepoint GPP consent for EC generation," but the diff also contains ~132 lines of `build-all.mjs` User-ID-submodule generation, a new `_user_ids.generated.ts`, ~290 lines of new Prebid test coverage, a 212-line design spec (`specs/2026-04-16-prebid-user-id-module-design.md`), and a 599-line implementation plan. Two independent features in one review is hard to evaluate. Can Prebid User ID Module move to its own PR?

## Non-blocking

### 📌 out of scope

- **No re-mirror after mid-session CMP interaction.** `mirrorSourcepointConsent()` runs once when the module is parsed. If a user opens the Sourcepoint CMP and updates consent mid-session, `__gpp` / `__gpp_sid` won't reflect the new state until the next page load — meaning the same session can continue to block EC creation even after consent is granted. Follow-up: subscribe to `__gppapi` events (or a `storage` listener) and re-run on change.

crates/trusted-server-core/src/cookies.rs
ChristianPavilonis: Fixed in 3bbdb1d2 — the test name is back to `test_parse_cookies_to_jar_empty`.

crates/js/lib/src/integrations/sourcepoint/index.ts
ChristianPavilonis: Aligned implementation with the approved spec in 3bbdb1d2 — dropped `Max-Age=86400` so the mirrored Sourcepoint cookies are session-scoped, and kept the spec in sync.

crates/js/lib/src/integrations/sourcepoint/index.ts
ChristianPavilonis: Addressed in 3bbdb1d2 with documentation rather than changing behavior: I added an inline comment plus a spec note that this integration assumes a single active Sourcepoint property and mirrors the first valid `_sp_user_consent_*` entry.

crates/js/lib/src/integrations/sourcepoint/index.ts
ChristianPavilonis: Added that assumption in 3bbdb1d2 — the code now documents that Trusted Server is the intended writer for these mirrored cookies, so clearing the origin-scoped cookie is the expected behavior here.

crates/trusted-server-core/src/consent/gpp.rs
ChristianPavilonis: Kept the conservative OR-on-opt-out behavior and left the existing rustdoc on `decode_us_sale_opt_out()` as the source of truth for those aggregation semantics.

crates/trusted-server-core/src/consent/gpp.rs
ChristianPavilonis: Added a maintenance note in 3bbdb1d2 next to the manual match so future `iab_gpp` US-state additions are easier to catch during updates.

crates/js/lib/test/integrations/sourcepoint/index.test.ts
ChristianPavilonis: Updated the test helper in 3bbdb1d2 to use `Max-Age=0`, matching the production cookie-clearing convention.

ChristianPavilonis:
Addressed the requested review feedback in 3bbdb1d2.

Summary:
- added `sourcepoint` to the JS-only bundle list and covered it with a registry test
- removed the orphaned `ec/admin.rs` file
- removed the dead-code clippy offenders in test support / integration tests
- scoped this PR back down by reverting the unrelated creative-lockdown and Prebid user ID changes
- aligned Sourcepoint cookie mirroring with the approved session-cookie spec
- fixed the cookie test-name typo and the Sourcepoint test helper nit

Validation run locally:
- `cargo test --workspace`
- `cd crates/js/lib && npx vitest run`

I also replied to and resolved the inline threads above.

docs/guide/integrations/prebid.md
```diff
@@ -223,6 +223,97 @@ The build script (`build-all.mjs`) validates that each adapter exists in `prebid
...
 :::
 
+## User ID Modules
+
+Prebid's User ID Module resolves cross-publisher identifiers (SharedID, ID5, LiveRamp RampID, UID2, etc.) and exposes them via `pbjs.getUserIdsAsEids()`. The TSJS Prebid integration bundles the core `userId.js` module and a configurable set of ID submodules. When the publisher's origin-side `pbjs.setConfig({ userSync: { userIds: [...] } })` call runs during `processQueue()`, each listed submodule activates and begins resolving its ID asynchronously. After each auction the shim writes the resolved EIDs to a `ts-eids` cookie, which the Rust backend ingests into the Edge Cookie identity graph.
+
+### How it works
+
+1. `userId.js` is statically imported in `index.ts` — always bundled, not operator-configurable.
+2. The set of ID submodules is controlled by `TSJS_PREBID_USER_IDS` at build time and emitted into `_user_ids.generated.ts`.
```
aram356: 🔧 **wrench** — Stale Prebid User ID docs reference a removed env var.

The "Scope Sourcepoint consent PR" commit (`8a6df3af`) removed `generatePrebidUserIds()` from `build-all.mjs` along with the Prebid User ID design/plan docs. But this guide still claims:

- L233: "controlled by `TSJS_PREBID_USER_IDS` at build time and emitted into `_user_ids.generated.ts`"
- L247–250: examples of slim builds via the env var
- L253: "`build-all.mjs` validates that each exists … and generates `_user_ids.generated.ts`"
- L256: claims a `liveIntentIdSystem` denylist exists

None of this is true after the scoping commit. The env var is silently ignored. Operators will set it expecting a slim build and ship the full bundle instead.

**Fix:** delete the "User ID Modules" section (or at least the env var subsections + denylist warning), keeping only what's accurate post-scoping. If the build-time configurability is meant to come back, hold these docs until the generator returns.

crates/js/lib/src/integrations/prebid/_user_ids.generated.ts
```diff
@@ -0,0 +1,22 @@
+// Auto-generated by build-all.mjs — manual edits will be overwritten at build time.
```
aram356: 🔧 **wrench** — File header claims it is auto-generated, but it is not.

```
// Auto-generated by build-all.mjs — manual edits will be overwritten at build time.
```

The `8a6df3af` scoping commit removed `generatePrebidUserIds()` from `build-all.mjs`. This file is now hand-edited but still warns future contributors that their edits will be overwritten — which will either discourage edits or cause confusion when the rebuild does not in fact overwrite them.

**Fix:** either (a) restore the generator in `build-all.mjs` (preferred, since the doc still pitches `TSJS_PREBID_USER_IDS` as a feature), or (b) rename to `_user_ids.ts` and replace the header with a static-file comment that documents the curated default and the rationale for `liveIntentIdSystem` exclusion.

crates/trusted-server-core/src/integrations/registry.rs
```diff
@@ -800,14 +782,14 @@ impl IntegrationRegistry {
...
-    /// Always includes "creative" (JS-only, no Rust-side registration).
+    /// Always includes JS-only modules with no Rust-side registration.
     /// Excludes integrations that have no JS module (e.g., "nextjs").
     #[must_use]
     pub fn js_module_ids(&self) -> Vec<&'static str> {
         // Rust-only integrations with no corresponding JS module
         const JS_EXCLUDED: &[&str] = &["nextjs", "aps", "adserver_mock"];
         // JS-only modules always included (no Rust-side registration)
-        const JS_ALWAYS: &[&str] = &["creative"];
+        const JS_ALWAYS: &[&str] = &["creative", "sourcepoint"];
```
aram356, line 806: 🤔 **thinking** — Sourcepoint hardcoded as always-shipped will clobber `__gpp`/`__gpp_sid` set by other CMPs.

```rust
const JS_ALWAYS: &[&str] = &["creative", "sourcepoint"];
```

Every Trusted Server deployment ships and auto-runs the Sourcepoint module on every page load — regardless of which CMP the publisher uses. When no `_sp_user_consent_*` localStorage entry is found, the module unconditionally clears `__gpp` and `__gpp_sid` cookies (see `crates/js/lib/src/integrations/sourcepoint/index.ts:60-72`). Publishers using OneTrust, Didomi, or any CMP that writes `__gpp` directly to cookies (allowed by the GPP spec) will have those cookies erased on every page load.

The design doc says "same pattern as `creative`," but `creative` is universally applicable while `sourcepoint` is publisher-CMP-specific.

**Suggested fixes:**
- (a) Make sourcepoint opt-in via `[integrations.sourcepoint] enabled = true` with a small Rust `IntegrationRegistration` whose only job is to gate JS inclusion.
- (b) Only clear cookies the module previously wrote — track via a marker (e.g. `_ts_gpp_src=sp`) written alongside `__gpp` and check it before clearing.

crates/js/lib/src/integrations/sourcepoint/index.ts
```diff
@@ -0,0 +1,95 @@
...
+  log.info('sourcepoint: mirrored GPP consent to cookies', {
+    gppLength: gppString.length,
+    sections: applicableSections,
+  });
+
+  return true;
+}
+
+if (typeof window !== 'undefined') {
+  mirrorSourcepointConsent();
```
aram356: 🤔 **thinking** — First page load race: mirror runs at script load before Sourcepoint CMP populates localStorage.

The integration runs once at script load and is included in the immediate (synchronous) bundle. If the bundle loads in `<head>` before Sourcepoint's CMP script populates `_sp_user_consent_*`, the mirror reads empty localStorage, writes `Max-Age=0` clearing cookies for `__gpp`/`__gpp_sid`, and gives up.

Combined with `allows_ec_creation` fail-closed for `UsState` with no signals (`consent/mod.rs:520-523`), users in regulated US states get **EC blocked on first page load** even though Sourcepoint is properly installed.

The design spec ("integration runs once — no polling or event listeners") acknowledges this but underestimates the impact for first-time visitors.

**Suggested fixes:**
- Register sourcepoint with `with_deferred_js()` so it loads after the CMP.
- Listen for Sourcepoint's `__sp__` consent-ready event.
- Retry once on `DOMContentLoaded` / short timer when initial scan finds nothing.

crates/trusted-server-core/src/consent/mod.rs
```diff
@@ -494,15 +497,35 @@ pub fn allows_ec_creation(ctx: &ConsentContext) -> bool {
...
-            // US: opt-out model — allow unless user explicitly opted out.
+            // GPC is an independent opt-out signal — it always blocks EC
+            // creation regardless of what the US Privacy string says.
+            if ctx.gpc {
+                return false;
+            }
+            // When a CMP uses TCF in the US (e.g. Didomi), respect the
+            // TCF Purpose 1 decision — this is an explicit opt-in signal.
+            if let Some(tcf) = effective_tcf(ctx) {
+                return tcf.has_storage_consent();
```
aram356, line 494: 🤔 **thinking** — TCF presence silently overrides explicit GPP US `sale_opt_out=Yes` in US states.

```rust
if let Some(tcf) = effective_tcf(ctx) {
    return tcf.has_storage_consent();
}
// Check GPP US section for sale opt-out.
```

In `Jurisdiction::UsState`, `effective_tcf(ctx)` short-circuits with `tcf.has_storage_consent()` and never falls through to the GPP US opt-out check. The test `ec_us_state_tcf_takes_priority_over_gpp_us` (L1228) asserts this is intentional.

It is defensible if you treat TCF Purpose 1 as the authoritative storage opt-in for the EC cookie itself — but the user has *explicitly* told the CMP "don't sell my info" via the GPP US section, and EC is itself a tracking identifier. Some CMPs (Didomi, Sourcepoint) will write both sections in the US for spec compliance. The current precedence may surprise privacy-counsel reviewers.

**Suggested fixes:**
- Document this precedence decision explicitly in `docs/superpowers/specs/2026-04-15-sourcepoint-gpp-consent-design.md` (currently silent on the conflict).
- Add a config knob: `consent.us_tcf_priority = strict|permissive` so publishers in stricter regimes can opt into US-section-wins behavior.

crates/js/lib/src/integrations/sourcepoint/index.ts
```diff
@@ -0,0 +1,95 @@
...
+      }
+    } catch {
+      log.debug('sourcepoint: failed to parse localStorage value', { key });
+    }
+  }
+  return null;
+}
+
+function writeCookie(name: string, value: string): void {
+  document.cookie = `${name}=${value}; path=/; Secure; SameSite=Lax`;
```
aram356, line 146: 🤔 **thinking** — Session-scoped cookie + run-once mirror leaves stale `__gpp` after mid-session retraction.

```ts
document.cookie = `${name}=${value}; path=/; Secure; SameSite=Lax`;
```

`writeCookie` deliberately omits `Max-Age` (per spec). Combined with "runs once on script load," if the user retracts consent mid-session via the Sourcepoint CMP UI, the localStorage entry updates but `__gpp` does not refresh until the next full page navigation. Subsequent same-tab requests read stale consent.

**Suggested fix:** hook Sourcepoint's `onConsentUpdate` callback or rerun the mirror on `visibilitychange` so the cookie tracks localStorage within a session.

crates/trusted-server-core/src/consent/gpp.rs
```diff
@@ -72,10 +72,13 @@ pub fn decode_gpp_string(gpp_string: &str) -> Result<GppConsent, Report<ConsentD
     let eu_tcf = decode_tcf_from_gpp(&parsed);
 
     // The GPP header version is always 1 for current spec.
+    let us_sale_opt_out = decode_us_sale_opt_out(&parsed);
```
aram356, line 74: ♻️ **refactor** — Stale comment displaced by `us_sale_opt_out` insertion.

```rust
// The GPP header version is always 1 for current spec.
let us_sale_opt_out = decode_us_sale_opt_out(&parsed);
```

The comment was originally above `version: 1` inside the struct literal, but the new `us_sale_opt_out` line was inserted between them. It now reads as if it documents `us_sale_opt_out`.

**Fix:** move the comment back next to `version: 1,` in the struct literal, or delete it — `version: 1` is self-documenting.

crates/js/lib/src/integrations/sourcepoint/index.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+function clearCookie(name: string): void {
+  // Trusted Server is the only intended writer for these mirrored cookies, so
+  // clearing the origin-scoped cookie is sufficient for this integration.
+  document.cookie = `${name}=; path=/; Secure; SameSite=Lax; Max-Age=0`;
+}
```
aram356, line 151: ♻️ **refactor** — Make the clearing logic CMP-safe by tracking write source.

If finding #3 (always-shipped sourcepoint) cannot be addressed in this PR, at least make the clearing safe in mixed-CMP scenarios:

1. On `writeCookie`, also write a marker cookie: `_ts_gpp_src=sp; path=/; Secure; SameSite=Lax`.
2. In `clearCookie`, check `document.cookie` for `_ts_gpp_src=sp` and skip clearing if absent.

This way, a publisher using a different CMP that happens to ship this module will not have their CMP's `__gpp` cookies clobbered.

crates/js/lib/src/integrations/sourcepoint/index.ts
```diff
@@ -0,0 +1,95 @@
...
+  });
+
+  return true;
+}
+
+if (typeof window !== 'undefined') {
+  mirrorSourcepointConsent();
+}
+
+export default mirrorSourcepointConsent;
```
aram356: ⛏ **nitpick** — Unused/inconsistent default export.

```ts
export default mirrorSourcepointConsent;
```

Nothing imports the default — the module self-initializes via `if (typeof window !== 'undefined')` at L91-93, and tests use the named export. No other Trusted Server integration exports a default. Drop it for consistency.

aram356 (changes requested):
## Summary

Follow-up review focused on areas not covered in the prior pass. Sourcepoint flow (mirror + GPP US decoding + EC gating) is functional and well-tested, but two doc/file-stale issues from the scoping commit (`8a6df3af`) need correction before merge, and the always-shipped Sourcepoint module has cross-CMP failure modes worth designing around.

PR effective scope note: vs `main` the diff is 80 commits / 111 files, because the intended base (`feature/edge-cookies-final`) has merged. The Sourcepoint-specific changes are a small subset (~14 files); reviewers landing on this PR via GitHub should read the description with that in mind.

## Blocking

### 🔧 wrench

- **Stale Prebid User ID docs reference removed `TSJS_PREBID_USER_IDS` env var** (`docs/guide/integrations/prebid.md:226-261`)
- **`_user_ids.generated.ts` claims to be auto-generated but is now hand-edited** (`crates/js/lib/src/integrations/prebid/_user_ids.generated.ts:1`)

## Non-blocking

### 🤔 thinking

- **Sourcepoint hardcoded as always-shipped will clobber `__gpp` set by other CMPs** (`registry.rs:792` + `sourcepoint/index.ts:60-72`)
- **First page load race: mirror runs before Sourcepoint CMP populates localStorage** (`sourcepoint/index.ts:91-93`)
- **TCF presence silently overrides explicit GPP US `sale_opt_out=Yes` in US states** (`consent/mod.rs:506-515`)
- **Session-scoped cookie + run-once mirror leaves stale `__gpp` after mid-session retraction** (`sourcepoint/index.ts:39`)

### ♻️ refactor

- **Stale comment displaced by `us_sale_opt_out` insertion** (`consent/gpp.rs:74-75`)
- **Make clearing logic CMP-safe by tracking write source** (`sourcepoint/index.ts:42-46`)

### ⛏ nitpick

- **Unused/inconsistent default export** (`sourcepoint/index.ts:95`)

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS (1001 lib + 21 misc)
- js tests: PASS (294)

docs/guide/integrations/prebid.md
ChristianPavilonis: Fixed in a2c6d831 — deleted the stale build-time User ID Module selection docs and replaced them with the current behavior. The guide no longer references `TSJS_PREBID_USER_IDS`, `_user_ids.generated.ts`, generator validation, or the `liveIntentIdSystem` denylist.

crates/js/lib/src/integrations/prebid/_user_ids.generated.ts
ChristianPavilonis: Confirmed this is already outdated in the current branch: `crates/js/lib/src/integrations/prebid/_user_ids.generated.ts` does not exist locally and is not present in the current PR diff. No generated-header claim remains to fix.

crates/trusted-server-core/src/integrations/registry.rs
ChristianPavilonis: Addressed in a2c6d831 with option (b): Sourcepoint remains in the always-shipped JS list, but the client now writes `_ts_gpp_src=sp` alongside Sourcepoint-mirrored GPP cookies and only clears `__gpp` / `__gpp_sid` when that marker is present. That prevents clobbering GPP cookies owned by other CMPs.

crates/js/lib/src/integrations/sourcepoint/index.ts
ChristianPavilonis: Addressed in a2c6d831 by adding bounded first-load retries. The module still tries immediately, but when no Sourcepoint payload is found it retries on `DOMContentLoaded` if still loading and once more after a short timeout. Added Vitest coverage for data appearing shortly after module initialization.

crates/trusted-server-core/src/consent/mod.rs
ChristianPavilonis: Documented in a2c6d831 rather than changing policy in this PR. The Sourcepoint design spec now explicitly states that in US-state jurisdictions GPC blocks first, then effective TCF Purpose 1 remains authoritative for EC storage consent, and GPP US sale opt-out is consulted only when no effective TCF signal is present. I also added a code comment pointing to that documented precedence decision.

crates/js/lib/src/integrations/sourcepoint/index.ts
ChristianPavilonis: Addressed in a2c6d831. Sourcepoint mirroring now re-runs on page focus and visible `visibilitychange`, so same-session updates/retractions in localStorage refresh the session cookies. Added tests for refreshing updated consent and clearing Sourcepoint-owned cookies after consent retraction.

crates/trusted-server-core/src/consent/gpp.rs
ChristianPavilonis: Fixed in a2c6d831 — moved the GPP version comment back next to `version: 1` in the `GppConsent` struct literal.

crates/js/lib/src/integrations/sourcepoint/index.ts
ChristianPavilonis: Addressed in a2c6d831. The module now tracks Sourcepoint ownership with `_ts_gpp_src=sp`; writes set the marker, and clearing is skipped unless the marker is present. Added tests covering both non-Sourcepoint cookies being preserved and Sourcepoint-owned cookies being cleared.

crates/js/lib/src/integrations/sourcepoint/index.ts
ChristianPavilonis: Fixed in a2c6d831 — removed the unused default export and kept the named exports used by tests/runtime initialization.

crates/js/lib/src/integrations/sourcepoint/index.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  const { gppString, applicableSections } = payload.gppData;
+  if (!gppString) {
+    clearSourcepointCookies();
+    log.debug('sourcepoint: gppString is empty');
+    return false;
+  }
+
+  writeCookie(GPP_SOURCE_COOKIE_NAME, GPP_SOURCE_SOURCEPOINT);
+  writeCookie(GPP_COOKIE_NAME, gppString);
```
aram356, line 236: 🔧 **wrench** — Sourcepoint mirror clobbers other CMPs' `__gpp` on the *write* path.

The `_ts_gpp_src=sp` marker added in `a2c6d831` correctly addresses the prior reviewer's *clearing*-clobber concern (#3164911741), but on the very first run with a valid Sourcepoint payload, this code unconditionally overwrites whatever `__gpp` and `__gpp_sid` another CMP wrote on the same origin. The marker only protects the symmetric *clear* path.

Suggested fix (one of):

```ts
// Option A: only write when we own the cookie or it's empty.
const existing = readCookie(GPP_COOKIE_NAME);
if (existing && existing !== gppString && !hasSourcepointMarker()) {
  log.debug('sourcepoint: __gpp already set by another writer — skipping');
  return false;
}
```

Option B: keep current behavior but document explicitly in the spec that Sourcepoint writes always win on its origin (the spec currently only documents *clearing* safety, not writing safety).

crates/js/lib/src/integrations/sourcepoint/index.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  if (Array.isArray(applicableSections) && applicableSections.length > 0) {
+    writeCookie(GPP_SID_COOKIE_NAME, applicableSections.join(','));
+  } else if (hasSourcepointMarker()) {
```
aram356: 🤔 **thinking** — `hasSourcepointMarker()` here is dead code.

This branch only executes after we just wrote `_ts_gpp_src=sp` on line 114, so the marker is always present. Either drop the conditional (`else { clearCookie(GPP_SID_COOKIE_NAME); }`) or hoist the marker check to before the marker write so it actually gates anything.

crates/js/lib/src/integrations/sourcepoint/index.ts
```diff
@@ -0,0 +1,151 @@
...
+function scheduleInitialRetry(): void {
+  const retry = (): void => {
+    mirrorSourcepointConsent();
+  };
+
+  if (document.readyState === 'loading') {
+    document.addEventListener('DOMContentLoaded', retry, { once: true });
+  }
+
+  window.setTimeout(retry, INITIAL_RETRY_DELAY_MS);
```
aram356: 🤔 **thinking** — Real `setTimeout` at module load leaks across Vitest tests.

`window.setTimeout(retry, 500)` is scheduled at import time with no `clearTimeout` and no idempotency guard beyond `initialized`. Every Vitest test that imports this module schedules one of these. Only the explicit retry test uses fake timers; other tests can have the deferred retry fire mid-assertion against whatever localStorage state they've set up. Tests pass today because the suite is fast, but the isolation is brittle.

Suggested fix: track the timeout id and clear it once a mirror succeeds (and on subsequent successful mirrors), e.g.:

```ts
let retryTimer: number | undefined;

function scheduleInitialRetry(): void {
  if (retryTimer !== undefined) return;
  retryTimer = window.setTimeout(() => {
    retryTimer = undefined;
    mirrorSourcepointConsent();
  }, INITIAL_RETRY_DELAY_MS);
  // …
}
```

crates/js/lib/src/integrations/sourcepoint/index.ts
```diff
@@ -0,0 +1,151 @@
...
+  }
+}
+
+function scheduleInitialRetry(): void {
+  const retry = (): void => {
+    mirrorSourcepointConsent();
+  };
+
+  if (document.readyState === 'loading') {
+    document.addEventListener('DOMContentLoaded', retry, { once: true });
```
aram356, line 194: 🤔 **thinking** — `DOMContentLoaded` + `setTimeout(500)` can fire `mirrorSourcepointConsent` twice.

When `readyState === 'loading'`, both handlers register and both run — they don't cancel each other. Idempotent today (reading localStorage and writing the same cookies has no observable side effect), but if the mirror later gains a non-idempotent step (event emit, beacon, counter), this would silently double. Cheap fix: a `retried` boolean inside `scheduleInitialRetry` that short-circuits the second call.

crates/js/lib/src/integrations/prebid/index.ts
```diff
@@ -451,6 +451,7 @@ function fitAuctionEidsToCookie(eids: AuctionEid[]): AuctionEid[] | undefined {
 function syncPrebidEidsCookie(): void {
   try {
     if (typeof pbjs.getUserIdsAsEids !== 'function') {
+      clearPrebidEidsCookie();
```
aram356, line 536: 🤔 **thinking** — Behavior change buried in a Sourcepoint PR.

Previously this branch was a silent no-op; now it clears the `ts-eids` cookie. With `userId.js` unconditionally imported at line 20, the missing-function branch should be unreachable, so this is defensive — but it's a behavior change that doesn't belong with the Sourcepoint feature. If anything else ever seeds `ts-eids` (server response, manual injection), this will now wipe it.

Either (a) split into a separate PR with its own justification, or (b) add a one-line comment here explaining *why* clearing on missing `getUserIdsAsEids` is correct (i.e. that no User ID Module = no EIDs to forward, so the cookie must not be stale).

crates/trusted-server-core/src/consent/gpp.rs
```diff
@@ -100,6 +103,84 @@ fn decode_tcf_from_gpp(parsed: &iab_gpp::v1::GPPString) -> Option<TcfConsent> {
...
+/// Iterates through section IDs looking for any in the US range (7–23),
+/// decodes each US section, and aggregates the result conservatively:
+///
+/// - `Some(true)` if any decodable US section says the user opted out of sale
+/// - `Some(false)` if at least one decodable US section says they did not opt
+///   out and none say they opted out
+/// - `None` if no US section is present or no decodable US section yields a
+///   usable `sale_opt_out` signal
+fn decode_us_sale_opt_out(parsed: &iab_gpp::v1::GPPString) -> Option<bool> {
+    let mut saw_not_opted_out = false;
```
aram356: ♻️ **refactor** — Aggregation can be expressed without `saw_not_opted_out`.

```rust
let mut result: Option<bool> = None;
for us_section_id in parsed
    .section_ids()
    .filter(|id| US_SECTION_ID_RANGE.contains(&(**id as u16)))
{
    match parsed.decode_section(*us_section_id) {
        Ok(section) => match us_sale_opt_out_from_section(&section) {
            Some(true) => return Some(true),
            Some(false) => result = Some(false),
            None => {}
        },
        Err(e) => log::warn!("Failed to decode US GPP section {us_section_id}: {e}"),
    }
}
result
```

Same behavior, drops the intermediate flag. Optional.

docs/superpowers/specs/2026-04-15-sourcepoint-gpp-consent-design.md
```diff
@@ -0,0 +1,170 @@
...
+existing CMP behavior where TCF Purpose 1 is the explicit storage/access signal
+for the EC cookie itself. Publishers that need US-section-wins behavior should
+raise that as a separate consent-policy configuration change.
+
+### 4. Files touched
+
+| File | Change |
+|---|---|
+| `crates/js/lib/src/integrations/sourcepoint/index.ts` | New — localStorage auto-discovery, cookie mirroring |
+| `crates/js/lib/src/integrations/sourcepoint/index.test.ts` | New — Vitest tests |
```
aram356: ⛏ **nitpick** — Wrong path. Tests live at `crates/js/lib/test/integrations/sourcepoint/index.test.ts` (per the project's `src` vs `test` split), not under `src/integrations/`.

aram356 (changes requested):
## Summary

Follow-up review focused on what's new since the last pass and on items I think were missed. The marker-cookie response in `a2c6d831` substantively addresses most prior feedback (clearing is now CMP-safe, retries cover the first-load race, the stale comment is fixed, the default export is gone). The remaining issues are: the symmetric *write-path* clobber that the marker doesn't cover, a real CI failure inherited from the merge-base, and continued scope creep beyond the Sourcepoint feature.

Local verification on the PR HEAD passes: `cargo check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test --workspace`, and `npx vitest run` (301 tests). CI fails because the `prepare integration artifacts` job hits a build-time validation panic.

## Blocking

### 🔧 wrench

- **CI: integration test passphrase below 32 chars.** `prepare integration artifacts` panics in [`build.rs:42`](crates/trusted-server-core/build.rs#L42) via [`Ec::validate_passphrase`](crates/trusted-server-core/src/settings.rs#L425) — `MIN_PASSPHRASE_LENGTH` was bumped from 8 to 32 in commit `8e08e642` (which is on the merge-base, hence inherited), but the integration env still sets `TRUSTED_SERVER__EC__PASSPHRASE=integration-test-ec-secret` (26 chars) in:
  - [`.github/actions/setup-integration-test-env/action.yml:83`](.github/actions/setup-integration-test-env/action.yml#L83)
  - [`.github/workflows/test.yml:58`](.github/workflows/test.yml#L58)
  - [`scripts/integration-tests.sh:56`](scripts/integration-tests.sh#L56)
  - [`scripts/integration-tests-browser.sh:35`](scripts/integration-tests-browser.sh#L35)

  These files are not in this PR's diff, so the fix is either land it here or in a small precursor PR — but the PR cannot merge with red CI. Suggested value: `integration-test-ec-secret-padded-32` (or longer).

- **Sourcepoint mirror clobbers other CMPs' `__gpp` on the *write* path** — see inline at [`sourcepoint/index.ts:115`](crates/js/lib/src/integrations/sourcepoint/index.ts#L115).

## Non-blocking

### 🤔 thinking

- **Scope: PR still bundles unrelated commits after the partial scope-down.** `8e3ff806` removed the Prebid User ID feature, but the diff vs `feature/edge-cookies-final` still contains `fedcc347` ("Wire request signing… and Move logging initialization") which re-applies already-merged PRs #609/#610 — once the base branch is rebased onto current `main`, this commit is duplicate work and a guaranteed merge-conflict source. `24d45156` ("Remove generated Prebid user ID shim") and the `clearPrebidEidsCookie` change at [`prebid/index.ts:454`](crates/js/lib/src/integrations/prebid/index.ts#L454) are also outside the Sourcepoint feature. Suggestion: rebase `feature/edge-cookies-final` onto current `main` first to drop `fedcc347`, and split the Prebid User ID shim removal into its own PR.

- **`__gpp_sid` marker-check on the write path is dead code** — see inline.
- **`scheduleInitialRetry` leaks a real `setTimeout` across Vitest tests** — see inline.
- **`DOMContentLoaded` + `setTimeout(500)` can double-fire `mirrorSourcepointConsent`** — see inline.
- **Prebid `clearPrebidEidsCookie` on missing `getUserIdsAsEids` is a behavior change** — see inline.

### ♻️ refactor

- **`decode_us_sale_opt_out` accumulator can drop `saw_not_opted_out`** — see inline.

### 🌱 seedling

- **Cookie size limits not enforced client-side.** Server caps at `MAX_GPP_STRING_LEN: 8192` in [`consent/gpp.rs:33`](crates/trusted-server-core/src/consent/gpp.rs#L33), but browsers cap individual cookies near ~4096 bytes. A long Sourcepoint GPP payload (compound multi-section) can be silently truncated by the browser, leaving the server unable to decode. Worth a follow-up: cap and log client-side instead of writing a doomed cookie.

### 📌 out of scope

- **TCF-vs-GPP-US precedence is documented but not configurable.** The spec section "TCF-before-GPP precedence is intentional" makes a permanent policy choice some publishers may want to override. Tracking issue would prevent this from getting buried.
- **Sourcepoint integration runs on every Trusted Server page** ([`registry.rs:803`](crates/trusted-server-core/src/integrations/registry.rs#L803), [`sourcepoint/index.ts:151`](crates/js/lib/src/integrations/sourcepoint/index.ts#L151)). With the marker cookie this is correctness-safe, but every visitor on every site pays ~3 KB of payload + a localStorage scan + cookie writes for a feature that only applies to Sourcepoint customers. Worth a follow-up: opt-in via publisher config, or only register listeners after a first valid mirror.

### ⛏ nitpick

- **Spec lists test file at the wrong path** — see inline.

## CI Status

- fmt: PASS (local)
- clippy: PASS (local, `--all-targets --all-features -- -D warnings`)
- rust tests: PASS (local, 1073 + 21 + 19)
- js tests: PASS (local, 301 tests / 24 files)
- integration tests: **FAIL** (`prepare integration artifacts` panics; see Blocking above)

crates/js/lib/src/integrations/sourcepoint/index.ts
ChristianPavilonis: Fixed in e7b561a2 with option A: Sourcepoint now reads any existing __gpp cookie before writing and skips mirroring when a different non-Sourcepoint-owned value is present. Added Vitest coverage that preserves another CMP's __gpp/__gpp_sid and leaves the Sourcepoint marker unset. The spec now notes that the marker protects both clearing and overwriting.

crates/js/lib/src/integrations/sourcepoint/index.ts
ChristianPavilonis: Fixed in e7b561a2 by dropping the dead marker check after the marker write; the no-applicable-sections branch now directly clears the stale __gpp_sid after Sourcepoint ownership has been established.

crates/js/lib/src/integrations/sourcepoint/index.ts
ChristianPavilonis: Fixed in e7b561a2. The initial retry now tracks the timeout id, clears the pending timer after a successful mirror, and guards against scheduling duplicate retry cycles. Added Vitest coverage for clearing a pending retry after a successful manual mirror.

crates/js/lib/src/integrations/sourcepoint/index.ts
ChristianPavilonis: Fixed in e7b561a2. The initial retry path now uses a shared initialRetryDone guard so DOMContentLoaded and the timeout cannot both run the mirror. Added Vitest coverage that fires DOMContentLoaded, advances the timer, and verifies only the first retry result persists.

crates/js/lib/src/integrations/prebid/index.ts
ChristianPavilonis: Addressed in e7b561a2 with option (b): added a one-line comment explaining that when Prebid cannot expose EIDs, stale auction fallback IDs must not persist in the ts-eids cookie.

crates/trusted-server-core/src/consent/gpp.rs
ChristianPavilonis: Implemented in e7b561a2. decode_us_sale_opt_out now uses an Option<bool> accumulator and preserves the same conservative aggregation behavior.

docs/superpowers/specs/2026-04-15-sourcepoint-gpp-consent-design.md
ChristianPavilonis: Fixed in e7b561a2. The spec now points to crates/js/lib/test/integrations/sourcepoint/index.test.ts.

PR #646, opened 2026-04-20
Add JA4/TLS fingerprint debug endpoint at /_ts/debug/ja4

### LINKED ISSUE #645: Add JA4/TLS fingerprint debug endpoint
## Description

Add a debug endpoint `GET /_ts/debug/ja4` to the Fastly adapter to inspect JA4 TLS fingerprints, H2 fingerprints, cipher suites, TLS versions, and Client Hints in production. This endpoint is needed to evaluate TLS fingerprinting as a cross-browser identity signal and as a validated-browser signal in OpenRTB ext fields.

## Done When

- [ ] Endpoint `GET /_ts/debug/ja4` returns plain-text TLS/browser fingerprint data

## Affected Area

Fastly runtime

prk-Jr:
## Summary

- Adds a temporary `GET /_ts/debug/ja4` endpoint to the Fastly adapter to surface JA4 TLS fingerprint, H2 fingerprint, cipher suite, TLS version, user-agent, and selected Client Hints for browser fingerprint evaluation
- Keeps the endpoint response readable when Fastly omits individual TLS or Client Hint values by using explicit fallback text
- Adds route-level regression coverage so the adapter test suite exercises both the response builder and the `/_ts/debug/ja4` router path

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-adapter-fastly/src/main.rs` | Added `build_ja4_debug_response`, registered `GET /_ts/debug/ja4`, and added unit coverage for fallback response fields |
| `crates/trusted-server-adapter-fastly/src/route_tests.rs` | Added a route-level test covering `/_ts/debug/ja4` dispatch, content type, and fallback body values |

## Closes

Closes #645

## Test plan

- [x] `cargo build --package trusted-server-adapter-fastly --release --target wasm32-wasip1`
- [x] GitHub Actions checks green for `cargo test`, `vitest`, `integration tests`, `browser integration tests`, `format-docs`, `format-typescript`, `cargo fmt`, and CodeQL

## Checklist

- [x] No `unwrap()` in production code - use `expect("should ...")`
- [x] New code has tests


crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -186,6 +213,9 @@ async fn route_request(
             }
         }
 
+        // Temporary JA4/TLS debug endpoint for browser fingerprint inspection.
+        (Method::GET, "/_ts/debug/ja4") => Ok(build_ja4_debug_response(&req)),
```
ChristianPavilonis: **Gate the debug endpoint behind explicit configuration**

The new `GET /_ts/debug/ja4` route is currently reachable without authentication or an explicit debug gate. Because it reflects Fastly-observed JA4/H2/TLS details back to same-origin browser JavaScript, any script running on the publisher page could fetch this endpoint and exfiltrate fingerprinting values that browser JS normally cannot read directly.

**Suggestion:** Add an explicit `trusted-server.toml` configuration flag for this debug endpoint, disabled by default, and return `404` or another non-disclosing response when the flag is off. That keeps the endpoint available for intentional Fastly/browser investigation while avoiding a public-by-default fingerprint reflection API.

ChristianPavilonis (commented):
## Summary

Thanks for adding the temporary JA4/TLS debug endpoint and route coverage. The implementation is straightforward and CI is green, but before this ships I think the endpoint should be explicitly gated so it cannot become a public same-origin fingerprint reflection API by default.

## Additional finding folded into body

**Document the debug endpoint and config gate**

If this endpoint is retained behind a new `trusted-server.toml` debug flag, please also document the flag and endpoint behavior in the API/config docs. The docs should make clear that this is temporary/debug-only, disabled by default, what fields are returned (`ja4`, `h2_fp`, `cipher`, `tls_version`, `user-agent`, `ch-mobile`, `ch-platform`), the fallback values, and that the response uses `Cache-Control: no-store, private`.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -186,6 +213,15 @@ async fn route_request(
             }
         }
 
+        // JA4/TLS debug endpoint — only active when debug.ja4_endpoint_enabled = true.
+        (Method::GET, "/_ts/debug/ja4") => {
```
aram356: 🔧 **wrench** — Public-by-default fingerprint reflection oracle

`enforce_basic_auth` only protects paths matched by an explicit `[[handlers]]` regex (see `crates/trusted-server-core/src/auth.rs:32`). The default config does not include `/_ts/debug`, so this route is reachable unauthenticated by anyone — including arbitrary same-origin scripts on the publisher page. It then reflects Fastly-observed JA4, H2 fingerprint, cipher, TLS version, and Client Hints back as same-origin readable text. Browser JS cannot normally read JA4/cipher/TLS-version directly, so this endpoint hands a fingerprinting third-party script exactly the values that being privileged-to-Fastly was protecting.

This is also at odds with trusted-server's privacy-preserving positioning.

This was raised in the prior review round and remains unaddressed.

**Fix (any of):**

1. Gate behind a new `[debug] ja4_endpoint = false` flag in `trusted-server.toml` (default off). Return 404 when off so the path does not even disclose its existence.
2. Restrict to staging only — short-circuit when `std::env::var(ENV_FASTLY_IS_STAGING).as_deref() != Ok("1")`.
3. Move the path under an existing admin handler (e.g. `/_ts/admin/debug/ja4` covered by an `^/_ts/admin` basic-auth handler) and document the required `[[handlers]]` entry.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -109,6 +109,33 @@ fn main() {
...
+         h2_fp:       {h2}\n\
+         cipher:      {cipher}\n\
+         tls_version: {tls_version}\n\
+         user-agent:  {ua}\n\
+         ch-mobile:   {ch_mobile}\n\
+         ch-platform: {ch_platform}\n"
+    );
+
+    Response::from_status(fastly::http::StatusCode::OK)
+        .with_header(header::CACHE_CONTROL, "no-store, private")
```
aram356, line 156: 🔧 **wrench** — `Cache-Control` is silently overridable by operator `[response_headers]`

The handler sets `Cache-Control: no-store, private`, but the response then flows through `finalize_response`, whose final loop unconditionally calls `response.set_header(key, value)` for every operator-configured `[response_headers]` entry (`crates/trusted-server-adapter-fastly/src/main.rs:349-351`). If an operator configures `Cache-Control = "public, max-age=..."` (a common edge default), our `no-store, private` is silently replaced and a sensitive fingerprint response can land in shared caches.

The route-level test does not catch this because `create_test_settings()` does not configure `[response_headers]`. The `finalize_response` doc says "operators can intentionally override any managed header" — that contract is acceptable for cosmetic headers, not for cache-control on a debug endpoint that exposes TLS metadata.

**Fix:** Handle this route the same way `/health` is handled — before `route_request`/`finalize_response` (top of `main()`, after `init_logger`):

```rust
if req.get_method() == Method::GET && req.get_path() == "/_ts/debug/ja4" {
    build_ja4_debug_response(&req).send_to_client();
    return;
}
```

That also avoids paying the cost of building the auction orchestrator and integration registry for a simple debug probe, and pairs naturally with the staging-only gate suggested in the other 🔧.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -320,3 +356,63 @@ fn finalize_response(settings: &Settings, geo_info: Option<&GeoInfo>, response:
...
+            body.contains("tls_version: unavailable"),
+            "should include TLS version fallback"
+        );
+        assert!(
+            body.contains("user-agent:  none"),
+            "should include user-agent fallback"
+        );
+        assert!(
+            body.contains("ch-mobile:   not sent"),
+            "should include sec-ch-ua-mobile fallback"
```
aram356, line 428: ♻️ **refactor** — Unit test duplicates the route-level test

This test and `ja4_debug_route_returns_plain_text_fallback_response` in `route_tests.rs:253-317` assert identical things — status 200, `text/plain; charset=utf-8`, `Cache-Control: no-store, private`, and the seven fallback substrings. The route-level test additionally exercises dispatch, which is the load-bearing signal. This unit test is strictly subsumed.

**Suggestion:** delete this test, or repurpose it to assert formatting output for *populated* TLS values — but that is currently not testable because `Request::get(...)` does not stub `get_tls_ja4()`. Deletion is the simpler call.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -320,3 +356,63 @@ fn finalize_response(settings: &Settings, geo_info: Option<&GeoInfo>, response:
...
+        let req = Request::get("https://example.com/_ts/debug/ja4");
+
+        let mut response = build_ja4_debug_response(&req);
+
+        assert_eq!(
+            response.get_status(),
+            fastly::http::StatusCode::OK,
+            "should return 200 OK"
+        );
+        assert_eq!(
```
aram356, line 393: ⛏ **nitpick** — Use `header::CACHE_CONTROL` constant

Production code uses `header::CACHE_CONTROL` (line 134); the test uses the literal `"cache-control"`. Use the constant here too for rename safety.

Same applies to `route_tests.rs:282`.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -109,6 +109,33 @@ fn main() {
...
+fn build_ja4_debug_response(req: &Request) -> Response {
+    let ja4 = req.get_tls_ja4().unwrap_or("unavailable");
+    let h2 = req.get_client_h2_fingerprint().unwrap_or("unavailable");
+    let cipher = req.get_tls_cipher_openssl_name().unwrap_or("unavailable");
+    let tls_version = req.get_tls_protocol().unwrap_or("unavailable");
+    let ua = req.get_header_str("user-agent").unwrap_or("none");
+    let ch_mobile = req.get_header_str("sec-ch-ua-mobile").unwrap_or("not sent");
+    let ch_platform = req
+        .get_header_str("sec-ch-ua-platform")
+        .unwrap_or("not sent");
```
aram356: ⛏ **nitpick** — Extract magic strings to `const`s

Fallback values (`"unavailable"`, `"not sent"`, `"none"`) and label strings (`"ja4:         "`, etc.) are duplicated across this function and both tests. A handful of `const`s near `build_ja4_debug_response` would prevent drift if the format changes.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -109,6 +109,33 @@ fn main() {
...
+         tls_version: {tls_version}\n\
+         user-agent:  {ua}\n\
+         ch-mobile:   {ch_mobile}\n\
+         ch-platform: {ch_platform}\n"
+    );
+
+    Response::from_status(fastly::http::StatusCode::OK)
+        .with_header(header::CACHE_CONTROL, "no-store, private")
+        .with_content_type(fastly::mime::TEXT_PLAIN_UTF_8)
+        .with_body(body)
```
aram356, line 162: 🤔 **thinking** — Defensive `Vary` header

Output varies by `User-Agent`, `sec-ch-ua-mobile`, `sec-ch-ua-platform`, and TLS metadata. With `no-store` set this is moot for shared caches — but if the cache-control gets overridden (see the 🔧 on line 134), `Vary: User-Agent, Sec-CH-UA-Mobile, Sec-CH-UA-Platform` would be a useful defense-in-depth backstop. Lower priority once the override path is fixed.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -109,6 +109,33 @@ fn main() {
     }
 }
 
+fn build_ja4_debug_response(req: &Request) -> Response {
+    let ja4 = req.get_tls_ja4().unwrap_or("unavailable");
+    let h2 = req.get_client_h2_fingerprint().unwrap_or("unavailable");
+    let cipher = req.get_tls_cipher_openssl_name().unwrap_or("unavailable");
+    let tls_version = req.get_tls_protocol().unwrap_or("unavailable");
```
aram356: 📝 **note** — Some of these fields are already on the platform abstraction

`tls_protocol` and `tls_cipher` are exposed by the platform abstraction via `runtime_services.client_info()` (see `route_tests.rs:175-176`). JA4 and H2 fingerprint are Fastly-specific and not on the abstraction yet, so direct Fastly calls are unavoidable for those. Calling `req.get_tls_protocol()` / `req.get_tls_cipher_openssl_name()` directly here is acceptable because this file lives in the adapter-fastly crate, but using the abstraction for fields that have one would be more consistent with how the rest of the adapter reads TLS metadata.

No action required if you'd rather keep one consistent source for all five values in this single handler.

aram356 (changes requested):
## Summary

Adds a `GET /_ts/debug/ja4` endpoint to the Fastly adapter for inspecting JA4/H2/TLS fingerprint metadata and Client Hints. CI is green and the route-level test correctly exercises dispatch. Two blocking concerns remain before this can ship: the endpoint is reachable unauthenticated by default (carried over from the prior review round), and its `Cache-Control: no-store, private` header is silently overridable by operator `[response_headers]` because the response flows through `finalize_response`.

## Blocking

### 🔧 wrench

- **Public-by-default fingerprint reflection oracle**: `/_ts/debug/ja4` is not protected by basic-auth in the default config and reflects values JS cannot otherwise read directly (main.rs:217). Gate behind a debug flag, restrict to staging, or move under an admin path.
- **`Cache-Control` overridable by operator `[response_headers]`**: handler-set `no-store, private` is replaced by `finalize_response` if operators set `Cache-Control` in `[response_headers]` (main.rs:134 → main.rs:349-351). Handle the route alongside `/health` before `route_request` runs.

## Non-blocking

### ♻️ refactor

- **Unit test duplicates the route-level test**: identical assertions; route-level test is strictly stronger (main.rs:359-411 vs route_tests.rs:253-317).

### 🤔 thinking

- **Defensive `Vary` header**: `User-Agent`, Client-Hints, and TLS metadata vary the response — useful backstop if the override on `Cache-Control` is fixed (main.rs:133-136).

### 🌱 seedling

- **Track removal**: PR description and code call this "temporary" but neither has a removal trigger or follow-up issue. Add a `// TODO: remove after JA4 evaluation completes — see #645` and consider a scheduled cleanup PR.

### 📝 note

- **Bypasses platform abstraction for fields that have one**: `tls_protocol` and `tls_cipher` are available via `runtime_services.client_info()` (main.rs:115-116). JA4/H2 are Fastly-specific so direct calls are unavoidable for those; mixing in the abstraction for fields that have one would be more consistent.

### ⛏ nitpick

- **Use `header::CACHE_CONTROL` in test assertions** instead of `"cache-control"` (main.rs:376, route_tests.rs:282).
- **Extract magic strings to `const`s**: `"unavailable"`, `"not sent"`, `"none"` and label strings are duplicated across production and both tests (main.rs:113-130).

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS
- js tests: PASS
- wasm32-wasip1 release build: PASS

ChristianPavilonis (commented):
## Summary

I reviewed the current branch again and am only leaving findings that do not appear to duplicate the existing pending review comments. These are non-blocking cleanup/documentation accuracy items.

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -249,3 +249,133 @@ fn configured_missing_consent_store_only_breaks_consent_routes() {
...
+    .expect("should route ja4 debug request");
+
+    assert_eq!(
+        response.get_status(),
+        StatusCode::NOT_FOUND,
+        "should return 404 when debug.ja4_endpoint_enabled is false"
+    );
+}
+
+fn create_debug_enabled_settings() -> Settings {
```
ChristianPavilonis: **Debug-enabled test settings duplicate the base test TOML**

`create_debug_enabled_settings` copies nearly the entire TOML from `create_test_settings` and only adds `[debug] ja4_endpoint_enabled = true`. Future changes to required settings, handlers, or integration defaults will need to be kept in sync in two TOML blobs.

**Suggestion:** Reuse the existing helper and mutate the debug flag instead:

```rust
fn create_debug_enabled_settings() -> Settings {
    let mut settings = create_test_settings();
    settings.debug.ja4_endpoint_enabled = true;
    settings
}
```

trusted-server.toml
```diff
@@ -184,6 +184,23 @@ enabled = false
...
+# Enable the JA4/TLS fingerprint debug endpoint at GET /_ts/debug/ja4.
+# Returns a plain-text response with the following fields (Fastly-observed values):
+#   ja4         — JA4 TLS client fingerprint
+#   h2_fp       — HTTP/2 client fingerprint
+#   cipher      — TLS cipher suite (OpenSSL name)
+#   tls_version — TLS protocol version
+#   user-agent  — User-Agent request header
+#   ch-mobile   — Sec-CH-UA-Mobile client hint
+#   ch-platform — Sec-CH-UA-Platform client hint
+# All fields fall back to "unavailable" or "not sent" when Fastly does not provide them.
```
ChristianPavilonis: **JA4 debug config comment states all fallbacks inaccurately**

This comment says all fields fall back to `"unavailable"` or `"not sent"`, but `user-agent` falls back to `"none"` in `build_ja4_debug_response`.

**Suggestion:** Make the fallback documentation field-specific, for example:

```toml
# Fastly TLS/fingerprint fields fall back to "unavailable"; client hints fall back
# to "not sent"; user-agent falls back to "none" when absent.
```

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Fixed — the handler is now gated behind `debug.ja4_endpoint_enabled = false` (default off in `trusted-server.toml`), returning 404 when disabled so the path does not disclose its existence. As part of resolving the `Cache-Control` finding, the handler was also moved to `main()` before `finalize_response`, so it short-circuits before building the orchestrator or touching the routing stack.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Fixed via option 1 — `ja4_endpoint_enabled = false` by default, returns 404 when off. Additionally, as part of resolving the `Cache-Control` finding below, the entire handler was moved to `main()` before `finalize_response`, so it short-circuits before any routing infrastructure runs when disabled.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Fixed — handler moved to `main()` after `get_settings()` but before `build_orchestrator()`, same pattern as `/health`. Calls `send_to_client()` directly and returns, bypassing `finalize_response` entirely. Both the enabled and disabled (404) cases are handled there.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Took the inverse approach: because fixing the `Cache-Control` issue required moving the handler out of `route_request` and into `main()`, the route-level ja4 tests (`ja4_debug_route_returns_404_when_disabled` and `ja4_debug_route_returns_plain_text_fallback_response`) became the redundant ones — they no longer exercise a meaningful dispatch path. Both were removed. The unit test for `build_ja4_debug_response` is kept as the authoritative test.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Fixed in the unit test. The `route_tests.rs` assertion was inside `ja4_debug_route_returns_plain_text_fallback_response`, which was removed (see comment above).

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Fixed — extracted `FALLBACK_UNAVAILABLE`, `FALLBACK_NOT_SENT`, and `FALLBACK_NONE` as module-level consts in `main.rs`, co-located with `build_ja4_debug_response`.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Added — `Vary: User-Agent, Sec-CH-UA-Mobile, Sec-CH-UA-Platform` is now set on the response.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Kept one consistent source — reading all five values from `req` directly is clearer in this self-contained handler than mixing `req` and `runtime_services` within the same function. The handler now also runs before `runtime_services` is built (it short-circuits in `main()` before `build_runtime_services` is called), so mixing sources is not an option anyway.

crates/trusted-server-adapter-fastly/src/route_tests.rs
prk-Jr: Went further — `create_debug_enabled_settings` and both ja4 route-level tests were removed entirely. Moving the handler out of `route_request` and into `main()` (to fix the `Cache-Control` finding) made those route-level tests meaningless, so deletion was cleaner than the suggested mutation pattern.

trusted-server.toml
prk-Jr: Fixed — updated to your exact suggestion.

aram356 (approved):

PR #647, opened 2026-04-20
Add config store runtime config spec

### LINKED ISSUE #665: Create trusted server config store spec

ChristianPavilonis:
## Summary
- add a concrete spec for moving application config from build-time embedding to runtime config store loading
- define production vs development config behavior, including Fastly/Viceroy local-dev projection into a simulated config store
- scope the doc to config-store architecture only, with hashing included and signatures/tooling UX deferred

Closes #665

## Testing
- not run (docs-only change)

### DIFF
--- a/docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
+++ b/docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
@@ -1,0 +1,515 @@
+# Runtime Config Store Architecture for `trusted-server.toml`
+
+> **Status:** Proposal
+> **Scope:** Config store architecture only
+> **Date:** April 2026
+
+## Summary
+
+This document specifies the application configuration changes required to move
+Trusted Server from **build-time embedded configuration** to **runtime-loaded
+configuration**.
+
+Today, Trusted Server reads `trusted-server.toml` during the build, merges
+`TRUSTED_SERVER__*` environment variable overrides, writes a generated TOML
+file to `target/trusted-server-out.toml`, and embeds that generated file into
+the WASM binary with `include_bytes!`.
+
+This proposal replaces that model with the following target architecture:
+
+- **Production** loads application config from a platform config store at runtime
+- **Development** uses a local TOML-authored workflow rooted at the repo root by default
+- The platform config store uses a **fixed key**: `ts-config`
+- The stored payload is **canonical TOML** representing the application config
+- The config schema remains substantially the same as the existing
+  `trusted-server.toml` schema
+- Build-time embedding and build-time environment-variable merging for
+  application settings are removed
+
+The goal is to make `trusted-server.toml` remain the single application config
+source of truth while changing **how it is deployed and loaded**.
+
+This document is intentionally limited to config-store behavior and does not
+specify attestation signatures, discovery endpoints, or a full CLI design.
+
+## Scope
+
+### In scope
+
+- Production runtime loading of application config from a platform config store
+- Development loading of application config from a local TOML file
+- Bootstrap contract for locating the platform config store in production
+- Deterministic validation and canonicalization of application config
+- Config hashing based on canonical TOML bytes
+- Repository file ownership changes related to `trusted-server.toml`
+- Minimal tooling responsibilities required to deploy config to the store
+- Explicit removal of the current build-time merge-and-embed model
+
+### Out of scope
+
+- Config signatures, DSSE, or signature verification
+- Runtime attestation endpoints or statement formats
+- Full CLI command design or operator UX
+- Hot reload, file watching, or runtime config refresh
+- Runtime mutation of application config
+- Broad config schema redesign
+- Migration rollout sequencing or temporary dual-source compatibility
+- Platform-specific implementation details beyond the abstract store contract
+
+## Current state
+
+Trusted Server currently treats application config as a **build input**.
+
+The current flow is:
+
+1. Read root `trusted-server.toml`
+2. Merge `TRUSTED_SERVER__*` environment variable overrides at build time
+3. Serialize the merged config to `target/trusted-server-out.toml`
+4. Embed that generated file into the WASM binary with `include_bytes!`
+5. Parse the embedded TOML at runtime into `Settings`
+
+In the current codebase, this behavior is centered in:
+
+- `crates/trusted-server-core/build.rs`
+- `crates/trusted-server-core/src/settings_data.rs`
+
+This means the deployed binary contains both:
+
+- application code
+- publisher/operator configuration
+
+## Problems with the current state
+
+### Code and config are tightly coupled
+
+A config-only change produces a different binary artifact. That makes it harder
+to reason about code provenance separately from config changes.
+
+### Config changes require a rebuild
+
+Operators cannot update configuration independently from the WASM build.
+Changing runtime behavior requires regenerating and redeploying the binary.
+
+### Build-time environment merging weakens the source-of-truth model
+
+The current model combines a TOML file with build-time environment variables.
+That makes the effective config less obvious and weakens the idea of a single,
+authoritative application config document.
+
+### Repository ownership is wrong for operator config
+
+A tracked root `trusted-server.toml` suggests application config should live in
+source control as a committed repository artifact. In practice, application
+config is operator-owned and deployment-specific.
+
+### Runtime behavior depends on build tooling decisions
+
+Because config is preprocessed during the build, the runtime is not the
+authoritative point where config loading and validity are determined.
+
+## Goals
+
+- Keep **one application config document**: `trusted-server.toml`
+- Make **production** load application config from a platform config store
+- Make **development** load application config from a local TOML file
+- Remove build-time embedding of application settings into the WASM binary
+- Remove build-time `TRUSTED_SERVER__*` application-setting merges
+- Define a deterministic canonical TOML representation suitable for hashing
+- Preserve the existing application config schema as much as possible
+- Keep the architecture generic enough to work with a future cross-platform
+  config store abstraction
+
+## Non-goals
+
+- Designing config signing or signature verification
+- Designing a runtime attestation document
+- Supporting multiple application config sources simultaneously in production
+- Supporting hidden fallback from one source to another after source selection
+- Supporting runtime writeback of application config
+- Preserving comments or operator formatting in stored canonical payloads
+- Supporting unknown fields in application config
+- Introducing broad schema cleanup unrelated to config-store loading
+
+## Target architecture
+
+### High-level model
+
+`trusted-server.toml` remains the single application configuration document.
+What changes is how that document is sourced:
+
+- **Production:** platform config store
+- **Development:** local file at repo root by default
+
+The authoritative production payload is stored in the platform config store under
+key `ts-config`.
+
+The value stored under `ts-config` is canonical TOML representing the application
+config document.
+
+### Production behavior
+
+In production, Trusted Server:
+
+1. Obtains a platform config store reference from deployment/bootstrap wiring
+2. Reads the fixed key `ts-config`
+3. Parses the TOML payload using the existing application config schema
+4. Rejects unknown fields
+5. Validates semantic config rules
+6. Produces a valid immutable `Settings` snapshot for request handling
+7. Derives canonical TOML bytes and a config hash from that valid config
+
+Production application behavior must treat the platform config store payload as
+authoritative.
+
+### Development behavior
+
+In development, `trusted-server.toml` remains the default local authoring file
+at the repository root.
+
+A flag may be used to choose a different TOML file path.
+
+The local TOML file is the development authoring source of truth. Runtime
+consumption may happen either directly from that file or via a
+platform-specific projection step, depending on platform constraints.
+
+The development pipeline is:
+
+1. Load `trusted-server.toml` from the repository root by default, or the
+   explicitly selected TOML file
+2. Parse the TOML payload using the existing application config schema
+3. Reject unknown fields according to the rules defined in this document
+4. Validate semantic config rules
+5. Derive canonical TOML bytes and a config hash from that valid config
+6. Project the canonical TOML into the local development runtime in the
+   platform-appropriate way when needed
+7. Produce a valid immutable `Settings` snapshot for request handling
+
+On platforms such as Fastly/Viceroy, the preferred local-development approach is
+to populate the local simulated config store with the canonical payload under
+`ts-config` before request handling, rather than relying on direct host-file
+reads from within the WASM guest.
+
+When running in development, tooling or runtime logs should identify which local
+TOML file path was loaded.
+
+Development loading does **not** automatically rewrite the source file into
+canonical form.
+
+### Request-level semantics
+
+Each request must be handled against **one valid, internally consistent,
+immutable `Settings` snapshot**.
+
+This document intentionally does **not** require a specific fetch or caching
+strategy. An implementation may fetch fresh config for each request or reuse
+previously loaded state, as long as:
+
+- each request sees one coherent snapshot
+- invalid or partially loaded config is never used
+- correctness does not depend on cross-request in-memory persistence
+
+Because platform lifecycle behavior varies, the architecture must not assume that
+in-memory state survives across requests.
+
+## Bootstrap contract
+
+Production requires a minimal bootstrap mechanism to locate the platform config
+store. That bootstrap mechanism is **not** part of the application config schema.
+
+Bootstrap is deployment plumbing only.
+
+### Bootstrap responsibilities
+
+Production bootstrap must provide:
+
+- the platform config store reference needed to open/read the store
+
+Production bootstrap must **not**:
+
+- inject application settings values
+- override application settings values
+- provide an alternate key name for the application config payload
+- create a secondary source of truth for runtime application behavior
+
+### Fixed key name
+
+The application config payload key is fixed globally:
+
+- `ts-config`
+
+The runtime always reads `ts-config` once the production store reference has been
+resolved.
+
+The key name is not configurable.
+
+## Config loading and canonicalization pipeline
+
+This section defines the semantic pipeline for both production and development.
+
+### Step 1: Select the source by environment
+
+- **Production** uses the platform config store
+- **Development** uses a local TOML-authored workflow rooted at
+  `trusted-server.toml` in the repository root by default, with a flag to
+  choose a different file
+
+This document does not define a separate first-class runtime `config_source`
+mode. It defines source behavior in terms of production vs. development.
+
+### Step 2: Load the payload
+
+The selected source must yield a UTF-8 TOML payload.
+
+- Production loads the payload from store key `ts-config`
+- Development loads the payload from the selected local TOML file path and may
+  then project the canonicalized result into a local platform-simulated config
+  store before runtime consumption
+
+### Step 3: Parse strictly using the existing schema
+
+The payload is parsed as the existing application configuration schema used to
+produce `Settings`.
+
+This spec is schema-preserving by default. It does not redesign the
+`trusted-server.toml` structure beyond what is necessary to support the new
+loading model.
+
+### Step 4: Reject unknown fields
+
+Unknown fields are rejected.
+
+The system must not silently preserve, ignore, or drop unsupported keys.
+
+This applies to strongly typed configuration sections in the application config
+schema. The `integrations` section continues to follow the existing integration
+configuration model, where integration IDs are discovered dynamically and each
+integration's typed validation rules govern the contents of its own config.
+
+### Step 5: Validate semantic constraints
+
+After parsing, the config must satisfy existing semantic validation rules.
+
+Examples include:
+
+- required fields must be present
+- invalid field combinations must fail
+- invalid regexes, store identifiers, or route coverage rules must fail
+
+### Step 6: Define the canonical TOML representation
+
+Valid config has a deterministic canonical TOML representation.
+
+Canonicalization is part of the config pipeline semantics, not merely a tooling
+implementation detail.
+
+Canonicalization is defined as a dedicated transformation over valid config. It
+must not rely on whatever output happens to fall out of naive derived
+serialization alone. Implementations may need explicit canonicalization logic to
+ensure the required output properties.
+
+Canonicalization is defined as:
+
+- parse valid config through the typed config model
+- serialize it in a deterministic TOML form
+- include explicitly declared settings only
+- do **not** expand the config into a full dump of all effective defaulted runtime values
+- define stable ordering for map-like structures so identical semantic config
+  produces identical canonical bytes
+
+As a consequence:
+
+- comments are not preserved in canonical form
+- original formatting is not preserved in canonical form
+- canonical stored payloads are intended to be tight and deterministic
+- additional implementation work may be required beyond current derived
+  `Serialize` behavior to satisfy these guarantees
+
+### Step 7: Compute the config hash
+
+The config hash is computed over the canonical TOML bytes.
+
+This provides a stable hashable representation of application config suitable
+for observability and future attestation work.
+
+### Step 8: Produce the runtime snapshot
+
+The runtime uses the validated config to produce the `Settings` snapshot used by
+request handling.
+
+Implementations may materialize canonical bytes eagerly or lazily, but the
+canonical form is part of the defined semantics.
+
+## Failure behavior
+
+The selected config source must produce one valid `Settings` snapshot.
+
+If loading, parsing, or validation fails, the runtime must **fail closed**.
+
+That includes failures such as:
+
+- missing store/bootstrap reference in production
+- inability to read the selected store or file
+- missing `ts-config` key in production
+- invalid UTF-8 payloads
+- malformed TOML
+- unknown fields
+- missing required fields
+- semantic validation failures
+
+### No fallback after source selection
+
+Once the source has been selected by environment, the runtime must not fall back
+to another source.
+
+Examples of disallowed behavior:
+
+- production store mode falling back to a local file
+- development file mode falling back to embedded config
+- falling back to a previously cached last-known-good config
+- loading partial config and continuing with defaults beyond normal schema behavior
+
+### Availability behavior
+
+A config failure means the service is not healthy for serving application
+traffic.
+
+This spec does not define a special "healthy but unusable" mode for config
+failure.
+
+## Repository and file ownership changes
+
+The repository layout should change to reflect the new ownership model.
+
+### Required changes
+
+- Remove tracked `trusted-server.toml` from source control
+- Add `trusted-server.toml` to `.gitignore`
+- Add or retain `trusted-server.example.toml` as a tracked template file
+
+### File roles
+
+#### `trusted-server.toml`
+
+- operator-owned local/deployment artifact
+- default local authoring file for development
+- not tracked in git
+
+#### `trusted-server.example.toml`
+
+- tracked template file
+- kept in sync with currently supported configuration features
+- intended to help operators create a real `trusted-server.toml`
+
+## Minimal tooling contract
+
+This document does not define a full CLI specification.
+
+It does define the minimum tooling responsibilities required by the target
+architecture.
+
+Tooling responsible for publishing production config must be able to:
+
+1. Load a local TOML file
+2. Parse it using the application config schema
+3. Reject unknown fields
+4. Validate semantic config rules
+5. Produce canonical TOML
+6. Compute a hash over canonical TOML bytes
+7. Write the canonical TOML payload to the platform config store under `ts-config`
+
+For local platform simulators such as `fastly compute serve`, tooling may also
+materialize that canonical payload into the simulator's local config-store input
+before starting the runtime.
+
+Tooling may support additional commands later, such as:
+
+- pull
+- diff
+- inspect
+- dry-run deployment
+
+Those capabilities are explicitly out of scope for this document.
+
+## Hashing
+
+Config hashing is part of this architecture because it depends on deterministic
+canonicalization.
+
+### Hash source
+
+The config hash is computed over the canonical TOML bytes.
+
+Because the hash is derived from canonical bytes, canonicalization must produce
+stable field and map ordering for semantically identical config.
+
+### Purpose
+
+The config hash exists to support:
+
+- observability
+- config comparison
+- deterministic deployment artifacts
+- future attestation and provenance work
+
+### Out of scope
+
+This document does not define:
+
+- signature formats
+- signed envelopes
+- signature verification behavior
+- runtime signature enforcement
+
+## Explicit removals from the current design
+
+This proposal explicitly removes the current application-config build pipeline.
+
+### Removed behaviors
+
+- build-time embedding of application config into the WASM binary
+- build-time generation of `target/trusted-server-out.toml` as the runtime app-config source
+- build-time merging of `TRUSTED_SERVER__*` application-setting overrides
+- production dependence on a repository-tracked `trusted-server.toml`
+- runtime mutation of application config
+
+### Resulting source-of-truth model
+
+After this change:
+
+- `trusted-server.toml` remains the canonical application config document
+- in production, the authoritative deployed copy is the platform-store payload under `ts-config`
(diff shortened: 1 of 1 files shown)


docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+resolved.
+
+The key name is not configurable.
+
+## Config loading and canonicalization pipeline
+
+This section defines the semantic pipeline for both production and development.
+
+### Step 1: Select the source by environment
+
```
aram356, line 250: ❓ **question** — Source-selection signal is unspecified. Step 1 says "Select the source by environment" but never describes the runtime mechanism. Cargo feature? Env var read at module init? Fastly service-ID check? Bootstrap config flag? This is the central pipeline decision and is undefined. Either pin the mechanism here or defer explicitly (e.g., "to be defined in the tooling spec, but must be a compile-time signal").

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+- canonical stored payloads are intended to be tight and deterministic
+- additional implementation work may be required beyond current derived
+  `Serialize` behavior to satisfy these guarantees
+
+### Step 7: Compute the config hash
+
+The config hash is computed over the canonical TOML bytes.
+
+This provides a stable hashable representation of application config suitable
+for observability and future attestation work.
```
aram356, line 332: 🔧 **wrench** — Hash algorithm not specified despite explicit "future attestation" goal. The doc commits to a config hash over canonical TOML bytes but never names SHA-256 vs BLAKE3 vs SHA-3, nor the encoding (hex / base64 / raw). Choosing here locks in the attestation contract in `2026-01-15-attestation-design.md`. Pin the algorithm + encoding (suggest SHA-256, hex) or explicitly defer to the attestation spec with a forward reference.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+
+Canonicalization is defined as:
+
+- parse valid config through the typed config model
+- serialize it in a deterministic TOML form
+- include explicitly declared settings only
+- do **not** expand the config into a full dump of all effective defaulted runtime values
+- define stable ordering for map-like structures so identical semantic config
+  produces identical canonical bytes
+
```
aram356, line 318: 🔧 **wrench** — Canonical TOML form is underspecified for hash stability. "Define stable ordering for map-like structures" leaves the following undefined and stability-critical:

- key ordering rule (lexicographic vs schema-declared vs insertion)
- string flavor (basic `"..."` vs literal `'...'`)
- numeric formatting (`1` vs `1.0`)
- table vs inline-table normalization
- array-of-tables vs inline-array
- Unicode normalization of string values

Two correct implementations (or two `toml-rs` versions) can produce different bytes from the same `Settings` and break hash stability. Either enumerate the rules explicitly or pin a canonicalizer crate + version.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+serialization alone. Implementations may need explicit canonicalization logic to
+ensure the required output properties.
+
+Canonicalization is defined as:
+
+- parse valid config through the typed config model
+- serialize it in a deterministic TOML form
+- include explicitly declared settings only
+- do **not** expand the config into a full dump of all effective defaulted runtime values
+- define stable ordering for map-like structures so identical semantic config
```
aram356, line 316: 🔧 **wrench** — "Explicitly declared settings only" contradicts the existing Serialize round-trip. Today `crates/trusted-server-core/build.rs:44-45` does `toml::to_string_pretty(&settings)`, which includes defaults. The new contract requires a "provided vs defaulted" distinction that `#[serde(default)]` erases. Implementers will need either custom `Serialize` impls per type or a default-elision pass that compares against a baseline `Settings`. Pick an approach or call this out as a known cost.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+store. That bootstrap mechanism is **not** part of the application config schema.
+
+Bootstrap is deployment plumbing only.
+
+### Bootstrap responsibilities
+
+Production bootstrap must provide:
+
+- the platform config store reference needed to open/read the store
+
```
aram356, line 226: 🔧 **wrench** — Bootstrap "store reference" is platform-undefined. Fastly Compute references config stores by **name** at the WASM guest level — the names live in `fastly.toml`, not in environment. The spec should at least state:

- On Fastly, the bootstrap form is a fixed store-name constant compiled into the adapter.
- `fastly.toml` must declare a `ts-config` `[local_server.config_stores]` entry for dev.

As written, the bootstrap contract is not actionable for the only currently supported platform.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+This spec is schema-preserving by default. It does not redesign the
+`trusted-server.toml` structure beyond what is necessary to support the new
+loading model.
+
+### Step 4: Reject unknown fields
+
+Unknown fields are rejected.
+
+The system must not silently preserve, ignore, or drop unsupported keys.
```
aram356, line 281: 🔧 **wrench** — `deny_unknown_fields` is a behavioral break with no migration path. Today's `Settings` (in `crates/trusted-server-core/src/settings.rs`) does **not** use `#[serde(deny_unknown_fields)]`. Operators with stale or typo'd keys are silently tolerated. The new contract fail-closes them at startup. Spec strictness is correct, but should explicitly state:

- This is a breaking change for any pre-existing operator TOML.
- The `integrations` carve-out covers per-integration payload internals only; the dynamic top-level `[integrations.<id>]` keyspace itself is not denied.

Without this, the migration risk is invisible to readers.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+strategy. An implementation may fetch fresh config for each request or reuse
+previously loaded state, as long as:
+
+- each request sees one coherent snapshot
+- invalid or partially loaded config is never used
+- correctness does not depend on cross-request in-memory persistence
+
+Because platform lifecycle behavior varies, the architecture must not assume that
+in-memory state survives across requests.
+
```
aram356, line 213: 🔧 **wrench** — Per-request load cost forecloses straightforward optimization. "Correctness does not depend on cross-request in-memory persistence" plus Fastly's per-request guest model means a worst-case implementation pays TOML parse + per-handler regex compile (via `prepare_runtime()`) every request. The spec allows in-memory caching as a perf optimization but forbids it for correctness. This is a hard implementation constraint that should be acknowledged so readers don't underestimate cost — and so reviewers of the implementation know the spec, not the implementer, set this ceiling.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+4. Validate semantic config rules
+5. Derive canonical TOML bytes and a config hash from that valid config
+6. Project the canonical TOML into the local development runtime in the
+   platform-appropriate way when needed
+7. Produce a valid immutable `Settings` snapshot for request handling
+
+On platforms such as Fastly/Viceroy, the preferred local-development approach is
+to populate the local simulated config store with the canonical payload under
+`ts-config` before request handling, rather than relying on direct host-file
+reads from within the WASM guest.
```
aram356, line 190: 🤔 **thinking** — "Production vs development" runtime distinction is largely cosmetic. After this paragraph commits to populating the local simulated config store with the canonical payload under `ts-config` before request handling, both runtime paths read the same key from the same kind of store. The actual difference is in tooling that *populates* the store, not in runtime logic. Collapsing the doc to one runtime path plus a separate "pre-flight" tooling step would simplify the surface area significantly.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+
+## Implementation notes
+
+These notes are informative, not additional scope.
+
+- The concrete platform config-store API may still be evolving while this is
+  implemented
+- Existing generic config-store abstractions in the codebase may be reused as
+  they mature
+- Runtime caching strategy is intentionally unspecified by this document
```
aram356, line 491: 🤔 **thinking** — Fastly Config Store value-size limit is not addressed. Fastly Config Stores cap each value (historically ~8 KB; recent expansions but still capped). Current canonicalized TOML is ~3.4 KB; source is ~6.8 KB. Future growth (more integrations, more handler regexes) could exceed it. Worth a one-line note in "Implementation notes" so this surfaces during design rather than at deploy time.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+### Fixed key name
+
+The application config payload key is fixed globally:
+
+- `ts-config`
+
+The runtime always reads `ts-config` once the production store reference has been
+resolved.
+
+The key name is not configurable.
```
aram356, line 243: 🤔 **thinking** — Multi-tenant constraint of fixed `ts-config` key. Hard-coding the key globally precludes per-tenant config via key multiplexing within a shared binary deployment. Probably intentional (one binary = one tenant), but call it out explicitly so future cross-platform abstractions don't assume multi-tenant key namespacing.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+  adjustments are necessary to support strict parsing or canonicalization
+
+## Future work enabled by this design
+
+This design is intended to enable, but not itself specify:
+
+- config attestation based on canonical payload hashes
+- runtime reporting of config hash metadata
+- richer deployment tooling around validation, diffing, and inspection
+- broader cross-platform config-store support behind a generic API
```
aram356, line 502: ♻️ **refactor** — Cross-link to existing attestation spec. The spec mentions "future attestation" four times. `docs/superpowers/specs/2026-01-15-attestation-design.md` already covers attestation. Add a `Related work` section linking it and noting the canonical hash here is the input to that spec's signing flow.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+- in development, the authoritative copy is the selected local TOML file
+
+## Implementation notes
+
+These notes are informative, not additional scope.
+
+- The concrete platform config-store API may still be evolving while this is
+  implemented
+- Existing generic config-store abstractions in the codebase may be reused as
+  they mature
```
aram356, line 490: ♻️ **refactor** — Name the existing platform abstraction. `crates/trusted-server-core/src/platform/traits.rs:12` already defines `PlatformConfigStore`. The spec mentions "existing generic config-store abstractions in the codebase may be reused as they mature" but never names the trait. Naming it makes the doc actionable.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+2. Parse it using the application config schema
+3. Reject unknown fields
+4. Validate semantic config rules
+5. Produce canonical TOML
+6. Compute a hash over canonical TOML bytes
+7. Write the canonical TOML payload to the platform config store under `ts-config`
+
+For local platform simulators such as `fastly compute serve`, tooling may also
+materialize that canonical payload into the simulator's local config-store input
+before starting the runtime.
```
aram356, line 422: ♻️ **refactor** — Specify how the dev `ts-config` payload reaches the simulator. On Viceroy, config-store contents come from `[local_server.config_stores.<name>.contents]` in `fastly.toml` (see `fastly.toml:42-49`). Either:

- dev tooling regenerates that section of `fastly.toml` from the canonical TOML on each `fastly compute serve`, or
- there is an alternate loading mechanism (external file pointer, etc.).

The spec is silent on the mechanism even though it commits to populating the local simulated config store with the canonical payload under `ts-config` before request handling. A one-paragraph sketch of the mechanism would close the loop.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+- loading partial config and continuing with defaults beyond normal schema behavior
+
+### Availability behavior
+
+A config failure means the service is not healthy for serving application
+traffic.
+
+This spec does not define a special "healthy but unusable" mode for config
+failure.
+
```
aram356, line 378: 🌱 **seedling** — Note `/health` carve-out from fail-closed. `crates/trusted-server-adapter-fastly/src/main.rs:55-61` short-circuits `/health` before settings load, which is a deliberate carve-out from the fail-closed model defined here. A one-liner acknowledging this would prevent a future reader from "fixing" the early-return as a bug.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+- operator-owned local/deployment artifact
+- default local authoring file for development
+- not tracked in git
+
+#### `trusted-server.example.toml`
+
+- tracked template file
+- kept in sync with currently supported configuration features
+- intended to help operators create a real `trusted-server.toml`
+
```
aram356, line 402: 📌 **out of scope** — Test fixtures + CI depend on tracked `trusted-server.toml`. Removing it touches `crate_test_settings_str()` consumers and 10+ tests in `crates/trusted-server-core/src/settings.rs`, plus CI workflows that rely on the file existing at the repo root. Worth a follow-up issue tracking the test/CI migration before this becomes implementable.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
+# Runtime Config Store Architecture for `trusted-server.toml`
+
+> **Status:** Proposal
+> **Scope:** Config store architecture only
+> **Date:** April 2026
```
aram356, line 5: ⛏ **nitpick** — Peer specs (e.g., `2026-01-15-attestation-design.md`) include an `Author:` line. Adding one keeps consistency.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+### High-level model
+
+`trusted-server.toml` remains the single application configuration document.
+What changes is how that document is sourced:
+
+- **Production:** platform config store
+- **Development:** local file at repo root by default
+
+The authoritative production payload is stored in the platform config store under
+key `ts-config`.
```
aram356, line 144: ⛏ **nitpick** — Inconsistent terminology. "platform config store" / "platform config-store" / "platform-store" / "config-store payload" / "platform store payload" appear interchangeably. Pick one form for the doc.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+must not rely on whatever output happens to fall out of naive derived
+serialization alone. Implementations may need explicit canonicalization logic to
+ensure the required output properties.
+
+Canonicalization is defined as:
+
+- parse valid config through the typed config model
+- serialize it in a deterministic TOML form
+- include explicitly declared settings only
+- do **not** expand the config into a full dump of all effective defaulted runtime values
```
aram356, line 315: ⛏ **nitpick** — `do **not** expand the config into a full dump of all effective defaulted runtime values` reads oddly mid-sentence inside this numbered list. Consider promoting it to its own bullet.

aram356 (changes requested):
## Summary

Spec for moving `trusted-server.toml` from build-time embedded config to runtime config-store loading. Scope and fail-closed posture are good, but several load-bearing details are unspecified or contradict the existing pipeline. Requesting changes on the items below before merge.

## Blocking

### 🔧 wrench

- **Hash algorithm not specified despite explicit "future attestation" goal** (lines 328–332): The doc commits to a config hash over canonical TOML bytes but never names SHA-256 vs BLAKE3 vs SHA-3. This locks in (or breaks) the attestation contract in `2026-01-15-attestation-design.md`. Pin the algorithm + encoding (suggest SHA-256, hex) or explicitly defer to the attestation spec with a forward reference.
- **Canonical TOML form is underspecified for hash stability** (lines 310–318): "Stable ordering" leaves these undefined: key ordering rule (lexicographic vs schema-declared vs insertion), string flavor (basic `"..."` vs literal `'...'`), numeric formatting (`1` vs `1.0`), table vs inline-table, array-of-tables vs inline-array, Unicode normalization. Two `toml-rs` versions can disagree. Either enumerate the rules or pin a canonicalizer crate + version.
- **"Explicitly declared settings only" contradicts the existing Serialize round-trip** (lines 313–316): Today `crates/trusted-server-core/build.rs:44-45` does `toml::to_string_pretty(&settings)`, including defaults. The new contract requires a "provided vs defaulted" distinction that `#[serde(default)]` erases. Implementers will need custom `Serialize` or a default-elision pass over a baseline `Settings`. Pick an approach or call this out as a known cost.
- **Bootstrap "store reference" is platform-undefined** (lines 222–226): Fastly Compute references config stores by **name** at the WASM guest. The bootstrap section should at least state: on Fastly the store-name is a fixed constant compiled into the adapter, and `fastly.toml` must declare a `ts-config` `[local_server.config_stores]` for dev. As written the "bootstrap contract" is not actionable for the only currently supported platform.
- **`deny_unknown_fields` is a behavioral break with no migration path** (lines 278–286): `Settings` does not currently use `#[serde(deny_unknown_fields)]`. The new contract fail-closes any operator TOML with stale or typo'd keys. Spec should explicitly state (a) this is a breaking change vs. existing operator configs and (b) the `integrations` carve-out applies only to per-integration payload internals — the dynamic top-level `[integrations.<id>]` keyspace itself is not denied.
- **Per-request load cost forecloses straightforward optimization** (lines 198–213): "Correctness does not depend on cross-request in-memory persistence" plus Fastly's per-request guest model means a worst-case implementation does TOML parse + per-handler regex compile (`prepare_runtime()`) every request. The spec allows in-memory caching as a perf optimization but forbids it for correctness. Worth flagging as a hard implementation constraint so readers don't underestimate cost.

### ❓ question

- **Source-selection signal is unspecified** (lines 249–258): Step 1 says "Select the source by environment" but never describes the runtime mechanism — cargo feature, env var, Fastly service-ID check, bootstrap flag? This is the central pipeline decision and is undefined. Either pin the mechanism or explicitly defer with a marker.
- **Secrets migration is missing**: `trusted-server.toml` carries `edge_cookie.secret_key` and `publisher.proxy_secret` (`crates/trusted-server-core/src/settings_data.rs:43-57`), overridden today via `TRUSTED_SERVER__EDGE_COOKIE__SECRET_KEY` and `TRUSTED_SERVER__PUBLISHER__PROXY_SECRET` at build time. Removing build-time env merging implies either (a) plaintext secrets in the (unencrypted) Fastly Config Store, or (b) splitting secrets into `PlatformSecretStore` and referencing them by name from `Settings`. The spec is silent — which is the intended model?

## Non-blocking

### 🤔 thinking

- **"Production vs development" runtime distinction is largely cosmetic** (lines 187–190): Both paths read `ts-config` from a (real or simulated) config store. The actual difference is *who populates the store*, not *how the runtime reads it*. Collapsing to one runtime path plus a tooling pre-flight step would simplify the doc.
- **Fastly Config Store value-size limit is not addressed**: Fastly caps each value (~8 KB historically). Current canonical TOML is ~3.4 KB; source is ~6.8 KB. Future growth could exceed the limit. One line in "Implementation notes" surfaces this at design time.
- **Multi-tenant constraint of fixed `ts-config` key** (lines 235–243): Hard-coding the key globally precludes per-tenant config via key multiplexing in a shared binary. Likely intentional — call it out so future cross-platform abstractions don't assume key namespacing.

### ♻️ refactor

- **Cross-link to existing attestation spec**: Spec mentions "future attestation" four times. `docs/superpowers/specs/2026-01-15-attestation-design.md` already exists. Add a "Related work" section linking it.
- **Name the existing platform abstraction**: `crates/trusted-server-core/src/platform/traits.rs:12` defines `PlatformConfigStore`. Spec says "existing generic config-store abstractions" but never names it. Naming makes implementation actionable.
- **Specify how the dev `ts-config` payload reaches Viceroy** (lines 418–422): Viceroy reads config-store contents from `[local_server.config_stores.<name>.contents]` in `fastly.toml` (`fastly.toml:42-49`). Either dev tooling regenerates that section per `fastly compute serve`, or an alternate loading mechanism. Worth one paragraph.

### 🌱 seedling

- **Note `/health` carve-out from fail-closed**: `crates/trusted-server-adapter-fastly/src/main.rs:55-61` short-circuits `/health` before settings load. Worth a one-liner so a future reader doesn't "fix" this as inconsistent with fail-closed semantics.

### 📌 out of scope

- **Test fixtures + CI depend on tracked `trusted-server.toml`**: Removing it touches `crate_test_settings_str()` consumers and 10+ tests in `crates/trusted-server-core/src/settings.rs`. Open a follow-up issue tracking the test/CI migration before implementation lands.

### ⛏ nitpick

- **Add Author line**: Peers use `> **Author:**`. This spec omits it.
- **Inconsistent terminology**: "platform config store" / "platform config-store" / "platform-store" used interchangeably. Pick one.
- **Bullet phrasing at line 314**: `do **not** expand…` mid-sentence reads oddly inside the numbered list.

## CI Status

- cargo fmt: PASS
- cargo test: PASS
- vitest: PASS
- format-typescript: PASS
- format-docs: PASS
- Analyze (rust / actions / js-ts): PASS
- integration tests: PASS
- CodeQL: PASS

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+- development file mode falling back to embedded config
+- falling back to a previously cached last-known-good config
+- loading partial config and continuing with defaults beyond normal schema behavior
+
+### Availability behavior
+
+A config failure means the service is not healthy for serving application
+traffic.
+
+This spec does not define a special "healthy but unusable" mode for config
```
aram356, line 376: 🔧 **wrench** — Availability section directly contradicts the existing `/health` carve-out.

Spec text says:

> A config failure means the service is not healthy for serving application traffic. This spec does not define a special "healthy but unusable" mode for config failure.

But `crates/trusted-server-adapter-fastly/src/main.rs:53-60` already short-circuits `/health` and returns `200 ok` *before* `get_settings()` runs. That is a deliberate "healthy but unusable" mode. An implementer reading this section in isolation would (correctly per the spec) delete the carve-out, breaking external readiness probes.

**Suggested rewording:**

```
A config failure means the service is not healthy for serving *application* traffic.

Liveness probes (e.g. `/health`) are explicitly carved out from this rule and remain available regardless of config-load state. This is a deliberate liveness-vs-readiness split, not a hidden fallback path.
```

This was a 🌱 in the prior pass — re-reading, the spec text is an active contradiction, not a footnote. Elevating.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
aram356, line 318: 🔧 **wrench** — Map-typed sections in `Settings` use `HashMap`, whose serde serialization is order-unstable. The current canonicalization wrench is abstract; this is the concrete code path it has to address.

- `crates/trusted-server-core/src/settings.rs:79` — `IntegrationSettings { entries: HashMap<String, JsonValue> }`. Every integration block flows through this.
- `crates/trusted-server-core/src/settings.rs:415` — `response_headers: HashMap<String, String>`. Used by every handler-config entry.

`toml::to_string_pretty(&settings)` (still the build.rs round-trip on `main`) emits these in iteration order, which is non-deterministic across runs. Two consecutive canonicalizations of the same `Settings` produce different bytes today.

**Suggested addition** under Step 6:

```
- map-like fields (`HashMap` in current `Settings`) MUST be serialized in lexicographic key order. Non-deterministic iteration of the underlying collection is a canonicalization bug, not an implementation choice.
```

This turns the abstract "stable ordering for map-like structures" bullet into something an implementer can grep for and verify.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+### Step 4: Reject unknown fields
+
+Unknown fields are rejected.
+
+The system must not silently preserve, ignore, or drop unsupported keys.
+
+This applies to strongly typed configuration sections in the application config
+schema. The `integrations` section continues to follow the existing integration
+configuration model, where integration IDs are discovered dynamically and each
+integration's typed validation rules govern the contents of its own config.
```
aram356, line 286: 🤔 **thinking** — The integration carve-out makes the fail-closed-on-typo guarantee weaker than the spec implies.

This paragraph exempts `[integrations.*]` from `deny_unknown_fields`, deferring strictness to each integration's typed config. Repo-wide grep on `main`:

```
grep -rn 'deny_unknown_fields' crates/trusted-server-core/src/
# (no matches)
```

No integration carries `deny_unknown_fields` today. So a typo in `[prebid].tiemout_ms` or `[google_tag_manager].containner_id` still passes silently — the very class of bug the rest of this spec is trying to close.

**Two consistent options:**

1. Require integration `Deserialize`s to also carry `deny_unknown_fields`. Hardens the contract to the integration boundary.
2. Add a Non-goal stating that the strict-parse contract stops at the integration boundary, and document the implication.

As written, the spec reads stronger than what the schema can deliver.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+- serialize it in a deterministic TOML form
+- include explicitly declared settings only
+- do **not** expand the config into a full dump of all effective defaulted runtime values
+- define stable ordering for map-like structures so identical semantic config
+  produces identical canonical bytes
+
+As a consequence:
+
+- comments are not preserved in canonical form
+- original formatting is not preserved in canonical form
```
aram356, line 322: 🤔 **thinking** — Canonical-bytes line-ending and trailing-newline conventions are unspecified. Both materially affect the hash.

- An operator authoring `trusted-server.toml` on Windows produces CRLF; the runtime canonicalizer reading that file and re-serializing via `toml::to_string_pretty` produces LF. Round-trip changes the hash.
- `toml::to_string_pretty` does not append a trailing newline. Tooling pipelines often do (`echo "$canon" > file`). The CLI hash and runtime hash drift on a single byte.

**Suggested addition** to Step 6:

```
- canonical bytes use LF line endings only (any CR/CRLF in source is normalized at canonicalization time)
- canonical bytes terminate with exactly one trailing LF
```

The kind of ambiguity that surfaces three months in as "the CLI computes a different hash than the runtime."

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+
+## Repository and file ownership changes
+
+The repository layout should change to reflect the new ownership model.
+
+### Required changes
+
+- Remove tracked `trusted-server.toml` from source control
+- Add `trusted-server.toml` to `.gitignore`
+- Add or retain `trusted-server.example.toml` as a tracked template file
```
aram356, line 387: 📝 **note** — `trusted-server.example.toml` does not exist in the repo today. `git ls-files | grep example` returns nothing; only `trusted-server.toml` is tracked. "Add or retain" should just read "Add" (or "Create").

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+
+### Step 1: Select the source by environment
+
+- **Production** uses the platform config store
+- **Development** uses a local TOML-authored workflow rooted at
+  `trusted-server.toml` in the repository root by default, with a flag to
+  choose a different file
+
+This document does not define a separate first-class runtime `config_source`
+mode. It defines source behavior in terms of production vs. development.
```
aram356, line 257: 🌱 **seedling** — Test path is implicit in the production-vs-development source-selection model.

The dominant call site for `Settings` numerically is the test fixture `crate_test_settings_str()` → `Settings::from_toml(...)` (30+ tests across `settings.rs`, `auth.rs`, `auction/orchestrator.rs`, integrations). Tests flow through neither production nor development as defined here.

A one-line acknowledgement under Step 1 — something like "unit tests construct `Settings` directly and bypass source selection by design" — would prevent a future contributor from "fixing" the test path to satisfy the source-selection contract.

docs/superpowers/specs/2026-04-20-config-store-runtime-config-design.md
```diff
@@ -0,0 +1,515 @@
...
+
+- missing store/bootstrap reference in production
+- inability to read the selected store or file
+- missing `ts-config` key in production
+- invalid UTF-8 payloads
+- malformed TOML
+- unknown fields
+- missing required fields
+- semantic validation failures
+
```
aram356, line 358: 📝 **note** — Failure-behavior list reads as exhaustive but isn't. Missing cases:

- store read timeouts (Fastly's config-store API can stall)
- partial / truncated reads
- transient store unavailability (retryable vs not)

Either prefix the list with "examples include" (current text says "includes") or add these. Failure-mode docs are load-bearing for fail-closed contracts; the implementer needs to know whether retry-on-timeout is a fallback (disallowed by line 361-362) or just normal I/O.

aram356 (changes requested):
## Summary

Second-pass review on the same SHA (`1ff07c95`) — the PR has had no new commits since the prior CHANGES_REQUESTED review on Apr 30, so all 18 prior findings still stand verbatim. This review adds findings the prior pass missed and elevates one finding that re-reading exposes as a contradiction rather than a footnote.

The load-bearing problem: the explicit downstream goal is **attestation**, which requires hash determinism. The spec demands hash stability but leaves the four ingredients open — algorithm, map-key ordering, EOL convention, "explicitly declared settings only" mechanism. Two correct independent implementations of this spec will produce different hashes today.

## Blocking

### 🔧 wrench

- **Availability section contradicts existing `/health` carve-out**: `crates/trusted-server-adapter-fastly/src/main.rs:53-60` already returns `200 ok` before settings load — that is a "healthy but unusable" mode the spec text claims it does not define. (line 376)
- **`HashMap` map-typed sections produce non-deterministic canonical bytes**: `IntegrationSettings.entries` (`settings.rs:79`) and `response_headers` (`settings.rs:415`) — concrete code paths the abstract canonical-ordering bullet has to address. (line 318)

## Non-blocking

### 🤔 thinking

- **Integration carve-out weakens fail-closed-on-typo**: no integration config carries `deny_unknown_fields` today; typos under `[prebid]`, `[google_tag_manager]`, etc. still pass silently. Either harden integrations or move the limitation into Non-goals. (line 286)
- **EOL and trailing-newline unspecified for canonical bytes**: CRLF/LF normalization and trailing-newline convention both affect the hash; both undefined. (line 322)

### 📝 note

- **`trusted-server.example.toml` does not exist today**; "Add or retain" should just say "Add". (line 387)
- **Failure-behavior list reads exhaustive but isn't**: missing store timeouts, partial reads, transient unavailability. (line 358)

### 🌱 seedling

- **Test path is implicit in source selection**: `crate_test_settings_str()` flows through neither production nor development as defined. One-line acknowledgement under Step 1. (line 257)

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS
- js tests: PASS
- format-docs: PASS

PR #648, opened 2026-04-20
Add test coverage for error mapping and tsjs helpers

### LINKED ISSUE #448: Add tests for TrustedServerError HTTP status mapping
## Description

`TrustedServerError::status_code()` in `crates/common/src/error.rs` has no direct tests asserting the HTTP status mapping for each error variant. A regression here would silently change client behavior and monitoring semantics.

## Done when

- [ ] Add deterministic tests covering each public error variant to status code mapping.
- [ ] Assert representative server-side and client-side errors map to the expected HTTP responses.
- [ ] Keep the scope limited to mapping coverage, not unrelated error refactors.

## Affected area

## Additional context

Source: `test_gap.md`, section `HTTP Status Code Mapping Not Validated`.
Parent issue: #396
Related: #437 covers leaked error context; this issue is specifically for missing mapping tests.

### LINKED ISSUE #450: Add unit tests for tsjs script source and tag generation
## Description

`crates/common/src/tsjs.rs` builds the injected TS/JS script URL and `<script>` tag used in publisher HTML responses, but the formatting behavior is untested.

## Done when

- [ ] Add unit tests for `tsjs_script_src` URL structure, hash placement, and module query string formatting.
- [ ] Add unit tests for `tsjs_script_tag` output attributes and valid single-tag structure.
- [ ] Cover realistic `IntegrationRegistry` input so enabled modules appear in the generated output.

## Affected area

## Additional context

Source: `test_gap.md`, section `Script Tag Generation`.
Parent issue: #396

### LINKED ISSUE #454: Add invalid cookie header tests for get_cookie_value
## Description

`get_cookie_value` in `crates/common/src/cookies.rs` does not explicitly test the invalid UTF-8 branch for the incoming `Cookie` header.

## Done when

- [ ] Add a unit test for invalid UTF-8 cookie header bytes returning `None`.
- [ ] Keep existing valid cookie parsing behavior unchanged.
- [ ] Add any small guardrail assertions needed for missing/empty cookie inputs if they clarify behavior.

## Affected area

## Additional context

Source: `test_gap.md`, section `Cookie Parsing Edge Cases`.
Parent issue: #396

### LINKED ISSUE #456: Consolidate duplicate callback serde rename tests
## Description

`crates/common/src/models.rs` currently uses overlapping callback deserialization tests to verify the same serde rename behavior. This is redundant and makes the test surface noisier than necessary.

## Done when

- [ ] Consolidate the duplicate callback rename tests into one focused test.
- [ ] Preserve the existing `type` -> `callback_type` serde assertion.
- [ ] Keep the resulting test intent clearer than the current duplicated coverage.

## Affected area

## Additional context

Source: `test_gap.md`, section `Duplicate Callback Deserialization Tests`.
Parent issue: #396

ChristianPavilonis:
## Superseded

This PR is superseded and should not be merged as the canonical solution for the #396 backlog cleanup.

- #448 and #450 are covered by #688.
- #454 and #456 were already handled on `main` by #660.
- This branch is currently conflicting and has unresolved requested changes, while #688 is the active consolidation branch.

Keeping this PR open only for historical review context unless/until we close it.

## Original Summary

- add exhaustive unit coverage for `TrustedServerError::status_code()` so each public variant is pinned to its expected HTTP response
- add direct `tsjs` formatting tests for unified and deferred script URL/tag helpers using computed hashes
- cover invalid UTF-8 cookie-header handling and simplify duplicate callback serde rename tests

## Original Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/error.rs` | Added a table-driven test covering HTTP status code mapping for every `TrustedServerError` variant. |
| `crates/trusted-server-core/src/tsjs.rs` | Added direct unit tests for unified and deferred TS/JS script URL and tag helpers. |
| `crates/trusted-server-core/src/cookies.rs` | Added a unit test for invalid UTF-8 `Cookie` header handling in `handle_request_cookies()`. |
| `crates/trusted-server-core/src/models.rs` | Consolidated duplicate callback serde rename tests into one focused test. |

## Duplicate closure note

This PR intentionally no longer uses `Closes` keywords. Closure ownership moved to:

- #688 for #448 and #450
- #660 for #454 and #456

## Test plan

- [ ] Other: 

## Checklist

- [x] New code has tests

### DIFF
--- a/crates/trusted-server-core/src/cookies.rs
+++ b/crates/trusted-server-core/src/cookies.rs
@@ -268,6 +268,8 @@ pub fn expire_ec_cookie(settings: &Settings, response: &mut fastly::Response) {
 
 #[cfg(test)]
 mod tests {
+    use fastly::http::HeaderValue;
+
     use crate::test_support::tests::create_test_settings;
 
     use super::*;
@@ -345,6 +347,28 @@ mod tests {
             .expect("should have cookie jar");
 
         assert!(jar.iter().count() == 0);
+    }
+
+    #[test]
+    fn test_handle_request_cookies_invalid_utf8_header_value() {
+        let mut req = Request::get("http://example.com");
+        req.set_header(
+            header::COOKIE,
+            HeaderValue::from_bytes(b"session=\xFF")
+                .expect("should allow non-UTF-8 cookie header bytes"),
+        );
+
+        let error = handle_request_cookies(&req).expect_err("should reject invalid UTF-8");
+        let trusted_server_error = error.current_context();
+
+        assert!(
+            matches!(
+                trusted_server_error,
+                TrustedServerError::InvalidHeaderValue { message }
+                    if message == "Cookie header contains invalid UTF-8"
+            ),
+            "should return InvalidHeaderValue for invalid UTF-8 cookie headers",
+        );
     }
 
     #[test]
--- a/crates/trusted-server-core/src/cookies.rs
+++ b/crates/trusted-server-core/src/cookies.rs
@@ -355,7 +355,7 @@ mod tests {
         req.set_header(
             header::COOKIE,
             HeaderValue::from_bytes(b"session=\xFF")
-                .expect("should allow non-UTF-8 cookie header bytes"),
+                .expect("should construct HeaderValue from non-UTF-8 bytes"),
         );
 
         let error = handle_request_cookies(&req).expect_err("should reject invalid UTF-8");
--- a/crates/trusted-server-core/src/error.rs
+++ b/crates/trusted-server-core/src/error.rs
@@ -142,6 +142,26 @@ mod tests {
 
     #[test]
     fn status_code_maps_each_error_variant_to_expected_http_response() {
+        // Compile-time guard: adding a TrustedServerError variant without
+        // updating this test will fail to compile.
+        let _exhaustive = |error: &TrustedServerError| match error {
+            TrustedServerError::BadRequest { .. }
+            | TrustedServerError::Configuration { .. }
+            | TrustedServerError::Auction { .. }
+            | TrustedServerError::Gam { .. }
+            | TrustedServerError::GdprConsent { .. }
+            | TrustedServerError::InvalidUtf8 { .. }
+            | TrustedServerError::InvalidHeaderValue { .. }
+            | TrustedServerError::KvStore { .. }
+            | TrustedServerError::Prebid { .. }
+            | TrustedServerError::Integration { .. }
+            | TrustedServerError::Proxy { .. }
+            | TrustedServerError::Forbidden { .. }
+            | TrustedServerError::AllowlistViolation { .. }
+            | TrustedServerError::Settings { .. }
+            | TrustedServerError::Ec { .. } => (),
+        };
+
         let cases = [
             (
                 TrustedServerError::BadRequest {
--- a/crates/trusted-server-core/src/error.rs
+++ b/crates/trusted-server-core/src/error.rs
@@ -136,7 +136,115 @@ impl IntoHttpResponse for TrustedServerError {
 
 #[cfg(test)]
 mod tests {
+    use http::StatusCode;
+
     use super::*;
+
+    #[test]
+    fn status_code_maps_each_error_variant_to_expected_http_response() {
+        let cases = [
+            (
+                TrustedServerError::BadRequest {
+                    message: "missing field".into(),
+                },
+                StatusCode::BAD_REQUEST,
+            ),
+            (
+                TrustedServerError::Configuration {
+                    message: "invalid config".into(),
+                },
+                StatusCode::INTERNAL_SERVER_ERROR,
+            ),
+            (
+                TrustedServerError::Auction {
+                    message: "bid timeout".into(),
+                },
+                StatusCode::BAD_GATEWAY,
+            ),
+            (
+                TrustedServerError::Gam {
+                    message: "upstream error".into(),
+                },
+                StatusCode::BAD_GATEWAY,
+            ),
+            (
+                TrustedServerError::GdprConsent {
+                    message: "missing consent string".into(),
+                },
+                StatusCode::BAD_REQUEST,
+            ),
+            (
+                TrustedServerError::InvalidUtf8 {
+                    message: "byte 0xff".into(),
+                },
+                StatusCode::INTERNAL_SERVER_ERROR,
+            ),
+            (
+                TrustedServerError::InvalidHeaderValue {
+                    message: "non-ascii".into(),
+                },
+                StatusCode::BAD_REQUEST,
+            ),
+            (
+                TrustedServerError::KvStore {
+                    store_name: "users".into(),
+                    message: "timeout".into(),
+                },
+                StatusCode::SERVICE_UNAVAILABLE,
+            ),
+            (
+                TrustedServerError::Prebid {
+                    message: "adapter error".into(),
+                },
+                StatusCode::BAD_GATEWAY,
+            ),
+            (
+                TrustedServerError::Integration {
+                    integration: "foo".into(),
+                    message: "connection refused".into(),
+                },
+                StatusCode::BAD_GATEWAY,
+            ),
+            (
+                TrustedServerError::Proxy {
+                    message: "upstream refused".into(),
+                },
+                StatusCode::BAD_GATEWAY,
+            ),
+            (
+                TrustedServerError::Forbidden {
+                    message: "missing credentials".into(),
+                },
+                StatusCode::FORBIDDEN,
+            ),
+            (
+                TrustedServerError::AllowlistViolation {
+                    host: "blocked.example.com".into(),
+                },
+                StatusCode::FORBIDDEN,
+            ),
+            (
+                TrustedServerError::Settings {
+                    message: "parse failed".into(),
+                },
+                StatusCode::INTERNAL_SERVER_ERROR,
+            ),
+            (
+                TrustedServerError::Ec {
+                    message: "seed missing".into(),
+                },
+                StatusCode::INTERNAL_SERVER_ERROR,
+            ),
+        ];
+
+        for (error, expected_status) in cases {
+            assert_eq!(
+                error.status_code(),
+                expected_status,
+                "should map {error:?} to the expected HTTP status",
+            );
+        }
+    }
 
     #[test]
     fn server_errors_return_generic_message() {
--- a/crates/trusted-server-core/src/tsjs.rs
+++ b/crates/trusted-server-core/src/tsjs.rs
@@ -67,25 +67,40 @@ pub fn tsjs_deferred_script_tags(module_ids: &[&str]) -> String {
 
 #[cfg(test)]
 mod tests {
-    use trusted_server_js::{all_module_ids, concatenated_hash, single_module_hash};
-
     use super::*;
+
+    fn assert_sha256_hex_hash(src: &str) {
+        let hash = src
+            .rsplit_once("?v=")
+            .expect("should have cache-busting query parameter")
+            .1;
+
+        assert_eq!(hash.len(), 64, "should use a SHA-256 hex hash");
+        assert!(
+            hash.chars().all(|character| character.is_ascii_hexdigit()),
+            "should use only ASCII hex digits",
+        );
+    }
 
     #[test]
     fn tsjs_script_src_formats_unified_bundle_url_with_hash() {
-        let module_ids = ["core", "creative"];
-        let expected_hash = concatenated_hash(&module_ids);
+        let creative_src = tsjs_script_src(&["creative"]);
+        let creative_prebid_src = tsjs_script_src(&["creative", "prebid"]);
 
-        assert_eq!(
-            tsjs_script_src(&module_ids),
-            format!("/static/tsjs=tsjs-unified.min.js?v={expected_hash}"),
-            "should include the unified bundle path and cache-busting hash",
+        assert!(
+            creative_src.starts_with("/static/tsjs=tsjs-unified.min.js?v="),
+            "should include the unified bundle path and cache-busting parameter",
+        );
+        assert_sha256_hex_hash(&creative_src);
+        assert_ne!(
+            creative_src, creative_prebid_src,
+            "should generate different cache-busting URLs for different module sets",
         );
     }
 
     #[test]
     fn tsjs_script_tag_wraps_source_in_a_single_tag() {
-        let module_ids = ["core", "creative"];
+        let module_ids = ["creative"];
         let expected_src = tsjs_script_src(&module_ids);
 
         assert_eq!(
@@ -97,20 +112,16 @@ mod tests {
 
     #[test]
     fn tsjs_unified_helpers_use_all_module_ids() {
-        let all_ids = all_module_ids();
-        let expected_src = format!(
-            "/static/tsjs=tsjs-unified.min.js?v={}",
-            concatenated_hash(&all_ids)
-        );
+        let src = tsjs_unified_script_src();
 
-        assert_eq!(
-            tsjs_unified_script_src(),
-            expected_src,
-            "should hash all compiled modules for the unified bundle",
+        assert!(
+            src.starts_with("/static/tsjs=tsjs-unified.min.js?v="),
+            "should include the unified bundle path and cache-busting parameter",
         );
+        assert_sha256_hex_hash(&src);
         assert_eq!(
             tsjs_unified_script_tag(),
-            format!("<script src=\"{expected_src}\" id=\"trustedserver-js\"></script>"),
+            format!("<script src=\"{src}\" id=\"trustedserver-js\"></script>"),
             "should wrap the unified bundle source in the standard script tag",
         );
     }
@@ -134,11 +145,11 @@ mod tests {
     }
 
     #[test]
-    fn tsjs_deferred_script_src_uses_empty_hash_for_unknown_module() {
+    fn tsjs_deferred_script_src_documents_unregistered_module_fallback() {
         assert_eq!(
             tsjs_deferred_script_src("unknown-module"),
             "/static/tsjs=tsjs-unknown-module.min.js?v=",
-            "should fall back to an empty hash for unknown deferred modules",
+            "should preserve the current fallback until callers validate registered modules",
         );
     }
 
--- a/crates/trusted-server-core/src/tsjs.rs
+++ b/crates/trusted-server-core/src/tsjs.rs
@@ -112,16 +112,21 @@ mod tests {
 
     #[test]
     fn tsjs_unified_helpers_use_all_module_ids() {
-        let src = tsjs_unified_script_src();
+        let unified = tsjs_unified_script_src();
+        let manual = tsjs_script_src(&all_module_ids());
 
+        assert_eq!(
+            unified, manual,
+            "unified helpers must hash every registered module",
+        );
         assert!(
-            src.starts_with("/static/tsjs=tsjs-unified.min.js?v="),
+            unified.starts_with("/static/tsjs=tsjs-unified.min.js?v="),
             "should include the unified bundle path and cache-busting parameter",
         );
-        assert_sha256_hex_hash(&src);
+        assert_sha256_hex_hash(&unified);
         assert_eq!(
             tsjs_unified_script_tag(),
-            format!("<script src=\"{src}\" id=\"trustedserver-js\"></script>"),
+            format!("<script src=\"{unified}\" id=\"trustedserver-js\"></script>"),
             "should wrap the unified bundle source in the standard script tag",
         );
     }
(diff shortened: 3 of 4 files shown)


crates/trusted-server-core/src/error.rs
```diff
@@ -136,8 +136,116 @@ impl IntoHttpResponse for TrustedServerError {
...
+        ];
+
+        for (error, expected_status) in cases {
+            assert_eq!(
+                error.status_code(),
+                expected_status,
+                "should map {error:?} to the expected HTTP status",
+            );
+        }
+    }
```
aram356, line 267: ♻️ **refactor** — The table-driven `cases` array is hand-maintained. If a new `TrustedServerError` variant is added and `status_code()` is updated but this test isn't, the test silently passes despite the new variant being uncovered. Since the whole point of this test is to lock in the variant→status mapping, a compile-time exhaustiveness guard would be much stronger.

**Fix** — add a non-exhaustive `match` somewhere in the test that fails to compile if a variant is added:

```rust
// Compile-time guard: adding a TrustedServerError variant without
// updating this test will fail to compile.
let _exhaustive = |e: &TrustedServerError| match e {
    TrustedServerError::BadRequest { .. }
    | TrustedServerError::Configuration { .. }
    | TrustedServerError::Auction { .. }
    | TrustedServerError::Gam { .. }
    | TrustedServerError::GdprConsent { .. }
    | TrustedServerError::InvalidUtf8 { .. }
    | TrustedServerError::InvalidHeaderValue { .. }
    | TrustedServerError::KvStore { .. }
    | TrustedServerError::Prebid { .. }
    | TrustedServerError::Integration { .. }
    | TrustedServerError::Proxy { .. }
    | TrustedServerError::Forbidden { .. }
    | TrustedServerError::AllowlistViolation { .. }
    | TrustedServerError::Settings { .. }
    | TrustedServerError::Ec { .. } => (),
};
```

crates/trusted-server-core/src/tsjs.rs
```diff
@@ -64,3 +64,96 @@ pub fn tsjs_deferred_script_tags(module_ids: &[&str]) -> String {
...
+    fn tsjs_script_src_formats_unified_bundle_url_with_hash() {
+        let module_ids = ["core", "creative"];
+        let expected_hash = concatenated_hash(&module_ids);
+
+        assert_eq!(
+            tsjs_script_src(&module_ids),
+            format!("/static/tsjs=tsjs-unified.min.js?v={expected_hash}"),
+            "should include the unified bundle path and cache-busting hash",
+        );
+    }
```
aram356, line 99: ♻️ **refactor** — `expected_hash` is derived by calling `concatenated_hash` — the same function the production code calls. The test verifies "function composition matches itself" instead of cache-busting *behavior*. If `concatenated_hash` silently regressed (e.g. always returned `""`) both production and test would change in lockstep and the test would still pass.

**Fix** — assert hash structure (64 hex chars for SHA-256) and that different module sets yield *different* URLs:

```rust
let a = tsjs_script_src(&["creative"]);
let b = tsjs_script_src(&["creative", "prebid"]);
assert_ne!(a, b, "different module sets must yield different cache-busting URLs");
let hash = a.rsplit_once("?v=").expect("should have v= param").1;
assert_eq!(hash.len(), 64, "should be SHA-256 hex");
assert!(hash.chars().all(|c| c.is_ascii_hexdigit()));
```

The same concern applies to `tsjs_unified_helpers_use_all_module_ids` below.

crates/trusted-server-core/src/tsjs.rs
```diff
@@ -64,3 +64,96 @@ pub fn tsjs_deferred_script_tags(module_ids: &[&str]) -> String {
         .map(|id| tsjs_deferred_script_tag(id))
         .collect::<String>()
 }
+
+#[cfg(test)]
+mod tests {
+    use trusted_server_js::{all_module_ids, concatenated_hash, single_module_hash};
```
aram356: ♻️ **refactor** — Duplicate of the import at the top of the file. Submodules can see private items of ancestors, so `use super::*;` at line 72 is sufficient. Verified locally: removing this line keeps both `cargo build` and `cargo test` green.

**Fix** — delete this line.

crates/trusted-server-core/src/tsjs.rs
```diff
@@ -64,3 +64,96 @@ pub fn tsjs_deferred_script_tags(module_ids: &[&str]) -> String {
...
+    }
+
+    #[test]
+    fn tsjs_deferred_script_src_uses_empty_hash_for_unknown_module() {
+        assert_eq!(
+            tsjs_deferred_script_src("unknown-module"),
+            "/static/tsjs=tsjs-unknown-module.min.js?v=",
+            "should fall back to an empty hash for unknown deferred modules",
+        );
+    }
```
aram356, line 158: 🤔 **thinking** — This test asserts that calling `tsjs_deferred_script_src("unknown-module")` produces `"/static/tsjs=tsjs-unknown-module.min.js?v="` — an empty hash that points at a non-existent file → silent 404 in the browser. Capturing this as "what we expect" makes future improvements to the production `unwrap_or_default()` (e.g. logging or returning `Option<String>`) look like regressions when they're actually fixes.

**Suggestion** — either re-frame the test so its name describes a documented contract (e.g. "caller is responsible for not generating tags for unregistered modules") and add a `log::warn!` in production for unknown IDs, or mark this as a known-bad behavior with a follow-up issue.

crates/trusted-server-core/src/cookies.rs
```diff
@@ -347,6 +349,28 @@ mod tests {
...
+
+        assert!(
+            matches!(
+                trusted_server_error,
+                TrustedServerError::InvalidHeaderValue { message }
+                    if message == "Cookie header contains invalid UTF-8"
+            ),
+            "should return InvalidHeaderValue for invalid UTF-8 cookie headers",
+        );
+    }
```
aram356, line 372: 📝 **note** — Issue #454 says "returning `None`" but `handle_request_cookies` returns `Err(InvalidHeaderValue)` for non-UTF-8 cookie headers. The test correctly verifies the *current* implementation. Just flagging that the issue text is stale (likely written when the function had a different signature) — the issue can still be closed; no action on this PR.

crates/trusted-server-core/src/tsjs.rs
```diff
@@ -64,3 +64,96 @@ pub fn tsjs_deferred_script_tags(module_ids: &[&str]) -> String {
...
+
+#[cfg(test)]
+mod tests {
+    use trusted_server_js::{all_module_ids, concatenated_hash, single_module_hash};
+
+    use super::*;
+
+    #[test]
+    fn tsjs_script_src_formats_unified_bundle_url_with_hash() {
+        let module_ids = ["core", "creative"];
```
aram356: ⛏ **nitpick** — `bundle.rs::concatenate_modules` always prepends `"core"` and skips it when present in `ids` (bundle.rs:31-39). Passing `["core", "creative"]` produces the same hash as `["creative"]`. Could be simplified to `["creative"]` to avoid implying caller-side intent that doesn't matter. Same applies to line 88.

crates/trusted-server-core/src/cookies.rs
```diff
@@ -347,6 +349,28 @@ mod tests {
         assert!(jar.iter().count() == 0);
     }
 
+    #[test]
+    fn test_handle_request_cookies_invalid_utf8_header_value() {
+        let mut req = Request::get("http://example.com");
+        req.set_header(
+            header::COOKIE,
+            HeaderValue::from_bytes(b"session=\xFF")
+                .expect("should allow non-UTF-8 cookie header bytes"),
```
aram356: ⛏ **nitpick** — The expect message reads like a property of `handle_request_cookies`, but it's actually about `HeaderValue::from_bytes` accepting any opaque bytes. Consider `"should construct HeaderValue from non-UTF-8 bytes"` for clarity.

aram356 (changes requested):
## Summary

Pure test additions/consolidation — no production code changes. Tests pass on `wasm32-wasip1`, clippy is clean, and CI is fully green. The substantive concerns below are about test *design* — what these tests would catch on regression — rather than correctness of what was added. Requesting changes primarily on the exhaustiveness and tautology issues; the rest are non-blocking.

## Blocking

### ♻️ refactor

- **Status code mapping test is not exhaustive at compile time** — `error.rs:144-247`. Hand-maintained `cases` array means a new variant added to `TrustedServerError` won't fail this test. A non-exhaustive `match` guard would force exhaustive coverage. See inline comment.
- **tsjs hash assertions are tautological** — `tsjs.rs:75-116`. `expected_hash` is derived by calling the same `concatenated_hash` the production code calls. If `concatenated_hash` regressed, both sides would change in lockstep. Suggest asserting hash structure (64 hex chars) and that different module sets yield different URLs. See inline comment.
- **Redundant import in tsjs test module** — `tsjs.rs:70`. Duplicates the top-of-file import; `use super::*` at line 72 is sufficient (verified locally).

## Non-blocking

### 🤔 thinking

- **`tsjs_deferred_script_src_uses_empty_hash_for_unknown_module` locks in silent failure** — `tsjs.rs:137-143`. Empty hash produces a 404-bound URL with no logging. Capturing this as "expected" makes future improvements look like regressions.

### 🌱 seedling

- **Production `unwrap_or_default()` for unknown deferred modules is invisible** — `tsjs.rs:44` (unchanged in this PR, so noted in the body). `single_module_hash(module_id).unwrap_or_default()` silently produces a 404-bound URL with no logging when the module isn't registered. Worth a follow-up issue to add `log::warn!` for unregistered module IDs.

### 📝 note

- **Issue #454 wording vs implementation** — `cookies.rs:353-372`. Issue says "returning `None`" but production returns `Err`. Test correctly verifies current implementation; issue text is stale.

### ⛏ nitpick

- **`["core", "creative"]` redundantly includes `core`** — `tsjs.rs:76, 88`. `concatenate_modules` always prepends core; could be simplified to `["creative"]`.
- **`HeaderValue::from_bytes` expect message is slightly misleading** — `cookies.rs:358`. Reads like a property of `handle_request_cookies` but is about `HeaderValue::from_bytes`.

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS (verified locally on `wasm32-wasip1`)
- vitest: PASS
- browser & integration tests: PASS

crates/trusted-server-core/src/error.rs
ChristianPavilonis: Addressed in 02ac32f2. Added an exhaustive, no-wildcard match guard covering every current TrustedServerError variant, while keeping the table-driven status assertions for readability.

crates/trusted-server-core/src/tsjs.rs
ChristianPavilonis: Addressed in 02ac32f2. Removed the tautological concatenated_hash expectations and now assert URL prefix, SHA-256 hex hash structure, and that distinct module sets produce distinct cache-busting URLs.

crates/trusted-server-core/src/tsjs.rs
ChristianPavilonis: Addressed in 02ac32f2. Removed the redundant test-module import and rely on use super::*.

crates/trusted-server-core/src/tsjs.rs
ChristianPavilonis: Addressed in 02ac32f2 by reframing the test as documentation of the current unregistered-module fallback rather than an ideal behavior. I left production behavior unchanged for this test-only PR; adding a warn for unregistered IDs is better as follow-up work.

crates/trusted-server-core/src/cookies.rs
ChristianPavilonis: Acknowledged. No PR code change needed here; the test continues to verify the current Err(InvalidHeaderValue) behavior despite the stale issue wording.

crates/trusted-server-core/src/tsjs.rs
ChristianPavilonis: Addressed in 02ac32f2. Simplified the test inputs from ["core", "creative"] to ["creative"] and use ["creative", "prebid"] where a distinct module set is needed.

crates/trusted-server-core/src/cookies.rs
ChristianPavilonis: Addressed in 02ac32f2. Updated the expect message to "should construct HeaderValue from non-UTF-8 bytes".

crates/trusted-server-core/src/tsjs.rs
```diff
@@ -64,3 +64,107 @@ pub fn tsjs_deferred_script_tags(module_ids: &[&str]) -> String {
...
+            tsjs_unified_script_tag(),
+            format!("<script src=\"{src}\" id=\"trustedserver-js\"></script>"),
+            "should wrap the unified bundle source in the standard script tag",
+        );
+    }
+
+    #[test]
+    fn tsjs_deferred_helpers_format_single_module_urls_and_tags() {
+        let module_id = "prebid";
+        let expected_hash = single_module_hash(module_id).expect("should hash known module");
```
aram356: 🔧 **wrench** — This re-introduces the same tautology pattern the prior review flagged for `concatenated_hash`, just shifted to `single_module_hash`. `expected_hash` is computed by calling the same hash function the production code calls — if `single_module_hash` regressed to return `String::new()`, both `tsjs_deferred_script_src(module_id)` and `expected_hash` would change in lockstep and this test would still pass (equivalent to the documented unregistered-module case at line 148-154).

**Fix** — mirror the unified-bundle approach: assert URL prefix and SHA-256 hex hash structure, not equality with another call to the same hash function. For the tag wrap test, derive `expected_src` from `tsjs_deferred_script_src` itself — that's verifying the wrapping format, not re-computing the hash.

```rust
let src = tsjs_deferred_script_src("prebid");
assert!(
    src.starts_with("/static/tsjs=tsjs-prebid.min.js?v="),
    "should include the deferred module path and cache-busting parameter",
);
assert_sha256_hex_hash(&src);

assert_eq!(
    tsjs_deferred_script_tag("prebid"),
    format!("<script src=\"{src}\" defer></script>"),
    "should render a deferred script tag for the module",
);
```

crates/trusted-server-core/src/tsjs.rs
```diff
@@ -64,3 +64,107 @@ pub fn tsjs_deferred_script_tags(module_ids: &[&str]) -> String {
...
+            tsjs_deferred_script_src("unknown-module"),
+            "/static/tsjs=tsjs-unknown-module.min.js?v=",
+            "should preserve the current fallback until callers validate registered modules",
+        );
+    }
+
+    #[test]
+    fn tsjs_deferred_script_tags_concatenates_tags_in_input_order() {
+        let module_ids = ["prebid", "creative"];
+        let expected = module_ids
```
aram356: 🔧 **wrench** — Test body is byte-identical to the production function body (tsjs.rs:62-65):

```rust
// production
module_ids.iter().map(|id| tsjs_deferred_script_tag(id)).collect::<String>()

// test
let expected = module_ids.iter().map(|id| tsjs_deferred_script_tag(id)).collect::<String>();
assert_eq!(tsjs_deferred_script_tags(&module_ids), expected, ...);
```

Any symmetric regression in either side passes the test. The test name claims to verify input-order concatenation, but the test never actually checks that the modules appear in input order — it just checks that re-running the same expression yields the same result.

**Fix** — assert the *property* (order preservation, count of tags), not equality with a re-execution of the same code:

```rust
let result = tsjs_deferred_script_tags(&["prebid", "creative"]);
let prebid_pos = result.find("tsjs-prebid").expect("should contain prebid script");
let creative_pos = result.find("tsjs-creative").expect("should contain creative script");
assert!(prebid_pos < creative_pos, "should preserve input order");
assert_eq!(
    result.matches("<script ").count(),
    2,
    "should emit one tag per module",
);
```

crates/trusted-server-core/src/tsjs.rs
```diff
@@ -64,3 +64,107 @@ pub fn tsjs_deferred_script_tags(module_ids: &[&str]) -> String {
...
+
+        assert_eq!(
+            tsjs_script_tag(&module_ids),
+            format!("<script src=\"{expected_src}\" id=\"trustedserver-js\"></script>"),
+            "should render the injected trustedserver script tag",
+        );
+    }
+
+    #[test]
+    fn tsjs_unified_helpers_use_all_module_ids() {
```
aram356, line 114: 🔧 **wrench** — Test name advertises "use all module IDs" but the body never verifies that contract. It only checks URL prefix, SHA-256 hex hash structure, and that `tsjs_unified_script_tag` wraps `tsjs_unified_script_src`. If `tsjs_unified_script_src()` were silently changed to `tsjs_script_src(&[])`, the test would still pass — the unified bundle path would still match, the empty hash would still be 64 hex chars (well, it would actually be the hash of just core, which is still a valid SHA-256 hex), and the tag wrap assertion would still hold.

**Fix** — make the contract executable by asserting equivalence with the manual all-modules call:

```rust
let unified = tsjs_unified_script_src();
let manual = tsjs_script_src(&all_module_ids());
assert_eq!(
    unified, manual,
    "unified helpers must hash every registered module",
);
assert!(
    unified.starts_with("/static/tsjs=tsjs-unified.min.js?v="),
    "should include the unified bundle path and cache-busting parameter",
);
assert_sha256_hex_hash(&unified);
assert_eq!(
    tsjs_unified_script_tag(),
    format!("<script src=\"{unified}\" id=\"trustedserver-js\"></script>"),
    "should wrap the unified bundle source in the standard script tag",
);
```

This closes the gap: any divergence between `tsjs_unified_script_src` and the all-modules contract is caught directly.

aram356 (changes requested):
## Summary

Pure test additions — no production code changes. CI is fully green and 809 `trusted-server-core` tests pass locally on `wasm32-wasip1`. The exhaustiveness guard added in 02ac32f for `status_code()` mapping is a strong improvement. However, two of the four "addressed" items still contain the same self-referential pattern the prior review flagged — the fix landed for `concatenated_hash` but the same shape was preserved (or freshly introduced) for `single_module_hash` and the deferred-tags collection helper. Plus one test whose name advertises a contract its body does not verify. Requesting changes on those three; the rest is non-blocking.

## Blocking

### 🔧 wrench

- **`tsjs_deferred_helpers_format_single_module_urls_and_tags` is tautological** — `expected_hash` is derived by calling the same `single_module_hash` the production code calls (tsjs.rs:130-145). Same critique the prior review raised against `concatenated_hash`, shifted helper. See inline comment.
- **`tsjs_deferred_script_tags_concatenates_tags_in_input_order` is byte-identical to production** — test body literally re-executes `module_ids.iter().map(|id| tsjs_deferred_script_tag(id)).collect::<String>()` and asserts equality with the production function (tsjs.rs:156-169). Any symmetric regression passes. See inline comment.
- **`tsjs_unified_helpers_use_all_module_ids` does not verify the contract its name advertises** — never asserts that `all_module_ids()` is consulted or that every module is in the hash (tsjs.rs:113-127). Would still pass if production silently degraded to `tsjs_script_src(&[])`. See inline comment.

## Non-blocking

### 🤔 thinking

- **Status code `cases` array still hand-maintained** — error.rs:165-258. The `_exhaustive` closure (lines 147-163) forces compile-fail on a new variant — good — but does not force adding the new variant to `cases`. A developer could update the closure to silence the compile error and forget the case. Stronger alternative: derive `cases` from the same exhaustive arm pattern, or `assert_eq!(cases.len(), <const tied to closure>)`. Not blocking — closure catches the common slip — but worth a follow-up.

### 🌱 seedling

- **Follow-up: log unregistered deferred module IDs** — tsjs.rs:44. `single_module_hash(module_id).unwrap_or_default()` silently produces a 404-bound URL with no logging when the module isn't registered. The PR now documents this as expected behavior at tsjs.rs:148-154; tracking issue would be useful so the test's "documented fallback" framing has a paired plan to remove it.

### ⛏ nitpick

- **Inconsistent test naming convention within this PR** — new tests in `error.rs`/`tsjs.rs` use the descriptive style without `test_` prefix; the new test in `cookies.rs` uses `test_` prefix matching its file-local neighbors. The codebase has both styles, so this is fine — just a note for future consistency.

### 📝 note

- **Issue #454 wording vs implementation** — already noted in the prior review; cookies.rs test correctly verifies current `Err(InvalidHeaderValue)` behavior despite the issue text saying "returns `None`". No PR action needed.
- **Branch is 4 commits behind `origin/main`** — no conflicting paths, but a rebase before merge is standard.

## CI Status

- fmt: PASS
- clippy: PASS
- cargo test: PASS (verified locally on `wasm32-wasip1`, 809 trusted-server-core tests)
- vitest: PASS
- browser & integration tests: PASS
- CodeQL: PASS

crates/trusted-server-core/src/tsjs.rs
ChristianPavilonis: Addressed in f31b00e1. Removed the direct single_module_hash expectation, assert the deferred URL prefix plus SHA-256 hex hash structure, and derive the tag expectation from tsjs_deferred_script_src so the assertion only verifies wrapping format.

crates/trusted-server-core/src/tsjs.rs
ChristianPavilonis: Addressed in f31b00e1. Replaced the byte-identical expected string construction with property assertions: the output contains both module URLs in the requested order, emits exactly one script tag per module, and marks each tag as defer.

crates/trusted-server-core/src/tsjs.rs
ChristianPavilonis: Addressed in f31b00e1. The test now compares tsjs_unified_script_src() directly against tsjs_script_src(&all_module_ids()) before retaining the URL-shape, SHA-256 hash, and wrapper assertions.

ChristianPavilonis:
Superseded during #396 backlog consolidation: #688 is now the canonical branch for #448/#450, and #454/#456 were already handled on `main` by #660. I updated this PR description to remove duplicate `Closes` keywords so issues have a single closure path.

PR #610, opened 2026-04-02
Move logging initialization into Fastly adapter (PR 10)

prk-Jr:
## Summary

- Move Fastly logging initialization into an adapter-local module so backend setup stays owned by the Fastly adapter instead of leaking through `main.rs`.
- Remove `log-fastly` from `trusted-server-core` so core remains backend-agnostic and future Edgezero adapters can rely on the shared `log` facade only.
- Capture the PR 10 design and implementation notes used to keep this stacked migration step scoped against the PR 9 base branch.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-adapter-fastly/src/logging.rs` | Add the adapter-local Fastly logger initializer, target-label helper, and unit test. |
| `crates/trusted-server-adapter-fastly/src/main.rs` | Call `logging::init_logger()` and remove the inline Fastly logger implementation. |
| `crates/trusted-server-core/Cargo.toml` | Remove the `log-fastly` dependency from core. |
| `Cargo.lock` | Update the lockfile after removing core's `log-fastly` dependency edge. |
| `docs/superpowers/specs/2026-04-02-pr10-logging-initialization-design.md` | Add the PR 10 design note and scope. |
| `docs/superpowers/plans/2026-04-02-pr10-logging-initialization.md` | Add the implementation plan for the logging boundary change. |

## Closes

Closes #491

## Test plan

- [x] `cargo test --workspace`
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [x] JS tests: `cd crates/js/lib && npx vitest run`
- [x] JS format: `cd crates/js/lib && npm run format`
- [x] Docs format: `cd docs && npm run format`
- [ ] WASM build: `cargo build --package trusted-server-adapter-fastly --release --target wasm32-wasip1`
- [ ] Manual testing via `fastly compute serve`
- [ ] Other:

## Hardening note

Not applicable. This PR does not touch config-derived regex or pattern compilation.

## Checklist

- [x] Changes follow [CLAUDE.md](/CLAUDE.md) conventions
- [x] No `unwrap()` in production code — use `expect("should ...")`
- [x] Uses `log` macros (not `println!`)
- [x] New code has tests
- [x] No secrets or credentials committed

PR #650, opened 2026-04-22
RSL AI crawler licensing design spec

### LINKED ISSUE #649: Write Spec for RSL Support
### Description

Write Spec for RSL Support

jevansnyc:
## Summary

Adds design spec for Trusted Server's RSL-compliant AI crawler detection and licensing enforcement layer.

- Edge-deployed AI crawler classification using six signals (UA, IP allowlist, JA4, ASN, H2, robots/license fetch correlation)
- RSL 1.0 standards-compliant license publishing (`/license.xml`, `robots.txt` augmentation, `Link` header)
- Public `license.toml` for RSL terms; private `license.private.toml` for enforcement secrets and commercial overrides
- Standards-compliant 402/403 enforcement responses with inline RSL fragments
- Permissive-by-default with per-publisher/per-route Strict override
- Debug endpoints (`/_ts/debug/rsl/summary`, `/_ts/debug/rsl/recent`, `/_ts/debug/rsl/license`) and structured logging
- Integrates with existing TS architecture; no changes needed to Edge Cookie, auction orchestrator, consent, or other integrations
- Phase 2 preview for Open License Protocol (OLP) token-based access

## Test plan

- [ ] Review spec for accuracy against current TS infrastructure (integration hooks, JA4 signals, bot gate)
- [ ] Verify RSL usage/payment vocabulary matches RSL 1.0 spec (https://rslstandard.org/rsl)
- [ ] Validate onboarding flow assumptions against an existing TS publisher deployment
- [ ] Confirm binary size estimates (~100 KB additional for IP allowlists + JA4 DB + new code)

Closes https://github.com/IABTechLab/trusted-server/issues/649

### DIFF
--- a/docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
+++ b/docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
@@ -1,0 +1,1055 @@
+# Trusted Server AI Crawler Licensing (RSL-compliant)
+
+*April 2026*
+
+---
+
+## 1. Product Positioning & Scope
+
+### 1.1 What It Is
+
+An edge-deployed AI crawler detection and RSL licensing enforcement layer for
+Trusted Server publishers. TS classifies incoming requests against published AI
+crawler fingerprints, serves publisher-defined RSL licensing terms as
+machine-readable XML, and enforces per-route access decisions with
+standards-compliant HTTP responses (402 Payment Required for honest crawlers,
+403 Forbidden for stealth or prohibited crawlers).
+
+### 1.2 What It Is Not
+
+- Not a payment/billing system (phase 2)
+- Not a proprietary licensing protocol — RSL-native, standards-aligned
+- Not a reverse proxy — runs in-line at the edge, no redirect, no subdomain
+- Not a bot-blocking product in general — scoped to AI crawler licensing
+
+### 1.3 Audience
+
+**Primary buyer:** Large publishers and publisher consortiums (Hearst, Arena
+Group, Condé Nast-tier) who already run TS or are evaluating it.
+
+**Secondary audience:** Entitlement platforms that want to transact with these
+publishers at scale without bespoke per-publisher integrations.
+
+### 1.4 Value Proposition
+
+1. **RSL-native, standards-compliant** — publishes `license.xml`, honors OLP
+   where applicable, interoperates with any compliant entitlement platform.
+2. **Edge-native, low-latency** — classification and enforcement at the CDN
+   layer, no proxy hop, no subdomain redirect.
+3. **Multi-signal detection including JA4** — TLS-layer fingerprinting catches
+   crawlers that spoof user-agent strings.
+4. **Publisher-owned config** — single `license.toml` file, version-controlled,
+   no lock-in to a vendor's dashboard.
+5. **Open source** — publishers can audit the enforcement behavior.
+
+### 1.5 Reference to IAB Tech Lab CoMP
+
+The IAB Tech Lab Content Monetization Protocol (CoMP) is a complementary
+commercial framework for AI content licensing. It is currently in working group
+formation. This spec references CoMP only as a future-compatible framework;
+this POC does not build against CoMP endpoints or schemas. When CoMP
+stabilizes, the TS RSL implementation can extend to emit or consume CoMP
+commercial signals without changing the core detection or enforcement layers.
+
+---
+
+## 2. Scope Decisions
+
+### 2.1 In Scope (POC / MVP)
+
+1. AI crawler detection via six signals (see Section 4)
+2. Publisher-authored public RSL terms via `license.toml`
+3. Publisher-authored private enforcement rules via `license.private.toml`
+4. `/license.xml` generation served from publisher's first-party domain
+5. `/robots.txt` augmentation with `License:` directive
+6. `Link: rel="license"` HTTP header on all responses
+7. Enforcement actions: 200 (allow), 402 (honest crawler blocked), 403 (stealth
+   or prohibited)
+8. Debug endpoints: `/_ts/debug/rsl/summary`, `/_ts/debug/rsl/recent`,
+   `/_ts/debug/rsl/license`
+9. Structured logging of every classified request
+10. Permissive-by-default mode with per-route or publisher-wide Strict override
+11. Crawler-specific overrides (allow/deny/enforce-default) via private config
+12. Integration with existing TS architecture without disrupting other
+    integrations (Monetize, Edge Cookie, consent, PBS, etc.)
+
+### 2.2 Out of Scope (Deferred)
+
+1. **OLP license server** — token issuance, `/token`/`/introspect`/`/key`
+   endpoints. Phase 2.
+2. **Billing, invoicing, payment rails** — no money moves in the POC. RSL 402
+   responses point to the publisher's contact information for out-of-band
+   negotiation.
+3. **Encrypted Media Standard (EMS)** — content encryption requires the OLP
+   `/key` endpoint. Phase 2.
+4. **Behavioral anomaly detection** — request rate/pattern analysis, path-depth
+   heuristics, referer-chain analysis. Future phase if POC metrics justify it.
+5. **Publisher SaaS dashboard** — structured logs + debug endpoints only.
+   Publishers render their own visualizations from the log stream.
+6. **AI-company cooperation agreements** — TS publishes standards-compliant
+   signals and assumes honest actors will respect them. TS does not negotiate
+   deals on publishers' behalf.
+7. **Multi-publisher consortium management** — single publisher at a time for
+   POC. Consortium config patterns (Hearst-style) come post-POC.
+8. **CoMP framework integration** — referenced only; not built against.
+9. **Creative / content transformation for AI consumers** — TS does not rewrite
+   content for AI, generate summaries, or serve different versions.
+10. **CAPTCHA / proof-of-human challenges** — Apple Private Access Tokens are
+    available in TS if a publisher wants cryptographic human attestation, but
+    not required for this POC. JA4 + IP + UA signals are sufficient for AI
+    crawler classification.
+
+---
+
+## 3. Architecture
+
+### 3.1 High-Level
+
+```text
+┌────────────────────────────────────────────────────────────────────┐
+│  Trusted Server — Edge Compute (WASM on Fastly today)              │
+│  Planned: Akamai EdgeWorkers, Cloudflare Workers                   │
+│                                                                    │
+│  ┌───────────────────────────────────────────────────────────┐     │
+│  │  Request arrives (human or crawler, any path)             │     │
+│  └──────┬────────────────────────────────────────────────────┘     │
+│         │                                                          │
+│         ▼                                                          │
+│  ┌──────────────────────┐   ┌──────────────────────┐               │
+│  │ RSL Classifier       │   │ License.toml Loader  │               │
+│  │ (JA4 + UA + IP + ASN)│   │ (route → terms)      │               │
+│  └──────┬───────────────┘   └──────┬───────────────┘               │
+│         │                          │                               │
+│         ▼                          ▼                               │
+│  ┌──────────────────────────────────────────────────────────┐      │
+│  │  Enforcement Decision                                    │      │
+│  │  - classified as: {human | honest_ai | stealth_ai}       │      │
+│  │  - mode: {permissive | strict}                           │      │
+│  │  - license terms for this route                          │      │
+│  └──────┬───────────────────────────────────────────────────┘      │
+│         │                                                          │
+│         ├──> [allow] pass through to origin/integration            │
+│         ├──> [402] RSL-compliant license-required response         │
+│         └──> [403] forbidden (stealth or prohibited)               │
+│                                                                    │
+│  Special routes (always served by TS):                             │
+│  /license.xml           — generated from license.toml              │
+│  /robots.txt            — augmented with License: directive        │
+│  /_ts/debug/rsl/summary — dashboard JSON                           │
+│  /_ts/debug/rsl/recent  — recent classifications                   │
+│  /_ts/debug/rsl/license — verify published terms                   │
+└────────────────────────────────────────────────────────────────────┘
+```
+
+### 3.2 Functional Units
+
+**RSL Classifier.** Given a request, returns a classification verdict:
+`{category, confidence, signals_matched, crawler_identity}`. Pure function over
+request features (TLS fingerprint, headers, source IP, ASN lookup, UA string).
+Consults static allowlists of published AI crawler fingerprints refreshed
+periodically from `openai.com/gptbot.json`, Anthropic published ranges, and
+equivalent.
+
+**License Resolver.** Given a request path, returns the applicable RSL terms
+from `license.toml` (default + most-specific route override). Produces a
+`LicenseTerms` struct used by both the enforcement layer and the
+`/license.xml` generator.
+
+**Enforcement Layer.** Takes the classification verdict, license terms, and
+publisher mode; produces an `Action`: `{Allow, Challenge402(reason, link),
+Forbid403(reason)}`. Applied to the response before hitting origin.
+
+### 3.3 Data Flow
+
+```text
+Request → Classifier → Verdict ──┐
+                                 ├──> Enforcement → Action → Response
+License.toml → Resolver → Terms ─┘                          │
+                                                            ▼
+                                                    Structured log
+                                                    + Debug endpoint state
+```
+
+### 3.4 State Footprint at Edge
+
+- AI crawler IP/UA/JA4 allowlists: compiled into the WASM binary at build
+  time, ~tens of KB, refreshed when TS is rebuilt with updated allowlists.
+- `license.toml` and `license.private.toml`: compiled into the WASM binary at
+  build time (same mechanism as other TS configs).
+- Per-request classification: emitted as structured log lines, optionally
+  counted in a small in-memory ring buffer for debug endpoints (no KV writes
+  on the hot path).
+
+### 3.5 Integration with Existing TS Infrastructure
+
+| Existing capability | How RSL uses it |
+|---|---|
+| `IntegrationRegistration` builder | New hook types: `with_request_classifier`, `with_special_route_augmenter` |
+| JA4 signal from edge TLS | Input to `classifier::classify()` |
+| Bot gate (H2 + JA4) | Supporting signal for stealth detection |
+| `/robots.txt` handling | Integration augments existing `robots.txt` response |
+| `/_ts/debug/*` auth pattern | Debug endpoints reuse existing token auth |
+| Structured logging (`log-fastly`) | Classification events emitted as structured log lines |
+| Settings (`trusted-server.toml`) | RSL config block added to existing settings parser |
+
+**No changes required to:** Edge Cookie, auction orchestrator / PBS integration,
+Monetize ad-server client, consent handling, existing integrations (Permutive,
+Lockr, Datadome, etc.), HTML processor, cache layer.
+
+### 3.6 Request-Path Insertion Point
+
+RSL classification runs **after** the bot gate (so RSL sees classified
+bot-or-not state) but **before** any integration that might touch the response
+body (so RSL can 402/403 before wasted work).
+
+Integration registration:
+
+```rust
+IntegrationRegistration::builder("rsl")
+    .with_request_classifier()       // runs classifier, attaches verdict to request context
+    .with_special_route("/license.xml")
+    .with_special_route_augmenter("/robots.txt")
+    .with_response_modifier()        // adds Link header on all responses
+    .with_debug_routes(&[
+        "/_ts/debug/rsl/summary",
+        "/_ts/debug/rsl/recent",
+        "/_ts/debug/rsl/license",
+    ])
+    .build()
+```
+
+`with_request_classifier()` and `with_special_route_augmenter()` are new hook
+types added to the integration framework. All other mechanisms are existing
+patterns.
+
+### 3.7 Module Structure
+
+```text
+crates/trusted-server-core/src/rsl/
+├── mod.rs               # public API + IntegrationRegistration
+├── classifier.rs        # AI crawler detection logic
+├── config/
+│   ├── public.rs        # license.toml parser → LicenseTerms
+│   └── private.rs       # license.private.toml parser → EnforcementRules
+├── xml_generator.rs     # LicenseTerms → RSL XML document
+├── fingerprints/
+│   ├── ua_patterns.rs   # static honest-UA matchers
+│   ├── ip_allowlists.rs # compiled-in crawler operator IP ranges
+│   └── ja4_db.rs        # JA4 fingerprint database for LLM fetchers
+├── enforcement.rs       # verdict + terms + mode → Action
+├── endpoints.rs         # /license.xml, /robots.txt augmentation, debug routes
+└── logging.rs           # structured log emission
+```
+
+### 3.8 Dependencies
+
+- `quick-xml` (or existing XML crate if one is already present) for
+  `license.xml` generation
+- No new heavy dependencies — the classifier is pure pattern-matching against
+  compiled-in data structures; no ML, no external services
+
+### 3.9 Binary Size Impact
+
+- IP allowlists compiled in: ~30-50 KB (a few thousand CIDR ranges from major
+  operators)
+- JA4 fingerprint database: ~5-10 KB (a few hundred common LLM fetcher
+  fingerprints)
+- UA pattern table: negligible
+- Total new code: estimated 2-3K lines of Rust, well within existing crate
+  structure
+
+---
+
+## 4. Bot Detection Signals
+
+### 4.1 Six Signals
+
+| # | Signal | Source | Strength | Coverage |
+|---|---|---|---|---|
+| 1 | **Honest User-Agent match** | HTTP `User-Agent` header | Definitive when paired with #2 | GPTBot, ClaudeBot, Claude-User, Claude-SearchBot, PerplexityBot, Perplexity-User, Google-Extended, CCBot, Bytespider, Amazonbot, Applebot-Extended, OAI-SearchBot, ChatGPT-User, Meta-ExternalAgent |
+| 2 | **Published IP allowlist match** | JSON lists from crawler operators | Definitive when paired with #1 | openai.com/gptbot.json, openai.com/searchbot.json, openai.com/chatgpt-user.json, Anthropic published ranges, Perplexity ranges |
+| 3 | **JA4 TLS fingerprint match** | TLS ClientHello at edge | Strong (catches spoofed UAs) | Common LLM fetcher libraries: Python `requests`, `aiohttp`, `httpx`, Go `net/http`, Node `fetch`, cURL, Scrapy, Playwright, Puppeteer |
+| 4 | **ASN classification** | IP → ASN lookup | Supporting signal only (never decisive alone) | Datacenter/hosting ASNs (AWS, GCP, Azure, DigitalOcean, Hetzner, OVH), VPN/proxy ASNs, residential ASNs |
+| 5 | **H2 handshake presence** | Edge TLS/HTTP layer | Supporting signal (humans nearly always H2; many scrapers still H1) | All traffic |
+| 6 | **`/robots.txt` and `/license.xml` fetch correlation** | TS request logs | Supporting signal (honest bots fetch before crawling) | All traffic |
+
+### 4.2 Classification Categories
+
+```rust
+pub enum Classification {
+    /// Confirmed human or browser-class traffic.
+    /// Signals: browser JA4 + H2 + consistent browsing patterns
+    Human,
+
+    /// Confirmed AI crawler with honest identity.
+    /// Signals: UA match AND IP allowlist match (or JA4 match for known library)
+    HonestAiCrawler {
+        operator: String,      // "openai", "anthropic", "perplexity", etc.
+        bot_name: String,      // "gptbot", "claude-user", "perplexitybot", etc.
+        purpose: AiPurpose,    // training, search, in-conversation, index
+    },
+
+    /// Strong AI crawler suspicion without honest identity.
+    /// Signals: datacenter ASN + LLM-library JA4 + no H2 or irregular headers
+    StealthAiCrawler {
+        signals: Vec<Signal>,
+        confidence: Confidence, // High | Medium | Low
+    },
+
+    /// Cannot classify with enough confidence.
+    /// Default action depends on publisher mode (Permissive vs Strict).
+    Ambiguous {
+        signals: Vec<Signal>,
+    },
+}
+```
+
+### 4.3 Signal Refresh Cadence
+
+- **IP allowlists from crawler operators:** fetched by a control-plane job
+  from each operator's published JSON endpoint (e.g., `openai.com/gptbot.json`).
+  Bundled into TS releases. Publishers pick up new IP ranges when they update
+  to a newer TS version. Recommended publisher refresh cadence: weekly.
+- **UA patterns:** static, updated via TS release.
+- **JA4 fingerprint database:** static, updated via TS release
+  (community-maintained list).
+- **ASN database:** updated via Maxmind or equivalent on publisher's own
+  schedule.
+
+### 4.4 Example Log Entry for a Classified Request
+
+```json
+{
+  "request_id": "01HXYZ...",
+  "timestamp": "2026-04-22T14:32:11Z",
+  "path": "/article/some-slug",
+  "classification": "honest_ai_crawler",
+  "operator": "anthropic",
+  "bot_name": "claudebot",
+  "purpose": "ai_train",
+  "signals_matched": ["ua_honest", "ip_allowlist:anthropic"],
+  "action": "403_forbidden",
+  "action_reason": "license.toml prohibits ai-train for this route",
+  "license_terms_applied": "default",
+  "mode": "permissive"
+}
+```
+
+### 4.5 Stealth Classification Example
+
+A scraper running on AWS using Python `requests` with a Chrome user-agent
+gets classified as `StealthAiCrawler`:
+
+- Signal: `asn:aws` ✓
+- Signal: `ja4:python_requests` ✓
+- Signal: `ua_spoofed_chrome` (UA claims Chrome but JA4 says Python) ✓
+- Confidence: High
+- Action depends on mode: Permissive → allow through (but logged for publisher
+  review); Strict → 403.
+
+### 4.6 Detection Posture (Default vs. Override)
+
+**Default:** Permissive. Block only confirmed crawlers whose license terms
+prohibit access. Stealth crawlers and ambiguous traffic are allowed through
+but logged for publisher review.
+
+**Override:** Strict, configurable per-publisher or per-route. Stealth and
+ambiguous traffic get 403. Use for high-value routes (premium content, APIs).
+
+### 4.7 Transparency Model
+
+- **Transparent to the publisher:** full classification detail in logs and
+  debug endpoints (which signals matched, what the decision was, why).
+- **Opaque to the crawler:** crawlers receive standards-compliant RSL responses
+  (401/402/403 with `WWW-Authenticate: License` and `Link` header pointing to
+  `license.xml`). Crawlers are not told which signals flagged them — this
+  prevents adversarial training against TS detection.
+
+---
+
+## 5. Configuration
+
+### 5.1 Two-File Split: Public vs Private
+
+**`license.toml`** — PUBLIC RSL terms. Safe to commit to a public git repo.
+Everything here ends up in `/license.xml`.
+
+**`license.private.toml`** — PRIVATE enforcement rules. Never exposed via any
+endpoint. Contains per-crawler commercial overrides, enforcement mode
+configuration, and any NDA-bound IP allowlist extensions.
+
+Both files are compiled into the WASM binary at build time (same mechanism as
+existing TS configs). Any change requires a rebuild and redeploy via the
+standard TS deploy pipeline. Deploy time is fast (standard
+`fastly compute publish` flow — typically under a minute for the publish,
+plus CI build time).
+
+**Optional future enhancement (not POC):** runtime config loading from edge KV
+or Fastly Config Store so terms can be updated without a redeploy. Deferred
+because it introduces failure modes (KV availability, eventual consistency,
+auth) that add complexity.
+
+### 5.2 Public `license.toml` — Minimal Example
+
+```toml
+# license.toml — publisher's RSL terms, read by Trusted Server
+# Served as /license.xml (auto-generated), referenced in /robots.txt
+
+[publisher]
+name = "Example Publisher, Inc."
+contact = "licensing@example.com"
+contact_url = "https://example.com/licensing"
+copyright_holder = "Example Publisher, Inc."
+copyright_type = "organization"
+
+# Default terms for all content not matching a more specific route
+[default]
+# What uses are allowed (RSL usage vocabulary)
+permits = ["search", "ai-input"]
+# What uses are explicitly prohibited
+prohibits = ["ai-train"]
+# Default payment model for permitted uses
+payment = "attribution"
+```
+
+### 5.3 Public `license.toml` — Full Example With Route Overrides
+
+```toml
+[publisher]
+name = "Example Publisher, Inc."
+contact = "licensing@example.com"
+contact_url = "https://example.com/licensing"
+copyright_holder = "Example Publisher, Inc."
+copyright_type = "organization"
+
+# Default terms — homepage, category pages, public articles
+[default]
+permits = ["search", "ai-input"]
+prohibits = ["ai-train", "ai-index"]
+payment = "attribution"
+
+# Premium/paywalled content — strict, contact for licensing
+[routes."/premium/*"]
+permits = []
+prohibits = ["ai-all", "search"]
+payment = "subscription"
+amount = "10.00"
+currency = "USD"
+
+# News archive — crawl fees apply
+[routes."/archive/*"]
+permits = ["ai-input", "ai-index"]
+prohibits = ["ai-train"]
+payment = "crawl"
+amount = "0.005"
+currency = "USD"
+
+# API endpoints — no AI use at all
+[routes."/api/*"]
+permits = []
+prohibits = ["ai-all"]
+payment = "free"
+```
+
+### 5.4 Private `license.private.toml` — Example
+
+```toml
+# Enforcement mode per route (not published — operational decision)
+[enforcement]
+default_mode = "permissive"
+
+[[enforcement.routes]]
+pattern = "/premium/*"
+mode = "strict"
+
+[[enforcement.routes]]
+pattern = "/api/*"
+mode = "strict"
+
+# Per-crawler commercial overrides — commercial secrets
+[[crawler_overrides]]
+bot_name = "gptbot"
+action = "allow"
+reason_internal = "Direct license agreement - contract #2026-OAI-001"
+
+[[crawler_overrides]]
+bot_name = "perplexitybot"
+action = "deny"
+reason_internal = "Pending commercial agreement"
+
+[[crawler_overrides]]
+bot_name = "claudebot"
+action = "enforce_default"  # apply license.toml terms as-is
+
+# Per-operator IP allowlist extensions (e.g., publisher has a direct feed from
+# a crawler operator not in the default public list)
+[[ip_allowlist_extensions]]
+operator = "example_ai_partner"
+cidrs = ["203.0.113.0/24", "198.51.100.0/24"]
+note = "Partner under NDA — not publicly disclosed"
+```
+
+### 5.5 Key Config Design Points
+
+1. **`permits` / `prohibits` use RSL's usage vocabulary** — `search`, `ai-all`,
+   `ai-train`, `ai-input`, `ai-index`. Prohibition always wins when both apply
+   (per RSL spec).
+2. **`payment` types match RSL's payment vocabulary** — `purchase`,
+   `subscription`, `training`, `crawl`, `use`, `contribution`, `attribution`,
+   `free`.
+3. **Route patterns are RFC 9309-compliant** (same syntax as robots.txt) —
+   wildcards supported, more specific paths override less specific.
+4. **`mode` per route lives in the private config** — permissive vs strict is
+   an operational decision, not a published term.
+5. **`[crawler_overrides]` in private config** — per-bot exceptions for
+   publishers with direct commercial deals. `"allow"` bypasses enforcement
+   entirely (they've paid via a separate contract); `"deny"` blocks regardless
+   of terms; `"enforce_default"` applies the route's public terms (default
+   behavior).
+6. **No secrets in the public file** — `license.toml` can live in a public git
+   repo if the publisher wants.
+
+### 5.6 Trusted Server Settings
+
+```toml
+# trusted-server.toml
+[integrations.rsl]
+enabled = true
+public_config = "license.toml"
+private_config = "license.private.toml"
+```
+
+---
+
+## 6. HTTP Response Behavior
+
+### 6.1 Allowed Request (GPTBot on Route That Permits `ai-input`)
+
+```http
+GET /article/hello-world HTTP/2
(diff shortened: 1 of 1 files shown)


docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+| Existing capability | How RSL uses it |
+|---|---|
+| `IntegrationRegistration` builder | New hook types: `with_request_classifier`, `with_special_route_augmenter` |
+| JA4 signal from edge TLS | Input to `classifier::classify()` |
+| Bot gate (H2 + JA4) | Supporting signal for stealth detection |
+| `/robots.txt` handling | Integration augments existing `robots.txt` response |
+| `/_ts/debug/*` auth pattern | Debug endpoints reuse existing token auth |
+| Structured logging (`log-fastly`) | Classification events emitted as structured log lines |
+| Settings (`trusted-server.toml`) | RSL config block added to existing settings parser |
+
```
aram356, line 194: 🔧 **wrench** — "Existing capability" table is materially inaccurate.

Four of six items in this column do not exist in the codebase today (verified by searching `crates/`):

- **JA4 signal from edge TLS** — no JA4 code, no `client_hello` access, no TLS-fingerprint plumbing anywhere
- **Bot gate (H2 + JA4)** — no bot gate exists
- **`/robots.txt` handling** — no robots.txt handler in `crates/trusted-server-core/`
- **`/_ts/debug/*` auth pattern** — no such route family or token-auth pattern exists

A reader walks away believing the implementation reuses four existing systems. It actually builds them all from scratch — a materially different effort estimate.

**Fix:** split the table into two columns:

- **Existing capability** — `IntegrationRegistration` builder, Settings (`trusted-server.toml`), structured logging
- **New infrastructure required** — JA4 acquisition path, bot gate, `/robots.txt` handler, `/_ts/debug/*` framework

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+    {"path": "/archive/*", "requests": 2104},
+    {"path": "/premium/*", "requests": 897}
+  ]
+}
+```
+
+### 7.3 `GET /_ts/debug/rsl/recent`
+
+Last N classified requests, newest first. Backed by an in-process ring buffer
+(no KV writes on hot path). Default 1000 entries, configurable.
```
aram356, line 768: 🔧 **wrench** — In-process ring buffer assumes long-lived process state that Fastly Compute does not provide.

Fastly Compute WASM instances are short-lived per-request — there is no in-memory state shared across requests. As specified, `/_ts/debug/rsl/recent` and `/_ts/debug/rsl/summary` would only see the single classification of the request that hit the debug endpoint itself.

The spec promises both "no KV writes on the hot path" *and* "live counters / recent classifications" — these are mutually exclusive on Fastly Compute today.

**Fix — pick one:**

1. Pipe `/summary` through an external aggregator over the structured log stream (Fastly log shipping → S3/BigQuery/Datadog), and document that the debug endpoints are *not* live edge state.
2. Commit to KV/Config Store reads/writes on the hot path with the trade-offs §5.1 explicitly defers (availability, eventual consistency, auth, write QPS limits).

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+
+## 4. Bot Detection Signals
+
+### 4.1 Six Signals
+
+| # | Signal | Source | Strength | Coverage |
+|---|---|---|---|---|
+| 1 | **Honest User-Agent match** | HTTP `User-Agent` header | Definitive when paired with #2 | GPTBot, ClaudeBot, Claude-User, Claude-SearchBot, PerplexityBot, Perplexity-User, Google-Extended, CCBot, Bytespider, Amazonbot, Applebot-Extended, OAI-SearchBot, ChatGPT-User, Meta-ExternalAgent |
+| 2 | **Published IP allowlist match** | JSON lists from crawler operators | Definitive when paired with #1 | openai.com/gptbot.json, openai.com/searchbot.json, openai.com/chatgpt-user.json, Anthropic published ranges, Perplexity ranges |
+| 3 | **JA4 TLS fingerprint match** | TLS ClientHello at edge | Strong (catches spoofed UAs) | Common LLM fetcher libraries: Python `requests`, `aiohttp`, `httpx`, Go `net/http`, Node `fetch`, cURL, Scrapy, Playwright, Puppeteer |
```
aram356, line 271: ❓ **question** — How does the WASM instance actually obtain JA4?

JA4 is the only signal that catches spoofed-UA stealth crawlers; §4.5 ("Stealth Classification Example") and the entire stealth branch in the §6.7 response matrix hinge on it.

Fastly Compute does not expose ClientHello bytes or a precomputed JA4 to the WASM instance today. Without a concrete acquisition path the stealth-detection branch is unimplementable.

Please specify which mechanism is assumed:

- VCL pre-stage that hashes ClientHello and forwards as `Fastly-JA4` request header?
- A closed-beta / experimental Fastly API?
- Upstream computation in a different layer?
- Phase-2 deferred until the edge platform exposes JA4?

Whichever it is, one paragraph describing it would unblock the design.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+```http
+HTTP/2 200 OK
+Link: <https://example.com/license.xml>; rel="license"; type="application/rsl+xml"
+Content-Type: text/html
+...
+<article>...full content...</article>
+```
+
+TS adds the `Link` header on every response so honest crawlers can discover
+license terms on any request, not just by fetching `robots.txt` first.
```
aram356, line 542: ❓ **question** — `Link: rel="license"` on every response — including non-HTML responses?

§6.1 says "TS adds the `Link` header on every response." §8.5 says existing integrations "continue working unchanged."

In practice every ad-server response, RTB endpoint, integration proxy response (Permutive, Lockr, Datadome, Didomi), 204 beacon, and OPTIONS preflight will gain this header. Is that intentional?

- For top-level navigation HTML responses: yes, it's the goal.
- For JSON RTB bid responses or analytics 204s: it's noise that competes with `Link` headers used for HTTP/2 push hints, preconnect, etc.

Suggest scoping to `Content-Type: text/html` (or top-level navigation responses) and explicitly stating the scope in §6.1.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+│   └── private.rs       # license.private.toml parser → EnforcementRules
+├── xml_generator.rs     # LicenseTerms → RSL XML document
+├── fingerprints/
+│   ├── ua_patterns.rs   # static honest-UA matchers
+│   ├── ip_allowlists.rs # compiled-in crawler operator IP ranges
+│   └── ja4_db.rs        # JA4 fingerprint database for LLM fetchers
+├── enforcement.rs       # verdict + terms + mode → Action
+├── endpoints.rs         # /license.xml, /robots.txt augmentation, debug routes
+└── logging.rs           # structured log emission
+```
```
aram356, line 242: ♻️ **refactor** — Module path inconsistent with project layout.

Spec proposes `crates/trusted-server-core/src/rsl/`. Every other integration in TS lives under `crates/trusted-server-core/src/integrations/{datadome,permutive,lockr,didomi,testlight,…}` — verified in the filesystem and CLAUDE.md.

**Fix:** `crates/trusted-server-core/src/integrations/rsl/{mod.rs, classifier.rs, …}` for consistency with `IntegrationRegistration` discovery and the existing module structure.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+
+### 3.4 State Footprint at Edge
+
+- AI crawler IP/UA/JA4 allowlists: compiled into the WASM binary at build
+  time, ~tens of KB, refreshed when TS is rebuilt with updated allowlists.
+- `license.toml` and `license.private.toml`: compiled into the WASM binary at
+  build time (same mechanism as other TS configs).
+- Per-request classification: emitted as structured log lines, optionally
+  counted in a small in-memory ring buffer for debug endpoints (no KV writes
+  on the hot path).
```
aram356, line 181: ♻️ **refactor** — IP allowlist lookup data structure unspecified.

"Compiled into the WASM binary at build time, ~tens of KB, a few thousand CIDR ranges." A naive `Vec` scan over thousands of CIDRs on every request would dominate hot-path latency.

Specify the structure: radix/patricia trie (e.g. `ip_network_table`), `iprange::IpRange` with sorted-Vec + binary search, or an explicit perfect-hash variant. This is a non-trivial implementation choice with measurable performance consequences and belongs in the design, not later.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+      <permits type="usage">search</permits>
+      <permits type="usage">ai-input</permits>
+      <prohibits type="usage">ai-train</prohibits>
+      <prohibits type="usage">ai-index</prohibits>
+      <payment type="attribution"/>
+    </license>
+    <copyright type="organization" contactEmail="licensing@example.com">
+      Example Publisher, Inc.
+    </copyright>
+  </content>
```
aram356, line 664: ♻️ **refactor** — `contact_url` config field doesn't appear in the rendered XML example.

`license.toml` (§5.2) defines `contact_url`. RSL 1.0's `<copyright>` element supports both `contactEmail` and `contactUrl` attributes. The example here renders only `contactEmail="…"` and silently drops `contact_url`.

**Fix:** render both attributes:

```xml
<copyright type="organization"
           contactEmail="licensing@example.com"
           contactUrl="https://example.com/licensing">
  Example Publisher, Inc.
</copyright>
```

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+[[ip_allowlist_extensions]]
+operator = "example_ai_partner"
+cidrs = ["203.0.113.0/24", "198.51.100.0/24"]
+note = "Partner under NDA — not publicly disclosed"
+```
+
+### 5.5 Key Config Design Points
+
+1. **`permits` / `prohibits` use RSL's usage vocabulary** — `search`, `ai-all`,
+   `ai-train`, `ai-input`, `ai-index`. Prohibition always wins when both apply
```
aram356, line 495: ♻️ **refactor** — Usage vocabulary list incomplete.

RSL 1.0 defines: `all`, `ai-all`, `ai-train`, `ai-input`, `ai-index`, `search`. The spec lists "`search`, `ai-all`, `ai-train`, `ai-input`, `ai-index`" — missing `all`.

Add `all` for completeness; §5.3 examples already use `ai-all` correctly so this is a doc-only fix in the prose.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+
+### 8.1 Assumptions
+
+- Publisher already runs Trusted Server at their edge.
+- Edge platform status:
+  - **Live today:** Fastly Compute (WASM)
+  - **Planned:** Akamai EdgeWorkers, Cloudflare Workers
+  - The RSL integration is pure Rust with no Fastly-specific dependencies in
+    the classifier, resolver, or XML generator. Porting follows the same
+    pattern as the rest of TS.
```
aram356, line 882: 🤔 **thinking** — "No Fastly-specific dependencies in the classifier, resolver, or XML generator" overstates the current reality.

`crates/trusted-server-core/Cargo.toml:25` already has `fastly = { workspace = true }` as a non-optional dependency. There is an in-progress platform-abstraction effort (PR #581 added `PlatformHttpClient` / `PlatformBackend` traits, PR #609 wired `RuntimeServices`).

The spec should commit to building against those traits (preferred), or honestly note that core is not yet platform-neutral and the RSL module is contributing to the abstraction work, not benefiting from a finished one.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+  `license.xml` generation
+- No new heavy dependencies — the classifier is pure pattern-matching against
+  compiled-in data structures; no ML, no external services
+
+### 3.9 Binary Size Impact
+
+- IP allowlists compiled in: ~30-50 KB (a few thousand CIDR ranges from major
+  operators)
+- JA4 fingerprint database: ~5-10 KB (a few hundred common LLM fetcher
+  fingerprints)
```
aram356, line 256: 🤔 **thinking** — ASN database not in the binary-size budget.

§4.3 says "ASN database: updated via Maxmind or equivalent on publisher's own schedule." MaxMind GeoLite2-ASN binary is ~10 MB; the §3.9 budget is "30–50 KB IPs + 5–10 KB JA4 + ~negligible UA patterns" — total <100 KB.

Reconcile one of:

1. Delegate ASN lookup to a Fastly-side feature (state which one).
2. Load from KV/Config Store at request time (and account for latency).
3. Use a curated datacenter-ASN allowlist (much smaller — size it explicitly).

As written, ASN is hand-waved relative to the binary-size promise.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+ambiguous traffic get 403. Use for high-value routes (premium content, APIs).
+
+### 4.7 Transparency Model
+
+- **Transparent to the publisher:** full classification detail in logs and
+  debug endpoints (which signals matched, what the decision was, why).
+- **Opaque to the crawler:** crawlers receive standards-compliant RSL responses
+  (401/402/403 with `WWW-Authenticate: License` and `Link` header pointing to
+  `license.xml`). Crawlers are not told which signals flagged them — this
+  prevents adversarial training against TS detection.
```
aram356, line 366: 🤔 **thinking** — "401/402/403" mentioned here, but only 402/403 appear elsewhere.

§6.7 response matrix has no 401 row. RSL 1.0 lists 401 as a `MAY` response when a License is missing/required. Either describe when 401 fires (likely the OLP token path in Phase 2 — missing/invalid `Authorization: License <token>`) or drop the 401 mention here to keep the response set self-consistent.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+    },
+}
+```
+
+### 4.3 Signal Refresh Cadence
+
+- **IP allowlists from crawler operators:** fetched by a control-plane job
+  from each operator's published JSON endpoint (e.g., `openai.com/gptbot.json`).
+  Bundled into TS releases. Publishers pick up new IP ranges when they update
+  to a newer TS version. Recommended publisher refresh cadence: weekly.
```
aram356, line 312: 🤔 **thinking** — "Recommended publisher refresh cadence: weekly" couples crawler IP freshness to the TS release train.

If OpenAI updates `gptbot.json` between TS releases, publishers stay stale until the next bundled allowlist ships. Worth acknowledging the staleness window or designing a Config-Store fallback that refreshes from operator JSON without a redeploy (with bundled list as static fallback for KV unavailability).

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+    <license>
+      <prohibits type="usage">ai-all</prohibits>
+      <payment type="free"/>
+    </license>
+  </content>
+</rsl>
+```
+
+30-day cache — matches RSL's `max-age` default. Crawlers cache this and only
+re-fetch when the ETag changes.
```
aram356, line 697: 🤔 **thinking** — `max-age` unit confusion (days vs seconds).

This section emits `max-age="30"` (RSL: integer days) on the same response as `Cache-Control: public, max-age=2592000` (HTTP: seconds, = 30 days). Both correct, but a single sentence noting the unit difference between RSL `max-age` and HTTP `Cache-Control` `max-age` would prevent future bugs in implementers.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+        signals: Vec<Signal>,
+        confidence: Confidence, // High | Medium | Low
+    },
+
+    /// Cannot classify with enough confidence.
+    /// Default action depends on publisher mode (Permissive vs Strict).
+    Ambiguous {
+        signals: Vec<Signal>,
+    },
+}
```
aram356, line 304: 🌱 **seedling** — `purpose` likely belongs on bot identity, not request classification.

`Classification::HonestAiCrawler { purpose: AiPurpose }` — but a bot's declared purpose is a property of its UA (GPTBot = training, ChatGPT-User = in-conversation, OAI-SearchBot = search), not of the individual request. Consider a static `BotId → AiPurpose` mapping table; the classification then carries `bot_identity: BotId` and `purpose` is a derived lookup.

Not blocking — flagging now so the type design isn't locked in before implementation.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+- **IP allowlists from crawler operators:** fetched by a control-plane job
+  from each operator's published JSON endpoint (e.g., `openai.com/gptbot.json`).
+  Bundled into TS releases. Publishers pick up new IP ranges when they update
+  to a newer TS version. Recommended publisher refresh cadence: weekly.
+- **UA patterns:** static, updated via TS release.
+- **JA4 fingerprint database:** static, updated via TS release
+  (community-maintained list).
+- **ASN database:** updated via Maxmind or equivalent on publisher's own
+  schedule.
+
```
aram356, line 318: 📌 **out of scope** — Control-plane refresh job is referenced but not designed.

"Fetched by a control-plane job from each operator's published JSON endpoint" — this control plane is new infrastructure outside the WASM edge. Reasonable to defer, but should appear explicitly in §2.2 (Out of Scope) as a dependency for the "publishers always have fresh allowlists" promise. Otherwise the freshness story is implicitly "manual TS release cadence".

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+---
+
+## 4. Bot Detection Signals
+
+### 4.1 Six Signals
+
+| # | Signal | Source | Strength | Coverage |
+|---|---|---|---|---|
+| 1 | **Honest User-Agent match** | HTTP `User-Agent` header | Definitive when paired with #2 | GPTBot, ClaudeBot, Claude-User, Claude-SearchBot, PerplexityBot, Perplexity-User, Google-Extended, CCBot, Bytespider, Amazonbot, Applebot-Extended, OAI-SearchBot, ChatGPT-User, Meta-ExternalAgent |
+| 2 | **Published IP allowlist match** | JSON lists from crawler operators | Definitive when paired with #1 | openai.com/gptbot.json, openai.com/searchbot.json, openai.com/chatgpt-user.json, Anthropic published ranges, Perplexity ranges |
```
aram356, line 270: 📝 **note** — Verified live (200 OK) for the IP-list URLs cited:

- `https://openai.com/gptbot.json`
- `https://openai.com/searchbot.json`
- `https://openai.com/chatgpt-user.json`

No action — recording the verification.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
+# Trusted Server AI Crawler Licensing (RSL-compliant)
+
+*April 2026*
```
aram356, line 3: ⛏ **nitpick** — format-docs CI failure is a trivial prettier whitespace diff.

- Italic syntax: `*April 2026*` → `_April 2026_` (this line)
- Markdown table column padding (§3.5 and §4.1)

Fix in one command:

```bash
cd docs && npx prettier --write superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+
+1. **RSL-native, standards-compliant** — publishes `license.xml`, honors OLP
+   where applicable, interoperates with any compliant entitlement platform.
+2. **Edge-native, low-latency** — classification and enforcement at the CDN
+   layer, no proxy hop, no subdomain redirect.
+3. **Multi-signal detection including JA4** — TLS-layer fingerprinting catches
+   crawlers that spoof user-agent strings.
+4. **Publisher-owned config** — single `license.toml` file, version-controlled,
+   no lock-in to a vendor's dashboard.
+5. **Open source** — publishers can audit the enforcement behavior.
```
aram356, line 43: ⛏ **nitpick** — "Trusted Server" / "TS" used interchangeably mid-paragraph throughout §1. Pick one and stick with it for readability. CLAUDE.md doesn't enforce, but the project prose elsewhere is consistent.

aram356 (changes requested):
## Summary

Design spec PR — single 1055-line markdown file in `docs/superpowers/specs/`, no code changes. The proposed RSL-compliant AI crawler licensing layer is well-scoped and the public/private config split is sound, but the spec misrepresents the current state of TS infrastructure and contains two architectural assumptions that don't hold on Fastly Compute. Requesting changes on those before merge.

## Blocking

### 🔧 wrench

- **§3.5 "Existing capability" table is materially inaccurate** — JA4 signal, bot gate, `/robots.txt` handling, and `/_ts/debug/*` auth pattern are all listed as existing but do not exist in `crates/`. Reframe as new infrastructure required.
- **§7.2 / §7.3 in-process ring buffer assumes long-lived process state** — Fastly Compute WASM instances are short-lived per-request; the debug endpoints can't aggregate without KV/Config Store or external log-stream aggregation. Pick one and document the trade-off.

### ❓ question

- **§4.1 / §3.5 — How does the WASM instance obtain JA4?** Fastly Compute does not expose ClientHello bytes today. Without a concrete acquisition path the entire stealth-detection branch is unimplementable.
- **§6.1 / §8.5 — `Link: rel="license"` on every response, including ad/RTB/integration responses?** §8.5 says integrations are unaffected, but every response gaining a header is a change. Suggest scoping to HTML responses.

## Non-blocking

### ♻️ refactor

- **§3.7 module path** — should be `crates/trusted-server-core/src/integrations/rsl/`, matching every other integration in the project.
- **§3.4 IP allowlist lookup structure unspecified** — naive Vec scan over thousands of CIDRs would dominate hot-path latency; specify a radix/trie structure.
- **§6.6 rendered XML drops `contact_url`** — `license.toml` defines it but the example only renders `contactEmail`.
- **§5.5 usage vocabulary missing `all`** — RSL 1.0 defines `all`, `ai-all`, `ai-train`, `ai-input`, `ai-index`, `search`.

### 🤔 thinking

- **§3.8 / §8.1 "no Fastly-specific dependencies in core"** overstates current reality — `crates/trusted-server-core/Cargo.toml` already has `fastly` as a non-optional dep; PR #581/#609 are the in-progress abstraction work.
- **§4.1 ASN database not in the §3.9 binary-size budget** — MaxMind GeoLite2-ASN is ~10 MB; reconcile with the <100 KB budget.
- **§4.7 mentions 401 but §6.7 matrix doesn't** — drop or describe when 401 fires.
- **§4.3 IP-allowlist refresh cadence couples to TS release train** — staleness window or KV-based refresh path worth acknowledging.
- **§6.6 RSL `max-age` is in days, HTTP `Cache-Control: max-age` in seconds** — note the unit difference.

### 🌱 seedling

- **§4.2 `purpose` likely belongs on bot identity, not request classification.**

### 📌 out of scope

- **§4.3 control-plane refresh job referenced but not designed** — should appear in §2.2 if deferred.

### 📝 note

- IP-list URLs (`openai.com/{gptbot,searchbot,chatgpt-user}.json`) verified live (200 OK).

### ⛏ nitpick

- format-docs CI failure is a trivial prettier whitespace diff (asterisk italics → underscore italics, table column padding); fix with `cd docs && npx prettier --write superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md`.
- "Trusted Server" / "TS" used interchangeably mid-paragraph; pick one.

## CI Status

- format-docs: **FAIL** (one-command prettier fix)
- cargo fmt: PASS
- cargo clippy: PASS
- cargo test: PASS
- vitest: PASS
- CodeQL: PASS

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+| 2 | **Published IP allowlist match** | JSON lists from crawler operators | Definitive when paired with #1 | openai.com/gptbot.json, openai.com/searchbot.json, openai.com/chatgpt-user.json, Anthropic published ranges, Perplexity ranges |
+| 3 | **JA4 TLS fingerprint match** | TLS ClientHello at edge | Strong (catches spoofed UAs) | Common LLM fetcher libraries: Python `requests`, `aiohttp`, `httpx`, Go `net/http`, Node `fetch`, cURL, Scrapy, Playwright, Puppeteer |
+| 4 | **ASN classification** | IP → ASN lookup | Supporting signal only (never decisive alone) | Datacenter/hosting ASNs (AWS, GCP, Azure, DigitalOcean, Hetzner, OVH), VPN/proxy ASNs, residential ASNs |
+| 5 | **H2 handshake presence** | Edge TLS/HTTP layer | Supporting signal (humans nearly always H2; many scrapers still H1) | All traffic |
+| 6 | **`/robots.txt` and `/license.xml` fetch correlation** | TS request logs | Supporting signal (honest bots fetch before crawling) | All traffic |
```
aram356, line 274: 🔧 **wrench** — Signal #6 (`/robots.txt` and `/license.xml` fetch correlation) needs cross-request state Fastly Compute does not provide.

This is the same root issue as the in-process ring buffer flagged earlier, but here it is a *classification signal* on the hot path, not just observability. Tracking "did this IP fetch `/robots.txt` recently?" requires cross-request lookups against Fastly KV / Edge Dictionary / Config Store, with cost and latency consequences that aren't acknowledged. As written, the signal is non-functional in the proposed architecture.

**Fix:** either drop signal #6 from the table, or re-spec it explicitly as a KV-backed correlation store with the request-budget impact called out in §3.4 and §3.9.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+Content-Type: text/plain
+
+License: https://example.com/license.xml
+
+User-agent: *
+Disallow: /admin/
+Sitemap: https://example.com/sitemap.xml
+```
+
+TS preserves the publisher's existing `robots.txt` content and prepends the
```
aram356, line 635: 🔧 **wrench** — `/robots.txt` augmentation needs origin-fetch + cache infrastructure that does not exist today.

"TS preserves the publisher's existing `robots.txt` content and prepends the `License:` directive" implies an origin subrequest, body mutation, and an edge cache with an invalidation strategy. The only synthetic-body endpoint pattern in TS today is `/static/tsjs=` ([publisher.rs:114-150](crates/trusted-server-core/src/publisher.rs#L114-L150)), which generates output entirely from in-binary modules — no origin fetch. §3.4 doesn't list `robots.txt` as compiled-in state, and §3.7 has no fetcher module.

**Fix:** specify the mechanism explicitly — origin subrequest path, cache key + TTL, surrogate-key purge on `license.toml` change, and behavior when origin returns 404 / 5xx (the §6.5 "TS generates a minimal one" case). Or constrain the design to a fully-generated `robots.txt` and acknowledge that publisher origin-side `robots.txt` content is dropped.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+
+### 5.5 Key Config Design Points
+
+1. **`permits` / `prohibits` use RSL's usage vocabulary** — `search`, `ai-all`,
+   `ai-train`, `ai-input`, `ai-index`. Prohibition always wins when both apply
+   (per RSL spec).
+2. **`payment` types match RSL's payment vocabulary** — `purchase`,
+   `subscription`, `training`, `crawl`, `use`, `contribution`, `attribution`,
+   `free`.
+3. **Route patterns are RFC 9309-compliant** (same syntax as robots.txt) —
```
aram356, line 500: 🔧 **wrench** — `License:` directive is not RFC 9309.

This bullet says "Route patterns are RFC 9309-compliant (same syntax as robots.txt)". RFC 9309 defines `User-agent`, `Disallow`, `Allow`, `Sitemap` — there is no `License:` directive. The `License:` line in §6.5 is an RSL 1.0 *extension* to the `robots.txt` format, not part of RFC 9309. Conflating the two will mislead implementers and reviewers about standards posture.

**Fix:** clarify that the route-pattern *syntax* (wildcards, longest-match) is RFC 9309-compliant, while the `License:` line is an RSL 1.0 extension to robots.txt — and link the RSL spec section.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+already declares future-compatibility via the `server` attribute when enabled.
+
+---
+
+## 10. Success Criteria for POC
+
+Set before running against real traffic:
+
+1. **Classification accuracy:** 100% of honest AI crawlers (OpenAI, Anthropic,
+   Perplexity, Google-Extended, CCBot) correctly identified by UA + IP
```
aram356, line 1009: 🔧 **wrench** — Success criterion #1 cannot hold given the §4.3 IP-allowlist refresh model.

"100% of honest AI crawlers correctly identified by UA + IP allowlist signals" is a hard claim, but §4.3 says IP allowlists are bundled into TS releases ("Recommended publisher refresh cadence: weekly"). Between an operator publishing new IPs and the publisher rolling a new TS build, requests from those IPs match UA but miss the allowlist — falling to Ambiguous or StealthAiCrawler. The criterion as stated is unmeetable for any week-long observation window.

**Fix:** either qualify the criterion ("…against the bundled allowlist current at deploy time, measured within X days of release") or commit to runtime-loading the IP allowlist from a Fastly Config Store / KV refreshed by the control-plane job mentioned in §4.3.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+```
+
+### 4.5 Stealth Classification Example
+
+A scraper running on AWS using Python `requests` with a Chrome user-agent
+gets classified as `StealthAiCrawler`:
+
+- Signal: `asn:aws` ✓
+- Signal: `ja4:python_requests` ✓
+- Signal: `ua_spoofed_chrome` (UA claims Chrome but JA4 says Python) ✓
```
aram356, line 345: ❓ **question** — How is `ua_spoofed_chrome` actually computed?

This stealth example detects "UA claims Chrome but JA4 says Python" — that requires both an LLM-library JA4 set *and* a known-good browser/Chrome JA4 reference set to detect the mismatch. §3.9 only budgets "~5–10 KB JA4 (a few hundred LLM fetcher fingerprints)" — there is no Chrome reference set in scope. Chrome JA4s additionally vary by version × OS × GREASE, so the reference set is non-trivial.

Without an answer here, the entire stealth branch (§4.5, §6.3) — which is the design's differentiator over plain UA matching — is non-functional as specified.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+- Confidence: High
+- Action depends on mode: Permissive → allow through (but logged for publisher
+  review); Strict → 403.
```
aram356, line 348: ❓ **question** — JA4 false-positive surface for legitimate non-AI clients.

Python `requests`, cURL, `httpx`, Go `net/http`, Node `fetch` are also used by CI scripts, monitoring probes, internal services, third-party API clients, embedded SDKs. A Strict-mode rule keyed on those JA4s would block far more than AI scrapers — including, plausibly, the publisher's own infrastructure.

How does the design distinguish legitimate non-browser clients from stealth crawlers? Without an answer, Strict mode is a liability and §10 success criterion #2 ("zero human-traffic impact") under-specifies the risk: the impact isn't on human traffic, it's on legitimate-machine traffic the spec doesn't model.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+pattern = "/premium/*"
+mode = "strict"
+
+[[enforcement.routes]]
+pattern = "/api/*"
+mode = "strict"
+
+# Per-crawler commercial overrides — commercial secrets
+[[crawler_overrides]]
+bot_name = "gptbot"
```
aram356, line 471: ❓ **question** — Bot-name normalization is unspecified.

`[[crawler_overrides]] bot_name = "gptbot"` (lowercase) vs UA token `GPTBot/1.1` in §6.1. The case-folding rule for the override lookup is not stated. Also unclear whether `bot_name` matches a substring, the full UA product token, or a canonical operator label maintained inside TS.

This is a typical source of "my override doesn't fire" bugs.

**Suggested:** lowercase substring match against the UA's product token, with the canonical bot names enumerated in this section. Document explicitly.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+        bot_name: String,      // "gptbot", "claude-user", "perplexitybot", etc.
+        purpose: AiPurpose,    // training, search, in-conversation, index
+    },
+
+    /// Strong AI crawler suspicion without honest identity.
+    /// Signals: datacenter ASN + LLM-library JA4 + no H2 or irregular headers
+    StealthAiCrawler {
+        signals: Vec<Signal>,
+        confidence: Confidence, // High | Medium | Low
+    },
```
aram356, line 297: 🤔 **thinking** — `Confidence: High | Medium | Low` semantics undefined.

The enum is declared but no thresholds tie signal combinations to confidence levels, and Strict-mode behavior at each level isn't specified. As written, `Confidence` is a placeholder.

At minimum, document the rule: e.g., "two definitive signals → High; one definitive + supporting → Medium; supporting only → Low" — and state which levels trigger 403 in Strict mode vs 402 vs allow-with-log.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+### 6.6 `/license.xml` Response (Public RSL Terms)
+
+```http
+GET /license.xml HTTP/2
+```
+
+```http
+HTTP/2 200 OK
+Content-Type: application/rsl+xml
+Cache-Control: public, max-age=2592000
```
aram356, line 648: 🤔 **thinking** — 30-day `Cache-Control: public, max-age=2592000` may strand publishers on stale terms after edits.

When `license.toml` changes and is redeployed, downstream caches (Fastly edge + crawler caches) hold the prior XML for up to 30 days. ETag-aware revalidation only helps clients that revalidate; the spec says crawlers "cache this and only re-fetch when the ETag changes," but they have no signal to revalidate during the 30-day window.

§11 q3 raises this for the RSL `max-age` attribute but doesn't connect it to `Cache-Control`. Either describe a purge-on-deploy mechanism (Fastly surrogate keys keyed off `license.toml` source hash) or recommend a configurable, shorter `Cache-Control` value.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+30-day cache — matches RSL's `max-age` default. Crawlers cache this and only
+re-fetch when the ETag changes.
+
+### 6.7 Full Response Matrix
+
+| Classification | Route permits? | Action | Status | Body |
+|---|---|---|---|---|
+| Human | (n/a) | Allow | 200 | Normal content + `Link` header |
+| Honest AI crawler | Yes | Allow | 200 | Normal content + `Link` header |
+| Honest AI crawler | No | Block (polite) | 402 | RSL fragment with terms |
```
aram356, line 705: 🤔 **thinking** — The actual decision logic — `bot_name → declared purpose → permits/prohibits` — is implicit.

The matrix abstracts to "Route permits? Yes/No," but the real question is: given GPTBot's declared purpose is `ai-train` and the route's `prohibits = ["ai-train", "ai-index"]`, what's the action? That mapping (declared bot purpose × usage vocabulary × `permits`/`prohibits` precedence) is the core of this design and is never written down.

Adjacent to but distinct from the prior 🌱 about `purpose` placement: this comment is about the missing decision-table itself. Recommend a worked-example table: rows = `(bot_name, purpose, route, permits, prohibits)` → columns = `(action, status, reason)`.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+  `Authorization: License <token>` on crawl requests
+- TS at the edge validates tokens via local HMAC check (fast path) or
+  optional `/introspect` callback (stronger security, higher latency)
+
+**Split architecture:**
+
+- **Hot path (WASM at edge):** token validation only. HMAC check against a
+  shared signing key. Sub-millisecond. No KV writes.
+- **Cold path (separate service):** token issuance, billing, dashboards, key
+  management for EMS-encrypted content. Writes to persistent storage.
```
aram356, line 996: 🤔 **thinking** — Phase-2 EMS contradicts the §3.5 "no changes required to" promise.

EMS-encrypted content needs the `/key` endpoint *and* content-side encryption — that touches the HTML processor and cache layer, both of which §3.5 line 195 explicitly says "No changes required to." Out-of-scope for POC, but worth acknowledging that the Phase-2 commitment isn't compatible with the §3.5 guarantee, so future scope discussions don't catch it as a surprise.

**Fix:** add a sentence to §3.5 noting the no-change guarantee applies to Phase 1 only, with EMS in Phase 2 explicitly flagged as touching those layers.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+   of terms; `"enforce_default"` applies the route's public terms (default
+   behavior).
+6. **No secrets in the public file** — `license.toml` can live in a public git
+   repo if the publisher wants.
+
+### 5.6 Trusted Server Settings
+
+```toml
+# trusted-server.toml
+[integrations.rsl]
```
aram356, line 516: 🤔 **thinking** — Two-config-file split deviates from the `trusted-server.toml`-everything pattern, no rationale given.

Other integrations declare config inline in `trusted-server.toml` (verified at [settings.rs](crates/trusted-server-core/src/settings.rs)). RSL introduces `license.toml` + `license.private.toml` as separate files. The likely reason — publishers want to git-version `license.toml` publicly while keeping internal settings private — is plausible, but the spec doesn't say so.

**Fix:** add a sentence to §5.1 explaining why RSL is the exception (public version-controlled terms vs. operational settings), or fold the contents under `[integrations.rsl.public]` / `[integrations.rsl.private]` blocks in `trusted-server.toml` for consistency.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+- Signal: `ua_spoofed_chrome` (UA claims Chrome but JA4 says Python) ✓
+- Confidence: High
+- Action depends on mode: Permissive → allow through (but logged for publisher
+  review); Strict → 403.
+
+### 4.6 Detection Posture (Default vs. Override)
+
+**Default:** Permissive. Block only confirmed crawlers whose license terms
+prohibit access. Stealth crawlers and ambiguous traffic are allowed through
+but logged for publisher review.
```
aram356, line 354: 🤔 **thinking** — Default Permissive may hide stealth-detection mis-fires from publishers.

Most publishers will enable this integration *to enforce*. Defaulting to Permissive means false-negatives (stealth crawlers misclassified as Ambiguous, or honest crawlers blocked by a logic bug) are invisible until the publisher reads structured logs.

Worth justifying the rollout posture explicitly (safety-first staging, then opt-in to Strict per route) or recommending Strict-on-premium-routes as a default for new deployments. As written, the publisher gets minimal protection until they manually flip to Strict.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+- Implements RSL OLP endpoints: `/token`, `/introspect`, `/key`
+- Entitlement platform obtains tokens programmatically, presents
+  `Authorization: License <token>` on crawl requests
+- TS at the edge validates tokens via local HMAC check (fast path) or
+  optional `/introspect` callback (stronger security, higher latency)
+
+**Split architecture:**
+
+- **Hot path (WASM at edge):** token validation only. HMAC check against a
+  shared signing key. Sub-millisecond. No KV writes.
```
aram356, line 994: 📌 **out of scope** — Phase-2 OLP key distribution / rotation unspecified.

HMAC against a shared signing key works only if (a) the WASM has the key, (b) the publisher can rotate it, and (c) revocation is possible after compromise. None of these are in scope today, but they should appear in §11 open questions or §2.2 explicitly so the dependency isn't lost between phases.

Likely answer for Fastly: load via `PlatformSecretStore` at request time. State this — and acknowledge that key rotation while tokens are in flight needs a grace window that the design doesn't yet specify.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+<rsl xmlns="https://rslstandard.org/rsl">
+  <content url="/premium/*">
+    <license>
+      <prohibits type="usage">ai-all</prohibits>
+      <payment type="subscription">
+        <amount currency="USD">10.00</amount>
+      </payment>
+    </license>
+  </content>
+</rsl>
```
aram356, line 568: ⛏ **nitpick** — 402 RSL fragment omits `<copyright>`.

The 402 body inlines `<content url="/premium/*">` with `<license>...<payment>` but no `<copyright>` element. The full `/license.xml` in §6.6 has it. RSL 1.0 attribution tracking expects `<copyright>` so consumers know the rights-holder for every fragment served.

**Fix:** keep at minimum a `<copyright type="organization" contactEmail="..." contactUrl="...">Example Publisher, Inc.</copyright>` element in 402 bodies.

docs/superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
```diff
@@ -0,0 +1,1055 @@
...
+# a crawler operator not in the default public list)
+[[ip_allowlist_extensions]]
+operator = "example_ai_partner"
+cidrs = ["203.0.113.0/24", "198.51.100.0/24"]
+note = "Partner under NDA — not publicly disclosed"
+```
+
+### 5.5 Key Config Design Points
+
+1. **`permits` / `prohibits` use RSL's usage vocabulary** — `search`, `ai-all`,
```
aram356, line 494: ⛏ **nitpick** — §5.3 example doesn't actually exercise "prohibition wins."

`permits = ["search", "ai-input"]` and `prohibits = ["ai-train", "ai-index"]` are disjoint sets, so the precedence rule stated here is never tested by the example.

**Fix:** add an example with overlapping vocab — e.g., `permits = ["ai-all"]` paired with `prohibits = ["ai-train"]` — to demonstrate that prohibition narrows the broader permission. Otherwise readers may not realize the precedence is load-bearing.

aram356 (changes requested):
## Summary

Re-review against unchanged head `8d081287` (no commits since the [2026-04-30 review](https://github.com/IABTechLab/trusted-server/pull/650#pullrequestreview-4203456118)). The 17 prior findings still stand and remain blocking — re-verified against the current `main` (post PR #581 / #609 / #610). This review adds 17 new findings surfaced by a deeper re-read; verdict remains `REQUEST_CHANGES`.

The new findings cluster around three themes:

1. **Hot-path infrastructure that doesn't exist** — Signal #6 fetch-correlation needs cross-request state, and `/robots.txt` augmentation needs an origin-fetch + cache pattern that TS doesn't have today (only synthetic body path is `/static/tsjs=`). Both are load-bearing and unaddressed.

2. **Stealth detection has gaps the spec doesn't acknowledge** — §4.5's `ua_spoofed_chrome` requires a Chrome JA4 reference set the design doesn't budget for, and the JA4 set as scoped will false-positive against legitimate non-AI clients (CI, monitoring, internal scripts). Combined with the unanswered prior question about how WASM obtains JA4 in the first place, the entire stealth branch is non-functional as specified.

3. **Decision logic is implicit and a success criterion is unmeetable** — The `bot_name → declared purpose → permits/prohibits precedence` table that drives every enforcement decision is never written down. §10 criterion #1 (\"100% identification\") is unmeetable given §4.3's release-bundled IP-allowlist refresh model.

## Blocking

### 🔧 wrench

- **Signal #6 needs cross-request state Fastly Compute lacks** — line 274
- **`/robots.txt` augmentation has no origin-fetch + cache mechanism** — line 635
- **`License:` directive is an RSL extension, not RFC 9309** — line 500
- **§10 success criterion #1 unmeetable given §4.3 refresh model** — line 1009

### ❓ question

- **Stealth detection requires a Chrome JA4 reference set not in scope** — line 345
- **JA4 false-positive surface for legitimate non-AI clients** — line 348
- **Bot-name normalization (case folding, match style) unspecified** — line 471

## Non-blocking

### 🤔 thinking

- **`Confidence` enum semantics undefined** — line 297
- **30-day `Cache-Control` may strand publishers on stale terms** — line 648
- **`bot_name → purpose → permits/prohibits` decision table missing** — line 705
- **Phase-2 EMS contradicts §3.5 \"no changes\" promise** — line 996
- **Two-config-file split deviates from `trusted-server.toml`, no rationale** — line 516
- **Default Permissive can hide stealth-detection mis-fires** — line 354

### 🌱 seedling

- **No test plan or testing strategy** — cross-cutting. 1055 lines of design with no §test-strategy. How are classifiers unit-tested without JA4 in the test runtime? How are integration tests structured against simulated Fastly geo / TLS metadata? Existing integrations test inline (e.g. `crates/trusted-server-core/src/integrations/permutive.rs:643+`) but there is no harness for TLS-fingerprint or ASN signals. A short §test-strategy (unit + integration + a sample-crawler end-to-end check) would close this.

### 📌 out of scope

- **Phase-2 OLP key distribution / rotation unspecified** — line 994

### ⛏ nitpick

- **402 RSL fragment omits `<copyright>`** — line 568
- **§5.3 example doesn't exercise the \"prohibition wins\" rule** — line 494

## CI Status (re-verified locally)

- cargo fmt: PASS
- cargo clippy: PASS
- cargo test: PASS
- vitest: PASS
- **format-docs: FAIL** — still red 14 days after the prior review flagged it. One-line fix:

  ```bash
  cd docs && npx prettier --write superpowers/specs/2026-04-22-rsl-ai-crawler-licensing-design.md
  ```

PR #653, opened 2026-04-23
Multi-backend routing

ChristianPavilonis:
## Summary
- split the multi-backend routing work into a standalone PR
- add backend routing config/build wiring and request-time origin selection
- add routing integration tests, harness support, and user-facing docs

## What changed
- add `BackendRouter` and backend routing settings/types
- merge `backends.toml` into embedded config at build time
- route publisher requests by host/path before proxying to origin
- add routing integration tests and test-backends support
- document multi-backend routing in the docs site

## Verification
- `cargo test --workspace`

## Follow-up
- DataDome server-side validation changes will land in a stacked follow-up PR.


crates/trusted-server-adapter-fastly/backends.toml
```diff
@@ -0,0 +1,176 @@
...
+#
+# This file is merged into the embedded binary config at build time by
+# crates/trusted-server-core/build.rs. It is separate from trusted-server.toml to keep
+# customer-specific configuration out of the shared application template.
+
+[[backends]]
+id = "raven"
+origin_url = "https://raven-public.prod.saymedia.com"
+certificate_check = true
+domains = [
```
aram356: 🔧  Please remove these domain and origin

aram356 (changes requested):

crates/trusted-server-adapter-fastly/backends.toml
ChristianPavilonis: Removed the checked-in `backends.toml` entirely. `build.rs` already treats it as optional, so shared builds continue to work without the file. I also updated the docs/comments to make that optional behavior explicit.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -364,17 +365,57 @@
...
+        .map_err(|e| {
+            log::error!("Failed to build backend router: {:?}", e);
+            e
+        })
+        .ok()
+    };
+
+    let (origin_url, certificate_check) = if let Some(ref router) = router {
+        let (url, cert_check) = router.select_origin(request_host, request_path);
+        log::info!(
```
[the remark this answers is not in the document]
ChristianPavilonis: Removed the routing log that emitted raw host/path/origin values and narrowed the remaining debug log to the dynamic backend name only. This keeps the routing behavior unchanged while avoiding cleartext request-derived logging.

aram356:
Closing since no longer needed

PR #660, opened 2026-04-24
Clarify cookie header tests and remove duplicate callback serde test

### LINKED ISSUE #454: Add invalid cookie header tests for get_cookie_value

### LINKED ISSUE #456: Consolidate duplicate callback serde rename tests

ChristianPavilonis:
## Summary

- add a focused unit test for the invalid UTF-8 `Cookie` header path in `handle_request_cookies`
- rename the existing malformed-cookie test so it matches the behavior it actually covers
- remove a redundant callback serde rename test while preserving direct and broader deserialization coverage

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/cookies.rs` | Renamed the malformed cookie-string test and added a new invalid UTF-8 cookie-header test using `HeaderValue::from_bytes(...)` to exercise the real `InvalidHeaderValue` branch. |
| `crates/trusted-server-core/src/models.rs` | Removed the duplicate callback serde rename test, keeping the existing focused direct test and broader callback deserialization coverage. |

## Closes

Closes #454
Closes #456

## Test plan

- [x] Other: Rust-only test cleanup; no production behavior changes expected

## Checklist

- [x] New code has tests


crates/trusted-server-core/src/cookies.rs
```diff
@@ -347,6 +350,26 @@ mod tests {
...
+            handle_request_cookies(&req).expect_err("should reject invalid UTF-8 cookie header");
+
+        assert!(
+            matches!(
+                err.current_context(),
+                TrustedServerError::InvalidHeaderValue { message }
+                    if message.contains("invalid UTF-8")
+            ),
+            "should return invalid header value error for non-UTF-8 cookie header"
+        );
```
aram356, line 371: 🔧 **wrench** — Brittle assertion: the `if message.contains("invalid UTF-8")` guard locks this test to the literal string set in `cookies.rs:108` (`"Cookie header contains invalid UTF-8"`). A reasonable rewording of that message (e.g. `"Cookie header is not valid UTF-8"`) would silently break this test even though behavior is unchanged. Production code branches on the variant, not the message — the variant check alone is sufficient.

**Fix:**
```rust
assert!(
    matches!(
        err.current_context(),
        TrustedServerError::InvalidHeaderValue { .. }
    ),
    "should return InvalidHeaderValue for non-UTF-8 cookie header"
);
```

crates/trusted-server-core/src/cookies.rs
```diff
@@ -347,6 +350,26 @@ mod tests {
         assert!(jar.iter().count() == 0);
     }
 
+    #[test]
+    fn test_handle_request_cookies_invalid_utf8_cookie_header() {
+        let invalid_cookie_value = HeaderValue::from_bytes(b"ts-ec=valid-prefix\xF0\x90\x80")
```
aram356: ⛏ **nitpick** — `\xF0\x90\x80` is a *truncated* 4-byte UTF-8 sequence: `\xF0` introduces a 4-byte code point but only two continuation bytes follow, so `to_str()` rejects it. A one-line comment above the literal would save future readers a trip to the Unicode tables.

```rust
// Truncated 4-byte UTF-8 sequence: `\xF0` starts a 4-byte code point but
// only two continuation bytes follow, so `to_str()` rejects it.
let invalid_cookie_value = HeaderValue::from_bytes(b"ts-ec=valid-prefix\xF0\x90\x80")
```

aram356: ⛏ **nitpick** — The `ts-ec=valid-prefix` portion adds no coverage. `HeaderValue::to_str()` validates the entire byte slice; there is no partial parsing to test. A bare `b"\xF0\x90\x80"` would exercise the same code path more clearly. Harmless either way.

aram356 (changes requested):
## Summary

Test-only change closing real coverage gaps; CI is green and production behavior is unchanged. One blocking concern around a brittle assertion in the new UTF-8 test, plus two nitpicks.

## Blocking

### 🔧 wrench

- **Brittle assertion on error message text** — `cookies.rs:370` couples the test to the literal message string set in `cookies.rs:108`. Asserting on the variant alone (`InvalidHeaderValue { .. }`) is sufficient and survives reasonable message rewording.

## Non-blocking

### 📝 note

- **Issue #454 wording is stale** — the "Done when" says invalid UTF-8 should return `None`, but `handle_request_cookies` actually returns `Err(InvalidHeaderValue)` (mapped to `400 Bad Request` in `error.rs:111`). The PR test correctly exercises the real API; worth mentioning when closing #454 so the discrepancy is documented.

### ⛏ nitpick

- **Explain why the bytes are invalid UTF-8** — `\xF0\x90\x80` is a truncated 4-byte sequence; a one-line comment above the literal helps future readers.
- **`ts-ec=valid-prefix` adds no coverage** — `to_str()` validates the entire byte slice, so a bare `b"\xF0\x90\x80"` would exercise the same path more clearly.

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS (cookies::tests 25/25, models::tests 10/10 verified locally)
- js tests: PASS

crates/trusted-server-core/src/cookies.rs
ChristianPavilonis: Fixed in 09b76948. The test now matches only on `TrustedServerError::InvalidHeaderValue { .. }` and no longer depends on the literal error message text.

crates/trusted-server-core/src/cookies.rs
ChristianPavilonis: Fixed in 09b76948. Added a short comment explaining that `\xF0\x90\x80` is a truncated 4-byte UTF-8 sequence rejected by `to_str()`.

ChristianPavilonis: Fixed in 09b76948. Simplified the fixture to the bare invalid byte sequence `b"\xF0\x90\x80"` so the test focuses on whole-header UTF-8 validation.

ChristianPavilonis:
Addressed the requested changes in 09b76948:

- Removed the brittle message-text assertion and now match only `TrustedServerError::InvalidHeaderValue { .. }`.
- Simplified the invalid UTF-8 fixture to `b"\xF0\x90\x80"`.
- Added a comment explaining that the byte sequence is a truncated 4-byte UTF-8 sequence.

Verification:
- `cargo test -p trusted-server-core test_handle_request_cookies_invalid_utf8_cookie_header`
- `cargo test --workspace`

Also noting for Issue #454: the original wording saying invalid UTF-8 should return `None` is stale. Current behavior rejects invalid UTF-8 cookie headers with `InvalidHeaderValue`, which maps to HTTP 400, and this PR keeps/tests that behavior.

aram356 (approved):
## Summary

Test-only change closing real coverage gaps; all prior review feedback addressed.

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS (cookies::tests 25/25, models::tests 10/10 verified locally)
- js tests: PASS

PR #661, opened 2026-04-24
Refactor overlapping test coverage for streaming and creative rewrites

### LINKED ISSUE #457: Refactor UTF-8 boundary tests in streaming_replacer
## Description

`crates/common/src/streaming_replacer.rs` contains several UTF-8 boundary tests with overlapping intent and setup, which makes the suite harder to maintain.

## Done when

- [ ] Replace overlapping UTF-8 boundary tests with a table-driven or otherwise consolidated structure.
- [ ] Preserve coverage for multibyte and emoji boundary handling.
- [ ] Keep the refactor limited to test organization, not runtime behavior changes.

## Affected area

## Additional context

Source: `test_gap.md`, section `UTF-8 Boundary Tests`.
Parent issue: #396

### LINKED ISSUE #458: Consolidate overlapping srcset attribute tests in creative module
## Description

`crates/common/src/creative.rs` has multiple `srcset` tests with overlapping coverage. Consolidating them would preserve behavior while reducing maintenance overhead.

## Done when

- [ ] Review the current `srcset` tests for overlapping cases.
- [ ] Merge redundant cases into a smaller set or parameterized structure.
- [ ] Preserve coverage for spacing, relative paths, and descriptor handling.

## Affected area

## Additional context

Source: `test_gap.md`, section `Srcset Attribute Tests`.
Parent issue: #396

ChristianPavilonis:
## Summary

- Refactor overlapping UTF-8 boundary tests in `streaming_replacer` into shared helpers and consolidated coverage-focused cases.
- Consolidate repetitive `srcset` and `imagesrcset` creative rewrite tests into shared helpers plus table-driven case matrices.
- Keep runtime behavior unchanged while reducing duplicated test setup and preserving multibyte, emoji, spacing, relative path, and descriptor coverage.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/streaming_replacer.rs` | Added reusable chunk-processing test helpers and replaced overlapping UTF-8 boundary tests with consolidated case-based coverage, including explicit multibyte and emoji boundary splits. |
| `crates/trusted-server-core/src/creative.rs` | Added shared `srcset`/`imagesrcset` test helpers and merged overlapping attribute rewrite tests into compact table-driven cases while keeping integration-specific tests separate. |

## Closes

Closes #457
Closes #458

## Test plan

- [x] Other: Focused Rust test runs for the new consolidated test cases

## Checklist

- [x] New code has tests


crates/trusted-server-core/src/streaming_replacer.rs
```diff
@@ -371,61 +404,85 @@ mod tests {
...
-        let chunk1 = &content[..23];
-        let chunk2 = &content[23..];
-
-        let mut result = Vec::new();
-        result.extend(replacer.process_chunk(chunk1, false));
-        result.extend(replacer.process_chunk(chunk2, true));
+        let result = process_with_explicit_splits(
+            create_url_replacer("example.com", "https://example.com", "new.com", "https"),
+            content,
+            &[23, content.len()],
```
aram356: 🔧 **wrench** — `test_process_chunk_boundary_in_multibyte_char` still doesn't split mid-multibyte.

Byte 23 falls between `ø` (bytes 21–22) and `r` — i.e., on a valid UTF-8 boundary, not mid-character. The test name claims to exercise a split inside a multibyte char, but it doesn't. Since you correctly fixed the analogous emoji case in this same PR (the original `test_process_chunk_emoji_boundary` never split at all), please apply the same fix here.

The replacer's host-fallback rule (`example.com` → `new.com`) keeps the assertion green even when the URL-prefix replacement is severed, so the existing assertion still works.

**Fix:**
```rust
let result = process_with_explicit_splits(
    create_url_replacer("example.com", "https://example.com", "new.com", "https"),
    content,
    &[22, content.len()],
);
```
(byte 22 is in the middle of `ø` — first byte was 21).

crates/trusted-server-core/src/creative.rs
```diff
@@ -775,6 +775,49 @@ mod tests {
...
+        expected_descriptors: &[&str],
+    ) {
+        let out = rewrite_srcset_attr(attr_name, attr_value);
+
+        assert!(
+            out.matches("/first-party/proxy?tsurl=").count() >= 2,
+            "case `{}` expected two rewritten {} candidates: {}",
+            case_name,
+            attr_name,
+            out
```
aram356, line 800: 🔧 **wrench** — Assertion message and check are inconsistent.

```rust
out.matches("/first-party/proxy?tsurl=").count() >= 2,
"case `{}` expected two rewritten {} candidates: {}",
```

The message says `expected two` but the predicate accepts ≥ 2. All current cases produce exactly two rewritten candidates, so tightening the check is the correct fix and would catch any future regression that quietly rewrites a third (e.g., relative-candidate) entry.

**Fix:**
```rust
assert_eq!(
    out.matches("/first-party/proxy?tsurl=").count(),
    2,
    "case `{}` expected exactly two rewritten {} candidates: {}",
    case_name,
    attr_name,
    out
);
```

crates/trusted-server-core/src/creative.rs
```diff
@@ -775,6 +775,49 @@ mod tests {
         rewrite_creative_html, rewrite_srcset, rewrite_style_urls, sanitize_creative_html, to_abs,
     };
 
+    fn rewrite_srcset_attr(attr_name: &str, attr_value: &str) -> String {
+        let settings = crate::test_support::tests::create_test_settings();
+        let html = format!(r#"<img {}="{}">"#, attr_name, attr_value);
+        rewrite_creative_html(&settings, &html)
```
aram356, line 781: 🤔 **thinking** — The new helper wraps both `srcset` and `imagesrcset` in ``. The old `imagesrcset` tests used `<div>`. Both are HTML-spec-illegal placements — `imagesrcset` belongs on `<link rel="preload" as="image">` — but both rely on the element-agnostic `[imagesrcset]` selector at [creative.rs:660](crates/trusted-server-core/src/creative.rs#L660), so runtime behavior is identical.

Coverage gap (pre-existing, not introduced here): the realistic `<link rel="preload" as="image" imagesrcset="…">` path also flows through the `link[href]` handler at [creative.rs:556-561](crates/trusted-server-core/src/creative.rs#L556-L561). Neither the old nor new tests exercise that path. Worth a follow-up issue rather than expanding this PR.

aram356 (changes requested):
## Summary

Pure test refactor consolidating overlapping `srcset`/`imagesrcset` attribute tests in `creative.rs` and UTF-8 boundary tests in `streaming_replacer.rs`. The 1:1 mapping from old tests to new table-driven cases is faithful, and several `imagesrcset` cases gain stronger assertions (`>= 2` rewritten candidates instead of `>= 1`). Two issues need attention before merge — see inline comments.

## Blocking

### 🔧 wrench

- **Multibyte boundary test still doesn't split mid-character**: see inline at `streaming_replacer.rs:468`. The PR fixes the analogous emoji bug but leaves the multibyte case mis-named.
- **Helper assertion message contradicts its check**: see inline at `creative.rs:798`. `>= 2` with message `expected two` should be `== 2`.

## Non-blocking

### 🤔 thinking

- **`<div>` → `` element switch in helper**: see inline at `creative.rs:781`. Functionally equivalent given the `[imagesrcset]` selector, but neither old nor new tests cover the realistic `<link rel="preload" as="image" imagesrcset="…">` path. Worth a follow-up issue.
- **Loss of `cargo test <single-case>` granularity**: collapsing 8 standalone tests into 2 table-driven tests means a failing inner case can't be re-run in isolation — only the umbrella test. The `case_name` in assertion messages mitigates this for log-reading, but bisecting harder failures is more friction. Acceptable tradeoff; flagging as a real downside without a parameterized-test macro (`rstest` or similar).

### 📝 note

- **Test-name surface shrinks**: `cargo test` output loses 6 named tests (8 srcset/imagesrcset → 2 umbrellas, plus the streaming consolidation). Any CI dashboard or flake tracker keyed on test names will need to be aware.

## CI Status

- cargo fmt: PASS
- cargo clippy: PASS
- cargo test: PASS
- vitest: PASS
- integration tests / browser integration tests: PASS

crates/trusted-server-core/src/streaming_replacer.rs
ChristianPavilonis: Fixed in 4b6586c: the explicit split now uses byte 22, so this test splits inside the multibyte `ø` character as intended. Verified with `cargo test -p trusted-server-core test_process_chunk_boundary_in_multibyte_char`.

crates/trusted-server-core/src/creative.rs
ChristianPavilonis: Fixed in 4b6586c: changed this to `assert_eq!(..., 2, ...)` and updated the message to say “expected exactly two”. Verified with `cargo test -p trusted-server-core attribute_cases`.

crates/trusted-server-core/src/creative.rs
ChristianPavilonis: Acknowledged. I left this PR scoped to the refactor/blocking fixes as suggested. The realistic `<link rel="preload" as="image" imagesrcset="…">` coverage gap is a good follow-up item rather than expanding this PR.

crates/trusted-server-core/src/creative.rs
```diff
@@ -941,26 +985,50 @@ mod tests {
...
-        // Descriptors preserved
-        assert!(out.contains(" 1x"));
-        assert!(out.contains(" 2x"));
+    fn rewrites_srcset_attribute_cases() {
+        struct SrcsetCase<'a> {
+            name: &'a str,
+            attr_value: &'a str,
+            expected_relative: &'a str,
+            expected_descriptors: &'a [&'a str],
+        }
```
aram356: ♻️ **refactor** — `SrcsetCase<'a>` is defined identically here and again at line 1101 in `rewrites_imagesrcset_attribute_cases`, and both functions run a structurally identical loop differing only in the attribute name passed to the helper.

Two paths to consolidate further:

```rust
// Hoist struct to module scope, single test loops over both attrs
struct SrcsetCase<'a> { name: &'a str, attr_value: &'a str, expected_relative: &'a str, expected_descriptors: &'a [&'a str] }

#[test]
fn rewrites_srcset_and_imagesrcset_attribute_cases() {
    for attr in ["srcset", "imagesrcset"] {
        for case in CASES { assert_rewritten_srcset_attr_case(case.name, attr, ...); }
    }
}
```

Splitting per attribute does give clearer panic messages, so this is a judgment call. Suggestion only — leaving as-is is fine.

crates/trusted-server-core/src/creative.rs
```diff
@@ -775,6 +775,50 @@ mod tests {
         rewrite_creative_html, rewrite_srcset, rewrite_style_urls, sanitize_creative_html, to_abs,
     };
 
+    fn rewrite_srcset_attr(attr_name: &str, attr_value: &str) -> String {
+        let settings = crate::test_support::tests::create_test_settings();
+        let html = format!(r#"<img {}="{}">"#, attr_name, attr_value);
```
aram356, line 780: 🌱 **seedling** — `rewrite_srcset_attr` wraps both `srcset` and `imagesrcset` in ``. `` is HTML-spec-illegal placement (`imagesrcset` belongs on `<link rel="preload" as="image">` or `<source>`).

This PR does not regress on that — the pre-refactor `<div imagesrcset>` was equally spec-illegal — and the tests still validate the rewrite logic correctly via the element-agnostic `[imagesrcset]` selector at line 660. Worth a follow-up issue to add a realistic `<source>`/`<link>` test path.

aram356 (approved):
## Summary

Pure test refactor consolidating overlapping `srcset`/`imagesrcset` and UTF-8 boundary tests. The fix-up commit `4b6586cc` correctly resolves both 🔧 findings from the prior review:

1. **Mid-multibyte split** — `streaming_replacer.rs:468` split point moved from byte 23 (UTF-8 boundary, after `ø` completes) to byte 22 (between `0xC3` and `0xB8`, genuinely inside `ø`).
2. **Exact-count assertion** — `creative.rs:794` changed from `assert!(... >= 2)` to `assert_eq!(..., 2)` with matching "expected exactly two" message.

I also confirmed `test_process_chunk_boundary_in_emoji` (split at byte 2 of 🎉, a 4-byte sequence) genuinely exercises a mid-emoji boundary; the pre-refactor version processed the entire content in a single `process_chunk` call despite its name. The renames are a real coverage improvement, not just naming.

No coverage was lost in the consolidation — every old assertion (count, descriptors, relative preserved) maps 1:1 to a table case and is preserved or strengthened.

Verified locally: `cargo test -p trusted-server-core` (creative 75 ok, streaming_replacer 21 ok), `cargo clippy --all-targets --all-features -- -D warnings` clean, `cargo fmt --all -- --check` clean.

## Non-blocking

### ♻️ refactor

- **Duplicate `SrcsetCase` struct**: `creative.rs:989` and `creative.rs:1101` define an identical struct and run an identical loop. See inline comment.

### 🌱 seedling

- **Realistic placement for `imagesrcset`**: helper wraps in ``, which is spec-illegal. Pre-existing pattern; worth a follow-up. See inline comment.

### 📝 note

- **Stale checklist items in PR body**: JS tests, JS format, docs format, WASM build are unchecked but don't apply to this Rust-test-only PR. All CI is green. Informational only.

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS
- js tests: PASS
- js format: PASS
- docs format: PASS

PR #671, opened 2026-04-29
Improve Prebid auction diagnostics

### LINKED ISSUE #736: Improve Prebid auction diagnostics

ChristianPavilonis:
## Summary

- Improve Prebid auction diagnostics so upstream HTTP failures surface status in `/auction` provider metadata, with capped body previews included only when Prebid debug mode is enabled.
- Treat successful empty Prebid responses (`204 No Content` or empty `200`) as `nobid` instead of provider `error`.
- Remove Rubicon from the default `client_side_bidders` sample config so it is not enabled by default.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/integrations/prebid.rs` | Adds upstream HTTP status metadata for non-2xx responses, gates capped Prebid response body previews behind `debug`, treats empty successful responses as no-bid, uses `PREBID_INTEGRATION_ID` for provider response IDs, and adds tests for truncation/non-UTF-8 preview behavior. |
| `crates/trusted-server-core/src/auction/orchestrator.rs` | Preserves provider launch/parse/conversion errors as provider response metadata using client-safe current-context display text, and adds tests for safe output and truncation. |
| `trusted-server.toml` | Changes default `client_side_bidders` from `["rubicon"]` to `[]`. |

## Compatibility notes

- `provider_details` may include launch-failure diagnostics before awaited provider responses. Bid selection is order-insensitive, but consumers should not rely on `provider_details` being ordered by response completion.
- Successful empty Prebid responses now emit `BidStatus::NoBid` with `reason = "empty_response"` instead of provider error, which may shift error-rate/no-bid-rate dashboards after merge.

## Closes

N/A — no issue filed.

## Test plan

- [x] Other: `cargo test -p trusted-server-core prebid -- --nocapture`; `cargo test -p trusted-server-core auction -- --nocapture`; targeted review-feedback tests for `provider_error`, `prebid_body_preview`, and `parse_response_non_success`

## Checklist

- [x] Uses `log` macros per [CLAUDE.md](/CLAUDE.md) (template says `tracing`, but this repo standard is `log`)
- [x] New code has tests


crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -13,6 +13,26 @@ use super::config::AuctionConfig;
...
 use super::types::{AuctionContext, AuctionRequest, AuctionResponse, Bid, BidStatus};
 
+const PROVIDER_ERROR_MESSAGE_CHARS: usize = 500;
+
+fn provider_error_message(error: &Report<TrustedServerError>) -> String {
+    format!("{error:?}")
+        .chars()
+        .take(PROVIDER_ERROR_MESSAGE_CHARS)
+        .collect()
+}
```
aram356, line 34: 🔧 **wrench** — Information disclosure: `format!("{error:?}")` on a `Report<TrustedServerError>` emits error-stack's full Debug stack, which includes `at <file>:<line>:<col>` location entries (and backtrace frames if compiled with them). That string flows through `AuctionResponse.metadata["message"]` → `ProviderSummary.metadata` → public JSON body of `POST /auction`, exposing internal source paths to any caller.

This contradicts the explicit policy in `error.rs:18-22` ("`message` field is included in client-facing HTTP responses … keep it free of internal implementation details") and the `IntoHttpResponse::user_message()` pattern ("Full error details are preserved in server logs").

**Fix** — use Display on the current context so the user-safe `#[display(...)]` text is preserved but file/line/stack info is dropped:

```rust
fn provider_error_message(error: &Report<TrustedServerError>) -> String {
    error
        .current_context()
        .to_string()
        .chars()
        .take(PROVIDER_ERROR_MESSAGE_CHARS)
        .collect()
}
```

The detailed `{error:?}` is still emitted to logs via the existing `log::warn!("…: {:?}", e)` calls.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -1093,27 +1101,45 @@ impl AuctionProvider for PrebidAuctionProvider {
...
+                log::warn!("Prebid returned non-success status: {status} body: {body_preview}");
+            }
+
+            let mut auction_response = AuctionResponse::error("prebid", response_time_ms)
+                .with_metadata("error_type", serde_json::json!("http_status"))
+                .with_metadata("http_status", serde_json::json!(status.as_u16()));
+            if !body_preview.is_empty() {
+                auction_response =
+                    auction_response.with_metadata("body_preview", serde_json::json!(body_preview));
             }
```
aram356: 🤔 **thinking** — The first 1000 chars of the upstream Prebid Server's error body are placed in `metadata["body_preview"]` and shipped to whoever called `/auction`. Upstream Prebid error bodies routinely contain bidder-config echoes, internal request IDs, and account hints.

The trace-level body logging at line 1146 is gated on `self.config.debug && trace_enabled`; consider gating `body_preview` similarly (debug flag or request-time header) so the default `/auction` response stays free of upstream content. The `warn!` log line keeps the operator-side signal regardless.

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -274,6 +294,7 @@ impl AuctionOrchestrator {
         let mut backend_to_provider: HashMap<String, (&str, Instant, &dyn AuctionProvider)> =
             HashMap::new();
         let mut pending_requests: Vec<PlatformPendingRequest> = Vec::new();
+        let mut responses = Vec::new();
```
aram356, line 293: 🤔 **thinking** — Moving `let mut responses = Vec::new();` above the launch loop means launch failures are appended before any awaited responses. Previously the list was ordered by completion. `select_winning_bids` is order-insensitive (HashMap), but external consumers reading `provider_details` will see a different ordering than they did pre-PR. Worth either calling out in the PR description or sorting deterministically before serialization.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -1093,27 +1101,45 @@ impl AuctionProvider for PrebidAuctionProvider {
...
-                );
+        if !status.is_success() {
+            let body_preview = prebid_body_preview(&body_bytes);
+            if body_preview.is_empty() {
+                log::warn!("Prebid returned non-success status: {status} with empty body");
+            } else {
+                log::warn!("Prebid returned non-success status: {status} body: {body_preview}");
+            }
+
+            let mut auction_response = AuctionResponse::error("prebid", response_time_ms)
```
aram356: ♻️ **refactor** — Use `PREBID_INTEGRATION_ID` (defined at line 36) instead of the literal `"prebid"`. Same applies to the `no_bid("prebid", …)` call below at line 1130. Other call sites in this file already use the constant.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -391,6 +392,13 @@ impl IntegrationHeadInjector for PrebidIntegration {
     }
 }
 
+fn prebid_body_preview(body: &[u8]) -> String {
+    String::from_utf8_lossy(body)
+        .chars()
+        .take(PREBID_ERROR_BODY_PREVIEW_CHARS)
+        .collect()
+}
```
aram356: ⛏ **nitpick** — No tests pin the truncation behavior (>1000 chars) or non-UTF-8 input handling for `prebid_body_preview`, nor for `provider_error_message` (500-char cap). Cheap to add and would lock in the security boundary that today's reviewers are relying on.

aram356 (changes requested):

aram356:
## Summary

Richer `/auction` diagnostics and correct no-bid classification for empty Prebid responses are well-motivated and well-tested. One blocker: error-stack `Report` Debug formatting bleeds internal source paths into the public `/auction` JSON response, which contradicts the explicit policy in `error.rs` separating client-facing messages from server-only details.

## Blocking

### 🔧 wrench

- **Information disclosure via `format!("{error:?}")`**: `provider_error_message` ships error-stack Debug output (including `at <file>:<line>` entries) into the public `/auction` `provider_details[].metadata.message`. See inline. Switch to `error.current_context().to_string()` to keep the user-safe Display text but drop file/line/stack.

## Non-blocking

### 🤔 thinking

- **Upstream `body_preview` proxied to public clients**: The first 1000 chars of upstream Prebid error bodies are returned in the `/auction` response. Consider gating on `self.config.debug` (mirrors the existing trace-log gate) or a request-time header. See inline.
- **`provider_details` ordering changed**: Launch failures now precede awaited responses in `responses`, whereas previously the list was ordered by completion. `select_winning_bids` is unaffected, but external consumers parsing `provider_details` see a different order. See inline.

### ♻️ refactor

- **Use `PREBID_INTEGRATION_ID` constant**: New `AuctionResponse::error("prebid", …)` / `::no_bid("prebid", …)` call sites should use the existing constant (`crates/trusted-server-core/src/integrations/prebid.rs:36`) instead of the literal. See inline.

### 🌱 seedling

- **Standardize diagnostic metadata across providers**: Orchestrator-level `error_type` tagging is provider-agnostic, but only Prebid produces `http_status` / `body_preview` for non-success upstream responses. APS/GAM/etc. would be silent on the same failure mode. Worth a follow-up.

### 📌 out of scope

- **`client_side_bidders = []` is a silent client-side capability change**: The JS bundle still ships the rubicon adapter at build time (`crates/js/lib/build-all.mjs:32` → `DEFAULT_PREBID_ADAPTERS = 'rubicon'`), but the runtime default in `trusted-server.toml` no longer enables it. Publishers copying the example config will lose rubicon client-side without an adapter-time signal. Worth a CHANGELOG / release note.

### ⛏ nitpick

- **Missing truncation tests**: Neither `prebid_body_preview` (1000-char cap, `String::from_utf8_lossy`) nor `provider_error_message` (500-char cap) has a test pinning truncation/non-UTF-8 behavior. See inline.
- **`format!("{error:?}")` allocates before truncation**: Long error chains materialize a multi-KB string before `chars().take(500)` discards the tail. Becomes a non-issue if the wrench above is fixed (Display output is much shorter).

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS (incl. 4 new prebid + 1 orchestrator)
- js tests: PASS
- format-typescript / format-docs / browser integration tests / CodeQL: PASS

crates/trusted-server-core/src/auction/orchestrator.rs
ChristianPavilonis: Fixed in 50cf2fdd: provider_error_message now uses error.current_context().to_string() and keeps the 500-char cap, so public /auction metadata gets only the client-safe Display context. Existing warn logs still use Debug for server-side diagnostics. Added regression coverage to ensure attached internal details are not included and long messages are capped.

crates/trusted-server-core/src/integrations/prebid.rs
ChristianPavilonis: Fixed in 50cf2fdd: body_preview is now only included in provider metadata when Prebid debug mode is enabled. The warn log path still records the capped preview for operators regardless of debug. Updated the default non-success test to assert body_preview is absent, and added a debug-mode test for capped inclusion.

crates/trusted-server-core/src/auction/orchestrator.rs
ChristianPavilonis: Called out in the PR description under Compatibility notes. I left the runtime ordering unchanged to avoid changing mediator-visible provider_responses semantics; bid selection remains order-insensitive, and consumers should not rely on provider_details completion ordering.

crates/trusted-server-core/src/integrations/prebid.rs
ChristianPavilonis: Fixed in 50cf2fdd: replaced the new AuctionResponse prebid literals with PREBID_INTEGRATION_ID, and also converted nearby production provider-name/response call sites that were using the same literal.

crates/trusted-server-core/src/integrations/prebid.rs
ChristianPavilonis: Fixed in 50cf2fdd: added provider_error_message truncation coverage, updated the provider error metadata test to verify attached internal details are excluded, and added direct prebid_body_preview tests for >1000-char truncation and non-UTF-8 lossy handling.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -1093,27 +1101,48 @@ impl AuctionProvider for PrebidAuctionProvider {
...
+                AuctionResponse::no_bid(PREBID_INTEGRATION_ID, response_time_ms)
+                    .with_metadata("reason", serde_json::json!("empty_response"))
+                    .with_metadata("http_status", serde_json::json!(status.as_u16())),
+            );
+        }
+
+        let response_json: Json = serde_json::from_slice(&body_bytes).map_err(|e| {
+            Report::new(TrustedServerError::Prebid {
+                message: format!(
+                    "Failed to parse Prebid response JSON (status {status}, {} bytes): {e}",
```
aram356: 🔧 **wrench** — Parse-error path bypasses the debug-gate that the 4xx path enforces.

The non-success branch above carefully gates `body_preview` behind `config.debug` so untrusted upstream content doesn't reach `/auction` consumers. But on a 200 with malformed JSON, the error message itself is `"Failed to parse Prebid response JSON (status {N}, {len} bytes): {serde_err}"`, which then flows through `provider_error_message` (capped at 500 chars but **not** debug-gated) into `provider_details[].metadata.message`. `serde_json::Error::Display` is normally just `"expected value at line 1 column 1"`, so the realistic leak is the body byte length + parser category — but the asymmetry undermines the boundary the PR set up for the 4xx path, and serde messages can grow if upstream JSON gets exotic.

**Suggested fix** — keep the rich detail in the operator log, but use a static public message:

```rust
let response_json: Json = match serde_json::from_slice(&body_bytes) {
    Ok(json) => json,
    Err(e) => {
        log::warn!(
            "Prebid: failed to parse response JSON (status {status}, {} bytes): {e}",
            body_bytes.len()
        );
        return Err(Report::new(TrustedServerError::Prebid {
            message: "Failed to parse Prebid response JSON".to_string(),
        }));
    }
};
```

That keeps full diagnostics for operators (via `warn!`) while ensuring `/auction` clients see a flat, non-leaky message regardless of debug mode.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -1093,27 +1101,48 @@ impl AuctionProvider for PrebidAuctionProvider {
...
-                log::trace!(
-                    "Prebid error response body: {}",
-                    &body_preview[..body_preview.floor_char_boundary(1000)]
-                );
+        if !status.is_success() {
+            let body_preview = prebid_body_preview(&body_bytes);
+            if body_preview.is_empty() {
+                log::warn!("Prebid returned non-success status: {status} with empty body");
+            } else {
+                log::warn!("Prebid returned non-success status: {status} body: {body_preview}");
```
aram356: 🔧 **wrench** — `warn!` now emits up to 1000 chars of upstream body unconditionally.

Pre-PR, the body preview was logged at `trace` only. Now it lands in the standard `warn` path on every non-success Prebid response, regardless of `config.debug`. Upstream Prebid 4xx/5xx bodies routinely contain bidder-config echoes, internal request IDs, account hints, and occasionally PII fragments — same concern that motivated debug-gating `body_preview` in metadata, just shifted to logs.

The PR description acknowledges this ("warn log path still records the capped preview for operators regardless of debug") which is a defensible operator-visibility choice, but production logs are persisted, indexed, and often shared more broadly than `/auction` response bodies. They deserve at least the same gate.

**Suggested fix** — log the status at `warn` always, but gate the body behind `debug`:

```rust
if !status.is_success() {
    let body_preview = prebid_body_preview(&body_bytes);
    log::warn!("Prebid returned non-success status: {status} ({} bytes)", body_bytes.len());
    if self.config.debug && !body_preview.is_empty() {
        log::debug!("Prebid non-success body: {body_preview}");
    }
    // ...
}
```

This preserves the operational signal (status + size) while keeping the upstream payload behind the same gate as the metadata path.

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -353,11 +376,18 @@ impl AuctionOrchestrator {
                     );
                 }
                 Err(e) => {
+                    let response_time_ms = start_time.elapsed().as_millis() as u64;
                     log::warn!(
                         "Provider '{}' failed to launch request: {:?}",
                         provider.provider_name(),
                         e
                     );
+                    responses.push(provider_error_response(
```
aram356: 🔧 **wrench** — Launch-failure messages can surface internal identifiers to `/auction` clients.

`provider_error_message` returns `error.current_context().to_string()` (capped at 500 chars). That's a clean fix for the prior `format!("{:?}")` issue — but it still exposes whatever string each provider chose to put into its `message` field, and those weren't designed as user-facing.

Concrete example: when `request_bids` fails because `RequestSigner::from_services()` couldn't load the active key, the error message is `"Configuration error: failed to get signing key for kid: <kid>"` ([request_signing/signing.rs:140-148](crates/trusted-server-core/src/request_signing/signing.rs#L140-L148)). With this change, that kid (an internal Secret Store key identifier) lands in the public `provider_details[].metadata.message`. Kids aren't strictly secret, but they're internal state that didn't previously surface — and the same pattern means anyone adding a new `change_context(...)` site upstream of `request_bids` is now shipping that detail to `/auction` callers by default.

**Suggested fix** — give `error_type=launch_failed` a static public message, and keep the rich context in the existing `warn!` log on line 380:

```rust
Err(e) => {
    let response_time_ms = start_time.elapsed().as_millis() as u64;
    log::warn!(
        "Provider '{}' failed to launch request: {:?}",
        provider.provider_name(),
        e
    );
    let mut error_response = AuctionResponse::error(provider.provider_name(), response_time_ms)
        .with_metadata("error_type", serde_json::json!("launch_failed"));
    // Optionally surface error_type only here; full detail stays in logs.
    responses.push(error_response);
}
```

Or — if you want a per-error-type policy — split `provider_error_response` into a public path (static description per `error_type`) and an operator path (rich `Report` formatting), so future call sites can't accidentally widen the public surface.

aram356 (changes requested):
## Summary

The diagnostic improvements are well-scoped and the prior round's findings are cleanly addressed (current-context truncation, debug-gated `body_preview`, `PREBID_INTEGRATION_ID` constant, regression tests for the 500/1000-char caps and attached-detail suppression). CI is green and the new tests pin the right boundaries.

The remaining concerns are about consistency of that boundary: a few error paths slip past the debug gate that the 4xx path now establishes. Inline comments below cover the three specific call sites; cross-cutting items are listed here.

## Blocking

### 🔧 wrench

- **Parse-error message bypasses debug gate**: `parse_response` returns serde-error detail + body byte length unconditionally, while the 4xx branch hides body content behind `config.debug` (`crates/trusted-server-core/src/integrations/prebid.rs:1141`).
- **`warn!` emits up to 1000 chars of upstream body without debug gate**: previously trace-only; this is a logging behavior change worth gating in line with the metadata-path treatment (`crates/trusted-server-core/src/integrations/prebid.rs:1112`).
- **Launch-failure metadata exposes internal `message` strings**: e.g. signing-key kid via `RequestSigner::from_services()`. `provider_error_message` is sound; the issue is at the call site choosing to emit `current_context()` verbatim for `error_type=launch_failed` (`crates/trusted-server-core/src/auction/orchestrator.rs:385`).

## Non-blocking

### 🌱 seedling

- **Network-layer `select()` failures still drop providers from `provider_responses`** (`crates/trusted-server-core/src/auction/orchestrator.rs:481-486`). The PR adds metadata for `launch_failed`, `parse_response`, and `unsupported_response_body`, but transport-level errors emerging from `select()` Err — and providers still pending after the deadline (`backend_to_provider` leftovers) — produce no entry. `/auction` clients can't distinguish "no bid" from "transport error" for those. Pre-existing, but the PR's stated intent is unified observability so it's a natural follow-up.

### 📌 out of scope

- **JS bundle still defaults to bundling the Rubicon adapter** (`crates/js/lib/build-all.mjs:32`). With `client_side_bidders = []` now the default in `trusted-server.toml`, the bundled Rubicon Prebid.js adapter is dead weight unless a publisher overrides the list. Either default `TSJS_PREBID_ADAPTERS=''` to match, or have `build-all.mjs` derive the adapter list from the merged TOML. Worth a follow-up issue.

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS
- js tests: PASS
- integration / browser tests: PASS
- CodeQL / format-typescript / format-docs: PASS

crates/trusted-server-core/src/integrations/prebid.rs
ChristianPavilonis: test

crates/trusted-server-core/src/integrations/prebid.rs
ChristianPavilonis: Implemented in commit 25eaca6d: invalid Prebid JSON now logs detailed parser diagnostics server-side (status/bytes/error) but returns a fixed client-safe error message `Failed to parse Prebid response JSON`, and does not include parse detail or byte-count in /auction metadata.

crates/trusted-server-core/src/auction/orchestrator.rs
ChristianPavilonis: Implemented in commit 25eaca6d: launch-failure path now uses a dedicated safe response helper so public metadata uses `error_type: "launch_failed"` and static message `Provider launch failed`; full internal details remain in operator warn logs only.

crates/trusted-server-core/src/integrations/prebid.rs
ChristianPavilonis: Implemented in commit 25eaca6d: non-success Prebid upstream body is now warn-logged as status + byte-count only; body preview is emitted only when `debug` is enabled (and capped), matching metadata behavior.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -391,6 +392,13 @@ impl IntegrationHeadInjector for PrebidIntegration {
     }
 }
 
+fn prebid_body_preview(body: &[u8]) -> String {
```
prk-Jr: 🔧 **wrench** — `prebid_body_preview` applies the 1000-character cap only after `String::from_utf8_lossy(body)` has processed the entire upstream body. For invalid UTF-8, that can allocate a lossy `String` for the full response before `.chars().take(...)` discards nearly all of it. The caller also computes `body_preview` before checking `self.config.debug`, so this uncapped work happens even when the preview will not be logged or returned in metadata.

This leaves a resource-exhaustion edge case on large Prebid 4xx/5xx bodies. The public/logged output is capped, but the pre-cap CPU/allocation cost is not.

**Suggested fix** — gate preview construction at the call site and cap bytes before lossy conversion, for example by taking a bounded slice and using `floor_char_boundary` for valid UTF-8 or a small lossy-buffer helper for invalid UTF-8. The important part is that debug-off non-success responses should only log status + byte count, and debug-on previews should never require processing the full body.

prk-Jr (changes requested):
## Summary

The remaining review concern is in the Prebid error-body preview path. The public preview is capped, but the helper still performs uncapped UTF-8/lossy processing before the cap is applied, and it is invoked even when debug mode is off.

## Blocking

### 🔧 wrench

- **Uncapped preview work before truncation**: `prebid_body_preview` converts the entire upstream body with `String::from_utf8_lossy(body)` before taking the first 1000 chars, and the non-success response path calls it before checking `config.debug`. Large or invalid UTF-8 error bodies can still drive uncapped validation/allocation even when no preview will be exposed. See inline.

## CI Status

- GitHub checks: Analyze (javascript-typescript) PASS
- Local fmt: PASS (`cargo fmt --all -- --check`)
- Local targeted Rust tests: PASS (`prebid_body_preview`; `provider_error_response`)

crates/trusted-server-core/src/integrations/prebid.rs
ChristianPavilonis: Implemented in 50e3000d: non-success Prebid responses now construct `body_preview` only inside the `config.debug` branch, so debug-off responses only log status + byte count and do not build a preview. `prebid_body_preview` now bounds input to `PREBID_ERROR_BODY_PREVIEW_CHARS * 4` bytes before `String::from_utf8_lossy`, then keeps the existing 1000-character output cap. Added regression coverage for bytes beyond the bounded slice, plus the existing truncation/non-UTF-8 and debug metadata tests pass.

aram356 (approved):
👍 Looks good

prk-Jr (approved):
👍 Looks good

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -13,6 +13,34 @@ use super::config::AuctionConfig;
...
+const PROVIDER_ERROR_MESSAGE_CHARS: usize = 500;
+
+fn provider_error_message(error: &Report<TrustedServerError>) -> String {
+    error
+        .current_context()
+        .to_string()
+        .chars()
+        .take(PROVIDER_ERROR_MESSAGE_CHARS)
+        .collect()
+}
```
aram356, line 34: 🔧 **wrench** — Forward-looking exposure invariant should be documented inline.

`current_context().to_string()` is safe today: every `change_context` site in the orchestrator uses static or operator-controlled message strings, and `platform_response_to_fastly`'s only error is a hard-coded `"streaming platform response body is not supported by Fastly response conversion"`. But this is now an *implicit* security invariant — a future provider whose `parse_response` does

```rust
.change_context(TrustedServerError::X { message: format!("Failed: {upstream_body}") })
```

would silently route `upstream_body` into `provider_details[*].metadata.message` on `/auction`, undoing the work this PR just did.

**Fix** — add a `// SECURITY:` block above `provider_error_message`:

```rust
// SECURITY: the returned string is included verbatim (truncated to
// PROVIDER_ERROR_MESSAGE_CHARS) in the public /auction response via
// ProviderSummary.metadata["message"]. Providers MUST NOT interpolate
// upstream-controlled content (response bodies, parse errors, headers)
// into their TrustedServerError::*.message fields. Use static text and
// log details server-side with `log::warn!` instead.
fn provider_error_message(error: &Report<TrustedServerError>) -> String {
```

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -13,6 +13,34 @@ use super::config::AuctionConfig;
...
+) -> AuctionResponse {
+    AuctionResponse::error(provider_name, response_time_ms)
+        .with_metadata("error_type", serde_json::json!(error_type))
+        .with_metadata("message", serde_json::json!(provider_error_message(error)))
+}
+
+fn provider_launch_failed_response(provider_name: &str, response_time_ms: u64) -> AuctionResponse {
+    AuctionResponse::error(provider_name, response_time_ms)
+        .with_metadata("error_type", serde_json::json!("launch_failed"))
+        .with_metadata("message", serde_json::json!("Provider launch failed"))
```
aram356, line 50: 🔧 **wrench** — Lift `error_type` strings to constants.

The values `"http_status"`, `"parse_response"`, `"unsupported_response_body"`, `"launch_failed"` are an implicit public contract with downstream consumers (dashboards, alerting, mediator logic). Keeping them as string literals at four call sites is a typo risk and lacks discoverability.

**Fix** — pull them up next to `PROVIDER_ERROR_MESSAGE_CHARS`:

```rust
const PROVIDER_ERROR_MESSAGE_CHARS: usize = 500;

const ERROR_TYPE_HTTP_STATUS: &str = "http_status";
const ERROR_TYPE_PARSE_RESPONSE: &str = "parse_response";
const ERROR_TYPE_UNSUPPORTED_BODY: &str = "unsupported_response_body";
const ERROR_TYPE_LAUNCH_FAILED: &str = "launch_failed";
```

Then replace the literals at orchestrator.rs:40, 458, 473 and the prebid.rs `"http_status"` site at prebid.rs:1509. Same idea applies to the `"empty_response"` reason literal in prebid.rs:1530.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -3115,28 +3160,44 @@ server_url = "https://prebid.example"
...
-            "should step back to the previous char boundary"
+            preview.chars().count(),
+            PREBID_ERROR_BODY_PREVIEW_CHARS,
+            "should keep the public preview capped"
+        );
+        assert!(
+            !preview.contains('\u{fffd}') && !preview.contains("tail"),
+            "should not process bytes beyond the bounded preview slice"
         );
     }
```
aram356, line 3277: ⛏ **nitpick** — UTF-8 boundary test gap.

`prebid_body_preview_ignores_bytes_after_bounded_slice` exercises the byte cap with ASCII input (`b'x'`). The negative assertion `!preview.contains('\u{fffd}')` only passes because no multi-byte character straddles the byte boundary.

The real adversarial case is a 3- or 4-byte UTF-8 character whose first byte lands at `PREBID_ERROR_BODY_PREVIEW_BYTES - 1`. `from_utf8_lossy` should substitute U+FFFD at the boundary instead of leaking the next bytes. Worth a small additional test:

```rust
#[test]
fn prebid_body_preview_replaces_partial_utf8_at_boundary() {
    let mut body = vec![b'a'; PREBID_ERROR_BODY_PREVIEW_BYTES - 1];
    // Three-byte U+2603 SNOWMAN straddles the cap by 2 bytes.
    body.extend_from_slice("\u{2603}".as_bytes());
    body.extend_from_slice(b"tail");

    let preview = prebid_body_preview(&body);

    assert!(
        preview.ends_with('\u{fffd}'),
        "partial multibyte char at byte cap should become U+FFFD, not leak following bytes"
    );
    assert!(
        !preview.contains("tail"),
        "should not include bytes beyond the bounded slice"
    );
}
```

aram356 (changes requested):
## Summary

Deep re-review against `main` at HEAD `50e3000d`. Two prior approvals (aram356, prk-Jr) and all CI green. The PR has iterated through three rounds of reviewer feedback and addressed every prior finding (Debug→Display, debug-gated body previews, static launch-failure messages, `PREBID_INTEGRATION_ID` constant, byte-bounded `from_utf8_lossy`, dedicated truncation tests). The remaining findings below are forward-looking invariants and one cross-cutting concern; none block correctness.

## Blocking

### 🔧 wrench

- **SECURITY invariant comment on `provider_error_message`**: safe today because all `change_context` sites use static or operator-controlled message strings, but a future provider interpolating upstream-controlled content into `TrustedServerError::*.message` would silently undo this PR's security work. See inline at `crates/trusted-server-core/src/auction/orchestrator.rs:25`.
- **Lift `error_type` strings to constants**: `"http_status"` / `"parse_response"` / `"unsupported_response_body"` / `"launch_failed"` / `"empty_response"` are now an implicit public contract — promote to module constants to prevent typos and centralize the taxonomy. See inline at `crates/trusted-server-core/src/auction/orchestrator.rs:41`.

## Non-blocking

### 🌱 seedling

- **Opt-in for provider diagnostics exposure**: `error_type` / `message` / `http_status` / `body_preview` are now unconditionally surfaced on `/auction`. Some production operators may prefer server-logs-only diagnostics and a minimal public response. Worth a follow-up `expose_provider_diagnostics: bool` flag in `AuctionConfig` (default `true` for backward compat). Out of scope for this PR.

### 📝 note

- **No-bid taxonomy shift**: a 200 OK response with empty body previously failed `serde_json::from_slice(&[])` and surfaced as a provider error; now it is `BidStatus::NoBid` with `reason = "empty_response"`. This will shift the error-rate ↔ no-bid-rate ratio in any dashboards comparing pre/post merge. Worth a one-line release note.

### ⛏ nitpick

- **UTF-8 boundary test gap**: `prebid_body_preview_ignores_bytes_after_bounded_slice` exercises the byte cap with ASCII input only; the adversarial case is a multi-byte character straddling the byte boundary. See inline at `crates/trusted-server-core/src/integrations/prebid.rs:3202`.

## CI Status

- fmt: PASS
- clippy: PASS
- cargo test: PASS
- vitest: PASS
- browser/integration: PASS

Locally verified `cargo test -p trusted-server-core --lib` against the touched test names (all 12 new tests pass) and `cargo clippy -p trusted-server-core --all-targets --all-features -- -D warnings` (clean).

crates/trusted-server-core/src/auction/orchestrator.rs
ChristianPavilonis: Implemented in 5f00e971: added the requested `SECURITY` block above `provider_error_message`, documenting that the returned message is copied into public `/auction` metadata and that providers must keep upstream-controlled details out of `TrustedServerError::*.message` fields. Rich details remain server-side via logs.

Verified with `cargo fmt --all -- --check`, targeted provider-error/launch-failure tests, `cargo test --workspace`, and `cargo clippy --workspace --all-targets --all-features -- -D warnings`.

crates/trusted-server-core/src/auction/orchestrator.rs
ChristianPavilonis: Implemented in 5f00e971: lifted the diagnostic `error_type` values next to `PROVIDER_ERROR_MESSAGE_CHARS`, exported `ERROR_TYPE_HTTP_STATUS` for Prebid, and replaced the production literals for `launch_failed`, `parse_response`, `unsupported_response_body`, and Prebid `http_status`. Also added `PREBID_REASON_EMPTY_RESPONSE` for the `empty_response` no-bid reason literal.

Verified with `cargo fmt --all -- --check`, targeted provider-error/launch-failure tests, `cargo test --workspace`, and `cargo clippy --workspace --all-targets --all-features -- -D warnings`.

crates/trusted-server-core/src/integrations/prebid.rs
ChristianPavilonis: Implemented in 5f00e971: added `prebid_body_preview_bounds_partial_utf8_at_byte_boundary`, covering a snowman byte sequence that straddles `PREBID_ERROR_BODY_PREVIEW_BYTES` and asserting bytes after the bounded slice do not leak. Because the helper applies both a byte cap and a character cap, the test pins the bounded lossy conversion directly and also verifies the public preview remains capped and excludes `tail`.

Verified with `cargo test -p trusted-server-core prebid_body_preview`, `cargo test --workspace`, and `cargo clippy --workspace --all-targets --all-features -- -D warnings`.

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -13,6 +13,45 @@ use super::config::AuctionConfig;
...
 use super::types::{AuctionContext, AuctionRequest, AuctionResponse, Bid, BidStatus};
 
+const PROVIDER_ERROR_MESSAGE_CHARS: usize = 500;
+
+pub(crate) const ERROR_TYPE_HTTP_STATUS: &str = "http_status";
+const ERROR_TYPE_PARSE_RESPONSE: &str = "parse_response";
+const ERROR_TYPE_UNSUPPORTED_BODY: &str = "unsupported_response_body";
+const ERROR_TYPE_LAUNCH_FAILED: &str = "launch_failed";
+
+// SECURITY: the returned string is included verbatim (truncated to
```
aram356, line 21: 🌱 **seedling** — The safety of exposing `metadata["message"]` in the unauthenticated `/auction` response (`ProviderSummary` → `provider_details`, no debug/auth gate) rests entirely on every provider keeping upstream-controlled content out of `TrustedServerError::*.message`. That invariant is enforced by this comment, not the type system. Today all providers comply (verified `prebid`, `aps`, `adserver_mock` — all static messages, and `current_context()` strips error-stack attachments). But a future provider doing `change_context(TrustedServerError::Auction { message: format!("bad body: {body}") })` would silently leak upstream bytes to all callers. Consider a dedicated safe-message newtype / redaction wrapper, or a test asserting the invariant across registered providers.

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -353,11 +393,16 @@ impl AuctionOrchestrator {
                     );
                 }
                 Err(e) => {
+                    let response_time_ms = start_time.elapsed().as_millis() as u64;
                     log::warn!(
                         "Provider '{}' failed to launch request: {:?}",
                         provider.provider_name(),
                         e
                     );
+                    responses.push(provider_launch_failed_response(
```
aram356, line 379: 🤔 **thinking** — Launch-failure diagnostics are partial: only the `request_bids` `Err` path emits a `launch_failed` provider entry. Providers skipped earlier in Phase 1 — `effective_timeout == 0` (budget exhausted), missing `backend_name`, disabled, or not-registered — still `continue` silently and never appear in `provider_details`. The budget-exhausted case is arguably the most useful diagnostic and is the one now silently dropped. Intentional scope, or worth emitting a skip entry there too?

trusted-server.toml
```diff
@@ -45,7 +45,7 @@ debug = false
 # Bidders that run client-side via native Prebid.js adapters instead of
 # being routed through the server-side auction. Their adapter modules must
 # be statically imported in the JS bundle.
-client_side_bidders = ["rubicon"]
+client_side_bidders = []
```
aram356, line 72: 📝 **note** — `client_side_bidders = []` means nothing uses rubicon client-side by default, but `crates/js/lib/build-all.mjs` still sets `DEFAULT_PREBID_ADAPTERS = 'rubicon'`, so the bundle continues to ship the rubicon Prebid.js adapter. Not a bug — just dead weight / a config inconsistency worth a follow-up.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -3115,28 +3162,78 @@ server_url = "https://prebid.example"
...
+            "should keep the public preview capped"
+        );
+        assert!(
+            !preview.contains('\u{fffd}') && !preview.contains("tail"),
+            "should not process bytes beyond the bounded preview slice"
+        );
+    }
+
+    #[test]
+    fn prebid_body_preview_bounds_partial_utf8_at_byte_boundary() {
```
aram356, line 3280: ⛏ **nitpick** — This test spends most of its body asserting `String::from_utf8_lossy` (std) behavior on a hand-sliced buffer, and its own comment notes the char cap dominates so the boundary case is "beyond the public character cap for ASCII input." The valuable assertions are on the helper output (capped char count, no `"tail"`); the bounded-slice pinning exercises std, not our code, and could be trimmed.

aram356 (approved):
## Summary

Improves Prebid auction diagnostics: non-2xx upstream responses surface `error_type`/`http_status` (and a debug-gated `body_preview`) in `/auction` provider metadata, empty `204`/`200` responses become `NoBid` with `reason=empty_response` instead of provider `error`, and `rubicon` is dropped from the default `client_side_bidders`. Correctness is solid, the public-leak surface is handled deliberately, and coverage is thorough. Approving; the notes below are non-blocking.

I verified the full data path `AuctionResponse.metadata` → `ProviderSummary` → `provider_details` → public `/auction` response (no auth/debug gate), and confirmed all providers (`prebid`, `aps`, `adserver_mock`) keep upstream content out of error messages, so the diagnostic `message` field is safe.

## Non-blocking

### 🌱 seedling

- **Public-leak invariant is comment-enforced**: exposing `metadata["message"]` is safe only while every provider keeps upstream content out of `TrustedServerError` messages — enforced by a comment, not types (`crates/trusted-server-core/src/auction/orchestrator.rs:23`).

### 🤔 thinking

- **Partial launch-failure diagnostics**: only the `request_bids` `Err` path emits a `launch_failed` entry; budget-exhausted / no-backend / disabled / unregistered skips vanish from `provider_details` (`crates/trusted-server-core/src/auction/orchestrator.rs:402`).

### 📝 note / ⛏ nitpick

- **rubicon adapter still bundled by default** despite empty `client_side_bidders` (`trusted-server.toml:48`, `crates/js/lib/build-all.mjs:32`).
- **Over-engineered UTF-8 boundary test** asserts std `from_utf8_lossy` behavior more than the helper (`crates/trusted-server-core/src/integrations/prebid.rs:3207`).

## CI Status

CI on the PR ran only CodeQL (Analyze: PASS); Rust gates did not run, so verified locally on head `5f00e971`:

- fmt: PASS
- clippy (`-D warnings`): PASS
- rust tests: PASS (966 passed, 0 failed)
- js tests: not run (no JS changes in this PR)

Issue #674, opened 2026-04-30
Investigate slash-command skill content caching for SKILL.md


jevansnyc:
During validation of `/generate-feature-docs` on 2026-04-30, two consecutive slash command invocations loaded a pre-amendment cached version of `.claude/skills/generate-feature-docs/SKILL.md` even though the on-disk file had been updated and committed (commits `637e4624`, `61e84a44`). The skill executor (Claude) had to recognize the staleness and follow the disk version instead of the loaded version, otherwise the new readiness gates added in Section 16 of the design spec would not have fired.

Concretely:

- The on-disk `SKILL.md` was 408 lines and contained Step 1.7a (verification-rate threshold) and Step 1.7b (branch-state heuristic).
- The slash command system message that the harness sent to Claude on each invocation showed a 365-line version that ended at Step 1.9 with no Step 1.7a or 1.7b.
- Claude verified the on-disk content via direct file read (`grep -n "Step 1.7a"` returned a match) and chose to follow the on-disk version.

This may be session-scoped caching, a marketplace-cache layer, or some other mechanism worth understanding. Risk: a team member updating a skill in their repo may not see the new behavior fire on subsequent invocations within the same session, leading to confusion or trust issues with skill updates.

Worth investigating before relying on slash-command-driven skill rollouts.

jevansnyc:
Skill not required with new workflow

Issue #673, opened 2026-04-30
Track: propose drafts/implemented split upstream to Superpowers brainstorming skill


jevansnyc:
Track work item: propose the `drafts/` and `implemented/` lifecycle split and `status:` frontmatter convention upstream to the Superpowers brainstorming skill.

The conventions adopted in this repo (`docs/superpowers/specs/drafts/` for brainstorm output, `docs/superpowers/specs/implemented/` for post-implementation truth, with required `status:` frontmatter) are useful beyond this project. A PR to the Superpowers plugin should:

1. Update the brainstorming skill to write new specs to `<spec-root>/drafts/` by default instead of the parent directory.
2. Add `status: draft` frontmatter to newly-written specs by default.
3. Document the lifecycle (drafts → implemented) in the brainstorming skill's README.

Reference: `docs/superpowers/specs/drafts/2026-04-28-generate-feature-docs-skill-design.md` Section 14 (Related work).

The Superpowers plugin source lives at the upstream `claude-plugins-official` repository. This issue is a tracking placeholder; the actual upstream PR is filed separately when the team is ready.

Issue #672, opened 2026-04-30
Build skill #2: spec-vs-reality gap analysis


jevansnyc:
Companion to the `generate-feature-docs` skill (skill #1). While skill #1 verifies that handles (config keys, endpoints, headers, error variants) referenced in a spec exist in `crates/`, it does not verify that the *behavior* of the code matches what the spec describes.

Skill #2 closes that gap. It consumes the same structured outline that skill #1 produces in stage 1 and adds behavior verification on top. Output is a roadmap-style "planned for later" list with rough engineering estimates: which spec promises are fully shipped, which are partial, and which are missing.

Reference: `docs/superpowers/specs/drafts/2026-04-28-generate-feature-docs-skill-design.md` Section 13 (Out of scope, deferred to skill #2).

## Optional v2 enhancement

Consume GitHub PR review comments via `gh pr view <implemented_in> --comments` to incorporate post-spec engineering decisions that never made it back into the spec body. This is an optional follow-up after the core skill #2 lands; not required for v1.

PR #675, opened 2026-04-30
Add /generate-feature-docs skill for spec-to-docs generation

jevansnyc:
## Summary

Adds a Claude Code skill that converts implemented engineering specs into publisher-facing documentation pages on the Trusted Server VitePress site. Closes a recurring gap where specs get written and code ships, but the docs at iabtechlab.github.io/trusted-server do not get updated.

The skill is invoked as `/generate-feature-docs <spec-path>` and runs in two interactive stages: an extraction pass that produces a structured outline for user review, and a generation pass that writes prose, applies mechanical updates to reference docs, and commits on user approval. Built entirely on Claude Code's existing skill and slash-command primitives. No new runtime infrastructure.

## What's in this PR

- Skill at `.claude/skills/generate-feature-docs/SKILL.md` and slash command at `.claude/commands/generate-feature-docs.md`.
- Spec lifecycle convention: `docs/superpowers/specs/drafts/` for brainstorm output, `docs/superpowers/specs/implemented/` for post-implementation truth, with `status:` frontmatter (`draft`, `in-progress`, `implemented`).
- One-time migration of 12 existing specs into `drafts/` with `status: draft`.
- VitePress `srcExclude` for `superpowers/**` so internal specs no longer render on the public docs site.
- Documentation of the convention in `CLAUDE.md`.
- Design spec at `docs/superpowers/specs/drafts/2026-04-28-generate-feature-docs-skill-design.md` and implementation plan at `docs/superpowers/plans/2026-04-28-generate-feature-docs-skill.md`.

## What is NOT in this PR

- No Rust application code changes.
- No JavaScript application code changes.
- No skill behavior changes to the trusted-server runtime.
- The only build-time change is the VitePress `srcExclude` config; the only runtime impact is on the public docs site (which removes internal spec pages, the desired outcome).

## Risk profile

Low. The skill is markdown-only, no compilation, no execution path through the runtime. The most material change is to the public docs site (internal specs no longer indexed). External links to those spec URLs will break, but they were never intended to be public.

## Test plan

- [ ] Pull this branch, open Claude Code in the repo, run `/generate-feature-docs docs/superpowers/specs/drafts/2026-04-28-generate-feature-docs-skill-design.md` to confirm the slash command is auto-discovered. Reply `abort` at the readiness prompt; expect a clean exit with no files written.
- [ ] After merge, deploy the docs site and confirm iabtechlab.github.io/trusted-server no longer serves any URL under `/superpowers/`.
- [ ] Verify CI passes: `cargo fmt`, `cargo clippy`, `vitest`, `npm run format` in `docs/`, `npm run build` in `docs/`. The `cargo test --workspace` Viceroy failure is pre-existing and unrelated to this branch.
- [ ] (Optional) Run a real spec through the skill end-to-end on a feature branch. Validation runs documented in the design spec Section 11.

## Follow-up issues

- #672 Build skill #2 (spec-vs-reality gap analysis)
- #673 Track: propose drafts/implemented split upstream to Superpowers brainstorming skill
- #674 Investigate slash-command skill content caching for SKILL.md

## How to use the skill

After this lands, any team member who pulls `main` can invoke `/generate-feature-docs <spec-path>` against any spec under `docs/superpowers/specs/implemented/` (with `status: implemented` frontmatter) to generate or augment its publisher-facing guide page. The skill verifies handles (config keys, endpoints, headers, error variants) against `crates/`, surfaces drift before writing prose, and commits the result on user approval. See the design spec for the full behavior contract.

### DIFF
--- a/.claude/skills/generate-feature-docs/SKILL.md
+++ b/.claude/skills/generate-feature-docs/SKILL.md
@@ -46,4 +46,131 @@ Invoked as `/generate-feature-docs <spec-path>`. The argument is a path to a spe
 
 If the spec file does not exist, abort with a clear error. If the spec file lives outside `docs/superpowers/specs/implemented/`, warn once and ask the user to confirm before proceeding.
 
-<!-- Tasks 7, 8, 9 will append stage 1, stage 2, and edge cases below this comment. Remove this comment when those sections are added. -->
+
+## Stage 1: Extraction pass
+
+Read-only. Produces a structured outline shown to the user in chat. Do not write any files during stage 1.
+
+### Step 1.1: Parse the spec
+
+Read the spec file. Extract:
+- The H1 title (treat as the feature name).
+- The intro paragraph (treat as the description).
+- All H2 and H3 section headings.
+- All fenced code blocks. Note the language tag of each block.
+
+### Step 1.2: Detect spec kind
+
+Heuristic on section names:
+- A spec with sections like "Configuration", "Public API", or "Endpoints" is a **feature spec**. Proceed normally.
+- A spec with sections like "Migration phases" or "Rollout plan" is a **migration spec**.
+- A spec with sections like "Pre-prod checklist" or "Production readiness" is a **readiness report**.
+- Anything else with no clear kind is **unknown**.
+
+For non-feature specs and unknown specs, ask:
+> "This looks like a `<kind>` spec, not a feature spec. Continue anyway, or abort?"
+
+Do not proceed without explicit confirmation.
+
+### Step 1.3: Resolve the target page path
+
+Slug the feature name to kebab-case (e.g., "RSL AI Crawler Licensing" becomes `ai-crawler-licensing`). The target page is `docs/guide/<slug>.md`.
+
+Check if the target page already exists:
+- If exists: this is an augmentation case. Note the existing file's section structure (H2/H3 walk).
+- If not: this is a greenfield case.
+
+If a near-match exists (e.g., the slug differs only by a word), surface it as a candidate before proceeding:
+> "I will write to `docs/guide/<slug>.md`. A similar page exists at `docs/guide/<other-slug>.md`. Augment the existing page, or create a new one?"
+
+### Step 1.4: Detect Sequence-section need
+
+Heuristic: scan the spec for numbered request-flow steps, or language like "first ... then ... finally", or sequence diagrams. If present, mark `needs_sequence_section: yes` for stage 2.
+
+### Step 1.5: Detect multi-feature specs
+
+If the spec has 2 or more top-level "Feature: X" sections, or the H1 is ambiguous (covers multiple distinct features), list candidate features and ask:
+> "This spec covers multiple features: <A>, <B>, <C>. Generate one page per feature, one combined page, or a subset?"
+
+No default. The user must pick.
+
+### Step 1.6: Extract handles
+
+Walk the spec body for:
+
+- **Config keys**: TOML keys (`section.key` or `key` inside a `[section]` block), and any inline references like `the X config key`.
+- **Endpoint paths**: URL strings starting with `/`, often inside code blocks or backtick-delimited.
+- **HTTP headers**: names matching `X-...` or shown as `Header: value`.
+- **Error variants**: Rust enum variants matching `SomethingError::Variant`, and any plain-text references to error codes.
+
+Record each handle with its surface form and any context the spec gave (purpose, valid values, defaults).
+
+### Step 1.7: Verify each handle against code
+
+For each handle, search the code:
+
+- Config keys: grep `crates/**/*.rs` and `trusted-server.toml` for the key name. Capture the file and line number when found.
+- Endpoint paths: grep `crates/**/*.rs` for the path string (try both quoted and unquoted forms).
+- HTTP headers: grep for the header name as a string literal, plus any const declarations.
+- Error variants: grep for the variant name, and locate its enum definition.
+
+Mark each handle as `verified` (with `file:line`) or `NOT FOUND`.
+
+### Step 1.8: Detect spec inconsistencies
+
+Look for:
+- Same config key spelled two ways across the spec (e.g., `rsl.enabled` and `rsl_enabled`).
+- Two endpoints with the same path but different descriptions.
+- Two error variants with conflicting trigger descriptions.
+
+Surface any findings under an "Inconsistencies" subsection in the outline.
+
+### Step 1.9: Render the outline
+
+Render a single chat message in this format. Use it verbatim, filling in the values you extracted:
+
+```markdown
+## Extraction summary for `<spec-filename>`
+
+**Feature:** <feature name>
+**Target page:** `docs/guide/<slug>.md` (NEW or EXISTING)
+**Spec kind:** <feature | migration | readiness | unknown>
+**Sequence section:** <yes (brief description) | no>
+
+### Config keys
+| Key             | Status              | Location              |
+| --------------- | ------------------- | --------------------- |
+| `<key>`         | verified | NOT FOUND | `<file:line>` or "spec only" |
+
+### Endpoints
+| Path            | Methods | Status              | Location              |
+| --------------- | ------- | ------------------- | --------------------- |
+| `<path>`        | <verbs> | verified | NOT FOUND | `<file:line>` or "spec only" |
+
+### Headers
+| Name            | Direction        | Status              | Location              |
+| --------------- | ---------------- | ------------------- | --------------------- |
+| `<name>`        | request|response | verified | NOT FOUND | `<file:line>` or "spec only" |
+
+### Error variants
+| Variant         | Status              | Location              |
+| --------------- | ------------------- | --------------------- |
+| `<variant>`     | verified | NOT FOUND | `<file:line>` or "spec only" |
+
+### Inconsistencies (if any)
+- <description of inconsistency>
+
+### Issues
+For each handle marked `NOT FOUND` or each inconsistency, list options:
+- (A) Mark inline as "planned, not yet shipped"
+- (B) Drop the row from the relevant reference doc
+- (C) Pause and let me fix the spec or the code first
+
+Reply `proceed`, redirect specific fields (e.g. "use slug `rsl-licensing`"), or pick A/B/C for each issue.
+```
+
+Omit empty subsections (e.g., if no headers were extracted, omit the Headers table). Always include at least one of: Config keys, Endpoints, or Error variants. If none of these exist, the spec may not be a feature spec.
+
+### Step 1.10: Wait for user response
+
+Do not proceed to stage 2 until the user replies with `proceed` or equivalent affirmative ("yes, go ahead", "ok proceed", etc.). Substantial redirects (different slug, different target, new feature scope) regenerate the outline; minor redirects (drop a handle, override a heuristic) are noted and the skill proceeds.
--- a/.claude/skills/generate-feature-docs/SKILL.md
+++ b/.claude/skills/generate-feature-docs/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: generate-feature-docs
-description: Use when generating, writing, or updating publisher-facing documentation from an implemented engineering spec. Activates on requests like "generate docs for spec X", "write a guide page for the RSL spec", "update docs for the EC KV extension". Operates on specs under docs/superpowers/specs/implemented/ with status: implemented frontmatter.
+description: "Use when generating, writing, or updating publisher-facing documentation from an implemented engineering spec. Activates on requests like \"generate docs for spec X\", \"write a guide page for the RSL spec\", \"update docs for the EC KV extension\". Operates on specs under docs/superpowers/specs/implemented/ with status implemented frontmatter."
 ---
 
 # Generate Feature Docs
@@ -14,7 +14,7 @@ You write only to:
 - One file under `docs/guide/<feature-slug>.md` (created or augmented).
 - Up to three additive updates to `docs/guide/configuration.md`, `docs/guide/api-reference.md`, and `docs/guide/error-reference.md`.
 
-You never write anything else, never open PRs, never push, never deploy, never modify code under `crates/`, and never modify the spec you are reading.
+Writes are confined to the four files listed above. You never write any other file, never open PRs, never push, never deploy, never modify code under `crates/`, and never modify the spec you are reading.
 
 ## Spec readiness check (run first, before anything else)
 
@@ -22,9 +22,9 @@ Before doing anything else, parse the spec's YAML frontmatter and check the `sta
 
 - `status: implemented`: proceed to the extraction pass.
 - Any other value, or missing `status`: stop. Print:
-  > "This spec has `status: <value>` (or no status). The skill operates on `status: implemented` specs. Continue without status: implemented? (y/N)"
+  > "This spec has `status: <value>` (or no status). The skill operates on `status: implemented` specs. Continue without status: implemented? Reply `y` to proceed."
 
-  Default is N. Empty input or anything other than literal `y` aborts. Explicit `y` proceeds with this warning printed once:
+  Wait for the user's reply. Treat any reply other than a single `y` (case-insensitive) as abort. On `y`, print this warning once before continuing:
   > "Proceeding without `status: implemented`. The generated docs may drift from product."
 
 You never add frontmatter on the user's behalf. If the file has no frontmatter, the user must add it before re-running.
@@ -46,4 +46,4 @@ Invoked as `/generate-feature-docs <spec-path>`. The argument is a path to a spe
 
 If the spec file does not exist, abort with a clear error. If the spec file lives outside `docs/superpowers/specs/implemented/`, warn once and ask the user to confirm before proceeding.
 
-(Stage 1 and Stage 2 instructions follow in subsequent sections.)
+<!-- Tasks 7, 8, 9 will append stage 1, stage 2, and edge cases below this comment. Remove this comment when those sections are added. -->
--- a/.claude/skills/generate-feature-docs/SKILL.md
+++ b/.claude/skills/generate-feature-docs/SKILL.md
@@ -140,22 +140,22 @@ Render a single chat message in this format. Use it verbatim, filling in the val
 ### Config keys
 | Key             | Status              | Location              |
 | --------------- | ------------------- | --------------------- |
-| `<key>`         | verified | NOT FOUND | `<file:line>` or "spec only" |
+| `<key>`         | verified or NOT FOUND | `<file:line>` or "spec only" |
 
 ### Endpoints
 | Path            | Methods | Status              | Location              |
 | --------------- | ------- | ------------------- | --------------------- |
-| `<path>`        | <verbs> | verified | NOT FOUND | `<file:line>` or "spec only" |
+| `<path>`        | <verbs> | verified or NOT FOUND | `<file:line>` or "spec only" |
 
 ### Headers
-| Name            | Direction        | Status              | Location              |
-| --------------- | ---------------- | ------------------- | --------------------- |
-| `<name>`        | request|response | verified | NOT FOUND | `<file:line>` or "spec only" |
+| Name            | Direction           | Status              | Location              |
+| --------------- | ------------------- | ------------------- | --------------------- |
+| `<name>`        | request or response | verified or NOT FOUND | `<file:line>` or "spec only" |
 
 ### Error variants
 | Variant         | Status              | Location              |
 | --------------- | ------------------- | --------------------- |
-| `<variant>`     | verified | NOT FOUND | `<file:line>` or "spec only" |
+| `<variant>`     | verified or NOT FOUND | `<file:line>` or "spec only" |
 
 ### Inconsistencies (if any)
 - <description of inconsistency>
--- a/.claude/skills/generate-feature-docs/SKILL.md
+++ b/.claude/skills/generate-feature-docs/SKILL.md
@@ -174,3 +174,148 @@ Omit empty subsections (e.g., if no headers were extracted, omit the Headers tab
 ### Step 1.10: Wait for user response
 
 Do not proceed to stage 2 until the user replies with `proceed` or equivalent affirmative ("yes, go ahead", "ok proceed", etc.). Substantial redirects (different slug, different target, new feature scope) regenerate the outline; minor redirects (drop a handle, override a heuristic) are noted and the skill proceeds.
+
+## Stage 2: Generation pass
+
+Runs only after the user types `proceed`. Inputs: the spec, the approved outline from stage 1, and the existing docs. Output: files written to disk; nothing is committed until the user approves the diff.
+
+### Step 2.1: Branch check (before any writes)
+
+Detect the current git branch:
+
+```bash
+git branch --show-current
+```
+
+If the result is `main` or `master`:
+- Stop. Do not write any files.
+- Propose a branch name in the form `docs/<feature-slug>` (e.g., `docs/ai-crawler-licensing`). Ask:
+  > "You are on `<branch>`. Create branch `docs/<slug>` and switch to it?"
+- The user can specify a different branch name.
+- The skill refuses to proceed on `main` or `master` under any circumstance, including override attempts.
+- After confirmation, run `git checkout -b <branch-name>`.
+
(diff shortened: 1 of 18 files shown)


.claude/skills/generate-feature-docs/SKILL.md
```diff
@@ -0,0 +1,408 @@
...
+You convert implemented engineering specs into publisher-facing documentation pages on the Trusted Server VitePress site. You run in two interactive stages: an extraction pass that produces a structured outline for the user, and a generation pass that writes prose, updates reference docs, and commits on user approval.
+
+## Output contract
+
+You write only to:
+
+- One file under `docs/guide/<feature-slug>.md` (created or augmented).
+- Up to three additive updates to `docs/guide/configuration.md`, `docs/guide/api-reference.md`, and `docs/guide/error-reference.md`.
+
+Writes are confined to the four files listed above. You never write any other file, never open PRs, never push, never deploy, never modify code under `crates/`, and never modify the spec you are reading.
```
ChristianPavilonis, line 17: **Output scope should be a default, not an absolute prohibition**: This section uses hard boundary language: writes are confined to exactly four files and the agent never writes anything else. For a skill, I think this should be phrased as a default scope with ask-before behavior. Some docs work may legitimately need a sidebar, index, or related guide update.

**Suggestion:**
```md
Default write scope is the generated feature page and relevant reference docs. If another docs file appears necessary, explain why and ask before editing it.
```

.claude/skills/generate-feature-docs/SKILL.md
```diff
@@ -0,0 +1,408 @@
...
+
+## Spec readiness check (run first, before anything else)
+
+Before doing anything else, parse the spec's YAML frontmatter and check the `status` field.
+
+- `status: implemented`: proceed to the extraction pass.
+- Any other value, or missing `status`: stop. Print:
+  > "This spec has `status: <value>` (or no status). The skill operates on `status: implemented` specs. Continue without status: implemented? Reply `y` to proceed."
+
+  Wait for the user's reply. Treat any reply other than a single `y` (case-insensitive) as abort. On `y`, print this warning once before continuing:
```
ChristianPavilonis, line 27: **Avoid exact-response gates for normal collaboration**: Requiring a single literal `y` makes the skill feel brittle. This is useful for irreversible actions like committing, but too strict for proceeding with a draft.

**Suggestion:** Treat non-implemented status as a warning and continue on ordinary affirmative replies such as `yes`, `proceed`, `continue`, or `draft it`. Reserve exact confirmation for commits or destructive operations.

.claude/skills/generate-feature-docs/SKILL.md
```diff
@@ -0,0 +1,408 @@
...
+
+Optional frontmatter fields the skill recognizes but does not enforce:
+
+- `implemented_in`: PR number where the implementation landed.
+- `last_reviewed`: date of the most recent engineering review, in `YYYY-MM-DD` format.
+- `verified_against_commit`: commit SHA the engineer asserts the spec was verified against at promotion time. Audit trail only.
+
+If `verified_against_commit` is present, surface its value in the Stage 1 outline header (alongside the other metadata) so the user can compare it against the current branch state if drift is suspected.
+
+## Style rules (apply to ALL output, both chat messages and written files)
```
ChristianPavilonis, line 40: **Style rules should apply to generated docs, not chat**: Applying these rules to all chat messages is likely one reason the agent feels stiff. The agent should be able to collaborate naturally in chat while still producing docs that match the site style.

**Suggestion:**
```md
Style rules apply to generated documentation files. In chat, be concise, clear, and collaborative.
```

I would also remove the duplicate self-check section near the bottom of the skill.

.claude/skills/generate-feature-docs/SKILL.md
```diff
@@ -0,0 +1,408 @@
...
+Mark each handle as `verified` (with `file:line`) or `NOT FOUND`.
+
+### Step 1.7a: Compute verification rate and gate on threshold
+
+After all handles have been verified in Step 1.7:
+
+1. Compute `verified_count` (handles marked `verified`) and `total_extracted_count` (all extracted handles, regardless of type).
+2. Compute `verification_rate = verified_count / total_extracted_count`.
+3. If `total_extracted_count` is zero, this is the "no shipped code" edge case (Section "Edge cases and failure modes"). Use the existing handling for that case.
+4. If `verification_rate < 0.50`, add a hard prompt to the Stage 1 "Issues" subsection:
```
ChristianPavilonis, line 134: **The verification threshold should be a confidence warning, not a gate**: A hard 50 percent verification threshold is likely to create false friction because handle extraction can be noisy. Specs may include examples, pseudocode, planned names, or docs-only concepts that do not grep cleanly.

**Suggestion:** Summarize handles as verified, unverified, and ambiguous. Ask how to treat unverified handles that would appear in final docs, but do not block generation solely based on a numeric threshold.

.claude/skills/generate-feature-docs/SKILL.md
```diff
@@ -0,0 +1,408 @@
...
+
+### Step 1.10: Wait for user response
+
+Do not proceed to stage 2 until the user replies with `proceed` or equivalent affirmative ("yes, go ahead", "ok proceed", etc.). Substantial redirects (different slug, different target, new feature scope) regenerate the outline; minor redirects (drop a handle, override a heuristic) are noted and the skill proceeds.
+
+## Stage 2: Generation pass
+
+Runs only after the user types `proceed`. Inputs: the spec, the approved outline from stage 1, and the existing docs. Output: files written to disk; nothing is committed until the user approves the diff.
+
+### Step 2.1: Branch check (before any writes)
```
ChristianPavilonis, line 225: **Move branch and working-tree checks to commit time**: Refusing to proceed before any writes makes the skill harder to use for drafting. Users often want to generate files, inspect the diff, and decide later. Branch hygiene matters most when committing.

**Suggestion:** Allow draft generation on any branch after user approval. Before committing, check the branch and working tree, stage only the generated docs files, and ask if a docs branch should be created.

.claude/skills/generate-feature-docs/SKILL.md
```diff
@@ -0,0 +1,408 @@
...
+
+Before sending any chat message or writing any file, scan your output for:
+- Em-dashes (`—` or `–`)
+- Emojis or decorative characters
+- Exclamation marks
+- The words: "powerful", "seamless", "robust", "efficiently", "appropriately", "leveraging"
+
+If any are present, rewrite. This includes prompts, status updates, summaries, the extraction outline, and the final diff-review message.
+
+## Out of scope
```
ChristianPavilonis, line 400: **Prefer default boundaries over an out-of-scope list**: This section is written entirely as negative constraints. In practice, those `do not` instructions can make the agent stubborn. I would reframe this as default boundaries and ask-before-expanding-scope guidance.

**Suggestion:**
```md
## Default boundaries

This skill is optimized for publisher-facing docs. By default, focus on guide pages, reference docs, and code search for grounding concrete claims. If the user asks for adjacent work, such as updating narrative docs, changing specs, or preparing a PR, pause and confirm the expanded scope before doing it.
```

ChristianPavilonis (commented):
## Summary

This PR is directionally useful: the two-stage docs workflow, code-grounded extraction, spec lifecycle convention, and VitePress exclusion are all good additions. My main concern is that the skill reads too much like a policy document and uses many hard-negative rules. That will likely make the agent rigid and less collaborative. I recommend rewriting the skill around defaults, warnings, and ask-before behavior, reserving hard stops for destructive actions like pushing, committing without approval, or editing code when the task is docs-only.

.claude/skills/generate-feature-docs/SKILL.md
```diff
@@ -0,0 +1,408 @@
+---
+name: generate-feature-docs
+description: "Use when generating, writing, or updating publisher-facing documentation from an implemented engineering spec. Activates on requests like \"generate docs for spec X\", \"write a guide page for the RSL spec\", \"update docs for the EC KV extension\". Operates on specs under docs/superpowers/specs/implemented/ with status implemented frontmatter."
```
aram356, line 3: ⛏ **nitpick** — The `description` value is a single ~700-char line. YAML allows folded (`>`) or literal (`|`) block scalars that render the same to the harness and edit much more cleanly:

```yaml
description: >
  Use when generating, writing, or updating publisher-facing documentation
  from an implemented engineering spec. Activates on requests like
  "generate docs for spec X", "write a guide page for the RSL spec",
  "update docs for the EC KV extension". Operates on specs under
  docs/superpowers/specs/implemented/ with status implemented frontmatter.
```

.claude/skills/generate-feature-docs/SKILL.md
```diff
@@ -0,0 +1,408 @@
...
+- Anything else with no clear kind is **unknown**.
+
+For non-feature specs and unknown specs, ask:
+> "This looks like a `<kind>` spec, not a feature spec. Continue anyway, or abort?"
+
+Do not proceed without explicit confirmation.
+
+### Step 1.3: Resolve the target page path
+
+Slug the feature name to kebab-case (e.g., "RSL AI Crawler Licensing" becomes `ai-crawler-licensing`). The target page is `docs/guide/<slug>.md`.
```
aram356, line 85: 🤔 **thinking** — Slug rule is non-deterministic, which undermines the idempotency requirement.

The example silently drops the leading acronym (`RSL AI Crawler Licensing` → `ai-crawler-licensing`). A different invocation could produce `rsl-ai-crawler-licensing` or `rsl-licensing`. If the slug differs across runs, the augment-vs-greenfield decision flips and the "zero diff on re-run" guarantee in the Idempotency section breaks.

Either codify a mechanical rule ("lowercase, replace runs of non-alphanumeric with single hyphens, strip leading/trailing hyphens") and treat the example as a stylistic shortening that requires user confirmation, or always confirm the slug with the user on first run and persist it (e.g., via a TODO marker in the spec frontmatter).

.claude/skills/generate-feature-docs/SKILL.md
```diff
@@ -0,0 +1,408 @@
...
+- **HTTP headers**: names matching `X-...` or shown as `Header: value`.
+- **Error variants**: Rust enum variants matching `SomethingError::Variant`, and any plain-text references to error codes.
+
+Record each handle with its surface form and any context the spec gave (purpose, valid values, defaults).
+
+### Step 1.7: Verify each handle against code
+
+For each handle, search the code:
+
+- Config keys: grep `crates/**/*.rs` and `trusted-server.toml` for the key name. Capture the file and line number when found.
```
aram356, line 120: ♻️ **refactor** — `grep crates/**/*.rs` requires `shopt -s globstar` in bash and silently expands wrong on plain `sh`. Recommend modeling the right pattern:

```bash
rg --type rust --fixed-strings -- '<key>' crates/
# or
find crates -name '*.rs' -exec grep -nH -F -- '<key>' {} +
```

Claude will likely reach for `rg` regardless, but the skill should show the pattern that works portably.

🤔 **thinking** — Verification semantics include test-only matches as `verified`. A handle that exists only inside `#[cfg(test)]`, mocks, or fixtures will pass the verification-rate gate even though no production code path uses it, hiding the very drift the gate is designed to catch. Either filter `*/tests/*` (and `#[cfg(test)]` blocks) from the search or add a third state ("verified in tests only") so the user can see it in the outline.

.claude/skills/generate-feature-docs/SKILL.md
```diff
@@ -0,0 +1,408 @@
...
+After all handles have been verified in Step 1.7:
+
+1. Compute `verified_count` (handles marked `verified`) and `total_extracted_count` (all extracted handles, regardless of type).
+2. Compute `verification_rate = verified_count / total_extracted_count`.
+3. If `total_extracted_count` is zero, this is the "no shipped code" edge case (Section "Edge cases and failure modes"). Use the existing handling for that case.
+4. If `verification_rate < 0.50`, add a hard prompt to the Stage 1 "Issues" subsection:
+
+   > "Stage 1 verified `<verified_count>` of `<total_extracted_count>` handles in this codebase (`<rate>%`). Below 50% suggests this spec may not be fully implemented in this branch. Options: (A) generate stubs for unverified handles, (B) abort and check status, (C) override and proceed normally."
+
+   The user must pick A, B, or C before the skill proceeds. The threshold of 50% is initial; tune based on real usage.
```
aram356, line 138: 🌱 **seedling** — The 50% threshold is a magic number, and the design itself flags it as "initial; tune based on real usage." After a few real runs, capture the actual verification rates (and the false-positive vs false-negative outcomes at 50%) and revise. Worth a follow-up issue so the calibration doesn't get forgotten.

.claude/skills/generate-feature-docs/SKILL.md
```diff
@@ -0,0 +1,408 @@
...
+
+5. The threshold gate is in addition to per-handle Issues from Step 1.7. Both get surfaced in the same Stage 1 outline.
+
+### Step 1.7b: Branch-state heuristic
+
+In addition to handle verification, check whether the current branch has touched code relevant to the feature:
+
+```bash
+git log --name-only $(git merge-base HEAD main 2>/dev/null || git merge-base HEAD master 2>/dev/null)..HEAD -- crates/ trusted-server.toml
```
aram356, line 147: 🤔 **thinking** — If both `git merge-base HEAD main` and `... master` fail, the `$(...)` subshell expands to empty and the line becomes:

```bash
git log --name-only ..HEAD -- crates/ trusted-server.toml
```

which git rejects as a bad revision. The skill text below this block says "skip this check silently and proceed" but the literal command doesn't enforce that. Capture the merge-base into a variable and gate on it:

```bash
MB=$(git merge-base HEAD main 2>/dev/null || git merge-base HEAD master 2>/dev/null)
if [ -n "$MB" ]; then
  git log --name-only "$MB"..HEAD -- crates/ trusted-server.toml
fi
```

.claude/skills/generate-feature-docs/SKILL.md
```diff
@@ -0,0 +1,408 @@
...
+git branch --show-current
+```
+
+If the result is `main` or `master`:
+- Stop. Do not write any files.
+- Propose a branch name in the form `docs/<feature-slug>` (e.g., `docs/ai-crawler-licensing`). Ask:
+  > "You are on `<branch>`. Create branch `docs/<slug>` and switch to it?"
+- The user can specify a different branch name.
+- The skill refuses to proceed on `main` or `master` under any circumstance, including override attempts.
+- After confirmation, run `git checkout -b <branch-name>`.
```
aram356, line 239: ♻️ **refactor** — `git checkout -b <branch-name>` will fail if `docs/<slug>` already exists (which it will on a re-run of the skill against the same spec on the same day). Modern git also prefers `git switch`. Suggested:

```bash
if git show-ref --verify --quiet "refs/heads/docs/<slug>"; then
  git switch "docs/<slug>"
else
  git switch -c "docs/<slug>"
fi
```

.claude/skills/generate-feature-docs/SKILL.md
```diff
@@ -0,0 +1,408 @@
...
+
+- `commit`: proceed to step 2.7.
+- `show diff`: run `git diff` against the modified files, paste the output inline, then re-prompt: "Reply `commit` or redirect a section."
+- A redirect ("Overview is too long, cut it in half" or "rewrite the Configuration section to use the new key"): apply the redirect to the named section only, re-show the diff for the affected file, then re-prompt.
+
+Do not proceed to commit without explicit `commit`.
+
+### Step 2.7: Commit
+
+Stage explicitly via path:
```
aram356, line 333: 🤔 **thinking** — No prettier run before commit. The repo enforces `cd docs && npm run format` in CI (the `format-docs` check). LLM-emitted markdown is unlikely to match prettier's rules exactly, so users who push the skill's commit will hit a CI failure and have to amend.

Suggest running the formatter on just the modified files before the diff-review message in Step 2.5:

```bash
(cd docs && npx prettier --write \
  guide/<slug>.md \
  guide/configuration.md \
  guide/api-reference.md \
  guide/error-reference.md)
```

or, simpler: instruct the user to run `npm run format` before pushing as part of the final "Committed as `<sha>`" message.

docs/.vitepress/config.mts
```diff
@@ -38,6 +38,7 @@ export default withMermaid(
     description:
       'Privacy-preserving edge computing for ad serving and edge cookie (EC) generation',
     base: '/trusted-server',
+    srcExclude: ['superpowers/**', '**/node_modules/**'],
```
aram356, line 41: 📌 **out of scope** — `srcExclude` works (verified locally: `cd docs && npm run build` produces 40 HTML files, zero under `*/superpowers/*` in `dist/`), but there is no automated guard against a future edit silently re-exposing internal specs. The plan's Task 1 Step 4 documents the manual check (`find docs/.vitepress/dist -path '*/superpowers/*'`) but it never made it into a CI assertion.

Suggest a follow-up issue to wire that check into the docs build job — a one-line post-build assertion that exits non-zero on any hit.

aram356 (changes requested):
## Summary

The skill design is solid and the two-stage flow (outline → user redirects → diff → explicit `commit`) with hard stops on `main`/`master` and uncommitted-tree state is well-thought-out. CI is green. I built the docs locally and confirmed `srcExclude` produces 40 HTML files with zero under `*/superpowers/*` in `dist/`.

One blocker: the migration of existing specs into `drafts/` is incomplete, and the body's count claim does not match the diff.

## Blocking

### 🔧 wrench

- **Incomplete migration: `docs/superpowers/specs/2026-04-02-pr10-logging-initialization-design.md` was not moved into `drafts/`.** The PR body claims "One-time migration of 12 existing specs into `drafts/`," and the implementation plan ([`docs/superpowers/plans/2026-04-28-generate-feature-docs-skill.md`](https://github.com/IABTechLab/trusted-server/blob/feature/generate-feature-docs-skill/docs/superpowers/plans/2026-04-28-generate-feature-docs-skill.md), Task 3) lists 12 files, with Step 1 verifying `ls docs/superpowers/specs/*.md | wc -l == 12`. The actual diff shows **8 renames + 4 net-new drafts**, and at PR HEAD the file `docs/superpowers/specs/2026-04-02-pr10-logging-initialization-design.md` still sits at the spec root, not under `drafts/`.

  This file landed on `main` in commit `73101982` (PR #610), which is in the merge base of this branch (`c7e1e692`). The migration commit `17a2207a Move existing specs to drafts/...` ran before this file was on the branch and was never re-run. CI didn't catch it because `srcExclude: 'superpowers/**'` masks the entire tree from the public build. The structural intent ("all specs live under `drafts/` or `implemented/`") is broken on `main` once this lands.

  **Fix:**
  ```bash
  git mv docs/superpowers/specs/2026-04-02-pr10-logging-initialization-design.md \
         docs/superpowers/specs/drafts/2026-04-02-pr10-logging-initialization-design.md
  # then prepend the same `---\nstatus: draft\n---\n\n` frontmatter the other 8 received
  ```

  *(This file is plausibly a candidate for `implemented/` since PR #610 has shipped, but the convention is "spec lands in `drafts/` first, gets promoted in a later body-update PR" — defaulting to `drafts/` is the safe choice for this PR.)*

### ❓ question

- **PR body says 12-spec migration; diff shows 8 renames + 4 net-new drafts.** Was "12" intended literally (in which case the wrench above plus three more — but I can only find one stray spec at `specs/`), or was it a stale plan reference and the actual count was always 8 renames? Please reconcile the body with what landed once the wrench is fixed.

## Non-blocking

### 🤔 thinking

- **SKILL Step 1.7b bash one-liner can fail open** (SKILL.md:147). Subshell expands to empty when both `main` and `master` are absent, producing an invalid revision range; gate explicitly. Inline comment with details.
- **Verification semantics include test-only matches as `verified`** (SKILL.md:120). A handle that exists only inside `#[cfg(test)]`/mocks passes the verification-rate gate. Filter or add a third state. Inline comment.
- **Slug rule is non-deterministic, undermining the idempotency claim** (SKILL.md:85). Inline comment with details.
- **No prettier run before commit** (SKILL.md:333). The skill commits LLM-emitted markdown that is unlikely to match prettier; pushing will fail CI's `format-docs` check. Inline comment.

### ♻️ refactor

- **`git checkout -b` doesn't handle existing branch; prefer `git switch`** (SKILL.md:239). Inline comment.
- **`grep crates/**/*.rs` relies on shell globstar** (SKILL.md:120). Inline comment with portable alternatives.

### 🌱 seedling

- **50% verification-rate threshold is a tunable magic number** (SKILL.md:138). Track and revisit after real usage. Inline comment.
- **Empty `implemented/` at merge time.** First-time use of `/generate-feature-docs` hits an empty directory. Consider a follow-up that promotes one shipped spec (e.g., the EC KV work, or the same PR10 logging spec referenced above) so the skill is exercisable end-to-end against this repo.

### 📌 out of scope

- **No automated guard against re-exposing `superpowers/` on the public docs site** (config.mts:41). Plan documents the manual `find` check but it isn't in CI; a small post-build assertion would prevent regressions. Inline comment.

## CI Status

- cargo fmt: PASS
- clippy (Analyze rust): PASS
- cargo test: PASS
- vitest: PASS
- format-docs: PASS
- format-typescript: PASS
- CodeQL: PASS
- integration tests: PASS

Local verification: `cd docs && npm run build` succeeds, `find docs/.vitepress/dist -path '*/superpowers/*'` returns zero results.

.claude/skills/generate-feature-docs/SKILL.md
jevansnyc: Ok, thats fine

.claude/skills/generate-feature-docs/SKILL.md
jevansnyc: fine ... i'm trying to enforce DEV discipline here so lets try this way to begin

.claude/skills/generate-feature-docs/SKILL.md
jevansnyc: Have you tried the agent? It doesn't feel stiff. Sorry, but this is an agent to agent response, not real experience

.claude/skills/generate-feature-docs/SKILL.md
jevansnyc: No, completely disagree. I want finalized documentation, not "notes as you go"

.claude/skills/generate-feature-docs/SKILL.md
jevansnyc: fine

.claude/skills/generate-feature-docs/SKILL.md
jevansnyc: Huh? @ChristianPavilonis rewrite this and tell me if you feel strongly about it.

.claude/skills/generate-feature-docs/SKILL.md
jevansnyc: You're saying you want the prompt to be more flexible and natural language? I avoided this to enforce developer structure and discipline to create good documentation, but we can have a convo about it.

.claude/skills/generate-feature-docs/SKILL.md
jevansnyc: Good point, we can open this up to be more flexbile

.claude/skills/generate-feature-docs/SKILL.md
jevansnyc: Ok, good point, up to you on improved approach

.claude/skills/generate-feature-docs/SKILL.md
jevansnyc: This is a better pushback than the last one and reasonable

.claude/skills/generate-feature-docs/SKILL.md
jevansnyc: great, thanks

.claude/skills/generate-feature-docs/SKILL.md
jevansnyc: Ok

.claude/skills/generate-feature-docs/SKILL.md
jevansnyc: Ok, makes sense, have the dev run 'npm run format'

docs/.vitepress/config.mts
jevansnyc: sounds good on follow-up issue

aram356:
Closing in place of https://github.com/IABTechLab/trusted-server/issues/1105

PR #621, opened 2026-04-07
Implement Edge Cookie identity system with KV graph and integration tests

ChristianPavilonis:
## Summary

- **Implement the Edge Cookie (EC) identity system** — a publisher-owned, first-party cookie-based identity mechanism that replaces the previous Synthetic ID concept. EC enables consent-gated identity generation, a KV-backed identity graph with CAS concurrency, partner sync (pixel, batch, pull), an identify endpoint, and auction bidstream decoration with partner EIDs.
- **Add 7 integration test scenarios** covering the full EC lifecycle: generation, consent withdrawal, identify edge cases, concurrent partner syncs, and batch sync flows.
- **Rename "Synthetic ID" → "Edge Cookie"** across the entire codebase, configuration, and documentation.

## Changes

| File / Area | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/ec/` (new module) | Full EC implementation: `mod.rs` (lifecycle orchestration), `generation.rs` (HMAC-SHA256), `cookies.rs` (cookie read/write), `consent.rs` (consent gating), `device.rs` (device signal derivation, bot gate), `kv.rs` + `kv_types.rs` (KV identity graph with CAS), `partner.rs` (partner registry + admin), `sync_pixel.rs` (pixel sync), `batch_sync.rs` (S2S batch sync), `pull_sync.rs` (background pull sync), `identify.rs` (identity lookup with CORS), `eids.rs` (EID encoding), `finalize.rs` (response finalization middleware), `admin.rs` (admin endpoints) |
| `crates/integration-tests/` | 7 new EC lifecycle test scenarios in `tests/common/ec.rs` and `tests/frameworks/scenarios.rs`; updated `integration.rs` test runner and Viceroy config fixtures |
| `crates/trusted-server-adapter-fastly/src/main.rs` | New EC endpoint routes namespaced under `/_ts/api/v1/` and `/_ts/admin/`; `send_to_client()` pattern with background pull sync dispatch |
| `crates/trusted-server-core/src/auction/` | `endpoints.rs` and `formats.rs` updated to decorate bid requests with partner EIDs from the KV identity graph (`user.id`, `user.eids`, `user.consent`) |
| `crates/trusted-server-core/src/integrations/prebid.rs` | Bidder deduplication fix; blank auction EC header handling |
| `crates/js/lib/src/integrations/prebid/index.ts` | JS-side Prebid integration updates for EC headers |
[...]

PR #687, opened 2026-05-12
Review fixes for #621

aram356:
## Summary

Implements the actionable review findings for #621 so the branch is closer to a clean merge candidate. Stacked on top of `feature/edge-cookies-final` — merge this into that branch to absorb the fixes.

## Findings addressed

- 🔧 **CORS on `/identify` responses** — `ec/identify.rs`: preflight set CORS headers but `handle_identify` didn't set them on the actual 200/204/401/403 responses, breaking the documented credentialed cross-origin browser use case. Now classifies the origin once and applies CORS headers to every returned response when the origin is allowed.
- 🔧 **Body-size cap before buffering** — `ec/batch_sync.rs`, `auction/endpoints.rs`, `http_util.rs`: both endpoints read the full request body into memory before (or without) checking the size limit. Added `http_util::read_body_bounded`, which reads at most `max_bytes + 1` so an oversized/hostile body can't OOM the wasm instance; both handlers now return `413` when exceeded.
- 🔧 **Remove docs for unimplemented endpoints** — `docs/guide/api-reference.md`, `ec/mod.rs`, `integration-tests/.../scenarios.rs`: dropped the `POST /_ts/admin/v1/partners/register` and `GET /_ts/api/v1/sync` sections and the matching doc-comment references (partners are configured statically in `trusted-server.toml`).
- ⛏ **Prebid EID length validation** — `ec/prebid_eids.rs`: added the `MAX_UID_LENGTH` check on `eid.id` for symmetry with `pull_sync`/`batch_sync`.
- ⛏ **Rename leftover `synthetic` test constant** — `test_support.rs`: `VALID_SYNTHETIC_ID` → `VALID_EC_ID`.

Not addressed here (need author/legal decisions or a rebase): the stale-base/rebase requirement, the `Jurisdiction::Unknown` fail-closed blast radius, and the TCF-vs-US-Privacy precedence question — see the review on #621.

## Test plan

- [x] `cargo test --workspace` (1020 tests, 0 failures)
- [x] `cd crates/js/lib && npx vitest run` (282 tests, 0 failures)
- [x] `prettier --check` on the modified doc


aram356:
Closing — stacked on a stale head of #621; the PR has since been rebased and most of these findings are already addressed upstream. Re-running the review against the current head.

Issue #455, opened 2026-03-06
Add tests for synthetic ID template error paths


prk-Jr:
## Description

Synthetic ID generation in `crates/common/src/synthetic.rs` exercises error paths for invalid templates and signing operations, but those paths are not directly tested.

## Done when

- [ ] Add a test for invalid template syntax returning a synthetic ID error.
- [ ] Add targeted coverage for error-path propagation in the template/signing flow.
- [ ] Keep the existing happy-path behavior unchanged.

## Affected area

- Core (synthetic IDs, cookies, GDPR)

## Additional context

Source: `test_gap.md`, section `Synthetic ID Generation Errors`.
Parent issue: #396

ChristianPavilonis:
Triage note from the #396 PR consolidation: this issue references the old `crates/common/src/synthetic.rs` / template-based synthetic ID path. That path no longer exists on current `main` after the EC ID migration. I removed `Closes #455` from #688 so this does not get auto-closed by unrelated test coverage. Recommend either closing this as obsolete or rewriting it around the current `edge_cookie.rs` EC ID behavior if there is still a specific gap to cover.

PR #688, opened 2026-05-13
Add focused coverage for production readiness test gaps

### LINKED ISSUE #448: Add tests for TrustedServerError HTTP status mapping

### LINKED ISSUE #449: Add boundary tests for host rewrite bare-host substitution
## Description

`rewrite_bare_host_at_boundaries` in `crates/common/src/host_rewrite.rs` has complex boundary detection logic and no direct tests. This leaves host rewriting vulnerable to regressions around subdomains, punctuation, word boundaries, and host:port forms.

## Done when

- [ ] Add unit tests for exact host matches at valid boundaries.
- [ ] Add negative tests for subdomains and embedded-word partial matches.
- [ ] Add coverage for multiple occurrences, empty input, and host:port cases.
- [ ] Keep the test names focused on rewrite semantics rather than implementation details.

## Affected area

## Additional context

Source: `test_gap.md`, section `URL Rewriting Boundary Detection`.
Parent issue: #396

### LINKED ISSUE #450: Add unit tests for tsjs script source and tag generation

### LINKED ISSUE #453: Add tests for auction format conversion helpers
## Description

Format conversion helpers in `crates/common/src/auction/formats.rs` are untested, leaving banner/video mapping behavior exposed to silent regressions.

## Done when

- [ ] Add tests for banner format conversion.
- [ ] Add tests for video/default-field behavior.
- [ ] Add negative tests for invalid format identifiers or malformed inputs.
- [ ] Keep coverage focused on conversion semantics used by auction handling.

## Affected area

- Ad serving (Equativ)

## Additional context

Source: `test_gap.md`, section `Format Conversion`.
Parent issue: #396

ChristianPavilonis:
## Summary

- Add focused unit coverage for production-readiness backlog test gaps under #396.
- Cover error status mapping, bare-host rewrite boundaries, TSJS script tag/source generation, and auction format conversion behavior.
- Document the current banner-only auction request conversion behavior without changing runtime behavior.
- Supersedes #648 for the overlapping #448/#450 coverage.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/error.rs` | Added direct status-code mapping coverage for every current `TrustedServerError` variant, including a compile-time exhaustive match guard for new variants. |
| `crates/trusted-server-core/src/host_rewrite.rs` | Added direct boundary tests for exact hosts, paths, punctuation, multiple occurrences, subdomains, suffixes, empty inputs, and ports. |
| `crates/trusted-server-core/src/tsjs.rs` | Added tests for unified/deferred script source formatting, hash shape, script tag attributes, empty deferred output, and deferred order preservation. |
| `crates/trusted-server-core/src/auction/formats.rs` | Added request/response conversion tests for banner formats, bidder params, context allowlisting, malformed sizes, unsupported media skip behavior, OpenRTB response fields, mediation strategy, missing prices, and out-of-range dimensions. |

## Closes

Closes #448
Closes #449
Closes #450
Closes #453

Note: #455 references the old synthetic-template path, which no longer exists on current `main` after the EC ID migration. It should be triaged separately rather than auto-closed by this PR.

## Test plan

- [x] Other: `cargo test -p trusted-server-core status_code_maps_each_error_variant_to_expected_http_response -- --nocapture`; `cargo test -p trusted-server-core tsjs_ -- --nocapture`; `cargo test -p trusted-server-core rewrite -- --nocapture`; `cargo test -p trusted-server-core convert_tsjs_to_auction_request -- --nocapture`; `cargo test -p trusted-server-core convert_to_openrtb_response -- --nocapture`

## Checklist

- [x] New code has tests

### DIFF
--- a/crates/trusted-server-core/src/auction/formats.rs
+++ b/crates/trusted-server-core/src/auction/formats.rs
@@ -524,7 +524,7 @@ mod tests {
                 page_url: Some("https://test-publisher.com".to_string()),
             },
             user: UserInfo {
-                id: "synthetic-id".to_string(),
+                id: "ec-id".to_string(),
                 fresh_id: "fresh-id".to_string(),
                 consent: Some(ConsentContext::default()),
             },
@@ -799,7 +799,7 @@ mod tests {
         );
         assert_eq!(
             response.get_header_str(HEADER_X_TS_EC),
-            Some("synthetic-id"),
+            Some("ec-id"),
             "should set EC ID header"
         );
         assert_eq!(
--- a/crates/trusted-server-core/src/auction/formats.rs
+++ b/crates/trusted-server-core/src/auction/formats.rs
@@ -304,38 +304,8 @@ pub fn convert_to_openrtb_response(
         ..Default::default()
     };
 
-    let mut response_json =
-        serde_json::to_value(&response_body).change_context(TrustedServerError::Auction {
-            message: "Failed to serialize auction response".to_string(),
-        })?;
-
-    if let Some(object) = response_json.as_object_mut() {
-        // The generated OpenRTB types skip empty vectors, but this endpoint's
-        // response contract uses explicit empty arrays for no-bid and no-adomain
-        // cases.
-        if response_body.seatbid.is_empty() {
-            object.insert("seatbid".to_string(), JsonValue::Array(Vec::new()));
-        }
-
-        if let Some(seatbids) = object.get_mut("seatbid").and_then(JsonValue::as_array_mut) {
-            for seatbid in seatbids {
-                let Some(bids) = seatbid.get_mut("bid").and_then(JsonValue::as_array_mut) else {
-                    continue;
-                };
-
-                for bid in bids {
-                    if let Some(bid_object) = bid.as_object_mut() {
-                        bid_object
-                            .entry("adomain".to_string())
-                            .or_insert_with(|| JsonValue::Array(Vec::new()));
-                    }
-                }
-            }
-        }
-    }
-
     let body_bytes =
-        serde_json::to_vec(&response_json).change_context(TrustedServerError::Auction {
+        serde_json::to_vec(&response_body).change_context(TrustedServerError::Auction {
             message: "Failed to serialize auction response".to_string(),
         })?;
 
@@ -368,162 +338,12 @@ pub fn convert_to_openrtb_response(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::auction::orchestrator::OrchestrationResult;
-    use crate::auction::types::{AdFormat, AdSlot, MediaType};
-    use crate::constants::{HEADER_X_TS_EC_CONSENT, HEADER_X_TS_EIDS, HEADER_X_TS_EIDS_TRUNCATED};
-    use crate::openrtb::{Eid, Uid};
-
-    fn make_minimal_auction_request() -> AuctionRequest {
-        AuctionRequest {
-            id: "test-auction".to_owned(),
-            slots: vec![AdSlot {
-                id: "slot-1".to_owned(),
-                formats: vec![AdFormat {
-                    media_type: MediaType::Banner,
-                    width: 300,
-                    height: 250,
-                }],
-                floor_price: None,
-                targeting: HashMap::new(),
-                bidders: HashMap::new(),
-            }],
-            publisher: PublisherInfo {
-                domain: "test.com".to_owned(),
-                page_url: None,
-            },
-            user: UserInfo {
-                id: Some("test-ec-id".to_owned()),
-                consent: None,
-                eids: None,
-            },
-            device: None,
-            site: None,
-            context: HashMap::new(),
-        }
-    }
-
-    fn make_empty_result() -> OrchestrationResult {
-        OrchestrationResult {
-            winning_bids: HashMap::new(),
-            provider_responses: Vec::new(),
-            mediator_response: None,
-            total_time_ms: 10,
-            metadata: HashMap::new(),
-        }
-    }
-
-    fn make_settings() -> Settings {
-        crate::test_support::tests::create_test_settings()
-    }
-
-    #[test]
-    fn response_includes_eid_headers_when_eids_present() {
-        let mut request = make_minimal_auction_request();
-        request.user.eids = Some(vec![Eid {
-            source: "ssp.com".to_owned(),
-            uids: vec![Uid {
-                id: "uid-1".to_owned(),
-                atype: Some(3),
-                ext: None,
-            }],
-        }]);
-
-        let settings = make_settings();
-        let result = make_empty_result();
-
-        let response = convert_to_openrtb_response(&result, &settings, &request, true)
-            .expect("should build response");
-
-        assert!(
-            response.get_header(HEADER_X_TS_EIDS).is_some(),
-            "should include x-ts-eids header when EIDs are present"
-        );
-        assert_eq!(
-            response
-                .get_header(HEADER_X_TS_EC_CONSENT)
-                .and_then(|v| v.to_str().ok()),
-            Some("ok"),
-            "should include x-ts-ec-consent: ok when ec_allowed is true"
-        );
-        assert!(
-            response.get_header(HEADER_X_TS_EIDS_TRUNCATED).is_none(),
-            "should not include truncated header for small payload"
-        );
-    }
-
-    #[test]
-    fn response_sets_consent_header_even_without_eids() {
-        let request = make_minimal_auction_request();
-        let settings = make_settings();
-        let result = make_empty_result();
-
-        let response = convert_to_openrtb_response(&result, &settings, &request, true)
-            .expect("should build response");
-
-        assert_eq!(
-            response
-                .get_header(HEADER_X_TS_EC_CONSENT)
-                .and_then(|v| v.to_str().ok()),
-            Some("ok"),
-            "should set x-ts-ec-consent: ok based on consent, not EID presence"
-        );
-        assert!(
-            response.get_header(HEADER_X_TS_EIDS).is_none(),
-            "should omit x-ts-eids when no EIDs available"
-        );
-    }
-
-    #[test]
-    fn response_omits_consent_header_when_not_allowed() {
-        let request = make_minimal_auction_request();
-        let settings = make_settings();
-        let result = make_empty_result();
-
-        let response = convert_to_openrtb_response(&result, &settings, &request, false)
-            .expect("should build response");
-
-        assert!(
-            response.get_header(HEADER_X_TS_EC_CONSENT).is_none(),
-            "should omit x-ts-ec-consent when ec_allowed is false"
-        );
-        assert!(
-            response.get_header(HEADER_X_TS_EIDS).is_none(),
-            "should omit x-ts-eids when no EIDs available"
-        );
-        assert!(
-            response.get_header("x-ts-ec").is_none(),
-            "should not emit x-ts-ec when a valid EC is present"
-        );
-    }
-
-    #[test]
-    fn response_omits_ec_header_when_ec_id_is_none() {
-        let mut request = make_minimal_auction_request();
-        request.user.id = None;
-
-        let settings = make_settings();
-        let result = make_empty_result();
-
-        let response = convert_to_openrtb_response(&result, &settings, &request, false)
-            .expect("should build response");
-
-        assert!(
-            response.get_header("x-ts-ec").is_none(),
-            "should omit x-ts-ec when no EC ID is available"
-        );
-    }
-}
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-    use crate::auction::orchestrator::OrchestrationResult;
     use crate::auction::types::{AuctionResponse, Bid, BidStatus};
-    use crate::platform::test_support::noop_services;
+    use crate::openrtb::{Eid, Uid};
     use crate::test_support::tests::create_test_settings;
     use fastly::http::HeaderValue;
     use serde_json::json;
-    use std::collections::{HashMap, HashSet};
+    use std::collections::HashSet;
 
     fn make_request() -> Request {
         let mut req = Request::new("POST", "https://publisher.example.com/auction");
@@ -532,6 +352,10 @@ mod tests {
             HeaderValue::from_str("Mozilla/5.0 test").expect("should create user-agent header"),
         );
         req
+    }
+
+    fn make_settings() -> Settings {
+        create_test_settings()
     }
 
     fn make_auction_request() -> AuctionRequest {
@@ -549,17 +373,27 @@ mod tests {
                 bidders: HashMap::new(),
             }],
             publisher: PublisherInfo {
-                domain: "test-publisher.com".to_string(),
-                page_url: Some("https://test-publisher.com".to_string()),
+                domain: "publisher.example.com".to_string(),
+                page_url: Some("https://publisher.example.com".to_string()),
             },
             user: UserInfo {
-                id: "ec-id".to_string(),
-                fresh_id: "fresh-id".to_string(),
+                id: Some("ec-id".to_string()),
                 consent: Some(ConsentContext::default()),
+                eids: None,
             },
             device: None,
             site: None,
             context: HashMap::new(),
+        }
+    }
+
+    fn make_empty_result() -> OrchestrationResult {
+        OrchestrationResult {
+            provider_responses: Vec::new(),
+            mediator_response: None,
+            winning_bids: HashMap::new(),
+            total_time_ms: 10,
+            metadata: HashMap::new(),
         }
     }
 
@@ -620,31 +454,123 @@ mod tests {
                 ]),
             }],
             config,
+            eids: None,
         }
     }
 
-    fn convert_body_to_auction_request(
-        body: &AdRequest,
-        settings: &crate::settings::Settings,
-    ) -> AuctionRequest {
+    fn convert_body_to_auction_request(body: &AdRequest, settings: &Settings) -> AuctionRequest {
         let req = make_request();
-        let services = noop_services();
 
         convert_tsjs_to_auction_request(
             body,
             settings,
-            &services,
             &req,
             ConsentContext::default(),
-            "existing-ec-id",
-            None,
+            Some("existing-ec-id"),
         )
         .expect("should convert banner request")
     }
 
     #[test]
+    fn response_includes_eid_headers_when_eids_present() {
+        let mut request = make_auction_request();
+        request.user.eids = Some(vec![Eid {
+            source: "ssp.com".to_owned(),
+            uids: vec![Uid {
+                id: "uid-1".to_owned(),
+                atype: Some(3),
+                ext: None,
+            }],
+        }]);
+
+        let settings = make_settings();
+        let result = make_empty_result();
+
+        let response = convert_to_openrtb_response(&result, &settings, &request, true)
+            .expect("should build response");
+
+        assert!(
+            response.get_header(HEADER_X_TS_EIDS).is_some(),
+            "should include x-ts-eids header when EIDs are present"
+        );
+        assert_eq!(
+            response
+                .get_header(HEADER_X_TS_EC_CONSENT)
+                .and_then(|v| v.to_str().ok()),
+            Some("ok"),
+            "should include x-ts-ec-consent: ok when ec_allowed is true"
+        );
+        assert!(
+            response.get_header(HEADER_X_TS_EIDS_TRUNCATED).is_none(),
+            "should not include truncated header for small payload"
+        );
+    }
+
+    #[test]
+    fn response_sets_consent_header_even_without_eids() {
+        let request = make_auction_request();
+        let settings = make_settings();
+        let result = make_empty_result();
+
+        let response = convert_to_openrtb_response(&result, &settings, &request, true)
+            .expect("should build response");
+
+        assert_eq!(
+            response
+                .get_header(HEADER_X_TS_EC_CONSENT)
+                .and_then(|v| v.to_str().ok()),
+            Some("ok"),
+            "should set x-ts-ec-consent: ok based on consent, not EID presence"
+        );
+        assert!(
+            response.get_header(HEADER_X_TS_EIDS).is_none(),
+            "should omit x-ts-eids when no EIDs available"
+        );
+    }
+
+    #[test]
+    fn response_omits_consent_header_when_not_allowed() {
+        let request = make_auction_request();
+        let settings = make_settings();
+        let result = make_empty_result();
+
+        let response = convert_to_openrtb_response(&result, &settings, &request, false)
+            .expect("should build response");
+
+        assert!(
+            response.get_header(HEADER_X_TS_EC_CONSENT).is_none(),
+            "should omit x-ts-ec-consent when ec_allowed is false"
+        );
+        assert!(
+            response.get_header(HEADER_X_TS_EIDS).is_none(),
+            "should omit x-ts-eids when no EIDs available"
+        );
+        assert!(
+            response.get_header("x-ts-ec").is_none(),
+            "should not emit x-ts-ec when a valid EC is present"
+        );
+    }
+
+    #[test]
+    fn response_omits_ec_header_when_ec_id_is_none() {
+        let mut request = make_auction_request();
+        request.user.id = None;
+
+        let settings = make_settings();
+        let result = make_empty_result();
+
+        let response = convert_to_openrtb_response(&result, &settings, &request, false)
+            .expect("should build response");
+
+        assert!(
+            response.get_header("x-ts-ec").is_none(),
+            "should omit x-ts-ec when no EC ID is available"
+        );
+    }
+
+    #[test]
     fn convert_tsjs_to_auction_request_maps_banner_sizes_to_formats() {
-        let settings = create_test_settings();
+        let settings = make_settings();
         let body = make_banner_body(None);
         let auction_request = convert_body_to_auction_request(&body, &settings);
 
@@ -671,7 +597,7 @@ mod tests {
 
     #[test]
     fn convert_tsjs_to_auction_request_preserves_bidder_params() {
-        let settings = create_test_settings();
+        let settings = make_settings();
         let body = make_banner_body(None);
         let auction_request = convert_body_to_auction_request(&body, &settings);
         let slot = &auction_request.slots[0];
@@ -690,7 +616,7 @@ mod tests {
 
     #[test]
     fn convert_tsjs_to_auction_request_populates_publisher_user_device_and_site_metadata() {
-        let settings = create_test_settings();
+        let settings = make_settings();
         let body = make_banner_body(None);
         let auction_request = convert_body_to_auction_request(&body, &settings);
 
@@ -707,16 +633,17 @@ mod tests {
             "should create site metadata from settings"
         );
         assert_eq!(
-            auction_request.user.id, "existing-ec-id",
+            auction_request.user.id.as_deref(),
+            Some("existing-ec-id"),
             "should use caller-provided EC ID"
-        );
-        assert!(
-            !auction_request.user.fresh_id.is_empty(),
-            "should generate a fresh EC ID"
         );
         assert!(
             auction_request.user.consent.is_some(),
             "should preserve consent context"
+        );
+        assert!(
+            auction_request.user.eids.is_none(),
+            "should not attach EIDs during request conversion"
         );
         assert_eq!(
             auction_request
@@ -730,7 +657,7 @@ mod tests {
 
     #[test]
     fn convert_tsjs_to_auction_request_filters_context_values() {
-        let mut settings = create_test_settings();
+        let mut settings = make_settings();
         settings.auction.allowed_context_keys = HashSet::from([
             "segments".to_string(),
             "lockr_id".to_string(),
@@ -776,7 +703,7 @@ mod tests {
 
     #[test]
     fn convert_tsjs_to_auction_request_rejects_banner_sizes_that_are_not_width_height_pairs() {
-        let settings = create_test_settings();
+        let settings = make_settings();
         let req = make_request();
         let body = AdRequest {
             ad_units: vec![AdUnit {
@@ -789,17 +716,15 @@ mod tests {
                 bids: None,
             }],
             config: None,
+            eids: None,
         };
 
-        let services = noop_services();
         let err = convert_tsjs_to_auction_request(
             &body,
             &settings,
-            &services,
             &req,
             ConsentContext::default(),
-            "existing-ec-id",
-            None,
+            Some("existing-ec-id"),
         )
         .expect_err("should reject malformed banner size");
 
@@ -811,7 +736,7 @@ mod tests {
 
     #[test]
     fn convert_tsjs_to_auction_request_skips_units_without_banner_media() {
-        let settings = create_test_settings();
+        let settings = make_settings();
         let req = make_request();
         let body = AdRequest {
             ad_units: vec![
@@ -827,17 +752,15 @@ mod tests {
                 },
             ],
             config: None,
+            eids: None,
         };
 
-        let services = noop_services();
         let auction_request = convert_tsjs_to_auction_request(
             &body,
             &settings,
-            &services,
             &req,
             ConsentContext::default(),
-            "existing-ec-id",
-            None,
+            Some("existing-ec-id"),
         )
         .expect("should skip unsupported media units");
 
@@ -848,12 +771,12 @@ mod tests {
     }
 
     #[test]
-    fn convert_to_openrtb_response_serializes_winning_bid_headers_and_orchestrator_ext() {
-        let settings = create_test_settings();
+    fn convert_to_openrtb_response_serializes_winning_bid_and_orchestrator_ext() {
+        let settings = make_settings();
         let auction_request = make_auction_request();
         let result = make_result(make_bid("div-gpt-top", "appnexus", Some(2.75)));
 
-        let response = convert_to_openrtb_response(&result, &settings, &auction_request)
+        let response = convert_to_openrtb_response(&result, &settings, &auction_request, true)
             .expect("should convert auction result to OpenRTB response");
 
         assert_eq!(response.get_status(), StatusCode::OK, "should return OK");
@@ -863,14 +786,13 @@ mod tests {
             "should set JSON content type"
         );
         assert_eq!(
-            response.get_header_str(HEADER_X_TS_EC),
-            Some("ec-id"),
-            "should set EC ID header"
+            response.get_header_str(HEADER_X_TS_EC_CONSENT),
+            Some("ok"),
+            "should set EC consent header when allowed"
         );
-        assert_eq!(
-            response.get_header_str(HEADER_X_TS_EC_FRESH),
-            Some("fresh-id"),
-            "should set fresh EC ID header"
+        assert!(
+            response.get_header("x-ts-ec").is_none(),
+            "should not emit removed EC ID header"
         );
 
         let json = response_json(response);
@@ -906,13 +828,13 @@ mod tests {
 
     #[test]
     fn convert_to_openrtb_response_serializes_missing_creative_as_empty_adm() {
-        let settings = create_test_settings();
+        let settings = make_settings();
         let auction_request = make_auction_request();
         let mut bid = make_bid("div-gpt-top", "appnexus", Some(2.75));
         bid.creative = None;
         let result = make_result(bid);
 
-        let response = convert_to_openrtb_response(&result, &settings, &auction_request)
+        let response = convert_to_openrtb_response(&result, &settings, &auction_request, false)
             .expect("should convert bid without creative HTML");
 
         assert_eq!(response.get_status(), StatusCode::OK, "should return OK");
@@ -925,28 +847,30 @@ mod tests {
     }
 
     #[test]
-    fn convert_to_openrtb_response_serializes_missing_adomain_as_empty_array() {
-        let settings = create_test_settings();
+    fn convert_to_openrtb_response_omits_missing_adomain() {
+        let settings = make_settings();
         let auction_request = make_auction_request();
         let mut bid = make_bid("div-gpt-top", "appnexus", Some(2.75));
         bid.adomain = None;
         let result = make_result(bid);
 
-        let response = convert_to_openrtb_response(&result, &settings, &auction_request)
+        let response = convert_to_openrtb_response(&result, &settings, &auction_request, false)
             .expect("should convert bid without advertiser domains");
 
         assert_eq!(response.get_status(), StatusCode::OK, "should return OK");
         let json = response_json(response);
-        assert_eq!(
-            json["seatbid"][0]["bid"][0]["adomain"],
-            json!([]),
-            "should serialize missing adomain as an empty array"
+        let bid = json["seatbid"][0]["bid"][0]
+            .as_object()
+            .expect("should serialize bid as object");
+        assert!(
+            !bid.contains_key("adomain"),
+            "should preserve current wire format by omitting empty adomain"
         );
     }
 
     #[test]
     fn convert_to_openrtb_response_allows_empty_winning_bids() {
-        let settings = create_test_settings();
+        let settings = make_settings();
         let auction_request = make_auction_request();
         let result = OrchestrationResult {
             provider_responses: vec![],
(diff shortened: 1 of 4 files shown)


crates/trusted-server-core/src/error.rs
```diff
@@ -200,4 +200,130 @@ mod tests {
         };
         assert_eq!(error.user_message(), "Invalid header value");
     }
+
+    #[test]
+    fn status_code_maps_each_error_variant_to_expected_http_response() {
+        // Compile-time guard: adding a TrustedServerError variant without
+        // updating this test will fail to compile.
+        let _exhaustive = |error: &TrustedServerError| match error {
```
prk-Jr: 🤔 **thinking** — `_exhaustive` is never called; it works because Rust type-checks closure bodies at compile time even for dead closures, but the intent is non-obvious to a reader. A comment above it explaining the compile-time guard purpose would help, or use the more explicit form:

```rust
let _guard: fn(&TrustedServerError) = |e| match e {
    TrustedServerError::BadRequest { .. }
    | ...
    | TrustedServerError::Ec { .. } => (),
};
```

crates/trusted-server-core/src/error.rs
```diff
@@ -200,4 +200,130 @@ mod tests {
...
+            | TrustedServerError::Prebid { .. }
+            | TrustedServerError::Integration { .. }
+            | TrustedServerError::Proxy { .. }
+            | TrustedServerError::Forbidden { .. }
+            | TrustedServerError::AllowlistViolation { .. }
+            | TrustedServerError::Settings { .. }
+            | TrustedServerError::Ec { .. } => (),
+        };
+
+        let cases = vec![
```
prk-Jr: ⛏ **nitpick** — `let cases = vec![...]` here; existing tests in this module use array literals (`let cases = [...]`). Minor inconsistency.

crates/trusted-server-core/src/host_rewrite.rs
```diff
@@ -47,3 +47,106 @@ pub(crate) fn rewrite_bare_host_at_boundaries(
...
+
+    fn assert_rewrite(input: &str, expected: &str) {
+        assert_eq!(
+            rewrite_bare_host_at_boundaries(input, ORIGIN_HOST, REQUEST_HOST),
+            Some(expected.to_string()),
+            "should rewrite bare host at valid boundaries"
+        );
+    }
+
+    fn assert_no_rewrite(input: &str) {
```
prk-Jr: ⛏ **nitpick** — The assertion message `"should not rewrite host embedded in a larger host token"` is inaccurate when called from `returns_none_when_origin_host_is_absent` (line 89) — there the host is simply absent, not embedded. Consider accepting a message parameter or a separate `assert_absent` helper.

prk-Jr (commented):
## Summary

Pure test additions across four files — `error.rs`, `host_rewrite.rs`, `tsjs.rs`, and `auction/formats.rs`. No production code changed. All CI checks pass. Test quality is high overall; a few missing edge cases and one misleading helper message noted inline.

## Non-blocking

### 🏕 camp site

- **`Forbidden` and `AllowlistViolation` user_message untested** — Both variants fall into the `_` catch-all in `user_message` and return `"An internal error occurred"` despite being 4xx responses. The existing `server_errors_return_generic_message` test does not include them. Worth adding both to explicitly document (and lock in) the intended behavior — or, if a more descriptive message is preferred for 403s, this is the place to catch it.

### 🌱 seedling

- **Host-in-path-segment behavior unspecified** (`host_rewrite.rs`) — Input `https://cdn.example.com/assets/origin.example.com/image.png` will be rewritten because `/` is not a `is_host_char`. This may be intentional for Next.js `__next_f` payloads but is undocumented and untested. A test explicitly asserting current behavior would prevent silent regressions.

- **Empty module list edge case** (`tsjs.rs`) — `tsjs_script_src(&[])` behavior is unspecified. What does `concatenated_hash(&[])` return? Low risk but a documentation test would lock in the contract.

## CI Status

- cargo fmt: PASS
- cargo clippy: PASS
- cargo test: PASS
- vitest: PASS
- integration tests: PASS

crates/trusted-server-core/src/host_rewrite.rs
```diff
@@ -47,3 +47,106 @@ pub(crate) fn rewrite_bare_host_at_boundaries(
...
+    }
+
+    #[test]
+    fn does_not_rewrite_suffix_domains_or_host_char_continuations() {
+        assert_no_rewrite("origin.example.com.uk");
+        assert_no_rewrite("origin.example.com-prod");
+    }
+
+    #[test]
+    fn rewrites_origin_host_with_port_when_origin_includes_port() {
```
prk-Jr, line 170: 👍 **praise** — The port inclusion/exclusion distinction (`origin.example.com:8443` as origin vs. `origin.example.com` alone) is a subtle and important boundary. Good coverage.

crates/trusted-server-core/src/tsjs.rs
```diff
@@ -64,3 +64,125 @@ pub fn tsjs_deferred_script_tags(module_ids: &[&str]) -> String {
...
+
+        assert!(
+            src.starts_with("/static/tsjs=tsjs-prebid.min.js?v="),
+            "should use per-module static bundle path"
+        );
+        assert_sha256_hex_hash(hash_query_value(&src));
+    }
+
+    #[test]
+    fn tsjs_deferred_script_src_uses_empty_hash_for_unknown_module() {
```
prk-Jr, line 182: 👍 **praise** — Documenting the `?v=` empty-hash fallback as spec (rather than asserting a computed value) is a clean use of tests as specification.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -318,3 +318,409 @@ pub fn convert_to_openrtb_response(
...
+                fresh_id: "fresh-id".to_string(),
+                consent: Some(ConsentContext::default()),
+            },
+            device: None,
+            site: None,
+            context: HashMap::new(),
+        }
+    }
+
+    fn make_bid(slot_id: &str, bidder: &str, price: Option<f64>) -> Bid {
```
prk-Jr, line 395: 🏕 **camp site** — `make_bid` always sets `creative: Some(...)` and `adomain: Some(...)`, leaving two paths uncovered:

1. `creative: None` — the `else` branch in `convert_to_openrtb_response` emits `log::warn!` and sets `adm: ""`. No test verifies the response still succeeds.
2. `adomain: None` — exercises `unwrap_or_default()`, which should serialize as `"adomain": []`.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -318,3 +318,409 @@ pub fn convert_to_openrtb_response(
...
+            bidder: bidder.to_string(),
+            width: 300,
+            height: 250,
+            nurl: None,
+            burl: None,
+            metadata: HashMap::new(),
+        }
+    }
+
+    fn make_result(bid: Bid) -> OrchestrationResult {
```
prk-Jr, line 411: 🏕 **camp site** — `make_result` always produces a single winning bid. No test exercises `convert_to_openrtb_response` with multiple slots. Adding a two-slot result would verify `seatbid` ordering and per-slot `impid` values are preserved.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -318,3 +318,409 @@ pub fn convert_to_openrtb_response(
...
+            metadata: HashMap::new(),
+        }
+    }
+
+    fn response_json(mut response: Response) -> JsonValue {
+        serde_json::from_slice(&response.take_body_bytes()).expect("should parse JSON response")
+    }
+
+    #[test]
+    fn convert_tsjs_to_auction_request_maps_banner_units_bidders_context_and_request_metadata() {
```
prk-Jr: ♻️ **refactor** — This 130-line test exercises ~9 behaviors in one function: slot mapping, format sizes, bidder params, publisher domain, site metadata, EC ID, fresh ID, device user-agent, and context filtering. When it fails the full setup must be re-read to isolate which invariant broke. Consider splitting into focused tests — e.g. `maps_banner_sizes_to_formats`, `copies_user_agent_to_device_info`, `filters_disallowed_context_keys`.

crates/trusted-server-core/src/tsjs.rs
```diff
@@ -64,3 +64,125 @@ pub fn tsjs_deferred_script_tags(module_ids: &[&str]) -> String {
...
+    #[test]
+    fn tsjs_script_src_hash_changes_with_module_set() {
+        let creative_src = tsjs_script_src(&["creative"]);
+        let creative_prebid_src = tsjs_script_src(&["creative", "prebid"]);
+
+        assert_ne!(
+            creative_src, creative_prebid_src,
+            "should include requested modules in cache-busting hash"
+        );
+    }
```
aram356, line 122: 🌱 **seedling** — `tsjs_script_src_hash_changes_with_module_set` only locks the weakest invariant: "adding a module changes the hash." Two stronger contracts of `concatenate_modules` are still unverified:

1. **Order matters** — `concatenated_hash(&["a","b"]) != concatenated_hash(&["b","a"])` (parts are pushed in iteration order).
2. **`core` is deduped** — `concatenated_hash(&["core","prebid"]) == concatenated_hash(&["prebid"])` (the `if *id == "core" { continue; }` branch in `bundle.rs`).

Locking both prevents silent cache-bust regressions if `concatenate_modules` is later refactored. Pairs naturally with the empty-list seedling already noted in the prior review.

crates/trusted-server-core/src/error.rs
aram356: 👍 **praise** — Even with the readability concern flagged separately, the underlying pattern — using a closure body to force the compiler into checking exhaustiveness against the `match` — is a clean way to make "add a new variant ⇒ failing build" the default. Worth keeping (with a one-line `// Compile-time guard` header).

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -318,3 +318,409 @@ pub fn convert_to_openrtb_response(
         .with_header(HEADER_X_TS_EC_FRESH, &auction_request.user.fresh_id)
         .with_body(body_bytes))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::auction::context::ContextValue;
```
aram356: ⛏ **nitpick** — `use crate::auction::context::ContextValue;` at the file top (line 15) is already in scope here via `use super::*;`. This inner `use` is redundant.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -318,3 +318,409 @@ pub fn convert_to_openrtb_response(
...
+    use crate::auction::orchestrator::OrchestrationResult;
+    use crate::auction::types::{AuctionResponse, Bid, BidStatus};
+    use crate::platform::test_support::noop_services;
+    use crate::test_support::tests::create_test_settings;
+    use fastly::http::HeaderValue;
+    use serde_json::json;
+    use std::collections::{HashMap, HashSet};
+
+    fn make_request() -> Request {
+        let mut req = Request::new("POST", "https://publisher.example/auction");
```
aram356: ⛏ **nitpick** — `publisher.example` (here) and `advertiser.example` (line 378) use the bare reserved `.example` TLD, while line 358 uses `test-publisher.com`. CLAUDE.md says "for urls use: `example.com` domains only" — picking one keeps the fixtures consistent. (Cheapest fix: `publisher.example.com`, `advertiser.example.com`.)

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -318,3 +318,409 @@ pub fn convert_to_openrtb_response(
...
+            ConsentContext::default(),
+            "existing-ec-id",
+            None,
+        )
+        .expect("should skip unsupported media units");
+
+        assert!(
+            auction_request.slots.is_empty(),
+            "should only create slots for banner media"
+        );
```
aram356, line 816: 👍 **praise** — Pinning the silent-skip behavior for non-banner media is high-value. When video / native eventually land, that change has to either preserve this contract or knowingly break it — exactly the kind of decision a future reviewer wants surfaced by a failing test.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -318,3 +318,409 @@ pub fn convert_to_openrtb_response(
...
+        let result = make_result(make_bid("div-gpt-top", "appnexus", None));
+
+        let err = convert_to_openrtb_response(&result, &settings, &auction_request)
+            .expect_err("should reject winning bid without decoded price");
+
+        assert!(
+            format!("{err:?}").contains("has no decoded price"),
+            "should explain missing decoded price"
+        );
+    }
```
aram356, line 1073: 🌱 **seedling** — While you're testing the error path, the **empty** `winning_bids` path is also unverified. In real auctions, bidder timeouts produce an `OrchestrationResult` with no winners. The expected response is `StatusCode::OK` with an empty `seatbid` array and `total_bids == 0` — a trivial third test alongside the happy and error cases here would lock that contract.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -318,3 +318,409 @@ pub fn convert_to_openrtb_response(
...
+
+    #[test]
+    fn convert_to_openrtb_response_omits_out_of_range_dimensions() {
+        let settings = create_test_settings();
+        let auction_request = make_auction_request();
+        let mut bid = make_bid("div-gpt-top", "appnexus", Some(2.75));
+        bid.width = u32::MAX;
```
aram356, line 1080: 🌱 **seedling** — Only `bid.width` is mutated. The symmetric `bid.height = u32::MAX` exercises the same `to_openrtb_i32` helper at line 234 — covering both call sites costs one extra line and guards against accidentally diverging width/height handling in the future.

```rust
bid.width = u32::MAX;
bid.height = u32::MAX;
// ...
assert!(bid.get("w").is_none(), "should omit out-of-range width");
assert!(bid.get("h").is_none(), "should omit out-of-range height");
```

aram356 (commented):
## Summary

Follow-up to the prior review on the same head commit — only **new** findings missed last pass. Still no blocking items; verdict is `COMMENT`.

- 3 🌱 seedling — coverage gaps: hash order/dedup invariants, empty `winning_bids` response, height out-of-range symmetry.
- 2 ⛏ nitpick — redundant `use` in the test module; mixed `*.example` / `example.com` test-fixture TLDs.
- 2 👍 praise — compile-time exhaustiveness guard pattern; banner-only behavior pinned by a test.

The prior review's findings remain open (no commits since `2a3edb77`) — they're tracked there, not re-posted here.

## CI Status

- cargo fmt: PASS
- cargo clippy: PASS
- cargo test: PASS
- vitest: PASS
- integration tests: PASS

aram356 (changes requested):
Requesting changes to address the 5 non-praise findings (3 🌱 + 2 ⛏) in [my prior review](https://github.com/IABTechLab/trusted-server/pull/688#pullrequestreview-4378072529) — and the 9 still-open items from the [earlier review](https://github.com/IABTechLab/trusted-server/pull/688#pullrequestreview-4291009593) that haven't been addressed yet.

crates/trusted-server-core/src/error.rs
ChristianPavilonis: Addressed in 37b41f0: switched this to the explicit `let _guard: fn(&TrustedServerError)` form while keeping the compile-time exhaustiveness guard.

crates/trusted-server-core/src/error.rs
ChristianPavilonis: Addressed in 37b41f0: changed the status-code cases from `vec![...]` to an array literal.

crates/trusted-server-core/src/host_rewrite.rs
ChristianPavilonis: Addressed in 37b41f0: `assert_no_rewrite` now accepts a caller-specific assertion message, and the call sites distinguish absent host, embedded host, suffix, and port-continuation cases.

crates/trusted-server-core/src/host_rewrite.rs
ChristianPavilonis: Thanks — kept this boundary coverage in place.

crates/trusted-server-core/src/tsjs.rs
ChristianPavilonis: Thanks — kept the unknown-module empty-hash behavior test in place.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in 37b41f0: added focused tests for `creative = None` (`adm: ""`) and `adomain = None` (`adomain: []`). I also preserved explicit empty arrays in response serialization to match that contract.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in 37b41f0: added a two-winner response test asserting both `seatbid` entries preserve bidder, `impid`, price, and creative. The assertions are order-insensitive because `winning_bids` is a `HashMap` and production code does not currently define deterministic ordering.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in 37b41f0: split the broad request-conversion test into focused tests for banner formats, bidder params, metadata, and context filtering.

crates/trusted-server-core/src/tsjs.rs
ChristianPavilonis: Addressed in 37b41f0: added tests for module-order-dependent hashes and explicit `core` deduping. I also added the empty module-list/core-only contract test alongside them.

crates/trusted-server-core/src/error.rs
ChristianPavilonis: Thanks — kept the compile-time guard and made it explicit with `_guard: fn(&TrustedServerError)`.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in 37b41f0: removed the redundant test-module `ContextValue` import.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in 37b41f0: normalized the fixtures/assertions to `publisher.example.com` and `advertiser.example.com`.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Thanks — kept this non-banner skip coverage as-is.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in 37b41f0: added the no-winner response test asserting OK, explicit `seatbid: []`, and `total_bids == 0`; response serialization now preserves the explicit empty `seatbid` array for this contract.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in 37b41f0: the test now sets both `width` and `height` to `u32::MAX` and asserts both `w` and `h` are omitted.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -318,3 +348,611 @@ pub fn convert_to_openrtb_response(
         .with_header(HEADER_X_TS_EC_FRESH, &auction_request.user.fresh_id)
         .with_body(body_bytes))
 }
+
+#[cfg(test)]
+mod tests {
```
aram356, line 338: 🔧 **wrench** — Duplicate `mod tests`. Current `main` (after #621, the Edge Cookie identity system) already defines a `#[cfg(test)] mod tests` in this file at `formats.rs:339`. Adding a second module produces `error[E0428]: the name 'tests' is defined multiple times` — this is the root cause of the failing `cargo test` CI job. Fold these tests into the existing module rather than introducing a new one.

This whole module is also written against the pre-#621 API and won't compile as-is:
- `convert_tsjs_to_auction_request` now takes **5** args `(body, settings, req, consent, ec_id)`, not 7 (no `services`, no separate fresh-id). Calls at lines ~468/630/668 pass 7 → `E0061`.
- `AdRequest` gained `eids: Option<JsonValue>`; the struct literals omit it → `E0063`.
- `UserInfo` no longer has `fresh_id`, and `id` is now `Option<String>` (lines ~392/549) → `E0560`/`E0609`/`E0308`.
- `crate::platform::test_support::noop_services` (line ~357) becomes a dead import once `services` is dropped.

**Fix**: rebase onto current `main` and rewrite these tests against the post-#621 signatures (`id: Some(...)`, `AdRequest { eids: None, .. }`, 5-arg conversion).

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -307,8 +307,38 @@ pub fn convert_to_openrtb_response(
         ..Default::default()
     };
 
+    let mut response_json =
```
aram356: 🔧 **wrench** — Undisclosed runtime behavior change. The PR description states the auction conversion is documented *"without changing runtime behavior"*, but this block re-serializes the response via `serde_json::to_value` and post-processes the JSON to inject explicit empty `seatbid: []` and `adomain: []` arrays. That **changes the wire format** (these were previously omitted by serde's skip-empty) and adds a **second full serialization pass** (`to_value` + `to_vec`) on the hot auction-response path.

On current `main` this function was reworked by #621 to attach `HEADER_X_TS_EC_CONSENT` / `HEADER_X_TS_EIDS` headers, so this change also needs re-evaluation after rebase.

**Fix**: split the runtime change into its own PR with a clear description, and if the explicit-empty-array contract is genuinely required, prefer customizing serialization on the OpenRTB response types over re-walking the JSON tree on every response.

crates/trusted-server-core/src/error.rs
```diff
@@ -200,4 +221,130 @@ mod tests {
         };
         assert_eq!(error.user_message(), "Invalid header value");
     }
+
+    #[test]
+    fn status_code_maps_each_error_variant_to_expected_http_response() {
+        // Compile-time guard: adding a TrustedServerError variant without
+        // updating this test will fail to compile.
+        let _guard: fn(&TrustedServerError) = |error| match error {
```
aram356, line 244: 🔧 **wrench** — Exhaustive guard references a removed variant and omits new ones, so it won't compile against current `main`. After #621, `TrustedServerError` has 16 variants: `Ec` was removed and `EdgeCookie`, `PartnerNotFound`, `InsecureDefault` were added. This match (and the `status_code` case table below, which references `TrustedServerError::Ec` at line ~297) must be updated to the current set.

The compile-time guard pattern itself is good — it's doing its job by forcing the variant set to stay in sync. It just needs to reflect the post-#621 enum.

**Fix**: drop the `Ec` arm/case and add `EdgeCookie`, `PartnerNotFound`, `InsecureDefault` to both the guard match and the status-mapping cases.

aram356 (changes requested):
## Summary

This test-coverage PR was branched from a stale `main` (merge-base `f275aba5`). Since then, **#621 (Edge Cookie identity system)** reshaped the auction APIs, and the PR no longer compiles against current `main` — which is why the `cargo test` CI job is failing. `cargo fmt` is also red on the merged tree. `host_rewrite.rs` and `tsjs.rs` are sound and land cleanly; `auction/formats.rs` and `error.rs` need a rebase and rework. Requesting changes on the three blocking items below.

## Blocking

### 🔧 wrench

- **Duplicate `mod tests` + stale auction API**: `main` already has a test module at `formats.rs:339`; the new module collides (`E0428`) and is written against pre-#621 signatures (7-arg conversion, `UserInfo.fresh_id`, missing `AdRequest.eids`, removed `HEADER_X_TS_EC`/`_FRESH`). (`auction/formats.rs:353`)
- **Undisclosed runtime behavior change**: `convert_to_openrtb_response` now emits explicit empty `seatbid`/`adomain` arrays via a second serialization pass, contradicting the PR's "without changing runtime behavior" claim. (`auction/formats.rs:310`)
- **Exhaustive error guard out of date**: references removed `TrustedServerError::Ec` and omits new `EdgeCookie`/`PartnerNotFound`/`InsecureDefault`. (`error.rs:229`)

## Non-blocking

### 🤔 thinking

- **Double serialization on response path**: if the explicit-empty-array contract is required, customize serialization on the OpenRTB response types rather than re-walking the JSON tree (`to_value` + `to_vec`) on every auction response. (`auction/formats.rs:310`)

### ♻️ refactor

- **On rebase**: fold both new test modules into the existing `mod tests`; update to `UserInfo { id: Some(...), .. }` (no `fresh_id`), `AdRequest { eids: None, .. }`, the 5-arg `convert_tsjs_to_auction_request`, and assert the current `HEADER_X_TS_EC_CONSENT`/`HEADER_X_TS_EIDS` headers.

### 🌱 seedling

- **Unknown-module hash**: `tsjs_deferred_script_src_uses_empty_hash_for_unknown_module` characterizes "unknown module → `?v=` with empty hash." Worth a follow-up on whether that should instead be rejected. (`tsjs.rs`)

### 📝 note

- `tsjs.rs` tests are independent of the EC migration and should merge + compile unchanged.

## CI Status

- fmt: FAIL
- clippy: PASS
- rust tests: FAIL (compile errors against merged `main` — see blocking findings)
- js tests (vitest): PASS

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in 8d63e24: rebased onto current `origin/main`, folded the PR-added tests into the existing `mod tests`, removed the stale `noop_services` path, added `eids: None` to `AdRequest` literals, updated the 5-arg `convert_tsjs_to_auction_request` calls, and changed user assertions/helpers to the post-#621 `id: Option<String>` / no-`fresh_id` model.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in 8d63e24: removed the `serde_json::to_value` re-walk and explicit empty-array injection. `convert_to_openrtb_response` is back to direct `serde_json::to_vec(&response_body)`, preserving current `main` wire behavior and the post-#621 `x-ts-ec-consent` / `x-ts-eids` header behavior. The tests now characterize the current omitted-empty-array contract instead of changing runtime behavior.

crates/trusted-server-core/src/error.rs
ChristianPavilonis: Addressed in 8d63e24: removed stale `TrustedServerError::Ec` references and updated the compile-time guard plus status-code table for `EdgeCookie`, `PartnerNotFound`, and `InsecureDefault`. I also expanded user-message coverage for the new/current catch-all variants.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -338,16 +338,31 @@ pub fn convert_to_openrtb_response(
...
+    use fastly::http::HeaderValue;
+    use serde_json::json;
+    use std::collections::HashSet;
 
-    fn make_minimal_auction_request() -> AuctionRequest {
+    fn make_request() -> Request {
+        let mut req = Request::new("POST", "https://publisher.example.com/auction");
+        req.set_header(
+            header::USER_AGENT,
+            HeaderValue::from_str("Mozilla/5.0 test").expect("should create user-agent header"),
```
prk-Jr: ⛏ **nitpick** — `HeaderValue::from_str(...).expect(...)` is unnecessary: fastly's `set_header` takes `impl ToHeaderValue`, so a plain `&str` works (production code does exactly this at line 320 with `set_header(HEADER_X_TS_EC_CONSENT, "ok")`). This also drops the `use fastly::http::HeaderValue;` import at line 344.

```rust
fn make_request() -> Request {
    let mut req = Request::new("POST", "https://publisher.example.com/auction");
    req.set_header(header::USER_AGENT, "Mozilla/5.0 test");
    req
}
```

prk-Jr (commented):
## Summary

Third-pass review of head `8d63e24` — all findings from the two prior review rounds are addressed. Only new findings below; nothing blocking. Verdict remains `COMMENT`.

- 2 ⛏ nitpick — redundant `HeaderValue::from_str`, under-enumerated `user_message` doc
- 1 🏕 camp site — stale `# Errors` doc on `convert_tsjs_to_auction_request` (below)
- 1 📌 out of scope — misleading generic message now pinned for 403/404 variants
- 3 🌱 seedling — `ec_id: None` propagation, empty banner `sizes`, host-rewrite case sensitivity
- 1 👍 praise — order-insensitive multi-winner assertions

## Non-blocking (body-level)

### 🏕 camp site

- **Stale `# Errors` doc on `convert_tsjs_to_auction_request`** (`crates/trusted-server-core/src/auction/formats.rs:82`, outside this diff): the entry "Fresh EC ID generation fails" is stale — the function never generates EC IDs; the caller does, per the doc paragraph at lines 76–77, and the only error path is the invalid-banner-size `ensure!`. Since this PR's purpose is documenting this function's behavior, removing the stale bullet fits here:

  ```rust
  /// # Errors
  ///
  /// Returns an error if the request contains invalid banner sizes
  /// (must be `[width, height]`).
  ```

## CI Status

- fmt: PASS
- clippy (Analyze rust): PASS
- rust tests (cargo test): PASS
- js tests (vitest): PASS
- integration + browser integration: PASS

crates/trusted-server-core/src/error.rs
```diff
@@ -103,8 +103,9 @@ pub trait IntoHttpResponse {
 
     /// Get a safe, user-facing error message.
     ///
-    /// Client errors (4xx) return a brief description; server/integration errors
-    /// return a generic message. Full error details are preserved in server logs.
+    /// Selected client errors return a brief description; forbidden and
+    /// server/integration errors return a generic message. Full error details
```
prk-Jr: ⛏ **nitpick** — The reworded doc says "forbidden and server/integration errors return a generic message", but `PartnerNotFound` (404) also returns the generic message, and it is neither forbidden nor a server/integration error. Enumerating categories here will keep drifting as variants are added. Simpler and future-proof:

```rust
/// Selected client errors return a brief description; all other errors
/// return a generic message. Full error details are preserved in server logs.
```

crates/trusted-server-core/src/error.rs
```diff
@@ -193,6 +197,29 @@ mod tests {
         }
     }
 
+    #[test]
+    fn other_client_errors_return_generic_user_message() {
```
prk-Jr, line 200: 📌 **out of scope** — This test now pins `"An internal error occurred"` for `Forbidden` / `AllowlistViolation` (403) and `PartnerNotFound` (404). That message is misleading for client-fault statuses: a blocked redirect or unknown partner is not an internal error, and clients debugging a 403/404 get actively wrong guidance. Pinning current behavior is right for this PR — but a follow-up issue could introduce distinct generic messages (e.g. `"Forbidden"`, `"Not found"`) without leaking detail.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -374,21 +389,91 @@ mod tests {
...
+                        params: json!({ "accountId": 456 }),
+                    },
+                ]),
+            }],
+            config,
+            eids: None,
+        }
+    }
+
+    fn convert_body_to_auction_request(body: &AdRequest, settings: &Settings) -> AuctionRequest {
```
prk-Jr, line 456: 🌱 **seedling** — Every conversion test routes through this helper with `Some("existing-ec-id")`, so the `ec_id: None` → `user.id == None` propagation (line 91 / 185 in production) is unpinned. One small test would cover it:

```rust
#[test]
fn convert_tsjs_to_auction_request_propagates_missing_ec_id() {
    let settings = make_settings();
    let req = make_request();
    let auction_request = convert_tsjs_to_auction_request(
        &make_banner_body(None),
        &settings,
        &req,
        ConsentContext::default(),
        None,
    )
    .expect("should convert request without EC ID");

    assert!(
        auction_request.user.id.is_none(),
        "should leave user ID unset when caller provides no EC ID"
    );
}
```

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -482,4 +567,480 @@ mod tests {
             "should omit x-ts-ec when no EC ID is available"
         );
     }
+
+    #[test]
+    fn convert_tsjs_to_auction_request_maps_banner_sizes_to_formats() {
```
prk-Jr, line 567: 🌱 **seedling** — A banner unit with empty `sizes: vec![]` passes the `len() == 2` check vacuously and produces a slot with zero formats. Whether downstream providers handle an empty-formats slot gracefully is unverified. Since this PR pins malformed-size rejection and non-banner skipping, the zero-sizes boundary is the remaining gap — a documenting test (slot created with empty `formats`, or a decision to reject) would complete the set.

crates/trusted-server-core/src/host_rewrite.rs
```diff
@@ -47,3 +47,135 @@ pub(crate) fn rewrite_bare_host_at_boundaries(
     out.push_str(&text[search..]);
     Some(out)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const ORIGIN_HOST: &str = "origin.example.com";
```
prk-Jr, line 55: 🌱 **seedling** — Case sensitivity is unpinned: `str::find` is case-sensitive, so `ORIGIN.EXAMPLE.COM` (or `Origin.Example.com` in markup) is never rewritten even though hostnames are case-insensitive per RFC 4343. That may be acceptable for `__next_f`/Flight payloads where hosts are machine-emitted lowercase, but a one-line documenting test would make the limitation explicit:

```rust
#[test]
fn does_not_rewrite_differently_cased_host() {
    assert_no_rewrite(
        "ORIGIN.EXAMPLE.COM/news",
        "should not rewrite differently-cased host occurrences",
    );
}
```

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -482,4 +567,480 @@ mod tests {
...
+        );
+        assert_eq!(
+            json["ext"]["orchestrator"]["total_bids"],
+            json!(0),
+            "should report zero total bids"
+        );
+    }
+
+    #[test]
+    fn convert_to_openrtb_response_serializes_multiple_winning_bids() {
```
prk-Jr, line 946: 👍 **praise** — Locating seatbids by `impid` lookup instead of array index makes these assertions immune to `HashMap` iteration order in `winning_bids` — exactly the kind of test that would otherwise flake intermittently.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in d74f870: simplified the test helper to pass the user-agent as a plain `&str` and removed the now-unused `HeaderValue` import.

crates/trusted-server-core/src/error.rs
ChristianPavilonis: Addressed in d74f870: replaced the category-specific wording with the broader "Selected client errors... all other errors..." doc text to avoid future drift.

crates/trusted-server-core/src/error.rs
ChristianPavilonis: Thanks — agreed this is better handled as a follow-up behavior change. I left the current generic 403/404 behavior pinned in this coverage PR.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in d74f870: added `convert_tsjs_to_auction_request_propagates_missing_ec_id` to assert `ec_id: None` leaves `user.id` unset.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in d74f870: added a documenting test for `sizes: vec![]` that pins the current behavior of creating one slot with empty `formats`.

crates/trusted-server-core/src/host_rewrite.rs
ChristianPavilonis: Addressed in d74f870: added `does_not_rewrite_differently_cased_host` to document the current case-sensitive matching behavior.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Thanks — kept the order-insensitive `impid` lookup assertions in place.

crates/trusted-server-core/src/tsjs.rs
```diff
@@ -64,3 +64,159 @@ pub fn tsjs_deferred_script_tags(module_ids: &[&str]) -> String {
...
+
+        assert!(
+            src.starts_with("/static/tsjs=tsjs-prebid.min.js?v="),
+            "should use per-module static bundle path"
+        );
+        assert_sha256_hex_hash(hash_query_value(&src));
+    }
+
+    #[test]
+    fn tsjs_deferred_script_src_uses_empty_hash_for_unknown_module() {
```
aram356, line 182: 🌱 **seedling** — This pins an unknown module ID → `?v=` (empty hash) → a `tsjs-unknown-module.min.js` path that won't resolve. The test correctly documents *current* behavior, but the underlying choice to silently produce a broken URL (rather than reject an unknown module) is worth a follow-up issue. Tracking-only; no change needed in this PR.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -374,21 +384,91 @@ mod tests {
...
+        Bid {
+            slot_id: slot_id.to_string(),
+            price,
+            currency: "USD".to_string(),
+            creative: Some("<div>Ad</div>".to_string()),
+            adomain: Some(vec!["advertiser.example.com".to_string()]),
+            bidder: bidder.to_string(),
+            width: 300,
+            height: 250,
+            nurl: None,
```
aram356, line 405: 🌱 **seedling** — `make_bid` always sets `nurl: None` / `burl: None`, so the win/billing-notice serialization branch in `convert_to_openrtb_response` stays uncovered. A future test setting these to `Some(...)` would lock that wire output. Low priority.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -338,16 +337,27 @@ pub fn convert_to_openrtb_response(
...
+    use std::collections::HashSet;
+
+    fn make_request() -> Request {
+        let mut req = Request::new("POST", "https://publisher.example.com/auction");
+        req.set_header(header::USER_AGENT, "Mozilla/5.0 test");
+        req
+    }
 
-    fn make_minimal_auction_request() -> AuctionRequest {
+    fn make_settings() -> Settings {
```
aram356, line 352: ⛏ **nitpick** — `make_settings()` is now a one-line passthrough to `create_test_settings()`. Could be inlined at call sites, or keep it if you expect per-test customization. Trivial.

aram356 (approved):
## Summary

Re-review of the production-readiness test-coverage PR. The diff is now test-only plus two doc/comment corrections — verified that `convert_to_openrtb_response`'s body is unchanged, so the "no runtime behavior change" claim holds. All three blocking items from the 2026-06-03 review are resolved, and every test assertion was cross-checked against the production source it pins. CI is fully green on the head commit. Approving.

## Prior blocking items — resolved

- Duplicate `mod tests` + stale 7-arg auction API → single test module on the current 5-arg signature (`no fresh_id`, `AdRequest.eids` present, asserts `HEADER_X_TS_EC_CONSENT`).
- Undisclosed runtime change (explicit empty `seatbid`/`adomain` via a second serialization pass) → reverted; tests now characterize the existing omit-when-empty wire format as spec.
- Exhaustive error guard → updated to current variants (`EdgeCookie`/`PartnerNotFound`/`InsecureDefault`, no `Ec`).

## Non-blocking

### 🌱 seedling
- **Unknown-module empty hash**: an unknown module ID yields a `?v=` (empty hash) URL to a non-existent bundle. Test documents it; worth a follow-up on whether to reject instead. (`tsjs.rs:182`)
- **`nurl`/`burl` uncovered**: always `None` in `make_bid`, so their response-serialization branch is untested. (`auction/formats.rs:405`)

### ⛏ nitpick
- **`make_settings()` passthrough**: one-line wrapper over `create_test_settings()`; could be inlined. (`auction/formats.rs:352`)

## CI Status

- cargo fmt: PASS
- cargo clippy: PASS
- cargo test: PASS
- vitest: PASS
- integration tests: PASS

prk-Jr (approved):

PR #689, opened 2026-05-13
Add route_request dispatch coverage

### LINKED ISSUE #451: Add integration coverage for route_request dispatch behavior
## Description

`route_request` in `crates/fastly/src/main.rs` owns request dispatch for static assets, admin endpoints, auction paths, and publisher fallback behavior, but it currently lacks direct integration coverage.

## Done when

- [ ] Add integration tests that exercise the main route branches through the Fastly entry path.
- [ ] Verify admin routes reject unauthenticated requests.
- [ ] Verify unknown routes fall back to publisher proxy behavior.
- [ ] Keep this issue scoped to route dispatch coverage rather than full frontend/container setup.

## Affected area

- Fastly runtime

## Additional context

Source: `test_gap.md`, section `Route Request Logic (Integration Test Required)`.
Parent issue: #396
Related: #398, #400

ChristianPavilonis:
## Summary

- Add focused Fastly adapter `route_request` dispatch coverage.
- Cover static TSJS success and unknown bundle handling.
- Cover public discovery and unauthenticated admin dispatch.
- Cover auction dispatch through consent-dependent setup.
- Cover unknown-route publisher fallback dispatch without configured consent persistence.

Closes #451

## Tests

- `cargo test -p trusted-server-adapter-fastly route_tests -- --nocapture`
- `cargo test --workspace`

### DIFF
--- a/crates/trusted-server-adapter-fastly/src/route_tests.rs
+++ b/crates/trusted-server-adapter-fastly/src/route_tests.rs
@@ -683,6 +683,24 @@ fn route_buffered_response(
     route_result_to_fastly_response(settings, services, &partner_registry, route_result)
 }
 
+fn route_with_settings(
+    settings: &Settings,
+    req: Request,
+    expect_message: &str,
+) -> fastly::Response {
+    let (orchestrator, integration_registry) = build_route_stack(settings);
+    let services = test_runtime_services(&req);
+
+    route_buffered_response(
+        settings,
+        &orchestrator,
+        &integration_registry,
+        &services,
+        req,
+        expect_message,
+    )
+}
+
 fn valid_banner_ad_unit_body() -> Vec<u8> {
     serde_json::to_vec(&json!({
         "adUnits": [
@@ -697,6 +715,94 @@ fn valid_banner_ad_unit_body() -> Vec<u8> {
         ]
     }))
     .expect("should serialize valid auction route test body")
+}
+
+#[test]
+fn static_tsjs_route_serves_unified_bundle() {
+    let settings = create_test_settings();
+    let req = Request::get("https://test.com/static/tsjs=tsjs-unified.min.js");
+
+    let mut response = route_with_settings(&settings, req, "should route static tsjs request");
+
+    assert_eq!(
+        response.get_status(),
+        StatusCode::OK,
+        "should serve the unified static bundle"
+    );
+    assert_eq!(
+        response.get_header_str(header::CONTENT_TYPE),
+        Some("application/javascript; charset=utf-8"),
+        "should serve the unified bundle as JavaScript"
+    );
+    assert!(
+        response.take_body_str().contains("requestAds"),
+        "should serve unified bundle content with the core requestAds API"
+    );
+}
+
+#[test]
+fn static_tsjs_route_returns_not_found_for_unknown_tsjs_bundle() {
+    let settings = create_test_settings();
+    let req = Request::get("https://test.com/static/tsjs=tsjs-doesnotexist.min.js");
+
+    let response = route_with_settings(&settings, req, "should route static tsjs request");
+
+    assert_eq!(
+        response.get_status(),
+        StatusCode::NOT_FOUND,
+        "should let the static tsjs branch own unknown bundle paths"
+    );
+}
+
+#[test]
+fn unknown_route_falls_back_to_publisher_proxy_path() {
+    let settings = create_test_settings();
+    let (orchestrator, integration_registry) = build_route_stack(&settings);
+
+    let req = Request::get("https://test.com/articles/example");
+    let http_client = Arc::new(RecordingHttpClient::new(StatusCode::BAD_GATEWAY));
+    let services = test_runtime_services_with_http_client(
+        &req,
+        Arc::new(FixedBackend),
+        Arc::clone(&http_client) as Arc<dyn PlatformHttpClient>,
+    );
+
+    let response = route_buffered_response(
+        &settings,
+        &orchestrator,
+        &integration_registry,
+        &services,
+        req,
+        "should route publisher fallback",
+    );
+
+    assert_eq!(
+        response.get_status(),
+        StatusCode::BAD_GATEWAY,
+        "should return the publisher origin response through fallback routing"
+    );
+    let calls = http_client
+        .calls
+        .lock()
+        .expect("should lock recorded calls");
+    assert_eq!(
+        calls.len(),
+        1,
+        "should send exactly one publisher fallback request"
+    );
+    assert_eq!(
+        calls[0].method,
+        Method::GET,
+        "should preserve the fallback request method"
+    );
+    assert_eq!(
+        calls[0].backend_name, "https-origin.test-publisher.com",
+        "should dispatch to the publisher origin backend"
+    );
+    assert_eq!(
+        calls[0].uri, "https://origin.test-publisher.com/articles/example",
+        "should send the fallback request to the publisher origin"
+    );
 }
 
 #[test]


crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -178,71 +212,143 @@ fn test_runtime_services(req: &Request) -> RuntimeServices {
...
+}
+
+#[test]
+fn auction_route_dispatches_to_consent_dependent_path() {
+    let settings = create_test_settings();
+    let req = Request::post("https://test.com/auction").with_body(r#"{"adUnits":[]}"#);
+
+    let resp = route_with_settings(&settings, req).expect("should route auction request");
+
+    assert_eq!(
```
prk-Jr, line 1624: 🔧 **wrench** — Auction route test can pass through the fallback path

`auction_route_dispatches_to_consent_dependent_path` asserts `503` with a missing consent store. That does not prove `/auction` dispatched to `handle_auction`, because the unknown-route publisher fallback also calls `runtime_services_for_consent_route` first and would return the same `503` if the `/auction` matcher were removed or typoed.

**Fix**: use a route-specific auction failure that fallback would not produce, for example settings without a consent store plus an invalid banner size:

```rust
let settings = create_test_settings_without_consent_store();
let req = Request::post("https://test.com/auction").with_body(
    r#"{"adUnits":[{"code":"slot","mediaTypes":{"banner":{"sizes":[[300]]}},"bids":[]}],"config":null}"#,
);

let mut resp = route_with_settings(&settings, req).expect("should route auction request");

assert_eq!(resp.get_status(), StatusCode::BAD_REQUEST);
assert!(resp.take_body_str().contains("Invalid banner size"));
```

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -178,71 +212,143 @@ fn test_runtime_services(req: &Request) -> RuntimeServices {
...
+        resp.get_status(),
+        StatusCode::SERVICE_UNAVAILABLE,
+        "should reach the auction route and fail opening configured consent persistence"
+    );
+}
+
+#[test]
+fn unknown_route_falls_back_to_publisher_proxy_path() {
+    let settings = create_test_settings_without_consent_store();
```
prk-Jr: 🤔 **thinking** — Publisher fallback test depends on ambient DNS failure

`unknown_route_falls_back_to_publisher_proxy_path` gets `502` because the configured origin cannot be resolved; the local run emitted a Viceroy DNS error. To make the failure mode intentional, consider a fallback-specific settings helper with an RFC-reserved `.invalid` origin host.

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -178,71 +212,143 @@ fn test_runtime_services(req: &Request) -> RuntimeServices {
         .build()
 }
 
+#[test]
+fn static_tsjs_route_serves_unified_bundle() {
```
prk-Jr, line 721: 🌱 **seedling** — Static route coverage misses deferred module success

The static route tests cover unified bundle success and unknown bundle failure. Since this config enables Prebid, adding `/static/tsjs=tsjs-prebid.min.js` would also cover the deferred-module branch in `handle_tsjs_dynamic`.

prk-Jr (changes requested):
## Summary

This PR adds focused Fastly adapter coverage for `route_request`. The overall direction is useful, but one assertion currently does not prove the route branch it is intended to cover because another branch can produce the same failure status.

## CI Status

- cargo fmt: PASS
- cargo test: PASS
- vitest: PASS
- integration tests: PASS
- CodeQL: PASS
- analysis jobs: PASS

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: Fixed in 3f7d3333. The auction dispatch test now uses settings without a consent store and asserts the route-specific invalid banner size `400` response/body, so the publisher fallback path would not satisfy the assertion.

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: Fixed in 3f7d3333. I added a test-only settings helper that allows overriding `origin_url`, and this fallback test now uses `https://publisher-origin.invalid` so the proxy failure is intentional.

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: Thanks. I am going to leave this one out for now: the adapter-level static route tests already cover successful `/static/tsjs=*` dispatch and not-found handling, while deferred module behavior is covered closer to the publisher/bundle logic. I would prefer not to add a Prebid-specific adapter test in this focused route dispatch PR.

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -179,73 +223,122 @@ fn test_runtime_services(req: &Request) -> RuntimeServices {
...
     );
+    assert!(
+        resp.get_header_str(header::WWW_AUTHENTICATE).is_some(),
+        "should advertise the Basic auth challenge"
+    );
+}
+
+#[test]
+fn auction_route_dispatches_to_auction_handler() {
+    let settings = create_test_settings_without_consent_store();
```
aram356: 🌱 **seedling** — Lost regression coverage for "auction returns 503 when consent_store is misconfigured"

The replaced test asserted that with `consent_store = "missing-consent-store"`, `/auction` returns `503 SERVICE_UNAVAILABLE`. Switching to `create_test_settings_without_consent_store()` is a strict improvement for proving handler dispatch (great `"Invalid banner size"` assertion), but it drops coverage that `runtime_services_for_consent_route` actually **gates** the auction route.

`runtime_services_for_consent_route` is invoked at two call sites: `main.rs:236` (auction) and `main.rs:273` (publisher fallback). Only the fallback gate is now under test (via `publisher_fallback_uses_consent_dependent_services`). A future change that removed the gate from the auction branch would slip through CI.

**Fix** — add a small companion test:

```rust
#[test]
fn auction_route_enforces_consent_store_when_configured() {
    let settings = create_test_settings();
    let req = Request::post("https://test.com/auction").with_body(r#"{\"adUnits\":[]}"#);

    let resp = route_with_settings(&settings, req).expect("should route auction request");

    assert_eq!(
        resp.get_status(),
        StatusCode::SERVICE_UNAVAILABLE,
        "should fail auction when consent persistence is configured but unavailable",
    );
}
```

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -118,7 +118,28 @@ impl PlatformGeo for NoopGeo {
...
 fn create_test_settings() -> Settings {
-    let settings = Settings::from_toml(
+    create_test_settings_with_consent_store(Some("missing-consent-store"))
+}
+
+fn create_test_settings_without_consent_store() -> Settings {
+    create_test_settings_with_consent_store(None)
+}
+
+fn create_test_settings_with_consent_store(consent_store: Option<&str>) -> Settings {
```
aram356: ♻️ **refactor** — Collapse the four-layer settings helper into two

The constructor chain currently grows by suffix concatenation:

```
create_test_settings()
create_test_settings_without_consent_store()
create_test_settings_with_consent_store(opt)
create_test_settings_with_consent_store_and_origin_url(opt, url)
```

The middle layer (`create_test_settings_with_consent_store`) has exactly two callers, both internal. The naming pattern will keep degrading when a third axis is added (e.g. `..._and_handlers`). Two helpers cover every existing call site cleanly:

```rust
fn create_test_settings() -> Settings {
    create_test_settings_with(Some("missing-consent-store"), "https://origin.test-publisher.com")
}

fn create_test_settings_without_consent_store() -> Settings {
    create_test_settings_with(None, "https://origin.test-publisher.com")
}

fn create_test_settings_with(consent_store: Option<&str>, origin_url: &str) -> Settings {
    // existing body of create_test_settings_with_consent_store_and_origin_url
}
```

Non-blocking; purely an ergonomics nudge before a third axis arrives.

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -179,73 +223,122 @@ fn test_runtime_services(req: &Request) -> RuntimeServices {
...
+
+#[test]
+fn unknown_route_falls_back_to_publisher_proxy_path() {
+    let settings = create_test_settings_with_consent_store_and_origin_url(
+        None,
+        "https://publisher-origin.invalid",
+    );
+    let req = Request::get("https://test.com/articles/example");
+
+    let resp = route_with_settings(&settings, req).expect("should route publisher fallback");
```
aram356: 🤔 **thinking** — `502` is consistent with `TrustedServerError::Proxy` but not exclusive to it

The `.invalid` change is a real improvement over the prior ambient-DNS reliance (RFC 2606 reserves `.invalid` precisely so it never resolves; local stderr confirms `hyper::Error(Connect, ConnectError("dns error", …))`). That said, five distinct error variants map to `BAD_GATEWAY` in `error.rs:104-121` (`Auction`, `Gam`, `Prebid`, `Integration`, `Proxy`), so a future regression that surfaces, say, an `Integration` 502 on this route would still pass the assertion silently.

A one-line intent comment is the cheapest mitigation:

```rust
// 502 here is specifically Proxy{} from req.send() against the .invalid backend;
// no other 502-mapping error fires on the publisher fallback for /articles/example.
assert_eq!(
    resp.get_status(),
    StatusCode::BAD_GATEWAY,
    ...
);
```

No-action is also reasonable — flagging this so the reasoning is visible to whoever next reads the test.

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -179,73 +223,122 @@ fn test_runtime_services(req: &Request) -> RuntimeServices {
...
-        "should keep discovery available when the consent store is unavailable"
+        "should serve the unified static bundle"
     );
+    assert_eq!(
+        resp.get_header_str(header::CONTENT_TYPE),
+        Some("application/javascript; charset=utf-8"),
+        "should serve the unified bundle as JavaScript"
+    );
+    assert!(
+        !resp.take_body_str().is_empty(),
```
aram356: ⛏ **nitpick** — `!is_empty()` is satisfied by any non-empty body

Would also pass if the bundle were ever replaced by an error page that happens to be JS-typed. `handle_tsjs_dynamic` already has direct unit coverage in `publisher.rs:1230-1343`, so this adapter-level test is really a *dispatch* check; asserting one canonical token (a known function name from the core bundle) would tighten the contract slightly without re-testing bundling logic. Not load-bearing — fine to leave as-is.

aram356 (changes requested):
## Summary

Decomposing the previous monolithic test into focused per-route dispatch tests is a clean improvement, and the prior blocking concerns from @prk-Jr (auction passing through fallback / ambient DNS dependency) are fully addressed in `3f7d3333` — the `"Invalid banner size"` body assertion and the `.invalid` origin URL both make the intended failure modes unambiguous. CI is green and the new tests pass locally.

Requesting changes on one point only: switching `auction_route_dispatches_to_auction_handler` to `consent_store = None` proves handler dispatch but **drops** the regression assertion that `runtime_services_for_consent_route` actually gates the `/auction` route. Only the publisher-fallback gate is now under test. A small companion test restores parity with the original coverage — see the inline 🌱.

## Non-blocking

### 🌱 seedling

- **Lost regression coverage for "auction returns 503 when consent_store is misconfigured"** — `route_tests.rs:296` (inline)

### ♻️ refactor

- **Collapse the four-layer settings helper into two** — `route_tests.rs:128` (inline)

### 🤔 thinking

- **`.invalid` 502 is consistent with `Proxy` but not exclusive to it** — `route_tests.rs:322` (inline)

### ⛏ nitpick

- **`!is_empty()` is satisfied by any non-empty body** — `route_tests.rs:243` (inline)

## CI Status

- cargo fmt: **PASS**
- cargo clippy: **PASS** (verified locally on `trusted-server-adapter-fastly`)
- cargo test: **PASS** (verified locally — 7/7 new tests pass)
- vitest: **PASS**
- format-docs / format-typescript: **PASS**
- browser integration tests / integration tests: **PASS**
- CodeQL: **PASS**

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: Fixed in 055ce1f9. Added `auction_route_enforces_consent_store_when_configured` using `create_test_settings()` and asserting `/auction` returns `503 SERVICE_UNAVAILABLE` when configured consent persistence is unavailable. The separate invalid-banner test remains in place for route-specific auction handler dispatch coverage.

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: Fixed in 055ce1f9. Collapsed the helper chain to `create_test_settings()`, `create_test_settings_without_consent_store()`, and `create_test_settings_with(consent_store, origin_url)`, then updated the `.invalid` fallback setup to use the parameterized helper.

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: Fixed in 055ce1f9. Added an intent comment above the `BAD_GATEWAY` assertion explaining that the `.invalid` publisher origin is intended to drive the publisher proxy `req.send()` failure and that no other 502-mapping branch should run for `/articles/example` in this test.

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: Fixed in 055ce1f9. The test now stores the bundle body and asserts it contains the stable core `requestAds` API token instead of only checking for any non-empty body.

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -162,6 +173,32 @@ fn create_test_settings() -> Settings {
     settings
 }
 
+fn build_route_stack(
```
aram356: ♻️ **refactor** — `build_route_stack` has exactly one caller (`route_with_settings` at line 190), so the tuple-returning seam adds indirection without reuse. Inlining its two lines into `route_with_settings` reads cleaner — the same single-caller helper smell the team already collapsed away in the settings helpers this round.

```rust
fn route_with_settings(settings: &Settings, req: Request) -> Option<Response> {
    let orchestrator = build_orchestrator(settings).expect("should build auction orchestrator");
    let integration_registry =
        IntegrationRegistry::new(settings).expect("should create integration registry");
    let runtime_services = test_runtime_services(&req);

    futures::executor::block_on(route_request(
        settings,
        &orchestrator,
        &integration_registry,
        &runtime_services,
        req,
    ))
}
```

Non-blocking.

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -162,6 +173,32 @@ fn create_test_settings() -> Settings {
...
+    settings: &Settings,
+) -> (
+    trusted_server_core::auction::AuctionOrchestrator,
```
aram356: ⛏ **nitpick** — The return type spells out `trusted_server_core::auction::AuctionOrchestrator` while `build_orchestrator` and `IntegrationRegistry` are already imported and used short. For consistency, fold it into the existing import:

```rust
use trusted_server_core::auction::{build_orchestrator, AuctionOrchestrator};
```

(Moot if `build_route_stack` is inlined per the refactor note above.)

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -179,73 +216,137 @@ fn test_runtime_services(req: &Request) -> RuntimeServices {
...
+    assert!(
+        body.contains("requestAds"),
+        "should serve unified bundle content with the core requestAds API"
+    );
+}
+
+#[test]
+fn static_tsjs_route_returns_not_found_for_unknown_bundle() {
+    let settings = create_test_settings();
+    let req = Request::get("https://test.com/static/tsjs=unknown.js");
```
aram356: 🌱 **seedling** — `unknown.js` has no `tsjs-` prefix, so it trips the *early* `NOT_FOUND` return in `handle_tsjs_dynamic` (`publisher.rs:129`) rather than the more interesting "`tsjs-`-prefixed but unknown/disabled module" branches (`publisher.rs:149` / `publisher.rs:161`). A `tsjs-`-prefixed unknown name (e.g. `tsjs-doesnotexist.min.js`) would prove the dispatch contract that actually matters. The test name is also slightly imprecise — this is really "non-bundle filename" rather than "unknown bundle". Non-blocking.

aram356 (approved):
## Summary

Re-review of `055ce1f9`. All prior feedback from both earlier rounds is resolved: auction dispatch now proves the route via the invalid-banner `400`, the `.invalid` origin makes the publisher-fallback `502` intentional (with an intent comment), consent-store gating is now verified at both call sites, and the settings helpers are collapsed to a single parameterized factory. I read the full test file and the `route_request` path, verified every assertion against the production status-code mappings, and ran the suite — 8/8 pass (the `hyper` DNS error logged by `unknown_route_falls_back_to_publisher_proxy_path` is the intended `.invalid`-origin failure that yields the `502`).

Approving. Remaining items are all non-blocking polish/coverage — see inline comments plus the out-of-scope note below.

## Non-blocking

### 📌 out of scope

- **Untested `route_request` branches** — `/verify-signature` (POST), the `/first-party/*` routes, integration-proxy dispatch via `has_route`, and the `enforce_basic_auth` `Err` arm are not exercised. Reasonable to defer given the focused scope of #451, but worth a follow-up so the gap stays visible.

## CI Status

- cargo fmt: PASS
- cargo clippy: PASS
- cargo test: PASS (verified locally — 8/8 route_tests pass on 055ce1f9)

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: Fixed in 73887d8d. Inlined `build_route_stack` into `route_with_settings` so the setup now lives at the only call site.

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: Moot after 73887d8d. `build_route_stack` was removed, so the explicit `trusted_server_core::auction::AuctionOrchestrator` return type is gone too.

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: Fixed in 73887d8d. Renamed the test to `static_tsjs_route_returns_not_found_for_unknown_tsjs_bundle` and switched the request to `/static/tsjs=tsjs-doesnotexist.min.js` so it exercises the `tsjs-`-prefixed unknown bundle path.

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -179,73 +205,137 @@ fn test_runtime_services(req: &Request) -> RuntimeServices {
...
+    assert!(
+        resp.take_body_str().contains("Invalid banner size"),
+        "should return the auction validation error"
+    );
+}
+
+#[test]
+fn auction_route_enforces_consent_store_when_configured() {
+    let settings = create_test_settings();
+    let req = Request::post("https://test.com/auction").with_body(r#"{"adUnits":[]}"#);
```
aram356: ⛏ **nitpick** — Inert request body reads as if it tests empty ad units

The `{"adUnits":[]}` body looks like it exercises empty-ad-unit handling, but the `503` comes purely from the consent gate failing in `runtime_services_for_consent_route` *before* the auction body is ever parsed (main.rs:236). A one-line comment prevents a future reader from over-reading it:

```rust
// Body is irrelevant: the consent gate (runtime_services_for_consent_route)
// fails before the auction handler parses the request.
let req = Request::post("https://test.com/auction").with_body(r#"{"adUnits":[]}"#);
```

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -179,73 +205,137 @@ fn test_runtime_services(req: &Request) -> RuntimeServices {
...
+        resp.get_status(),
         StatusCode::SERVICE_UNAVAILABLE,
-        "should fail auction requests when consent persistence is configured but unavailable"
+        "should fail auction when consent persistence is configured but unavailable"
     );
+}
+
+#[test]
+fn unknown_route_falls_back_to_publisher_proxy_path() {
+    let settings = create_test_settings_with(None, "https://publisher-origin.invalid");
```
aram356: 📝 **note** — Expected stderr noise, no action needed

This test intentionally prints `hyper::Error(Connect, ConnectError("dns error", ...))` to stderr on every run because the `.invalid` origin never resolves. It's harmless and the inline comment below already explains the intent — just flagging that the scary-looking error line in `cargo test` output is expected, not a failure.

aram356 (changes requested):
## Summary

Re-review of the current head (`73887d8d`) after the post-approval polish commit. The test design is sound — every assertion traces to real dispatch behavior, and the route-specific assertions correctly discriminate dispatch from the publisher fallback. Requesting changes only to flag a CI gating gap before merge; the remaining items are non-blocking.

## Blocking

### 🔧 wrench

- **Rust CI never ran on the current head**: `Run Tests`, `Run Format`, and `Integration Tests` last ran on commit `3f7d3333` (2026-05-15). The two commits after it — `055ce1f9` and `73887d8d` (post-approval polish) — only carry CodeQL's `Analyze (javascript-typescript)` check. For an entirely-Rust PR, the merge gate ("fmt/clippy/test pass") is not actually satisfied by CI for the code that would merge (likely the external-contributor "approve and run workflows" gate). Please re-trigger the Rust workflows on `73887d8d` so the green checks reflect the merged commit. _Verified locally on head: `cargo fmt --all -- --check` clean, `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean, all 8 route tests pass._

## Non-blocking

### ⛏ nitpick

- **Inert auction body** (route_tests.rs:301): the `503` is from the consent gate, not the empty `adUnits`; a clarifying comment would help. See inline.

### 📝 note

- **Expected stderr DNS error** (route_tests.rs:313-320): the `.invalid` origin prints a `hyper` connect error on every run; harmless and intentional. See inline.

### 📌 out of scope

- **Misplaced `JWKS_CONFIG_STORE_NAME` assertion** (route_tests.rs:168-171, pre-existing on `main` — not touched by this PR): the `assert_eq!(JWKS_CONFIG_STORE_NAME, "jwks_store", "...keep the stub discovery store aligned...")` lives inside the settings factory, so it re-runs on every settings construction, and it doesn't actually couple to the stub — `StubJwksConfigStore::get` ignores `_store_name` and matches on `key`, so the discovery test would still pass if the constant changed. Worth a follow-up: move it to a dedicated `#[test]`, or make the stub assert/match the store name so the check becomes load-bearing.

## CI Status (verified locally on head `73887d8d`)

- fmt: PASS (local) — not run on head in CI
- clippy: PASS (local) — not run on head in CI
- rust tests: PASS (8/8 local) — not run on head in CI
- js analyze (CodeQL): PASS (CI)

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -637,6 +655,58 @@ fn valid_banner_ad_unit_body() -> Vec<u8> {
...
+        response.get_status(),
+        StatusCode::NOT_FOUND,
+        "should let the static tsjs branch own unknown bundle paths"
+    );
+}
+
+#[test]
+fn unknown_route_falls_back_to_publisher_proxy_path() {
+    let mut settings = create_test_settings();
+    settings.publisher.origin_url = "https://publisher-origin.invalid".to_string();
```
aram356: 🔧 **wrench** — This test no longer verifies what its name claims; the `.invalid` origin is inert after the merge

Setting `origin_url = "https://publisher-origin.invalid"` to force the failure is an idiom from the pre-merge version, where the publisher fallback used Fastly `req.send()` and `.invalid` genuinely caused a DNS failure. In the merged (request-local consent / platform-http-client) architecture it no longer drives the failure:

- `handle_publisher_request` calls `services.backend().ensure(...)` (publisher.rs:496-507) **before** any origin send;
- default test services use `NoopBackend`, whose `ensure()` returns `Err(Unsupported)` → `Proxy{"backend registration failed"}` → `502`;
- running this test with `--nocapture` produces **no** DNS/connect error (the pre-merge version always emitted `hyper ConnectError("dns error")`), confirming `.invalid` is never dialed.

So the `502` comes from the Noop stub, not from reaching the publisher proxy — the assertion can't distinguish "fell back to publisher proxy" from any all-Noop failure, and the `.invalid` setup is misleading.

**Fix** — mirror the asset tests: inject `FixedBackend` + a `RecordingHttpClient` and assert exactly one call to the publisher origin backend/uri, which proves the fallback actually dispatched to the publisher proxy:

```rust
let req = Request::get("https://test.com/articles/example");
let http_client = Arc::new(RecordingHttpClient::new(StatusCode::BAD_GATEWAY));
let services = test_runtime_services_with_http_client(
    &req,
    Arc::new(FixedBackend),
    Arc::clone(&http_client) as Arc<dyn PlatformHttpClient>,
);
let resp = route_buffered_response(&settings, &orchestrator, &integration_registry, &services, req, "should route publisher fallback");
assert_eq!(resp.get_status(), StatusCode::BAD_GATEWAY, "...");
let calls = http_client.calls.lock().expect("should lock recorded calls");
assert_eq!(calls.len(), 1, "should reach the publisher origin via fallback");
assert!(calls[0].backend_name.contains("origin.test-publisher.com"), "should dispatch to the publisher origin, got {}", calls[0].backend_name);
```

Reference: main's own `asset_routes_skip_non_get_head_requests` (route_tests.rs:1329) already forces a deterministic publisher-fallback `502` (via `origin_url = "https://"` → missing-host parse error) and asserts `calls.is_empty()`.

aram356 (changes requested):
## Summary

Re-review after the `main` merge (`56d29cf4`). The branch now fully integrates current `main` (request-local consent, `RouteResult`/`HandlerOutcome`, platform-http-client proxying), and the net contribution over main is 3 tests + a `route_with_settings` helper. The merge correctly retired the obsolete consent-KV-gating tests. One test needs strengthening before merge; the previously-flagged CI gap is now resolved.

## Blocking

### 🔧 wrench

- **`unknown_route_falls_back_to_publisher_proxy_path` doesn't verify what it claims** (route_tests.rs:695-708): the `.invalid` origin is inert in the merged architecture — the `502` comes from `NoopBackend.ensure()` ("backend registration failed"), not from reaching the publisher origin (verified: no DNS/connect error under `--nocapture`, unlike the pre-merge version). The assertion can't distinguish a real publisher fallback from any all-Noop failure. See inline for a `FixedBackend` + `RecordingHttpClient` fix that proves dispatch.

## Non-blocking

### 📝 note

- **Prior CI gap resolved**: the merge re-triggered full Rust CI. `cargo fmt`, `cargo test`, `Analyze (rust)`, `vitest`, and integration tests all pass on head `56d29cf4`. Re-verified locally: 20/20 route tests pass, fmt clean.

## CI Status (head `56d29cf4`)

- cargo fmt: PASS (CI + local)
- cargo test: PASS (CI; 20/20 route tests local)
- Analyze (rust): PASS (CI)
- vitest: PASS (CI)
- integration tests: PASS (CI)

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: Fixed in 372e0b61. `unknown_route_falls_back_to_publisher_proxy_path` now keeps the normal publisher origin and injects `FixedBackend` plus `RecordingHttpClient`, then asserts exactly one fallback request reaches `https-origin.test-publisher.com` with URI `https://origin.test-publisher.com/articles/example` and method `GET`. This removes the inert `.invalid`/`NoopBackend` failure mode and proves publisher proxy dispatch directly.

Verified:
- `cargo test -p trusted-server-adapter-fastly unknown_route_falls_back_to_publisher_proxy_path`
- `cargo test -p trusted-server-adapter-fastly route_tests`
- `cargo test --workspace`

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: No code change needed here after the main merge: the obsolete consent-store gating test was retired with the request-local consent architecture, so the inert `{"adUnits":[]}` body is no longer present in the current route dispatch tests.

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: No longer applicable after 372e0b61: the fallback test no longer uses a `.invalid` origin. It now injects `FixedBackend` plus `RecordingHttpClient` and asserts the recorded publisher-origin request directly, so the test should not produce the old expected DNS stderr noise.

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -637,6 +655,94 @@ fn valid_banner_ad_unit_body() -> Vec<u8> {
...
+    );
+}
+
+#[test]
+fn unknown_route_falls_back_to_publisher_proxy_path() {
+    let settings = create_test_settings();
+    let (orchestrator, integration_registry) = build_route_stack(&settings);
+
+    let req = Request::get("https://test.com/articles/example");
+    let http_client = Arc::new(RecordingHttpClient::new(StatusCode::BAD_GATEWAY));
```
aram356, line 763: ⛏ **nitpick** (optional, non-blocking) — Returning `BAD_GATEWAY` from the stubbed origin makes the status assertion mildly ambiguous: a `502` is both a plausible origin response *and* the status a proxy failure would produce, so `assert_eq!(status, BAD_GATEWAY)` alone can't distinguish "origin returned 502" from "proxy errored." The new `calls.len()/method/backend_name/uri` assertions fully disambiguate it in practice, so this is purely cosmetic — using a non-error status (e.g. `StatusCode::OK` or `ACCEPTED`, as the sibling `asset_routes_*` tests do) would make "origin response passed through fallback" unambiguous at a glance.

aram356 (approved):
## Summary

Re-review of `372e0b61`. The prior blocking finding — `unknown_route_falls_back_to_publisher_proxy_path` relying on an inert `.invalid` origin that didn't actually exercise the publisher proxy — is resolved. The test now injects `FixedBackend` + `RecordingHttpClient` and asserts the request reached the publisher origin (one call, method preserved, backend `https-origin.test-publisher.com`, URI `https://origin.test-publisher.com/articles/example`). Approving. Two non-blocking notes below.

## Non-blocking

### ⛏ nitpick

- **Stubbed origin returns `BAD_GATEWAY`** (route_tests.rs:701): the `502` status assertion is slightly ambiguous vs a proxy error; the call/backend/uri assertions disambiguate it, so this is cosmetic only. See inline.

### 📝 note

- **Branch is behind main; re-trigger Rust CI before merge**: the branch base is `cbce8b07` but `origin/main` is now `8dc26fa2`, which added DataDome route tests to this same file. Those additions are additive and non-overlapping with this PR's three tests (no redundancy, conflict-free merge expected) — just worth a fresh merge. CI on this head shows only the JS `Analyze` check; the Rust workflows (`cargo test`/`fmt`/clippy) have not re-run on `372e0b61`. Verified locally that they pass.

## CI Status (head `372e0b61`)

- cargo fmt: PASS (local) — not re-run on head in CI
- cargo test: PASS (20/20 route tests, local) — not re-run on head in CI
- clippy: PASS (local) — not re-run on head in CI
- js analyze (CodeQL): PASS (CI)

prk-Jr (approved):

PR #669, opened 2026-04-29
Add the Trusted Server CLI with JS Asset Auditor

ChristianPavilonis:
## Reviewer quick links

Start here if you want to test the new CLI directly:

- CLI guide: [`docs/guide/cli.md`](https://github.com/IABTechLab/trusted-server/blob/feature/ts-cli/docs/guide/cli.md)
- Fastly provisioning map: [`docs/guide/fastly-provisioning.md`](https://github.com/IABTechLab/trusted-server/blob/feature/ts-cli/docs/guide/fastly-provisioning.md)
- Fastly setup guide: [`docs/guide/fastly.md`](https://github.com/IABTechLab/trusted-server/blob/feature/ts-cli/docs/guide/fastly.md)
- Getting started guide: [`docs/guide/getting-started.md`](https://github.com/IABTechLab/trusted-server/blob/feature/ts-cli/docs/guide/getting-started.md)

The most useful review path is:

1. Read the CLI command reference.
2. Read the provisioning map to understand which `trusted-server.toml` changes map to Fastly resources.
3. Follow the manual test guide below against a disposable or test Fastly service.

## Manual reviewer test plan

Use a disposable or test Fastly Compute service. Run the commands one by one so you can inspect each CLI step and its output.

### 1. Set up local CLI access

Install the host-target CLI from the checkout:

```bash
cargo install_cli
```

Confirm the command surface is available:

```bash
ts --help
ts config --help
ts audit --help
ts auth fastly --help
ts provision fastly --help
```

If you are not on Apple Silicon macOS, use the explicit host-target form from [`docs/guide/cli.md`](https://github.com/IABTechLab/trusted-server/blob/feature/ts-cli/docs/guide/cli.md#cargo-aliases) instead of `cargo install_cli`.

### 2. Authenticate to Fastly

For local review, store a Fastly API token in the host credential store:

```bash
ts auth fastly login
```

Inspect credential status in both human and JSON forms:

```bash
ts auth fastly status
ts auth fastly status --json
```

Expected result: `effective_source` is `secure-storage`, unless you set `FASTLY_API_KEY`, in which case it should be `environment`.

For CI-style review, skip `login` and use:

```bash
export FASTLY_API_KEY="your-fastly-token"
ts auth fastly status --json
```

### 3. Audit the test site

Create a temp directory so generated audit files do not overwrite your checkout files:

```bash
WORKDIR="$(mktemp -d)"
echo "$WORKDIR"
```

Run the browser-backed audit against the test site:

```bash
[...]

PR #690, opened 2026-05-13
Add auction endpoint request handling coverage

### LINKED ISSUE #452: Add integration tests for auction endpoint request handling
## Description

`handle_auction` in `crates/common/src/auction/endpoints.rs` has no end-to-end verification for request parsing, timeout behavior, or response shaping.

## Done when

- [ ] Add integration coverage for a valid auction request returning a success response.
- [ ] Add coverage for invalid request payload returning `400`.
- [ ] Add coverage for timeout or orchestrator failure returning the expected gateway timeout/error response.
- [ ] Assert the response contract used by callers stays stable.

## Affected area

- Ad serving (Equativ)

## Additional context

Source: `test_gap.md`, section `Auction Handler`.
Parent issue: #396

ChristianPavilonis:
## Summary

- Add route-level coverage for `/auction` request handling
- Return `400 Bad Request` for malformed auction JSON payloads
- Cover semantic validation failures, orchestration failures, and a stable no-network success path
- Note `/auction` now returns `502 Bad Gateway` when every configured provider is skipped, disabled, or fails to launch instead of returning `200` with an empty `seatbid`.

Closes #452

## Testing

- `cargo test -p trusted-server-adapter-fastly route_tests -- --nocapture`
- `cargo test -p trusted-server-core auction:: -- --nocapture`
- `cargo test --workspace`

### DIFF
--- a/crates/trusted-server-adapter-fastly/src/route_tests.rs
+++ b/crates/trusted-server-adapter-fastly/src/route_tests.rs
@@ -7,12 +7,16 @@ use fastly::http::request::PendingRequest;
 use fastly::http::{header, StatusCode};
 use fastly::Request;
 use serde_json::json;
+<<<<<<< HEAD
 use trusted_server_core::auction::{
     build_orchestrator, AuctionContext, AuctionOrchestrator, AuctionProvider, AuctionRequest,
     AuctionResponse,
 };
 use trusted_server_core::ec::registry::PartnerRegistry;
 use trusted_server_core::error::TrustedServerError;
+=======
+use trusted_server_core::auction::{build_orchestrator, AuctionOrchestrator};
+>>>>>>> 680579cf (Fail auction when no providers launch)
 use trusted_server_core::integrations::IntegrationRegistry;
 use trusted_server_core::platform::{
     ClientInfo, GeoInfo, PlatformBackend, PlatformBackendSpec, PlatformConfigStore, PlatformError,
--- a/crates/trusted-server-adapter-fastly/src/route_tests.rs
+++ b/crates/trusted-server-adapter-fastly/src/route_tests.rs
@@ -3,10 +3,16 @@ use std::sync::Arc;
 
 use edgezero_core::key_value_store::NoopKvStore;
 use error_stack::Report;
-use fastly::http::StatusCode;
+use fastly::http::request::PendingRequest;
+use fastly::http::{header, StatusCode};
 use fastly::Request;
-use trusted_server_core::auction::build_orchestrator;
+use serde_json::json;
+use trusted_server_core::auction::{
+    build_orchestrator, AuctionContext, AuctionOrchestrator, AuctionProvider, AuctionRequest,
+    AuctionResponse,
+};
 use trusted_server_core::ec::registry::PartnerRegistry;
+use trusted_server_core::error::TrustedServerError;
 use trusted_server_core::integrations::IntegrationRegistry;
 use trusted_server_core::platform::{
     ClientInfo, GeoInfo, PlatformBackend, PlatformBackendSpec, PlatformConfigStore, PlatformError,
@@ -118,9 +124,44 @@ impl PlatformGeo for NoopGeo {
     }
 }
 
-fn create_test_settings() -> Settings {
-    let settings = Settings::from_toml(
-        r#"
+struct DisabledRouteProvider;
+
+impl AuctionProvider for DisabledRouteProvider {
+    fn provider_name(&self) -> &'static str {
+        "disabled-route"
+    }
+
+    fn request_bids(
+        &self,
+        _request: &AuctionRequest,
+        _context: &AuctionContext<'_>,
+    ) -> Result<PendingRequest, Report<TrustedServerError>> {
+        Err(Report::new(TrustedServerError::Auction {
+            message: "disabled route provider should not launch requests".to_string(),
+        }))
+    }
+
+    fn parse_response(
+        &self,
+        _response: fastly::Response,
+        _response_time_ms: u64,
+    ) -> Result<AuctionResponse, Report<TrustedServerError>> {
+        Err(Report::new(TrustedServerError::Auction {
+            message: "disabled route provider should not parse responses".to_string(),
+        }))
+    }
+
+    fn timeout_ms(&self) -> u32 {
+        2000
+    }
+
+    fn is_enabled(&self) -> bool {
+        false
+    }
+}
+
+fn base_route_settings_toml() -> &'static str {
+    r#"
             [[handlers]]
             path = "^/_ts/admin"
             username = "admin"
@@ -139,18 +180,32 @@ fn create_test_settings() -> Settings {
             enabled = false
             config_store_id = "test-config-store-id"
             secret_store_id = "test-secret-store-id"
+        "#
+}
 
+fn prebid_integration_toml() -> &'static str {
+    r#"
             [integrations.prebid]
             enabled = true
             server_url = "https://test-prebid.com/openrtb2/auction"
+        "#
+}
+
+fn create_test_settings() -> Settings {
+    let base = base_route_settings_toml();
+    let prebid = prebid_integration_toml();
+    let config = format!(
+        r#"{base}
+
+{prebid}
 
             [auction]
             enabled = true
             providers = ["prebid"]
             timeout_ms = 2000
         "#,
-    )
-    .expect("should parse adapter route test settings");
+    );
+    let settings = Settings::from_toml(&config).expect("should parse adapter route test settings");
 
     assert_eq!(
         JWKS_CONFIG_STORE_NAME, "jwks_store",
@@ -158,6 +213,32 @@ fn create_test_settings() -> Settings {
     );
 
     settings
+}
+
+fn create_auction_test_settings(providers: &str) -> Settings {
+    let base = base_route_settings_toml();
+    let prebid = prebid_integration_toml();
+    let config = format!(
+        r#"{base}
+
+{prebid}
+
+            [auction]
+            enabled = true
+            providers = {providers}
+            timeout_ms = 2000
+        "#,
+    );
+
+    Settings::from_toml(&config).expect("should parse adapter auction route test settings")
+}
+
+fn build_route_stack(settings: &Settings) -> (AuctionOrchestrator, IntegrationRegistry) {
+    let orchestrator = build_orchestrator(settings).expect("should build auction orchestrator");
+    let integration_registry =
+        IntegrationRegistry::new(settings).expect("should create integration registry");
+
+    (orchestrator, integration_registry)
 }
 
 fn test_runtime_services(req: &Request) -> RuntimeServices {
@@ -174,6 +255,54 @@ fn test_runtime_services(req: &Request) -> RuntimeServices {
             tls_cipher: req.get_tls_cipher_openssl_name().map(str::to_string),
         })
         .build()
+}
+
+fn route_auction(settings: &Settings, body: impl Into<Vec<u8>>) -> fastly::Response {
+    let (orchestrator, integration_registry) = build_route_stack(settings);
+
+    route_auction_with_stack(settings, &orchestrator, &integration_registry, body)
+}
+
+fn route_auction_with_stack(
+    settings: &Settings,
+    orchestrator: &AuctionOrchestrator,
+    integration_registry: &IntegrationRegistry,
+    body: impl Into<Vec<u8>>,
+) -> fastly::Response {
+    let partner_registry =
+        PartnerRegistry::from_config(&settings.ec.partners).expect("should build partner registry");
+    let req = Request::post("https://test.com/auction")
+        .with_header(header::CONTENT_TYPE, "application/json")
+        .with_body(body.into());
+    let services = test_runtime_services(&req);
+
+    futures::executor::block_on(route_request(
+        settings,
+        orchestrator,
+        integration_registry,
+        &partner_registry,
+        &services,
+        req,
+    ))
+    .expect("should route auction request")
+    .response
+    .expect("should buffer auction response in tests")
+}
+
+fn valid_banner_ad_unit_body() -> Vec<u8> {
+    serde_json::to_vec(&json!({
+        "adUnits": [
+            {
+                "code": "div-gpt-ad-1",
+                "mediaTypes": {
+                    "banner": {
+                        "sizes": [[300, 250]]
+                    }
+                }
+            }
+        ]
+    }))
+    .expect("should serialize valid auction route test body")
 }
 
 #[test]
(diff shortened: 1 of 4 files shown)


crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -249,3 +329,72 @@ fn configured_missing_consent_store_only_breaks_consent_routes() {
...
+
+    assert_eq!(
+        response.get_status(),
+        StatusCode::BAD_GATEWAY,
+        "should surface no-provider orchestration failures as gateway errors"
+    );
+}
+
+#[test]
+fn valid_auction_request_with_unregistered_provider_returns_bad_gateway() {
```
prk-Jr: 🔧 **wrench** — Missing auction provider should fail startup, not become request-time 502

This test codifies `auction.providers = ["missing-provider"]` as a successful startup followed by `502 Bad Gateway` on `/auction`. That conflicts with the project rule that invalid enabled integrations/providers must not be silently logged and disabled during startup or registration. With the current behavior, a deployment can start with a typo or missing provider config and only fail when traffic hits `/auction`.

**Fix:** validate configured auction providers after registration in `build_orchestrator`, including the mediator if configured, and return a startup/configuration error when a configured name was not registered. Keep the new `No provider requests launched` runtime guard for true launch-time failures, but update this route test to expect startup failure or add a core unit test with a registered fake provider whose `request_bids` returns `Err`.

```rust
if settings.auction.enabled {
    for provider_name in settings
        .auction
        .providers
        .iter()
        .chain(settings.auction.mediator.iter())
    {
        if !registered_provider_names.contains(provider_name) {
            return Err(Report::new(TrustedServerError::Configuration {
                message: format!("Auction provider `{provider_name}` is configured but not registered"),
            }));
        }
    }
}
```

prk-Jr (changes requested):
## Summary

The route coverage is useful, but one new test now codifies a configuration typo as a request-time 502 instead of a startup/configuration failure.

## Blocking

### 🔧 wrench

- **Missing auction provider should fail startup, not become request-time 502**: `auction.providers = ["missing-provider"]` currently builds the route stack and only fails when `/auction` is called. Invalid enabled providers should fail during startup/registration so deployments do not start with a mistyped provider list.

## CI Status

- cargo fmt / clippy: PASS
- rust tests: PASS
- js tests: PASS
- integration tests: PASS

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: I disagree.

I'll make a proposal to remove our misconfiguration fails startup, and move to config validation before config persistence in or after pr #669.

tooling should be in place to prevent these issues, but hosing a production site should probably be unacceptable vs failing with good observability.

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -249,3 +329,72 @@ fn configured_missing_consent_store_only_breaks_consent_routes() {
...
+    let settings = create_auction_test_settings_without_consent_store(r#"["missing-provider"]"#);
+
+    let response = route_auction(&settings, valid_banner_ad_unit_body());
+
+    assert_eq!(
+        response.get_status(),
+        StatusCode::BAD_GATEWAY,
+        "should fail when configured providers cannot be launched"
+    );
+}
```
aram356, line 437: 🔧 **wrench** — This test codifies a configuration typo (`providers = ["missing-provider"]`) as a successful startup followed by a request-time 502. That conflicts with the project rule in `CLAUDE.md`:

> Invalid enabled integrations/providers must not be silently logged-and-disabled during startup or registration.

Reinforcing concerns beyond the prior review:

1. **The new orchestrator guard only catches the all-skipped case.** A mixed config like `providers = ["prebid", "missing-provider"]` would still return 200 with prebid bids only — the orchestrator silently `continue`s for unregistered names at `orchestrator.rs:279-285`. The typo is masked, not caught.
2. **Locking in the disputed behavior with an asserting test will make the #669 follow-up harder** — anyone moving validation to startup has to delete this test, which reads as a regression in a diff.
3. **Inconsistent with existing rule enforcement.** Invalid *config values* (e.g. `prebid.server_url = "not a url"`) already fail startup (`settings.rs:1611-1620`); invalid *names* shouldn't be the lone exception.

**Suggested fix (pick one):**

- **A — validate at startup** (matches the project rule). In `build_orchestrator`, after registering all providers, ensure every name in `settings.auction.providers` and `settings.auction.mediator` exists in `orchestrator.providers`; return `TrustedServerError::Configuration` otherwise. Update this test to assert startup failure.
- **B — drop this single test pending #669.** Keep the other three (they cover the genuinely valuable JSON-parse-400 and empty-providers-502 paths). Don't assert either side of the debated behavior until #669 settles it.

If A is too large for this PR, B is the smaller move that doesn't lock in either side.

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -362,6 +362,12 @@ impl AuctionOrchestrator {
             }
         }
 
+        if pending_requests.is_empty() {
+            return Err(Report::new(TrustedServerError::Auction {
+                message: "No provider requests launched".to_string(),
```
aram356: 🤔 **thinking** — This guard fires whenever `pending_requests.is_empty()`, which includes three distinct cases beyond the typo case the tests cover:

- All providers unregistered
- All registered providers `.is_enabled() == false`
- All registered providers' `request_bids` returns `Err` (transient backend issue)

Pre-PR, all three returned `Ok(vec![])` → 200 with empty `seatbid`. Post-PR, all three return 502. The first two are misconfiguration (502 is fine). The third is a transient failure that previously surfaced as an empty auction; now it surfaces as a gateway error — any client that swallows empty 200s will start seeing 502s. Not a blocker, but worth a line in the PR description so downstream consumers aren't surprised.

♻️ **refactor** — The error message itself doesn't tell an operator *why* — unregistered? disabled? launch failure? The loop already logs the cause as a warn for each provider. Consider including the count:

```rust
return Err(Report::new(TrustedServerError::Auction {
    message: format!(
        "All {} configured provider(s) skipped or failed to launch",
        provider_names.len()
    ),
}));
```

The client-facing 502 body stays generic via `user_message()`; only the server log gets richer — exactly where you want the detail.

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -249,3 +329,72 @@ fn configured_missing_consent_store_only_breaks_consent_routes() {
...
+
+    assert_eq!(
+        response.get_status(),
+        StatusCode::BAD_REQUEST,
+        "should reject semantically invalid banner sizes as a client request error"
+    );
+}
+
+#[test]
+fn valid_auction_request_with_no_providers_returns_bad_gateway() {
```
aram356: 🤔 **thinking** — `valid_auction_request_with_no_providers_returns_bad_gateway` reads as "no providers in the system." The test actually sets `providers = []`. Consider `auction_request_with_empty_provider_list_returns_bad_gateway` for clarity, especially next to the sibling test that also configures providers but doesn't register them.

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -162,6 +163,46 @@ fn create_test_settings() -> Settings {
...
+
+            [auction]
+            enabled = true
+            providers = {providers}
+            timeout_ms = 2000
+        "#,
+    );
+
+    Settings::from_toml(&config).expect("should parse adapter auction route test settings")
+}
```
aram356, line 234: ♻️ **refactor** — `create_test_settings` (above) and this helper share ~25 lines of TOML. Not blocking, but as more route tests land this duplication will grow. Consider extracting a `base_settings_toml()` returning the common prefix, and have each helper append its differentiating block (`[consent] + [integrations.prebid] + [auction with prebid]` for the existing helper, `[auction with `{providers}`]` for this one).

aram356 (changes requested):
## Summary

The endpoint change correctly fixes a real bug — malformed `/auction` JSON now returns 400 instead of 502 — and the three accompanying tests for that path are solid. The contentious piece (already flagged by @prk-Jr) is the fourth test, which codifies a configuration typo (`providers = ["missing-provider"]`) as a request-time 502. That conflicts with the project rule in `CLAUDE.md` and locks in disputed behavior with an asserting test ahead of the #669 follow-up. Inline notes cover the blocking item, two thinking notes, two refactor suggestions, and one seedling.

## Blocking

### 🔧 wrench

- **Test codifies a config typo as request-time 502, conflicting with the project rule** (`crates/trusted-server-adapter-fastly/src/route_tests.rs:389-400`). See inline comment for the two suggested resolutions (validate at startup, or drop just this test pending #669).

## Non-blocking

### 🤔 thinking

- **Orchestrator guard behavior change is broader than the typo case** (`crates/trusted-server-core/src/auction/orchestrator.rs:365-369`). Inline.
- **Test name "no_providers" is ambiguous** (`crates/trusted-server-adapter-fastly/src/route_tests.rs:377`). Inline.

### ♻️ refactor

- **Error message lacks diagnostic info** (`crates/trusted-server-core/src/auction/orchestrator.rs:367`). Inline alongside the thinking note.
- **TOML duplication between test settings helpers** (`crates/trusted-server-adapter-fastly/src/route_tests.rs:166-196`). Inline.

### 🌱 seedling

- **Add a unit-level test for the new orchestrator guard.** The existing TODO at `orchestrator.rs:765-778` already notes the gap. A fake `AuctionProvider` whose `request_bids` returns `Err` would cover the "registered-but-failed-to-launch" path that the route-level test doesn't reach. Could land with the #669 follow-up.

## CI Status

- cargo fmt: PASS
- cargo clippy: PASS
- cargo test (Rust): PASS
- vitest (JS): PASS
- integration tests + browser integration tests: PASS
- CodeQL / Analyze: PASS

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: Addressed in fef75fc6. `build_orchestrator` now validates configured `auction.providers` plus the optional mediator after provider registration and returns a `TrustedServerError::Configuration` when a configured name is not registered. I also removed the route-level test that expected `missing-provider` to become a request-time 502 and added core coverage for missing provider, mixed valid+missing providers, missing mediator, and the registered-provider launch-failure guard.

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: Addressed in fef75fc6 with option A. `build_orchestrator` now rejects unregistered configured provider names at startup, including mixed configs like `["prebid", "missing-provider"]`, and also validates the configured mediator. The disputed request-time 502 route test for `missing-provider` was removed; the remaining route tests use registered `prebid` where startup should succeed.

crates/trusted-server-core/src/auction/orchestrator.rs
ChristianPavilonis: Addressed in fef75fc6. The runtime guard message now includes the configured provider count (`All N configured provider(s) skipped or failed to launch`), and I added a unit test with a registered fake provider whose `request_bids` fails so the true launch-failure path is covered. I also updated the PR description with the behavior-change note for the 200-empty-to-502 case.

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: Addressed in fef75fc6. Renamed the test to `auction_request_with_empty_provider_list_returns_bad_gateway` so it clearly describes `providers = []`.

crates/trusted-server-adapter-fastly/src/route_tests.rs
ChristianPavilonis: Addressed in fef75fc6. Extracted shared TOML into `base_route_settings_toml()` plus `prebid_integration_toml()`, then each settings helper now appends only its differing `[consent]` / `[auction]` sections.

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -249,3 +328,59 @@ fn configured_missing_consent_store_only_breaks_consent_routes() {
...
+
+#[test]
+fn auction_request_with_empty_provider_list_returns_bad_gateway() {
+    let settings = create_auction_test_settings_without_consent_store("[]");
+
+    let response = route_auction(&settings, valid_banner_ad_unit_body());
+
+    assert_eq!(
+        response.get_status(),
+        StatusCode::BAD_GATEWAY,
```
aram356, line 412: 🤔 **thinking** — This test uses `providers = []`, which hits the pre-existing `"No providers configured"` guard in `orchestrator.rs` (`run_providers_parallel`, ~line 284), **not** the new `pending_requests.is_empty()` branch (~line 391) this PR adds. Both return `TrustedServerError::Auction` → 502, so the route-level mapping is verified — but the *new* branch's end-to-end route mapping isn't covered here (only the `LaunchFailingProvider` orchestrator unit test exercises it).

Consider a route test with a **registered-but-disabled** provider so the request reaches the new branch end-to-end.

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -52,6 +52,32 @@ impl AuctionOrchestrator {
...
+        for provider_name in self
+            .config
+            .providers
+            .iter()
+            .chain(self.config.mediator.iter())
+        {
+            if !self.providers.contains_key(provider_name) {
+                return Err(Report::new(TrustedServerError::Configuration {
+                    message: format!(
+                        "Auction provider `{provider_name}` is configured but not registered"
```
aram356: 🌱 **seedling** — The message `"configured but not registered"` leans on internal vocabulary. The most common real cause is a *disabled (or absent) integration*, which "not registered" doesn't hint at. A more actionable phrasing, e.g.:

```rust
message: format!(
    "Auction provider `{provider_name}` is listed in [auction] but no enabled integration provides it"
),
```

would point an operator straight at the fix (enable the integration / fix the name).

crates/trusted-server-adapter-fastly/src/route_tests.rs
```diff
@@ -178,6 +218,45 @@ fn test_runtime_services(req: &Request) -> RuntimeServices {
...
+    serde_json::to_vec(&json!({
+        "adUnits": [
+            {
+                "code": "div-gpt-ad-1",
+                "mediaTypes": {
+                    "banner": {
+                        "sizes": [[300, 250]]
+                    }
+                },
+                "bids": [
```
aram356: ⛏ **nitpick** — This `bids` array references bidder `"missing-provider"` that is never asserted on, and the only caller runs with `providers = []`, so it's irrelevant to the assertion. For a helper named `valid_banner_ad_unit_body` it's a bit confusing — dropping the `bids` block would read cleaner.

aram356 (approved):
## Summary

Adds route- and orchestrator-level coverage for `/auction`, reclassifies malformed-body parsing as `400`, fails the auction with `502` when no provider request launches, and validates configured provider/mediator names at startup. The changes are correct, well-tested, and convention-aligned. Verified locally: `cargo fmt --check`, `cargo clippy --all-targets --all-features -D warnings`, and the affected `auction::` / `route_tests` suites all pass. No blocking issues — approving with a few non-blocking observations.

## Non-blocking

### 🤔 thinking

- **Whole-service blast radius for an auction config typo** — `build_orchestrator` runs in the per-request entrypoint (`main.rs:85`) *before* routing, so the new `validate_configured_provider_names` makes an unregistered/typo'd provider return a `Configuration` error (500) for **every route**, not just `/auction`. This matches the pre-existing fail-closed posture of `build_orchestrator` (a bad `server_url` already does this) and aligns with the "don't silently disable invalid providers" convention — but it's the opposite philosophy from the adjacent consent-store handling, which deliberately *scopes* failures to consent routes (enshrined by `configured_missing_consent_store_only_breaks_consent_routes`). Worth a conscious maintainer thumbs-up that taking the whole service down on an `[auction].providers` typo is the intended tradeoff. (`crates/trusted-server-core/src/auction/orchestrator.rs:56`, `crates/trusted-server-adapter-fastly/src/main.rs:85`)
- **Empty-provider route test covers the old guard, not the new branch** — see inline comment on `route_tests.rs:383`.

### 🌱 seedling / ⛏ nitpick

- **Validation message wording** — see inline comment on `orchestrator.rs:72`.
- **Irrelevant `bids` in `valid_banner_ad_unit_body`** — see inline comment on `route_tests.rs:248`.

## CI Status

- fmt: PASS (local)
- clippy: PASS (local, core + adapter, `-D warnings`)
- rust tests: PASS (local — `auction::` 29 passed, `route_tests` 4 passed)
- js tests: N/A (no JS changes)
- CodeQL (GitHub): PASS

prk-Jr (approved):
👍

PR #707, opened 2026-05-17
Rework pr-reviewer workflow around suggestions

### LINKED ISSUE #706: Make pr-reviewer agent iterate to an ideal merge candidate
## Description

Today the `pr-reviewer` agent (`.claude/agents/pr-reviewer.md`) does one read-only pass: gather context, read changed files, classify findings, present them to the user, and submit a GitHub review with inline comments. The findings live as text on the PR; the author then has to interpret and implement them.

This issue tracks teaching the agent to **iterate to an ideal merge candidate**:

1. Each review pass starts by **re-resolving the PR's current head** so the diff being reviewed is the diff that would actually merge. PRs move between passes (force-pushes, rebases, base merges); a review against a stale head wastes everyone's time.
2. The user **triages each finding twice**: include it in the GitHub review, *and* implement it as code in a stacked fix-up PR. Findings with a concrete actionable fix (🔧 ♻️ ⛏ 🏕) are eligible; questions, thinking-aloud, seedlings, out-of-scope, and praise stay comment-only.
3. The agent opens **one fix-up PR per review engagement**, stacked on top of the PR under review and reused across passes:
   - branch: `review/<timestamp>-<pr-number>` (UTC `YYYYMMDD-HHMMSS`, fixed at engagement start)
   - title: `<timestamp> Review fixes for #<pr-number>`
   - base: the PR's head ref (not `main`)
   - later passes rebase the branch onto the PR's current head and force-push-with-lease; new commits are appended, the PR body's "Findings addressed" list grows
4. The GitHub review **references the fix-up PR** in the body summary and in each inline comment for findings that were implemented; non-implemented findings keep the existing suggestion format.
5. The verdict logic is updated so 🔧 findings that were **addressed in the fix-up PR** no longer force `REQUEST_CHANGES`.
6. After each pass the agent loops back to step 1 and re-reviews the combined (PR + fix-up) state. It **stops** when:
   - a full pass surfaces no new actionable findings (ideal merge candidate), or
   - the only remaining items are blocked on the author (open ❓, decisions, required rebase), or
   - the user says so.
7. Hard rules forbid pushing fix-up commits or submitting reviews without explicit user approval of which findings to implement, force-pushing to `main`, or letting a fix-up PR target `main` instead of the PR's head branch.

## Motivation

End state we want: a stacked fix-up PR that, once merged into the PR under review, leaves nothing further a reviewer would need to change. Today the author bears the full burden of translating review text into commits, and a second review pass against a moved head is manual. With the iterative workflow the agent shortens the review cycle from "review → discuss → author implements → re-review" to "review → implement → re-review until clean".

This was exercised end-to-end on #621 across two passes — a stale-PR situation, several upstream fixes between passes, and a final stacked fix-up PR — and the workflow held up.

## Acceptance criteria

- `.claude/agents/pr-reviewer.md` documents the iterative workflow as the agent's default behavior, including the goal statement up front, the `review/<timestamp>-<pr#>` / `<timestamp> Review fixes for #<pr#>` naming, the re-fetch-head step, the per-pass loop, and the stop conditions.
- The verdict matrix accounts for findings fixed in the stacked PR.
- The rules section forbids skipping user approval, force-pushing without `--force-with-lease` after a rebase, and targeting `main` from a fix-up PR.

aram356:
## Summary

- Reworks the `pr-reviewer` agent around a single review pass that posts GitHub review comments and user-approved `suggestion` blocks, instead of pushing fixes or opening a stacked fix-up PR.
- Adds explicit PR, branch-remote, and branch-local modes with private review refs, a repo-root-anchored reviewer worktree, merge-base-aware diff handling (`$DIFF_RANGE`/`$DIFF_BASE`), branch-only artifact rendering, and a head re-check before submission.
- Hardens CI handling: checks are classified from the **full** `gh pr checks --json` bucket set (required checks annotated separately), any failed check becomes a blocking body-level finding, and reviewer-side tooling failures (network/auth/pending) are recorded in CI Status only so they can never escalate to `REQUEST_CHANGES` on the author's PR.
- Makes suggestions trustworthy: diff-hunk anchoring rules (RIGHT/LEFT side, no cross-hunk spans), a deterministic fence-widening rule for replacements containing backtick runs, per-suggestion isolated scratch verification in the reviewer worktree with a byte-exact post-verify drift check, and repo-correct verification commands (per-target cargo aliases including the cross-adapter parity suite, `crates/trusted-server-js` paths).
- Makes submission safe: verdicts are downgraded to `COMMENT` when the PR author is the authenticated user (GitHub rejects self-review verdicts) **before** any pending-review handling, existing pending reviews are never deleted without explicit user consent, and the review payload is built as a single strict-JSON `--input` file pinned to the reviewed head OID.

## Changes

| File | Change |
| ---- | ------ |
| `.claude/agents/pr-reviewer.md` | Replaces the previous fix-up-PR loop with a one-pass review workflow. Documents mode resolution, idempotent worktree setup, full changed-file reads, CI classification, finding triage with two-decision user approval (include? exact suggestion bytes?), GitHub suggestion eligibility, scratch verification and CI-equivalent gates, verdict precedence, branch-only rendering (8a) vs GitHub submission (8b), PR head drift handling, self-review and pending-review guards, and safe `gh api` review submission. |

## Closes

Closes #706

## Test plan

This is a workflow/documentation-only change under `.claude/agents/`; no Rust or JS source is affected.

- [x] `git diff --check -- .claude/agents/pr-reviewer.md`
- [x] All file paths and commands cross-checked against the tree at head: `crates/trusted-server-js/{lib,dist,build.rs}`, every cargo alias in `.cargo/config.toml`, the parity suite manifest, and CLAUDE.md's CI gate list
- [x] `gh` CLI behavior verified against live help/API: `pr checks --json` fields and bucket values, exit code 8 semantics, `api --input` vs `-f` body handling, review-submission payload shape
- [x] Mocked CI capture cases: nonzero exit with JSON → classify from JSON; exit 8 with no JSON → pending note; nonzero exit with no JSON → CI Status diagnostic (never a finding)
- [ ] Rust/JS/docs CI gates — n/a, no source affected

## Checklist

- [x] No production code changed


.claude/agents/pr-reviewer.md
```diff
@@ -151,32 +171,157 @@ organized by severity, with:
...
 
 #### Determine the review verdict
 
-- If any 🔧 (wrench) findings are included: `REQUEST_CHANGES`
-- If any ❓ (question) findings are included: `COMMENT` (questions need answers, not change requests)
+- If any 🔧 (wrench) findings remain **un-implemented** in the review: `REQUEST_CHANGES`
+- If any ❓ (question) findings are included: `COMMENT`
+- If all 🔧 findings were addressed in the fix-up PR and only ❓ / non-blocking remain: `COMMENT` (note the fix-up PR in the summary)
 - If only non-blocking findings (🤔 ♻️ 🌱 📝 ⛏ 🏕 📌 👍): `COMMENT`
 - If no findings (or only 👍 praise): `APPROVE`
```
ChristianPavilonis: Automated Yesman review: consider tightening this for the new stacked-fix workflow. If the only clean state is `original PR + fix-up PR`, an `APPROVE` review on the original PR can be read as safe to merge without first applying the fix-up PR. A safer rule would keep the original PR at `COMMENT` (or `REQUEST_CHANGES` for unimplemented blocking findings) while required fixes live only in the stacked PR, and reserve `APPROVE` for when the original PR head itself contains them.

ChristianPavilonis (changes requested):
Automated review by Yesman.

I found one blocking correctness issue in the updated pr-reviewer instructions. The PR adds a stale-head guardrail in Step 1, but Step 2 still says to enumerate changed files with `git diff main...HEAD --name-only` (`.claude/agents/pr-reviewer.md`, current line 59). Because that line is not part of the submitted diff, I could not attach this as an inline comment.

This should use the just-fetched PR head, not the reviewer worktree's local `HEAD`. Otherwise later passes can enumerate files from an old checkout, a fix-up branch, or another branch entirely, causing the agent to miss current PR changes before submitting a review. Please make it consistent with Step 1, for example `git diff origin/<baseRefName>...origin/<headRefName> --name-only` after fetching the base, or `git diff main...origin/<headRefName> --name-only` if `main` is intentionally fixed.

CI is mostly passing with some checks pending at review time.

.claude/agents/pr-reviewer.md
```diff
@@ -2,60 +2,413 @@
...
+
+```bash
+if [ -n "$NUMBER" ]; then
+    # Use structured output. `gh pr checks` has non-zero exits for different
+    # states. Classify from JSON whenever stdout exists; use the exit code only
+    # to distinguish pending/no-output from command failures.
+    ci_error=$(mktemp)
+    checks_json=$(gh pr checks "$NUMBER" \
+        --repo IABTechLab/trusted-server \
+        --required \
```
ChristianPavilonis: Automated review: this required-only query drops several Trusted Server CI gates from the reviewer's view. On this PR, `gh pr checks --required` returns only four checks (`cargo test`, `format-typescript`, `cargo fmt`, `format-docs`), while the full check set also includes `Analyze (rust)`, `vitest`, CodeQL, browser integration tests, integration tests, etc. Since `CLAUDE.md` treats those workflows as PR gates, a failure in one of the omitted checks could be missed and the reviewer could still report clean CI.

Suggested fix: query the full check set for reporting/classification, and separately annotate which checks are branch-protection-required if needed.

ChristianPavilonis (commented):
Automated review: I reviewed PR #707 against `origin/main` at head `5ea9211305c040fe8ddce01da70b392aa63db554`. The workflow rewrite is generally careful around stale heads, suggestion safety, pending reviews, and JSON review submission, but I found one CI-coverage gap that can cause the reviewer to miss failing Trusted Server gates.

## Findings

### P1 / High

1. **Do not limit CI classification to branch-protection-required checks** — `.claude/agents/pr-reviewer.md:370`
   - Issue: The new CI collection uses `gh pr checks --required` and the body template later says to render only checks reported by that required-only query. In this repository, that currently returns only `cargo test`, `format-typescript`, `cargo fmt`, and `format-docs`; it omits checks such as `Analyze (rust)`, `vitest`, CodeQL, browser integration tests, and integration tests.
   - Why it matters: `CLAUDE.md` treats those broader workflows as PR gates. If one of the omitted checks fails, the reviewer can still report clean CI and potentially approve/comment with no failed-CI finding, hiding a real regression in Trusted Server's Rust/JS/integration validation.
   - Suggested fix: Query and report all PR checks (for example, `gh pr checks "$NUMBER" --json name,bucket,state,link`), then separately mark which are required if branch-protection status matters. At minimum, classify failures/cancellations from the full check set as review findings, while preserving the existing required-check handling for merge-blocking branch protection.

## CI / Existing Reviews

CI is currently passing. Existing review feedback included an older stale-HEAD concern that appears addressed by the current `$DIFF_RANGE` / private-ref workflow; I did not duplicate that finding.

.claude/agents/pr-reviewer.md
```diff
@@ -2,60 +2,434 @@
...
+    elif git worktree list --porcelain | grep -qx "worktree $(realpath "$WT")"; then
+        # Registered worktree from a prior invocation. The parent repo
+        # already ran the shared fetch above and verified $HEAD_OID_EXPECTED
+        # — linked worktrees share `.git`, so $HEAD_REF and
+        # `refs/remotes/origin/$BASE_BRANCH` are already current here. A
+        # second fetch would re-pull whatever the PR head looks like *now*
+        # (which could have moved between the verify and the reset), so
+        # don't fetch again; just reset to the already-verified $HEAD_REF.
+        git -C "$WT" reset --hard "$HEAD_REF"
+        git -C "$WT" clean -fd
```
ChristianPavilonis, line 326: Automated review: ♻️ **refactor** — Reused reviewer worktrees should clear ignored JS bundles before a new pass.

The reuse path resets tracked files and removes untracked non-ignored files, but it leaves ignored artifacts such as `crates/js/dist/` behind. Later in the workflow we explicitly call out that a stale `crates/js/dist/` can leak into Rust compilation via the JS bundle build path, and the per-suggestion loop only removes it when the *previous iteration in the same invocation* set `prev_iteration_ran_js_build=1`. If an earlier invocation is interrupted after running `node build-all.mjs`, the next invocation reuses the same `$WT` and can start scratch verification with stale ignored bundles from the old head/suggestion, which can produce misleading compile/test results.

Suggested fix: in the existing-worktree setup path, clear known ignored generated inputs that affect verification before starting the new pass, e.g. add `git -C "$WT" clean -fdx crates/js/dist` (or an equivalent targeted removal) alongside this reset/clean sequence.

ChristianPavilonis (commented):
Automated review: I reviewed PR #707 at head `a7b267c955d45a27a309b23e7bef33b787ae288f`. The workflow rewrite is careful overall and the previously noted stale-head and required-only CI issues appear addressed in this revision. I found one non-blocking robustness gap around reusing the agent-owned worktree after an interrupted suggestion-verification pass; no blocking correctness, security, data-loss, authorization, or severe compatibility issues found.

## Findings

- P2: Reused reviewer worktrees should clear ignored JS bundles before a new pass — see inline comment.

## CI / Existing Reviews

All GitHub checks currently pass. Existing review feedback covered the stale `HEAD` diff issue and the required-only CI query gap; I did not duplicate those findings.

.claude/agents/pr-reviewer.md
aram356: Addressed in 131ea513. Step 3 now queries the **full** check set (`gh pr checks "$NUMBER" --json name,bucket,state,link`, no `--required`) and classifies any `fail`/`cancel` bucket as a blocking 🔧 body-level finding — so a failure in `Analyze (rust)`, `vitest`, CodeQL, or the integration tests can no longer fall through to a clean-CI `APPROVE`. A separate best-effort `required_names` capture annotates which failures are merge-blocking under branch protection, and the CI Status body template renders the full set with a `(required)` tag.

.claude/agents/pr-reviewer.md
aram356: Addressed in 5c6b8f27. The worktree-reuse path now runs `git -C "$WT" clean -fdx crates/js/dist` after the `reset --hard` + `clean -fd`, so a stale `crates/js/dist/` left by an interrupted prior invocation that ran `node build-all.mjs` can't leak into the next pass's Rust compile via `include_str!()`. The `-x` clean is scoped to `crates/js/dist` so the `target/` and `node_modules/` caches that speed up verification are preserved.

.claude/agents/pr-reviewer.md
aram356: This concern was about the earlier stacked fix-up PR shape, which the rework removed entirely. The agent no longer pushes fixes or opens a fix-up branch — it only posts review comments and one-click `suggestion` blocks, so required fixes never live in a separate PR. The verdict rules in step 8a now state explicitly that a suggestion is a *proposal*, not a fix: any open 🔧/❓ finding forces `REQUEST_CHANGES`, and `APPROVE` is reserved for when there are no findings — i.e. the reviewed head itself is clean. So the "APPROVE on the original PR while the fix lives elsewhere" reading no longer applies.

.claude/agents/pr-reviewer.md
```diff
@@ -2,60 +2,445 @@
...
+    fi
+fi
 ```
 
-If no PR exists, review the branch diff directly and skip the GitHub review
-submission (report findings as text instead).
+After this step every later step uses `${WT:-.}` for cwd (so BRANCH-LOCAL
+implicitly runs from the project root) and `$DIFF_RANGE` for diffs. There
+are no more per-mode forks until step 7e (which checks `[ -n "$WT" ]` for
+scratch verification) and step 8 (which is skipped entirely for BRANCH-*
```
ChristianPavilonis: Automated review: P2 — This says branch modes skip step 8 entirely, but step 8a is where the branch-only review artifact and would-have-been verdict are composed. Later text correctly says BRANCH-* modes still run 8a and skip only 8b submission, so an agent following this earlier sentence could stop without rendering any branch-only findings.

Suggested fix: change this to say that only **step 8b / GitHub submission** is skipped for BRANCH-* modes, while 8a still runs to compose and render the local review artifact.

.claude/agents/pr-reviewer.md
```diff
@@ -138,159 +523,845 @@ whether a comment requires action, is a suggestion, or is informational.
...
+
+Per finding, show:
 
 - Emoji tag and title
-- File path and line number
+- File path and the precise diff range the suggestion would target
+  (`<file>:<start_line>-<line>` if multi-line, `<file>:<line>` if single)
 - Description and suggested fix
-- Whether it would be an inline comment or body-level finding
+- For suggestion-eligible findings, the literal `\`\`\`suggestion … \`\`\``
```
ChristianPavilonis, line 584: Automated review: P2 — The workflow promises that users approve exact replacement bytes and also supports docs/Markdown suggestions, but it always describes the suggestion as a literal triple-backtick fence. If the replacement itself contains a line with triple backticks, the outer `suggestion` block can close early, so the rendered one-click suggestion may be truncated or malformed rather than matching the bytes the user approved.

Suggested fix: before displaying or submitting a suggestion, scan the replacement for backtick runs and use an outer fence longer than any run in the replacement if GitHub parses that form correctly; otherwise demote that finding to prose-only. Apply the same rule to both the triage preview and the final inline comment body.

ChristianPavilonis (commented):
Automated review: I reviewed PR #707 at head `ba8a52113214d0502da7b4e2df822010c7eef059`. The revised pr-reviewer workflow is much safer than the previous fix-up-PR approach, and I did not find blocking correctness, security, data-loss, authorization, or severe compatibility issues. I left two medium-severity inline comments for branch-only flow clarity and Markdown suggestion-fence safety.

## CI / Existing Reviews

Current GitHub checks are mostly passing; `integration tests` and `integration tests (EdgeZero entry point)` were still pending at review time. Existing review feedback on stale HEAD handling, full CI classification, and stale JS bundle cleanup appears addressed; I did not duplicate those findings.

aram356:
@aram356 Check if workflow creates worktree twice

.claude/agents/pr-reviewer.md
aram356: Addressed in 4f4e1eea7. Reworded the step 1 orientation note so it scopes the skip to **step 8b (GitHub submission)** only, and states that 8a still runs in every mode to compose the artifact and render the would-have-been verdict/findings into chat for BRANCH-* modes.

.claude/agents/pr-reviewer.md
aram356: Addressed in 4f4e1eea7. Added a fence-length rule in step 7a: scan the replacement for the longest backtick run `N`; use the normal three-backtick fence when `N < 3`, and an `N + 1` backtick outer fence when `N >= 3` (CommonMark closes a fence only on a run of *at least* as many backticks). Step 7e verifies the widened form still renders as a one-click suggestion; if it doesn't, the finding is demoted to prose-only rather than posting a malformed suggestion. The rule is referenced from both the step 6 triage preview and the 7b inline-comment body so the approved bytes/fence and the submitted ones stay identical.

.claude/agents/pr-reviewer.md
```diff
@@ -138,159 +554,878 @@ whether a comment requires action, is a suggestion, or is informational.
...
+**JS/TS suggestions** (run from the package root in a subshell so the
+worktree's cwd is unaffected). CLAUDE.md's JS-side build pipeline also runs
+`node build-all.mjs` to re-bundle the per-integration IIFEs; skipping it
+means a suggestion that edits `src/integrations/*/index.ts` could compile
+under `vitest` but break the runtime bundle the Rust crate `include_str!`s
+at build time. Run the build whenever the suggestion touches files under
+`crates/js/lib/src/`:
+
+```bash
+(cd "$WT/crates/js/lib" && npx vitest run)
```
ChristianPavilonis: Automated review: 🔧 **wrench** — The JS verification commands still use `$WT/crates/js/lib`, but this repository's JS package is under `crates/trusted-server-js/lib` (as documented in `CLAUDE.md` and present in the tree). The same wrong `crates/js/...` prefix is also used for the `lib/src` trigger and `dist` cleanup earlier/later in this file. As written, JS/TS suggestions would `cd` into a missing directory, fail to detect real TS source edits, and clean a non-existent `crates/js/dist` instead of the ignored `crates/trusted-server-js/dist` that `crates/trusted-server-js/build.rs` consumes. That breaks the scratch-verification and stale-bundle guarantees this workflow is adding.

Suggested fix: replace the `crates/js/...` references throughout this workflow with the actual package paths: `crates/trusted-server-js/lib`, `crates/trusted-server-js/lib/src/`, `crates/trusted-server-js/dist`, and `crates/trusted-server-js/build.rs`.

ChristianPavilonis (changes requested):
Automated review: I reviewed PR #707 against `origin/main` at head `db9b3f33d1e87cb620bac5e291d0e8cdad8edb01`. The workflow rewrite is detailed and the previously reported stale-head, required-only CI, branch-mode, suggestion-fence, and worktree-registration issues appear addressed. I found one blocking workflow correctness issue: the new JS scratch-verification and stale-bundle cleanup instructions point at a non-existent `crates/js/...` tree instead of this repository's `crates/trusted-server-js/...` package, so the reviewer would fail or skip the intended JS checks and leave stale bundles behind.

## Findings

### P1 / High

- **JS suggestion verification uses the wrong package path** — see inline comment.

## CI / Existing Reviews

`gh pr checks 707` currently reports one failing check: `prepare integration artifacts`. The job failed while Docker was fetching `php:8.3-cli-alpine` due to a Docker Hub TLS handshake timeout, so I am treating it as CI context rather than attributing it to this docs-only change. Existing automated review feedback on stale `HEAD` diffing, full-check CI classification, stale JS bundle cleanup, branch-only step 8 handling, suggestion-fence length, and idempotent worktree setup appears addressed; I did not duplicate those resolved comments.

.claude/agents/pr-reviewer.md
aram356: Addressed in 5a2a04290. All `crates/js/...` references now point at `crates/trusted-server-js/...` — the `lib` package root for vitest/format/build-all.mjs, `lib/src/` for the JS-build trigger, and the crate-root `dist/` + `build.rs` for the stale-bundle cleanup (verified against the tree: `dist/` sits at the crate root, not under `lib/`). The same restructure also invalidated the workspace-wide cargo commands in this section (`cargo check/clippy/test --workspace` no longer run at the workspace root), so the preflight, full CI-equivalent gate, and local-mode disclaimers now use the per-target `check-*`/`clippy-*`/`test-*` aliases from CLAUDE.md, and the Rust test checklist adds `cargo test-spin`.

ChristianPavilonis (approved):

aram356:
Self-review round at head 5a2a04290, fixes pushed in ba3c959cf:

- **Self-review rejection guard** (new known-issue 1 in step 8b): GitHub 422s `APPROVE`/`REQUEST_CHANGES` when the PR author is the authenticated user — which is always possible here since the agent shares the user's identity. The check runs before the pending-review handling, so a consented draft deletion can no longer precede a POST that fails, and the would-have-been verdict is preserved in the review body.
- **Reviewer-side failures can't block the PR**: step 3 now routes `gh pr checks` command/auth failures and pending checks to the CI Status section only — previously the "diagnostic ❓" wording combined with 8a's "any ❓ → REQUEST_CHANGES" rule to let an expired token block the author's PR.
- **Fence-length rule stated as deterministic**: step 7a delegated a rendering check to 7e's scratch pass, but 7e verifies replacement bytes locally and cannot observe GitHub's server-side suggestion rendering; the impossible delegation is replaced with an explicit caveat.
- **`$WT` anchored to the repo root** (`git rev-parse --show-toplevel`): a relative path resolved against a drifted cwd could create the reviewer worktree inside a crate directory, where the root-anchored `.claude/*` gitignore rule wouldn't cover it. Also `grep -Fqx` for the literal worktree-path match.
- **Parity gate added**: the "full CI-equivalent gate" omitted CLAUDE.md gate 4 (`cargo test --manifest-path crates/trusted-server-integration-tests/Cargo.toml --test parity`) — the gate most relevant when a suggestion touches `trusted-server-core`.
- Cleanups: explicit `user_choice` placeholder in the pending-review snippet, dead flag reset removed, garbled `clean -fd` comment fixed, step 8a subsection headings demoted to the correct level.

One reviewer finding was refuted during verification: `crates/trusted-server-core/src/auction/types.rs` (used in the 7c example) does exist in the tree, so the example stands.

prk-Jr (approved):

Issue #684, opened 2026-05-08
Add secret-store backed config references for secret values


ChristianPavilonis:
## Problem

Several application settings are secret material today, but they are authored directly in `trusted-server.toml` and then stored as part of the canonical runtime config payload. This increases plaintext exposure and makes it harder for provisioning to manage secrets safely.

We should establish a first-class secret reference pattern for config fields that should be backed by a provider secret store. The goal is for secret-bearing fields to be clearly marked in code, represented in TOML as references, and optionally generated/provisioned by the CLI instead of requiring users to paste plaintext values into config files.

## Current secret-bearing config fields

Current obvious candidates include:

```toml
[publisher]
proxy_secret = "..."

[edge_cookie]
secret_key = "..."

[[handlers]]
username = "..."
password = "..."
```

Notes:

- `publisher.proxy_secret` is secret material.
- `edge_cookie.secret_key` is secret material.
- `handlers.password` is secret material.
- `handlers.username` may be sensitive/redacted, but is not necessarily secret-store material.
- Request-signing private keys are already intended to live in `signing_keys`.
- The Fastly runtime API token is already intended to live in `api-keys/api_key`, but that location is hardcoded and should become provider-scoped.
- Future integrations may add API keys, client secrets, shared secrets, etc.

Today `Redacted<String>` helps avoid logging/displaying secret values, but it does not encode the stronger semantic that a value should be sourced from a secret store.

## Proposed direction

Introduce a dedicated secret value/reference type for config fields that should be backed by a secret store.

A near-term TOML shape could be:

```toml
[providers.fastly.secrets]
default_store_name = "ts_secrets"

[publisher]
domain = "example.com"
cookie_domain = ".example.com"
origin_url = "https://origin.example.com"
proxy_secret = { secret = "publisher/proxy_secret" }

[edge_cookie]
secret_key = { secret = "edge_cookie/secret_key" }

[[handlers]]
id = "admin"
path = "^/admin"
username = "admin"
password = { secret = "handlers/admin/password" }
```

With a provider default secret store, this:

```toml
proxy_secret = { secret = "publisher/proxy_secret" }
```

means:

```text
provider = fastly
store = ts_secrets
key = publisher/proxy_secret
```

For advanced cases, a field could override the default store:

```toml
proxy_secret = { store = "publisher_secrets", key = "proxy_secret" }
```

## Default secret refs and generation

Each secret-bearing field should have metadata describing:

- config path
- default secret key
[...]

PR #715, opened 2026-05-19
Add secret-store config references spec

### LINKED ISSUE #714: Create secret-store config references spec
Parent: #684\n\nCreate a proposal/spec document for secret-store backed configuration references.\n\nScope:\n- Capture the v1 config shape for inline local/dev secrets and provider secret refs.\n- Define Fastly production provisioning behavior.\n- Define runtime resolution and fail-closed behavior for missing secrets.\n- Keep v1 narrow: no automatic generation, no handler IDs, no request-signing migration.

ChristianPavilonis:
## Summary

- Adds a design spec for secret-store backed config references from #684.
- Captures the narrowed v1 scope from the design discussion: refs plus provisioning/runtime policy, no generation or handler IDs.
- Documents local inline-secret support and production provisioning rejection of inline secrets.

## Changes

| File | Change |
| ---- | ------ |
| `docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md` | Adds the proposal/spec for secret-store backed config refs, Fastly provider behavior, runtime resolution semantics, provisioning validation, and acceptance criteria. |

## Closes

Closes #714
Related #684

## Test plan

- [ ] `cargo test --workspace --exclude trusted-server-cli`
- [ ] `cargo test --package trusted-server-cli --target "$(rustc -vV | sed -n 's/^host: //p')"`
- [ ] `cargo clippy --workspace --exclude trusted-server-cli --all-targets --all-features -- -D warnings`
- [ ] `cargo clippy --package trusted-server-cli --target "$(rustc -vV | sed -n 's/^host: //p')" --all-targets -- -D warnings`
- [ ] `cargo fmt --all -- --check`
- [x] Docs format: `cd docs && npm run format -- superpowers/specs/2026-05-19-secret-store-config-references-design.md`
- [x] Other: docs-only change; Rust/JS/runtime checks not run.

## Checklist

- [x] New code has tests 

### DIFF
--- a/docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
+++ b/docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
@@ -1,0 +1,537 @@
+# Secret-Store Backed Config References Design
+
+> **Status:** Proposal
+> **Issue:** [#684](https://github.com/IABTechLab/trusted-server/issues/684)
+> **Scope:** Secret-bearing application config values, runtime resolution policy, and provisioning validation
+> **Date:** May 2026
+
+## Summary
+
+Trusted Server currently represents several secret-bearing application settings as inline string values in `trusted-server.toml`. `Redacted<String>` prevents accidental display in logs and debug output, but it does not express that a value should be backed by a provider secret store, and it does not prevent plaintext secret material from being uploaded as part of the canonical runtime configuration payload.
+
+This design introduces a first-class secret reference model for application config fields that should be backed by a provider secret store.
+
+In v1:
+
+- Secret-bearing config fields may be authored either as inline strings or as secret refs.
+- Inline strings remain supported for local development, tests, and migration workflows.
+- Fastly production provisioning rejects inline secret values for the v1 secret-bearing fields.
+- Production provisioning verifies referenced secret stores and secret items exist before applying config.
+- Runtime resolves secret refs eagerly for each loaded config snapshot, but missing runtime secrets do not make the entire service fail to load.
+- Missing runtime secrets are logged and cause only the affected capability to fail closed or degrade.
+- Secret refs participate in the config hash; resolved secret values do not.
+
+This proposal intentionally keeps v1 narrow. It does not add automatic secret generation, handler IDs, or request-signing migration into the generic secret-ref model.
+
+## Scope
+
+### In scope
+
+- A first-class `SecretString` / `SecretRef` representation for selected application config fields.
+- TOML syntax for inline secrets and secret refs.
+- Provider default secret-store configuration for Fastly provisioning.
+- A fixed Fastly runtime alias for the default application secret store.
+- Runtime secret-ref resolution semantics.
+- Feature-scoped runtime fail-closed behavior for missing secrets.
+- CLI/provisioning validation that rejects inline production secrets and missing referenced secrets.
+- Config-hash behavior for secret refs.
+- Initial v1 secret-bearing field list.
+
+### Out of scope
+
+- Automatic generation of `publisher.proxy_secret`, `edge_cookie.secret_key`, handler passwords, request-signing keys, or runtime API tokens.
+- Adding `id` to `[[handlers]]`.
+- Moving request-signing private keys or runtime API tokens into the generic `SecretString` model.
+- Storing `handlers.username` in a secret store.
+- Hot reload or background refresh of resolved secrets.
+- Cross-provider secret-store resource design beyond the Fastly provider shape needed for v1.
+- A derive macro or annotation system for secret field descriptors.
+- Rewriting local authoring files automatically.
+
+## Current state
+
+Several application settings are secret material but are currently authored directly in `trusted-server.toml`:
+
+```toml
+[publisher]
+proxy_secret = "..."
+
+[edge_cookie]
+secret_key = "..."
+
+[[handlers]]
+username = "..."
+password = "..."
+```
+
+Known semantics today:
+
+- `publisher.proxy_secret` is secret material.
+- `edge_cookie.secret_key` is secret material.
+- `handlers.password` is secret material.
+- `handlers.username` may be sensitive and should remain redacted, but it is not secret-store material for v1.
+- Request-signing private keys already live in `signing_keys` and are out of scope for v1.
+- The Fastly runtime API token already lives in `api-keys/api_key` for request-signing rotation and is out of scope for v1.
+
+The config model uses `Redacted<String>` for these values. Redaction protects display surfaces, but it does not distinguish local inline development values from production secret-store references.
+
+## Goals
+
+- Make secret-store-backed config values explicit in the schema.
+- Keep local development simple by allowing inline dev secrets.
+- Prevent production provisioning from uploading inline secret material in canonical runtime config.
+- Let production provisioning verify secret-store readiness before apply.
+- Keep runtime availability resilient if a secret is deleted or corrupted out-of-band after provisioning.
+- Ensure affected runtime behavior fails closed rather than bypassing protections.
+- Keep the first implementation small and understandable.
+
+## Non-goals
+
+- Requiring secret refs for all local development workflows.
+- Inferring default secret refs from omitted fields.
+- Automatically generating or rotating secret values.
+- Introducing a broad secret-management CLI surface in v1.
+- Reworking request-signing storage in v1.
+- Designing a multi-provider secrets abstraction beyond the existing platform secret-store interface.
+
+## V1 secret-bearing fields
+
+V1 applies the secret-ref model to these application config fields:
+
+| Field                    | V1 behavior                                                   | Default shorthand key recommendation |
+| ------------------------ | ------------------------------------------------------------- | ------------------------------------ |
+| `publisher.proxy_secret` | Inline for local/dev; ref required by production provisioning | `publisher/proxy_secret`             |
+| `edge_cookie.secret_key` | Inline for local/dev; ref required by production provisioning | `edge_cookie/secret_key`             |
+| `handlers[].password`    | Inline for local/dev; ref required by production provisioning | No derived default in v1             |
+
+`handlers.username` remains `Redacted<String>` and inline-only in v1.
+
+Request-signing resources remain on their current dedicated path:
+
+- private signing keys in the `signing_keys` runtime alias
+- runtime Fastly API token in the `api-keys` runtime alias under `api_key`
+
+## TOML representation
+
+### Inline local/dev form
+
+Inline strings remain accepted by the config parser and runtime:
+
+```toml
+[publisher]
+domain = "example.com"
+cookie_domain = ".example.com"
+origin_url = "https://origin.example.com"
+proxy_secret = "dev-proxy-secret"
+
+[edge_cookie]
+secret_key = "dev-edge-cookie-secret"
+
+[[handlers]]
+path = "^/admin"
+username = "admin"
+password = "dev-password"
+```
+
+This form is intended for local development, tests, and migration only. Fastly production provisioning must reject it for v1 secret-bearing fields.
+
+### Default-store secret ref shorthand
+
+The common production form uses the default application secret-store alias:
+
+```toml
+[publisher]
+domain = "example.com"
+cookie_domain = ".example.com"
+origin_url = "https://origin.example.com"
+proxy_secret = { secret = "publisher/proxy_secret" }
+
+[edge_cookie]
+secret_key = { secret = "edge_cookie/secret_key" }
+
+[[handlers]]
+path = "^/admin"
+username = "admin"
+password = { secret = "handlers/admin/password" }
+```
+
+`secret` means “read this key from the default application secret-store runtime alias.”
+
+### Explicit store override form
+
+Advanced cases may use an explicit runtime store alias:
+
+```toml
+[publisher]
+proxy_secret = { store = "publisher_secrets", key = "proxy_secret" }
+```
+
+Rules:
+
+- `{ secret = "..." }` and `{ store = "...", key = "..." }` are the only supported ref shapes in v1.
+- `secret` must not be combined with `store` or `key`.
+- `key` must not be empty.
+- `store`, when present, must not be empty.
+- The `store` value is a runtime secret-store alias, not necessarily the provider's underlying resource name.
+
+### No omitted-field default refs in v1
+
+V1 does not infer secret refs from omitted fields.
+
+This is invalid for required secret-bearing fields:
+
+```toml
+[publisher]
+domain = "example.com"
+cookie_domain = ".example.com"
+origin_url = "https://origin.example.com"
+# proxy_secret omitted
+```
+
+Fields remain visibly required. Tooling may later scaffold explicit refs, but omission does not imply a default.
+
+## Fastly provider configuration
+
+Provider-specific resource names remain outside canonical runtime application config. For Fastly, the source config may include:
+
+```toml
+[providers.fastly.secrets]
+store_name = "customer_ts_secrets"
+```
+
+Provisioning uses this provider setting to create or locate the underlying Fastly Secret Store resource and bind it to a fixed runtime alias.
+
+### Fixed runtime alias
+
+The default application secret store uses a fixed runtime alias:
+
+```text
+ts_secrets
+```
+
+Provisioning may link any underlying Fastly Secret Store resource as `ts_secrets`. Runtime code opens the alias, not the underlying provider resource name.
+
+A shorthand ref:
+
+```toml
+proxy_secret = { secret = "publisher/proxy_secret" }
+```
+
+therefore resolves as:
+
+```text
+store alias = ts_secrets
+key = publisher/proxy_secret
+```
+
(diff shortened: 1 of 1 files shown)


docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
```diff
@@ -0,0 +1,537 @@
...
+proxy_secret = { secret = "publisher/proxy_secret" }
+```
+
+to this:
+
+```toml
+proxy_secret = { secret = "publisher/proxy_secret_v2" }
+```
+
+must change the config hash.
```
aram356, line 397: 🔧 **wrench** — Hash determinism: shorthand and explicit-default refs resolve to the same secret but hash differently.

`{ secret = "publisher/proxy_secret" }` and `{ store = "ts_secrets", key = "publisher/proxy_secret" }` both point at the same physical secret, but as written they produce different canonical TOML bytes and therefore different `config_hash` values. Attestation/diff tooling will report spurious config changes when an operator normalizes shorthand → explicit (or back).

**Fix**: either (a) require canonicalization to normalize one form to the other before hashing, or (b) state explicitly that the two forms are intentionally non-equivalent for hashing and document why. Either way, add the chosen rule to Acceptance Criteria.

docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
```diff
@@ -0,0 +1,537 @@
...
+Fields remain visibly required. Tooling may later scaffold explicit refs, but omission does not imply a default.
+
+## Fastly provider configuration
+
+Provider-specific resource names remain outside canonical runtime application config. For Fastly, the source config may include:
+
+```toml
+[providers.fastly.secrets]
+store_name = "customer_ts_secrets"
+```
```
aram356, line 200: 🔧 **wrench** — `providers.fastly.secrets.store_name` must be added to the CLI's `validate_distinct_names`.

[`crates/trusted-server-cli/src/config.rs:213-226`](../../../crates/trusted-server-cli/src/config.rs#L213-L226) already enforces distinct Fastly Secret Store resource names across `signing_secret_store_name` and `runtime_api_secret_store_name`. Adding a third Secret Store without extending that check lets two different runtime aliases (`ts_secrets`, `signing_keys`, `api-keys`) silently link the same underlying resource, collapsing the isolation contract the spec assumes between application secrets and request-signing/runtime-API secrets.

**Fix**: spec should require provisioning to reject overlap across all three secret-store resource names, and Acceptance Criteria should add a checkbox for it.

docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
```diff
@@ -0,0 +1,537 @@
...
+
+`ts provision fastly apply` must fail before any mutating action if any v1 secret-bearing field is inline:
+
+- `publisher.proxy_secret`
+- `edge_cookie.secret_key`
+- `handlers[].password`
+
+There is no v1 `--allow-inline-secrets` production escape hatch.
+
+Rationale: local developers should not be uploading local inline config to production, and production provisioning should prevent accidental plaintext secret upload into the canonical runtime config store.
```
aram356, line 338: ❓ **question** — Production boundary: where exactly is inline-secret rejection enforced?

The spec gates inline-secret rejection at `ts provision fastly apply`. But the canonical TOML payload is what actually lands in the Config Store. Is `ts provision fastly apply` the only code path that writes the canonical app config payload to a Fastly Config Store? If a future `ts deploy`, manual upload, or alternate CLI path can write that payload, inline plaintext can still reach production.

Clarify: is the gate at the command boundary sufficient, or should it also live at the payload-write boundary (the function that serializes + uploads canonical TOML)?

docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
```diff
@@ -0,0 +1,537 @@
...
+
+Expected behavior:
+
+| Missing secret           | Runtime behavior                                                                                                                                             |
+| ------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `handlers[].password`    | Matching protected route must not bypass auth. It should deny access or return an auth/config error for that route.                                          |
+| `publisher.proxy_secret` | First-party proxy signing, encryption/decryption, and URL-rewrite behavior requiring the proxy secret should fail or be disabled. Unrelated routes continue. |
+| `edge_cookie.secret_key` | Edge Cookie generation/validation should fail or be disabled. Unrelated request handling continues.                                                          |
+
+The exact HTTP status codes can be implementation-specific, but behavior must not silently skip security checks or generate weak fallback secrets.
```
aram356, line 314: ❓ **question** — Fail-closed HTTP semantics: 401 vs 503?

"Deny access or return an auth/config error" is ambiguous. Today, [`crates/trusted-server-core/src/auth.rs`](../../../crates/trusted-server-core/src/auth.rs) returns 401 with `WWW-Authenticate` for failed basic auth. A probe receiving 401 can't distinguish "route exists, wrong creds" from "route is misconfigured"; 503/500 reveals "this route is broken". The choice has security (oracle attacks) and ops (paging signal) consequences.

Same question for `proxy_secret`-dependent endpoints (encrypt/decrypt URL): 4xx (treat as unsigned) or 5xx (refuse to serve)?

Lock specific status codes / failure modes into the spec; downstream callers will rely on them.

docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
```diff
@@ -0,0 +1,537 @@
...
+
+Provisioning uses this provider setting to create or locate the underlying Fastly Secret Store resource and bind it to a fixed runtime alias.
+
+### Fixed runtime alias
+
+The default application secret store uses a fixed runtime alias:
+
+```text
+ts_secrets
+```
```
aram356, line 210: ♻️ **refactor** — Alias naming inconsistency: `ts_secrets` vs `ts_config_store`.

Existing aliases: `ts_config_store` ([`crates/trusted-server-core/src/runtime_config.rs:22`](../../../crates/trusted-server-core/src/runtime_config.rs#L22)) for the application Config Store; `signing_keys`/`api-keys` for the existing Secret Stores. The new alias breaks both patterns: it's neither `ts_<resource>_store` (which would be `ts_secret_store`) nor an unprefixed purpose-name (which would be `app_secrets`/`application_secrets`).

Pick one convention and justify in the spec. `ts_secret_store` would parallel `ts_config_store` cleanly.

docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
```diff
@@ -0,0 +1,537 @@
...
+
+Fields remain visibly required. Tooling may later scaffold explicit refs, but omission does not imply a default.
+
+## Fastly provider configuration
+
+Provider-specific resource names remain outside canonical runtime application config. For Fastly, the source config may include:
+
+```toml
+[providers.fastly.secrets]
+store_name = "customer_ts_secrets"
```
aram356, line 199: ♻️ **refactor** — Provider group key inconsistency: `[providers.fastly.secrets]` vs the purpose-scoped pattern.

Existing CLI groups are purpose-scoped: `[providers.fastly.application_config]`, `[providers.fastly.request_signing]` ([`crates/trusted-server-cli/src/config.rs:43-50`](../../../crates/trusted-server-cli/src/config.rs#L43-L50)). `[providers.fastly.secrets]` is generic by contrast. Recommend `[providers.fastly.application_secrets]` to parallel `application_config` and leave room for future purpose-scoped secret groups.

docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
```diff
@@ -0,0 +1,537 @@
...
+```
+
+Rules:
+
+- `{ secret = "..." }` and `{ store = "...", key = "..." }` are the only supported ref shapes in v1.
+- `secret` must not be combined with `store` or `key`.
+- `key` must not be empty.
+- `store`, when present, must not be empty.
+- The `store` value is a runtime secret-store alias, not necessarily the provider's underlying resource name.
+
```
aram356, line 176: 🤔 **thinking** — Inline-table canonical form for refs.

TOML inline tables can be re-serialized as `{ ... }` or as expanded subtables with no semantic difference. The existing `retain_declared_fields` canonicalization in `runtime_config.rs` preserves declared field shape, but if refs must hash identically across formats, the spec needs to state the canonical layout (inline vs expanded; key order). Otherwise hash determinism breaks subtly when a formatter or editor rewrites the TOML. Couples with the shorthand-vs-explicit hash issue above.

docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
```diff
@@ -0,0 +1,537 @@
...
+proxy_secret = "dev-proxy-secret"
+
+[edge_cookie]
+secret_key = "dev-edge-cookie-secret"
+
+[[handlers]]
+path = "^/admin"
+username = "admin"
+password = "dev-password"
+```
```
aram356, line 134: 🤔 **thinking** — Inline string that looks like a ref path.

`proxy_secret = "publisher/proxy_secret"` is a valid inline dev secret string that happens to mimic a ref path. An operator who forgets the `{ secret = ... }` wrapper writes a plaintext secret whose value is the literal path, and parsing accepts it silently.

Spec should require `ts config validate` to flag inline values that look like ref paths (heuristic: contains `/`, no whitespace, matches a path-like regex) — at minimum as a warning.

docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
```diff
@@ -0,0 +1,537 @@
...
+  status: present
+
+- edge_cookie.secret_key -> ts_secrets/edge_cookie/secret_key
+  status: missing
+  action: create the secret before apply
+
+- handlers[0].password -> ts_secrets/handlers/admin/password
+  status: missing
+  action: create the secret before apply
+```
```
aram356, line 363: 🤔 **thinking** — Plan-output `status: missing` per secret item depends on the underlying API.

Some secret-store backends (including some Fastly Secret Store visibility/policy modes) expose item names but never values, and some don't enumerate items at all. The spec should specify which API surface provisioning relies on (list-items metadata vs read-then-test), so `status: missing` accuracy is bounded by the documented Fastly API contract rather than assumed.

docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
```diff
@@ -0,0 +1,537 @@
...
+1. Parse and validate the canonical TOML payload.
+2. Identify all v1 secret refs.
+3. Attempt to resolve refs through `RuntimeServices.secret_store()`.
+4. Store resolved redacted values when available.
+5. Store unavailable markers when refs cannot be resolved.
+6. Log a warning or error for each unavailable secret ref.
+7. Continue serving with the resolved snapshot.
+
+Missing secret refs at runtime must not make the entire service fail config loading. This protects production availability when a secret is deleted or changed out-of-band through the Fastly UI/CLI after a successful provisioning flow.
+
```
aram356, line 301: 🤔 **thinking** — "Eager but non-fatal" + no retry = persistent fail-closed until next snapshot.

The runtime section reads as decisive ("eager per loaded config snapshot"); open question #6 punts retry. Operationally this is "affected feature stays fail-closed indefinitely until the next config refresh / cold start." That's defensible for v1, but the spec should state this explicitly here instead of leaving it to readers to infer from the open question list.

docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
```diff
@@ -0,0 +1,537 @@
...
+    Unavailable { reference: SecretRef },
+}
+```
+
+Names are illustrative. The important distinction is:
+
+- parsed/raw settings can contain inline values or refs
+- resolved settings know whether each secret is available
+- logging/debug output never prints resolved secret values
+
```
aram356, line 274: 📝 **note** — Debug/Display semantics for `SecretRef`.

`Redacted<T>` redacts in Debug/Display. `SecretRef` carries `store`/`key` strings, which are operational metadata, not secret material — but readers will reasonably wonder whether they leak in logs. Spec should state explicitly that `SecretRef` Debug/Display is non-redacting (paths only), and `ResolvedSecretString::Unavailable { reference }` is safe to log.

docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
```diff
@@ -0,0 +1,537 @@
...
+## Validation rules
+
+### Parser/schema validation
+
+- V1 secret-bearing fields accept either a string or one supported ref table shape.
+- Unknown keys inside ref tables are rejected.
+- Empty strings are rejected for inline secret values.
+- Empty `secret`, `store`, or `key` values are rejected.
+- Mixed ref shapes are rejected.
```
aram356, line 444: 📝 **note** — Tie `SecretString` validation into the existing `validator::Validate` pattern.

Existing fields use `validator::Validate` with custom validators like `validate_redacted_not_empty` ([`crates/trusted-server-core/src/settings.rs:332-335`](../../../crates/trusted-server-core/src/settings.rs#L332-L335)) and `EdgeCookie::validate_secret_key` ([`crates/trusted-server-core/src/settings.rs:282-286`](../../../crates/trusted-server-core/src/settings.rs#L282-L286)). The spec should note that `SecretString` validation slots into this pattern (custom validator for `SecretString::Inline` empty-check, parse-time rejection for ref shape errors) rather than introducing a parallel validation framework.

docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
```diff
@@ -0,0 +1,537 @@
...
+```
+
+Local runtime behavior should use the same resolution semantics as production:
+
+- inline values are immediately available
+- refs resolve through the configured local/platform secret-store abstraction
+- missing refs log warnings and mark the field unavailable
+- affected features fail closed
+
+`ts config validate` may warn when inline v1 secrets are present, but inline dev config remains valid.
```
aram356, line 434: ⛏ **nitpick** — "may warn" should be "warns".

Production *rejects* inline v1 secrets; local *may warn* is too soft. Make the local-dev contract a definite warning (still permitting the inline value) so the two layers tell the same story.

docs/superpowers/specs/2026-05-19-secret-store-config-references-design.md
```diff
@@ -0,0 +1,537 @@
...
+- [ ] Resolve refs eagerly per loaded config snapshot without failing the entire service for missing runtime secrets.
+- [ ] Log missing runtime secrets clearly.
+- [ ] Ensure affected runtime features fail closed when a required secret is unavailable.
+- [ ] Reject inline v1 secrets during Fastly production provisioning before mutation.
+- [ ] Verify referenced secret stores and secret items during Fastly provisioning.
+- [ ] Fail provisioning before mutation when required refs are missing.
+- [ ] Do not add automatic secret generation in v1.
+- [ ] Do not add `handlers.id` in v1.
+- [ ] Do not migrate request-signing secrets into the generic secret-ref model in v1.
+- [ ] Document local/dev inline-secret support and production ref requirements.
```
aram356, line 526: ⛏ **nitpick** — Acceptance Criteria gaps.

Add checkboxes for:

- the ref canonicalization rule (matches the hash-determinism wrench)
- extending `validate_distinct_names` to include `providers.fastly.secrets.store_name`
- rejecting refs whose explicit `store` alias has not been provisioned (mentioned in §"Provisioning validation" body, missing from criteria)

aram356 (changes requested):
## Summary

Docs-only PR adding a v1 design spec for secret-store backed config references. The proposal is well-scoped (no auto-generation, no `handlers.id`, no request-signing migration) and the inline/ref dichotomy with strict production gating is a clean separation. Two correctness gaps and two open semantics questions block before merge; the rest are improvements to harden the spec against ambiguous implementation choices.

## Blocking

### 🔧 wrench

- **Hash determinism**: shorthand and explicit-default refs hash differently for the same logical secret (see inline at §"Config hashing and canonicalization").
- **Distinct-names validation**: `providers.fastly.secrets.store_name` must be reconciled with the CLI's existing `validate_distinct_names` (see inline at §"Fastly provider configuration").

### ❓ question

- **Production boundary**: which exact code path enforces inline-secret rejection at the canonical-payload write boundary, not just the `apply` command boundary?
- **Fail-closed HTTP semantics**: 401 vs 503 (and the same for `proxy_secret`-backed endpoints)?

## Non-blocking

### 📌 out of scope

- **`handlers.id` deferral**: open question #3 leaves `handlers.id` for a later version. Without it, the operator-chosen secret key is the only stable link between a `[[handlers]]` entry and its secret; reordering the TOML is fine, but renames/duplicates break the link. Worth a follow-up issue so v1.x can decide stability semantics before the user surface ossifies.

## CI Status

- integration tests: PASS

aram356:
Closing in place of Edgezero handling secret references https://github.com/stackpop/edgezero/pull/269

PR #724, opened 2026-05-20
Spec: Server side ad templates creative-opportunities cli

### LINKED ISSUE #701: Build a standalone validation tool for configuring creative opportunities
## Context

Current validation of `creative-opportunities.toml` is minimal:

- **Build-time only:** slot ID charset check (`^[A-Za-z0-9_-]+$`) in `build.rs`
- `validate_slot_id()` exists in `creative_opportunities.rs` but is not called at runtime
- No CI check, no standalone tool, no auditor-facing feedback

This creates friction for ad ops and JS auditors (raised by Aram) who need to validate slot config changes without running a full Rust build.

## Missing validations

- Duplicate slot IDs
- Pattern overlap / unreachable patterns (e.g. `/*` shadows `/article/**`)
- Format size sanity (`width > 0`, `height > 0`)
- `floor_price >= 0`
- Bidder names against the `[integrations.prebid].bidders` list
- APS `slot_id` presence when APS is enabled
- PBS stored-request cross-check: does each slot ID have a corresponding entry in the configured PBS instance?

## Proposed tool

A `ts-validate` CLI (or `cargo xtask validate`) that:
1. Parses `creative-opportunities.toml` and runs all structural checks
2. Optionally probes PBS for stored-request existence (requires PBS URL in config)
3. Outputs a human-readable report suitable for ad ops review
4. Exits non-zero on any error — runnable in CI without a full WASM build

## Action

- Define the full validation rule set
- Implement as a standalone binary or xtask
- Add to CI pipeline alongside `cargo clippy` and `cargo test`

Parent: #677

prk-Jr:
## Summary

- Spec for `ts-config`, a standalone CLI that validates, matches, lints, and checks `creative-opportunities.toml` without requiring a Rust toolchain — directly addressing the JS auditor onboarding gap raised in #701
- Defines a two-tool architecture: Rust binary (validate/match/lint/check, mirrors runtime glob and slot-ID logic) and Node.js tool (generates slot config from live publisher pages via Playwright)
- Covers all design decisions, edge cases, error type scaffold, JSON output contracts, and concrete test shapes for both Phase 1A (Rust) and Phase 1B (Node.js)

## Changes

| File | Change |
| ---- | ------ |
| `docs/superpowers/specs/2026-05-19-ts-config-cli-design.md` | New: full design spec for `ts-config` CLI (1440 lines) |

## Closes

Closes #701

## Test plan

> Spec-only PR — no production code changed. WASM build and manual testing not applicable.

## Checklist

- [x] New code has tests


docs/superpowers/specs/2026-05-19-ts-config-cli-design.md
```diff
@@ -0,0 +1,2122 @@
...
+Two standalone tools with a shared schema understanding, no shared binary:
+
+```
+crates/trusted-server-cli/       ← Rust binary (native target, excluded from workspace)
+  src/main.rs                    ← ts-config validate / match / lint / check subcommands
+  src/validate.rs                ← validation rules
+  src/match_cmd.rs               ← glob matching
+  src/report.rs                  ← output formatting (human + JSON)
+
+packages/js-asset-auditor/       ← Node.js package (feature/js-asset-auditor branch)
```
ChristianPavilonis, line 93: the claude code package will be removed, in favor of the ts cli.

the ts cli / edgezero cli will have auditing  and config store primitives, so I'd hold off and rework this when those are in place

docs/superpowers/specs/2026-05-19-ts-config-cli-design.md
```diff
@@ -0,0 +1,2122 @@
...
+
+_Author · 2026-05-19_
+
+---
+
+## 1. Problem Statement
+
+`creative-opportunities.toml` is the runtime contract between Trusted Server and the
+publisher's ad stack. Every slot definition in this file drives three load-bearing
+systems simultaneously: URL-pattern matching at the edge (which pages fire auctions),
```
aram356, line 11: ⛏ **nitpick — "GPT slot definition injected into `<head>`"**

`creative-opportunities.toml` doesn't call `defineSlot`; the runtime injects `__ts_ad_slots` and the GPT integration consumes it client-side. As written, this implies the TOML schema drives GPT directly. Tighten the prose so readers don't form that mental model.

docs/superpowers/specs/2026-05-19-ts-config-cli-design.md
```diff
@@ -0,0 +1,2122 @@
...
+ts-config check    [--config PATH] <PATH> (--expected-slots SLOT_ID,... | --expect-no-slots) [--json]
+```
+
+`<PATH>` in `match` and `check` is a URL path (e.g. `/2024/01/my-article/`), not a
+full URL. Full URLs are accepted but only the path component is used after stripping
+the scheme and host.
+
+`$TS_CONFIG_PATH` is honored by **all subcommands** as the default config path when
+`--config` is not provided. `--config` takes precedence when specified. If neither
+`--config` nor `$TS_CONFIG_PATH` is set and no `./creative-opportunities.toml` exists,
```
aram356, line 138: ❓ **question — `--config` vs `$TS_CONFIG_PATH` conflict semantics undefined**

Spec says `--config` takes precedence when specified. But: if `--config /missing.toml` is passed *and* `$TS_CONFIG_PATH=/exists.toml` is set, does the tool fall back to `$TS_CONFIG_PATH`, or surface the explicit path's not-found error?

Standard convention is explicit > env with no fallback (so the not-found error is on `/missing.toml`), but the spec doesn't say so. State this explicitly — otherwise implementers will guess, and the behavior becomes hard to change later.

docs/superpowers/specs/2026-05-19-ts-config-cli-design.md
```diff
@@ -0,0 +1,2122 @@
...
+    },
+  })
+
+  // Immediate pass — handles the case where either library is already on window
+  // (e.g. synchronous script above the page fold) before addInitScript runs.
+  patchGoogletag()
+  patchCmdPush()
+  patchApstag()
+
+  const interval = setInterval(() => {
```
aram356, line 648: 🤔 **thinking — `setInterval` runs for `settle + 2000`ms with no early-exit**

With default `settle = 6000`, the interval runs ~8s × 20Hz = 160 fires. All patches are idempotent via `__ts_patched`, so this is wasted scheduler work — not a correctness bug. Cleaner pattern:

```javascript
const interval = setInterval(() => {
  patchGoogletag()
  patchCmdPush()
  patchApstag()
  if (allPatched()) clearInterval(interval)
}, 50)
```

…where `allPatched()` checks the three `__ts_patched` flags. Optional polish.

docs/superpowers/specs/2026-05-19-ts-config-cli-design.md
```diff
@@ -0,0 +1,2122 @@
...
+The property setter on `window.googletag` is the critical addition. Publishers
+universally write `window.googletag = window.googletag || { cmd: [] }` before
+GPT loads. The setter fires the moment the page assigns this stub, immediately
+patching `cmd.push` so any `cmd.push(fn)` calls that follow are captured — even
+if they happen before the first 50ms interval. The 50ms polling remains as a
+fallback for late-loading or dynamically injected GPT. `orig.apply(this, arguments)`
+is used for `cmd.push` (not `orig.call(this, fn)`) because GPT's `cmd.push`
+accepts multiple callbacks in some builds.
+
+**Layer 2: read-back after page settles**
```
aram356, line 666: 🌱 **seedling — no per-URL Playwright navigation timeout**

Spec calls `waitForTimeout(settle)` after `{ waitUntil: 'load' }` but doesn't set a Playwright navigation timeout. Default is 30s — fine for one URL, but a multi-URL crawl with 10 dead URLs hangs CI for 5 minutes.

**Fix:** specify `page.setDefaultNavigationTimeout(navTimeout)` with a documented default (suggest 15s) and a `--nav-timeout` flag. Note in §4.2 and the CLI interface block.

docs/superpowers/specs/2026-05-19-ts-config-cli-design.md
```diff
@@ -0,0 +1,2122 @@
...
+
+The user can rename in the output file. The comment on the generated `id` line should
+cite which source was used: `# derived from semantic segments in div_id` or
+`# derived from unitPath`.
+
+**Collision resolution:** multiple placements can produce the same base slug (e.g.
+two homepage slots both deriving `publisher-homepage`). `ts-config validate` requires
+unique IDs, so the generator must guarantee uniqueness. Rule: if a slug is already
+taken, append a hex suffix derived from `sha256(unitPath + "\0" + divId)` — the null
+byte separator ensures `unitPath` and `divId` cannot be confused. Start with 6 hex
```
aram356, line 770: ♻️ **refactor — hash-suffix collision terminal condition unbounded**

"Lengthen by 2 hex chars at a time until unique" has no documented stopping point. In practice 6 hex (24 bits) handles every realistic case (birthday-bound collision probability is ~0.003% at 1000 slots), but the loop is theoretically unbounded.

**Fix:** specify an upper bound — e.g. "extend up to 16 hex (64 bits); error out if still colliding." That makes the error path documented rather than implicit, and keeps the generator deterministic in the (effectively unreachable) failure case.

docs/superpowers/specs/2026-05-19-ts-config-cli-design.md
```diff
@@ -0,0 +1,2122 @@
...
+
+1. **Exact token match** — `slotID === divId` (case-insensitive). If exactly one GPT
+   slot matches, that is the name candidate. Stop.
+2. **Semantic token match** — tokenize both `slotID` and `divId` by splitting on `-`,
+   `_`, and digit runs. If both share a token from the set
+   `{atf, btf, header, footer, sidebar, leaderboard, mrec, interstitial}`, that GPT
+   slot is a candidate. If exactly one candidate survives, that is the name candidate.
+   If zero or multiple survive, `name_candidate = None`.
+
+**Step B — compute `size_candidate`** (always, independent of step A): 3. **Size-unique fallback** — the APS `sizes` are a subset of exactly one GPT slot's
```
aram356, line 795: 🤔 **thinking — APS "no-overlap" condition is ambiguous**

"APS `sizes` are a subset of exactly one GPT slot's sizes AND no other GPT slot has a size that overlaps the APS sizes."

Worked case: APS `[[300,250]]`, Slot A `[[300,250],[728,90]]`, Slot B `[[300,250]]`. Both are supersets. Under the strict "any overlap defeats" reading, `size_candidate = None`. Under "only full-subset overlap defeats," `size_candidate = None` still (both are full supersets).

Now consider: APS `[[300,250],[728,90]]`, Slot A has both, Slot B has only `[300,250]`. Slot A is the unique full superset; Slot B overlaps in one size. Strict reading says `size_candidate = None` because B overlaps — but the intuitive answer is A.

**Fix:** add a worked example for the disambiguation case (the one where exactly one slot fully contains APS sizes and other slots partially overlap), so the implementation matches intent.

docs/superpowers/specs/2026-05-19-ts-config-cli-design.md
```diff
@@ -0,0 +1,2122 @@
...
+(`id.chars().all(|c| c.is_ascii_alphanumeric() || c == '_' || c == '-')`). Note:
+`build.rs` uses a regex (`^[A-Za-z0-9_\-]+$`) which is semantically equivalent, but
+the CLI mirrors `creative_opportunities.rs` — not `build.rs` — to stay aligned with
+the runtime matching authority. A comment in both files cross-references the other.
+
+The glob normalization behavior is reproduced exactly:
+
+```rust
+/// Result of matching a single pattern against a path.
+enum MatchResult {
```
aram356, line 1121: ♻️ **refactor — `MatchResult` enum exists only in the CLI; exposes drift surface**

Runtime `matches_path` returns plain `bool` ([creative_opportunities.rs:81](crates/trusted-server-core/src/creative_opportunities.rs#L81-L93)). This spec introduces a richer `MatchResult { Match, NormalisedMatch, UncompilablePattern }` only in the CLI. That's exactly the drift-prone surface §4.4 warns about — every future change to runtime matching semantics has to be mirrored manually here, and the round-trip test only catches drift exercised by the production config.

**Fix:** expose `matches_path_with_normalisation` from `creative_opportunities.rs` as the canonical entry point (returning the enum); have runtime call a `.is_match()` / `.was_match()` helper on it; let the CLI consume the same enum directly. If `wasm32-wasip1` blocks sharing the module today (per §4.4's EdgeZero note), document that blocker explicitly — but don't introduce a CLI-only variant of the type that *defines* matching behavior.

docs/superpowers/specs/2026-05-19-ts-config-cli-design.md
```diff
@@ -0,0 +1,2122 @@
...
+
+Validation rules in `ts-config validate` are derived from
+`crates/trusted-server-core/src/creative_opportunities.rs` and
+`crates/trusted-server-core/build.rs`. Rule drift between the CLI and the runtime is
+a bug.
+
+**Intentional asymmetry:** `build.rs` validates slot IDs only (rule 3 above). The CLI
+validates all 11 rules. This is intentional — the CLI is a richer pre-flight, not a
+clone of `build.rs`. The round-trip CI test confirms that the checked-in
+`creative-opportunities.toml` passes both tools, but does not prove that the two tools
```
aram356, line 1106: 🔧 **wrench — copy-paste drift defended only by comments and single-config round-trip test**

The spec correctly identifies that the CLI duplicates `validate_slot_id` and `matches_path_with_normalisation` from `creative_opportunities.rs`. The defense is:

1. `// NOTE: keep in sync` comments in both files
2. A round-trip test that runs the CLI against the production `creative-opportunities.toml` (currently 3 slots, all clean)

The round-trip test cannot detect drift in any rule the production config doesn't exercise. The spec mentions "the test fixture matrix in §9... should be run against both the CLI and the equivalent runtime logic in `creative_opportunities.rs` tests" — but only as aspirational future work, not Phase 1A scope.

**Fix:** promote that fixture-matrix proposal to Phase 1A scope. Concrete shape:

- `tests/fixtures/parity/<kind>.toml` for each `kind` value in §6 (one fixture per `invalid_slot_id`, `uncompilable_pattern`, `missing_leading_slash`, `non_string_targeting`, etc.)
- A shared loader in both `creative_opportunities.rs::tests` and the CLI's integration tests that iterates the directory and asserts identical pass/fail outcomes per fixture.

Without this, "CLI passes ≠ runtime accepts" can ship silently.

docs/superpowers/specs/2026-05-19-ts-config-cli-design.md
```diff
@@ -0,0 +1,2122 @@
...
+| `2`  | Input failure (file not found, invalid TOML, invalid path argument) |
+
+### JSON output contract
+
+All subcommands accept `--json` and emit a single JSON object on stdout. `schema_version: 1`
+allows downstream consumers to detect breaking changes. The `subcommand` field
+identifies which output shape to expect.
+
+All diagnostic entries carry: `slot` (ID string or `null` for file-level issues),
+`field` (full TOML path string, e.g. `"slots[0].page_patterns[0]"`), `kind` (machine-readable tag),
```
aram356, line 1383: ♻️ **refactor — kind-specific JSON fields not documented in the contract**

§4.1 examples show `glob_normalised` warnings carrying `original` and `effective` keys ([line 222-227](docs/superpowers/specs/2026-05-19-ts-config-cli-design.md#L222-L227)). The §6 contract states all entries carry `slot/field/kind/message` — implying that's the complete shape.

**Fix:** either (a) enumerate kind-specific fields per `kind` value in a small table, or (b) add an explicit "additional kind-specific keys may be present" clause and pin which kinds carry which extras. As written, a downstream JSON consumer building strictly against §6 will silently miss the `original`/`effective` fields and break when they appear.

docs/superpowers/specs/2026-05-19-ts-config-cli-design.md
```diff
@@ -0,0 +1,2122 @@
...
+before the 50ms polling interval fires. The Layer 1 shim patches `googletag.cmd.push`
+to call `patchGoogletag()` before each queued callback executes, closing this race.
+Publishers that call `defineSlot` directly (not inside `cmd.push`) are also covered:
+`patchGoogletag` installs a property setter on `googletag.defineSlot` when the function
+is not yet present, so the moment GPT attaches `defineSlot` to the object the setter
+fires, wraps it with a value descriptor, and all subsequent direct calls are captured
+regardless of when the next 50ms tick fires.
+
+**Pattern glob normalization: `/20**`→`/20*`** — `/20**`follows the same rule as`/b**`: the `**` immediately follows a non-separator character (`0`), which is invalid
+glob syntax. `Pattern::new("/20**")`fails; the normalization branch fires and replaces
```
aram356, line 1479: ⛏ **nitpick — malformed markdown escapes**

This paragraph mixes literal underscores, escaped asterisks, and code spans in ways that won't render cleanly:

> `/20**`follows the same rule as`/b**`: the `**` immediately follows … `/20\*\*`get the correct match outcome … the generated output uses`/20_`directly. The`equivalent_pattern`lint fires if both`/20\*_`and`/20_`appear

Intent reads, but the rendered HTML will mangle the underscores and asterisks. Reformat with consistent code spans (`/20**`, `/20*`, etc.) and proper spacing around them.

docs/superpowers/specs/2026-05-19-ts-config-cli-design.md
```diff
@@ -0,0 +1,2122 @@
...
+
+8. `cd packages/js-asset-auditor && node --test test/generate-slots.test.mjs` → all tests pass
+9. Playwright integration test: crawl `fake-publisher.html` (sync GPT load) → TOML has `id = "test-atf"` and `id = "test-btf"` with correct formats
+10. Playwright integration test: crawl `fake-publisher-pre-gpt.html` (async GPT, pre-push) → same slots captured via property-setter path
+11. Playwright integration test: crawl `fake-publisher-aps.html` (late `fetchBids`) → APS `slot_id` present in output TOML
+12. Playwright integration test: crawl `fake-publisher-direct.html` (direct `defineSlot`, no `cmd.push`) → same slot IDs and formats as fixture A, confirming Layer 1 `patchGoogletag` `defineSlot` setter fires
+
+### Manual (local only — requires network + browser)
+
+8. `./target/"$HOST"/debug/ts-config match /2024/01/my-article/` → matches `atf_sidebar_ad` only
```
aram356, line 2117: ⛏ **nitpick — verification step numbering collides**

Phase 1B CI uses steps 8–12 (lines 2109-2113); Manual then restarts at 8 (this line). Renumber Manual as 13–17 so the section reads top-to-bottom without ambiguity.

aram356 (changes requested):
## Summary

Spec is detailed and well-iterated, but two structural concerns make this hard to land as-is: (1) it appears to be superseded by [#669 "Add the Trusted Server CLI"](https://github.com/IABTechLab/trusted-server/pull/669), which targets `main` and introduces a `ts config` namespace; (2) the PR is stacked on `server-side-ad-templates-impl` rather than `main`, so the claim of "Phase 1A ships as PR A independently" doesn't actually hold. Several spec-internal issues are flagged inline.

## Blocking

### 🔧 wrench

- **Spec collides with PR #669 (`ts` CLI)**: PR #669 introduces a `ts` CLI with `ts config` and `ts audit` subcommands, targeting `main` and currently open. This spec proposes a *separate* `ts-config` binary in a *separate* `crates/trusted-server-cli/` crate. Christian's prior inline comment ("hold off and rework this when [the ts cli / edgezero cli] are in place") points to the same collision. Either rework so the four subcommands are designed as `ts config validate|match|lint|check` under the namespace from #669, or explicitly cite #669 in the spec and justify why two CLI binaries are intentional.
- **PR base is `server-side-ad-templates-impl`, not `main`** (PR metadata): A spec-only PR should be reviewable and mergeable independently. As stacked, this PR can't land until the much larger feature branch lands first, and reviewers have to mentally subtract the 30+ unrelated files in the base diff to see this PR's actual content. The PR body itself claims "Phase 1A ships as PR A independently" — that's contradicted by the stacked base. Rebase onto `main`.
- **Copy-paste drift defense is insufficient** (inline at §4.4): see inline comment.

### ❓ question

- **Why a top-level `ts-config` binary rather than `ts config <subcmd>`?** Given #669 establishes the `ts` namespace, this needs an answer in the spec. If the answer is "because EdgeZero/`ts` isn't merged yet," document the migration path so we don't ship a binary that's renamed/removed within one release cycle.
- **`--config` vs `$TS_CONFIG_PATH` conflict semantics** (inline at §4.1): see inline comment.

## Non-blocking

### 🤔 thinking

- **2122-line spec for a Phase 1 CLI** (file-level): actual binary scope (4 subcommands, ~12 rules, ~600 lines of Rust) is much smaller than the spec implies. ~140 lines of JavaScript shim and four full HTML fixtures in §4.2 belong in the Node.js package's docs once implementation lands. Recommend splitting into a Phase 1A spec (~700 lines, Rust binary) and Phase 1B spec (~900 lines, Node.js generator) so each phase is independently reviewable.
- **APS \"no-overlap\" wording is ambiguous** (inline at §4.2 lines 795-797): see inline comment.
- **`setInterval` runs without early-exit** (inline at §4.2 lines 648-654): see inline comment.

### ♻️ refactor

- **`MatchResult` enum exists only in the CLI** (inline at §4.4 line 1121): see inline comment.
- **JSON contract omits kind-specific fields** (inline at §6 line 1383): see inline comment.
- **Hash-suffix collision terminal condition unbounded** (inline at §4.2 line 770): see inline comment.
- **No stable ordering for `errors[]` / `warnings[]`** (cross-cutting): HashMap iteration order means two CI runs against the same file produce diff churn. Specify lexicographic ordering by `field` then `kind`, or by slot-file order then field, in the §6 JSON contract. Diff-friendly output matters once this lands in CI.
- **`lint` output dedup vs `validate` rules** (cross-cutting §4.1, §6): `lint`'s `kind` enumeration is a superset of `validate`. If a file has an invalid slot ID, both should report it once — but spec is silent on whether `lint --json` puts validate-level errors in `errors[]` or `warnings[]`. Recommend: validate-level errors always in `errors[]`; lint-only findings always in `warnings[]`; no duplication.

### 🌱 seedling

- **No per-URL Playwright navigation timeout** (inline at §4.2 line 666): see inline comment.

### 📌 out of scope

- **Runtime should also enforce config invariants once KV-backed config lands**: today `matches_path` silently treats an uncompilable pattern as `false` ([creative_opportunities.rs:84-91](crates/trusted-server-core/src/creative_opportunities.rs#L84-L91)) — fine while config is baked into the WASM binary via `include_str!()`, but harmful once it comes from KV (mentioned as a non-goal in §3). The CLI is pre-flight; runtime needs its own enforcement once that surface opens.
- **Confirm generator's slug derivation is a strict subset of `[A-Za-z0-9_-]+`**: §7 says slot IDs are an HTML-injection safety surface. The validate step will catch any drift, but a one-line invariant statement in §4.2 ("tokenizer output is a strict subset of `[A-Za-z0-9_-]+` by construction") would close the loop and let reviewers verify by inspection.

### ⛏ nitpick

- **§1 prose: "GPT slot definition injected into `<head>`"** (inline at line 11)
- **Malformed markdown at §8 \"Pattern glob normalization\"** (inline at line 1479)
- **§11 verification step numbering collides** (inline at line 2117)

## CI Status

- browser integration tests: **PASS** (1m54s)
- integration tests: **PASS** (1m34s)
- prepare integration artifacts: **PASS** (2m1s)

prk-Jr:
Closing as superseded by #823. The standalone creative-opportunities CLI spec is stale now that configuration flows through the unified `ts` CLI from #799 and effective `trusted-server.toml` / env overlay semantics.

Issue #495, opened 2026-03-12
Fastly entry point switch (dual-path with flag)


aram356:
## Description

Add feature-flagged dual-path entry point: `edgezero_enabled` config flag with `unwrap_or(false)` safe default. EdgeZero path dispatches through `TrustedServerApp::build_app()`; legacy path preserved as `legacy_main()`.

## Done when

- Both paths compile
- Legacy cleanup issue filed
- Per-PR gates pass

## Blocked by

PR 13

## References

- `docs/internal/EDGEZERO_MIGRATION.md` — Phase 3, PR 14
- Part of #480

PR #727, opened 2026-05-21
Add EdgeZero canary routing (rollout_pct flag + FNV-1a bucket assignment) (PR 19)

### LINKED ISSUE #500: Cutover and canary rollout
## Description

Deploy EdgeZero-backed Fastly entry point behind `edgezero_enabled` config flag. Run canary traffic progression: 1% → 10% → 50% → 100% with hold points. Validate rollback by setting `edgezero_rollout_pct` to 0.

## Done when

- Config flag live
- Canary at 1% → 10% → 50% → 100% with hold points
- Rollback tested

## Blocked by

Phase 5: Verification gates

## References

- `docs/internal/EDGEZERO_MIGRATION.md` — Phase 5: Cutover plan
- Part of #480

prk-Jr:
## Summary

- Adds `edgezero_rollout_pct` (0–100) to the `trusted_server_config` store so EdgeZero traffic can be shifted incrementally without a deploy
- Uses FNV-1a 32-bit hash of `client_ip` for deterministic, sticky per-user bucket assignment (0–99); routing predicate is `bucket < rollout_pct`
- Fail-safe defaults: read error or invalid value → 0 (all legacy); absent key (backward compat) → 100 (all EdgeZero); `rollout_pct = 0` is the immediate no-deploy rollback

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-adapter-fastly/src/main.rs` | Added `parse_rollout_pct`, `fnv1a_bucket`, `canary_routes_to_edgezero`, `read_rollout_pct`; updated `main()` with canary dispatch; 11 unit tests (pinned FNV-1a vectors, distribution, boundary routing) |
| `fastly.toml` | Added `edgezero_rollout_pct = "0"` to local Viceroy config store with ops comment |
| `docs/internal/EDGEZERO_MIGRATION.md` | New ops runbook: config store keys, failure modes, pre-flight activation, 4 canary stages, pass/fail thresholds, rollback procedure, monitoring |

## Closes

Closes [#500](https://github.com/IABTechLab/trusted-server/issues/500)

## Test plan

- [x] `cargo test-fastly` — 65 + 956 + 21 + 3 tests green
- [x] `cargo test-axum` — 31 tests green
- [x] `cargo clippy-fastly && cargo clippy-axum` — no warnings
- [x] `cargo fmt --all -- --check` — clean
- [x] JS tests: `cd crates/js/lib && npx vitest run` — 291 tests green
- [x] JS format: `cd crates/js/lib && npm run format` — clean
- [x] Docs format: `cd docs && npm run format` — clean
- [ ] Manual testing via `fastly compute serve` with `edgezero_rollout_pct = "1"` and `"0"` (rollback)

## Checklist

- [x] New code has tests


crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -113,6 +158,42 @@ fn is_edgezero_enabled(config_store: &ConfigStoreHandle) -> Result<bool, fastly:
...
+                    value
+                );
+                0
+            }
+        },
+        Ok(None) => {
+            // Fires per-request when key is absent and edgezero_enabled=true.
+            // At production scale this creates one warn per request until the key is set.
+            // Resolution: set edgezero_rollout_pct = "0" before setting edgezero_enabled = "true".
+            log::warn!(
```
aram356: 🔧 **wrench** — Per-request warn log on absent `edgezero_rollout_pct`.

This emits a `log::warn!` for **every request** when the key is absent. The code comment acknowledges "At production scale this creates one warn per request until the key is set" but does not fix it. Combined with the absent-key default = 100% rollout, this is a real operational hazard: a single ops ordering mistake (setting `edgezero_enabled = "true"` before `edgezero_rollout_pct`) results in **both** an immediate full cutover **and** a sustained warn flood. The warn doesn't tell anyone anything is broken (traffic flows fine), so it's likely to be silenced/ignored — not acted on.

**Fix (pick one):**
- Demote to `log::debug!` — the runbook + setup procedure is the real safety net, not this log line:
  ```rust
  Ok(None) => {
      log::debug!(
          "edgezero_rollout_pct key absent, defaulting to 100 (full rollout — backward compat)"
      );
      100
  }
  ```
- Better: change the absent-key default to `0` (fail-safe) — see the design question in the review body — and drop the warn entirely. An absent key would then surface as zero canary traffic, observable from real-time stats, rather than as a log flood.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -146,14 +227,36 @@ fn main() {
...
         log::warn!("failed to read edgezero_enabled flag, falling back to legacy path: {e}");
         false
     }) {
-        log::debug!("routing request through EdgeZero path");
+        log::debug!("routing request through legacy path (edgezero_enabled=false)");
+        legacy_main(req);
+        return;
+    }
+
+    let rollout_pct = read_rollout_pct(&edgezero_config_store);
```
aram356: 🤔 **thinking** — Two config-store reads per request.

Each request now hits `config_store.get(...)` twice (`is_edgezero_enabled` then `read_rollout_pct`). Fastly Config Store reads are local-edge and fast, but at full QPS that's a measurable doubling. Consider consolidating into one helper that returns `(enabled, rollout_pct)` after a single read sequence, or wrapping both behind a cached snapshot if the SDK supports it. Not urgent — flag only.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -113,6 +158,42 @@ fn is_edgezero_enabled(config_store: &ConfigStoreHandle) -> Result<bool, fastly:
...
 
+/// Reads `edgezero_rollout_pct` from the config store.
+///
+/// | Config store state              | Return value | Effect                     |
+/// |---------------------------------|--------------|----------------------------|
+/// | Key absent                      | `100`        | Full rollout (backward compat) |
+/// | Key present, valid 0–100        | parsed value | Partial or full rollout    |
+/// | Key present, invalid            | `0`          | All legacy (safe default)  |
+/// | Key read error                  | `0`          | All legacy (safe default)  |
+fn read_rollout_pct(config_store: &ConfigStoreHandle) -> u8 {
```
aram356, line 202: ♻️ **refactor** — Add unit tests for `read_rollout_pct` using an in-memory `ConfigStore`.

The plan asserts `ConfigStoreHandle` cannot be constructed in unit tests; this is not accurate at the pinned `edgezero-core` rev. The `ConfigStore` trait is public, and `ConfigStoreHandle::new(Arc::new(...))` is public — edgezero-core's own tests use exactly this pattern.

The four documented branches are safety-critical; only the first is indirectly covered (via `parse_rollout_pct` tests).

**Suggested addition** in the existing `#[cfg(test)] mod tests`:

```rust
use edgezero_core::config_store::{ConfigStore, ConfigStoreError};

struct TestConfigStore {
    response: Result<Option<String>, ConfigStoreError>,
}

impl ConfigStore for TestConfigStore {
    fn get(&self, _key: &str) -> Result<Option<String>, ConfigStoreError> {
        self.response.clone()
    }
}

fn handle(response: Result<Option<String>, ConfigStoreError>) -> ConfigStoreHandle {
    ConfigStoreHandle::new(Arc::new(TestConfigStore { response }))
}

#[test]
fn read_rollout_pct_absent_defaults_to_full_rollout() {
    assert_eq!(read_rollout_pct(&handle(Ok(None))), 100);
}

#[test]
fn read_rollout_pct_invalid_value_defaults_to_zero() {
    assert_eq!(read_rollout_pct(&handle(Ok(Some("abc".into())))), 0);
}

#[test]
fn read_rollout_pct_out_of_range_defaults_to_zero() {
    assert_eq!(read_rollout_pct(&handle(Ok(Some("101".into())))), 0);
}

#[test]
fn read_rollout_pct_read_error_defaults_to_zero() {
    assert_eq!(
        read_rollout_pct(&handle(Err(ConfigStoreError::unavailable("boom")))),
        0,
    );
}
```

These tests pin the safety-critical defaults under refactoring — the part of the design that *matters* in incidents.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -146,14 +227,36 @@ fn main() {
...
+    let routing_key = match req.get_client_ip_addr() {
+        Some(ip) => ip.to_string(),
+        None => {
+            log::debug!(
+                "no client IP available, using empty routing key (deterministic bucket 61)"
+            );
+            String::new()
+        }
+    };
+    let bucket = fnv1a_bucket(&routing_key);
```
aram356: 🌱 **seedling** — Skip the FNV hash + `IpAddr.to_string()` allocation when `rollout_pct == 0 || == 100`.

`routing_key = ip.to_string()` allocates a `String` per request, and the FNV computation runs even when the decision is degenerate (0 → always legacy, 100 → always EdgeZero). For steady-state rollback (0) and post-cutover (100) — which together cover most of the canary's lifetime — this work is wasted.

```rust
let rollout_pct = read_rollout_pct(&edgezero_config_store);

let route_to_edgezero = match rollout_pct {
    0 => false,
    100 => true,
    pct => {
        let routing_key = req.get_client_ip_addr()
            .map(|ip| ip.to_string())
            .unwrap_or_default();
        let bucket = fnv1a_bucket(&routing_key);
        canary_routes_to_edgezero(bucket, pct)
    }
};
```

If you want to skip the `to_string()` allocation in the canary path too, FNV-1a can be fed `IpAddr` octets directly:
```rust
fn fnv1a_bucket_octets(bytes: &[u8]) -> u8 { /* same algorithm */ }
let bucket = match req.get_client_ip_addr() {
    Some(IpAddr::V4(v)) => fnv1a_bucket_octets(&v.octets()),
    Some(IpAddr::V6(v)) => fnv1a_bucket_octets(&v.octets()),
    None => fnv1a_bucket_octets(&[]),
};
```

Not urgent.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -85,6 +86,50 @@ fn parse_edgezero_flag(value: &str) -> bool {
...
+    }
+    (hash % 100) as u8
+}
+
+/// Returns `true` if the given bucket should be routed to the `EdgeZero` path.
+///
+/// `bucket` must be in `0..100`; `rollout_pct` in `0..=100`.
+/// When `rollout_pct = 0` no bucket ever routes to `EdgeZero` (instant rollback).
+/// When `rollout_pct = 100` every bucket routes to `EdgeZero` (full cutover).
+fn canary_routes_to_edgezero(bucket: u8, rollout_pct: u8) -> bool {
```
aram356: ⛏ **nitpick** — `canary_routes_to_edgezero` is just `bucket < rollout_pct`; "canary" is misleading at the call site.

The function works at any `rollout_pct` (0, 50, 100), not only canary stages. Reads awkwardly at the call site: `if canary_routes_to_edgezero(bucket, rollout_pct)`.

Suggest `routes_to_edgezero(bucket, rollout_pct)` — shorter and accurate at any rollout value.

aram356 (changes requested):

docs/internal/EDGEZERO_MIGRATION.md
```diff
@@ -0,0 +1,140 @@
...
+> (100) kicks in immediately, routing all traffic to EdgeZero without a staged canary.
+
+### Pre-flight activation
+
+Before advancing any stage, activate the canary switch:
+
+1. Confirm `edgezero_rollout_pct = "0"` is already set in the production config store
+   (set it now if not — the pre-condition above explains why this must come first).
+2. Set `edgezero_enabled = "true"` in the production config store.
+3. Verify via log tailing that `routing request through legacy path (bucket=N, rollout_pct=0)`
```
ChristianPavilonis: **P3: Clarify that these verification logs require debug-level logging**

This is useful operational guidance, but the referenced `routing request through ...` messages are emitted with `log::debug!` in the Fastly adapter while the normal Fastly logger is capped at `Info`. If operators are expected to verify this via log tailing, it may be worth noting that this requires a debug/log-level deployment or another equivalent canary observability signal.

ChristianPavilonis (commented):
## Summary

One low-severity docs/ops note: the canary runbook references routing log lines that are emitted at debug level, so production verification may require a debug/log-level deployment or equivalent observability setup.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Fixed in 9743550c. Demoted the absent-key log from `log::warn!` to `log::debug!` so it no longer floods at production QPS. Kept the 100%-on-absent default (documented backward-compat): changing it to a fail-safe 0 would silently roll back any existing `edgezero_enabled = "true"` deployment that has no `edgezero_rollout_pct` key (full EdgeZero → all legacy), so that path is left to the setup procedure + runbook as the safety net.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Acknowledged — leaving as-is. The two reads are two *distinct* keys (`edgezero_enabled` and `edgezero_rollout_pct`), so they cannot collapse into a single `config_store.get(...)`. The common pre-cutover state (`edgezero_enabled = "false"`) already does a single read and short-circuits to legacy before `read_rollout_pct` runs; the second read only happens once EdgeZero is enabled. A combined `(enabled, rollout_pct)` helper would either still issue two gets or read `rollout_pct` unnecessarily on the disabled path, so it would not reduce read count where it matters. Config Store reads are local-edge, and per the note this is non-urgent.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Fixed in 30002e2c. Added an in-memory `ConfigStore` stub via `ConfigStoreHandle::new(Arc::new(...))` and pinned all four branches: absent → 100, valid → parsed, invalid → 0, read error → 0 (plus a valid in-range value). One deviation from the suggested snippet: `ConfigStoreError` is not `Clone` at the pinned rev (it has an `Internal { source: AnyError }` variant), so the stub stores a small response enum and constructs the `Err` fresh inside `get` rather than cloning a stored `Result`.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Fixed in 7b420cc3. The routing decision now matches on `rollout_pct`: `0 => false`, `100 => true`, and only the partial-rollout arm allocates the client-IP routing key and runs the FNV hash. Steady-state rollback (0) and post-cutover (100) skip the per-request `to_string()` + hash entirely.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Fixed in 7b420cc3. Renamed `canary_routes_to_edgezero` → `routes_to_edgezero`; it is just `bucket < rollout_pct` and reads accurately at the call site for any rollout value.

docs/internal/EDGEZERO_MIGRATION.md
prk-Jr: Fixed in cbaa4a17. The runbook now states the routing log lines are emitted at `log::debug!` while the Fastly logger is capped at `Info`, so verifying via log tailing requires a debug-level deployment (or the Fastly real-time stats traffic split as an equivalent signal). Also reconciled the documented log format: after the degenerate-value optimization, `rollout_pct` 0 and 100 no longer compute a bucket, so those lines now read `routing request through {legacy,EdgeZero} path (rollout_pct=N)` without a `bucket=`.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -146,14 +228,52 @@ fn main() {
...
+        legacy_main(req);
+        return;
+    }
+
+    let rollout_pct = read_rollout_pct(&edgezero_config_store);
+
+    // Skip the per-request routing-key allocation and FNV hash for the degenerate
+    // rollout values (0 = full rollback, 100 = full cutover), which together cover
+    // most of the canary's lifetime. Only the partial-rollout path needs a bucket.
+    let route_to_edgezero = match rollout_pct {
```
ChristianPavilonis: Automated review: **P2: Add coverage for the entry-point canary dispatch, not just the helper functions.**

The safety-critical behavior is the wiring from `edgezero_enabled` + `read_rollout_pct()` + client-IP bucket into the final `edgezero_main` vs `legacy_main` call. The new tests pin parsing, hashing, the predicate, and `read_rollout_pct()` branches, but they do not exercise this decision matrix at the entry-point boundary (`enabled=false`, `rollout_pct=0`, `rollout_pct=100`, partial rollout bucket hit/miss, invalid/read-error fallback). A future refactor could accidentally invert the match, ignore `rollout_pct=0`, or route the absent/error cases incorrectly while the current helper tests still pass.

Suggested fix: extract the route-selection decision into a pure helper that takes the enabled state/rollout value/client IP (or a precomputed bucket) and unit-test the full matrix, then keep `main()` as thin I/O glue. If feasible, add a Viceroy smoke test for `edgezero_enabled=true` with `edgezero_rollout_pct=0` and `100` once there is an observable route marker.

ChristianPavilonis (commented):
Automated review:

## Review Summary

Reviewed PR #727 against `feature/edgezero-pr19-spin-adapter`, focusing on the Fastly entry-point canary routing, config-store failure behavior, rollout docs, and CI/review context. I did not find any blocking correctness, security, data-loss, authorization, or severe compatibility issues. I left one P2 inline comment about strengthening coverage for the safety-critical entry-point dispatch matrix.

## Findings

### P0 / Blockers

None.

### P1 / High

None.

### P2 / Medium

1 inline comment submitted.

### P3 / Low

None.

## CI / Existing Reviews

All reported PR checks are passing. Existing inline feedback about absent-key logging, config-store branch tests, degenerate rollout hashing, naming, and debug-level log documentation appears to have been addressed in later commits. Local `cargo test-fastly rollout -- --nocapture` could not run in this environment because Viceroy failed to satisfy `fastly_http_downstream::downstream_bot_analyzed`; CI's Fastly job is green.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Fixed in 722c6dc6. Extracted the dispatch decision into a pure `should_route_to_edgezero(rollout_pct, client_ip)` helper so the rollout-state → EdgeZero/legacy wiring is unit-tested directly, not just the parsing/hashing helpers it composes. `main()` keeps the `edgezero_enabled` gate and config reads as thin I/O glue.

Added tests covering the full matrix: `rollout_pct=0` (always legacy), `rollout_pct=100` (always EdgeZero), and partial-rollout bucket hit/miss for both present and absent client IPs (pinned FNV vectors: `1.2.3.4`→85, empty key→61).

Left `edgezero_enabled=false` as a trivial early return rather than folding it into the helper — doing so would force a `read_rollout_pct` config read even when disabled. The read-error fallback is covered by composition: `read_rollout_pct` fails safe to `0` (tested) and `should_route_to_edgezero(0, ..)` routes to legacy (tested).

docs/internal/EDGEZERO_MIGRATION.md
```diff
@@ -0,0 +1,150 @@
...
+
+- **Error rate:** `5xx / total_requests` by edge PoP
+- **Latency p95:** use log search for `routing request through EdgeZero path` to identify EdgeZero traffic (`x-edgezero-path` instrumentation header does not exist yet — follow-up task)
+- **Auction win-rate:** downstream SSP reporting, compare same-day prior week
+- **Timeout rate:** `504 / total_requests`
+
+> **The routing log lines are emitted at `log::debug!`, while the Fastly logger
+> is capped at `Info` in production.** Verifying the canary via log tailing
+> therefore requires a debug-level logging deployment (or another equivalent
+> observability signal, such as the Fastly real-time stats traffic split). With
```
ChristianPavilonis: Automated review: **P2: Make the production canary route signal explicit before relying on it operationally.**

This paragraph now correctly calls out that the route-decision logs are `debug!` while the Fastly logger is capped at `Info`, but the fallback it names (`Fastly real-time stats traffic split`) is not something this PR appears to create. The code only emits the EdgeZero/legacy decision at debug level, and the runbook also says the `x-edgezero-path` marker does not exist yet. Without a production-visible route marker (or an exact documented dashboard/query that already distinguishes the two branches), operators cannot reliably confirm that `rollout_pct=0` is actually all legacy, that 1%/10% traffic is flowing on the intended path, or attribute latency/error changes to the canary branch.

Suggested fix: either add/document a concrete production-safe signal before running the canary (for example a sampled/rate-limited info log field, a metrics/log field, or an internally gated diagnostic response header), or change the runbook to make a debug-level logging deployment an explicit precondition and remove the unsupported real-time-stats fallback unless there is a specific existing split/dashboard to reference.

ChristianPavilonis (commented):
Automated review:

## Review Summary

Reviewed PR #727 against `feature/edgezero-pr19-spin-adapter`, focusing on the Fastly canary dispatch, rollout percentage parsing/defaults, config-store failure behavior, tests, local Fastly config, and the migration runbook. I did not find any blocking correctness, security, data-loss, authorization, or severe compatibility issues. I left one P2 inline comment about making the production canary observability path explicit before operators rely on the runbook.

## Findings

### P0 / Blockers

None.

### P1 / High

None.

### P2 / Medium

1 inline comment submitted.

### P3 / Low

None.

## CI / Existing Reviews

All reported PR checks are passing. Existing review feedback about absent-key log volume, degenerate rollout hashing, config-store branch coverage, naming, and debug-level log documentation appears addressed by later commits. Local `cargo test-fastly rollout -- --nocapture` compiled the changed Fastly test binary but could not execute under the installed Viceroy because it lacks `fastly_http_downstream::downstream_bot_analyzed`; CI's Fastly test job is green.

docs/internal/EDGEZERO_MIGRATION.md
prk-Jr: Fixed in 706b077d (doc-only). You're right that neither named signal is usable in production — and the debug-logging fallback isn't either: `should_route_to_edgezero` logs the decision only at `debug!`, and `logging::init_logger` pins the Fastly logger to `Info` with no runtime override, so those lines never reach the production log endpoint (only local Viceroy). There's no `x-edgezero-path` marker and no real-time-stats split for this decision.

Rather than document a precondition the code can't satisfy, the runbook now states there is no production per-branch route signal yet, bases canary verification on aggregate service metrics tracking the staged `rollout_pct` changes, and scopes the debug route-decision lines to local pre-prod validation. Removed the real-time-stats fallback and the debug-logging-deployment guidance. A production-safe signal (e.g. the deferred `x-edgezero-path` header or a configurable log level) stays a follow-up.

docs/internal/EDGEZERO_MIGRATION.md
```diff
@@ -0,0 +1,165 @@
...
+Fastly real-time stats dashboard — aggregate service signals (not split by
+branch). Watch each as `rollout_pct` is increased stage by stage; a regression
+that appears and tracks the rollout steps implicates the EdgeZero branch:
+
+- **Error rate:** `5xx / total_requests` by edge PoP
+- **Latency p95:** service-wide
+- **Auction win-rate:** downstream SSP reporting, compare same-day prior week
+- **Timeout rate:** `504 / total_requests`
+
+> For local pre-production validation under Viceroy (where `debug!` is visible),
```
ChristianPavilonis: Automated review: **P2: The local Viceroy route-decision logs are still filtered out.**

This says the `debug!` route-decision lines are visible under local Viceroy, but `logging::init_logger()` builds the Fastly logger with `max_level(log::LevelFilter::Info)`. That level also applies to the `echo_stdout(true)` path, so the `log::debug!` calls in `should_route_to_edgezero()` will not be emitted locally unless the logger configuration changes.

Why it matters: the runbook's pre-production validation points operators at a branch-level signal they cannot actually observe, which can waste incident/cutover time and leave the canary validated only by aggregate metrics.

Suggested fix: either add a documented dev/test-only way to raise the Fastly logger to `Debug` for Viceroy validation, or remove/reword this local-log validation section to say these lines are currently code-only/test-only and not observable with the shipped logger configuration.

ChristianPavilonis (commented):
Automated review:

## Review Summary

Reviewed PR #727 against `feature/edgezero-pr19-spin-adapter`, focusing on the Fastly canary dispatch, rollout percentage parsing/defaults, config-store failure behavior, local Fastly config, tests, and the EdgeZero migration runbook. I did not find any blocking correctness, security, data-loss, authorization, or severe compatibility issues. I left one P2 inline comment about a runbook validation signal that is not actually emitted with the current logger configuration.

## Findings

### P0 / Blockers

None.

### P1 / High

None.

### P2 / Medium

1 inline comment submitted.

### P3 / Low

None.

## CI / Existing Reviews

All reported PR checks are passing. Existing review feedback about absent-key logging, degenerate rollout hashing, config-store branch coverage, dispatch-matrix tests, naming, and production route observability appears addressed in later commits. Local `cargo test-fastly rollout -- --nocapture` compiled the Fastly test binary but could not execute under the installed Viceroy 0.16.4 because it lacks `fastly_http_downstream::downstream_bot_analyzed`; CI's Fastly job is green.

docs/internal/EDGEZERO_MIGRATION.md
prk-Jr: Fixed in 5f81e80b. You're right — `init_logger` capped the Fastly logger at `Info`, which also gated the `echo_stdout(true)` path, so the `should_route_to_edgezero` `debug!` lines were emitted nowhere, including under Viceroy.

Took the dev/test-override option: `init_logger` now reads an `EDGEZERO_LOG_LEVEL` env var (resolved by a pure, unit-tested `resolve_max_level` helper) and defaults to `Info`. Local Viceroy runs can set `EDGEZERO_LOG_LEVEL=debug` (e.g. `EDGEZERO_LOG_LEVEL=debug fastly compute serve`) to surface the route-decision lines; production Fastly Compute does not surface arbitrary process env vars, so the level stays at the safe `Info` default there. Viceroy already reads process env (the local `api-keys` secret store uses `env = "FASTLY_KEY"`).

Reworded all three runbook references (Stage 0, Monitoring, and the local-validation list) to state the override explicitly rather than implying the lines are visible by default.

docs/superpowers/plans/2026-05-21-pr19-cutover-canary-rollout.md
```diff
@@ -0,0 +1,713 @@
...
+   full release cycle after reaching 100% with zero regressions.
+
+---
+
+## Monitoring
+
+Fastly real-time stats dashboard. Key signals at each canary stage:
+
+- **Error rate:** `5xx / total_requests` by edge PoP
+- **Latency p95:** use log search for `routing request through EdgeZero path` to identify EdgeZero traffic (`x-edgezero-path` instrumentation header does not exist yet — follow-up task)
```
ChristianPavilonis: Automated review: **P2: Update the duplicated monitoring guidance to match the runbook.**

This plan still embeds stale operational guidance telling operators to use Fastly log search / Fastly log tailing for `routing request through EdgeZero path` to identify EdgeZero traffic. The implemented logger defaults to `Info`, while the route-decision lines are emitted only at `debug!`, and the new runbook explicitly says those lines do not reach production and that no `x-edgezero-path`/real-time-stats branch split exists yet.

Why it matters: if someone uses this plan as the handoff/checklist during cutover, it gives false confidence that production has per-branch route attribution when it does not.

Suggested fix: update or remove this duplicated Monitoring section so it mirrors `docs/internal/EDGEZERO_MIGRATION.md`: production canary verification uses aggregate service metrics only, and route-decision logs are local Viceroy-only with `EDGEZERO_LOG_LEVEL=debug` until a production-safe branch signal is added.

ChristianPavilonis (commented):
Automated review:

## Review Summary

Reviewed PR #727 against `feature/edgezero-pr19-spin-adapter`, focusing on the Fastly canary routing path, rollout percentage parsing/fallbacks, logger-level changes, tests, local config, and EdgeZero migration documentation. I did not find any blocking correctness, security, data-loss, authorization, or severe compatibility issues. I left one P2 inline comment about stale operational guidance duplicated in the implementation plan.

## Findings

### P0 / Blockers

None.

### P1 / High

None.

### P2 / Medium

1 inline comment submitted.

### P3 / Low

None.

## CI / Existing Reviews

All reported PR checks are passing. Existing automated-review feedback about absent-key log volume, config-store branch coverage, degenerate rollout hashing, dispatch-matrix tests, production route observability, and local Viceroy debug logs appears addressed in later commits. Local `cargo test -p trusted-server-adapter-fastly --target wasm32-wasip1 rollout_pct --no-run` passed; a native `cargo test -p trusted-server-adapter-fastly rollout_pct -- --nocapture` attempt failed at link time on Fastly hostcall symbols, which is expected for the Fastly adapter outside the wasm/Viceroy path and not counted as a PR finding.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -135,6 +181,84 @@ fn is_edgezero_enabled(config_store: &ConfigStoreHandle) -> Result<bool, fastly:
...
+                0
+            }
+        },
+        Ok(None) => {
+            // Fires per-request when the key is absent and edgezero_enabled=true, so
+            // emit at debug to avoid a per-request warn flood at production QPS. The
+            // setup procedure (set edgezero_rollout_pct = "0" before edgezero_enabled
+            // = "true") and the migration runbook are the operational safety net here.
+            log::debug!(
+                "edgezero_rollout_pct key absent, defaulting to 100 (full rollout — backward compat)"
```
aram356: 🔧 **wrench** — Absent `edgezero_rollout_pct` defaults to `100` (fail-OPEN), inverting the fail-safe rule every other branch follows.

Every other failure mode in `read_rollout_pct` / `is_edgezero_enabled` resolves to the safe path: config store unreachable → legacy, `edgezero_enabled` unreadable → legacy, `rollout_pct` invalid → 0 (legacy), `rollout_pct` read error → 0 (legacy). Only this one case — `enabled = true` + key **absent** — resolves to **100% EdgeZero**, the maximum blast radius. A read *error* and an *absent* key are operationally the same failure class ("config isn't what I expect"), yet they produce opposite, maximally-divergent outcomes.

The mitigation chosen for this is *procedural*: a runbook precondition ("set `rollout_pct = 0` before `enabled = true`"), a ⚠️ "don't delete the key" box, and the demoted log. Flipping the default to **0** makes that safety *automatic* and lets all three compensating mechanisms be deleted.

The "backward compat" rationale protects a hypothetical: `fastly.toml` ships `edgezero_enabled = "false"`, EdgeZero has never been at 100% in production, and legacy removal is a separate PR (#495). No deployment relies on absent = 100, while the footgun is real (delete the key → instant full cutover).

**Fix:**
```rust
Ok(None) => {
    log::debug!("edgezero_rollout_pct key absent, defaulting to 0 (legacy path — fail safe)");
    0
}
```
and update the doc table above + the runbook failure-modes table accordingly. This also removes the need for the setup-ordering precondition.

docs/internal/EDGEZERO_MIGRATION.md
```diff
@@ -0,0 +1,170 @@
...
+
+Config store name: **`trusted_server_config`** (Fastly service config store)
+
+| Key                    | Type                 | Effect                                                                                                                            |
+| ---------------------- | -------------------- | --------------------------------------------------------------------------------------------------------------------------------- |
+| `edgezero_enabled`     | `"true"` / `"false"` | Master on/off switch. Set `"false"` to disable EdgeZero entirely, regardless of rollout_pct.                                      |
+| `edgezero_rollout_pct` | `"0"` – `"100"`      | Percentage of traffic (by client IP bucket) routed to EdgeZero. Only read when `edgezero_enabled = "true"`. Key absent = `"100"`. |
+
+**Routing logic:** `fnv1a_bucket(client_ip) < edgezero_rollout_pct` → EdgeZero, else legacy.
+Same client IP always gets the same bucket — routing is sticky per user.
```
aram356: 🔧 **wrench** — "sticky per user" overstates the guarantee; bucketing is sticky per **client IP**, not per user.

The routing key is `client_ip`, so:
- Multiple users behind one NAT/CGNAT egress share a bucket.
- A single user whose IP changes (mobile roaming, ISP re-assignment) is **re-bucketed** and can flip legacy↔EdgeZero mid-session. If the two paths differ in EC cookie / identity handling, an oscillating user could observe inconsistent identity during the rollout.

Suggested wording:
> Same client IP always gets the same bucket — routing is sticky **per client IP** (not per user; a user whose IP changes may re-bucket and switch paths).

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -179,14 +303,21 @@ fn main() {
...
         false
     }) {
-        log::debug!("routing request through EdgeZero path");
+        log::debug!("routing request through legacy path (edgezero_enabled=false)");
+        legacy_main(req);
+        return;
+    }
+
+    let rollout_pct = read_rollout_pct(&edgezero_config_store);
+    let route_to_edgezero = should_route_to_edgezero(rollout_pct, req.get_client_ip_addr());
```
aram356: 🤔 **thinking** — Two config-store reads per request.

Each request hits `config_store.get(...)` twice: once in `is_edgezero_enabled` (line 313) and once in `read_rollout_pct` (line 315). Fastly Config Store reads are local-edge and cheap, but at full QPS that's a measurable doubling. Consider consolidating into a single helper that returns `(enabled, rollout_pct)` from one read sequence. Not urgent — flag only.

crates/trusted-server-adapter-fastly/src/logging.rs
```diff
@@ -16,21 +16,46 @@ fn target_label(target: &str) -> &str {
...
+}
+
 /// Initialises the Fastly-backed `fern` logger and installs it as the global logger.
 ///
 /// Log records are forwarded to the `tslog` Fastly endpoint and echoed to stdout.
 /// Each line is prefixed with an RFC 3339 timestamp, level, and the final segment
 /// of the record's target module path.
 ///
+/// The maximum level defaults to `Info`. Setting the [`LOG_LEVEL_ENV`]
+/// environment variable (e.g. `EDGEZERO_LOG_LEVEL=debug`) raises it for local
```
aram356: ⛏ **nitpick** — Doc says the override "raises" the level, but it can also *lower* it.

`resolve_max_level` sets the max level to whatever `EDGEZERO_LOG_LEVEL` parses to, so `EDGEZERO_LOG_LEVEL=error` (or `off`) would *suppress* `Info` logs, not raise anything. Harmless (local-only), but the wording is imprecise — say "overrides" rather than "raises it."

crates/trusted-server-adapter-fastly/src/logging.rs
```diff
@@ -16,21 +16,46 @@ fn target_label(target: &str) -> &str {
     }
 }
 
+/// Environment variable that overrides the Fastly logger's maximum level.
+///
+/// Production ships at `Info`; this override exists for local pre-production
+/// validation under Viceroy, where raising the level to `debug` makes the
+/// `should_route_to_edgezero` route-decision lines observable. Production Fastly
+/// Compute does not surface arbitrary process environment variables, so the
+/// override is effectively local-only and the level stays at the safe default
```
aram356: 📝 **note** — The local-only guarantee rests on a Fastly-platform property.

This is correct for the Fastly adapter (Compute does not surface arbitrary process env vars at runtime). Flagging for awareness only: the repo now has axum/spin/cloudflare adapters that *do* run where env vars exist, so this `EDGEZERO_LOG_LEVEL` knob must not be copied verbatim into those adapters — there it would reintroduce exactly the per-request prod debug-flood this PR just removed. No change needed here.

aram356 (changes requested):
## Summary

Re-review of the canary routing changes against the PR base (`feature/edgezero-pr19-spin-adapter`). The prior review's findings on the per-request warn flood, degenerate-rollout short-circuit, predicate rename, `read_rollout_pct` test coverage, and debug-log observability are all resolved. Two findings remain open (both carried over) and three new minor items are noted inline. CI is green (12/12).

## Blocking

### 🔧 wrench

- **Fail-open default for absent `edgezero_rollout_pct`**: absent key → `100` (full cutover) is the only failure branch that does not fail to legacy. Recommend defaulting to `0`. (`main.rs` absent branch + doc table; runbook failure-modes table)
- **Runbook overstates stickiness**: "sticky per user" should read "sticky per client IP" — NAT sharing and IP changes break the per-user assumption. (`docs/internal/EDGEZERO_MIGRATION.md:19`)

## Non-blocking

### 🤔 thinking

- **Two config-store reads per request** — `is_edgezero_enabled` then `read_rollout_pct`; consolidatable into one read. (`main.rs:316`)

### ⛏ nitpick

- **`resolve_max_level` doc says "raises"** but the override can also lower the level. (`logging.rs:44`)

### 📝 note

- **Env-var override safety rests on a Fastly-platform property** — must not be copied verbatim into the axum/spin/cloudflare adapters. (`logging.rs:25`)

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS
- js tests: PASS

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Fixed in 17925a7c. Absent `edgezero_rollout_pct` now resolves to 0 (legacy), matching every other failure branch — deleting the key can never trigger a full cutover. Updated the `read_rollout_pct` doc table and the `read_rollout_pct_absent_defaults_to_legacy` test, plus the runbook failure-modes table. Removed the now-unnecessary setup-ordering precondition and the do-not-delete warning, as you noted.

docs/internal/EDGEZERO_MIGRATION.md
prk-Jr: Fixed in 17925a7c. Reworded to "sticky per client IP (not per user; a user whose IP changes, e.g. mobile roaming or ISP reassignment, may re-bucket and switch paths)", and noted the possible inconsistent-identity observation if the two paths differ in EC handling.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Leaving as-is. The two reads are distinct keys — `edgezero_enabled` then `edgezero_rollout_pct` — and the disabled path already early-returns after a single read, so the second read only happens once the flag is on. Both keys are genuinely required on the enabled path, so a combined helper would still issue two `get()` calls; it would not reduce the count, only couple two independent concerns. Open to revisiting if profiling shows the second local-edge read is material.

crates/trusted-server-adapter-fastly/src/logging.rs
prk-Jr: Fixed in 17925a7c. The doc now says the override "overrides" the level — used as-is, so `error`/`off` lowers it just as `debug` raises it.

crates/trusted-server-adapter-fastly/src/logging.rs
prk-Jr: Captured the awareness in 17925a7c: added a note on the `LOG_LEVEL_ENV` doc that this knob is Fastly-only (safe here because Compute hides runtime env vars) and must not be copied verbatim into the axum/spin/cloudflare adapters, where it would reintroduce the per-request prod debug flood this path avoids.

docs/superpowers/plans/2026-05-21-pr19-cutover-canary-rollout.md
prk-Jr: Fixed in 17925a7c. Replaced the stale Monitoring section in the plan with aggregate-metrics-only verification, mirroring `docs/internal/EDGEZERO_MIGRATION.md`: route-decision logs are local Viceroy only (`EDGEZERO_LOG_LEVEL=debug`), production surfaces no per-branch attribution, and no `x-edgezero-path`/real-time-stats split exists yet.

fastly.toml
```diff
@@ -64,6 +64,13 @@ build = """
             # EC API routes and the EC identity lifecycle are ported; the remaining
             # intentional deviations are documented in adapter-fastly/src/app.rs.
             edgezero_enabled = "false"
+            # Integer 0-100. Effective only when edgezero_enabled = "true".
+            #   0    -> all traffic to legacy (instant rollback — no deploy needed)
+            #   1-99 -> canary: clients whose fnv1a_bucket(client_ip) < this value go EdgeZero
+            #   100  -> all traffic to EdgeZero (full cutover)
+            # Key absent when edgezero_enabled = "true" is treated as 100 (full rollout).
```
ChristianPavilonis: Automated review: **P2: Local config comments still document the old fail-open absent-key behavior.**

The Fastly entry point now treats an absent `edgezero_rollout_pct` as `0`/all legacy, but this shipped Viceroy config still says an absent key is treated as `100` and that production must set `0` before enabling. That contradicts `read_rollout_pct()` and the new runbook. Because this block is the closest example operators/developers will copy when preparing the production config store, stale comments can cause confusion during rollback/cutover drills.

Suggested fix: change these comments to say that an absent key fails safe to legacy (`0`), and remove the now-obsolete setup-ordering warning.

docs/superpowers/plans/2026-05-21-pr19-cutover-canary-rollout.md
```diff
@@ -0,0 +1,721 @@
...
+---
+
+## Config store keys
+
+Config store name: **`trusted_server_config`** (Fastly service config store)
+
+| Key                    | Type                 | Effect                                                                                                                            |
+| ---------------------- | -------------------- | --------------------------------------------------------------------------------------------------------------------------------- |
+| `edgezero_enabled`     | `"true"` / `"false"` | Master on/off switch. Set `"false"` to disable EdgeZero entirely, regardless of rollout_pct.                                      |
+| `edgezero_rollout_pct` | `"0"` – `"100"`      | Percentage of traffic (by client IP bucket) routed to EdgeZero. Only read when `edgezero_enabled = "true"`. Key absent = `"100"`. |
```
ChristianPavilonis: Automated review: **P2: The embedded runbook/plan still contradicts the implemented rollout semantics.**

This implementation plan still says absent `edgezero_rollout_pct` means `100`, repeats the `Do NOT delete` warning/setup-ordering precondition below, and still describes routing as sticky per user. The actual code and `docs/internal/EDGEZERO_MIGRATION.md` now say absent/invalid/unreadable all fail safe to legacy and that stickiness is per client IP.

Why it matters: this file is explicitly written as an agentic-worker handoff and includes an embedded ops runbook; if it is used later for follow-up work or cutover prep, it can reintroduce the exact fail-open and stickiness assumptions this PR just fixed.

Suggested fix: update the duplicated sections (background flow, embedded runbook/failure-mode table, production setup precondition, invariants, and sticky-routing wording) to match the final implementation, or replace them with a short pointer to `docs/internal/EDGEZERO_MIGRATION.md` so there is only one source of operational truth.

ChristianPavilonis (commented):
Automated review:

## Review Summary

Reviewed PR #727 against `feature/edgezero-pr19-spin-adapter`, focusing on the Fastly canary routing implementation, rollout defaults, config-store failure behavior, local Fastly config, migration runbook, implementation plan docs, and CI/review context. I did not find any blocking correctness, security, data-loss, authorization, or severe compatibility issues. I left two P2 inline comments about stale operational guidance that still contradicts the current fail-safe absent-key behavior.

## Findings

### P0 / Blockers

None.

### P1 / High

None.

### P2 / Medium

2 inline comments submitted.

### P3 / Low

None.

## CI / Existing Reviews

All reported PR checks are passing. Existing review feedback about fail-open absent-key behavior, sticky-per-user wording, route-decision observability, logging override wording/scope, and stale monitoring guidance appears addressed in code/runbook and replies; however, some duplicated guidance in `fastly.toml` and the implementation plan remains stale. Locally, `cargo test -p trusted-server-adapter-fastly --target wasm32-wasip1 read_rollout_pct --no-run` completed successfully.

fastly.toml
prk-Jr: Fixed in 403de35f. The fastly.toml comments now say an absent key fails safe to legacy (0), and the obsolete "set 0 before enabling" precondition is removed.

docs/superpowers/plans/2026-05-21-pr19-cutover-canary-rollout.md
prk-Jr: Fixed in 403de35f. Updated every stale spot in the plan: background flow, the read_rollout_pct doc-table snippet, the rustdoc summary, the embedded fastly.toml snippet, the embedded runbook (config table, failure-modes table, do-not-delete warning, setup-order precondition), the sticky-routing note, and the invariants — all now match the implementation (absent/invalid/unreadable fail safe to 0; sticky per client IP). Also added a pointer marking docs/internal/EDGEZERO_MIGRATION.md as the canonical runbook so the embedded copy cannot silently drift again.

docs/superpowers/plans/2026-05-21-pr19-cutover-canary-rollout.md
```diff
@@ -0,0 +1,720 @@
...
+            Some(pct) => pct,
+            None => {
+                log::warn!(
+                    "invalid edgezero_rollout_pct value {:?}, defaulting to 0 (legacy path)",
+                    value
+                );
+                0
+            }
+        },
+        Ok(None) => 100,
```
ChristianPavilonis: Automated review: **P2: This snippet still documents the old fail-open absent-key behavior.**

The table immediately above says an absent `edgezero_rollout_pct` returns `0`, the implementation now returns `0`, and the final invariant says absent/invalid/unreadable must never route traffic to EdgeZero. This embedded code block still says `Ok(None) => 100`, so anyone reusing the implementation plan would reintroduce the exact fail-open behavior this PR just fixed: deleting or omitting the key would trigger full EdgeZero cutover.

Suggested fix: update this arm to return `0` (and preferably include the same debug log/comment as the real implementation), or mark the whole implementation snippet as historical/superseded so it cannot be copied as current guidance.

docs/superpowers/plans/2026-05-21-pr19-cutover-canary-rollout.md
```diff
@@ -0,0 +1,720 @@
...
+- **`docs/README.md` runbook pointer**: Add a line referencing `docs/internal/EDGEZERO_MIGRATION.md` so ops can discover it without grepping.
+
+---
+
+## Invariants to preserve
+
+- **No new dependencies** — FNV-1a is inlined, not a crate import.
+- **Fail-safe** — absent, invalid, or unreadable `edgezero_rollout_pct` all produce 0 (all legacy), never 100. A missing or misconfigured key can never accidentally route traffic to EdgeZero; rollout requires an explicit percentage.
+- **No handler or core changes** — all changes are confined to `main.rs` entry-point logic and `fastly.toml`.
+- **Sticky routing** — client IP hash is deterministic; the same user always gets the same path at a given `rollout_pct`, preventing split-session bugs.
```
ChristianPavilonis: Automated review: **P2: This invariant still overstates stickiness as per-user.**

The runtime hashes `client_ip`, and the canonical runbook now correctly says routing is sticky per client IP, not per user: NAT/CGNAT users can share a bucket, and a user whose IP changes can be re-bucketed and switch paths. This final invariant still promises that the same user always gets the same path and that split-session bugs are prevented, which contradicts the implemented routing key and the fixed runbook wording.

Suggested fix: align this invariant with the runbook, e.g. `client IP hash is deterministic; the same client IP gets the same path at a given rollout_pct, but NAT sharing and IP changes can re-bucket users.`

ChristianPavilonis (commented):
Automated review:

## Review Summary

Reviewed PR #727 against `feature/edgezero-pr19-spin-adapter`, focusing on the Fastly canary-routing code, fail-safe rollout defaults, logging changes, local config, migration runbook, implementation-plan documentation, CI, and existing review feedback. I did not find any blocking correctness, security, data-loss, authorization, or severe compatibility issues. I left two P2 inline comments on stale plan text that still contradicts the implemented fail-safe and per-client-IP semantics.

## Findings

### P0 / Blockers

None.

### P1 / High

None.

### P2 / Medium

2 inline comments submitted.

### P3 / Low

None.

## CI / Existing Reviews

All reported PR checks are passing. Existing feedback about absent-key fail-open behavior, route-decision observability, logging override wording/scope, and stale monitoring guidance appears addressed in the runtime code, `fastly.toml`, and the canonical migration runbook; the remaining issues are duplicated/stale text in the implementation-plan document. Locally, `cargo test -p trusted-server-adapter-fastly --target wasm32-wasip1 read_rollout_pct --no-run` and `cargo fmt --all -- --check` passed; `cd docs && npm run format -- --check` could not run because `prettier` is not installed in this environment, while CI's docs format check is green.

crates/trusted-server-adapter-fastly/src/logging.rs
```diff
@@ -16,21 +16,52 @@ fn target_label(target: &str) -> &str {
...
+/// environment variable (e.g. `EDGEZERO_LOG_LEVEL=debug`) overrides it for local
+/// Viceroy validation — the value is used as-is, so `error`/`off` lowers it just
+/// as `debug` raises it; see [`resolve_max_level`].
+///
 /// # Panics
 ///
 /// Panics if the Fastly logger cannot be built or if the global logger has already
 /// been set.
 pub(crate) fn init_logger() {
+    let configured = std::env::var(LOG_LEVEL_ENV).ok();
```
ChristianPavilonis: 🔧 **wrench — `EDGEZERO_LOG_LEVEL` is not visible to Viceroy guest code.**

`std::env::var(LOG_LEVEL_ENV)` reads the guest WASI environment, but `fastly compute serve`/Viceroy does not pass arbitrary shell environment variables like `EDGEZERO_LOG_LEVEL` into the Compute guest. That means the runbook command `EDGEZERO_LOG_LEVEL=debug fastly compute serve` still initializes at `Info`, so the route-decision `debug!` lines remain unavailable in the only documented branch-level validation path.

One minimal way to make the local validation path real is to key off the Fastly-provided local hostname and keep production at the safe default:

```suggestion
    let configured = std::env::var(LOG_LEVEL_ENV).ok().or_else(|| {
        (std::env::var("FASTLY_HOSTNAME").as_deref() == Ok("localhost"))
            .then_some("debug".to_owned())
    });
```

docs/internal/EDGEZERO_MIGRATION.md
```diff
@@ -0,0 +1,167 @@
...
+that appears and tracks the rollout steps implicates the EdgeZero branch:
+
+- **Error rate:** `5xx / total_requests` by edge PoP
+- **Latency p95:** service-wide
+- **Auction win-rate:** downstream SSP reporting, compare same-day prior week
+- **Timeout rate:** `504 / total_requests`
+
+> For local pre-production validation under Viceroy, start the simulator with
+> `EDGEZERO_LOG_LEVEL=debug` (the logger defaults to `Info`, which suppresses
+> these lines) — for example `EDGEZERO_LOG_LEVEL=debug fastly compute serve`.
```
ChristianPavilonis: 🔧 **wrench — make the local log validation instructions match a guest-visible signal.**

After changing the logger to use a Viceroy/Compute-provided signal (for example `FASTLY_HOSTNAME=localhost`) instead of an arbitrary shell env var, this runbook should stop telling operators to set `EDGEZERO_LOG_LEVEL`, which the guest will not see.

```suggestion
> For local pre-production validation under Viceroy, start the simulator normally.
> Viceroy exposes `FASTLY_HOSTNAME=localhost` to guest code, and the Fastly
> logger raises the route-decision level in that local environment while
> production stays at `Info`.
```

docs/superpowers/plans/2026-05-21-pr19-cutover-canary-rollout.md
ChristianPavilonis: 🔧 **wrench — this embedded snippet still has the old fail-open absent-key behavior.**

The table above, the runtime code, and the runbook all say an absent `edgezero_rollout_pct` fails safe to `0`; this snippet still returns `100`. Because this file is an agent handoff, copying it later would reintroduce the exact full-cutover footgun the PR fixed.

```suggestion
        Ok(None) => {
            log::debug!(
                "edgezero_rollout_pct key absent, defaulting to 0 (legacy path — fail safe)"
            );
            0
        }
```

docs/superpowers/plans/2026-05-21-pr19-cutover-canary-rollout.md
ChristianPavilonis: 🔧 **wrench — keep the invariant to per-client-IP stickiness.**

The runtime hashes `client_ip`, not a durable user identifier. NAT/CGNAT users can share a bucket, and a user whose IP changes can re-bucket, so the plan should not promise that the same user always gets the same path.

```suggestion
- **Sticky routing** — client IP hash is deterministic; the same client IP gets the same path at a given `rollout_pct`, but NAT sharing and IP changes can re-bucket users.
```

docs/superpowers/plans/2026-05-21-pr19-cutover-canary-rollout.md
```diff
@@ -0,0 +1,720 @@
...
+Expected: all tests pass.
+
+```bash
+cargo fmt --all -- --check 2>&1 | tail -10
+```
+
+Expected: no diff. If there is output, run `cargo fmt --all` to fix, then re-check.
+
+```bash
+cargo clippy -p trusted-server-adapter-fastly --all-targets --all-features -- -D warnings 2>&1 | tail -20
```
ChristianPavilonis: 🔧 **wrench — use the target-matched Fastly clippy gate.**

This host-targeted clippy command conflicts with `CLAUDE.md`'s Fastly gate. Also, the surrounding `| tail`/`| head` validation commands should be removed or wrapped in `set -o pipefail`, otherwise a failed `cargo`/`npm` command can look green because the final pipeline command succeeded.

```suggestion
cargo clippy-fastly
```

docs/superpowers/plans/2026-05-21-pr19-cutover-canary-rollout.md
```diff
@@ -0,0 +1,720 @@
...
+```
+
+---
+
+## Final verification
+
+- [ ] **Run full test suite**
+
+```bash
+cargo test-fastly 2>&1 | tail -20
```
ChristianPavilonis: 🔧 **wrench — don't pipe validation through `tail` without `pipefail`.**

As written, a failing test command can still exit successfully if `tail` succeeds. Keep the command's real exit status visible.

```suggestion
cargo test-fastly
```

docs/superpowers/plans/2026-05-21-pr19-cutover-canary-rollout.md
```diff
@@ -0,0 +1,720 @@
...
+```
+
+Expected: all tests pass, including the 10 new canary routing tests.
+
+- [ ] **Run fmt and clippy**
+
+```bash
+cargo fmt --all -- --check 2>&1 | tail -10
+cargo clippy -p trusted-server-adapter-fastly --all-targets --all-features -- -D warnings 2>&1 | tail -10
```
ChristianPavilonis: 🔧 **wrench — final verification should use the Fastly clippy alias and preserve exit status.**

This is the PR handoff gate, so it should match the repo's target-specific CI command and not hide failures behind `tail`.

```suggestion
cargo clippy-fastly
```

ChristianPavilonis (changes requested):
## Review Summary

🔧 Requesting changes. The runtime fail-safe rollout behavior is much improved and CI is green, but local canary validation still relies on a log-level mechanism that Viceroy/Compute guest code will not actually see. I also left accept-ready suggestions for stale plan snippets and validation commands so the PR artifacts stay safe to reuse.

Additional note folded into the body because the exact insertion point is unchanged context in the diff: `max_level` is currently applied only to the inner `log_fastly::Logger`. Please also set the same level on the outer `fern::Dispatch` (e.g. `fern::Dispatch::new().level(max_level)`) so production `debug!` route decisions are filtered before timestamp/format construction, not just dropped by the child logger.

docs/superpowers/plans/2026-05-21-pr19-cutover-canary-rollout.md
```diff
@@ -0,0 +1,731 @@
...
+
+**No** `edgezero_rollout_pct` key exists anywhere yet. After this PR the flow becomes:
+
+1. Open config store — on error → `legacy_main`
+2. Check `edgezero_enabled` — if false/err → `legacy_main`
+3. Read `rollout_pct` from `read_rollout_pct()` (absent / invalid / unreadable all = 0, fail safe to legacy)
+4. Compute `bucket = fnv1a_bucket(client_ip_string)`
+5. If `bucket < rollout_pct` → `edgezero_main`, else → `legacy_main`
+
+`ConfigStoreHandle` comes from `edgezero_core::config_store` and cannot be constructed in unit tests (needs Fastly runtime). Test only pure functions.
```
ChristianPavilonis: Automated review: **P2: The implementation plan still says the config-store rollout helper cannot be unit-tested.**

This is now stale and contradicts the PR's final code: `read_rollout_pct` is unit-tested with an in-memory `ConfigStoreHandle` stub in `crates/trusted-server-adapter-fastly/src/main.rs`, and the helper was also renamed from `canary_routes_to_edgezero` to `routes_to_edgezero`. Because this file is explicitly framed as an agentic-worker handoff, leaving the old constraint in place can cause future follow-up work to skip the safety-critical config-store branch tests or copy the outdated helper names.

Suggested fix: either mark this plan as a historical snapshot that must not be reused for implementation, or sync the affected sections with the final PR: remove the claim that `ConfigStoreHandle` cannot be constructed in unit tests, mention the in-memory `ConfigStore` stub coverage for `read_rollout_pct`, and update the old `canary_routes_to_edgezero` references to `routes_to_edgezero`.

ChristianPavilonis (commented):
Automated review:

## Review Summary

Reviewed PR #727 against `feature/edgezero-pr19-spin-adapter`, focusing on the Fastly canary routing implementation, rollout defaults/failure behavior, logging-level changes, local config, migration runbook, duplicated implementation-plan guidance, CI, and existing review feedback. I did not find any blocking correctness, security, data-loss, authorization, or severe compatibility issues. I left one P2 inline comment on stale plan guidance that now contradicts the implemented tests/helper names and could mislead future agent handoffs.

## Findings

### P0 / Blockers

None.

### P1 / High

None.

### P2 / Medium

1 inline comment submitted.

### P3 / Low

None.

## CI / Existing Reviews

All reported PR checks are passing. Existing review feedback about absent-key fail-safe behavior, per-client-IP stickiness, route-decision observability, logging override scope, stale fail-open snippets, and validation command exit-status masking appears addressed in the runtime code, `fastly.toml`, the canonical migration runbook, and portions of the implementation-plan document. Locally, `cargo test -p trusted-server-adapter-fastly --target wasm32-wasip1 rollout_pct --no-run` completed successfully.

crates/trusted-server-adapter-fastly/src/logging.rs
```diff
@@ -16,25 +16,77 @@ fn target_label(target: &str) -> &str {
...
 ///
 /// Panics if the Fastly logger cannot be built or if the global logger has already
 /// been set.
 pub(crate) fn init_logger() {
+    let configured = std::env::var(LOG_LEVEL_ENV).ok().or_else(|| {
+        // Viceroy reports the guest host as `localhost`; production cache nodes
+        // report their real hostname. Raise to debug only in that local
+        // environment so route-decision lines surface under `fastly compute
+        // serve` while production stays at the safe `Info` default.
+        (std::env::var(LOCAL_HOSTNAME_ENV).as_deref() == Ok(LOCAL_HOSTNAME))
```
aram356: ♻️ **refactor** — The log-level *detection* logic is untested; the production-safety property isn't pinned.

`resolve_max_level` (the level parse) is covered, but the branch that actually governs production safety is this inline `or_else`: explicit override wins, `FASTLY_HOSTNAME == "localhost"` → debug, and crucially **a real production hostname must stay at `Info`**. That last property is exactly what a future refactor could silently break (debug-log flood in production), and nothing tests it.

Suggest extracting a pure helper and testing the precedence/detection matrix:
```rust
fn detect_max_level(explicit: Option<&str>, hostname: Option<&str>) -> log::LevelFilter {
    let configured = explicit
        .map(str::to_owned)
        .or_else(|| (hostname == Some(LOCAL_HOSTNAME)).then(|| "debug".to_owned()));
    resolve_max_level(configured.as_deref())
}
```
Tests: explicit wins over hostname; `Some("localhost")` → Debug; `Some("cache-xyz")` and `None` → Info.

crates/trusted-server-adapter-fastly/src/logging.rs
```diff
@@ -16,25 +16,77 @@ fn target_label(target: &str) -> &str {
...
+/// value takes precedence where the guest environment is populated — the value is
+/// used as-is, so `error`/`off` lowers it just as `debug` raises it; see
+/// [`resolve_max_level`].
+///
 /// # Panics
 ///
 /// Panics if the Fastly logger cannot be built or if the global logger has already
 /// been set.
 pub(crate) fn init_logger() {
+    let configured = std::env::var(LOG_LEVEL_ENV).ok().or_else(|| {
```
aram356: ⛏ **nitpick** — An unparseable explicit `EDGEZERO_LOG_LEVEL` silently suppresses the local auto-debug.

`.ok()` makes any set value (e.g. a typo like `debugg`) short-circuit the `or_else`, so under Viceroy you'd get `Info` instead of the auto-raised `debug`, with no signal as to why. Precedence (explicit wins) is defensible — flagging only as a minor dev-surprise.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -191,14 +316,21 @@ fn main() {
...
         log::warn!("failed to read edgezero_enabled flag, falling back to legacy path: {e}");
         false
     }) {
-        log::debug!("routing request through EdgeZero path");
+        log::debug!("routing request through legacy path (edgezero_enabled=false)");
+        legacy_main(req);
+        return;
+    }
+
+    let rollout_pct = read_rollout_pct(&edgezero_config_store);
```
aram356: 🤔 **thinking** — Two config-store reads per request (carried over).

`is_edgezero_enabled` (line 319) then `read_rollout_pct` (here) each call `config_store.get(...)`. Local-edge reads are cheap, so leaving this as-is is reasonable; a single helper returning `(enabled, rollout_pct)` would halve it if it ever matters. Flag only.

aram356 (approved):
## Summary

Re-review against the PR base (`feature/edgezero-pr19-spin-adapter`). All findings from the prior review rounds are resolved:

- Absent `edgezero_rollout_pct` now fails safe to `0` (legacy), with the doc table, runbook failure-modes table, and setup precondition updated to match.
- The runbook stickiness claim is corrected to "sticky per client IP (not per user)".
- The log-level override docs no longer imply it only raises (it can lower too), and the `FASTLY_HOSTNAME` signal is documented as Fastly-specific and not to be copied into the axum/spin/cloudflare adapters.
- `.level(max_level)` on the fern `Dispatch` now filters route-decision debug logs out of production (without it, fern's default would have overridden the Logger's `Info`).

Approving. Three non-blocking items remain inline: a test-coverage gap on the log-level detection logic (♻️), the two-per-request config-store reads (🤔, carried), and an `EDGEZERO_LOG_LEVEL` precedence nit (⛏).

## Non-blocking

### ♻️ refactor

- **Log-level detection logic is untested** — extract a pure helper and pin the production-safety property (real hostname must stay `Info`). (`logging.rs:76`)

### 🤔 thinking

- **Two config-store reads per request** — consolidatable into one read. (`main.rs:328`)

### ⛏ nitpick

- **Unparseable `EDGEZERO_LOG_LEVEL` suppresses local auto-debug** — minor dev-surprise. (`logging.rs:71`)

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS
- js tests: PASS

docs/internal/EDGEZERO_MIGRATION.md
```diff
@@ -0,0 +1,168 @@
...
+2. Optionally set `edgezero_enabled = "false"` as belt-and-suspenders.
+3. Investigate root cause before re-advancing the canary.
+4. Keep the legacy entry point (`legacy_main()`) available until at least one
+   full release cycle after reaching 100% with zero regressions.
+
+---
+
+## Monitoring
+
+> **There is no production signal that splits traffic by EdgeZero-vs-legacy
```
ChristianPavilonis, line 128: 📌 **pushpin — Production canary still needs branch-level attribution before real cutover.**

The runbook is now honest that no production EdgeZero-vs-legacy split exists yet. That is safe documentation, but operationally it means the staged rollout can only be inferred from aggregate metrics, not verified or attributed per branch.

Before running the production canary, please add a production-safe low-cardinality signal (sampled info log, counter/metric, or captured response/log field) and update this runbook to make that the verification source. Not blocking this code path approval, but it should be tracked before ops uses the runbook.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -144,6 +190,85 @@ fn is_edgezero_enabled(config_store: &ConfigStoreHandle) -> Result<bool, fastly:
...
+/// | Key read error                  | `0`          | All legacy (safe default)  |
+fn read_rollout_pct(config_store: &ConfigStoreHandle) -> u8 {
+    match config_store.get(EDGEZERO_ROLLOUT_PCT_KEY) {
+        Ok(Some(value)) => match parse_rollout_pct(&value) {
+            Some(pct) => pct,
+            None => {
+                log::warn!(
+                    "invalid edgezero_rollout_pct value {:?}, defaulting to 0 (legacy path)",
+                    value
+                );
```
ChristianPavilonis: 🌱 **seedling — Avoid logging the raw invalid rollout value.**

This value is operator-controlled config and this branch can run on every request until the config is fixed. Logging the raw value can flood logs and can persist an accidentally pasted secret. Consider redacting the value here; sampling/rate-limiting repeated warnings can be a follow-up if this remains noisy.

```suggestion
                log::warn!(
                    "invalid edgezero_rollout_pct value, defaulting to 0 (legacy path)"
                );
```

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -191,14 +316,21 @@ fn main() {
...
         false
     }) {
-        log::debug!("routing request through EdgeZero path");
+        log::debug!("routing request through legacy path (edgezero_enabled=false)");
+        legacy_main(req);
+        return;
+    }
+
+    let rollout_pct = read_rollout_pct(&edgezero_config_store);
+    let route_to_edgezero = should_route_to_edgezero(rollout_pct, req.get_client_ip_addr());
```
ChristianPavilonis: ♻️ **recycle — Cover the combined entry-point gate, not only its pieces.**

The helper tests are good, but the public behavior depends on this composition: `edgezero_enabled` must be true, `edgezero_rollout_pct` must be read from the real key, and the final branch must dispatch to the intended entry point. A small pure selector test (or Viceroy smoke) covering disabled, absent/invalid, `0`, partial hit/miss, and `100` would catch a future wiring typo/inversion that helper-only tests might miss.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -1240,6 +1372,284 @@ mod tests {
...
+
+    struct TestConfigStore {
+        response: StubResponse,
+    }
+
+    impl edgezero_core::config_store::ConfigStore for TestConfigStore {
+        fn get(
+            &self,
+            _key: &str,
+        ) -> Result<Option<String>, edgezero_core::config_store::ConfigStoreError> {
```
ChristianPavilonis, line 1668: ♻️ **recycle — Pin the config-store key in the stub.**

This in-memory store is exactly where a future typo in `EDGEZERO_ROLLOUT_PCT_KEY` can be caught cheaply.

```suggestion
        fn get(
            &self,
            key: &str,
        ) -> Result<Option<String>, edgezero_core::config_store::ConfigStoreError> {
            assert_eq!(
                key, EDGEZERO_ROLLOUT_PCT_KEY,
                "stub should pin the rollout config key"
            );
```

ChristianPavilonis (approved):
## Summary

👍 Approving. The Fastly canary runtime path now fails safe to legacy, the critical rollout helper branches are covered, and reported CI is green. I left non-blocking inline comments using the CREG style from `CONTRIBUTING.md`.

## Folded follow-ups

🔧 **wrench — Configuration guide still describes the old one-key rollout flow** — `docs/guide/configuration.md:1163`

This file is not in the PR diff, so I could not leave an inline suggestion. The guide says `edgezero_enabled=true` routes traffic through EdgeZero and the examples only set that key. With this PR, absent `edgezero_rollout_pct` defaults to `0`, so following the guide produces 0% EdgeZero. Please update the guide with both config keys, staged values, fallback behavior, and rollback via `edgezero_rollout_pct = "0"`.

📝 **memo — Spec drift to reconcile later** — `docs/superpowers/specs/2026-03-19-edgezero-migration-design.md:1512` still says rollout hashes the request ID, while the implementation and runbook hash client IP. Please align that source doc when updating rollout documentation.

PR #743, opened 2026-05-28
Prebid external first-party bundle loading

### LINKED ISSUE #692: Better TSJS prebid bundle onboarding
### Description

Need to think of a better solution to build and bundle custom prebid js

ChristianPavilonis:
## Summary
- Hard cut Prebid to the external first-party bundle flow: no `bundle_mode`, no embedded/deferred `tsjs-prebid` loading, and `external_bundle_url` is required whenever Prebid is enabled.
- Always inject the same-origin `/integrations/prebid/bundle.js` script and keep publisher Prebid script interception/removal to prevent duplicate instances.
- Proxy the generated bundle with static-asset behavior: no EC forwarding, no copied request headers, streaming passthrough, HTTPS-only targets/redirects, and `proxy.allowed_domains` enforcement.
- Keep `external_bundle_sha256` optional; when present it drives versioned first-party URLs, immutable cache headers, and `sha256:` ETags. SRI is validated if configured but is not required.
- Move Prebid out of the embedded Cargo TSJS build and document/generate publisher-specific external bundles with `build-prebid-external.mjs`.

## Quick how-to: generate and configure a Prebid bundle

1. Generate a publisher-specific external Prebid bundle:

```bash
cd crates/js/lib
npm run build:prebid-external -- \
  --adapters=rubicon,appnexus,openx \
  --user-id-modules=sharedIdSystem,uid2IdSystem \
  --out=dist/prebid
```

2. Upload the generated `trusted-prebid-<sha256>.js` from `dist/prebid/` to an HTTPS asset host that is allowed by `proxy.allowed_domains`.

3. Copy values from `dist/prebid/manifest.json` into config:

```toml
[integrations.prebid]
enabled = true
server_url = "https://prebid-server.example.com/openrtb2/auction"
external_bundle_url = "https://assets.example.com/prebid/trusted-prebid-<sha256>.js"

# Optional but recommended for versioned first-party URL + immutable cache headers.
external_bundle_sha256 = "<manifest sha256>"

# Optional browser SRI. Validated if present, but not required.
external_bundle_sri = "<manifest sri>"

# Include any client-side bidders whose adapters were included above.
client_side_bidders = ["rubicon"]
```

Trusted Server injects the same-origin `/integrations/prebid/bundle.js` route for browsers; publishers should not reference the external asset URL directly in page markup.

## Testing
- `TSJS_SKIP_BUILD=1 cargo check -p trusted-server-core`
- `TSJS_SKIP_BUILD=1 cargo test -p trusted-server-core prebid --lib`
- `TSJS_SKIP_BUILD=1 cargo test -p trusted-server-core publisher::tests --lib`
- `TSJS_SKIP_BUILD=1 cargo test -p trusted-server-core registry::tests --lib`
- `TSJS_SKIP_BUILD=1 cargo test --workspace`
- `TSJS_SKIP_BUILD=1 cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo check -p trusted-server-js`
- `cd crates/js/lib && npm run build`
- `cd crates/js/lib && npm run format`
- `cd crates/js/lib && npx vitest run`
- `cd crates/js/lib && npm run build:prebid-external -- --adapters=rubicon --user-id-modules=sharedIdSystem --out=/tmp/trusted-server-prebid-external-hard-cutover`
- `cd docs && npm ci && npm run format`


docs/guide/integrations/prebid.md
```diff
@@ -280,22 +288,22 @@ client_side_bidders = ["rubicon"]             # native browser adapters
 
 The two lists are independent — the operator manages both explicitly. If a bidder appears in both lists, a warning is logged at startup (the bidder will run in both paths, which is likely unintended).
 
-### Build-time adapter selection
+### External bundle adapter selection
 
-Client-side bidders need their Prebid.js adapter modules bundled in the JS output. This is controlled by the `TSJS_PREBID_ADAPTERS` environment variable at build time:
+Client-side bidders need their Prebid.js adapter modules included in the generated external bundle:
```
ChristianPavilonis, line 324: The bundling should be moved to CLI when ready.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -192,6 +216,24 @@ impl IntegrationConfig for PrebidIntegrationConfig {
...
+/// schema validation, or bidder-param override compilation.
+pub fn validate_config_for_startup(
+    settings: &Settings,
+) -> Result<Option<PrebidIntegrationConfig>, Report<TrustedServerError>> {
+    let Some(config) =
+        settings.integration_config::<PrebidIntegrationConfig>(PREBID_INTEGRATION_ID)?
+    else {
+        return Ok(None);
+    };
+    BidParamOverrideEngine::try_from_config(&config)?;
```
aram356, line 252: 🔧 **wrench** — `ts config validate` skips external-bundle validation, so the CLI green-lights a config the running server then rejects at boot.

`validate_config_for_startup` (called by the CLI at `config_command.rs:215`) only runs `BidParamOverrideEngine::try_from_config`; it never calls `validate_external_bundle_config`, which `build()`/`register()` enforces at line 736. A Prebid config that is `enabled=true` but missing `external_bundle_url`, using `http://`, or pointing outside `proxy.allowed_domains` passes `ts config validate` yet hard-fails server startup — defeating the pre-deploy gate. Scratch-verified: adding the call keeps all 130 prebid/registry/settings tests green (including the `disabled_invalid_default_enabled_prebid_skips_validation` path).

```suggestion
    BidParamOverrideEngine::try_from_config(&config)?;
    validate_external_bundle_config(&config, &settings.proxy.allowed_domains)?;
```

crates/trusted-server-js/lib/build-prebid-external.mjs
```diff
@@ -0,0 +1,266 @@
...
+  fs.writeFileSync(
+    path.join(args.outDir, 'manifest.json'),
+    `${JSON.stringify(manifest, null, 2)}\n`
+  );
+
+  console.log('[build-prebid-external] Built external Prebid bundle:', bundle.filename);
+  console.log('[build-prebid-external] SHA-256:', bundle.sha256);
+  console.log('[build-prebid-external] SRI:', bundle.sri);
+  console.log('[build-prebid-external] Manifest:', path.join(args.outDir, 'manifest.json'));
+} finally {
```
aram356: ❓ **question** — The builder overwrites the git-tracked `_adapters.generated.ts` / `_user_ids.generated.ts` (lines 150, 178) and restores them only in this `finally` block. `finally` covers thrown exceptions, but a `SIGINT`/`SIGKILL`/crash bypasses it, leaving a **publisher-specific adapter set committed as the repo default** — which a later unrelated `cargo build` would embed. Is the in-tree mutation intentional?

Safer options: generate to a temp path and alias via Vite `resolve.alias`/`input`, or trap signals to run the restore:

```js
for (const sig of ['SIGINT', 'SIGTERM']) process.on(sig, () => { restoreFile(ADAPTERS_FILE, originalAdapters); restoreFile(USER_IDS_FILE, originalUserIds); process.exit(1); });
```

Apply manually — spans setup outside this hunk.

crates/trusted-server-cli/src/edgezero_delegate.rs
```diff
@@ -0,0 +1,436 @@
...
+        report_error("manifest has no `[stores.config]` section; declare it before pushing config")
+    })?;
+    if !declaration.ids.iter().any(|id| id == &request.store) {
+        return cli_error(format!(
+            "--store={:?} is not in [stores.config].ids ({:?})",
+            request.store, declaration.ids
+        ));
+    }
+
+    let env_config = EnvConfig::from_env();
```
aram356: ❓ **question** — `push_config_entries` resolves `EnvConfig::from_env()` and calls the real `push_config_entries`/`push_config_entries_local` for both dry-run and real pushes, delegating dry-run honoring entirely to the pinned `edgezero-adapter` rev with no local guard or test. If that adapter's dry-run contract ever regresses, `ts config push --dry-run` would mutate remote state with no backstop here. Can you add a local test that pins the dry-run semantics, or short-circuit the obviously-mutating setup before the adapter call when `request.dry_run`?

crates/trusted-server-core/src/request_signing/endpoints.rs
```diff
@@ -24,6 +24,13 @@ fn json_response(status: StatusCode, body: String) -> Response<EdgeBody> {
         .expect("should build json response")
 }
 
+fn request_body_bytes(
```
aram356, line 27: ❓ **question** — `request_body_bytes` returns `Result` but never errs, ignores its `_endpoint` arg, and — unlike the auction handler's guard at `auction/endpoints.rs:81` — has no `is_stream()` check. In the pinned edgezero rev `into_bytes()` collapses a stream body to empty, so a streamed `rotate-key` would parse as `RotateKeyRequest { kid: None }` and silently rotate the **default** key. It's unreachable today (the Fastly adapter always buffers), hence a question not a blocker — but the helper is asymmetric with the auction fix and earns its `Result` only if it guards:

```rust
if body.is_stream() {
    return Err(Report::new(TrustedServerError::BadRequest {
        message: "Streaming request bodies are not supported".to_string(),
    }));
}
Ok(body.into_bytes())
```

Apply manually — the `_endpoint` param and call sites need updating together. (Or just delete the helper and call `into_bytes()` directly.)

crates/trusted-server-core/src/publisher.rs
```diff
@@ -32,8 +32,10 @@ use crate::streaming_replacer::create_url_replacer;
 const SUPPORTED_ENCODING_VALUES: [&str; 3] = ["gzip", "deflate", "br"];
 const DEFAULT_PUBLISHER_FIRST_BYTE_TIMEOUT: Duration = Duration::from_secs(15);
 
-fn body_as_reader(body: EdgeBody) -> std::io::Cursor<bytes::Bytes> {
-    std::io::Cursor::new(body.into_bytes())
+fn body_as_reader(
```
aram356: 🤔 **thinking** — `body_as_reader` (and the sibling `request_body_bytes` in `request_signing/endpoints.rs`) wrap an infallible `into_bytes()` in `Result<_, Report<...>>`, adding `?` noise at every call site for no behavioral gain — it reads like an abandoned intent to enforce a max-body bound. Either wire a real `into_bytes_bounded(MAX)` cap so the `Result` is earned, or revert to the plain infallible signature.

crates/trusted-server-cli/src/edgezero_delegate.rs
```diff
@@ -0,0 +1,436 @@
...
+        if let Some(command_text) = manifest_command(loader.manifest(), adapter_name, command) {
+            let manifest = loader.manifest();
+            let root = manifest.root().unwrap_or_else(|| Path::new("."));
+            let environment = manifest.environment_for(adapter_name);
+            let adapter_bind = adapter_bind_from_manifest(manifest, adapter_name);
+            return run_shell(command_text, root, &environment, adapter_bind, passthrough);
+        }
+    }
+
+    let adapter = adapter_registry::get_adapter(adapter_name).ok_or_else(|| {
```
aram356: 🤔 **thinking** — Env/secret handling diverges by path: the manifest-`command_text` branch runs `apply_environment` (which fails on unset declared secrets and binds `EDGEZERO__ADAPTER__HOST/PORT`), but this in-process `adapter.execute(...)` branch never does. It's **not** a bypass — the in-process adapter resolves its own credentials — but the asymmetry is non-obvious. A one-line comment noting that manifest-declared secrets/env bindings apply only to the shell-command path would keep a future reader from 'hardening' this incorrectly.

crates/trusted-server-core/src/proxy.rs
```diff
@@ -1372,6 +1412,12 @@ async fn proxy_with_redirects(
 
         let next_scheme = next_url.scheme().to_ascii_lowercase();
```
aram356, line 1406: 🤔 **thinking** — The redirect HTTPS enforcement is split across two overlapping blocks (the non-http/https `require_https` Forbidden at 1415-1420, then the `require_https && next_scheme != "https"` block at 1428-1433). Functionally correct but hard to reason about. Consider one early check right after `next_scheme` is computed:

```rust
if redirect_policy.require_https && next_scheme != "https" {
    log::warn!("redirect to `{}` blocked: HTTPS is required", next_url);
    return Err(Report::new(TrustedServerError::Forbidden {
        message: "non-HTTPS redirect blocked".to_string(),
    }));
}
```

then keep the non-http/https passthrough for the non-`require_https` path. Apply manually — restructures two hunks.

crates/trusted-server-core/src/tester_cookie.rs
```diff
@@ -57,6 +65,44 @@ pub fn handle_set_tester(
...
+///
+/// Returns `404 Not Found` while `[tester_cookie].enabled` is false. When the
+/// feature is enabled, returns `204 No Content` with `Set-Cookie` expiring the
+/// `ts-tester` cookie scoped to `publisher.cookie_domain`.
+///
+/// # Errors
+///
+/// Returns [`TrustedServerError::InvalidHeaderValue`] if the configured cookie
+/// domain cannot be rendered as an HTTP header value.
+pub fn handle_clear_tester(
```
aram356, line 80: 🤔 **thinking** — `/_ts/clear-tester` (like the pre-existing set endpoint) is gated only by `tester_cookie.enabled`, not by `enforce_basic_auth` (the auth regex matches `^/(_ts/)?admin`, which this path doesn't). The cookie value is a fixed literal, so there's no injection/response-splitting risk, and it only affects tester routing — but anyone can flip the tester cookie while the feature is enabled. Confirm this open posture is intended; if the tester cookie ever gates sensitive behavior, gate these routes behind auth or a signed value.

crates/trusted-server-core/src/integrations/osano.rs
```diff
@@ -0,0 +1,96 @@
...
+use crate::error::TrustedServerError;
+use crate::settings::{IntegrationConfig, Settings};
+
+use super::IntegrationRegistration;
+
+const OSANO_INTEGRATION_ID: &str = "osano";
+
+/// Configuration for the Osano consent mirror integration.
+#[derive(Debug, Clone, Deserialize, Validate)]
+pub struct OsanoConfig {
```
aram356, line 22: 🌱 **seedling** — Most config structs gained `#[serde(deny_unknown_fields)]` in this PR (Publisher, Ec, Proxy, consent types, `Settings`). `OsanoConfig` (`enabled` only) was left without it — adding it would catch typo'd keys consistently with the rest of the PR. (The top-level `PrebidIntegrationConfig` omission is correct — it must tolerate the CLI-only `[integrations.prebid.bundle]` sub-table.)

crates/trusted-server-cli/src/config_command.rs
```diff
@@ -0,0 +1,466 @@
...
+        .map(|_| ())
+        .map_err(|error| {
+            report_error(format!(
+                "invalid app config: EC partner registry startup failed: {error:?}"
+            ))
+        })?;
+    Ok(())
+}
+
+fn validate_enabled_integrations(
```
aram356: ⛏ **nitpick** — `validate_enabled_integrations` hardcodes the integration list (`aps`, `testlight`, …). When a new integration is added to core, `ts config validate` silently won't validate it until someone remembers to edit this list — a misconfigured new integration passes CLI validation but fails at runtime. Consider driving this from the integration registry, or add a test asserting this list matches the registry's known IDs.

crates/trusted-server-cli/Cargo.toml
```diff
@@ -0,0 +1,33 @@
+[package]
+name = "trusted-server-cli"
+version = "0.1.0"
+authors = []
+edition = "2021"
```
aram356: 📝 **note** — This crate declares `edition = "2021"` while CLAUDE.md and the core/adapter crates mandate the 2024 edition. Likely deliberate (host-toolchain differences), but it silently diverges from the documented standard — bump to 2024 if the pinned toolchain supports it, or add a comment explaining why 2021 is required.

aram356 (changes requested):
## Summary

Reworks Prebid to load an external first-party bundle proxied through the server (hard cut: `external_bundle_url` required when enabled, no embedded `tsjs-prebid`), adds the `trusted-server-cli` crate, the Osano consent mirror, and a hash-verified config-store payload. Overall this is high-quality, security-conscious work — the proxy path in particular is well-built. Verdict is **REQUEST_CHANGES** for one validation-parity bug plus a few open questions; the blocking fixes are small.

> Reviewed against base `feature/ts-cli-next` (this is a **stacked PR** — the range also carries Osano #773 and tester-cookie-clear #797). 1 of the inline comments carries a one-click, scratch-verified GitHub `suggestion`; the rest describe fixes in prose because they span multiple files, touch new files, or are design questions.

## Blocking

### 🔧 wrench
- **`ts config validate` skips external-bundle validation (CLI passes, server boot-fails)** — see inline at `crates/trusted-server-core/src/integrations/prebid.rs:233` *(one-click suggestion, verified: 130 tests green)*

### ❓ question
- **External-bundle builder mutates git-tracked generated files; restore is `finally`-only, not signal-safe** — see inline at `crates/js/lib/build-prebid-external.mjs:263`
- **`config push --dry-run` safety rests entirely on the pinned edgezero-adapter rev** — see inline at `crates/trusted-server-cli/src/edgezero_delegate.rs:329`
- **`request_body_bytes` dead scaffolding with a latent stream-drop (rotate-key → default-key rotation)** — see inline at `crates/trusted-server-core/src/request_signing/endpoints.rs:27`

## Non-blocking

### 🤔 thinking / 🌱 seedling / 📝 note / ⛏ nitpick
- **Infallible `Result` + unused arg churn on body helpers** — see inline at `crates/trusted-server-core/src/publisher.rs:35`
- **CLI env/secret handling diverges between manifest-command and in-process paths** — see inline at `crates/trusted-server-cli/src/edgezero_delegate.rs:108`
- **Redundant/split HTTPS redirect check** — see inline at `crates/trusted-server-core/src/proxy.rs:1413`
- **Tester set/clear endpoints unauthenticated (design carry-over)** — see inline at `crates/trusted-server-core/src/tester_cookie.rs:78`
- **`OsanoConfig` lacks `deny_unknown_fields`** — see inline at `crates/trusted-server-core/src/integrations/osano.rs:21`
- **`validate_enabled_integrations` hardcodes the integration list** — see inline at `crates/trusted-server-cli/src/config_command.rs:187`
- **CLI crate is `edition = "2021"` vs documented 2024** — see inline at `crates/trusted-server-cli/Cargo.toml:5`

- 🤔 **Open-proxy fallback when `proxy.allowed_domains` is empty** — `validate_external_bundle_config` (`prebid.rs:245`) only enforces the allowlist when non-empty, and `redirect_is_permitted` (`proxy.rs:1211`) treats an empty allowlist as "permit all". With an empty list, a redirecting origin can send the bundle fetch to any HTTPS host (≤4 hops) and it's streamed as first-party JS with immutable cache; `require_https` is the only backstop. The host match itself is correctly dot-boundary-safe (`evil-example.com` ≠ `*.example.com` — verified). Consider requiring a non-empty allowlist whenever `external_bundle_url` is set.
- 🤔 **Two `edgezero-core` revs in the dependency graph** — `trusted-server-cli` pins `@2eeccc97` (edgezero-cli/adapter) while its dep `trusted-server-core` pins `@38198f98`, so the CLI's graph compiles two full copies of `edgezero-core`. Verified benign for correctness today (only `String`/`BTreeMap` cross the boundary), but it's build-cost + a latent footgun the moment an `edgezero_core` type crosses that seam. Align to one rev — the split looks accidental.
- 🌱 **Test-coverage gaps** — no end-to-end test for the proxy fetch+sanitize happy path (`handle_external_bundle` with a mock streaming client); no mjs-level test asserting `filename sha256 == hash(bytes)` and SRI derivation (the load-bearing CLI↔runtime contract); `buildMirrorPlan` untested with USP+GPP+TCF all present at once.
- 🤔 **Build-tooling robustness (benign today)** — `--out` resolves against cwd in the Rust CLI but `__dirname` in the mjs (safe only because the CLI always passes an absolute path); the mjs temp file is a fixed name with `emptyOutDir:false`, so concurrent/overlapping runs into one dir can hash a partial file. Align the mjs base to `process.cwd()` and PID-suffix the temp file.
- 📝 **Insecure-`certificate_check` warning relocated** — the old startup `log::warn!("INSECURE: proxy.certificate_check is disabled")` in `load_settings` is gone; the warning still fires per-backend at connect time (`adapter-fastly/src/backend.rs:220`), so it's not lost but the semantics shifted from once-at-startup to per-connection. Confirm operators still see it.

## 👍 Praise
- Proxy path disables **both** EC-ID and client-header forwarding for the static-asset fetch (`without_ec_id().without_forward_headers()`), rebuilds responses from a strict header allowlist, strips `Set-Cookie`, and forces `no-store` on non-200 to protect the immutable cache. Cache-mode/version logic robustly resists cache poisoning (exact-hash immutable mode, multi-`v` rejection).
- sha256/SRI formats line up across all three layers (mjs emitter → CLI manifest validator → runtime config validator) — the invariant this whole PR rests on.
- Osano consent mirror is genuinely fail-safe (never fabricates a permissive signal on missing/malformed input) with a real `mirrorGeneration` race guard and test.
- `write_atomic`/`ensure_output_dir_writable` and the argv-array `npm` invocation (no shell) are careful, well-tested systems code; the config payload is hash-verified with no panics on operator input; real domains/secrets were removed from source control.

## CI Status

- **All GitHub checks: not run** — the host-target CLI clippy/test workflows are *added* by this PR but haven't executed on this branch. Recommend pushing to trigger CI before merge, since the two-rev edgezero graph and the new CLI crate are only exercised there.

crates/trusted-server-js/lib/build-prebid-external.mjs
ChristianPavilonis: Addressed in the current branch: the external Prebid builder now generates adapter/user-id modules in a temporary directory and aliases those imports through Vite, so it no longer mutates the git-tracked generated files. The temp bundle filename is also PID/UUID-suffixed and relative `--out` paths resolve from the CLI cwd.

docs/guide/integrations/prebid.md
ChristianPavilonis: Done — Prebid external bundle generation now lives behind the `ts prebid bundle` CLI command.

crates/trusted-server-core/src/integrations/prebid.rs
ChristianPavilonis: Addressed: `validate_config_for_startup` now calls `validate_external_bundle_config`, so `ts config validate` exercises the same external-bundle startup checks as runtime registration.

crates/trusted-server-core/src/proxy.rs
ChristianPavilonis: Addressed: the redirect handling now performs one early `require_https && next_scheme != "https"` check immediately after `next_scheme` is computed, then falls through to the non-http/https passthrough case only for non-HTTPS-required policies.

crates/trusted-server-cli/src/edgezero_delegate.rs
ChristianPavilonis: Obsolete after the CLI refactor: `edgezero_delegate.rs` was removed and `ts config push` now delegates through EdgeZero typed config APIs directly (`edgezero_cli::run_config_push_typed::<TrustedServerAppConfig>`).

crates/trusted-server-core/src/request_signing/endpoints.rs
ChristianPavilonis: Addressed: `request_body_bytes` now checks `body.is_stream()` and returns a bad-request error for streamed admin bodies; rotate/deactivate streaming-body tests cover this path.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Addressed: `body_as_reader` is infallible again. `request_body_bytes` still returns `Result` now because it performs the streaming-body guard before converting to bytes.

crates/trusted-server-cli/src/edgezero_delegate.rs
ChristianPavilonis: Obsolete after the CLI refactor: the old in-process adapter delegate path was removed, so the manifest-command vs local adapter env/secret asymmetry no longer exists in this crate.

crates/trusted-server-core/src/tester_cookie.rs
ChristianPavilonis: Confirmed this posture is intentional. Added module docs in b82ab1e6 clarifying these are self-service tester routes gated by `tester_cookie.enabled`, that the cookie only affects tester routing, and that it must not gate sensitive behavior.

crates/trusted-server-core/src/integrations/osano.rs
ChristianPavilonis: Addressed in b82ab1e6: added `#[serde(deny_unknown_fields)]` to `OsanoConfig` plus tests for unknown-field rejection and deploy-time validation.

crates/trusted-server-cli/src/config_command.rs
ChristianPavilonis: Addressed in b82ab1e6: added registry builder IDs and a deploy-validation parity test asserting every registered integration builder is covered by deploy validation. Also added Osano to deploy validation.

crates/trusted-server-cli/Cargo.toml
ChristianPavilonis: Addressed in b82ab1e6: bumped `trusted-server-cli` to edition 2024 and verified with CLI tests/clippy.

crates/trusted-server-js/lib/build-all.mjs
```diff
@@ -252,11 +34,12 @@ const integrationModules = fs.existsSync(integrationsDir)
...
         return (
-          fs.statSync(fullPath).isDirectory() && fs.existsSync(path.join(fullPath, 'index.ts'))
+          name !== 'prebid' &&
+          fs.statSync(fullPath).isDirectory() &&
+          fs.existsSync(path.join(fullPath, 'index.ts'))
         );
       })
-      .sort()
-  : [];
+      : [];
```
aram356, line 43: ⛏ **nitpick** — Integration discovery lost its ordering: `.readdirSync().filter(...)` has no `.sort()`, so build order and the discovery log are filesystem-dependent (arbitrary on Linux CI). Final `dist/tsjs-<name>.js` artifacts are unaffected, but reproducible logs/builds are nicer. Prettier-clean, verified:

```suggestion
      })
      .sort()
  : [];
```

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -362,16 +551,185 @@ impl PrebidIntegration {
...
+        if let Some(content_encoding) = content_encoding {
+            sanitized
+                .headers_mut()
+                .insert(header::CONTENT_ENCODING, content_encoding);
+        }
+
+        if status == StatusCode::OK {
+            self.apply_external_bundle_headers(&mut sanitized, mode);
+        } else {
+            sanitized.headers_mut().insert(
```
aram356, line 706: 🌱 **seedling** — The non-200 branch of `sanitize_external_bundle_response` sets only `Cache-Control: no-store`, so an origin error body is served under the first-party path `/integrations/prebid/bundle.js` with no `Content-Type` and no `X-Content-Type-Options: nosniff`. Low impact (origin is https-allowlisted), but a compromised allowlisted origin could serve a sniffable first-party body. Consider setting `X-Content-Type-Options: nosniff` on both branches and an explicit `Content-Type` on the error branch.

crates/trusted-server-cli/src/prebid_bundle.rs
```diff
@@ -0,0 +1,916 @@
...
+    let request = PrebidBundleGenerateRequest {
+        js_lib_dir,
+        out_dir: out_dir.clone(),
+        adapters: config.adapters,
+        user_id_modules: config.user_id_modules,
+    };
+
+    generator.generate(&request, out, err)?;
+
+    let manifest_path = out_dir.join("manifest.json");
```
aram356, line 150: 🌱 **seedling** — `patch_config_metadata` rewrites `external_bundle_sha256`/`external_bundle_sri` but never `external_bundle_url`. Because the generator names the file `trusted-prebid-{sha256}.js` (content-addressed), any content change yields a new filename while the persisted URL still points at the old bundle — a config the runtime then rejects on integrity mismatch. The "Next:" hint warns generically; since the tool already knows `manifest.filename`, echo the exact new filename so the required URL edit is obvious.

trusted-server.example.toml
```diff
@@ -45,6 +45,10 @@ bidders = []
 debug = false
 client_side_bidders = []
 
+[integrations.prebid.bundle]
```
aram356, line 52: 📝 **note** — The template shows `[integrations.prebid.bundle]` (CLI input) but not the runtime keys an operator must set once they flip `enabled = true` — and enabling Prebid now *requires* `external_bundle_url` at deploy time (`validate_external_bundle_config`). A couple of commented example keys here would prevent a first-run validation surprise:

```toml
# external_bundle_url = "https://assets.example.com/prebid/trusted-prebid-<sha256>.js"
# external_bundle_sha256 = "<manifest sha256>"
# external_bundle_sri = "<manifest sri>"
```

Apply manually — outside the changed hunk.

aram356 (approved):
## Summary — Approve, conditioned on green CI ✅ (pending 🔧)

Second review pass (retargeted to `main`; force-pushed since pass 1). The Prebid external-bundle code is **merge-ready** and every finding from the first review was addressed correctly. I'm approving the code — but **this approval is contingent on the two required CI gates below going green**; branch protection should hold the merge until they do.

### Prior review fully addressed (verified)
- **CLI ↔ runtime validation parity** — `validate_config_for_startup` now calls `validate_external_bundle_config` (`prebid.rs:234`); runs at both deploy-audit and registry-build time.
- **Streaming-body guard** — `request_body_bytes` now rejects `is_stream()` with `BadRequest`.
- **Open-proxy on empty allowlist** — now a hard startup error requiring `proxy.allowed_domains` when an external bundle is configured (`prebid.rs:436`), so the open-mode path is unreachable for this route.
- **`OsanoConfig` `deny_unknown_fields`** — added, with a rejection test **and** deploy-validation coverage.
- **mjs hash test** — added and genuine (asserts `filename == trusted-prebid-${sha256(bytes)}.js` and the sha384 SRI against independently-computed digests).

Security posture verified sound: EC-ID + client-header forwarding disabled for the bundle fetch, `https_only` on initial **and** redirect hops, origin `Set-Cookie` stripped, `no-store` on non-200, multi-`v`/off-version → 404, and the one `expect()` on `external_bundle_sha256` is guarded by schema validation. sha256/SRI consistent across mjs → manifest → CLI validator → runtime.

## Blocking on CI (must be green before merge)

- 🔧 **`cargo fmt --check` fails (required gate).** Confirmed on CI and locally with the pinned rustfmt 1.95.0 — `main` passes locally, this branch does not (e.g. the PR-added `use super::{register, OsanoConfig};` at `osano.rs:53`, which rustfmt wants as `{OsanoConfig, register}`). The reordering pattern spans nearly every changed file, indicating the branch was formatted with a **newer rustfmt (1.96/nightly) than pinned 1.95.0**. **Fix:** run `cargo fmt --all` under the pinned toolchain and commit.
- 🔧 **`prepare integration artifacts` fails — dependency drift (required gate).** This PR bumped the workspace `Cargo.lock` (aws-lc-rs → 1.17.1, time → 0.3.53, …) but left `crates/trusted-server-integration-tests/Cargo.lock` on the old versions, so `check-integration-dependency-versions.sh` fails on 5 packages. **Fix** (from the CI output):
```bash
cargo update --manifest-path crates/trusted-server-integration-tests/Cargo.toml -p aws-lc-rs --precise 1.17.1
cargo update --manifest-path crates/trusted-server-integration-tests/Cargo.toml -p aws-lc-sys --precise 0.42.0
cargo update --manifest-path crates/trusted-server-integration-tests/Cargo.toml -p rustls-pki-types --precise 1.15.0
cargo update --manifest-path crates/trusted-server-integration-tests/Cargo.toml -p time --precise 0.3.53
cargo update --manifest-path crates/trusted-server-integration-tests/Cargo.toml -p time-macros --precise 0.2.31
```

## Non-blocking (optional polish — see inline)
- 🌱 Content-addressed rebuild leaves `external_bundle_url` stale (`patch_config_metadata` rewrites hash/SRI but not the URL) — see inline.
- 🌱 Error-path bundle responses drop `Content-Type` / set no `nosniff` — see inline.
- ⛏ `build-all.mjs` discovery is filesystem-ordered — see inline (one-click suggestion).
- 📝 Example config omits the runtime `external_bundle_*` keys — see inline.
- ⛏ `HashSet::with_capacity(DEPLOY_VALIDATED_INTEGRATION_IDS.len())` at `config.rs:136` oversizes a set that holds ≤3 auction-provider IDs — cosmetic; `HashSet::new()` reads clearer.
- 🌱 The bundle generator buffers all stdout/stderr (`prebid_bundle.rs:70`) and shows nothing until esbuild exits — consider inherited/streamed stdio.

## 👍 Praise
Strong, responsive iteration — all pass-1 findings fixed correctly. The `IntegrationBuilder`/`registered_builder_ids()` refactor plus the `deploy_validation_covers_registered_integration_builders` test is nice defensive design; the argv-array `npm` invocation stays injection-free; the mjs hash test is a real known-answer test.

## CI Status (against `main`)
- `cargo fmt`: **FAIL** (required) · `prepare integration artifacts`: **FAIL** (required)
- CodeQL, `cargo test`, `cargo test (axum native)`, `cargo check (cloudflare)`, `cargo check/build/test (spin)`, `cargo test (cross-adapter parity)`, `vitest`, `format-typescript`, `format-docs`, `Analyze (rust/js/actions)`: **PASS**
- `integration tests`, `browser integration tests`, `integration tests (EdgeZero entry point)`: **SKIPPED**

crates/trusted-server-js/lib/build-all.mjs
ChristianPavilonis: Addressed in 59045980: added `.sort()` to integration discovery so the build order/logs are deterministic. Verified with `cd crates/trusted-server-js/lib && npm run format` and `npx vitest run`.

crates/trusted-server-core/src/integrations/prebid.rs
ChristianPavilonis: Addressed in 59045980: external bundle responses now set `X-Content-Type-Options: nosniff` on both success and error paths, and the non-200 path now sets an explicit `text/plain; charset=utf-8` `Content-Type`. Updated the sanitization tests and verified with `cargo test-fastly`.

crates/trusted-server-cli/src/prebid_bundle.rs
ChristianPavilonis: Addressed in 59045980: the `Next:` guidance now includes `manifest.filename`, so operators see the exact content-addressed bundle filename to upload/reference. Added CLI test coverage for the hint.

trusted-server.example.toml
ChristianPavilonis: Addressed in 59045980: added commented `external_bundle_url`, `external_bundle_sha256`, and `external_bundle_sri` runtime keys to `trusted-server.example.toml`.

ChristianPavilonis:
Follow-up commit 59045980 addresses the approval-pass follow-ups:

- Ran pinned `cargo fmt --all` and fixed the rustfmt drift.
- Updated `crates/trusted-server-integration-tests/Cargo.lock` to match workspace dependency versions.
- Applied the optional Prebid polish from the inline comments: deterministic JS integration discovery, `nosniff`/explicit content type for external-bundle error responses, exact generated filename in the CLI next-step hint, and commented runtime bundle keys in the example config.

Local verification:
- `./scripts/check-integration-dependency-versions.sh`
- `./scripts/test-cli.sh`
- `cargo test-fastly`
- `cargo clippy-fastly`
- host CLI clippy with `-D warnings`
- `cd crates/trusted-server-js/lib && npm run format && npx vitest run`

PR #753, opened 2026-06-06
Avoid allocation when detecting content encoding

### LINKED ISSUE #434: Compression::from_content_encoding allocates String for case comparison
## Summary

Uses `.to_lowercase()` (allocating a String) for case-insensitive matching on every response. Should use `eq_ignore_ascii_case` instead.

## Refs

- `crates/common/src/streaming_processor.rs` lines 46-53

## Recommendation

Replace `.to_lowercase().as_str()` match with `eq_ignore_ascii_case` comparisons.

## Context

ChristianPavilonis:
## Summary

- Avoid allocating a lowercase `String` when detecting response compression from `Content-Encoding`.
- Preserve existing ASCII case-insensitive matching for `gzip`, `deflate`, and `br`.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/streaming_processor.rs` | Replaced `to_lowercase()` matching in `Compression::from_content_encoding` with non-allocating `eq_ignore_ascii_case` checks. |

## Closes

Closes #434

## Test plan

- [x] Other: `cargo test -p trusted-server-core streaming_processor::tests::test_compression_detection`; `cargo test -p trusted-server-core`

## Checklist

- [x] New code has tests

### DIFF
--- a/crates/trusted-server-core/src/streaming_processor.rs
+++ b/crates/trusted-server-core/src/streaming_processor.rs
@@ -52,11 +52,14 @@ impl Compression {
     /// Detect compression from content-encoding header
     #[must_use]
     pub fn from_content_encoding(encoding: &str) -> Self {
-        match encoding.to_lowercase().as_str() {
-            "gzip" => Self::Gzip,
-            "deflate" => Self::Deflate,
-            "br" => Self::Brotli,
-            _ => Self::None,
+        if encoding.eq_ignore_ascii_case("gzip") {
+            Self::Gzip
+        } else if encoding.eq_ignore_ascii_case("deflate") {
+            Self::Deflate
+        } else if encoding.eq_ignore_ascii_case("br") {
+            Self::Brotli
+        } else {
+            Self::None
         }
     }
 }
--- a/crates/trusted-server-core/src/streaming_processor.rs
+++ b/crates/trusted-server-core/src/streaming_processor.rs
@@ -52,14 +52,11 @@ impl Compression {
     /// Detect compression from content-encoding header
     #[must_use]
     pub fn from_content_encoding(encoding: &str) -> Self {
-        if encoding.eq_ignore_ascii_case("gzip") {
-            Self::Gzip
-        } else if encoding.eq_ignore_ascii_case("deflate") {
-            Self::Deflate
-        } else if encoding.eq_ignore_ascii_case("br") {
-            Self::Brotli
-        } else {
-            Self::None
+        match encoding {
+            s if s.eq_ignore_ascii_case("gzip") => Self::Gzip,
+            s if s.eq_ignore_ascii_case("deflate") => Self::Deflate,
+            s if s.eq_ignore_ascii_case("br") => Self::Brotli,
+            _ => Self::None,
         }
     }
 }


crates/trusted-server-core/src/streaming_processor.rs
```diff
@@ -52,11 +52,14 @@ impl Compression {
...
-            _ => Self::None,
+        if encoding.eq_ignore_ascii_case("gzip") {
+            Self::Gzip
+        } else if encoding.eq_ignore_ascii_case("deflate") {
+            Self::Deflate
+        } else if encoding.eq_ignore_ascii_case("br") {
+            Self::Brotli
+        } else {
+            Self::None
         }
```
aram356: ⛏ **nitpick** — Optional, purely stylistic: the original `match` shape can be kept without re-introducing an allocation by matching with guard arms. Identical behavior and performance to the `if`/`else if` chain — take it only if you prefer the `match` form; the PR is good as-is.

_(scratch-verified in an isolated worktree: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, and `cargo test --workspace` all pass.)_

```suggestion
        match encoding {
            s if s.eq_ignore_ascii_case("gzip") => Self::Gzip,
            s if s.eq_ignore_ascii_case("deflate") => Self::Deflate,
            s if s.eq_ignore_ascii_case("br") => Self::Brotli,
            _ => Self::None,
        }
```

aram356 (approved):
## Summary

Clean, behavior-preserving micro-optimization: `Compression::from_content_encoding` drops the per-call `to_lowercase()` `String` allocation in favor of non-allocating `eq_ignore_ascii_case` checks. Since the only match targets are ASCII tokens (`gzip`/`deflate`/`br`), ASCII case-folding is exactly equivalent to the prior Unicode lowercasing for every possible input — and is the standards-correct choice for HTTP content-coding tokens. The existing `test_compression_detection` (including the mixed-case `GZIP` case) already guards the behavior, and all CI checks pass. **Approving.**

> 1 of the inline comments below carries a one-click GitHub `suggestion` (optional, purely stylistic). The remaining finding is a body-level note about pre-existing surrounding code.

## Non-blocking

### ⛏ nitpick

- **Optional `match`-with-guards form** — see inline at `crates/trusted-server-core/src/streaming_processor.rs:55`

- 🌱 **Pre-existing case-sensitivity mismatch around encoding handling (out of scope — flagging for a follow-up, no change needed here)** — `from_content_encoding` matches case-insensitively (`GZIP` → `Gzip`), but the gate `is_supported_content_encoding` (`crates/trusted-server-core/src/publisher.rs:625`) uses a case-*sensitive* `matches!(encoding, "" | "identity" | "gzip" | "deflate" | "br")`. In the publisher path the case-sensitive gate runs first (`publisher.rs:352`, `publisher.rs:539`), so a mixed-case `GZIP` is treated as *unsupported* and bypasses rewriting — while the proxy path (`proxy.rs:235`) would decompress it. This predates and is orthogonal to this PR; noting it only because a deep pass was requested. Worth a follow-up issue rather than expanding this PR's scope.

## CI Status

- integration tests: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS
- cargo test: PASS (required)
- cargo fmt: PASS (required)
- vitest: PASS

PR #754, opened 2026-06-06
Use valid image proxy content type fallback

### LINKED ISSUE #429: image/* is not a valid Content-Type for responses
## Summary

When the origin omits a Content-Type for image responses, the fallback is set to `image/*`. Wildcard MIME types are only valid in Accept headers, not Content-Type. Browsers may not render the image correctly.

## Refs

- `crates/common/src/proxy.rs` lines 247-248

## Recommendation

Use `application/octet-stream` as the fallback.

## Context

ChristianPavilonis:
## Summary

- Replaces the invalid `image/*` response `Content-Type` fallback with `application/octet-stream`.
- Covers both proxy response finalization paths and updates the first-party proxy guide.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/proxy.rs` | Added a shared image fallback content-type constant, used it in normal and streaming finalization paths, and updated tests. |
| `docs/guide/first-party-proxy.md` | Updated image handling docs to describe the valid fallback and trigger behavior. |

## Closes

Closes #429

## Test plan

- [x] Other: `cargo test -p trusted-server-core proxy`; `cargo test -p trusted-server-core`

## Checklist

- [x] Uses tracing macros (not `println!`)
- [x] New code has tests

### DIFF
--- a/crates/trusted-server-core/src/proxy.rs
+++ b/crates/trusted-server-core/src/proxy.rs
@@ -18,6 +18,9 @@ use crate::streaming_processor::{Compression, PipelineConfig, StreamProcessor, S
 /// Chunk size used for streaming content through the rewrite pipeline.
 const STREAMING_CHUNK_SIZE: usize = 8192;
 
+/// Fallback `Content-Type` for image-like responses missing an origin type.
+const IMAGE_FALLBACK_CONTENT_TYPE: &str = "application/octet-stream";
+
 #[derive(Deserialize)]
 struct ProxySignReq {
     url: String,
@@ -256,7 +259,7 @@ fn finalize_proxied_response(
         );
     }
 
-    // Image handling: set generic content-type if missing and log pixel heuristics
+    // Image handling: set a valid fallback content type if missing and log pixel heuristics
     let req_accept_images = req
         .get_header(HEADER_ACCEPT)
         .and_then(|h| h.to_str().ok())
@@ -265,7 +268,7 @@ fn finalize_proxied_response(
 
     if ct.starts_with("image/") || req_accept_images {
         if beresp.get_header(header::CONTENT_TYPE).is_none() {
-            beresp.set_header(header::CONTENT_TYPE, "image/*");
+            beresp.set_header(header::CONTENT_TYPE, IMAGE_FALLBACK_CONTENT_TYPE);
         }
 
         // Heuristics to log likely tracking pixels without altering response
@@ -344,7 +347,7 @@ fn finalize_proxied_response_streaming(
 
     if ct.starts_with("image/") || req_accept_images {
         if beresp.get_header(header::CONTENT_TYPE).is_none() {
-            beresp.set_header(header::CONTENT_TYPE, "image/*");
+            beresp.set_header(header::CONTENT_TYPE, IMAGE_FALLBACK_CONTENT_TYPE);
         }
 
         let mut is_pixel = false;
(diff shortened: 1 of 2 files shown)


aram356 (approved):
## Summary

Correct, well-scoped fix. `image/*` is not a valid response `Content-Type` (the
`*` wildcard belongs in an `Accept` header, not a concrete media type), so
falling back to `application/octet-stream` when the origin omits a type is the
right standard choice. Both finalization paths are covered via a shared
constant, the streaming path gains a previously-missing test, and the docs are
updated accurately. No blocking findings — the change is merge-ready. A few
non-blocking notes below for awareness only.

- 🌱 **Future: content-sniff for a precise image fallback** — The fallback fires
  only when the origin sent no `Content-Type` at all (the `is_none()` guard at
  `crates/trusted-server-core/src/proxy.rs:499` / `:586`), so
  `application/octet-stream` ("unknown binary") is the honest answer today. A
  future enhancement could sniff the leading bytes (`\x89PNG`, `GIF8`, `\xFF\xD8`
  for JPEG) to emit a concrete `image/...` type. Not worth doing in this PR —
  it'd only affect direct-navigation display, and ``/CSS contexts sniff
  bytes regardless of the declared type.

- 🤔 **Direct-navigation behavior with `octet-stream`** — Worth a one-line
  confirmation that download-on-direct-navigation (rather than inline display)
  is acceptable for this proxy's use cases. It is for the `` / tracking-pixel
  flows this path serves (browsers sniff for ``, and pixels don't care about
  the declared type), so there's no functional regression versus the old invalid
  value — flagging only for explicit awareness.

- 📝 **Incidental blank line** — `crates/trusted-server-core/src/proxy.rs:270`
  adds a formatting-only blank line between `stream_asset_body` and
  `ProxySignReq`, unrelated to the fix. Harmless (`cargo fmt` passes); mention
  only.

## CI Status

- cargo test: PASS (required)
- cargo fmt: PASS (required)
- vitest: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS
- CodeQL: PASS
- integration tests: PASS

crates/trusted-server-core/src/proxy.rs
```diff
@@ -493,9 +497,10 @@ fn finalize_proxied_response(
 
     if ct.starts_with("image/") || req_accept_images {
         if beresp.headers().get(header::CONTENT_TYPE).is_none() {
-            beresp
-                .headers_mut()
-                .insert(header::CONTENT_TYPE, HeaderValue::from_static("image/*"));
+            beresp.headers_mut().insert(
+                header::CONTENT_TYPE,
+                HeaderValue::from_static(IMAGE_FALLBACK_CONTENT_TYPE),
```
aram356, line 502: 🌱 **seedling** — Future: content-sniff for a precise image fallback.

This insert fires only when the origin sent **no** `Content-Type` at all (the `is_none()` guard just above), so `application/octet-stream` ("unknown binary") is the honest answer here. A future enhancement could sniff the leading bytes (`\x89PNG`, `GIF8`, `\xFF\xD8` for JPEG) to emit a concrete `image/...` type. Only matters for direct-navigation display — `` / CSS contexts sniff bytes regardless of the declared type. Out of scope for this PR.

aram356, line 502: 🤔 **thinking** — `application/octet-stream` makes a browser *download* on direct navigation rather than render inline.

That's fine for the `` / tracking-pixel flows this path serves (browsers sniff bytes for ``, and pixels ignore the declared type), and it's strictly better than the old invalid `image/*`. Flagging only to confirm no use case here depends on direct inline display of the proxied image.

prk-Jr (approved):

PR #759, opened 2026-06-08
feat: make integration proxy paths configurable per customer

### LINKED ISSUE #741: Make integration proxy paths configurable per customer
## Context

Several integrations currently use fixed, predictable proxy paths:

- `/integrations/didomi/consent` (Didomi integration)
- `/integrations/addingwell/*` (Addingwell integration)

Using shared, hardcoded paths increases the risk of ad blockers adding them to their block lists, which would defeat the purpose of first-party proxying.

## Proposal

Allow the proxy path to be configurable per customer for each integration, for example:

```toml
[integrations.didomi]
proxy_path = "my-custom-consent-path" # default: "integrations/didomi"

[integrations.addingwell]
proxy_path = "my-custom-addingwell-path" # default: "integrations/addingwell"
```

## Trade-offs to discuss

- A randomized or customer-specific path is harder for ad blockers to target
- It adds configuration complexity for the customer setup
- All script serving snippets and redirects must reflect the configured path at render time

vasujain00:
Fixes #741

## Summary

- Adds a `proxy_prefix()` method to the `IntegrationProxy` trait allowing integrations to use customer-configured paths instead of hardcoded `/integrations/{name}` paths
- Implements configurable `proxy_path` for the Didomi integration as the first adopter
- Backwards-compatible: default behavior unchanged when `proxy_path` is not set

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/integrations/registry.rs` | Added `proxy_prefix()` method to `IntegrationProxy` trait; updated `get()`, `post()`, `put()`, `delete()`, `patch()` helpers to use it |
| `crates/trusted-server-core/src/integrations/didomi.rs` | Added `proxy_path` config field; overrides `proxy_prefix()` when set; updated routes and path stripping |

## Closes

Closes #741

## Test plan

- [x] Other: New test `registers_custom_proxy_path` verifies custom path routes correctly and default path no longer matches

## Checklist

- [x] No `unwrap()` in production code -- use `expect("should ...")`
- [x] New code has tests

### DIFF
--- a/crates/trusted-server-core/src/integrations/didomi.rs
+++ b/crates/trusted-server-core/src/integrations/didomi.rs
@@ -15,7 +15,7 @@ use crate::platform::RuntimeServices;
 use crate::settings::{IntegrationConfig, Settings};
 
 const DIDOMI_INTEGRATION_ID: &str = "didomi";
-const DIDOMI_PREFIX: &str = "/integrations/didomi/consent";
+const DIDOMI_DEFAULT_PREFIX: &str = "/integrations/didomi/consent";
 
 /// Configuration for the Didomi consent notice reverse proxy.
 #[derive(Debug, Clone, Deserialize, Serialize, Validate)]
@@ -23,6 +23,10 @@ pub struct DidomiIntegrationConfig {
     /// Whether the integration is enabled.
     #[serde(default = "default_enabled")]
     pub enabled: bool,
+    /// Custom proxy path prefix to avoid ad-blocker detection.
+    /// Defaults to "integrations/didomi/consent" if not set.
+    #[serde(default)]
+    pub proxy_path: Option<String>,
     /// Base URL for the Didomi SDK origin.
     #[serde(default = "default_sdk_origin")]
     #[validate(url)]
@@ -192,8 +196,15 @@ impl IntegrationProxy for DidomiIntegration {
         DIDOMI_INTEGRATION_ID
     }
 
+    fn proxy_prefix(&self) -> String {
+        match &self.config.proxy_path {
+            Some(custom) => format!("/{}", custom.trim_start_matches('/')),
+            None => DIDOMI_DEFAULT_PREFIX.to_string(),
+        }
+    }
+
     fn routes(&self) -> Vec<IntegrationEndpoint> {
-        vec![self.get("/consent/*"), self.post("/consent/*")]
+        vec![self.get("/*"), self.post("/*")]
     }
 
     async fn handle(
@@ -203,7 +214,8 @@ impl IntegrationProxy for DidomiIntegration {
         req: Request,
     ) -> Result<Response, Report<TrustedServerError>> {
         let path = req.get_path();
-        let consent_path = path.strip_prefix(DIDOMI_PREFIX).unwrap_or(path);
+        let prefix = self.proxy_prefix();
+        let consent_path = path.strip_prefix(&prefix).unwrap_or(path);
         let backend = self.backend_for_path(consent_path);
         let base_origin = match backend {
             DidomiBackend::Sdk => self.config.sdk_origin.as_str(),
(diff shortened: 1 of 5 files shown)


crates/trusted-server-core/src/integrations/didomi.rs
```diff
@@ -14,14 +14,18 @@ use crate::integrations::{IntegrationEndpoint, IntegrationProxy, IntegrationRegi
...
 /// Configuration for the Didomi consent notice reverse proxy.
 #[derive(Debug, Clone, Deserialize, Serialize, Validate)]
 pub struct DidomiIntegrationConfig {
     /// Whether the integration is enabled.
     #[serde(default = "default_enabled")]
     pub enabled: bool,
+    /// Custom proxy path prefix to avoid ad-blocker detection.
+    /// Defaults to "integrations/didomi/consent" if not set.
+    #[serde(default)]
+    pub proxy_path: Option<String>,
```
ChristianPavilonis, line 33: **Custom path also needs to reach the client-side URL builder**

Thanks for adding the config knob here. One important gap: this field only changes the Rust route registration, but the Didomi browser module still builds `sdkPath` from the hard-coded `/integrations/didomi/consent/` path in `crates/js/lib/src/integrations/didomi/index.ts`. With `proxy_path` set, this PR registers only the custom route and the new test asserts the old route no longer matches, so pages would still request the old prefix and miss the proxy.

Could we pass the resolved prefix to the JS runtime (for example via a small head-injected `window.__tsjs_didomi.proxyPath` config) and have the TypeScript fall back to the current default when it is absent?

crates/trusted-server-core/src/integrations/didomi.rs
```diff
@@ -191,8 +195,15 @@ impl IntegrationProxy for DidomiIntegration {
         DIDOMI_INTEGRATION_ID
     }
 
+    fn proxy_prefix(&self) -> String {
+        match &self.config.proxy_path {
+            Some(custom) => format!("/{}", custom.trim_start_matches('/')),
```
ChristianPavilonis: **Please validate or canonicalize `proxy_path` before using it for routes**

This normalizes leading slashes, but malformed values can still register surprising routes. For example, `proxy_path = "my-custom-consent/"` becomes `/my-custom-consent//*`, so the intended `/my-custom-consent/loader.js` request would not match. Empty/root paths, `//`, query/fragment characters, and matchit route metacharacters would also be risky here.

I’d suggest validating/canonicalizing once at config load or construction time: allow an optional leading slash, store/return exactly `/{path}`, and reject empty/root/trailing-slash/query/fragment/wildcard/route-metacharacter values. A few tests around leading slash normalization and invalid values would make this much safer for a customer-facing config field.

ChristianPavilonis (commented):
Thanks for the contribution — making these proxy paths configurable is a good direction, and I appreciate that the default route behavior is covered. I found two implementation gaps that look worth addressing before relying on custom Didomi paths.

I left the main items inline. One additional follow-up: once the behavior is wired through, please update `docs/guide/integrations/didomi.md` to document `proxy_path`, its default, accepted format, and the custom endpoint examples.

crates/trusted-server-core/src/integrations/didomi.rs
vasujain00: Addressed ✅ — `DidomiIntegration` now implements `IntegrationHeadInjector` and injects:

`<script>window.__tsjs_didomi={proxyPath:"/custom-path/"};</script>`

The TypeScript module reads `window.__tsjs_didomi?.proxyPath` at runtime, falling back to `/integrations/didomi/consent/` when absent. Follows the same pattern as `window.__tsjs_prebid`.

crates/trusted-server-core/src/integrations/didomi.rs
vasujain00: Done ✅ — Added `validate_proxy_path` custom validator (runs at config load via `config.validate()`). Rejects:

- Empty / root-only paths
- Trailing slashes
- Double slashes (`//`)
- matchit metacharacters: `?`, `#`, `{`, `}`, `*`, spaces

Leading slash is optional in config — normalized to `/{path}` internally. 5 unit tests added covering valid/invalid inputs.

ChristianPavilonis (approved):

PR #769, opened 2026-06-12
Add DataDome server-side protection

### LINKED ISSUE #317: Add server-side bot protection via DataDome Protection API

ChristianPavilonis:
## Summary

- Adds DataDome server-side Protection API validation before route matching, with fail-open behavior and challenge short-circuiting.
- Adds request-filter infrastructure for integrations to mutate upstream request headers and final response headers.
- Moves the DataDome server-side key to runtime Secret Store configuration and documents staging/deployment behavior.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/integrations/registry.rs` | Adds integration request-filter traits, decisions, header effects, mutation semantics, and registry execution. |
| `crates/trusted-server-core/src/integrations/datadome.rs` | Adds DataDome protection config, Secret Store key references, auto-injection, and request-filter registration. |
| `crates/trusted-server-core/src/integrations/datadome/protection.rs` | Implements protection matching, payload construction, Secret Store key loading/cache, Protection API calls, response classification, and pointer-header extraction. |
| `crates/trusted-server-adapter-fastly/src/main.rs` | Runs request filters after basic auth and before route matching; applies response effects after finalization. |
| `crates/trusted-server-adapter-fastly/src/platform.rs` / `crates/trusted-server-core/src/platform/types.rs` | Extends client metadata with JA4/H2 fingerprint and edge server metadata. |
| `crates/trusted-server-adapter-fastly/src/route_tests.rs` | Adds route-level coverage for DataDome protection behavior. |
| `crates/trusted-server-core/src/platform/test_support.rs` / `src/http_util.rs` | Updates test support/default client metadata helpers. |
| `crates/trusted-server-core/Cargo.toml` | Adds dependencies needed by the DataDome protection implementation. |
| `trusted-server.toml` | Documents new DataDome protection, Secret Store, and auto-injection settings. |
| `docs/guide/integrations/datadome.md` | Documents first-party proxy, server-side protection, fail-open behavior, headers, Secret Store config, and GraphQL v1 limitation. |
| `docs/superpowers/specs/2026-06-11-datadome-server-side-protection-design.md` | Adds and updates the implementation spec/status for the DataDome server-side protection work. |

## Closes

Closes #317

## Test plan

- [x] Other: `cd docs && npx prettier --check guide/integrations/datadome.md superpowers/specs/2026-06-11-datadome-server-side-protection-design.md`

## Checklist

- [x] Uses logging macros (not `println!`)
- [x] New code has tests

### DIFF
--- a/crates/trusted-server-core/src/integrations/datadome.rs
+++ b/crates/trusted-server-core/src/integrations/datadome.rs
@@ -64,17 +64,22 @@ use http::header;
 use http::{Method, StatusCode};
 use regex::Regex;
 use serde::Deserialize;
+use serde_json::Value as JsonValue;
 use validator::Validate;
 
 use crate::error::TrustedServerError;
 use crate::integrations::{
     collect_body_bounded, collect_response_bounded, ensure_integration_backend,
     AttributeRewriteAction, IntegrationAttributeContext, IntegrationAttributeRewriter,
-    IntegrationEndpoint, IntegrationProxy, IntegrationRegistration, INTEGRATION_MAX_BODY_BYTES,
-    UPSTREAM_SDK_MAX_RESPONSE_BYTES,
+    IntegrationEndpoint, IntegrationHeadInjector, IntegrationHtmlContext, IntegrationProxy,
+    IntegrationRegistration, IntegrationRequestFilter, RequestFilterDecision, RequestFilterInput,
+    INTEGRATION_MAX_BODY_BYTES, UPSTREAM_SDK_MAX_RESPONSE_BYTES,
 };
 use crate::platform::{PlatformHttpRequest, RuntimeServices};
+use crate::redacted::Redacted;
 use crate::settings::{IntegrationConfig, Settings};
+
+mod protection;
 
 const DATADOME_INTEGRATION_ID: &str = "datadome";
 
@@ -127,6 +132,52 @@ pub struct DataDomeConfig {
     /// Whether to rewrite `DataDome` script URLs in HTML to first-party paths
     #[serde(default = "default_rewrite_sdk")]
     pub rewrite_sdk: bool,
+
+    /// Whether to call DataDome Protection API before route matching.
+    #[serde(default)]
+    pub enable_protection: bool,
+
+    /// DataDome server-side key used for Protection API validation.
+    #[serde(default)]
+    pub server_side_key: Redacted<String>,
+
+    /// Base URL for the DataDome Protection API.
+    #[serde(default = "default_protection_api_origin")]
+    #[validate(url)]
+    pub protection_api_origin: String,
+
+    /// First-byte timeout for Protection API calls, in milliseconds.
+    #[serde(default = "default_timeout_ms")]
+    #[validate(range(min = 1, max = 10000))]
+    pub timeout_ms: u32,
+
+    /// Regex for URLs to exclude from Protection API validation.
+    #[serde(default = "default_url_pattern_exclusion")]
+    pub url_pattern_exclusion: String,
+
+    /// Regex for URLs to include in Protection API validation.
+    #[serde(default)]
+    pub url_pattern_inclusion: String,
+
+    /// Reserved flag for future GraphQL payload extraction.
+    #[serde(default)]
+    pub enable_graphql_support: bool,
+
+    /// DataDome client-side key used for auto-injecting the browser tag.
+    #[serde(default)]
+    pub client_side_key: String,
+
+    /// Whether to auto-inject the DataDome browser tag when a client-side key exists.
+    #[serde(default = "default_inject_client_side_tag")]
+    pub inject_client_side_tag: bool,
+
+    /// URL used for the injected DataDome browser tag.
+    #[serde(default = "default_client_side_tag_url")]
+    pub client_side_tag_url: String,
+
+    /// Options assigned to `window.ddoptions` before loading the browser tag.
+    #[serde(default = "default_client_side_configuration")]
+    pub client_side_configuration: JsonValue,
 }
 
 fn default_enabled() -> bool {
@@ -149,6 +200,30 @@ fn default_rewrite_sdk() -> bool {
     true
 }
 
+fn default_protection_api_origin() -> String {
+    "https://api-fastly.datadome.co".to_string()
+}
+
+fn default_timeout_ms() -> u32 {
+    1500
+}
+
+fn default_url_pattern_exclusion() -> String {
+    r"\.(avi|flv|mka|mkv|mov|mp4|mpeg|mpg|mp3|flac|ogg|ogm|opus|wav|webm|webp|bmp|gif|ico|jpeg|jpg|png|svg|svgz|swf|eot|otf|ttf|woff|woff2|css|less|js|map)$".to_string()
+}
+
+fn default_inject_client_side_tag() -> bool {
+    true
+}
+
+fn default_client_side_tag_url() -> String {
+    "/integrations/datadome/tags.js".to_string()
+}
+
+fn default_client_side_configuration() -> JsonValue {
+    serde_json::json!({ "ajaxListenerPath": true })
+}
+
 impl Default for DataDomeConfig {
     fn default() -> Self {
         Self {
@@ -157,6 +232,17 @@ impl Default for DataDomeConfig {
             api_origin: default_api_origin(),
             cache_ttl_seconds: default_cache_ttl(),
             rewrite_sdk: default_rewrite_sdk(),
+            enable_protection: false,
+            server_side_key: Redacted::default(),
+            protection_api_origin: default_protection_api_origin(),
+            timeout_ms: default_timeout_ms(),
+            url_pattern_exclusion: default_url_pattern_exclusion(),
+            url_pattern_inclusion: String::new(),
+            enable_graphql_support: false,
+            client_side_key: String::new(),
+            inject_client_side_tag: default_inject_client_side_tag(),
+            client_side_tag_url: default_client_side_tag_url(),
+            client_side_configuration: default_client_side_configuration(),
         }
     }
 }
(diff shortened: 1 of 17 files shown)


crates/trusted-server-core/src/integrations/datadome/protection.rs
```diff
@@ -0,0 +1,795 @@
...
+    pub(super) async fn filter_protection_request(
+        &self,
+        input: RequestFilterInput<'_>,
+    ) -> RequestFilterDecision {
+        if !self.config.enable_protection || !self.is_request_protected(&input) {
+            return RequestFilterDecision::Continue(RequestFilterEffects::default());
+        }
+
+        match self.filter_protection_request_inner(input).await {
+            Ok(decision) => decision,
```
aram356, line 44: 🤔 **thinking** — Persistent misconfiguration fails open silently, indistinguishable from a transient API error.

`filter_protection_request_inner` returns `Err` both for transient Protection API failures *and* for permanent setup problems — an empty/missing secret **value** (startup validation in `try_new` can only confirm the store name and secret name are configured, not that the secret exists), or an unparseable backend URL. All of them land on this single per-request `warn` and continue. With `enable_protection = true` and the secret absent from the store, protection is effectively disabled for the lifetime of every instance while the operator believes requests are being validated.

Consider distinguishing setup/config errors (log at `error!`, and/or surface them once rather than once-per-request) from transient API timeouts so a permanent misconfiguration is loud. Apply manually — the change belongs in the inner-call error classification, not on this one line.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -473,6 +689,55 @@ impl IntegrationProxy for DataDomeIntegration {
...
+            .replace("</", "<\\/");
+        let options = serde_json::to_string(&self.config.client_side_configuration)
+            .unwrap_or_else(|err| {
+                log::warn!("[datadome] Failed to serialize client-side configuration: {err}");
+                "{}".to_string()
+            })
+            .replace("</", "<\\/");
+
+        vec![format!(
+            "<script>window.ddjskey={key};window.ddoptions={options};</script><script src={tag_url} async></script>"
```
aram356: 🌱 **seedling** — `client_side_tag_url` is JSON-escaped but rendered into an HTML *attribute*, not a JS string.

`key` and `options` sit in `<script>` (JS) context, where `serde_json` plus the `</` → `<\/` replacement is the correct mitigation. `tag_url`, however, is emitted as `src={tag_url}` — an HTML attribute. JSON escaping renders an embedded quote as `\"`, which does **not** prevent attribute breakout: a value such as `x" onerror="…` closes the `src` attribute and injects a new one. The value is operator-controlled config so exploitability is low, but the escaping is inconsistent with the JS-context fields right beside it.

Suggested fix (apply manually — needs validation/escaping logic, not a single-line swap): validate `client_side_tag_url` as a relative path or `https` origin at config time, or HTML-attribute-escape it before interpolation.

crates/trusted-server-core/src/integrations/datadome/protection.rs
```diff
@@ -0,0 +1,795 @@
...
+const HEADER_DATADOME_X_SET_COOKIE: &str = "x-datadome-x-set-cookie";
+const DATADOME_COOKIE_NAME: &str = "datadome";
+
+#[derive(Debug, Clone, Eq, Hash, PartialEq)]
+struct DataDomeServerSideKeyCacheKey {
+    secret_store: String,
+    secret_name: String,
+}
+
+static DATADOME_SERVER_SIDE_KEY_CACHE: LazyLock<
```
aram356: 📝 **note** — Unlike `IP_CIDR_SOURCE_CACHE` (TTL-based), this server-side key cache never expires within an instance. That is fine for short-lived Fastly instances, but the asymmetry with the CIDR cache is worth a one-line rationale comment so a future reader doesn't mistake it for an oversight relative to key rotation.

crates/trusted-server-core/src/integrations/datadome/protection.rs
```diff
@@ -0,0 +1,795 @@
...
+                if body.is_stream() {
+                    log::warn!(
+                        "[datadome] Protection API challenge body was streaming; failing open"
+                    );
+                    return RequestFilterDecision::Continue(RequestFilterEffects::default());
+                }
+                let body_bytes = body.into_bytes();
+                EdgeBody::from(body_bytes.as_ref().to_vec())
+            };
+            let challenge = Response::builder()
```
aram356, line 395: 🤔 **thinking** — The rebuilt challenge response carries only status + body; every other header (Content-Type, and `Location` for the `301`/`302` cases) must arrive via DataDome's `x-datadome-headers` pointer and be applied as response effects. That matches DataDome's contract, but it's an implicit external dependency: if the pointer omits `Content-Type` the challenge page renders without it, and a `301`/`302` whose `Location` is not listed in the pointer becomes a no-op redirect. A `301`/`302` challenge test asserting `Location` lands from effects would lock this behavior in (only `403` is covered today).

aram356 (approved):
## Summary

Adds DataDome server-side Protection API validation plus generic integration request-filter infrastructure. The implementation is solid: response effects are applied on every `HandlerOutcome` path, integration config is validated at load time, fail-open is deliberate and tested, and the new filter infra is cleanly factored and unit-tested. All 14 CI checks pass, and I separately verified the `wasm32-wasip1` release build (which CI does not gate) compiles. Approving — every finding below is non-blocking.

## Non-blocking

### 🤔 thinking / 🌱 seedling / 📝 note

- **Config/setup errors fail open silently** — see inline at `crates/trusted-server-core/src/integrations/datadome/protection.rs:51`
- **Head-injector attribute-context escaping** — see inline at `crates/trusted-server-core/src/integrations/datadome.rs:736`
- **Server-side key cache has no TTL** — see inline at `crates/trusted-server-core/src/integrations/datadome/protection.rs:37`
- **Challenge response depends on the pointer header** — see inline at `crates/trusted-server-core/src/integrations/datadome/protection.rs:406`

- 🤔 **Test coverage gap** — Behavioral tests cover the `403` challenge and `200` allow paths well, but not the `301/302` redirect-challenge path nor `x-datadome-request-headers` upstream enrichment routed through a real `classify_protection_response` result. Adding both would close the gap and lock in the redirect/`Location` behavior.
- 📌 **No `wasm32-wasip1` build gate in CI** — This PR adds wasm-sensitive runtime dependencies (`chrono::Utc::now()`, `std::time::Instant`, `std::env::var`, `urlencoding`) on the request hot path. I verified the `wasm32-wasip1` release build compiles locally, but nothing in CI guards it. Worth a follow-up issue to add a wasm build job. (Out of scope for this PR.)

## CI Status

- cargo fmt: PASS (required)
- cargo test: PASS (required)
- vitest: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS
- integration tests: PASS

PR #773, opened 2026-06-17
Add Osano consent mirror integration

### LINKED ISSUE #772: Add Osano CMP consent mirror integration for EC gating
## User story

As a publisher using Osano as its consent management platform, I want Trusted Server to read the visitor's Osano consent choices at the edge so that Edge Cookie generation, withdrawal, and auction identity behavior honor the user's consent state.

## Problem

Trusted Server's consent pipeline currently reads request-visible standard consent signals:

- `euconsent-v2`
- `__gpp`
- `__gpp_sid`
- `us_privacy`
- `Sec-GPC`

Osano exposes consent in the browser through `window.Osano.cm` and the IAB APIs (`__tcfapi`, `__uspapi`, `__gpp`), but some Osano deployments persist consent primarily in Osano-managed browser storage and do not automatically write the standard cookies Trusted Server reads on the next request.

That creates a transport gap: the browser has a valid CMP decision, but the edge request may not include a consent cookie. In regulated jurisdictions, Trusted Server then fails closed and skips EC creation even after the user has made an allowed choice. Conversely, explicit opt-out/withdrawal choices need to be transported to the edge so an existing EC cookie can be expired and tombstoned.

## Proposed solution

Build a JS-only Osano integration modeled after the existing Sourcepoint consent mirror:

1. Initialize when `window.Osano.cm` is available.
2. Listen for Osano consent lifecycle events, including saved/new consent and returning-consent initialization.
3. Read IAB-compatible signals from Osano's browser APIs:
   - `__uspapi('getUSPData', 1, ...)` for US Privacy string.
   - `__gpp('ping', ...)` for GPP string and applicable sections when ready.
   - `__tcfapi('getTCData', 2, ...)` for GDPR/TCF consent when applicable.
4. Mirror those values into first-party standard cookies for the next request:
   - `us_privacy=<uspString>`
   - `__gpp=<gppString>`
   - `__gpp_sid=<comma-separated applicable section IDs>`
   - `euconsent-v2=<tcString>` when a TCF TC string is present
5. Use a Trusted Server marker cookie, similar to `_ts_gpp_src=sp`, so the Osano mirror only updates or clears cookies it owns and does not clobber another CMP's values.
6. Keep the mirror bounded and fail-safe: if Osano or an IAB API is unavailable/not ready, do not fabricate consent; retry briefly and then wait for future Osano events/focus/visibility refresh.

## Acceptance criteria

- A new Osano JS integration mirrors Osano consent to standard cookies readable by Trusted Server.
- US no-opt-out state results in `us_privacy` indicating no sale opt-out, allowing EC generation on subsequent eligible requests.
- US opt-out state results in `us_privacy` indicating sale opt-out, blocking EC generation and allowing existing EC withdrawal handling to run.
- GDPR/TCF state mirrors `euconsent-v2` when a TC string is available.
- GPP state mirrors `__gpp` and `__gpp_sid` only when the GPP API reports a ready signal with a non-empty string.
- The integration preserves consent cookies written by another CMP unless a Trusted Server Osano marker indicates ownership.
- Unit tests cover accept, deny/opt-out, unavailable API, not-ready GPP, marker ownership, and clearing stale Osano-owned cookies.
- No real publisher domains, customer names, or production config values are added to tests, docs, or examples.

## Affected area

- Integrations (prebid, lockr, permutive, etc.)
- Core (Edge Cookies, GDPR)
- JS build pipeline

ChristianPavilonis:
## Summary

- Adds an explicitly-enabled Osano consent mirror that copies Osano IAB API output into the standard consent cookies Trusted Server already reads.
- Refactors CMP mirror module inclusion so Sourcepoint and Osano are loaded through normal integration configuration instead of unconditional JS inclusion.
- Documents the implemented design and adds Rust/Vitest coverage for registration, module selection, ownership, retries, and failure handling.

## Changes

| File | Change |
| ---- | ------ |
| `crates/js/lib/src/integrations/osano/index.ts` | New Osano browser consent mirror for `us_privacy`, `__gpp`, `__gpp_sid`, and `euconsent-v2` with `_ts_consent_src=osano` ownership. |
| `crates/js/lib/test/integrations/osano/index.test.ts` | Vitest coverage for mirroring behavior, ownership preservation, stale-cookie clearing, failures, timeouts, retries, and initialization. |
| `crates/trusted-server-core/src/integrations/osano.rs` | Minimal enabled/disabled integration registration for the JS-only Osano module. |
| `crates/trusted-server-core/src/integrations/mod.rs` | Adds Osano to the integration builder list. |
| `crates/trusted-server-core/src/integrations/registry.rs` | Tracks enabled integration IDs so opt-in JS-only registrations can be included; removes Sourcepoint from unconditional JS inclusion. |
| `trusted-server.toml` | Adds disabled-by-default `[integrations.osano]` config block. |
| `docs/superpowers/specs/2026-06-17-osano-consent-mirror-design.md` | Captures the implemented design. |

## Closes

Closes #772

## Test plan

- [x] Other: `cd crates/js/lib && node build-all.mjs`
- [x] Other: `cd docs && npx prettier --check .`

## Checklist

- [x] Uses project logging helpers/macros (not `println!`)
- [x] New code has tests


crates/js/lib/src/integrations/osano/index.ts
```diff
@@ -0,0 +1,506 @@
...
+ *
+ * Returns `true` when any cookie was written or cleared, `false` otherwise.
+ */
+export async function mirrorOsanoConsent(): Promise<boolean> {
+  if (typeof document === 'undefined') return false;
+
+  const win = getWindow();
+  if (!win) return false;
+
+  const plan = await buildMirrorPlan(win);
```
aram356, line 396: 🔧 **wrench** — Overlapping mirror attempts can apply stale consent.

`mirrorOsanoConsent()` awaits up to three async IAB API reads, each with a 500ms timeout, but nothing prevents an older in-flight attempt from applying after a newer event-driven attempt. A delayed old `__uspapi`/`__tcfapi` callback can overwrite a newer opt-out/accept state.

**Fix**: add a monotonically increasing mirror generation and skip `applyMirrorPlan()` unless the completed attempt is still current. Please add a test with two overlapping calls resolving out of order.

crates/js/lib/src/integrations/osano/index.ts
```diff
@@ -0,0 +1,506 @@
...
+            writes.push({ name: GPP_SID_COOKIE_NAME, value: applicableSections.join(',') });
+          } else {
+            clears.push(GPP_SID_COOKIE_NAME);
+          }
+
+          done(signalResult(writes, clears));
+          return;
+        }
+
+        done(signalResult([], [GPP_COOKIE_NAME, GPP_SID_COOKIE_NAME]));
```
aram356: 🔧 **wrench** — GPP empty-value clearing bypasses the Osano readiness guard.

USP and TCF empty values use `emptyAfterOsanoReadyResult()`, but GPP ready-with-empty immediately clears `__gpp` and `__gpp_sid`. That can clear stale Osano-owned GPP cookies during startup before an Osano readiness event, which conflicts with the PR's preserve-until-ready behavior.

**Fix**: gate the empty-GPP clear on `osanoReadyForClears` too, and add the GPP equivalent of the stale-cookie preservation test.

crates/trusted-server-core/src/integrations/registry.rs
```diff
@@ -1131,15 +1142,14 @@ impl IntegrationRegistry {
     pub fn js_module_ids(&self) -> Vec<&'static str> {
         // Rust-only integrations with no corresponding JS module
         const JS_EXCLUDED: &[&str] = &["nextjs", "aps", "adserver_mock"];
```
aram356: ♻️ **refactor** — Derive JS module availability instead of hardcoding integration IDs.

`JS_EXCLUDED` is another integration-ID list that has to stay in sync with registration and the TSJS build output. We can avoid that drift by asking the generated JS bundle registry whether a module exists for the enabled integration ID, rather than hardcoding Rust-only integrations here.

**Suggested direction**:
```rust
for id in &self.inner.enabled_integration_ids {
    if trusted_server_js::module_bundle(id).is_some() && !ids.contains(id) {
        ids.push(*id);
    }
}
```

`creative` may still need an explicit always-on path unless it gets a registration, but this removes the brittle `nextjs` / `aps` / `adserver_mock` exclusion list.

crates/trusted-server-core/src/integrations/registry.rs
```diff
@@ -1131,15 +1142,14 @@ impl IntegrationRegistry {
...
-        // JS-only modules always included (no Rust-side registration).
-        // Sourcepoint's JS guards cookie clearing with a Sourcepoint-owned marker.
-        const JS_ALWAYS: &[&str] = &["creative", "sourcepoint"];
+        // Core JS-only modules that do not have a Rust-side registration.
+        const JS_ALWAYS: &[&str] = &["creative"];
```
aram356, line 1105: 🤔 **thinking** — Sourcepoint mirror is now opt-in, which is a behavior change.

Removing Sourcepoint from `JS_ALWAYS` means existing deployments that relied on the unconditional Sourcepoint GPP mirror will stop shipping that JS unless they explicitly set `[integrations.sourcepoint].enabled = true`. The PR says this is intentional, but I would recommend a migration note in public docs or release notes.

aram356 (changes requested):
## Summary

The Osano mirror is a good direction and the PR has solid CI coverage, but I found two consent-state correctness issues that should be fixed before merge. I also included a small registry refactor request and two non-blocking follow-ups around migration/docs.

## Blocking

### 🔧 wrench

- **Overlapping mirror attempts can apply stale consent**: `mirrorOsanoConsent()` awaits async IAB API reads without guarding against older attempts completing after newer consent events, so a delayed callback can overwrite the current consent state. (`crates/js/lib/src/integrations/osano/index.ts:394`)
- **GPP empty-value clearing bypasses the Osano readiness guard**: USP and TCF empty values wait for Osano readiness before clearing stale cookies, but GPP ready-with-empty clears immediately. (`crates/js/lib/src/integrations/osano/index.ts:290`)

## Non-blocking

### ♻️ refactor

- **Derive JS module availability instead of hardcoding integration IDs**: `JS_EXCLUDED` has to stay in sync with Rust registrations and generated TSJS modules. Prefer checking whether `trusted_server_js::module_bundle(id)` exists for each enabled integration. (`crates/trusted-server-core/src/integrations/registry.rs:1144`)

### 🤔 thinking

- **Sourcepoint mirror is now opt-in, which is a behavior change**: existing deployments that relied on the unconditional Sourcepoint GPP mirror will stop shipping that JS unless they explicitly enable `[integrations.sourcepoint]`. This likely deserves a migration note. (`crates/trusted-server-core/src/integrations/registry.rs:1146`)

### 🌱 seedling

- **Public integration docs should mention Osano enablement**: the design spec is useful, but operators will likely look in public guide/configuration docs for `[integrations.osano] enabled = true` and the first-request/next-request limitation.

## CI Status

- Analyze (actions): PASS
- Analyze (rust): PASS
- CodeQL: PASS
- cargo fmt: PASS
- cargo test: PASS
- format-docs: PASS
- format-typescript: PASS
- integration tests: PASS
- vitest: PASS

crates/js/lib/src/integrations/osano/index.ts
ChristianPavilonis: Fixed in e62f8e08 by adding a monotonically increasing mirror generation guard. Older overlapping mirror attempts now return false without applying their plan when a newer attempt has started. Added the out-of-order regression test `does not let older overlapping mirror attempts overwrite newer consent`.

crates/js/lib/src/integrations/osano/index.ts
ChristianPavilonis: Fixed in e62f8e08 by routing ready-but-empty GPP through the same Osano readiness gate as USP/TCF. Added GPP stale-preservation and post-readiness clearing tests for `__gpp` and `__gpp_sid`.

crates/trusted-server-core/src/integrations/registry.rs
ChristianPavilonis: Fixed in e62f8e08 by removing the hardcoded `JS_EXCLUDED` list and deriving enabled integration JS availability with `trusted_server_js::module_bundle(id).is_some()`. Added coverage showing enabled Rust-only `nextjs` remains registered but is excluded from JS module IDs.

crates/trusted-server-core/src/integrations/registry.rs
ChristianPavilonis: Documented in e62f8e08: added a Sourcepoint migration note in the public integration guide and a breaking-change note in `CHANGELOG.md` explaining that `[integrations.sourcepoint].enabled = true` is now required for the Sourcepoint browser module.

ChristianPavilonis:
Also addressed the non-inline docs feedback in e62f8e08 by adding public Osano enablement docs, linking the Osano guide from the sidebar and integrations overview, and documenting the first-request/subsequent-request limitation.

crates/js/lib/src/integrations/osano/index.ts
```diff
@@ -0,0 +1,514 @@
...
+function readCookie(name: string): string | undefined {
+  if (typeof document === 'undefined') return undefined;
+
+  const prefix = `${name}=`;
+  const cookie = document.cookie.split('; ').find((entry) => entry.startsWith(prefix));
+  return cookie?.slice(prefix.length);
+}
+
+function writeCookie(name: string, value: string): void {
+  document.cookie = `${name}=${value}; Path=/; Secure; SameSite=Lax`;
```
aram356, line 121: 🌱 **seedling** — Mirrored cookies are written without `Max-Age`/`Expires`, so they are session cookies and must be re-mirrored on each new browser session. A persistent cookie bounded to the consent's validity would partially close the documented first-request gap for *returning* visitors (the server could see consent on the very first request of a return visit, before this browser module runs). Not required for v1 — just noting for a future iteration.

crates/js/lib/src/integrations/osano/index.ts
```diff
@@ -0,0 +1,514 @@
...
+
+function hasAnyTargetCookie(): boolean {
+  return TARGET_COOKIE_NAMES.some((name) => readCookie(name) !== undefined);
+}
+
+function ownsConsentCookies(): boolean {
+  return readCookie(MARKER_COOKIE_NAME) === MARKER_COOKIE_VALUE;
+}
+
+function canWriteConsentCookies(): boolean {
```
aram356, line 136: 📝 **note** — Ownership is first-writer-wins: if both the Osano and Sourcepoint mirrors are enabled simultaneously, whichever sets `_ts_consent_src` first owns the cookies and the other backs off indefinitely. This is correct behavior (the mirrors don't fight), but it would be worth a one-line note in the docs that operators should run a single CMP mirror at a time.

aram356 (approved):
## Summary

Re-review. Both prior 🔧 blockers and all three non-blocking items from the earlier review are resolved, each with regression coverage, and CI is fully green.

- **Stale overlapping consent** — fixed via the `mirrorGeneration` guard; only the latest in-flight attempt applies (`mirrorOsanoConsent`). Covered by a dedicated overlap test.
- **GPP empty-value clearing** — now routes through the same Osano-readiness guard as USP/TCF (`emptyAfterOsanoReadyResult`). Covered by preserve-until-ready and clear-when-ready tests.
- **JS module derivation** — `js_module_ids()` now derives availability from `trusted_server_js::module_bundle(id)` over enabled integrations instead of a hardcoded denylist. Verified behavior-preserving against `main` (only the intended Sourcepoint change).
- **Sourcepoint opt-in** — documented as a Breaking CHANGELOG entry plus a migration note in the Sourcepoint guide.
- **Public docs** — added the Osano guide, configuration section, sidebar/overview links, and the first-request limitation.

I independently traced the cookie ownership/clearing state machine, the multi-attempt generation race, and the denylist→allowlist refactor — no blocking issues. Approving. Two optional non-blocking notes are left inline.

## CI Status

- cargo fmt: PASS
- cargo clippy / Analyze (rust): PASS
- cargo test: PASS
- vitest: PASS
- browser + integration tests: PASS

prk-Jr (approved):

Issue #269, opened 2026-02-09
As a buyer, I want the first bid request to include IAB cohorts, signals, and synthetic ID


aram356:
## Story

**As a** buyer,
**I want** the first bid request to include IAB cohorts, publisher signals, and the TS synthetic ID,
**so that** even before ID sync, I have meaningful data to inform my bidding and can initiate the sync process.

**Parent Epic:** #265

## Acceptance Criteria

- [ ] Every bid request includes the TS synthetic ID (already implemented via `user.id`)
- [ ] If IAB audience taxonomy cohorts are available, they are included in the request
- [ ] If publisher-defined signals are available, they are included in the request
- [ ] This data is present regardless of whether a buyer ID mapping exists

## Open Questions

- What is the source of IAB cohorts? (Publisher-provided? TS-computed? Third-party integration?)
- What signal format? (IAB Seller Defined Audiences? Custom key-values?)
- Where in the OpenRTB request should cohorts/signals go? (`user.data`? `site.content.data`?)

PR #783, opened 2026-06-18
Harden edgezero #269 runtime config-store load (HTTP layer)

prk-Jr:
## Summary

- **HTTP-layer (runtime) half of edgezero `stackpop/edgezero#269`.** Stacked on `feature/ts-cli-next` (which carries the [#269](https://github.com/stackpop/edgezero/pull/269) repin, `Body` fixes, the Fastly adapter migration, and config-store-backed `Settings` load). **Draft** — base is unmerged; retarget to `main` once `ts-cli-next` lands.
- **Core change (spec "option Y"):** a config-store **read failure** (store unseeded, transient backend, or a listed key missing) now maps to a new `TrustedServerError::ConfigStoreUnavailable` → **503**, while a **reconstruct/verify failure** (`settings_from_config_entries`: hash mismatch / unparseable) stays **500**. One new error variant; no platform-layer change.
- **Security-aware:** the actionable hint (`run \`ts config push\``) goes to **server logs** (error chain); the public 503 **body stays generic** by design.
- Includes the design spec + implementation plan, and the upstream finding/plan docs for context.

**Base note:** diff is against `feature/ts-cli-next`, so it shows only this branch's work. Against `main` it would include all of `ts-cli-next`.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/error.rs` | New `ConfigStoreUnavailable { store_name, message }` variant → 503 (+ exhaustiveness guard, unit test) |
| `crates/trusted-server-core/src/settings_data.rs` | `read_config_entry` read failures → `ConfigStoreUnavailable`; tests (unseeded→503, malformed-hash→500, missing-listed-key→503, hint-in-chain) |
| `crates/trusted-server-adapter-fastly/src/error.rs` | Test: `ConfigStoreUnavailable` renders 503 to client via `to_error_response` |
| `crates/integration-tests/Cargo.lock` | Reconciled to edgezero [#269](https://github.com/stackpop/edgezero/pull/269) |
| `docs/superpowers/specs|plans/*` | Runtime design spec + plan (+ cherry-picked [#269](https://github.com/stackpop/edgezero/pull/269) finding/plan) |

## Closes

n/a — issue linking skipped by request.

## Test plan

- [x] `cargo test --workspace` (core 1376 / adapter 39, 0 fail)
- [x] `cargo build -p trusted-server-adapter-fastly --release --target wasm32-wasip1`
- [x] integration-tests workspace builds; docs format clean

## Checklist

- [x] Changes follow [CLAUDE.md](/CLAUDE.md) conventions (error-stack, `log`, colocated tests)
- [x] No secrets committed
- [x] New code has tests


ChristianPavilonis (commented):
Automated review:

## Review Summary

Reviewed PR #783 against `feature/ts-cli-audit`, focusing on the runtime config-store load path, `error-stack` context/status behavior, Fastly/EdgeZero HTTP error rendering, public error-body leakage, and the added regression tests. I did not find any blocking correctness, security, data-loss, authorization, or severe compatibility issues in the changed code.

## Findings

No blocking findings.

## CI / Existing Reviews

- GitHub checks currently show `prepare integration artifacts` failing, with downstream jobs skipped. The failure is dependency-lock parity drift in `crates/trusted-server-integration-tests/Cargo.lock` (for example `anyhow`, `cssparser`, `html5ever`, `scraper`, `uuid`, and related transitive dependencies). This should be resolved or confirmed as inherited from the stacked/base branch before merge.
- Existing PR reviews: none returned by the GitHub API. Existing inline review comments: none.
- Local focused check passed: `cargo test -p trusted-server-core settings_data --target wasm32-wasip1`. A local Fastly adapter-focused test could not complete because this environment has `viceroy 0.16.4` while the repo pins `0.17.0`.

crates/trusted-server-core/src/settings_data.rs
```diff
@@ -1,223 +1,408 @@
...
+    store_name: &StoreName,
+    key: &str,
+) -> Result<String, Report<TrustedServerError>> {
+    config_store
+        .get(store_name, key)
+        .change_context(TrustedServerError::ConfigStoreUnavailable {
+            store_name: store_name.to_string(),
+            message: format!(
+                "unavailable or not seeded (failed to read `{key}`) — run `ts config push`"
+            ),
```
aram356: ⛏ **nitpick** — `read_config_entry` is the shared read seam for both the top-level blob key and each chunk key, so this "unavailable or not seeded" wording is slightly off for a *missing chunk* (the blob key was seeded; a referenced chunk is absent). A phrasing that covers unseeded, missing-key, and transient reads keeps it accurate for both call sites:

```suggestion
            message: format!(
                "read failed for `{key}` (unseeded, missing, or transient) — run `ts config push` to (re)seed"
            ),
```

_(compile- and test-verified in a scratch worktree: `cargo fmt --all --check`, `clippy -p trusted-server-core -D warnings`, and the `settings_data` tests all pass with this applied.)_

aram356 (changes requested):
## Summary

Scoped, well-tested runtime hardening: a config-store **read** failure now maps to `TrustedServerError::ConfigStoreUnavailable` → **503** (retryable), while **verify/parse** failures stay **500**. The real change is **4 files / +279 −13**; verified locally at head (all new core + adapter tests pass). The two blocking items concern the PR's **base/CI plumbing**, not the code itself.

> 1 of the inline comments below carries a one-click GitHub `suggestion`. The remaining findings are body-level because they concern the PR base, CI, or code outside the diff hunks and can't be auto-applied.

## Blocking

### 🔧 wrench

- **Wrong PR base inflates the diff to 66 files** — see Cross-cutting below.
- **CI `prepare integration artifacts` is failing** — see Cross-cutting below.

## Non-blocking

### ⛏ nitpick

- **Chunk-read failures reuse the top-level "not seeded" hint** — see inline at `crates/trusted-server-core/src/settings_data.rs:95`.

### 🤔 thinking / 🌱 seedling

- **503 body says "An internal error occurred"** — see Cross-cutting.
- **`with_capacity(envelope_len)` trusts an unvalidated length** — see Cross-cutting.

- 🔧 **Wrong PR base → unreviewable diff & merge risk.** The PR targets `feature/ts-cli-audit`, but the branch is stacked on `feature/ts-cli-next` (per the PR body) and is diverged from both (vs `ts-cli-audit`: 18 ahead / 25 behind). GitHub shows **66 files / +9,550 −1,772**; the actual change is **4 files / +279 −13** (the 4 topmost commits: the `ConfigStoreUnavailable` variant, the read-failure reclassification, the adapter 503 lock, and the design doc). Merging against this base risks pulling ~60 unrelated files into the target, and the design-doc header ("stacked on `feature/ts-cli-audit`") disagrees with the PR body ("stacked on `feature/ts-cli-next`"). Recommend rebasing onto the true parent's current tip (so it's a clean ancestor), then retargeting the base — or `main` once the stack lands — so the PR shows only the 4 files.

- 🔧 **CI `prepare integration artifacts` is FAILING.** Transitive-dependency parity drift (`markup5ever`, `match_token`, `scraper`, `selectors`, `uuid`, `wasm-bindgen-futures`, `web-sys`) between the workspace and `trusted-server-integration-tests`. The failure originates in the stacked base (integration-tests lock), not the 4-file 503 change, and isn't branch-protection-required here (base is a feature branch) — but it's red on the PR and CLAUDE.md treats it as a gate. It should clear once the branch sits on the correct, lock-reconciled parent (see above).

- 🌱 **`String::with_capacity(pointer.envelope_len)` (`settings_data.rs:119`) pre-allocates from an unvalidated length.** Pre-existing and outside this diff, but squarely in this PR's "harden the config-store load" theme: a corrupted pointer with a huge `envelope_len` triggers a large allocation on the **per-request** settings-load path — a potential trap/abort instead of the graceful 503/500 this PR adds. Consider bounding `envelope_len` (and cross-checking it against the sum of `chunk.len`) before pre-allocating. Follow-up, not this PR's job.

- 🤔 **503 body reads "An internal error occurred".** The generic catch-all `user_message()` describes a 500, but 503 is *retryable*. The generic body is deliberate (design §4 — no `user_message` arm added, which correctly avoids leaking internals), so this is only a suggestion: a 503-appropriate phrase (e.g. "Service temporarily unavailable") would let clients/monitoring distinguish retryable from terminal without leaking detail.

## CI Status

- integration tests: SKIPPED
- integration tests (EdgeZero entry point): SKIPPED
- browser integration tests: SKIPPED
- prepare integration artifacts: FAIL
- .github/dependabot.yml: PASS

crates/trusted-server-core/src/error.rs
```diff
@@ -123,6 +128,7 @@ impl IntoHttpResponse for TrustedServerError {
             Self::InvalidHeaderValue { .. } => StatusCode::BAD_REQUEST,
             Self::InvalidUtf8 { .. } => StatusCode::INTERNAL_SERVER_ERROR,
             Self::KvStore { .. } => StatusCode::SERVICE_UNAVAILABLE,
+            Self::ConfigStoreUnavailable { .. } => StatusCode::SERVICE_UNAVAILABLE,
```
aram356, line 131: 🤔 **thinking** — This arm maps `ConfigStoreUnavailable` to **503 Service Unavailable** (retryable), which is right. The follow-on nuance is the *body*: `user_message()` deliberately has no arm for this variant, so the client receives the generic catch-all `"An internal error occurred"`. Keeping the body generic is correct (design §4 — no internal leak), but that phrasing describes a **500**, not a retryable **503**. A 503-flavored generic phrase (e.g. `"Service temporarily unavailable"`) would let clients/monitoring distinguish *retry me* from *terminal* without leaking any detail.

Not blocking, and intentionally out of this hunk's scope — just flagging the status/body mismatch next to where the 503 is established.

crates/trusted-server-core/src/settings_data.rs
prk-Jr: Applied in ca2e77a60. Reworded to `read failed for `{key}` (unseeded, missing, or transient) — run `ts config push` to (re)seed`, so it reads correctly for both the top-level blob key and a missing chunk key. fmt + `clippy -p trusted-server-core -D warnings` + settings_data tests pass.

crates/trusted-server-core/src/error.rs
prk-Jr: Applied in ca2e77a60. Added a `user_message()` arm for the retryable 503 variants (`ConfigStoreUnavailable | KvStore`) returning "Service temporarily unavailable" — retry-flavored, still leaks no internal detail (full detail stays in the logged error chain). Updated `config_store_unavailable_maps_to_503`, dropped `KvStore` from the generic-body case list, and added `retryable_503_variants_return_service_unavailable_body`.

crates/trusted-server-core/src/settings_data.rs
```diff
@@ -28,10 +28,36 @@ struct FastlyChunkRef {
...
+    let config_key = default_config_key();
+    get_settings_from_config_store(services.config_store(), &store_name, &config_key)
+}
+
 /// Returns the default `EdgeZero` app-config store name.
 #[must_use]
 pub fn default_config_store_name() -> StoreName {
-    StoreName::from(EnvConfig::from_env().store_name("config", DEFAULT_CONFIG_STORE_ID))
+    StoreName::from(
+        std::env::var("EDGEZERO__STORES__CONFIG__APP_CONFIG__NAME")
```
ChristianPavilonis: Automated review: **Store-name env resolution no longer honors EdgeZero fallback semantics.**

This used to go through `EnvConfig::store_name("config", "app_config")`, which intentionally falls back to `app_config` when `EDGEZERO__STORES__CONFIG__APP_CONFIG__NAME` is unset, blank/whitespace-only, or contains control characters. Reading the env var directly preserves blank/control values as the runtime `StoreName`, so a deployment with an accidentally empty override now tries to open/read `""` instead of falling back to the logical store and will surface as a config-store 503. `default_config_key()` still uses `EnvConfig`, so name/key handling is also inconsistent.

Suggested fix: restore the typed helper, e.g. `StoreName::from(EnvConfig::from_env().store_name("config", DEFAULT_CONFIG_STORE_ID))`, and add a regression test for blank/whitespace values.

crates/trusted-server-integration-tests/tests/common/ec.rs
```diff
@@ -271,6 +271,37 @@ fn mappings_to_json(mappings: &[BatchMapping]) -> Vec<Value> {
...
+        .expect("should build EdgeZero canary client");
+    let response = client
+        .request(
+            reqwest::Method::OPTIONS,
+            format!("{base_url}/_ts/api/v1/batch-sync"),
+        )
+        .send()
+        .change_context(TestError::HttpRequest)
+        .attach("OPTIONS /_ts/api/v1/batch-sync (EdgeZero entry-point probe)")?;
+    if response.status().as_u16() != 405 {
```
ChristianPavilonis: Automated review: **The EdgeZero integration job can pass without proving the EdgeZero entry point was used.**

The workflow opts into `EXPECT_EDGEZERO_ENTRY_POINT`, but this probe only logs when the response is not 405 and then returns `Ok(())`. The surrounding comments say the EC lifecycle scenarios pass on either the legacy or EdgeZero path, so a broken config-store flag, failed config-store open, or silent fallback to `legacy_main` would still leave the EdgeZero CI job green.

Suggested fix: make the job assert a stable positive EdgeZero-path signal instead of a non-fatal unsupported-method status (for example a test-only response/header/route that only exists after EdgeZero dispatch, or a startup/config condition that fails the job when `edgezero_enabled` cannot be honored).

docs/superpowers/specs/2026-06-27-edgezero-http-config-503-design.md
```diff
@@ -0,0 +1,78 @@
...
+## 4. Mechanism (one new variant)
+
+`TrustedServerError::ConfigStoreUnavailable { store_name, message }` →
+`StatusCode::SERVICE_UNAVAILABLE` (precedent: the existing `KvStore` 503 arm).
+Only `read_config_entry`'s `change_context` target changes from `Configuration`
+to the new variant; all verify/parse paths are untouched. No `PlatformError`
+or `PlatformConfigStore` change.
+
+**Security:** the actionable hint rides the error chain (`Display`) to the
+**server log** only. The public 503 body stays the generic
```
ChristianPavilonis: Automated review: **The design doc no longer matches the implementation.**

The spec says the public 503 body stays on the generic `user_message()` catch-all and that no `user_message()` arm is added, but the implementation now adds a `ConfigStoreUnavailable | KvStore` arm returning `"Service temporarily unavailable"`. The implemented behavior is reasonable and still does not leak internals; the doc should be updated so future reviewers do not treat the current code as violating the design.

ChristianPavilonis (commented):
Automated review:

## Review Summary

Reviewed PR #783 against `main`, focusing on the config-store settings load path, HTTP status/body mapping, Fastly EdgeZero routing behavior, CI coverage, and the previously raised review threads. The core 503 classification is generally well covered and I did not find a blocking correctness, security, data-loss, authorization, or severe compatibility issue. I did find a few non-blocking issues worth addressing before merge, especially an env-config compatibility regression in store-name resolution.

## Findings

### P0 / Blockers

None.

### P1 / High

1. **Store-name env resolution no longer honors EdgeZero fallback semantics** — see inline comment on `crates/trusted-server-core/src/settings_data.rs`.

### P2 / Medium

1. **The JA4 debug preflight still returns 500 for config-store read failures** — `crates/trusted-server-adapter-fastly/src/main.rs:401`. This line is outside the PR diff, so I could not leave it inline, but the new `ConfigStoreUnavailable` behavior is incomplete on this HTTP path: `GET /_ts/debug/ja4` calls `load_settings_from_config_store()` before app construction and hard-codes `500 Internal Server Error` on every load failure instead of using `e.current_context().status_code()` / `user_message()`. In an unseeded or transient config-store case, normal EdgeZero startup responses now return 503 while this debug endpoint still returns 500. Suggested fix: map the `Report<TrustedServerError>` through the same `IntoHttpResponse` methods used by the startup/error response path.
2. **The EdgeZero integration job can pass without proving the EdgeZero entry point was used** — see inline comment on `crates/trusted-server-integration-tests/tests/common/ec.rs`.

### P3 / Low

1. **The implemented 503 body behavior contradicts the new design doc** — see inline comment on `docs/superpowers/specs/2026-06-27-edgezero-http-config-503-design.md`.
2. **Duplicate CI/local verification steps were added** — the diff adds exact duplicate host CLI clippy/test steps in `.github/workflows/format.yml` and `.github/workflows/test.yml`, duplicate Viceroy config generation in `.github/workflows/integration-tests.yml` / `scripts/integration-tests.sh`, and duplicate CLI-test docs in `CLAUDE.md`. This is not correctness-blocking, but it increases CI time and maintenance noise.

## CI / Existing Reviews

- GitHub checks are currently passing.
- Existing review context: earlier automated reviews/comments raised the chunk-read hint and 503 body wording; the latest commits appear to have addressed those specific comments.
- Local focused checks run during this review: `cargo test -p trusted-server-core settings_data --target x86_64-unknown-linux-gnu`, `cargo test -p trusted-server-core config_store_unavailable --target x86_64-unknown-linux-gnu`, and `git diff --check $(git merge-base HEAD origin/main)...HEAD` all passed.

ChristianPavilonis (commented):
Automated review:

## Review Summary

Reviewed PR #783 at head `03fc21c00528544717b398d481556842e3757026` against `main`, focusing on the config-store load/error-classification changes, Fastly/Axum/Spin HTTP response behavior, Spin runtime config-store integration, manifests, integration workflow changes, and prior review feedback. Overall risk looks moderate: CI is green and the main 503-vs-500 split is well covered, but I found one non-blocking correctness gap in the new Spin KV-backed app-config path where corrupt stored bytes are reported as retryable store unavailability.

## Findings

### P0 / Blockers

None.

### P1 / High

None.

### P2 / Medium

1. **Spin app-config values that are present but not UTF-8 are classified as retryable 503s** — see inline comment on `crates/trusted-server-adapter-spin/src/platform.rs`.

### P3 / Low

None.

## CI / Existing Reviews

- GitHub checks are currently passing.
- Existing review context: the previously raised store-name fallback, JA4 preflight status mapping, design-doc body mismatch, duplicate CI-step drift, and 503 body wording comments appear addressed by later commits. The earlier EdgeZero-entry-point integration probe comment remains a non-blocking coverage concern, but I am not duplicating it here.
- I did not run additional local test commands beyond inspecting the diff and CI results.

crates/trusted-server-adapter-spin/src/platform.rs
```diff
@@ -61,6 +61,57 @@ impl PlatformConfigStore for NoopConfigStore {
...
+                Report::new(PlatformError::ConfigStore).attach(format!(
+                    "Spin key-value lookup for `{key}` in `{label}` failed: {e}"
+                ))
+            })?
+            .ok_or_else(|| {
+                Report::new(PlatformError::ConfigStore).attach(format!(
+                    "key `{key}` not found in Spin key-value store `{label}`"
+                ))
+            })?;
+        String::from_utf8(bytes).map_err(|e| {
```
ChristianPavilonis: Automated review: **Present-but-corrupt Spin app-config bytes are reported as retryable store unavailability.**

This `String::from_utf8` failure means the `app_config` key was successfully opened and read, but the stored value is corrupt/not a valid text config blob. Because the error is returned as `PlatformError::ConfigStore`, `read_config_entry()` wraps it in `TrustedServerError::ConfigStoreUnavailable`, so Spin returns the new retryable 503 path instead of the intended 500-class "read succeeded but reconstruct/verify failed" path. In practice, an operator who accidentally seeds binary/corrupt bytes gets misleading retryable behavior and clients may keep retrying a terminal bad config.

Suggested fix: keep open/get/missing-key failures on the 503 path, but make a present value that cannot be decoded feed the verification/configuration path (for example, have this adapter convert invalid UTF-8 into a `TrustedServerError::Configuration`, or decode lossily/otherwise pass bytes forward so `settings_from_config_blob` fails as a 500), and add a Spin-specific regression test for non-UTF-8 bytes.

aram356:
Closing in place of https://github.com/IABTechLab/trusted-server/pull/839

PR #787, opened 2026-06-19
Honor publisher Host override and bypass IO for SVG assets

### LINKED ISSUE #786: Honor publisher origin Host header override in Fastly backends
## Problem

A test deployment that proxies to a publisher origin returns `404 Not Found` even when `publisher.origin_host_header_override` is configured.

Direct origin checks show the upstream origin serves the page when the required Host header is present, but returns the same 404 shape when the backend host is used as Host. This indicates the configured override is not being applied at the Fastly backend layer.

## Root cause

The publisher proxy computes and inserts the configured Host header on the outbound request, but the platform backend abstraction does not carry a host-header override into Fastly dynamic backend registration.

`BackendConfig` already supports `host_header_override` and uses it to call Fastly `override_host`, but the newer `PlatformBackendSpec` path does not expose or pass this value.

## Expected behavior

When `publisher.origin_host_header_override` is configured:

- publisher-origin dynamic backend registration uses the override Host value
- Fastly sends the configured Host header to the origin
- origins that route by Host return the expected content instead of backend-host 404s

## Acceptance criteria

- Add host-header override support to `PlatformBackendSpec`.
- Pass `publisher.origin_host_header_override` when registering the publisher origin backend.
- Wire the override into the Fastly platform backend implementation.
- Keep non-publisher backend registrations unchanged by using no override.
- Add/adjust tests covering override-aware backend naming.

### LINKED ISSUE #789: Bypass Fastly Image Optimizer for SVG asset routes
## Problem

Asset proxy routes can attach Fastly Image Optimizer metadata whenever an asset route has Image Optimizer enabled. SVG and SVGZ assets should not be sent through Fastly Image Optimizer, but the current routing path can still attach IO metadata for SVG requests.

This is especially visible when a route rewrite changes the origin target path: the rewritten target may hide the original `.svg`/`.svgz` extension from the later IO decision, causing SVG assets to be processed by IO instead of passed through directly.

## Expected behavior

When an asset proxy request is for an SVG or SVGZ asset:

- do not attach Fastly Image Optimizer metadata
- proxy the asset directly to the origin
- preserve normal asset proxy behavior for non-SVG assets
- treat both the incoming request path and the rewritten target path as SVG bypass signals

## Acceptance criteria

- Detect `.svg` and `.svgz` asset paths case-insensitively.
- Bypass Image Optimizer when either the incoming path or rewritten target path is SVG/SVGZ.
- Keep Image Optimizer enabled for supported non-SVG asset types.
- Add tests covering direct SVG paths, rewritten SVG paths, and non-SVG paths.

ChristianPavilonis:
## Summary

- Thread publisher origin Host header overrides through the platform backend abstraction.
- Ensure Fastly dynamic backends call `override_host` when `publisher.origin_host_header_override` is configured.
- Keep integration, asset, and generic proxy backend registrations unchanged by using no override.
- Bypass Fastly Image Optimizer for SVG asset requests when either the incoming path or rewritten target path ends in `.svg`/`.svgz`.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/platform/types.rs` | Added optional `host_header_override` to `PlatformBackendSpec`. |
| `crates/trusted-server-core/src/publisher.rs` | Passes `publisher.origin_host_header_override` when registering the publisher origin backend. |
| `crates/trusted-server-adapter-fastly/src/platform.rs` | Wires `PlatformBackendSpec::host_header_override` into `BackendConfig::host_header_override`; adds override-aware backend name test. |
| `crates/trusted-server-core/src/integrations/mod.rs` | Sets no host override for integration backends. |
| `crates/trusted-server-core/src/integrations/datadome/protection.rs` | Sets no host override for DataDome protection backend. |
| `crates/trusted-server-core/src/proxy.rs` | Sets no host override for asset and signed proxy backends; skips Image Optimizer for incoming SVG paths even when route rewrites remove the extension. |
| `crates/trusted-server-core/src/platform/test_support.rs` | Updates test backend spec construction for the new field. |

## Closes

Closes #786
Closes #789

## Test plan

- [x] Manual testing: verified deployed SVG logo loads after bypassing Image Optimizer
- [ ] Other: Not run; Rust-only change.

## Checklist

- [x] Uses logging macros (not `println!`)
- [x] New code has tests

### DIFF
--- a/crates/trusted-server-adapter-fastly/src/platform.rs
+++ b/crates/trusted-server-adapter-fastly/src/platform.rs
@@ -309,6 +309,11 @@ fn edge_request_to_fastly(
     let (parts, body) = request.into_parts();
     let mut fastly_req = fastly::Request::new(parts.method, parts.uri.to_string());
     for (name, value) in parts.headers.iter() {
+        // `fastly::Request::new` derives a Host header from the request URI, so
+        // appending the edge request's own Host would leave a duplicate. Replace
+        // it instead to keep the in-memory request well-formed. The Host actually
+        // sent on the wire is still governed by the backend's `override_host`
+        // (see `BackendConfig::ensure`), which forces the value regardless.
         if name == edgezero_core::http::header::HOST {
             fastly_req.set_header(name.as_str(), value.as_bytes());
         } else {
--- a/crates/trusted-server-adapter-fastly/src/platform.rs
+++ b/crates/trusted-server-adapter-fastly/src/platform.rs
@@ -157,6 +157,7 @@ pub struct FastlyPlatformBackend;
 fn backend_config_from_spec(spec: &PlatformBackendSpec) -> BackendConfig<'_> {
     BackendConfig::new(&spec.scheme, &spec.host)
         .port(spec.port)
+        .host_header_override(spec.host_header_override.as_deref())
         .certificate_check(spec.certificate_check)
         .first_byte_timeout(spec.first_byte_timeout)
 }
@@ -619,6 +620,7 @@ mod tests {
             scheme: "https".to_string(),
             host: "origin.example.com".to_string(),
             port: None,
+            host_header_override: None,
             certificate_check: true,
             first_byte_timeout: Duration::from_secs(15),
         };
@@ -634,12 +636,35 @@ mod tests {
     }
 
     #[test]
+    fn predict_name_includes_host_header_override_suffix() {
+        let backend = FastlyPlatformBackend;
+        let spec = PlatformBackendSpec {
+            scheme: "https".to_string(),
+            host: "origin.example.com".to_string(),
+            port: None,
+            host_header_override: Some("www.example.com".to_string()),
+            certificate_check: true,
+            first_byte_timeout: Duration::from_secs(15),
+        };
+
+        let name = backend
+            .predict_name(&spec)
+            .expect("should compute backend name for host header override");
+
+        assert_eq!(
+            name, "backend_https_origin_example_com_443_oh_www_example_com_t15000",
+            "should match BackendConfig naming convention with host header override"
+        );
+    }
+
+    #[test]
     fn predict_name_includes_nocert_suffix_when_cert_check_disabled() {
         let backend = FastlyPlatformBackend;
         let spec = PlatformBackendSpec {
             scheme: "https".to_string(),
             host: "origin.example.com".to_string(),
             port: None,
+            host_header_override: None,
             certificate_check: false,
             first_byte_timeout: Duration::from_secs(15),
         };
@@ -661,6 +686,7 @@ mod tests {
             scheme: "https".to_string(),
             host: String::new(),
             port: None,
+            host_header_override: None,
             certificate_check: true,
             first_byte_timeout: Duration::from_secs(15),
         };
@@ -677,6 +703,7 @@ mod tests {
             scheme: "https".to_string(),
             host: "origin.example.com".to_string(),
             port: None,
+            host_header_override: None,
             certificate_check: true,
             first_byte_timeout: Duration::from_millis(2000),
         };
--- a/crates/trusted-server-adapter-fastly/src/platform.rs
+++ b/crates/trusted-server-adapter-fastly/src/platform.rs
@@ -309,7 +309,11 @@ fn edge_request_to_fastly(
     let (parts, body) = request.into_parts();
     let mut fastly_req = fastly::Request::new(parts.method, parts.uri.to_string());
     for (name, value) in parts.headers.iter() {
-        fastly_req.append_header(name.as_str(), value.as_bytes());
+        if name == edgezero_core::http::header::HOST {
+            fastly_req.set_header(name.as_str(), value.as_bytes());
+        } else {
+            fastly_req.append_header(name.as_str(), value.as_bytes());
+        }
     }
     match body {
         edgezero_core::body::Body::Once(bytes) => {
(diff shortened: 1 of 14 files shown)


ChristianPavilonis:
Follow-up after testing the deployed branch: the first patch deployed as `x-ts-version: 281` still produced the origin 404. I found an additional Host handling issue in the Fastly request conversion path: `edge_request_to_fastly()` constructed a Fastly request from the absolute origin URL, then appended forwarded headers. If Fastly had already populated Host/authority from the URL, appending the configured Host could leave the backend host in effect. The latest commit replaces the URL-derived Host header instead of appending a duplicate and adds a targeted test for that conversion behavior. Please redeploy latest branch commit `65a475c9`.

crates/trusted-server-adapter-fastly/src/platform.rs
```diff
@@ -308,7 +309,11 @@ fn edge_request_to_fastly(
     let (parts, body) = request.into_parts();
     let mut fastly_req = fastly::Request::new(parts.method, parts.uri.to_string());
     for (name, value) in parts.headers.iter() {
-        fastly_req.append_header(name.as_str(), value.as_bytes());
+        if name == edgezero_core::http::header::HOST {
+            fastly_req.set_header(name.as_str(), value.as_bytes());
+        } else {
+            fastly_req.append_header(name.as_str(), value.as_bytes());
+        }
```
aram356, line 321: ♻️ **refactor** — This `set_header`-vs-`append_header` split has no effect on the bytes sent to origin: `BackendConfig::ensure` *always* calls `override_host(...)`, and per the Fastly SDK that "will force the value of the 'Host' header ... when sending out the origin request." So this only keeps the in-memory request from carrying a duplicate Host — worth a comment so it doesn't read as load-bearing for the override feature (and note it applies to every backend, not just the publisher origin).

```suggestion
        // `fastly::Request::new` derives a Host header from the request URI, so
        // appending the edge request's own Host would leave a duplicate. Replace
        // it instead to keep the in-memory request well-formed. The Host actually
        // sent on the wire is still governed by the backend's `override_host`
        // (see `BackendConfig::ensure`), which forces the value regardless.
        if name == edgezero_core::http::header::HOST {
            fastly_req.set_header(name.as_str(), value.as_bytes());
        } else {
            fastly_req.append_header(name.as_str(), value.as_bytes());
        }
```

aram356 (commented):
## Summary

Correct, well-tested, and CI-green. The publisher Host-override is implemented through the right mechanism — `origin_host_header_override` → `PlatformBackendSpec.host_header_override` → `BackendConfig::ensure()` → `Backend::override_host(...)`, which the Fastly SDK documents as forcing the outbound Host. The SVG Image Optimizer bypass is a genuine correctness fix with a focused regression test. No blocking issues; two non-blocking suggestions (inline) to make intent explicit and add the unit-testable half of the override coverage.

## Non-blocking

### ♻️ refactor

- **Document the Host `set_header` normalization** — it's defensive only; the wire Host is forced by the backend's `override_host`, and it applies to every backend, not just the publisher origin. (`crates/trusted-server-adapter-fastly/src/platform.rs:312` — suggestion inline)
- **Add request-side override coverage** — assert a configured `origin_host_header_override` reaches the outbound request's Host header via `StubHttpClient`. (`crates/trusted-server-core/src/publisher.rs` — suggested test inline)

### 👍 praise

- **SVG bypass checks both incoming and rewritten paths** with a targeted regression test. (`crates/trusted-server-core/src/proxy.rs`)
- **Backend name encodes the override** (`_oh_` suffix) so different overrides don't collide or get poisoned by first-registration-wins, and the value is validated against control characters. (`crates/trusted-server-core/src/backend.rs`)

### ⛏ nitpick

- History includes "Load Prebid split bundle synchronously" and its immediate revert (net-zero in the diff) — consider squashing on merge.

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS
- js tests: PASS
- integration / browser tests: PASS

crates/trusted-server-core/src/publisher.rs
```diff
@@ -499,6 +499,7 @@ pub async fn handle_publisher_request(
             scheme: origin_scheme.clone(),
             host: origin_host_without_port.to_string(),
             port: parsed_origin.port(),
+            host_header_override: settings.publisher.origin_host_header_override.clone(),
```
aram356, line 502: ♻️ **refactor** — The override wiring is currently only covered indirectly (backend-name encoding). The request-side half is unit-testable with the existing `StubHttpClient`: assert that a configured `origin_host_header_override` reaches the outbound request's Host header. (The Fastly `override_host` *forcing* still needs manual verification — it only runs in the Fastly runtime.) Suggested test for the `#[cfg(test)] mod tests` block in this file:

```rust
#[tokio::test]
async fn publisher_request_sends_configured_host_header_override() {
    let mut settings = create_test_settings();
    settings.publisher.origin_host_header_override = Some("www.example.com".to_string());
    let registry =
        IntegrationRegistry::new(&settings).expect("should create integration registry");
    let stub = Arc::new(StubHttpClient::new());
    stub.push_response(200, b"origin response".to_vec());
    let services = build_services_with_http_client(
        Arc::clone(&stub) as Arc<dyn crate::platform::PlatformHttpClient>
    );
    let req = HttpRequest::builder()
        .method(Method::GET)
        .uri("https://publisher.example/page")
        .header(header::HOST, "publisher.example")
        .body(EdgeBody::empty())
        .expect("should build request");

    handle_publisher_request(&settings, &registry, &services, req)
        .await
        .expect("should proxy publisher request");

    let recorded_headers = stub.recorded_request_headers();
    let outbound = recorded_headers
        .first()
        .expect("should record one outbound origin request");
    let host_header = outbound
        .iter()
        .find(|(name, _)| name.eq_ignore_ascii_case("host"))
        .map(|(_, value)| value.as_str());
    assert_eq!(
        host_header,
        Some("www.example.com"),
        "should send the configured origin host header override to the backend"
    );
}
```

crates/trusted-server-core/src/proxy.rs
```diff
@@ -1028,12 +1028,15 @@ pub async fn handle_asset_proxy_request(
...
+    let skip_image_optimizer = asset_path_skips_image_optimizer(incoming_path)
+        || asset_path_skips_image_optimizer(target_url.path());
     let image_optimizer = if skip_image_optimizer {
         log::debug!(
-            "Skipping Image Optimizer for unsupported SVG asset path: {}",
+            "Skipping Image Optimizer for unsupported SVG asset path: incoming={}, target={}",
+            incoming_path,
             target_url.path()
         );
         None
```
aram356, line 1042: 👍 **praise** — Checking `incoming_path` in addition to the rewritten `target_url.path()` correctly handles routes whose `target_path` strips the extension, and the new `handle_asset_proxy_request_skips_image_optimizer_for_incoming_svg` test pins exactly that case. The `&url::Url` → `&str` refactor is a clean simplification too.

aram356 (approved):
👍 Looks good

PR #798, opened 2026-06-22
Add `ts dev proxy`: a local MITM dev proxy serving production hostnames from staging

### LINKED ISSUE #824: Add `ts dev proxy`: local MITM dev proxy for serving production hostnames from staging
## Summary

Provide a local developer tool — **`ts dev proxy`** — that lets you open a
**production publisher hostname** in a real browser and have it served by a
**dev/staging upstream**, without touching production DNS, VCL, or certificates.

It's a TLS-terminating (MITM) forward proxy: for the configured host(s) it swaps
the TLS **SNI** to the upstream while the address bar keeps showing the
production hostname, so you can validate routing, cookies, `Host`-sensitive
logic, CMP/consent flows, and first-party context against a new Trusted Server
deployment before any DNS cutover.

## Scope

- New `ts` CLI (`crates/trusted-server-cli`, native binary excluded from the
  wasm workspace): `ts dev proxy` + `ts dev proxy ca install|uninstall|regenerate`.
- Flags: `--from/--to/--map`, `--rewrite-host`, `--resolve HOST:IP` (curl-style
  DNS pin), `--basic-auth`/`--basic-auth-file`, `--launch chrome|firefox|safari`,
  `--listen`, `--insecure`, `--allow-non-loopback`.
- CONNECT MITM with a per-machine local CA (trust once); blind-tunnel unmatched
  hosts; `421`/`403` guards; loopback-only by default; PAC route; browser + CA
  orchestration. macOS-only (deps cfg-scoped).

## Known follow-up (not in this issue)

Through the production Fastly adapter, `--rewrite-host`'s `X-Forwarded-Host:
FROM` signal is stripped as spoofable, so absolute first-party URLs anchor to
`TO`. The proper fix is a constrained, authenticated first-party-host override
in core/adapter (or relying on root-relative URLs) — tracked separately.

Implemented in #798.

aram356:
Closes #824

## Summary

Adds **`ts dev proxy`** — a local TLS-terminating (MITM) developer proxy that serves a **production publisher hostname from a dev/staging upstream** by swapping the TLS SNI. A real browser shows the production domain in the address bar while a Compute/staging service answers, so you can exercise Trusted Server's host-anchored URL rewriting against real first-party behavior without DNS or `/etc/hosts` changes.

> This PR started as the design spec + implementation plan and now contains the **full implementation** (new `trusted-server-cli` crate, CA, MITM server, browser/PAC orchestration, e2e tests, docs). ~9k lines across 22 files.

## What's included

**New `ts` developer CLI** (`crates/trusted-server-cli`, a native binary excluded from the wasm workspace):

```
ts dev proxy --from www.example-publisher.com --to staging.example.net --launch chrome
ts dev proxy ca install | uninstall | regenerate
```

**Proxy flags**
- `--map FROM=TO` (repeatable) or `-f/--from` + `-t/--to HOST[:PORT]` — the rewrite rule(s); rules are **explicit** (no `trusted-server.toml` inference).
- `--rewrite-host` — send `Host: TO` upstream (default keeps `Host: FROM` so core's URL rewriting stays anchored to the production host).
- `--resolve HOST:IP` (repeatable) — curl-style DNS pin; dials the given IP on every upstream path while keeping SNI/`Host` on the hostname, so the cert still validates. Makes the proxy self-contained (no hosts-file edits).
- `--basic-auth USER:PASS` / `--basic-auth-file PATH` — inject `Authorization` when absent (clears a gated upstream).
- `--listen ADDR` (default `127.0.0.1:18080`), `--launch chrome|firefox|safari`, `--insecure` (accept self-signed upstreams), `--allow-non-loopback`.

## Example

```bash
# Serve the production host from a staging upstream, pinning the upstream
# connection to a specific staging IP (no DNS / /etc/hosts changes), injecting
# Basic auth, and opening Chrome at the production URL:
ts dev proxy \
  --from www.example-publisher.com \
  --to   staging.example.net \
  --resolve staging.example.net:192.0.2.10 \
  --rewrite-host \
  --basic-auth "$DEV_USER:$DEV_PASS" \
  --launch chrome
```

`--resolve staging.example.net:192.0.2.10` dials `192.0.2.10` for the upstream
leg while leaving the TLS **SNI** and `Host` as `staging.example.net`, so the
upstream certificate still validates — the curl `--resolve` model. Repeat the
flag to pin multiple hosts. IPv6 targets work too (the value is split on the
first `:`), e.g. `--resolve staging.example.net:2001:db8::10`.

## Architecture

- **CONNECT MITM**: matched hosts get a freshly minted leaf (SNI→`TO`) signed by a per-machine local CA; the decrypted stream is proxied request-by-request over keep-alive. Unmatched hosts are **blind-tunnelled on loopback** (never MITM'd), and a `Host` matching no rule is refused `421` (no smuggling through the CONNECT-authority rule).
- **Per-machine CA** (`ca install` adds it to the login keychain; `uninstall`/`regenerate` revoke prior trust before replacing key material).
- **Browser + PAC orchestration**: HTTPS-only proxying via a served `/proxy.pac` route; Chrome/Firefox/Safari launch helpers (Safari persists & restores system-proxy state across hard kills).
- **`X-Forwarded-Host`/`X-Orig-Host = FROM`** so the upstream can keep emitting first-party URLs on the production host even when `--rewrite-host` sends `Host: TO`.

## Safety

- Binds **loopback-only** by default; off-loopback unmatched CONNECT/forward is refused `403` (can't become an open proxy) unless `--allow-non-loopback`.
- **macOS-only**: dependencies are `cfg`-scoped to macOS, so non-macOS targets compile to an empty shell with a clear message instead of a hard `compile_error!` (which previously broke `cargo build` under the repo's wasm default target).

## Tests, CI, docs

- ~43 unit + e2e tests (`tests/proxy_e2e.rs`): MITM rewrite/forward, blind tunnel cert identity, `--resolve` pinning, `--rewrite-host` Forwarded-Host behavior, injected Basic auth, keep-alive multi-request, `421`/`403` guards.
- New **native CI job** for the CLI crate (built with an explicit host target, since the repo pins `wasm32-wasip1` by default).
- `docs/guide/ts-dev-proxy.md` (setup, trust, troubleshooting) + the design spec and implementation plan under `docs/superpowers/`.

## Notes

- The spec/plan went through several review rounds; resolved items are recorded in-document.


ChristianPavilonis (commented):
Automated review: I reviewed the new `trusted-server-cli` dev proxy implementation, CA/browser orchestration, CI wiring, and the related docs. CI is currently green and I did not find a blocking issue that requires requesting changes, but I found several correctness/security edge cases that should be addressed before relying on this tool broadly.

crates/trusted-server-cli/src/commands/dev/proxy/mod.rs
```diff
@@ -0,0 +1,230 @@
...
+                // stays trusted would contradict the "invalidates prior trust"
+                // promise and leave an exfiltrated old key usable.
+                if !browser::ca_uninstall() {
+                    return Err(error_stack::Report::new(ProxyError::CertAuthority).attach(
+                        "could not revoke the previously-installed CA from the keychain; \
+                         aborting regenerate so on-disk key material still matches OS trust. \
+                         Remove the old CA manually (Keychain Access), then retry.",
+                    ));
+                }
+                std::fs::remove_file(&cert_path).ok();
```
ChristianPavilonis: Automated review: **P1 — `ca regenerate` can silently reuse the old CA if deletion fails.**

These `remove_file(...).ok()` calls discard failures, and the following `load_or_generate()` will happily reload the existing cert/key if they are still readable. That means `ts dev proxy ca regenerate` can print `regenerated CA` even though the old key material remains in use, which breaks the security promise users rely on after a suspected key compromise. Please propagate deletion failures (or verify the new cert/key differs from the old pair) and only print success after the files were actually replaced.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
```diff
@@ -0,0 +1,705 @@
...
+    // Remove it first so the `X-Forwarded-Host` we stamp is the one core reads,
+    // keeping emitted first-party URLs on the production host even when
+    // `--rewrite-host` sends `Host: TO` for routing/validation (spec §8.3). The
+    // `insert`s below already overwrite any inbound `X-Forwarded-Host`/`X-Orig-Host`.
+    headers.remove("forwarded");
+    if let Ok(value) = HeaderValue::from_str(&outcome.orig_host) {
+        headers.insert(HeaderName::from_static(X_FORWARDED_HOST), value.clone());
+        headers.insert(HeaderName::from_static(X_ORIG_HOST), value);
+    }
+    if let Some(auth) = basic_auth
```
ChristianPavilonis, line 634: Automated review: **P1 — `--allow-non-loopback` plus `--basic-auth` exposes the injected credentials to any network client that can reach the proxy.**

Matched CONNECTs are MITM'd even on non-loopback binds, and this branch injects the developer's upstream `Authorization` header for every matched request that lacks one. A LAN client can therefore proxy through `https://<FROM>/` and reach the gated staging upstream using the developer's credentials. Please reject `--basic-auth`/`--basic-auth-file` with non-loopback listens unless a separate explicit unsafe flag or proxy authentication is configured, or otherwise restrict matched MITM traffic to trusted clients.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
```diff
@@ -0,0 +1,705 @@
...
+        headers.insert(hyper::header::HOST, value);
+    }
+    // Tell the upstream the original first-party host (always `FROM`). Trusted
+    // Server resolves the request host from `Forwarded` → `X-Forwarded-Host` →
+    // `Host`, so a client-supplied `Forwarded` would outrank the value we inject.
+    // Remove it first so the `X-Forwarded-Host` we stamp is the one core reads,
+    // keeping emitted first-party URLs on the production host even when
+    // `--rewrite-host` sends `Host: TO` for routing/validation (spec §8.3). The
+    // `insert`s below already overwrite any inbound `X-Forwarded-Host`/`X-Orig-Host`.
+    headers.remove("forwarded");
```
ChristianPavilonis, line 619: Automated review: **P2 — plaintext upstreams lose the original HTTPS scheme signal.**

For `--upstream-plaintext`, the browser still visited `https://FROM`, but the upstream leg is plain HTTP and the proxy only stamps host headers. Trusted Server derives URL-rewrite scheme from TLS metadata, `Forwarded`, `X-Forwarded-Proto`, or `Fastly-SSL`; with a local plaintext upstream it will default to `http` (or honor a spoofed inbound scheme header), so generated first-party URLs can be downgraded. Please strip inbound scheme-forwarding headers here and stamp an authoritative `X-Forwarded-Proto: https` (and/or the signal the local adapter expects) to preserve the browser-facing scheme.

crates/trusted-server-cli/src/commands/dev/proxy/browser.rs
```diff
@@ -0,0 +1,817 @@
...
+    }
+
+    // Import CA into the profile's NSS DB via certutil. If certutil is missing
+    // or fails, Firefox would launch with no CA trust, so warn with the exact
+    // manual command instead of silently continuing.
+    let cert_path = super::ca::CertAuthority::cert_path(&cfg.ca_dir);
+    if cert_path.exists() {
+        let cert = cert_path.to_string_lossy();
+        let profile = tmpdir.to_string_lossy();
+        let certutil = Command::new("certutil")
```
ChristianPavilonis, line 383: Automated review: **P2 — Firefox CA import is attempted against an uninitialized NSS database.**

`tmpdir` is a freshly-created profile directory, but `certutil -A -d <dir>` expects an existing NSS DB (and without `sql:` it may also target the legacy DB format). On a clean temp profile this commonly fails with `SEC_ERROR_BAD_DATABASE`, after which Firefox launches without trusting the dev CA and proxied HTTPS fails despite `--launch firefox`. Please initialize the profile DB first (for example `certutil -N --empty-password -d sql:<profile>`) and import with `-d sql:<profile>`, or launch Firefox once to create the DB before importing.

crates/trusted-server-cli/src/commands/dev/proxy/browser.rs
```diff
@@ -0,0 +1,817 @@
...
+    }
+}
+
+/// Launches Chrome in a temporary profile configured for HTTPS-only proxying.
+///
+/// Uses `--proxy-server="https=<addr>"` so only HTTPS traffic goes through the
+/// proxy; HTTP and other schemes bypass it (spec §9).
+fn launch_chrome(cfg: &ResolvedConfig) {
+    let port = cfg.listen.port();
+    let proxy_arg = format!("https=127.0.0.1:{port}");
```
ChristianPavilonis: Automated review: **P2 — browser launch ignores the configured listen address.**

The CLI accepts arbitrary loopback/non-loopback `--listen` addresses, but Chrome (here), Firefox, and Safari hard-code `127.0.0.1:<port>` when configuring the proxy/PAC URL. Launching with `--listen [::1]:...`, `127.0.0.2:...`, or a specific non-loopback address can bind successfully while every launched browser points at the wrong endpoint. Please derive a browser-connect address from `cfg.listen` (normalizing wildcard binds intentionally) and use it consistently in Chrome, Firefox, Safari, and the manual instructions.

aram356:
Thanks for the review — all five findings addressed in `1f529a1ba`:

- **P1 (`ca regenerate` silent reuse)** — old cert/key deletions now propagate failures (already-absent tolerated) and the success message only prints after the files were actually removed; a failed delete aborts instead of reloading the stale pair.
- **P1 (creds on non-loopback)** — `config::resolve` now rejects `--basic-auth`/`--basic-auth-file` on a non-loopback `--listen` (`BasicAuthNonLoopback`), since a matched CONNECT is MITM'd off loopback too.
- **P2 (plaintext scheme downgrade)** — `rewrite_headers` stamps an authoritative `X-Forwarded-Proto: https` (browser leg is always TLS) and strips spoofable `Fastly-SSL`; inbound `Forwarded` was already stripped.
- **P2 (Firefox NSS DB)** — initialise an empty `sql:` DB (`certutil -N --empty-password -d sql:<profile>`) before importing, and import with `-d sql:<profile>`; manual hint updated to both commands.
- **P2 (listen address ignored)** — added `proxy_connect_addr`/`proxy_connect_host` (wildcard→loopback, IPv6 bracketed) and used them in Chrome/Firefox/Safari/PAC + manual hints, so a non-default `--listen` is honored.

Unit tests added for each; CLI crate fmt/clippy/test green.

crates/trusted-server-cli/src/commands/dev/proxy/mod.rs
aram356: Fixed in 1f529a1ba: the old cert/key deletions now propagate failures (already-absent tolerated via NotFound) and abort the regenerate; "regenerated CA" only prints after the files were actually removed, so a failed delete can't leave the stale key silently in use.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
aram356: Fixed in 1f529a1ba: `config::resolve` now rejects `--basic-auth`/`--basic-auth-file` on a non-loopback `--listen` (new `ConfigError::BasicAuthNonLoopback`), since a matched CONNECT is MITM'd off loopback too. Test: `basic_auth_on_non_loopback_listen_is_rejected`.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
aram356: Fixed in 1f529a1ba: `rewrite_headers` now stamps an authoritative `X-Forwarded-Proto: https` (the browser leg is always TLS), overwriting any inbound value, and strips spoofable `Fastly-SSL`; inbound `Forwarded` was already stripped. Test: `rewrite_headers_stamps_https_scheme_and_drops_spoofed_signals`.

crates/trusted-server-cli/src/commands/dev/proxy/browser.rs
aram356: Fixed in 1f529a1ba: we now initialise an empty modern DB first (`certutil -N --empty-password -d sql:<profile>`) and import with `-d sql:<profile>`, instead of importing against an uninitialised profile DB. The manual hint prints both commands.

crates/trusted-server-cli/src/commands/dev/proxy/browser.rs
aram356: Fixed in 1f529a1ba: added `proxy_connect_addr`/`proxy_connect_host` (wildcard bind → loopback, IPv6 bracketed) and used them in Chrome, Firefox, Safari, the PAC, and the manual hints, so a non-default `--listen` is honored. Tests: `proxy_connect_addr_normalizes_*`, `pac_uses_normalized_connect_address_for_wildcard_bind`.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
```diff
@@ -0,0 +1,748 @@
...
+    );
+    headers.remove("fastly-ssl");
+    if let Some(auth) = basic_auth
+        && !headers.contains_key(hyper::header::AUTHORIZATION)
+        && let Ok(value) = HeaderValue::from_str(&auth.header_value())
+    {
+        headers.insert(hyper::header::AUTHORIZATION, value);
+    }
+    // Drop the hop-by-hop proxy header so it never reaches the upstream (spec §8.3).
+    headers.remove("proxy-connection");
```
ChristianPavilonis: Automated review: **P2 — Strip all inbound hop-by-hop headers before stamping authoritative proxy headers.**

`rewrite_headers` overwrites `Host`, `X-Forwarded-Host`, `X-Orig-Host`, and `X-Forwarded-Proto`, but leaves the inbound `Connection` header and any headers named by it in place (only `Proxy-Connection` is removed). A client can send something like `Connection: X-Forwarded-Host, X-Forwarded-Proto, Authorization`; any downstream HTTP intermediary/framework that honors RFC hop-by-hop semantics may then discard exactly the headers this proxy relies on for first-party host anchoring, scheme preservation, or Basic-auth injection.

Please parse the inbound `Connection` tokens before rewriting, remove `Connection`, the standard hop-by-hop request headers (`Keep-Alive`, `Proxy-Authenticate`, `Proxy-Authorization`, `TE`, `Trailer`, `Transfer-Encoding`, `Upgrade`), and every token-named header, then stamp the authoritative headers. A regression test with `Connection: X-Forwarded-Host` should prove the upstream still receives the proxy-injected `X-Forwarded-Host: FROM`.

crates/trusted-server-cli/src/commands/dev/proxy/browser.rs
```diff
@@ -0,0 +1,901 @@
...
+
+/// Creates a unique temp directory under the system temp dir.
+///
+/// Returns `None` and prints a warning with `label` if the directory cannot be created.
+fn make_temp_dir(label: &str) -> Option<PathBuf> {
+    let ts = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map_or(0, |d| d.as_nanos());
+    let dir = std::env::temp_dir().join(format!("ts-dev-proxy-{label}-{ts}"));
+    match std::fs::create_dir_all(&dir) {
```
ChristianPavilonis: Automated review: **P2 — Browser profiles are created in a predictable temp path with `create_dir_all`.**

`make_temp_dir` uses `/tmp/ts-dev-proxy-{label}-{timestamp}` and `create_dir_all`, which succeeds if the path already exists. On a shared macOS machine, another local user/process can pre-create or race that predictable directory and cause Chrome/Firefox to run with an attacker-controlled profile directory, exposing or influencing the dev browser session.

Please use a secure temp-directory primitive (for example `tempfile::Builder` as a runtime dependency and keep the `TempDir` alive until the browser exits), or create a random 0700 directory with `create_dir` and fail on `AlreadyExists`.

crates/trusted-server-cli/src/commands/dev/proxy/ca.rs
```diff
@@ -0,0 +1,285 @@
...
+    /// with mode `0600`.  On first run a trust hint is logged.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`CaError`] on directory, I/O, or certificate generation failures.
+    pub fn load_or_generate(ca_dir: &Path) -> Result<Self, Report<CaError>> {
+        let cert_path = ca_dir.join(CA_CERT_FILE);
+        let key_path = ca_dir.join(CA_KEY_FILE);
+
+        let (cert_pem, key_pem) = if cert_path.exists() && key_path.exists() {
```
ChristianPavilonis, line 79: Automated review: **P2 — Re-secure existing CA material before reusing it.**

The first-generation path creates the CA directory/key with 0700/0600, but the reload path immediately reads any existing `ca-cert.pem` + `ca-key.pem` without checking or fixing permissions. If an earlier build, manual copy, backup restore, or partial recovery left the trusted root CA private key group/world-readable, the tool will continue using that exposed key indefinitely.

Please ensure `ca_dir` is 0700 and `ca-key.pem` is a regular file with 0600 permissions (or reject with a clear error) before loading existing CA material.

crates/trusted-server-cli/src/commands/dev/proxy/browser.rs
```diff
@@ -0,0 +1,901 @@
...
+/// the CA's common name after repeated installs, so it deletes until none are
+/// found. On non-macOS systems, prints a manual note and returns `true`. Never
+/// panics.
+pub fn ca_uninstall() -> bool {
+    #[cfg(target_os = "macos")]
+    {
+        // Delete every keychain entry matching the CA's CN; stop when none remain.
+        for _ in 0..16 {
+            let present = Command::new("security")
+                .args(["find-certificate", "-c", CA_COMMON_NAME])
```
ChristianPavilonis: Automated review: **P2 — Scope CA uninstall to the same login keychain used for install.**

`ca_install` explicitly trusts the certificate in `~/Library/Keychains/login.keychain-db`, but `ca_uninstall` runs `security find-certificate/delete-certificate -c ...` without passing that keychain. That searches the default keychain list, so a matching stale/duplicate cert in another keychain can make regenerate/uninstall fail even after the login-keychain entry is gone, or can delete an unintended matching certificate before reaching the installed one.

Please pass the login keychain path to both `find-certificate` and `delete-certificate` (and ideally delete by fingerprint/DER match rather than CN alone) so uninstall/regenerate affects the exact trust location this tool manages.

ChristianPavilonis (commented):
Automated review: I reviewed the current head of PR #798 (new `trusted-server-cli` dev proxy, CA/browser orchestration, tests/CI, and docs) against `main`. CI is green and I did not find a blocking correctness/security issue requiring REQUEST_CHANGES, but I found several medium-risk hardening/correctness issues worth addressing before broad use of this MITM dev tool.

crates/trusted-server-cli/src/main.rs
```diff
@@ -0,0 +1,16 @@
...
+fn main() {
+    env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info")).init();
+    std::process::exit(Cli::parse().run());
+}
+
+// `ts dev proxy` is macOS-only (its deps are macOS-scoped in `Cargo.toml`), so
+// on other targets the crate is an empty shell; this trivial entry point just
+// keeps the binary target's shape valid.
+#[cfg(not(target_os = "macos"))]
+fn main() {}
```
ChristianPavilonis, line 13: Automated review: **P2 — Unsupported platforms currently get a successful no-op `ts` binary.**

On non-macOS targets the binary `main` is empty, so `cargo run --manifest-path crates/trusted-server-cli/Cargo.toml -- dev proxy --help` exits 0 and prints nothing. Because the crate still builds a `ts` binary on those platforms, contributors/users can interpret a successful command as success even though no CLI was parsed and no proxy started.

Please make the non-macOS entry point print a clear unsupported-platform error and exit nonzero (or expose a minimal Clap surface that reports the macOS-only limitation), rather than silently succeeding.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
aram356: Fixed in 59bb30434: added `strip_hop_by_hop` which removes `Connection` + every header it names, plus the standard hop-by-hop set (`Keep-Alive`, `Proxy-Authenticate`, `Proxy-Authorization`, `TE`, `Trailer`, `Transfer-Encoding`, `Upgrade`, `Proxy-Connection`), run at the top of `rewrite_headers` before the authoritative headers are stamped. Regression test `rewrite_headers_strips_connection_named_headers_but_keeps_injected` proves `Connection: x-forwarded-host` still yields the proxy-injected `X-Forwarded-Host: FROM`.

crates/trusted-server-cli/src/commands/dev/proxy/browser.rs
aram356: Fixed in 59bb30434: `make_temp_dir` now uses `tempfile::Builder` (random name, 0700, O_EXCL) instead of a predictable timestamp path + `create_dir_all`; the returned `TempDir` is kept alive in the cleanup thread until the browser exits, then dropped to remove the profile. `tempfile` promoted to a runtime dependency.

crates/trusted-server-cli/src/commands/dev/proxy/ca.rs
aram356: Fixed in 59bb30434: the reload path now re-applies dir `0700` and key `0600` (propagating failures) before reading existing CA material, matching the freshly-generated posture. Test `reload_resecures_loosened_key_permissions` loosens the key to 0644 and asserts reload restores 0600.

crates/trusted-server-cli/src/commands/dev/proxy/browser.rs
aram356: Fixed in 59bb30434: added `login_keychain()` and pass it to both `find-certificate` and `delete-certificate` in `ca_uninstall` (and `ca_install` uses it too), so uninstall/regenerate operate on exactly the login keychain this tool trusts into.

crates/trusted-server-cli/src/main.rs
aram356: Fixed in 59bb30434: the non-macOS `main` now prints a clear unsupported-platform error via `output::error` and `exit(2)` instead of an empty no-op that exits 0.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
```diff
@@ -0,0 +1,829 @@
...
+    rewrite_headers(req.headers_mut(), outcome, basic_auth);
+
+    // Dial the `--resolve` pin when the upstream host has one; the SNI/`Host`
+    // (set above) stay the hostname, so the certificate still validates.
+    let tcp = connect_upstream(upstream_host, upstream_port, resolve)
+        .await
+        .change_context(ProxyError::Server)?;
+
+    let response = if outcome.scheme_is_tls {
+        let connector = TlsConnector::from(client_config(insecure));
```
prk-Jr, line 517: ♻️ **refactor** — `client_config(insecure)` runs per upstream request, rebuilding a `rustls::ClientConfig` and re-parsing the entire `webpki_roots::TLS_SERVER_ROOTS` store each time. The service closure at server.rs:341 also clones `rules`/`basic_auth`/`resolve` per request. Since none of this changes across requests, build the client config once (store an `Arc<rustls::ClientConfig>` in `ResolvedConfig`, or a `OnceLock`) and share the immutable inputs. Not urgent at dev-tool scale, but it's constant per-request work.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
```diff
@@ -0,0 +1,829 @@
...
+    // No match on loopback: connect upstream FIRST, then reply 200 (else 502).
+    blind_tunnel(client, &host, port, &cfg.resolve).await
+}
+
+/// Opens an upstream TCP connection, honoring a `--resolve` pin for `host`.
+///
+/// When `host` (matched case-insensitively) has a pin, the socket dials that IP
+/// instead of resolving the name via DNS. The TLS SNI / `Host` set by the caller
+/// are unaffected, so the certificate still validates against the hostname.
+async fn connect_upstream(
```
prk-Jr, line 237: 🌱 **seedling** — `connect_upstream` has no connect/read timeout. A black-holed upstream (easy to hit with a wrong `--resolve` pin) will stall the browser tab until the OS TCP timeout. Wrapping the dial (and, for the TLS case, the handshake) in a `tokio::time::timeout` would fail fast into the existing `502` path.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
```diff
@@ -0,0 +1,550 @@
...
+        resolve,
+    })
+}
+
+/// Credential precedence: `--basic-auth-file` > `--basic-auth`.
+fn resolve_basic_auth(args: &ProxyArgs) -> Result<Option<BasicAuth>, ConfigError> {
+    if let Some(path) = &args.basic_auth_file {
+        let raw = std::fs::read_to_string(path)
+            .map_err(|_| ConfigError::BasicAuthFile { path: path.clone() })?;
+        return Ok(Some(BasicAuth::parse(raw.trim())?));
```
prk-Jr: ⛏ **nitpick** — `BasicAuth::parse(raw.trim())` on the file contents strips the trailing newline (good) but also any meaningful leading/trailing whitespace in the password. Trimming only trailing `\r`/`\n` would preserve such passwords.

prk-Jr (approved):
## Summary

`ts dev proxy` is a well-built, security-conscious addition. I read the full proxy surface (server, rewrite, config, ca, browser, mod, run), the e2e suite, module gating, dependency scoping, and CI wiring. The audit/config changes are pure moves (96–100% similarity — import-path + rustfmt only, no logic change). No blocking issues found; approving. The notes below are all optional follow-ups.

## What stands out (👍)

- **Security posture**: loopback-only default with `403` off-loopback refusal, `421` misdirected-request guard against Host smuggling over a MITM tunnel, hop-by-hop + inbound `Forwarded` stripping so the injected `X-Forwarded-Host` wins, authoritative `X-Forwarded-Proto: https` + `Fastly-SSL` drop, refusing `--basic-auth` on a non-loopback bind, `0600` CA key with re-secure-on-reload, and `ca regenerate` aborting if OS trust cannot first be revoked.
- **Detail**: byte-at-a-time head read so the TLS ClientHello is never consumed before the CONNECT `200`; IP-literal hosts get an IP-type SAN.
- **Test + CI design**: 43 unit/e2e tests including the blind-tunnel cert-identity assertion (proves an unmatched host is not MITM'd), IPv6 `--resolve`, keep-alive multi-request, and the 421/403 guards; macOS-native CI exercises the proxy while the Linux-native job covers audit and the non-macOS fallback arms.

## Non-blocking notes

### ♻️ refactor / perf
- **Per-request TLS client config** — `crates/trusted-server-cli/src/commands/dev/proxy/server.rs:461` (`proxy_to_upstream`) rebuilds a `rustls::ClientConfig` and re-parses the whole `webpki_roots` store on every upstream hop; the service closure at `server.rs:341` also clones `rules`/`basic_auth`/`resolve` per request. Build the client config once (Arc in `ResolvedConfig` or a `OnceLock`) and share the immutable inputs. Dev-tool scale, so not urgent.

### 🌱 seedling
- **No upstream timeout** — `connect_upstream` (`server.rs:212`) has no connect/read timeout; a black-holed upstream (easy with a wrong `--resolve` pin) stalls the browser tab until the OS TCP timeout. A `tokio::time::timeout` around dial + handshake would fail fast into the existing `502` path.
- **No upstream connection reuse** — a fresh TCP + TLS handshake per request even within one keep-alive client tunnel (`server.rs:438`). Acknowledged tradeoff; noting for a future pass.

### 🤔 thinking
- **Concurrent Safari instances** — two concurrent `ts dev proxy --launch safari` runs collide: the second's startup `restore_system_proxy_if_pending` reverts the first's live system PAC and deletes its restore file (`browser.rs:456`). Single-instance is an implicit assumption; worth a doc note or a lockfile guard.

### ⛏ nitpick
- **Credential trimming** — `resolve_basic_auth` (`config.rs:305`) trims the whole value/file; correct for the trailing newline but it also strips a password's meaningful leading/trailing whitespace. Trimming only trailing `\r`/`\n` is safer.
- **`is_valid_host` leniency** — accepts leading/trailing `.` and `..` (`config.rs:208`). Safe for the PAC/URL/Host interpolation contexts (the stated purpose), just not a strict hostname.

## CI Status

All checks green at time of review (fmt, clippy across adapters, rust tests incl. ts-CLI native + cross-adapter parity, vitest, docs format).

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
```diff
@@ -0,0 +1,829 @@
...
+        headers.insert(hyper::header::HOST, value);
+    }
+    // Tell the upstream the original first-party host (always `FROM`). Trusted
+    // Server resolves the request host from `Forwarded` → `X-Forwarded-Host` →
+    // `Host`, so a client-supplied `Forwarded` would outrank the value we inject.
+    // Remove it first so the `X-Forwarded-Host` we stamp is the one core reads,
+    // keeping emitted first-party URLs on the production host even when
+    // `--rewrite-host` sends `Host: TO` for routing/validation (spec §8.3). The
+    // `insert`s below already overwrite any inbound `X-Forwarded-Host`/`X-Orig-Host`.
+    headers.remove("forwarded");
```
ChristianPavilonis, line 619: 🔧 **P1 — `--rewrite-host` does not preserve the production host on real Trusted Server adapters**

Issue: the proxy relies on `X-Forwarded-Host: FROM` to keep Trusted Server URL rewriting anchored to the production host, and the docs say `--rewrite-host` is safe with TS upstreams. However, the real Fastly and Spin adapter paths strip `X-Forwarded-Host` before routing, so with `--rewrite-host` core falls back to `Host: TO`.

Why it matters: when a staging/upstream service requires `Host: TO`, first-party URLs/cookies can be generated on the staging host instead of the production host, which breaks the core dev-proxy validation scenario.

Suggested fix: either document/restrict `--rewrite-host` as incompatible with real TS adapters until there is a trusted original-host channel, or add a dev-only trusted mechanism in the adapters (for example a secret-gated/signed original-host header preserved before sanitization). Please add an adapter-level test proving `--rewrite-host` still yields `request_host == FROM`.

edgezero.toml
```diff
@@ -1,9 +1,48 @@
...
+# -- Stores ------------------------------------------------------------------
+# Logical store ids only. The union covers every adapter's binding; `default`
+# names the primary id (Fastly production for kv/config/secrets). Runtime store
+# names resolve per platform outside this manifest.
+
+[stores.kv]
+ids = ["trusted_server_kv", "ec_identity_store"]
+default = "trusted_server_kv"
+
+[stores.config]
```
ChristianPavilonis, line 27: 🔧 **P1 — `ts config push/diff` now resolve the app config to the wrong store**

Issue: the unified manifest declares only `trusted_server_config` as the config store default, but Trusted Server's runtime default app-config store remains `app_config`, and existing configs still reference `config_store_id = "app_config"`. I verified `ts config push ... --store app_config` now fails because `app_config` is not in `[stores.config].ids`, while omitting `--store` dry-runs against `trusted_server_config`.

Why it matters: operators can push/diff the app config into the rollout/config-flag store instead of the runtime app-config store, and cannot override back to the existing `app_config` store. That can break deploy/config parity and request-signing config lookup.

Suggested fix: keep `app_config` in `[stores.config].ids` and make it the default for app-config operations. Keep `trusted_server_config` only if it is intentionally a separate declared store. Also reconcile the `secrets`/`trusted_server_secrets` naming or migrate runtime configs together.

crates/trusted-server-cli/src/commands/dev/proxy/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+    // never blocks on a sudo password prompt.
+    browser::restore_system_proxy_if_pending(&config::ca_dir(args), false);
+
+    let cfg = Arc::new(config::resolve(args).change_context(ProxyError::Config)?);
+
+    let ca = Arc::new(
+        ca::CertAuthority::load_or_generate(&cfg.ca_dir)
+            .change_context(ProxyError::CertAuthority)?,
+    );
+    let pac: Arc<str> = Arc::from(browser::generate_pac(&cfg.rules, cfg.listen).as_str());
```
ChristianPavilonis: 🔧 **P2 — `--listen :0` launches browsers/PAC on port 0 instead of the bound port**

Issue: the PAC is generated from `cfg.listen` before binding, and browser launch also uses `cfg.listen`. If a user passes `--listen 127.0.0.1:0`, the OS binds an ephemeral real port, but the proxy logs/configures port `0`.

Why it matters: port `0` is a common local-dev way to avoid collisions; with `--launch`, every browser points at an unusable endpoint.

Suggested fix: bind first, read `listener.local_addr()`, then use that effective listen address for PAC generation, logging, `serve_on`, and browser launch. Add coverage for the `--listen 127.0.0.1:0 --launch`/PAC path.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
```diff
@@ -0,0 +1,829 @@
...
+        let tls = connector
+            .connect(server_name, tcp)
+            .await
+            .change_context(ProxyError::Server)?;
+        send_over(TokioIo::new(tls), req).await?
+    } else {
+        send_over(TokioIo::new(tcp), req).await?
+    };
+
+    Ok(response.map(http_body_util::BodyExt::boxed))
```
ChristianPavilonis, line 533: 🔧 **P2 — strip upstream hop-by-hop response headers before returning to the browser tunnel**

Issue: request hop-by-hop headers are stripped, but upstream response headers are boxed and returned unchanged. Headers like `Connection: close` or `Connection: x-debug` can leak onto the browser↔proxy connection.

Why it matters: one upstream response can close or corrupt an otherwise reusable MITM tunnel, undermining keep-alive behavior even though each upstream request uses a fresh upstream connection.

Suggested fix: make the response mutable before boxing, strip response `Connection`, standard hop-by-hop response headers, and every header named by `Connection`. Add a regression test where the upstream sends `Connection: close, x-debug` plus `x-debug` and the client can still issue a second request on the same MITM tunnel.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
```diff
@@ -0,0 +1,829 @@
...
+        blind_forward_http(client, &head, &cfg.resolve).await
+    } else {
+        respond_status_line(&mut client, StatusCode::FORBIDDEN).await
+    }
+}
+
+/// Splits `host:port`, defaulting the port to 443.
+fn split_authority(authority: &str) -> (String, u16) {
+    match authority.rsplit_once(':') {
+        Some((host, port)) => (host.to_string(), port.parse().unwrap_or(443)),
```
ChristianPavilonis: 🌱 **P3 — malformed CONNECT heads should be rejected cleanly**

Issue: invalid ports currently fall back to `443`, and oversized request heads stop at the 8192-byte cap but still route a partially-read request.

Why it matters: bad client input can produce misleading `200 Connection Established` / TLS failures instead of a clear proxy parse error, which makes debugging much harder.

Suggested fix: make request-head / authority parsing return explicit errors and respond with `400 Bad Request` or `431 Request Header Fields Too Large` before entering MITM/tunnel mode.

crates/trusted-server-cli/src/commands/dev/proxy/browser.rs
```diff
@@ -0,0 +1,921 @@
...
+                let _ = Command::new("open").args(["-a", "Safari", &url]).status();
+                output::info(&format!("Safari: PAC set for '{service}'; opened {url}"));
+            } else {
+                output::info(&format!(
+                    "Safari: PAC URL set for '{service}'; open Safari and browse to a proxied host"
+                ));
+            }
+        }
+        _ => {
+            output::warn(&format!(
```
ChristianPavilonis, line 538: 🌱 **P3 — shell-quote the Safari setup fallback command**

Issue: this fallback message interpolates the macOS network service name inside double quotes. Service names are local config values and can contain shell metacharacters; double quotes still allow command substitution like `$()` and backticks.

Why it matters: a developer copy/pasting the fallback could execute unintended shell code.

Suggested fix: use the existing `shell_quote()` helper for both the service name and PAC URL, or reuse a helper analogous to `manual_restore_command`.

docs/guide/ts-dev-proxy.md
```diff
@@ -0,0 +1,291 @@
...
+**Primary use case.** Validate routing and behavior of a new or changed Trusted
+Server deployment at the publisher's real domain — cookies, `Host`-sensitive
+logic, CMP/consent flows, first-party context — before any DNS cutover.
+
+**Non-goals.** Not a production proxy or load-testing tool. Does not modify any
+Fastly service. Local only, developer-facing.
+
+## Build and run
+
+`ts dev proxy` is part of `crates/trusted-server-cli`, a native binary excluded
```
ChristianPavilonis: 🌱 **P3 — workspace docs/lockfile model is stale**

Issue: this guide says `trusted-server-cli` is excluded from the workspace, but the root workspace now includes it. Related cleanup: `crates/trusted-server-cli/Cargo.lock` is ignored by Cargo because the CLI is a workspace member, and the integration-test README still describes the old excluded-crate/nested-lockfile/dependency-check workflow.

Why it matters: this can mislead maintainers about which lockfile pins dependencies and how integration-test dependencies are maintained.

Suggested fix: delete the nested CLI lockfile unless the crate is intentionally removed from the workspace, and update the dev-proxy guide plus integration-test README to state that the root `Cargo.lock` / `[workspace.dependencies]` model is authoritative.

ChristianPavilonis (approved):
## Summary

📝 Approved with comments. I reviewed the current PR #798 diff against `main`, including the new `ts dev proxy`, CA/browser orchestration, workspace/EdgeZero manifest changes, CI, and docs. CI is green and I did not find any P0 blockers.

I left inline notes using the CREG guide: 🔧 for changes worth addressing before broad use, and 🌱 for lower-risk follow-ups/hardening.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
aram356: Fixed in 2eb346fd: `client_config` now caches the built config per verification mode with a `OnceLock`, so the webpki root store is parsed once instead of on every request.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
aram356: Fixed in 2eb346fd: added `--connect-timeout` (default 10s) that bounds the upstream dial with `tokio::time::timeout`, failing fast into the existing 502 path.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
aram356: Fixed in 2eb346fd: the `--basic-auth-file` contents are now stripped only of a trailing `\r`/`\n` (`trim_end_matches(['\r','\n'])`), so leading/trailing whitespace in the password survives.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
aram356: Corrected in 2eb346fd. Dropped the blanket "safe with Trusted Server upstreams" claim from the guide and softened the code comment: the proxy still stamps `X-Forwarded-Host: FROM`, but since the Fastly/Spin adapter paths strip it before routing, `--rewrite-host` falls back to `Host: TO` there and first-party URLs emit on `TO`. Documented as a caveat and recommended the default (no `--rewrite-host`) with TS upstreams.

edgezero.toml
aram356: The unified manifest now declares the logical `trusted_server_config` id, with an inline note that store ids are overridable per adapter/runtime (a deliberate choice to keep the portable Trusted Server names rather than a platform-specific one). So `ts config push`/`diff` resolve against `trusted_server_config` by default. Open question I want your read on: if the deploy workflow must keep `--store app_config` working, the runtime override needs to map `trusted_server_config → app_config`, or `app_config` should be pinned in `[stores.config].ids`. Happy to pin `app_config` if the override doesn't cover it.

crates/trusted-server-cli/src/commands/dev/proxy/mod.rs
aram356: Fixed in 2eb346fd: `run()` binds first and sets `cfg.listen = listener.local_addr()` before generating the PAC and launching browsers, so an ephemeral `--listen …:0` uses the real OS-assigned port.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
aram356: Fixed in 2eb346fd: `strip_hop_by_hop` is now applied to the upstream response headers too, so a `Connection: close` (or named connection token) can't leak onto the reusable browser↔proxy MITM tunnel.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
aram356: Fixed in 2eb346fd: `split_authority` returns `None` on a non-numeric port (→ 400), and a head that hits EOF/the 8192-byte cap without `\r\n\r\n` is marked incomplete, which `handle_connection` rejects with 400 instead of routing a partially-read request.

crates/trusted-server-cli/src/commands/dev/proxy/browser.rs
aram356: Fixed in 2eb346fd: the Safari fallback now runs the service name and PAC URL through the existing `shell_quote()` before interpolating them into the copy-pasteable `networksetup` command.

docs/guide/ts-dev-proxy.md
aram356: Fixed across 6a35b7f3/2eb346fd: the guide and the integration-tests README now describe the CLI and integration crates as workspace members (shared root lockfile, no drift-check), and the stale tracked `crates/trusted-server-cli/Cargo.lock` is removed.

PR #799, opened 2026-06-22
Add ts cli

### LINKED ISSUE #770: Port EdgeZero CLI to TS
### Description

Start porting edge zero pr 269 to trusted server

ChristianPavilonis:
## Summary

- Adds the host-target `ts` operator CLI crate and binary.
- Wires EdgeZero-backed lifecycle delegation and Trusted Server config init/validate/push flows.
- Moves runtime settings loading to flattened app-config entries and updates docs/CI for the host-target CLI.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-cli/` | Adds the base CLI crate, clap args, dispatch, EdgeZero delegate, config commands, and tests. |
| `crates/trusted-server-core/src/config_payload.rs` | Adds canonical flattened config payload generation and hash metadata. |
| `crates/trusted-server-core/src/settings_data.rs` | Loads settings from the platform config store instead of embedded build output. |
| `crates/trusted-server-adapter-fastly/src/app.rs`, `src/main.rs` | Reads settings through the Fastly config-store adapter. |
| `.github/workflows/*`, `.cargo/config.toml` | Adds host-target CLI check/test coverage and local aliases. |
| `edgezero.toml`, `trusted-server.example.toml`, `.gitignore` | Adds EdgeZero manifest/template config and ignores operator-owned config. |
| `docs/guide/cli.md`, `docs/guide/getting-started.md`, `README.md` | Documents base CLI install/config workflow. |

## Closes

No issue provided; this PR is split out from the combined `feature/ts-cli-next` branch.

## Test plan

- [ ] `cargo test --workspace`
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] Other: `cargo test --package trusted-server-cli --target aarch64-apple-darwin` — 12 passed

## Checklist

- [x] Uses tracing/log macros (not `println!`)
- [x] New code has tests


crates/trusted-server-core/src/proxy.rs
```diff
@@ -37,8 +37,15 @@ const IMAGE_FALLBACK_CONTENT_TYPE: &str = "application/octet-stream";
...
+fn body_as_reader(body: EdgeBody) -> Result<Cursor<bytes::Bytes>, Report<TrustedServerError>> {
+    Ok(Cursor::new(body.into_bytes().unwrap_or_default()))
+}
+
+fn request_body_bytes(
+    body: EdgeBody,
+    _endpoint: &str,
+) -> Result<bytes::Bytes, Report<TrustedServerError>> {
+    Ok(body.into_bytes().unwrap_or_default())
 }
```
prk-Jr, line 56: ♻️ **refactor** — `request_body_bytes` takes an `_endpoint: &str` it never uses and returns a `Result` it never fails — it just maps a streaming body to empty bytes via `unwrap_or_default()` (the pinned edgezero `Body::into_bytes()` returns `None` for streams). The signature implies the intent — matching the new `auction_rejects_streaming_body_instead_of_treating_as_empty` test — was to reject streaming bodies. Make it actually reject, naming the endpoint:

```suggestion
fn request_body_bytes(
    body: EdgeBody,
    endpoint: &str,
) -> Result<bytes::Bytes, Report<TrustedServerError>> {
    if body.is_stream() {
        return Err(Report::new(TrustedServerError::BadRequest {
            message: format!("{endpoint} request body must be buffered, not streamed"),
        }));
    }
    Ok(body.into_bytes().unwrap_or_default())
}
```

_(Scratch-verified: `cargo fmt --check`, `cargo clippy -p trusted-server-core --all-targets --all-features -D warnings`, `cargo test -p trusted-server-core` — 1426 passed, 0 failed.)_

crates/trusted-server-core/src/request_signing/endpoints.rs
```diff
@@ -24,6 +24,13 @@ fn json_response(status: StatusCode, body: String) -> Response<EdgeBody> {
         .expect("should build json response")
 }
 
+fn request_body_bytes(
+    body: EdgeBody,
+    _endpoint: &str,
+) -> Result<bytes::Bytes, Report<TrustedServerError>> {
+    Ok(body.into_bytes().unwrap_or_default())
+}
```
prk-Jr, line 38: ♻️ **refactor** — Same incomplete abstraction as in `proxy.rs`: `_endpoint` is unused and the `Result` never fails, so a streaming body becomes empty bytes. This matters more here — `handle_rotate_key` treats an empty body as a valid no-kid rotation, so a streamed `rotate-key` body would be silently accepted rather than rejected. Make the helper actually reject streaming bodies:

```suggestion
fn request_body_bytes(
    body: EdgeBody,
    endpoint: &str,
) -> Result<bytes::Bytes, Report<TrustedServerError>> {
    if body.is_stream() {
        return Err(Report::new(TrustedServerError::BadRequest {
            message: format!("{endpoint} request body must be buffered, not streamed"),
        }));
    }
    Ok(body.into_bytes().unwrap_or_default())
}
```

_(Scratch-verified: `cargo fmt --check`, `cargo clippy -p trusted-server-core --all-targets --all-features -D warnings`, `cargo test -p trusted-server-core` — 1426 passed, 0 failed.)_

crates/trusted-server-core/src/settings_data.rs
```diff
@@ -1,223 +1,286 @@
...
+        ));
+    }
 
-    // Validate the settings
-    settings
-        .validate()
-        .change_context(TrustedServerError::Configuration {
-            message: "Failed to validate configuration".to_string(),
-        })?;
+    let mut envelope_json = String::with_capacity(pointer.envelope_len);
```
prk-Jr, line 122: 🤔 **thinking** — `envelope_len` comes straight from the operator-written config-store pointer and is used to pre-size the buffer before any chunk is read or validated. A corrupt/huge value triggers a large allocation before the post-concat length check can reject it. Operator-controlled surface, so low severity — but the capacity hint buys nothing here; `push_str` grows as needed and the existing length/sha checks still validate the result:

```suggestion
    let mut envelope_json = String::new();
```

_(Scratch-verified: `cargo fmt --check`, `cargo clippy -p trusted-server-core --all-targets --all-features -D warnings`, `cargo test -p trusted-server-core` — 1426 passed, 0 failed.)_

crates/trusted-server-integration-tests/tests/common/ec.rs
```diff
@@ -269,30 +271,35 @@ fn mappings_to_json(mappings: &[BatchMapping]) -> Vec<Value> {
...
-/// `405` for methods outside its registered set (e.g. `TRACE`), whereas the
-/// legacy path proxied every method through to the publisher origin. Without it,
-/// a fixture/env/config-store regression could green the EdgeZero CI job while
-/// it actually exercises legacy.
+/// This probe used to assert a router-level `405` for unsupported methods, but
+/// Viceroy/Fastly method handling can fall through to the publisher fallback.
+/// Keep the request as a non-fatal diagnostic so the EdgeZero CI job still runs
+/// the EC lifecycle scenarios instead of failing on a routing canary that is not
+/// stable across runtime versions.
 pub fn assert_edgezero_entry_point(base_url: &str) -> TestResult<()> {
```
prk-Jr: 🤔 **thinking** / 📌 **out of scope** — This canary was downgraded from a hard `405` assertion to a non-fatal `log::warn!` + `Ok(())`. Its original purpose was to prove the `integration-tests-edgezero` CI job actually exercises `edgezero_main` rather than silently falling back to `legacy_main` (the EC lifecycle scenarios pass on **both** paths). As a warning it can no longer fail CI, so a fixture/config-store regression that drops the EdgeZero job onto the legacy path would green both jobs — which is the exact failure this guard existed to catch, and directly relevant given this PR rewires runtime config loading.

The flaky-`405` concern is legitimate, but consider restoring a *deterministic* EdgeZero-only signal (e.g. a header or route that only `edgezero_main` emits) rather than removing the guard, or track it as a follow-up issue. Was the downgrade intentional, or just to get CI green?

_Apply manually — no single-line fix; needs a stable EdgeZero-only assertion._

prk-Jr (commented):
## Summary

Introduces the host-target `ts` operator CLI and, more significantly, re-architects runtime settings loading: config moves from a build-time-embedded TOML (`include_bytes!(target/trusted-server-out.toml)`, baked by `build.rs`) to a runtime read of an EdgeZero `BlobEnvelope` from a platform config store, with per-chunk + envelope SHA-256 integrity verification and a Fastly chunk-pointer reassembly path. `build.rs` becomes a no-op, the `config` crate drops to a dev-dependency, and EdgeZero git deps are pinned to a fixed `rev`.

Clean, well-tested work — strong unit coverage (round-trip, tamper-rejection, chunk reassembly, placeholder rejection) and good error messages. No blocking issues; all findings below are non-blocking.

> 3 of the inline comments carry a one-click GitHub `suggestion` (verified in a scratch worktree: `cargo fmt --check`, `cargo clippy -p trusted-server-core --all-targets --all-features -D warnings`, and `cargo test -p trusted-server-core` — 1426 passed). The canary comment is prose because it has no single clean fix.

## Non-blocking

### ♻️ refactor
- **`request_body_bytes` is an incomplete abstraction (proxy)** — see inline at `crates/trusted-server-core/src/proxy.rs:44`
- **`request_body_bytes` is an incomplete abstraction (request signing)** — see inline at `crates/trusted-server-core/src/request_signing/endpoints.rs:27`

### 🤔 thinking
- **Unbounded pre-allocation from operator-controlled `envelope_len`** — see inline at `crates/trusted-server-core/src/settings_data.rs:113`
- **EdgeZero entry-point canary downgraded to a no-op** — see inline at `crates/trusted-server-integration-tests/tests/common/ec.rs:283`

- 🤔 **Startup `INSECURE: proxy.certificate_check is disabled` warning dropped** — the old `load_settings` emitted a startup `log::warn!` when `proxy.certificate_check` was off; the new blob path (`settings_from_config_blob` / `get_settings_from_config_store`) does not. A per-backend warning still exists at request time (`adapter-fastly/src/backend.rs:219`), so coverage isn't zero, but the loud startup signal is gone. Cheap to re-add in `get_settings_from_config_store`.
- 📝 **`#[serde(deny_unknown_fields)]` added broadly** (`settings.rs`, `consent_config.rs`, `auction_config_types.rs`) — good hardening (catches typo'd keys; the `from_toml` test flipped from "extra fields ignored" to "rejected"), but it's a behavior change: any deployed config carrying extra/forward-compat keys now fails to load. Validated at `ts config push` time, which is the right place — flagging the migration implication for existing operator configs.
- 🌱 **Minor**: `adapter-fastly/src/main.rs` opens the same `TRUSTED_SERVER_CONFIG_STORE` twice per request (`open_trusted_server_config_store` for the flag + `edgezero_config_store_handle` for dispatch) — harmless, mildly wasteful. Separately, `.github/workflows/test.yml`'s "Verify Fastly WASM release build" step still sets `TRUSTED_SERVER__PUBLISHER__ORIGIN_URL`, now vestigial since `build.rs` no longer bakes env-derived config.

### 👍 praise
- EdgeZero git deps pinned from `branch = "main"` to a fixed `rev` — reproducible builds.
- Integrity model in `resolve_fastly_chunk_pointer` + `settings_from_config_blob`: per-chunk len+sha, envelope len+sha, **and** `BlobEnvelope::verify()` — solid defense-in-depth, with `tampered_blob_hash_is_rejected` and `strings_that_look_like_json_scalars_round_trip_as_strings` guarding the TOML→JSON→TOML secret round-trip.
- `build.rs` reduced from 72 lines of `#[path]`-included module hacks to a 3-line no-op.

## CI Status — all 14 checks PASS

- cargo test: PASS (required)
- cargo fmt: PASS (required)
- Analyze (rust): PASS
- Analyze (actions): PASS
- CodeQL: PASS
- vitest: PASS
- integration tests: PASS
- integration tests (EdgeZero entry point): PASS

Cargo.toml
```diff
@@ -30,14 +31,17 @@ build-print = "1.0.1"
...
-edgezero-adapter-fastly = { git = "https://github.com/stackpop/edgezero", branch = "main", default-features = false }
-edgezero-core = { git = "https://github.com/stackpop/edgezero", branch = "main", default-features = false }
+edgezero-adapter = { git = "https://github.com/stackpop/edgezero", rev = "89f592665a8b386111995da5cbf3fcc068113dfc", default-features = false }
+edgezero-adapter-axum = { git = "https://github.com/stackpop/edgezero", rev = "89f592665a8b386111995da5cbf3fcc068113dfc", default-features = false }
+edgezero-adapter-cloudflare = { git = "https://github.com/stackpop/edgezero", rev = "89f592665a8b386111995da5cbf3fcc068113dfc", default-features = false }
+edgezero-adapter-fastly = { git = "https://github.com/stackpop/edgezero", rev = "89f592665a8b386111995da5cbf3fcc068113dfc", default-features = false }
+edgezero-cli = { git = "https://github.com/stackpop/edgezero", rev = "89f592665a8b386111995da5cbf3fcc068113dfc" }
+edgezero-core = { git = "https://github.com/stackpop/edgezero", rev = "89f592665a8b386111995da5cbf3fcc068113dfc", default-features = false }
```
aram356: ♻️ **refactor** — Track the in-flight EdgeZero CLI branch rather than the superseded rev. PR #799 depends on EdgeZero CLI APIs that have since moved to `feature/extensible-cli` ([stackpop/edgezero#269](https://github.com/stackpop/edgezero/pull/269)). Point all six crates at it:

```suggestion
edgezero-adapter = { git = "https://github.com/stackpop/edgezero", branch = "feature/extensible-cli", default-features = false }
edgezero-adapter-axum = { git = "https://github.com/stackpop/edgezero", branch = "feature/extensible-cli", default-features = false }
edgezero-adapter-cloudflare = { git = "https://github.com/stackpop/edgezero", branch = "feature/extensible-cli", default-features = false }
edgezero-adapter-fastly = { git = "https://github.com/stackpop/edgezero", branch = "feature/extensible-cli", default-features = false }
edgezero-cli = { git = "https://github.com/stackpop/edgezero", branch = "feature/extensible-cli" }
edgezero-core = { git = "https://github.com/stackpop/edgezero", branch = "feature/extensible-cli", default-features = false }
```

Verified locally against branch HEAD `bf177b13`: `cargo check` passes for `trusted-server-core` (wasm32-wasip1), `trusted-server-adapter-fastly` (wasm32-wasip1), and `trusted-server-cli` (host); CLI tests 8/8. ⚠️ A `branch =` ref is not reproducible across CI machines — pin back to a rev before merge.

aram356 (commented):
## Summary

Review of #799. The substantive change is sound: runtime config moves from build-time-baked TOML to a config-store blob envelope with solid integrity checks (per-chunk + envelope len/SHA-256 + `verify()` + post-load placeholder rejection), and the host-only `ts` CLI is cleanly isolated from the wasm build. CI is green. Findings are mostly non-blocking — no 🔧 blockers — with one open ❓ question on secret handling worth resolving before merge.

## Blocking

### ❓ question

- **Plaintext secrets in the app-config blob** — `crates/trusted-server-core/src/config.rs:92` (inline).

## Non-blocking

### 🤔 thinking

- **Settings reloaded up to 3×/request** — `crates/trusted-server-adapter-fastly/src/main.rs:355,383` (inline).
- **Dead build-time env exports** — `scripts/integration-tests.sh:54-58` (inline).
- **EdgeZero entry-point canary downgraded to a warning** — `crates/trusted-server-integration-tests/tests/common/ec.rs:296` (inline).
- **Auction test name overstates coverage** — `crates/trusted-server-core/src/auction/endpoints.rs:808` (inline).
- **Broad `#[serde(deny_unknown_fields)]`** (`settings.rs` + ~20 config structs): unknown/typo'd keys now hard-fail load, which at runtime drops the whole app to the startup-error router. Good safety, but potentially breaking for existing operator configs that carried extra keys — worth a release note.

### ♻️ refactor

- **No-op `Result` body wrappers with unused `_endpoint`** — `crates/trusted-server-core/src/proxy.rs:41`, `crates/trusted-server-core/src/request_signing/endpoints.rs:27` (inline).
- **Track the EdgeZero `feature/extensible-cli` branch** — `Cargo.toml:39-44`, `crates/trusted-server-integration-tests/Cargo.toml:13` (inline suggestions). Verified compiling against branch HEAD `bf177b13`.

### 🌱 seedling / 📝 note

- `prebid::validate_config_for_startup` lacks a direct unit test — `crates/trusted-server-core/src/integrations/prebid.rs:201` (inline).
- Integration `cargo test` now depends on a generated config artifact — `crates/trusted-server-integration-tests/tests/environments/fastly.rs:82` (inline).
- **Unused workspace dependency** `edgezero-adapter` (`Cargo.toml:39`) — no first-party crate references it (only transitive via `edgezero-cli`). Drop it or document intent.
- **Reproducibility** — a `branch =` ref (per the suggestions above) is not reproducible across machines/CI; pin back to a specific rev before merge.

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS
- js tests (vitest): PASS
- integration tests (incl. EdgeZero entry point): PASS

crates/trusted-server-integration-tests/Cargo.toml
```diff
@@ -9,14 +9,18 @@ name = "integration"
 path = "tests/integration.rs"
 harness = true
 
-[dev-dependencies]
+[dependencies]
+edgezero-core = { git = "https://github.com/stackpop/edgezero", rev = "89f592665a8b386111995da5cbf3fcc068113dfc", default-features = false }
```
aram356: ♻️ **refactor** — Same branch switch for the integration-tests crate.

```suggestion
edgezero-core = { git = "https://github.com/stackpop/edgezero", branch = "feature/extensible-cli", default-features = false }
```

This crate builds without `--locked` (`scripts/integration-tests.sh`), so its `Cargo.lock` re-resolves on the next run.

crates/trusted-server-core/src/config.rs
```diff
@@ -0,0 +1,275 @@
...
+
+impl Validate for TrustedServerAppConfig {
+    fn validate(&self) -> Result<(), ValidationErrors> {
+        validate_settings_for_deploy(&self.settings)
+            .map_err(|report| report_to_validation_errors(&report))
+    }
+}
+
+impl edgezero_core::app_config::AppConfigMeta for TrustedServerAppConfig {
+    const SECRET_FIELDS: &'static [edgezero_core::app_config::SecretField] = &[];
```
aram356, line 97: ❓ **question** — `SECRET_FIELDS` is empty, so `ts config push` serializes the entire `Settings` — including `ec.passphrase`, `publisher.proxy_secret`, `ec.partners[].api_token`, and `handlers[].password` — as plaintext JSON into the `app_config` config store. `edgezero.toml` declares a dedicated `[stores.secrets]` that nothing routes to. Is keeping these in the app-config blob intentional (relying on config-store ACLs / parity with the previous wasm-baked secrets), or should they be declared as `SecretField`s so EdgeZero routes them to the secrets store? Worth resolving before merge.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -327,7 +352,7 @@ fn edgezero_main(mut req: FastlyRequest, config_store: ConfigStoreHandle) {
         // verbs) carry TS/geo headers. Middleware-finalized responses are
         // skipped here to avoid a second settings read and geo lookup on the
         // normal registered-route path.
-        match get_settings() {
+        match load_settings_from_config_store() {
```
aram356: 🤔 **thinking** — Settings are fully reloaded here, again at line 383 (EC finalize), and once more in `build_state()` — up to 3× per request. Each `load_settings_from_config_store()` does a config-store read → JSON envelope parse → SHA-256 over the whole blob → `serde_json` deserialize → `finalize_deserialized` (handler-regex recompile + full validation). The removed `OnceLock` only memoized *within* a request (Fastly is fresh-instance-per-request, so it never cached across requests) — but it did collapse these repeats to one parse. Consider threading a single request-scoped `Settings` (or reusing `AppState.settings`) to the finalize points instead of reloading.

crates/trusted-server-core/src/proxy.rs
```diff
@@ -37,8 +37,15 @@ const IMAGE_FALLBACK_CONTENT_TYPE: &str = "application/octet-stream";
 const SIGN_MAX_BODY_BYTES: usize = 65536;
 const REBUILD_MAX_BODY_BYTES: usize = 65536;
 
-fn body_as_reader(body: EdgeBody) -> Cursor<bytes::Bytes> {
-    Cursor::new(body.into_bytes().unwrap_or_default())
+fn body_as_reader(body: EdgeBody) -> Result<Cursor<bytes::Bytes>, Report<TrustedServerError>> {
+    Ok(Cursor::new(body.into_bytes().unwrap_or_default()))
```
aram356: ♻️ **refactor** — `request_body_bytes`/`body_as_reader` return `Result` and take `_endpoint: &str`, but `body.into_bytes().unwrap_or_default()` can never fail and `_endpoint` is unused, so the new `?` at the call sites is a no-op. Streaming/oversized bodies are still silently turned into empty bytes (the size check runs *after* full collection). This reads as half-finished work toward `Body::into_bytes_bounded(max)`. Either finish that (reject oversized/streaming bodies) or drop the `Result`/param churn. Same pattern in `request_signing/endpoints.rs:27`.

crates/trusted-server-core/src/request_signing/endpoints.rs
```diff
@@ -24,6 +24,13 @@ fn json_response(status: StatusCode, body: String) -> Response<EdgeBody> {
         .expect("should build json response")
 }
 
+fn request_body_bytes(
```
aram356, line 27: ♻️ **refactor** — Same no-op `Result` wrapper as `proxy.rs:41` (`_endpoint` unused, `unwrap_or_default()` can't fail, `?` is a no-op). Fold both into a real bounded reader or revert the churn.

crates/trusted-server-integration-tests/tests/common/ec.rs
```diff
@@ -269,30 +271,35 @@ fn mappings_to_json(mappings: &[BatchMapping]) -> Vec<Value> {
...
+        )
         .send()
         .change_context(TestError::HttpRequest)
-        .attach("TRACE / (EdgeZero entry-point canary)")?;
-    assert_status(&response, 405).attach(
-        "EdgeZero canary: TRACE should return a router-level 405; a non-405 status \
-         means main() fell back to the legacy entry point",
-    )
+        .attach("OPTIONS /_ts/api/v1/batch-sync (EdgeZero entry-point probe)")?;
+    if response.status().as_u16() != 405 {
```
aram356: 🤔 **thinking** — Downgrading the entry-point check from a hard `405` assertion to `log::warn!` removes the guard that this CI job actually exercises the EdgeZero entry point: it can now go green even if `main()` silently falls back to `legacy_main`. Consider asserting a positive EdgeZero-only signal (a header/behavior only the EdgeZero path emits) instead of dropping the assertion entirely.

scripts/integration-tests.sh
```diff
@@ -58,6 +58,10 @@ TRUSTED_SERVER__EC__PARTNERS='[{"name":"Integration Test Partner","source_domain
 TRUSTED_SERVER__PROXY__CERTIFICATE_CHECK=false \
```
aram356, line 60: 🤔 **thinking** — This `TRUSTED_SERVER__*` block (lines 54-58) no longer affects the binary: this PR removes build-time config baking (`core/build.rs` is now a stub), and runtime config comes from the generated Viceroy `app_config` blob. Origin port etc. already flow through `generate-integration-viceroy-configs.sh`, so these are dead/misleading — recommend removing them.

crates/trusted-server-core/src/auction/endpoints.rs
```diff
@@ -803,4 +803,51 @@ mod tests {
             );
         });
     }
+
+    #[test]
+    fn auction_rejects_streaming_body_instead_of_treating_as_empty() {
```
aram356, line 808: 🤔 **thinking** — The name overstates what's verified. `handle_auction` does `into_bytes().unwrap_or_default()`, so a streaming body *is* treated as empty; it only surfaces as `BadRequest` because empty input fails JSON parse. This test would still pass if streaming bodies were silently emptied. Use a body that would parse as `{}` when drained, or rename to reflect that it asserts empty-body parse failure.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -192,6 +192,24 @@ impl IntegrationConfig for PrebidIntegrationConfig {
     }
 }
 
+/// Validate enabled Prebid config using the same startup-only checks as runtime registration.
+///
+/// # Errors
+///
+/// Returns a configuration error if enabled Prebid settings fail typed parsing,
+/// schema validation, or bidder-param override compilation.
+pub fn validate_config_for_startup(
```
aram356, line 201: 🌱 **seedling** — `validate_config_for_startup` is only covered indirectly. A small direct test locking the no-silent-disable contract — enabled+invalid → `Err`, enabled+valid → `Ok(Some)`, absent → `Ok(None)` — would be worthwhile.

crates/trusted-server-integration-tests/tests/environments/fastly.rs
```diff
@@ -63,25 +63,23 @@ impl RuntimeEnvironment for FastlyViceroy {
...
-                return std::path::PathBuf::from(path);
-            }
+        if let Ok(path) = std::env::var("VICEROY_CONFIG_PATH")
+            && !path.is_empty()
+        {
+            return std::path::PathBuf::from(path);
         }
         std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"))
-            .join("fixtures/configs/viceroy-template.toml")
+            .join("../../target/integration-test-artifacts/configs/viceroy-legacy.toml")
```
aram356, line 88: 📝 **note** — The default `VICEROY_CONFIG_PATH` now points at a generated artifact (`target/integration-test-artifacts/configs/viceroy-legacy.toml`) that doesn't exist until `generate-integration-viceroy-configs.sh` runs. Running `cargo test` directly (outside the wrapper) without setting `VICEROY_CONFIG_PATH` fails to find it; the old default was source-controlled and self-sufficient. A clearer panic message pointing at the generator would help.

crates/trusted-server-core/src/proxy.rs
ChristianPavilonis: Fixed in fd0322c2. `request_body_bytes` now rejects `Body::Stream` with `TrustedServerError::BadRequest` using the endpoint name, and the first-party sign path has a streaming-body regression test. Response-body helpers were left unchanged because origin responses may intentionally be streamed.

crates/trusted-server-core/src/request_signing/endpoints.rs
ChristianPavilonis: Fixed in fd0322c2. The request-signing body helper now rejects streaming bodies instead of treating them as empty, with targeted tests for `verify-signature` and `rotate-key`.

crates/trusted-server-core/src/settings_data.rs
ChristianPavilonis: Fixed in fd0322c2. Replaced the operator-controlled capacity hint with `String::new()` while keeping the existing chunk length, envelope length, and SHA-256 validation unchanged.

crates/trusted-server-integration-tests/tests/common/ec.rs
ChristianPavilonis: Fixed in fd0322c2. The EdgeZero entry point now marks responses with `x-ts-entry-point: edgezero`, and the `EXPECT_EDGEZERO_ENTRY_POINT=true` integration canary hard-asserts that positive signal instead of relying on method/status behavior.

Cargo.toml
ChristianPavilonis: Understood, wont merge until edge zero pr is merged

crates/trusted-server-integration-tests/Cargo.toml
ChristianPavilonis: Same resolution as the workspace deps: keeping the integration-tests EdgeZero dependency pinned to a fixed `rev` for reproducible CI instead of using the moving feature branch.

crates/trusted-server-core/src/config.rs
ChristianPavilonis: Answered/explicitly deferred for this PR. The current `ts config push` behavior intentionally preserves the existing inline-settings model rather than introducing a partial secret-management feature here. Full secret refs/provisioning require EdgeZero support and are tracked in trusted-server#684 plus stackpop/edgezero#284; this PR should not half-implement nested/array secret extraction.

crates/trusted-server-adapter-fastly/src/main.rs
ChristianPavilonis: Fixed in fd0322c2. Normal EdgeZero responses now reuse the request-scope `AppState.settings` snapshot for entry-point finalization and EC finalization; settings reload is preserved only as a fallback for startup-error routers that have no valid state.

crates/trusted-server-core/src/proxy.rs
ChristianPavilonis: Fixed in fd0322c2. The proxy request-body helper now uses the endpoint parameter and rejects streaming request bodies with 400. Buffered bodies still flow through the existing max-size checks; response-body stream handling was intentionally left alone.

crates/trusted-server-core/src/request_signing/endpoints.rs
ChristianPavilonis: Fixed in fd0322c2. The request-signing helper now has a real error path for streamed bodies, and tests cover streamed `verify-signature` and `rotate-key` requests.

crates/trusted-server-integration-tests/tests/common/ec.rs
ChristianPavilonis: Fixed in fd0322c2. The EdgeZero path emits `x-ts-entry-point: edgezero`, and the EdgeZero integration job hard-fails if that header is missing.

scripts/integration-tests.sh
ChristianPavilonis: Fixed in fd0322c2. Removed the dead `TRUSTED_SERVER__*` build-time env block from `scripts/integration-tests.sh` and the vestigial workflow env for the Fastly release build; integration runtime config now comes from generated Viceroy `app_config`.

crates/trusted-server-core/src/auction/endpoints.rs
ChristianPavilonis: Acknowledged and left as a follow-up per the agreed scope. This PR only tightened the small request-body helper endpoints; auction stream semantics/test naming are separate P3 polish unless we touch auction body handling directly.

crates/trusted-server-core/src/integrations/prebid.rs
ChristianPavilonis: Acknowledged as follow-up coverage. I kept this out of the review-fix commit to avoid expanding scope beyond the agreed P1/P2 items.

crates/trusted-server-integration-tests/tests/environments/fastly.rs
ChristianPavilonis: Fixed in fd0322c2. `FastlyViceroy::spawn` now checks the resolved config path before spawning Viceroy and returns `RuntimeSpawn` with guidance to run `scripts/generate-integration-viceroy-configs.sh`, `scripts/integration-tests.sh`, or set `VICEROY_CONFIG_PATH`.

crates/trusted-server-core/src/config.rs
prk-Jr, line 97: ❓ **question** — `SECRET_FIELDS = &[]` opts out of EdgeZero secret-store routing. EdgeZero treats a `#[secret]` field's value as a *reference* into `[stores.secrets]`. With none declared, `ec.passphrase`, `publisher.proxy_secret`, `handler.password`, and `ec.partners[].api_token` are embedded as plaintext JSON in the `app_config` **config** store — even though `edgezero.toml` already declares a `secrets` store and `request_signing` resolves keys from it.

Pre-PR these secrets were baked into the WASM binary (arguably worse), so this is likely not a regression. Is leaving `SECRET_FIELDS` empty an intentional phase-1 decision, or should the app-level secrets migrate to the `secrets` store?

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -409,6 +430,68 @@ fn take_finalize_sentinel(response: &mut HttpResponse) -> bool {
...
+    request_filter_effects: Option<&RequestFilterEffects>,
+) {
+    if let Some(effects) = request_filter_effects {
+        effects.apply_to_response(&mut response);
+    }
+    mark_edgezero_entry_point(&mut response);
+    compat::to_fastly_response(response).send_to_client();
+}
+
+fn mark_edgezero_entry_point(response: &mut HttpResponse) {
```
prk-Jr: 🤔 **thinking** — `x-ts-entry-point: edgezero` is added to every EdgeZero-path response, including those served to real clients. It usefully drives the CI canary in `assert_edgezero_entry_point`, but it also exposes internal routing / migration state externally. Consider gating it to non-prod, namespacing it, or documenting it as an intentional diagnostic header.

crates/trusted-server-core/src/proxy.rs
```diff
@@ -37,8 +37,21 @@ const IMAGE_FALLBACK_CONTENT_TYPE: &str = "application/octet-stream";
 const SIGN_MAX_BODY_BYTES: usize = 65536;
 const REBUILD_MAX_BODY_BYTES: usize = 65536;
 
-fn body_as_reader(body: EdgeBody) -> Cursor<bytes::Bytes> {
-    Cursor::new(body.into_bytes().unwrap_or_default())
+fn body_as_reader(body: EdgeBody) -> Result<Cursor<bytes::Bytes>, Report<TrustedServerError>> {
```
prk-Jr: ♻️ **refactor** — `body_as_reader` now returns `Result` but always yields `Ok`; the `?` at call sites adds no real error path. For response bodies a streamed body is still silently emptied by `into_bytes().unwrap_or_default()`. Either add the stream-rejection guard (symmetry with the new request-body helpers) or keep the function infallible to avoid misleading error plumbing. Same applies to `publisher.rs:35`.

crates/trusted-server-core/src/settings_data.rs
```diff
@@ -1,223 +1,286 @@
...
-/// - [`TrustedServerError::InvalidUtf8`] if the embedded TOML file contains invalid UTF-8
-/// - [`TrustedServerError::Configuration`] if the configuration is invalid or missing required fields
-pub fn get_settings() -> Result<Settings, Report<TrustedServerError>> {
-    if let Some(settings) = SETTINGS.get() {
-        return Ok(settings.clone());
-    }
+/// Returns [`TrustedServerError::Configuration`] when the config blob is
+/// missing, cannot be read, fails envelope verification, or fails Trusted
+/// Server settings validation.
+pub fn get_settings_from_services(
```
prk-Jr: 🌱 **seedling** — `get_settings_from_services` is `pub` but has no in-repo caller (the Fastly adapter uses `get_settings_from_config_store` via `load_settings_from_config_store`). Presumably staged for future Cloudflare/Axum/Spin adapters that carry `RuntimeServices` — confirm that intent or drop it to avoid an unused public surface.

crates/trusted-server-core/src/settings_data.rs
```diff
@@ -1,223 +1,286 @@
...
 
-    let settings = load_settings()?;
-    if SETTINGS.set(settings.clone()).is_err() {
-        if let Some(settings) = SETTINGS.get() {
-            return Ok(settings.clone());
-        }
-    }
+/// Returns the default `EdgeZero` app-config store name.
+#[must_use]
+pub fn default_config_store_name() -> StoreName {
```
prk-Jr, line 33: ⛏ **nitpick** — `default_config_store_name` reads the raw env var directly while `default_config_key` (just below) resolves through `EnvConfig::from_env().store_key(...)`. Routing both through `EnvConfig` keeps a single source of truth for the `app_config` store identifiers.

crates/trusted-server-core/src/settings_data.rs
```diff
@@ -1,223 +1,286 @@
...
+    }
 
-    // Validate the settings
-    settings
-        .validate()
-        .change_context(TrustedServerError::Configuration {
-            message: "Failed to validate configuration".to_string(),
-        })?;
+    let mut envelope_json = String::new();
+    for chunk in pointer.chunks {
```
prk-Jr, line 124: 🌱 **seedling** — Chunks are concatenated into `envelope_json` before the total `envelope_len` check, so there's no upper bound on reassembled size until after the fact. Only an operator with config-store write access could abuse this, so risk is low — but an early cumulative-size cap would harden the path.

crates/trusted-server-cli/src/run.rs
```diff
@@ -0,0 +1,197 @@
...
+    Validate(ConfigValidateArgs),
+}
+
+/// Run the CLI using process arguments.
+///
+/// # Errors
+///
+/// Returns an error when command parsing, config validation, `EdgeZero`
+/// delegation, or config initialization fails.
+pub fn run_from_env() -> Result<(), String> {
```
prk-Jr, line 54: 📝 **note** — The CLI uses `Result<(), String>` rather than the repo's `error-stack` `Report<E>` convention. This is constrained by `edgezero_cli`'s String-based public API, so it's a defensible deviation for a thin delegation binary — noting for awareness, no action required.

scripts/check-integration-dependency-versions.sh
```diff
@@ -85,6 +85,27 @@ transitive_parity_allowlist=(
     "wasm-bindgen-shared"
     # Forced newer by an integration-only dependency.
     "num-conv"
+    # The workspace now pulls these through EdgeZero / production-only config
+    # tooling while the integration crate's native test stack resolves different
+    # compatible major/minor lines, or no longer resolves the old line after
+    # trusted-server-core stopped depending on the config crate.
+    "convert_case"
```
prk-Jr, line 92: 🤔 **thinking** — This PR adds ~18 crates to the parity allowlist (reqwest, hashbrown, winnow, toml_datetime, convert_case, windows-*). Pragmatic, but each entry weakens the shared-dependency drift guard. Worth confirming none of these are actually runtime-shared between the workspace and the integration crate (most look dev/test- or transitive-only).

prk-Jr (commented):
## Summary

Strong, well-tested PR. Unifies runtime config loading onto the EdgeZero `app_config` blob, adds the host-target `ts` operator CLI, pins all edgezero git deps to a fixed rev, and hardens request-body handling. Read every changed file; no bugs found. Findings below are non-blocking — one open question on secret-at-rest storage and several consistency/hardening suggestions.

## Blocking

### ❓ question

- **Secrets stored in the config-store blob, not the secret store** — `TrustedServerAppConfig::SECRET_FIELDS = &[]` keeps `ec.passphrase`, `publisher.proxy_secret`, `handler.password`, and `ec.partners[].api_token` as plaintext JSON inside the `app_config` config store. Intentional for this phase? (See inline comment on `config.rs:92`.)

## Non-blocking

### ♻️ refactor

- **Auction stream-body handling** — `auction/endpoints.rs:81` silently treats a streamed body as empty; add an explicit `is_stream()` guard for parity with `proxy.rs` / `request_signing`.
- **Vacuous `Result` on `body_as_reader`** — `proxy.rs:40` and `publisher.rs:35` return `Result` but never error; add the stream guard or keep infallible.

### 🤔 thinking

- **`x-ts-entry-point: edgezero` on all responses** — exposes internal routing/migration state to clients; used as a CI canary (`main.rs:488`). Consider gating/documenting.
- **Dependency-parity allowlist grew ~18 crates** — `check-integration-dependency-versions.sh`; each entry weakens the drift guard.

### 🌱 seedling

- **`get_settings_from_services` unused** — `settings_data.rs:42`, `pub` with no in-repo caller; confirm it's for future adapters or drop.
- **Unbounded blob reassembly** — `settings_data.rs:114`, chunks concatenated before the total-length check.

### 📝 note

- **`#[serde(deny_unknown_fields)]` is a behavior change** — existing `trusted-server.toml` files with extra/typo'd/legacy keys now hard-fail validation and runtime load (the round-trip test flipped from "extra fields ignored" to "rejected"). Good hardening — call it out in migration/release notes (`settings.rs`, plus the consent/auction config structs).
- **CLI error type** — `Result<(), String>` deviates from the `error-stack` convention but is constrained by `edgezero_cli`'s API (`run.rs:54`).

### ⛏ nitpick

- **Store-id resolution** — `default_config_store_name` reads raw env while `default_config_key` uses `EnvConfig` (`settings_data.rs:52`).

### 👍 praise

- Streaming-body rejection on sign/verify/rotate/deactivate request paths closes a real correctness gap (streamed bodies were silently treated as empty — could bypass signature verification), with thorough tests.
- Pinning all edgezero deps from `branch = "main"` to a fixed rev — reproducible builds.
- Defense-in-depth on Fastly chunk reconstruction: per-chunk len + sha, envelope len + sha, then envelope integrity verify.
- Reusing the request-scope `AppState` settings snapshot instead of re-reading the config store on every finalize / EC step — removes redundant reads and guarantees finalize uses the same settings that built the router.
- `is_edgezero_enabled` switched to `try_get` (no panic on backend error) with graceful legacy fallback.
- Removing build-time secret env injection from CI now that runtime loads from the config store.

## CI Status

- fmt: PASS
- clippy (workspace + host-target CLI): PASS
- rust tests (workspace + host-target CLI): PASS
- js tests (vitest): PASS
- integration / edgezero / browser: PASS
- docs + ts format, CodeQL: PASS

crates/trusted-server-core/src/proxy.rs
aram356: ⛏ **nitpick** (follow-up to the `request_body_bytes` fix) — `body_as_reader` still returns `Result` but `body.into_bytes().unwrap_or_default()` can't fail, so the `?` at `proxy.rs:424` is a no-op. Unlike `request_body_bytes`, it didn't get the `is_stream()` guard. This is the upstream-response buffering path (pre-existing behavior), so not a regression — but the dead `Result` could be dropped, or given the same stream check if a streamed upstream body should be rejected rather than silently emptied.

crates/trusted-server-adapter-fastly/src/main.rs
aram356: 📝 **note** — `x-ts-entry-point: edgezero` is now added to every normal EdgeZero response, so it ships to all production clients, not just the CI probe. That's a small always-on info-disclosure / fingerprint surface (reveals the routing path). If it's only needed as a test/rollout signal, consider gating it (debug flag or a specific endpoint); if it's intentionally a public rollout marker, all good — just flagging the visibility.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -351,55 +376,51 @@ fn edgezero_main(mut req: FastlyRequest, config_store: ConfigStoreHandle) {
...
-    // recognized browsers. When settings or the partner registry cannot be
-    // loaded the response is sent without EC finalization rather than
-    // dropped.
+    // recognized browsers. Reuse the request-scope settings from AppState;
+    // startup-error responses have no valid state and keep the previous
+    // best-effort reload fallback.
     if let Some(ec_state) = ec_state {
-        match get_settings() {
-            Ok(settings) => match PartnerRegistry::from_config(&settings.ec.partners) {
+        if let Some(settings) = settings_snapshot.as_deref() {
```
aram356, line 551: ♻️ **refactor** (optional) — The snapshot branch (`settings_snapshot`) and the reload fallback below repeat the same finalize → `send_edgezero_response` → pull-sync → `return` sequence. Resolving settings into a single value first (e.g. an owned `Settings`/`Cow`) and running one code path would remove the duplication. Non-blocking.

Cargo.toml
```diff
@@ -30,14 +31,17 @@ build-print = "1.0.1"
...
+clap = { version = "4", features = ["derive"] }
 config = "0.15.19"
 cookie = "0.18.1"
 derive_more = { version = "2.0", features = ["display", "error"] }
 ed25519-dalek = { version = "2.2", features = ["rand_core"] }
-edgezero-adapter-axum = { git = "https://github.com/stackpop/edgezero", branch = "main", default-features = false }
-edgezero-adapter-cloudflare = { git = "https://github.com/stackpop/edgezero", branch = "main", default-features = false }
-edgezero-adapter-fastly = { git = "https://github.com/stackpop/edgezero", branch = "main", default-features = false }
-edgezero-core = { git = "https://github.com/stackpop/edgezero", branch = "main", default-features = false }
+edgezero-adapter = { git = "https://github.com/stackpop/edgezero", rev = "89f592665a8b386111995da5cbf3fcc068113dfc", default-features = false }
```
aram356: ⛏ **nitpick** — `edgezero-adapter` is declared in `[workspace.dependencies]` but no first-party crate references it (only transitive via `edgezero-cli`). Drop it, or add a comment if it's a deliberate placeholder.

aram356 (approved):
## Re-review — follow-up `fd0322c2`

The follow-up commit resolves the substantive findings from the prior review, with tests:

- **Per-request settings reload** → `edgezero_main` now reuses the request-scope `AppState.settings` snapshot for entry-point + EC finalization (one load per request), with a reload fallback only for startup-error routers.
- **Streaming request bodies** → `request_body_bytes` rejects `Body::Stream` with `BadRequest`; sign / verify / rotate streaming-rejection tests added.
- **EdgeZero entry-point guard** → new `x-ts-entry-point: edgezero` header is hard-asserted by the EdgeZero CI job, replacing the dropped `405` canary.
- **Dead build-time env exports** removed from `scripts/integration-tests.sh` and `test.yml`; **strict key validation** documented; **missing Viceroy config** now errors with guidance.
- Bonus hardening: chunk-pointer reconstruction no longer pre-allocates from the operator-controlled `envelope_len` (`String::new()`), removing an untrusted-capacity vector. 👍

The open ❓ on `SECRET_FIELDS = &[]` is answered (intentional inline-secret behavior for this PR; secrets-store routing deferred), and the auction-test-name / prebid-direct-test items are acknowledged as scoped follow-ups.

Approving. A few non-blocking nits inline: a leftover no-op `Result` in `body_as_reader`, the always-on entry-point header's client visibility, EC-finalize branch duplication, and an unused workspace dependency.

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS
- js tests (vitest): PASS
- integration tests (incl. EdgeZero entry point): PASS

crates/trusted-server-integration-tests/src/bin/generate-viceroy-config.rs
```diff
@@ -0,0 +1,313 @@
...
+        )));
+    }
+
+    let generated_stores = generated_config_store_blocks(envelope_json, edgezero_enabled);
+    Ok(template.replace(GENERATED_STORES_MARKER, &generated_stores))
+}
+
+fn generated_config_store_blocks(envelope_json: &str, edgezero_enabled: bool) -> String {
+    let edgezero_enabled_value = if edgezero_enabled { "true" } else { "false" };
+    let edgezero_rollout_pct = if edgezero_enabled { "100" } else { "0" };
```
aram356, line 174: 🤔 **thinking** — Forcing `edgezero_rollout_pct = "100"` here is a good way to make the EdgeZero CI job take the EdgeZero path. But note `cfcd35885` also removed `assert_edgezero_entry_point`, the `x-ts-entry-point` header, and the `EXPECT_EDGEZERO_ENTRY_POINT` gate — so there's no longer a *positive assertion* that `test_ec_lifecycle_fastly` actually routed through `edgezero_main`. If `main()` silently falls back to `legacy_main` (e.g. the config store can't be opened), the EC scenarios still pass and the `integration tests (EdgeZero entry point)` job stays green. Forcing rollout makes that unlikely, but the explicit guard is gone. If keeping it matters, a debug-gated signal or a startup log-assertion would restore the guarantee without the always-on client-visible header. Non-blocking.

crates/trusted-server-adapter-cloudflare/src/app.rs
```diff
@@ -49,10 +58,37 @@ pub struct AppState {
 /// Returns an error when settings, the auction orchestrator, or the integration
 /// registry fail to initialise.
 fn build_state() -> Result<Arc<AppState>, Report<TrustedServerError>> {
-    let settings = get_settings()?;
+    #[cfg(target_arch = "wasm32")]
+    if let Some(settings) = settings_from_cloudflare_config_json()? {
+        return build_state_with_settings(settings);
+    }
+
+    let settings = Settings::from_toml(include_str!("../../../trusted-server.example.toml"))?;
```
aram356: 📝 **note** — When `TRUSTED_SERVER_CONFIG` is unset, `build_state` falls back to `Settings::from_toml(example.toml)`. Unlike the blob path (`settings_from_config_blob`), `from_toml` does not run `reject_placeholder_secrets`, so a misconfigured Worker boots with the template's placeholder secrets instead of failing fast. Dev-only adapter and pre-existing pattern, but worth a guard (or at least a `log::warn!`) so a missing config surfaces loudly. Non-blocking.

crates/trusted-server-integration-tests/tests/environments/cloudflare.rs
```diff
@@ -22,6 +23,31 @@ pub struct CloudflareWorkers;
...
+fn write_generated_ci_config(wrangler_dir: &Path) -> TestResult<String> {
+    let template_path = wrangler_dir.join(CI_CONFIG_TEMPLATE);
+    let template = std::fs::read_to_string(&template_path)
+        .change_context(TestError::RuntimeSpawn)
+        .attach(format!(
+            "failed to read Cloudflare CI wrangler config at {}",
+            template_path.display()
+        ))?;
+    let config_json = cloudflare_config_json(origin_port())?;
+    let generated = template.replace(
```
aram356: ⛏ **nitpick** — `template.replace("TRUSTED_SERVER_CONFIG = \"{}\"", …)` silently no-ops if the template placeholder ever drifts, in which case the generated config omits `TRUSTED_SERVER_CONFIG` and the Worker quietly falls back to example config. Consider asserting the replacement changed the string (e.g. `assert_ne!(generated, template)`). Test-only.

aram356 (approved):
## Re-review (3rd pass) — head `37718b7`

Everything actionable from the two prior review rounds is resolved, and the new work is well-tested. CI is green across all adapter jobs (fastly, cloudflare, spin, axum, cross-adapter parity).

**Prior nits addressed:** unused `edgezero-adapter` dep removed; the no-op `Result` in `body_as_reader` fixed in **both** `proxy.rs` and `publisher.rs`; `SECRET_FIELDS = &[]` documented with a Phase-1 rationale; the `x-ts-entry-point` header removed.

**New work reviewed:**
- **Chunk-pointer hardening** (`settings_data.rs`): validates the ≤8000-byte Fastly entry limit, per-chunk lengths, and a `checked_add` sum against `envelope_len` before allocating, with a running overflow guard and a new mismatch test. Switched to `EnvConfig` consistently.
- **Cloudflare runtime config loading**: clean `cfg(wasm32)` `OnceLock` injection via `env.var("TRUSTED_SERVER_CONFIG")` → `settings_from_config_blob`, with proper `Report` error handling.
- Route-dispatch test coverage, rebase API-drift fixes, and an EdgeZero dep rev bump (still pinned).

Approving. Three non-blocking notes inline: the removed EdgeZero entry-point CI guard (now relying on forced `rollout_pct=100` rather than a positive assertion), the Cloudflare example-config fallback shipping placeholder secrets, and a brittle test-config string substitution.

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests (incl. cross-adapter parity): PASS
- js tests (vitest): PASS
- integration tests (incl. EdgeZero entry point): PASS

prk-Jr (approved):

PR #800, opened 2026-06-22
Add Trusted Server audit command

ChristianPavilonis:
## Summary

- Adds `ts audit` as a Trusted Server-specific page audit command on top of the base CLI.
- Collects rendered page/script evidence with Chrome/Chromium and writes draft audit/config artifacts.
- Adds audit-specific docs, specs, tests, and host dependencies.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-cli/src/audit.rs` | Adds audit orchestration, output planning, artifact writing, and draft config generation. |
| `crates/trusted-server-cli/src/audit/analyzer.rs` | Detects JS assets, first/third-party classification, redirects, titles, and known integrations. |
| `crates/trusted-server-cli/src/audit/browser_collector.rs` | Adds headless Chrome/Chromium collection for DOM scripts and network requests. |
| `crates/trusted-server-cli/src/audit/collector.rs` | Defines audit collection traits/data shapes for testability. |
| `crates/trusted-server-cli/src/args.rs`, `src/run.rs`, `src/lib.rs` | Wires `ts audit` into CLI parsing and dispatch. |
| `Cargo.toml`, `crates/trusted-server-cli/Cargo.toml`, `Cargo.lock` | Adds audit-only dependencies. |
| `README.md`, `docs/guide/cli.md`, `docs/guide/getting-started.md` | Documents audit workflow and Chrome/Chromium prerequisite. |
| `docs/superpowers/...audit...` | Adds audit design and implementation plan artifacts. |

## Closes

No issue provided; this PR is split out from the combined `feature/ts-cli-next` branch.

## Test plan

- [ ] `cargo test --workspace`
- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] Other: `cargo test --package trusted-server-cli --target aarch64-apple-darwin` — 42 passed

## Checklist

- [x] Uses tracing/log macros (not `println!`)
- [x] New code has tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ build-print = "1.0.1"
 bytes = "1.11"
 chacha20poly1305 = "0.10"
 chrono = "0.4.44"
+chromiumoxide = "0.9.1"
 clap = { version = "4", features = ["derive"] }
 config = "0.15.19"
 cookie = "0.18.1"
@@ -73,6 +74,7 @@ mime = "0.3"
 rand = "0.8"
 regex = "1.12.3"
 reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls"] }
+scraper = "0.24.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.149"
 simple_logger = "5"
--- a/crates/trusted-server-cli/src/audit/analyzer.rs
+++ b/crates/trusted-server-cli/src/audit/analyzer.rs
@@ -11,6 +11,24 @@ use crate::error::{report_error, CliResult};
 
 static GTM_REGEX: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r"\bGTM-[A-Z0-9]+\b").expect("should compile GTM regex"));
+static GPT_INLINE_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r"(?i)\b(?:googletag|gpt\.js|googletagservices|securepubads)\b")
+        .expect("should compile GPT inline regex")
+});
+static DIDOMI_INLINE_REGEX: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"(?i)\bdidomi\b").expect("should compile Didomi inline regex"));
+static DATADOME_INLINE_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r"(?i)\bdatadome\b").expect("should compile DataDome inline regex")
+});
+static PERMUTIVE_INLINE_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r"(?i)\bpermutive\b").expect("should compile Permutive inline regex")
+});
+static LOCKR_INLINE_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r"(?i)(?:\blockr\b|\bloc\.kr\b)").expect("should compile Lockr inline regex")
+});
+static PREBID_INLINE_REGEX: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r"(?i)(?:\bprebid\b|\bpbjs\b)").expect("should compile Prebid inline regex")
+});
 
 pub(crate) fn analyze_collected_page(collected: &CollectedPage) -> CliResult<AuditArtifact> {
     let final_url = collected
@@ -22,7 +40,6 @@ pub(crate) fn analyze_collected_page(collected: &CollectedPage) -> CliResult<Aud
 
     let document = Html::parse_document(&collected.html);
     let title_selector = Selector::parse("title").expect("should parse title selector");
-    let script_selector = Selector::parse("script").expect("should parse script selector");
     let derived_title = document
         .select(&title_selector)
         .next()
@@ -46,29 +63,14 @@ pub(crate) fn analyze_collected_page(collected: &CollectedPage) -> CliResult<Aud
         ));
     }
 
-    for element in document.select(&script_selector) {
-        if let Some(src) = element.value().attr("src") {
+    for tag in &collected.script_tags {
+        if let Some(src) = &tag.src {
             if let Ok(asset_url) = final_url.join(src) {
                 let integration = detect_integration_from_url(&asset_url);
                 record_integration(&mut integrations, &integration, asset_url.as_str());
                 insert_asset(&mut assets_by_url, &final_url, &asset_url, integration);
             } else {
                 warnings.push(format!("could not resolve script URL `{src}`"));
-            }
-        } else {
-            let inline_text = element.text().collect::<Vec<_>>().join(" ");
-            for (integration_id, evidence) in detect_integrations_from_inline_script(&inline_text) {
-                integrations.entry(integration_id).or_insert(evidence);
-            }
-        }
-    }
-
-    for tag in &collected.script_tags {
-        if let Some(src) = &tag.src {
-            if let Ok(asset_url) = Url::parse(src) {
-                let integration = detect_integration_from_url(&asset_url);
-                record_integration(&mut integrations, &integration, asset_url.as_str());
-                insert_asset(&mut assets_by_url, &final_url, &asset_url, integration);
             }
         }
 
@@ -167,6 +169,7 @@ pub(crate) fn classify_party(page_url: &Url, asset_url: &Url) -> AssetParty {
 }
 
 fn host_matches(page_host: &str, asset_host: &str) -> bool {
+    // This is an advisory heuristic, not public-suffix-aware eTLD+1 classification.
     asset_host == page_host
         || asset_host
             .strip_suffix(page_host)
@@ -213,9 +216,15 @@ pub(crate) fn detect_integrations_from_inline_script(script: &str) -> Vec<(Strin
         ));
     }
 
-    let lowered = script.to_ascii_lowercase();
-    for integration in ["gpt", "didomi", "datadome", "permutive", "lockr", "prebid"] {
-        if lowered.contains(integration) {
+    for (integration, regex) in [
+        ("gpt", &*GPT_INLINE_REGEX),
+        ("didomi", &*DIDOMI_INLINE_REGEX),
+        ("datadome", &*DATADOME_INLINE_REGEX),
+        ("permutive", &*PERMUTIVE_INLINE_REGEX),
+        ("lockr", &*LOCKR_INLINE_REGEX),
+        ("prebid", &*PREBID_INLINE_REGEX),
+    ] {
+        if regex.is_match(script) {
             matches.push((
                 integration.to_string(),
                 format!("inline script matched `{integration}`"),
@@ -259,16 +268,20 @@ mod tests {
             requested_url: "https://publisher.example/page".to_string(),
             final_url: "https://publisher.example/page".to_string(),
             page_title: Some("Browser Title".to_string()),
-            html: r#"<html><head><title>HTML Title</title><script src="https://www.googletagmanager.com/gtm.js?id=GTM-ABCD123"></script></head></html>"#.to_string(),
-            script_tags: vec![CollectedScriptTag {
-                src: Some("https://securepubads.g.doubleclick.net/tag/js/gpt.js".to_string()),
-                inline_text: None,
-            }],
+            html: r#"<html><head><title>HTML Title</title></head></html>"#.to_string(),
+            script_tags: vec![
+                CollectedScriptTag {
+                    src: Some("https://www.googletagmanager.com/gtm.js?id=GTM-ABCD123".to_string()),
+                    inline_text: None,
+                },
+                CollectedScriptTag {
+                    src: Some("https://securepubads.g.doubleclick.net/tag/js/gpt.js".to_string()),
+                    inline_text: None,
+                },
+            ],
             network_requests: vec![CollectedRequest {
                 url: "https://cdn.example.com/dynamic.js".to_string(),
-                method: "GET".to_string(),
                 resource_type: Some("Script".to_string()),
-                status: Some(200),
             }],
             warnings: vec!["partial settle".to_string()],
         };
@@ -345,9 +358,7 @@ mod tests {
             }],
             network_requests: vec![CollectedRequest {
                 url: "https://cdn.example.com/prebid.js".to_string(),
-                method: "GET".to_string(),
                 resource_type: Some("script".to_string()),
-                status: Some(200),
             }],
             warnings: Vec::new(),
         };
@@ -371,8 +382,17 @@ mod tests {
             requested_url: "https://publisher.example/page".to_string(),
             final_url: "https://publisher.example/path/page".to_string(),
             page_title: None,
-            html: r#"<html><head><script src="/static/app.js"></script><script src="http://[invalid"></script></head></html>"#.to_string(),
-            script_tags: Vec::new(),
+            html: "<html><head></head></html>".to_string(),
+            script_tags: vec![
+                CollectedScriptTag {
+                    src: Some("/static/app.js".to_string()),
+                    inline_text: None,
+                },
+                CollectedScriptTag {
+                    src: Some("http://[invalid".to_string()),
+                    inline_text: None,
+                },
+            ],
             network_requests: Vec::new(),
             warnings: Vec::new(),
         };
@@ -461,6 +481,22 @@ mod tests {
         assert!(matches
             .iter()
             .any(|(integration, _)| integration == "didomi"));
+    }
+
+    #[test]
+    fn detect_integrations_from_inline_script_avoids_short_substring_matches() {
+        let matches = detect_integrations_from_inline_script("const svgptimize = blockrResult;");
+
+        assert!(
+            !matches.iter().any(|(integration, _)| integration == "gpt"),
+            "should not match incidental GPT substrings"
+        );
+        assert!(
+            !matches
+                .iter()
+                .any(|(integration, _)| integration == "lockr"),
+            "should not match lockr inside a larger token"
+        );
     }
 
     #[test]
--- a/crates/trusted-server-cli/src/audit/analyzer.rs
+++ b/crates/trusted-server-cli/src/audit/analyzer.rs
@@ -1,0 +1,547 @@
+use std::collections::BTreeMap;
+use std::sync::LazyLock;
+
+use regex::Regex;
+use scraper::{Html, Selector};
+use url::Url;
+
+use crate::audit::collector::CollectedPage;
+use crate::audit::{AssetParty, AuditArtifact, AuditedAsset, DetectedIntegration};
+use crate::error::{report_error, CliResult};
+
+static GTM_REGEX: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"\bGTM-[A-Z0-9]+\b").expect("should compile GTM regex"));
+
+pub(crate) fn analyze_collected_page(collected: &CollectedPage) -> CliResult<AuditArtifact> {
+    let final_url = collected
+        .final_url()
+        .map_err(|error| report_error(format!("invalid final URL: {error}")))?;
+    let requested_url = collected
+        .requested_url()
+        .map_err(|error| report_error(format!("invalid requested URL: {error}")))?;
+
+    let document = Html::parse_document(&collected.html);
+    let title_selector = Selector::parse("title").expect("should parse title selector");
+    let script_selector = Selector::parse("script").expect("should parse script selector");
+    let derived_title = document
+        .select(&title_selector)
+        .next()
+        .map(|element| {
+            element
+                .text()
+                .collect::<Vec<_>>()
+                .join(" ")
+                .trim()
+                .to_string()
+        })
+        .filter(|title| !title.is_empty());
+
+    let mut assets_by_url = BTreeMap::<String, AuditedAsset>::new();
+    let mut integrations = BTreeMap::<String, String>::new();
+    let mut warnings = collected.warnings.clone();
+
+    if requested_url != final_url {
+        warnings.push(format!(
+            "page redirected from `{requested_url}` to `{final_url}`"
+        ));
+    }
+
+    for element in document.select(&script_selector) {
+        if let Some(src) = element.value().attr("src") {
+            if let Ok(asset_url) = final_url.join(src) {
+                let integration = detect_integration_from_url(&asset_url);
+                record_integration(&mut integrations, &integration, asset_url.as_str());
+                insert_asset(&mut assets_by_url, &final_url, &asset_url, integration);
+            } else {
+                warnings.push(format!("could not resolve script URL `{src}`"));
+            }
+        } else {
+            let inline_text = element.text().collect::<Vec<_>>().join(" ");
+            for (integration_id, evidence) in detect_integrations_from_inline_script(&inline_text) {
+                integrations.entry(integration_id).or_insert(evidence);
+            }
+        }
+    }
+
+    for tag in &collected.script_tags {
+        if let Some(src) = &tag.src {
+            if let Ok(asset_url) = Url::parse(src) {
+                let integration = detect_integration_from_url(&asset_url);
+                record_integration(&mut integrations, &integration, asset_url.as_str());
+                insert_asset(&mut assets_by_url, &final_url, &asset_url, integration);
+            }
+        }
+
+        if let Some(inline_text) = &tag.inline_text {
+            for (integration_id, evidence) in detect_integrations_from_inline_script(inline_text) {
+                integrations.entry(integration_id).or_insert(evidence);
+            }
+        }
+    }
+
+    for request in &collected.network_requests {
+        let is_script = request
+            .resource_type
+            .as_deref()
+            .is_some_and(|resource_type| resource_type.eq_ignore_ascii_case("script"));
+        if !is_script {
+            continue;
+        }
+        if let Ok(asset_url) = Url::parse(&request.url) {
+            let integration = detect_integration_from_url(&asset_url);
+            record_integration(&mut integrations, &integration, asset_url.as_str());
+            insert_asset(&mut assets_by_url, &final_url, &asset_url, integration);
+        }
+    }
+
+    let assets = assets_by_url.into_values().collect::<Vec<_>>();
+    let third_party_asset_count = assets
+        .iter()
+        .filter(|asset| asset.party == AssetParty::ThirdParty)
+        .count();
+
+    let page_title = collected
+        .page_title
+        .as_deref()
+        .map(str::trim)
+        .filter(|title| !title.is_empty())
+        .map(ToOwned::to_owned)
+        .or(derived_title);
+
+    Ok(AuditArtifact {
+        audited_url: final_url.to_string(),
+        page_title,
+        js_asset_count: assets.len(),
+        third_party_asset_count,
+        detected_integrations: integrations
+            .into_iter()
+            .map(|(id, evidence)| DetectedIntegration { id, evidence })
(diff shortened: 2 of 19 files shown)


Cargo.toml
```diff
@@ -59,6 +60,7 @@ matchit = "0.9"
 mime = "0.3"
 rand = "0.8"
 regex = "1.12.3"
+scraper = "0.24.0"
```
prk-Jr, line 77: 🔧 **wrench** — This `scraper = "0.24.0"` (plus `chromiumoxide`) is the direct cause of the red `prepare integration artifacts` CI job. The workspace now resolves markup5ever 0.35 / selectors 0.31 / match_token 0.35, but `crates/trusted-server-integration-tests` pins `scraper = "0.21"` (markup5ever 0.14 / selectors 0.26), so the transitive-parity check fails (also flags uuid, web-sys, wasm-bindgen-futures).

Must be green before merge. Options:
- **Preferred:** align scraper across the workspace and integration-tests (bump integration-tests to `0.24`, or pin audit to `0.21`) so the trees converge.
- If the split is intentional: add `markup5ever`, `match_token`, `scraper`, `selectors` to `transitive_parity_allowlist` in `scripts/check-integration-dependency-versions.sh`, and run the suggested `cargo update --manifest-path crates/trusted-server-integration-tests/Cargo.toml -p <crate> --precise <v>` for the patch-level drifts (uuid, web-sys, wasm-bindgen-futures).

crates/trusted-server-cli/src/audit/analyzer.rs
```diff
@@ -0,0 +1,547 @@
...
+
+    if let Some(container_id) = GTM_REGEX.find(script) {
+        matches.push((
+            "google_tag_manager".to_string(),
+            container_id.as_str().to_string(),
+        ));
+    }
+
+    let lowered = script.to_ascii_lowercase();
+    for integration in ["gpt", "didomi", "datadome", "permutive", "lockr", "prebid"] {
```
prk-Jr: 🤔 **thinking** — Lowercased substring matching is false-positive prone, especially `"gpt"` (3 chars) which can appear in unrelated script text. Hits for `gpt`/`didomi`/`datadome` later auto-set `enabled = true` in the draft config (`audit.rs build_draft_config`). Blast radius is low (the draft is reviewed and must pass `ts config validate`), but a word-boundary or more specific marker per integration would cut noise.

crates/trusted-server-cli/src/audit/analyzer.rs
```diff
@@ -0,0 +1,547 @@
...
+    let asset_host = asset_url.host_str().unwrap_or_default();
+
+    if host_matches(page_host, asset_host) {
+        AssetParty::FirstParty
+    } else {
+        AssetParty::ThirdParty
+    }
+}
+
+fn host_matches(page_host: &str, asset_host: &str) -> bool {
```
prk-Jr, line 171: 🤔 **thinking** — The `page_host.strip_suffix(asset_host)` direction classifies assets served from a parent or public-suffix host as first-party — e.g. page `foo.co.uk` with an asset on `co.uk` resolves to FirstParty. Acceptable as a heuristic for an advisory tool, but a public-suffix-aware check (psl/publicsuffix) would avoid treating eTLD/parent domains as same-party.

crates/trusted-server-cli/src/audit/browser_collector.rs
```diff
@@ -0,0 +1,353 @@
...
+
+async fn collect_page_from_browser(
+    browser: &mut Browser,
+    target_url: &Url,
+) -> CliResult<CollectedPage> {
+    let page = browser.new_page("about:blank").await.map_err(|error| {
+        report_error(format!("failed to create browser page for audit: {error}"))
+    })?;
+
+    page.goto(target_url.as_str())
```
prk-Jr: 🤔 **thinking** — `wait_for_page_settle` caps at 6s, but `goto` / `wait_for_navigation_response` (and `browser.close()`) have no timeout. A slow or hanging origin can stall `ts audit` indefinitely. Consider wrapping navigation in `tokio::time::timeout` so the command fails cleanly with a partial-result warning instead of hanging.

crates/trusted-server-cli/src/error.rs
```diff
@@ -0,0 +1,9 @@
+pub(crate) type CliResult<T> = Result<T, String>;
+
+pub(crate) fn cli_error<T>(message: impl Into<String>) -> CliResult<T> {
+    Err(message.into())
+}
+
+pub(crate) fn report_error(message: impl Into<String>) -> String {
```
prk-Jr, line 7: ♻️ **refactor** — `report_error` returns its input unchanged with no logging or other side effect, so the name overpromises and it overlaps `cli_error`. Either give it real behavior (e.g. `log::error!` before returning) or drop it and construct the `String` / use `cli_error` directly.

crates/trusted-server-cli/src/audit/analyzer.rs
```diff
@@ -0,0 +1,547 @@
...
+            let inline_text = element.text().collect::<Vec<_>>().join(" ");
+            for (integration_id, evidence) in detect_integrations_from_inline_script(&inline_text) {
+                integrations.entry(integration_id).or_insert(evidence);
+            }
+        }
+    }
+
+    for tag in &collected.script_tags {
+        if let Some(src) = &tag.src {
+            if let Ok(asset_url) = Url::parse(src) {
```
prk-Jr: ⛏ **nitpick** — DOM scripts above use `final_url.join(src)` (relative-aware) while collector `script_tags` here use `Url::parse(src)` (absolute-only). A relative `src` coming from `script_tags` is silently dropped with no warning, unlike the DOM path which warns on unresolvable URLs. Consider `final_url.join` here too for symmetry.

prk-Jr (changes requested):
## Summary

Adds `ts audit`: loads a public page in a fresh headless Chrome/Chromium session, inventories rendered JS assets, detects known integrations, and writes a JS-asset report plus a draft `trusted-server.toml`. Clean `AuditCollector` trait makes the analysis browser-free and well-tested. One blocking item: the new `scraper` dependency breaks the integration dependency-parity CI gate.

## Blocking

### 🔧 wrench

- **Dependency-parity CI failure** — `scraper = "0.24.0"` added to the workspace conflicts with integration-tests' pinned `scraper = "0.21"`; the `prepare integration artifacts` job fails on markup5ever / match_token / scraper / selectors (plus uuid, web-sys, wasm-bindgen-futures). Align the versions or extend the allowlist — see inline on `Cargo.toml`.

## Non-blocking

### 🤔 thinking

- **Inline substring integration detection is false-positive prone** — `analyzer.rs:217`; auto-enables gpt/didomi/datadome in the draft.
- **First-party classifier over-matches parent/eTLD domains** — `analyzer.rs:169`.
- **No overall navigation timeout** — `browser_collector.rs:98`; a hanging origin stalls `ts audit`.

### ♻️ refactor

- **`report_error` is a misleading no-op wrapper** — `error.rs:7`; overlaps `cli_error`.

### ⛏ nitpick

- **Asymmetric URL resolution** — `analyzer.rs:68`; relative `src` from collector script tags is silently dropped.

### 👍 praise

- `AuditCollector` trait → browser-free unit tests via `FakeCollector`; strong testability.
- Browser hygiene: fresh temp user-data-dir per run, `close()` always called, handler task aborted on close failure, no forced `--no-sandbox`.
- `parse_audit_url` restricts to http/https (blocks file://, data:, chrome://), with a test.
- Overwrite protection + a pre-collection conflict check, so a refusal doesn't even launch Chrome.
- GTM container id constrained by `GTM-[A-Z0-9]+` → no TOML injection from page content into the draft.
- audit module and all host-only deps correctly gated behind `cfg(not(target_arch = "wasm32"))`.

## CI Status

- prepare integration artifacts (dependency parity): FAIL (caused by this PR)
- integration / edgezero / browser tests: SKIPPED (blocked by the failed prepare job)
- fmt / clippy / cargo test / vitest: not run — those workflows trigger only on PRs targeting `main`, and this PR targets `feature/ts-cli-base`. Author reports `cargo test --package trusted-server-cli` passing (42 tests) locally.

Cargo.toml
```diff
@@ -31,6 +31,7 @@ build-print = "1.0.1"
 bytes = "1.11"
 chacha20poly1305 = "0.10"
 chrono = "0.4.44"
+chromiumoxide = "0.9.1"
```
aram356, line 46: 🔧 **wrench — PR breaks CI: integration-tests dependency drift**

Adding `chromiumoxide` (here) and `scraper` (line 63) bumped/added `cssparser, ego-tree, html5ever, markup5ever, selectors` and transitively `wasm-bindgen-futures` (0.4.75→0.4.76) and `web-sys` in the root `Cargo.lock`. The separate `crates/trusted-server-integration-tests` lockfile was not updated, so `scripts/check-integration-dependency-versions.sh` reports parity drift and the **prepare integration artifacts** job exits 1.

**Fix** — apply the commands the CI log prints, then commit the integration-tests lockfile:
```
cargo update --manifest-path crates/trusted-server-integration-tests/Cargo.toml -p wasm-bindgen-futures --precise 0.4.76
cargo update --manifest-path crates/trusted-server-integration-tests/Cargo.toml -p web-sys --precise 0.3.103
# (+ anyhow, cssparser, ego-tree, html5ever, markup5ever, match_token, scraper, selectors, uuid as listed in the job log)
```

crates/trusted-server-cli/src/audit/analyzer.rs
```diff
@@ -0,0 +1,547 @@
...
+
+fn host_matches(page_host: &str, asset_host: &str) -> bool {
+    asset_host == page_host
+        || asset_host
+            .strip_suffix(page_host)
+            .is_some_and(|prefix| prefix.ends_with('.'))
+        || page_host
+            .strip_suffix(asset_host)
+            .is_some_and(|prefix| prefix.ends_with('.'))
+}
```
aram356, line 180: 🤔 **thinking — first/third-party classification ignores the public suffix list**

`host_matches` treats any host-suffix relationship as first-party, so `https://example/` is classified first-party to `publisher.example` (asserted at line 435), and e.g. `pub.github.io` vs `github.io` would also be first-party. This feeds `third_party_asset_count`, which a publisher may read as a real signal. An eTLD+1 comparison (`publicsuffix`/`psl`) would be accurate; at minimum a comment noting this is a heuristic would help.

crates/trusted-server-cli/src/audit/analyzer.rs
```diff
@@ -0,0 +1,547 @@
...
+
+    let lowered = script.to_ascii_lowercase();
+    for integration in ["gpt", "didomi", "datadome", "permutive", "lockr", "prebid"] {
+        if lowered.contains(integration) {
+            matches.push((
+                integration.to_string(),
+                format!("inline script matched `{integration}`"),
+            ));
+        }
+    }
```
aram356, line 233: 🤔 **thinking — loose substring detection can auto-enable integrations**

`lowered.contains("gpt")` / `"didomi"` / `"datadome"` on raw inline JS later flips `enabled = true` in the draft config (`audit.rs:282-291`). `"gpt"` is a 3-char substring that appears incidentally in minified code, and `"lockr"` is a substring of `"blockr…"`. The draft is human-reviewed, which mitigates it, but tighter markers (word boundaries, or known globals like `googletag`/`Didomi`) would cut false positives that auto-enable a module.

crates/trusted-server-cli/src/audit/analyzer.rs
```diff
@@ -0,0 +1,547 @@
...
+    let mut integrations = BTreeMap::<String, String>::new();
+    let mut warnings = collected.warnings.clone();
+
+    if requested_url != final_url {
+        warnings.push(format!(
+            "page redirected from `{requested_url}` to `{final_url}`"
+        ));
+    }
+
+    for element in document.select(&script_selector) {
```
aram356: ♻️ **refactor — scripts collected twice via two paths**

The rendered HTML from `page.content()` already contains every `<script>`, and `collected.script_tags` (`document.scripts`) is the same set; both loops run and dedupe by URL. It works, but it's redundant collection — a single source would be simpler.

crates/trusted-server-cli/src/audit/browser_collector.rs
```diff
@@ -0,0 +1,353 @@
...
+            report_error(format!(
+                "failed to read browser performance resource entries: {error}"
+            ))
+        })?
+        .into_value()
+        .map_err(|error| {
+            report_error(format!(
+                "failed to decode browser performance resource data: {error}"
+            ))
+        })?;
```
aram356, line 184: 🤔 **thinking — network capture is bounded and loses status/method**

`performance.getEntriesByType('resource')` caps results at the resource-timing buffer (~250 entries) and exposes no HTTP status, so `CollectedRequest.status` is always `None` and `method` is hardcoded `"GET"`. Large pages can silently drop assets. The CDP Network domain (already available via chromiumoxide) would give complete requests with real status/method.

crates/trusted-server-cli/src/audit/browser_collector.rs
```diff
@@ -0,0 +1,353 @@
...
+
+    if is_successful_navigation_status(response.status) {
+        return Ok(());
+    }
+
+    Err(report_error(format!(
+        "audit request returned HTTP {} {} for `{}`",
+        response.status, response.status_text, response.url
+    )))
+}
```
aram356, line 276: 🤔 **thinking — any 4xx/5xx aborts the whole audit**

This errors out on status ≥ 400, producing no artifacts. Bot-protected publishers (including DataDome — one of the things this tool detects) commonly return 403 to headless Chrome. Consider degrading to a warning + partial artifact instead of a hard failure.

crates/trusted-server-cli/src/audit/browser_collector.rs
```diff
@@ -0,0 +1,353 @@
...
+            }
+        }
+    });
+
+    let result = collect_page_from_browser(&mut browser, target_url).await;
+
+    let close_result = browser
+        .close()
+        .await
+        .map_err(|error| report_error(format!("failed to close browser after audit: {error}")));
```
aram356: 🌱 **seedling — no timeout around `browser.close()`**

An unresponsive browser could hang the command here. A bounded `tokio::time::timeout` would harden the teardown path.

crates/trusted-server-cli/src/audit/collector.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub(crate) inline_text: Option<String>,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub(crate) struct CollectedRequest {
+    pub(crate) url: String,
+    pub(crate) method: String,
+    pub(crate) resource_type: Option<String>,
+    pub(crate) status: Option<u16>,
+}
```
aram356, line 31: ♻️ **refactor — `method`/`status` are dead fields**

`method` is always `"GET"` and `status` always `None` (set in `browser_collector.rs:179-187`), and the analyzer reads neither. Drop them, or populate them via CDP (ties into the network-capture note).

crates/trusted-server-cli/src/error.rs
aram356, line 7: ⛏ **nitpick — `report_error` vs `cli_error` overlap**

Both just `.into()` a message (`report_error` → `String`, `cli_error` → `Err`). The name `report_error` implies logging it doesn't do. Consider collapsing to one helper or renaming to something like `error_message`.

aram356 (changes requested):
## Summary

`ts audit` is a well-structured, genuinely well-tested addition: a headless-Chrome page auditor that emits a `js-assets.toml` inventory and a draft `trusted-server.toml`. The `AuditCollector` trait makes the orchestration unit-testable without a browser, WASM gating is clean, and overwrite/`--force` safety is careful. One blocking issue: the new dependencies break CI. The remaining notes are heuristic-quality and design observations.

Local verification on `aarch64-apple-darwin`: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` (37 passed) all pass for `trusted-server-cli`.

## Blocking

### 🔧 wrench

- **PR breaks CI — integration-tests dependency drift**: new `scraper`/`chromiumoxide` deps bumped the root `Cargo.lock` but the separate `crates/trusted-server-integration-tests` lockfile was not updated, failing the parity check. See inline comment on `Cargo.toml:34`.

## Non-blocking

### 🤔 thinking

- **Public-suffix-unaware party classification** (`analyzer.rs:169-177`): suffix matching treats `example` as first-party to `publisher.example`.
- **Loose substring integration detection** (`analyzer.rs:216-224` → `audit.rs:282-291`): incidental substrings like `gpt` can auto-enable a module in the draft config.
- **Bounded network capture** (`browser_collector.rs:147-165`): resource-timing buffer cap (~250) and no HTTP status; large pages may drop assets.
- **Hard-fail on 4xx/5xx** (`browser_collector.rs:247-255`): bot-protected pages (e.g. DataDome 403 to headless Chrome) yield no artifacts.

### ♻️ refactor

- **Dead `method`/`status` fields** (`collector.rs:27-33`): always `"GET"`/`None`, never read.
- **Redundant script collection** (`analyzer.rs:49-95`): parsed HTML and `document.scripts` are the same set, deduped by URL.

### 🌱 seedling

- **No timeout on `browser.close()`** (`browser_collector.rs:75-78`): could hang on an unresponsive browser.
- **Heavy dependency tree**: `chromiumoxide` pulls async-tungstenite/rustls/reqwest (+458 lockfile lines). Acceptable for a host-only operator CLI (no WASM/runtime impact); good that the fetcher feature is off so there's no Chromium auto-download.

### ⛏ nitpick

- **`report_error`/`cli_error` overlap** (`error.rs:7-9`): both just `.into()`; `report_error` doesn't actually report.

### 📝 note

- ~1.9k lines of design/plan docs added under `docs/superpowers/`. Scanned both files — only example/known-vendor domains, no secrets. Flagging the volume only.

### 👍 praise

- The `AuditCollector` trait + `FakeCollector` seam and the thorough analyzer/orchestration tests (redirect warnings, dedup, relative-URL resolution, no-output guard, and collect-only-after-overwrite-check ordering) are excellent. Clean `#[cfg(not(target_arch = "wasm32"))]` gating throughout.

## CI Status

- prepare integration artifacts: **FAIL** (dependency drift — see blocking)
- fmt (trusted-server-cli): PASS (local)
- clippy (trusted-server-cli): PASS (local)
- rust tests (trusted-server-cli): PASS (local, 37 passed)

Cargo.toml
ChristianPavilonis: Fixed in dfa8a298: bumped integration-tests scraper to 0.24.0, refreshed its lockfile, and verified ./scripts/check-integration-dependency-versions.sh passes.

crates/trusted-server-cli/src/audit/analyzer.rs
ChristianPavilonis: Fixed in dfa8a298: inline detection now uses explicit regex markers instead of loose lowercase substring matching, with a regression test for incidental gpt/lockr substrings.

crates/trusted-server-cli/src/audit/analyzer.rs
ChristianPavilonis: Addressed in dfa8a298 with documentation only per the follow-up decision: host matching is now explicitly labeled as an advisory heuristic, not public-suffix-aware eTLD+1 classification.

crates/trusted-server-cli/src/audit/browser_collector.rs
ChristianPavilonis: Fixed in dfa8a298: wrapped goto and wait_for_navigation_response in tokio::time::timeout, and also bounded browser.close().

crates/trusted-server-cli/src/error.rs
ChristianPavilonis: Fixed in dfa8a298: report_error now logs with log::error! before returning the message.

crates/trusted-server-cli/src/audit/analyzer.rs
ChristianPavilonis: Fixed in dfa8a298: collected script tags now resolve src values with final_url.join(src), preserving relative URL support and warnings for malformed src values.

Cargo.toml
ChristianPavilonis: Fixed in dfa8a298: aligned integration-tests scraper to 0.24.0, updated the integration-tests lockfile including the listed transitive drifts, and verified ./scripts/check-integration-dependency-versions.sh passes.

crates/trusted-server-cli/src/audit/analyzer.rs
ChristianPavilonis: Addressed in dfa8a298 with documentation only per the follow-up decision: the host relationship check now explicitly notes it is an advisory heuristic and not public-suffix-aware eTLD+1 classification.

crates/trusted-server-cli/src/audit/analyzer.rs
ChristianPavilonis: Fixed in dfa8a298: replaced loose inline contains checks with tighter regex markers/known globals and added a regression test covering incidental gpt/lockr substrings.

crates/trusted-server-cli/src/audit/analyzer.rs
ChristianPavilonis: Fixed in dfa8a298: analyzer now uses collected document.scripts data as the single script source; rendered HTML parsing remains only for title fallback.

crates/trusted-server-cli/src/audit/browser_collector.rs
ChristianPavilonis: Partially addressed in dfa8a298: dropped the dead method/status fields and added a warning when the resource timing buffer reaches the default cap. Full CDP Network capture is left out of this PR to keep the change bounded.

crates/trusted-server-cli/src/audit/browser_collector.rs
ChristianPavilonis: Fixed in dfa8a298: non-2xx/3xx main document status now becomes a warning with partial artifacts instead of aborting the entire audit.

crates/trusted-server-cli/src/audit/browser_collector.rs
ChristianPavilonis: Fixed in dfa8a298: browser.close() is now wrapped in a bounded tokio::time::timeout.

crates/trusted-server-cli/src/audit/collector.rs
ChristianPavilonis: Fixed in dfa8a298: removed CollectedRequest.method and status since they were not populated or read.

crates/trusted-server-cli/src/error.rs
ChristianPavilonis: Fixed in dfa8a298: report_error now performs the behavior its name implies by logging with log::error! before returning the message.

aram356 (approved):
## Re-review — all prior findings addressed

Re-reviewed at `20f867451` ("Address audit review feedback"). Every item from my earlier `CHANGES_REQUESTED` is resolved, and all CI checks are now green (including **prepare integration artifacts**, previously failing).

### Prior findings — status

| Finding | Status |
| ------- | ------ |
| 🔧 CI: integration-tests dependency drift | **Fixed** — `crates/trusted-server-integration-tests` `Cargo.lock`/`Cargo.toml` updated; CI green |
| 🤔 Loose substring detection auto-enabling modules | **Fixed** — word-boundary regexes replace `contains()`; new `..._avoids_short_substring_matches` test |
| 🤔 4xx/5xx hard-aborts the audit | **Fixed** — `validate_navigation_response` now returns `Option<String>`; HTTP errors warn and still write partial artifacts |
| 🤔 Bounded resource-timing capture (silent drop) | **Mitigated** — warns when entries reach the 250-entry buffer threshold |
| 🤔 Public-suffix-unaware party classification | **Acknowledged** — documented as an advisory heuristic |
| ♻️ Dead `method`/`status` fields | **Fixed** — removed |
| ♻️ Redundant two-path script collection | **Fixed** — consolidated to the single `script_tags` source |
| 🌱 No timeout on `browser.close()` | **Fixed** — wrapped in `timeout(...)`; navigation timeouts added too |
| ⛏ `report_error` didn't report | **Fixed** — now `log::error!`s the message |

Local verification on `aarch64-apple-darwin`: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` (38 passed) all pass for `trusted-server-cli`. Correctness spot-checks: `\bgoogletag\b` correctly does not misfire on `googletagmanager`; the `timeout(...).await.map_err(..)?.map_err(..)?` unwrapping is correct; dropping the HTML `<script>` parse is safe because browser `document.scripts` `.src` values are already absolute (relative-URL and invalid-URL tests still pass).

## Non-blocking notes

### 📝 note

- The new warning paths — `validate_navigation_response` returning `Some(..)` and the resource-timing-buffer warning — have no dedicated unit test.
- Party classification remains a host-suffix heuristic (now documented as such); an eTLD+1 comparison via a public-suffix list would be a future nicety.

## CI Status

- cargo fmt: PASS
- cargo test: PASS
- vitest: PASS
- integration tests (all adapters): PASS

crates/trusted-server-cli/src/audit/browser_collector.rs
```diff
@@ -0,0 +1,375 @@
...
+
+        previous_count = Some(resource_count);
+        sleep(SETTLE_POLL_INTERVAL).await;
+        elapsed += SETTLE_POLL_INTERVAL;
+    }
+
+    Ok(false)
+}
+
+fn validate_navigation_response(navigation_response: ArcHttpRequest) -> CliResult<Option<String>> {
```
aram356, line 254: 📝 **note** — This new warn-and-continue path (returning `Some(..)` on a 4xx/5xx main document instead of erroring) has no dedicated unit test. The existing `successful_navigation_status_...` test only covers `is_successful_navigation_status`. A small test asserting the `Some(warning)` branch would lock in the behavior. Non-blocking.

crates/trusted-server-cli/src/audit/browser_collector.rs
```diff
@@ -0,0 +1,375 @@
...
+        .into_value()
+        .map_err(|error| {
+            report_error(format!(
+                "failed to decode browser performance resource data: {error}"
+            ))
+        })?;
+
+    if network_requests.len() >= RESOURCE_TIMING_BUFFER_WARNING_THRESHOLD {
+        warnings.push(
+            "browser resource timing buffer reached its default size; some network assets may be missing"
```
aram356: 📝 **note** — The resource-timing-buffer warning (entries ≥ 250) is untested. Worth a small test so a future refactor doesn't silently drop the warning. Non-blocking.

crates/trusted-server-cli/src/audit/analyzer.rs
```diff
@@ -0,0 +1,583 @@
...
+
+    if host_matches(page_host, asset_host) {
+        AssetParty::FirstParty
+    } else {
+        AssetParty::ThirdParty
+    }
+}
+
+fn host_matches(page_host: &str, asset_host: &str) -> bool {
+    // This is an advisory heuristic, not public-suffix-aware eTLD+1 classification.
```
aram356, line 172: 📝 **note** — Good that this is documented as advisory. As a future nicety, an eTLD+1 comparison via a public-suffix list (`psl`) would make first/third-party classification accurate (e.g. avoid treating `example` as first-party to `publisher.example`). Non-blocking.

aram356 (commented):
Follow-up to my approval: anchoring the two non-blocking 📝 notes to their exact lines. Approval stands.

prk-Jr (approved):

PR #819, opened 2026-06-24
Make first-party integration URLs root-relative

### LINKED ISSUE #820: First-party integration URLs leak the routing host instead of being root-relative
## Summary

Several integration attribute rewriters (**GPT, DataDome, Permutive, Lockr, Sourcepoint**) rewrite a publisher's `<script src>` / `<link href>` to an **absolute** first-party URL built from `request_host`:

```rust
format!("{}://{}/integrations/gpt/script", ctx.request_scheme, ctx.request_host)
```

When `request_host` is not the page's own host, the wrong host leaks into the rendered HTML. For example, behind a host-rewriting dev proxy the page renders `https://ts.example.com/integrations/gpt/script` while the address bar shows `www.example.com`.

## Impact

- Integration assets (GPT, DataDome SDK, Permutive, Lockr, Sourcepoint) point at the routing host rather than the production/first-party host.
- Breaks first-party context whenever `request_host` != the page host — e.g. dev-proxy testing against a staging hostname, or any deployment where the program sees the routing host. `X-Forwarded-Host` does not reliably fix it (it can be dropped/overwritten at the edge).

## Expected

These page-facing first-party URLs should be **root-relative** (`/integrations/...`) so the browser resolves them against the page's own host — matching what GTM, Didomi, Testlight, and these integrations' own inline-script rewriters already do.

## Root cause

`{scheme}://{request_host}/integrations/...` in the attribute rewriters:
- `gpt.rs` (`rewrite`)
- `datadome.rs` (`rewrite`)
- `permutive.rs` (`rewrite`)
- `lockr.rs` (`rewrite`)
- `sourcepoint.rs` (`build_first_party_url`)

## Scope

An audit of every integration's page-facing URL emission found these five were the only host leaks. NextJS RSC/flight payload rewriting is intentionally absolute and out of scope; the DataDome protection module's raw-`Host` read is server-side (validation API body), not page-facing.

aram356:
Closes #820

## Problem

The **GPT, DataDome, Permutive, Lockr, and Sourcepoint** attribute rewriters rewrite a publisher's `<script src>` / `<link href>` to an **absolute** first-party URL built from `request_host`:

```rust
format!("{}://{}/integrations/gpt/script", ctx.request_scheme, ctx.request_host)
```

`ctx.request_host` is the host TS derived for the request (`extract_request_host` → `Host`/`X-Forwarded-Host`). When that isn't the page's own host, these URLs leak the wrong host into the HTML — e.g. the page renders `https://ts.example.com/integrations/gpt/script` while the address bar shows `www.example.com`. This shows up behind a host-rewriting dev proxy, and anywhere `request_host` resolves to the routing host rather than the production domain.

## Fix

Emit these URLs **root-relative** (`/integrations/...`) so the **browser** resolves them against the page's own host:

```rust
"/integrations/gpt/script".to_string()
```

Correct both in production and behind a proxy, and it removes the dependency on `request_host` / `X-Forwarded-Host` derivation entirely. This matches what GTM, Didomi, Testlight, and these integrations' own inline-script rewriters already do.

## Scope — full integration audit

Audited **every** integration's page-facing URL emission (attribute rewriters, head injectors, inline-script rewriters, config-rendered URLs):

| Verdict | Integrations |
|---|---|
| **Fixed** (were absolute from `request_host`) | gpt, datadome, permutive, lockr, sourcepoint |
| Already root-relative | google_tag_manager, didomi, testlight, datadome `client_side_tag_url` (default), sourcepoint trap prefix |
| Correctly absolute (real external/upstream origin) | GPT/GTM/Permutive/Sourcepoint upstream fetch targets |
| Intentionally absolute — **left unchanged** | NextJS RSC/flight payload host rewriting |
| Server-side only — not page-facing | DataDome protection module's raw-`Host` read (validation API body) |

These five were the only page-facing first-party URLs built from a host.

## Tests

Unit tests asserting the old absolute form updated to the root-relative form (no weakening). `cargo fmt`, `cargo clippy --workspace --all-targets --all-features -D warnings`, and `cargo test --workspace` all pass.

### DIFF
--- a/crates/trusted-server-core/src/integrations/gpt.rs
+++ b/crates/trusted-server-core/src/integrations/gpt.rs
@@ -435,17 +435,15 @@ impl IntegrationAttributeRewriter for GptIntegration {
         &self,
         _attr_name: &str,
         attr_value: &str,
-        ctx: &IntegrationAttributeContext<'_>,
+        _ctx: &IntegrationAttributeContext<'_>,
     ) -> AttributeRewriteAction {
         if !self.config.rewrite_script {
             return AttributeRewriteAction::keep();
         }
 
         if Self::is_gpt_script_url(attr_value) {
-            AttributeRewriteAction::replace(format!(
-                "{}://{}/integrations/gpt/script",
-                ctx.request_scheme, ctx.request_host
-            ))
+            // Root-relative so the browser resolves it against the page host.
+            AttributeRewriteAction::replace("/integrations/gpt/script".to_string())
         } else {
             AttributeRewriteAction::keep()
         }
@@ -587,8 +585,8 @@ mod tests {
         match result {
             AttributeRewriteAction::Replace(url) => {
                 assert_eq!(
-                    url, "https://edge.example.com/integrations/gpt/script",
-                    "should rewrite to first-party script endpoint"
+                    url, "/integrations/gpt/script",
+                    "should rewrite to root-relative first-party script endpoint"
                 );
             }
             other => panic!("Expected Replace action, got {:?}", other),
--- a/crates/trusted-server-core/src/integrations/gpt.rs
+++ b/crates/trusted-server-core/src/integrations/gpt.rs
@@ -443,6 +443,10 @@ impl IntegrationAttributeRewriter for GptIntegration {
 
         if Self::is_gpt_script_url(attr_value) {
             // Root-relative so the browser resolves it against the page host.
+            // Note: a page-level `<base href>` participates in this resolution,
+            // so on pages that set an external base URL these resolve against
+            // that base rather than the address-bar origin — an accepted
+            // tradeoff, matching GTM/Didomi/Testlight which are also relative.
             AttributeRewriteAction::replace("/integrations/gpt/script".to_string())
         } else {
             AttributeRewriteAction::keep()
--- a/crates/trusted-server-core/src/integrations/datadome.rs
+++ b/crates/trusted-server-core/src/integrations/datadome.rs
@@ -52,7 +52,7 @@
 //! automatically rewrite `DataDome` script URLs in HTML responses:
 //!
 //! - `<script src="https://js.datadome.co/tags.js">` becomes
-//!   `<script src="https://publisher.com/integrations/datadome/tags.js">`
+//!   `<script src="/integrations/datadome/tags.js">`
 //! - Handles both `src` and `href` attributes (for preload/prefetch links)
 
 use std::sync::{Arc, LazyLock};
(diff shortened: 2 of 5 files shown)


crates/trusted-server-core/src/integrations/gpt.rs
```diff
@@ -435,17 +435,15 @@ impl IntegrationAttributeRewriter for GptIntegration {
...
             return AttributeRewriteAction::keep();
         }
 
         if Self::is_gpt_script_url(attr_value) {
-            AttributeRewriteAction::replace(format!(
-                "{}://{}/integrations/gpt/script",
-                ctx.request_scheme, ctx.request_host
-            ))
+            // Root-relative so the browser resolves it against the page host.
+            AttributeRewriteAction::replace("/integrations/gpt/script".to_string())
```
ChristianPavilonis, line 450: Automated review: One compatibility edge case with this cross-cutting change: root-relative attribute values are resolved against the document base URL. If a publisher page contains something like `<base href="https://assets.example/">`, then `<script src="/integrations/gpt/script">` resolves to `https://assets.example/integrations/gpt/script`, not the page's address-bar host. The old absolute form was immune to that, so this can break these integrations on pages that use an external base URL. This applies to the other new root-relative attribute rewrites in this PR too.

Suggested fix: either add coverage for a `<base href>` document and handle/neutralize that case for first-party integration URLs, or explicitly document the limitation if root-relative URLs are the intended tradeoff. Where a base-safe URL is required, emit in a way that uses `window.location.origin` at runtime rather than relying on HTML relative URL resolution.

ChristianPavilonis (commented):
Automated review: I reviewed the root-relative integration URL changes for GPT, DataDome, Lockr, Permutive, and Sourcepoint. I did not find any blocking correctness, security, data-loss, authorization, or severe compatibility issues. CI is passing. I left one non-blocking compatibility comment about `<base href>` semantics because root-relative attribute URLs are still resolved against the document base URL, not always the address-bar origin.

ChristianPavilonis (approved):
## Review Summary

Approved. I reviewed the root-relative first-party integration URL change and did not find any blocker/high/medium correctness, security, or compatibility issues beyond the already-noted `<base href>` edge case.

## Findings

### P3 / Low

- `crates/trusted-server-core/src/integrations/datadome.rs:55` — DataDome module docs still show the old absolute rewrite output (`https://publisher.com/integrations/datadome/tags.js`). Please update the example to the new root-relative form (`/integrations/datadome/tags.js`) when convenient.

prk-Jr (approved):

crates/trusted-server-core/src/integrations/gpt.rs
aram356: Documented the limitation as the intended tradeoff in 38aee5367. Added a note at all five rewrite sites (gpt, datadome, lockr, permutive, sourcepoint) explaining that a page-level `<base href>` participates in resolution, so on pages that set an external base URL these resolve against that base rather than the address-bar origin — an accepted tradeoff that matches GTM/Didomi/Testlight, which are already root-relative.

aram356:
Addressed review feedback in 38aee5367:

- **P3 (datadome.rs:55)** — updated the module doc example from the old absolute `https://publisher.com/integrations/datadome/tags.js` to the root-relative `/integrations/datadome/tags.js`.
- **`<base href>` compatibility note** — documented the tradeoff at all five rewrite sites (gpt, datadome, lockr, permutive, sourcepoint), noting that a page-level `<base href>` participates in URL resolution. This is the intended, lighter option from the review and is consistent with GTM/Didomi/Testlight, which are already root-relative.

Comment/doc-only change; `cargo fmt` and `cargo clippy` pass.

PR #822, opened 2026-06-25
Fix Sourcepoint first-party consent: iframe asset 404s + same-origin message guard

### LINKED ISSUE #821: Sourcepoint privacy-manager iframe assets 404: root-absolute paths not rewritten
## Summary

When the Sourcepoint CMP is served through Trusted Server's first-party proxy,
the **privacy-manager iframe documents** 404 their own assets, so the
privacy-manager UI cannot load.

## Root cause

The privacy-manager iframe pages (e.g. `us_pm/index.html`) reference their
assets with **root-absolute paths**:

```html
<script src="/PrivacyManagerUS.89867.js"></script>
<link  href="/PrivacyManagerUS.b9d1f.css" rel="stylesheet">
<script src="/polyfills.01516.js"></script>
<link  rel="manifest" href="/manifest.json">
```

On `cdn.privacy-mgmt.com` the iframe origin is the CDN, so these resolve
correctly. Served **first-party** through TS, the iframe origin is the
publisher, so `/PrivacyManagerUS.89867.js` resolves to the **publisher root →
404**. (`<base href>` can't help — it doesn't affect root-absolute paths.)

TS's Sourcepoint proxy already rewrites JS response bodies, but did not rewrite
the **HTML** iframe documents' root-absolute asset references.

## Evidence

- `cdn.privacy-mgmt.com/PrivacyManagerUS.89867.js` → 200; proxied
  `…/integrations/sourcepoint/cdn/PrivacyManagerUS.89867.js` → 200.
- In a browser through the proxy the iframe requested `/PrivacyManagerUS.89867.js`
  (publisher root) → 404. The request initiator was the iframe **parser**, not a
  JS public-path — i.e. it's an HTML asset-path issue.

## Fix

Rewrite root-absolute `src`/`href` in proxied Sourcepoint HTML responses to the
`/integrations/sourcepoint/cdn` prefix (leaving protocol-relative / absolute
URLs untouched). See the linked PR. Verified 404 → 200 in-browser.

## Related / not covered here

Fixing these 404s lets the privacy-manager app render, but does **not** by
itself restore page scroll: the **first-layer consent message iframe renders
0×0**, leaving the page behind an effectively-invisible consent overlay. That
0×0 message-iframe behavior has no missing-asset cause and is a separate
first-party-proxy problem (wrapper↔message sizing/postMessage) — to be tracked
separately and likely needs an upstream Sourcepoint change.

aram356:
Closes #821

> **Stacked on #819** — base branch is `fix-integration-relative-urls`. Review/merge after #819 (or rebase onto `main` once #819 lands). (Branch name is legacy — the fix is HTML asset paths + the message-origin guard, not webpack publicPath.)

Makes Sourcepoint's CMP actually work through Trusted Server's first-party proxy: the consent dialog renders and the page scrolls. Two distinct first-party-proxy breakages, both verified in a real browser through the proxy.

## 1. Privacy-manager iframe assets 404

The privacy-manager iframe documents (`us_pm/index.html`) reference assets root-absolute: `<script src="/PrivacyManagerUS.<hash>.js">`, `/polyfills.<hash>.js`, `/manifest.json`. On `cdn.privacy-mgmt.com` those hit the CDN; served first-party the iframe origin is the publisher, so they resolve to the publisher root → **404**, and the privacy-manager UI can't load.

**Fix:** an HTML response rewrite that prefixes root-absolute `src`/`href` with `/integrations/sourcepoint/cdn` (protocol-relative/absolute URLs untouched). Verified **404 → 200**.

## 2. Consent dialog never shows / page stays scroll-locked

Even with assets at 200, the dialog stayed hidden and `<body>` stayed `position:fixed`. The wrapper validates messages from its own iframe with `e.origin === params.msgOrigin || e.origin === params.pmOrigin`, where `msgOrigin` is `baseEndpoint` used verbatim. First-party, `baseEndpoint` is a **path** (`/integrations/sourcepoint/cdn`), so `msgOrigin = https://<publisher>/integrations/sourcepoint/cdn` — which never equals the iframe's **bare** origin `https://<publisher>`. So the wrapper adds `html.sp-message-open` (locks scroll) but never shows the dialog or removes the lock.

Verified dead ends: an **absolute** `baseEndpoint` keeps the path → still no match; a **bare-origin** `baseEndpoint` needs a dedicated subdomain (not available) or claiming colliding site-root paths (`/manifest.json`, `/polyfills.*`).

**Fix:** since the iframe is genuinely **same-origin** when proxied first-party, a third script rewrite lets the guard also accept `e.origin === location.origin`. This only *additionally* trusts a same-origin frame — which already has full page access — so it adds **no attack surface**; it adapts an origin check written for a cross-origin CDN to first-party serving. Anchored on the semantic `.pmOrigin)` close, capturing the minified event identifier; it only adds an OR branch (existing checks untouched).

## Verification (in-browser, through the proxy)

- Assets `PrivacyManagerUS.*` / `polyfills.*` → 200.
- Consent dialog ("PRIVACY AND LEGAL TERMS / Agree & Continue") **renders**.
- Clicking **Agree & Continue** → `sp-message-open` removed → `body` back to `static` → **page scrolls** (verified scrollTop moves, was stuck at 0).

## Tests

`rewrites_root_absolute_asset_paths_in_html`, `html_rewrite_preserves_absolute_and_protocol_relative_urls`, `is_likely_html_path_matches_iframe_documents`, `rewrites_message_origin_guard_to_accept_same_origin`, `message_origin_guard_rewrite_handles_minified_identifiers`, `message_origin_guard_rewrite_leaves_unrelated_origin_checks_untouched`. All 48 Sourcepoint tests pass; fmt/clippy/check clean.

## Note on fragility

Both rewrites pattern-match Sourcepoint's minified JS/HTML (as the pre-existing CDN-URL and `/unified/` rewrites already do), so a Sourcepoint recompile could require updating an anchor. Anchored on semantic tokens (`src`/`href`, `.pmOrigin`) to minimise that.

### DIFF
--- a/crates/trusted-server-core/src/integrations/sourcepoint.rs
+++ b/crates/trusted-server-core/src/integrations/sourcepoint.rs
@@ -285,11 +285,7 @@ impl SourcepointIntegration {
         Ok(target.to_string())
     }
 
-    fn build_first_party_url(
-        &self,
-        source_url: &str,
-        ctx: &IntegrationAttributeContext<'_>,
-    ) -> Option<String> {
+    fn build_first_party_url(&self, source_url: &str) -> Option<String> {
         let parsed = parse_sourcepoint_url(source_url)?;
         if parsed.host_str()? != SOURCEPOINT_CDN_HOST {
             return None;
@@ -301,10 +297,8 @@ impl SourcepointIntegration {
             .map(|value| format!("?{value}"))
             .unwrap_or_default();
 
-        Some(format!(
-            "{}://{}{}{}{}",
-            ctx.request_scheme, ctx.request_host, SOURCEPOINT_CDN_PREFIX, path, query
-        ))
+        // Root-relative so the browser resolves it against the page host.
+        Some(format!("{SOURCEPOINT_CDN_PREFIX}{path}{query}"))
     }
 
     fn copy_headers(
@@ -851,11 +845,11 @@ impl IntegrationAttributeRewriter for SourcepointIntegration {
         &self,
         _attr_name: &str,
         attr_value: &str,
-        ctx: &IntegrationAttributeContext<'_>,
+        _ctx: &IntegrationAttributeContext<'_>,
     ) -> AttributeRewriteAction {
         // `handles_attribute()` already gates on `rewrite_sdk`, so this
         // method is only called when rewriting is enabled.
-        if let Some(rewritten) = self.build_first_party_url(attr_value, ctx) {
+        if let Some(rewritten) = self.build_first_party_url(attr_value) {
             return AttributeRewriteAction::replace(rewritten);
         }
 
@@ -986,7 +980,7 @@ mod tests {
         assert_eq!(
             rewritten,
             AttributeRewriteAction::replace(
-                "https://edge.example.com/integrations/sourcepoint/cdn/mms/v2/get_site_data?account_id=821",
+                "/integrations/sourcepoint/cdn/mms/v2/get_site_data?account_id=821",
             )
         );
     }
--- a/crates/trusted-server-core/src/integrations/sourcepoint.rs
+++ b/crates/trusted-server-core/src/integrations/sourcepoint.rs
@@ -9,6 +9,7 @@
 //! |-------|-----------|-----------------|
 //! | HTML attributes | `IntegrationAttributeRewriter` | Static `<script src>` / `<link href>` tags |
 //! | JS response bodies | `rewrite_script_content` | Webpack chunk paths + hardcoded CDN URLs |
+//! | HTML response bodies | `rewrite_html_content` | Root-absolute `src`/`href` in proxied iframe documents |
 //! | Runtime config | `IntegrationHeadInjector` | `window._sp_` assignments from Next.js chunks |
 //! | Dynamic DOM | TS script guard (`script_guard.ts`) | Script/link elements inserted after page load |
 //!
@@ -109,6 +110,29 @@ static SP_CDN_URL_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
 static SP_ORIGIN_UNIFIED_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(r#"\.origin\s*\+\s*(['"])/unified/"#)
         .expect("Sourcepoint origin+unified regex should compile")
+});
+
+/// Matches a root-absolute `src`/`href` attribute value in a proxied HTML
+/// document (e.g. the privacy-manager iframe `us_pm/index.html`).
+///
+/// These iframe documents reference their assets relative to the CDN root:
+/// ```html
+/// <script src="/PrivacyManagerUS.89867.js"></script>
+/// <link href="/PrivacyManagerUS.b9d1f.css" rel="stylesheet">
+/// ```
+/// On `cdn.privacy-mgmt.com` that resolves to the CDN; served first-party
+/// through Trusted Server the iframe origin is the publisher, so
+/// `/PrivacyManagerUS.<hash>.js` resolves to the publisher root and 404s —
+/// leaving the consent UI unable to render. We prefix these with the CDN proxy
+/// path so they load through `/integrations/sourcepoint/cdn/…`.
+///
+/// Group 1 is the attribute up to the opening quote and leading slash; group 2
+/// is the rest of the path. The `[^/"]` after the leading slash excludes
+/// protocol-relative `//host` URLs (and absolute `https://…` never starts with
+/// `/`), so only root-absolute paths are rewritten.
+static SP_HTML_ROOT_ABSOLUTE_ASSET_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r#"((?:src|href)=")/([^/"][^"]*)""#)
+        .expect("Sourcepoint HTML root-absolute asset regex should compile")
 });
 
 /// Configuration for the Sourcepoint first-party proxy.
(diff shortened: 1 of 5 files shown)


crates/trusted-server-core/src/integrations/sourcepoint.rs
```diff
@@ -111,6 +112,54 @@ static SP_ORIGIN_UNIFIED_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
...
+/// leaving the page rendered-but-unscrollable.
+///
+/// When proxied first-party the message iframe is genuinely **same-origin**, so
+/// we append `|| <event>.origin === location.origin` to the guard. This only
+/// *additionally* trusts a same-origin frame — which already has full access to
+/// the page — so it adds no attack surface; it teaches an origin check written
+/// for a cross-origin CDN about first-party serving. The match is anchored on
+/// the semantic `.pmOrigin)` close of the guard; group 1 is the (possibly
+/// minified) event identifier and group 2 the `…pmOrigin` operand.
+static SP_MESSAGE_ORIGIN_GUARD_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
```
ChristianPavilonis, line 166: **Non-blocking: same-origin guard rewrite may miss the documented/current wrapper guard shape**

This rewrite only matches guards that end in `.pmOrigin)`. The documented Sourcepoint wrapper path (`/wrapperMessagingWithoutDetection.js`) currently has a different handler shape (`if (<event>.origin === <origin-var>)`), where the origin variable is derived from `baseEndpoint`. Under this proxy, that value can be `/integrations/sourcepoint/cdn`, while `MessageEvent.origin` is the bare publisher origin, so the consent-message postMessages can still be rejected for that wrapper.

Suggested fix: add a second narrowly anchored rewrite for the legacy/current `if (<event>.origin === <origin-var>)` handler, ideally anchored near Sourcepoint-specific message-handler tokens, and cover it with a realistic wrapper snippet test.

crates/trusted-server-core/src/integrations/sourcepoint.rs
```diff
@@ -539,27 +639,47 @@ impl SourcepointIntegration {
...
 
     fn rewrite_javascript_response(&self, response: &mut Response<EdgeBody>, rewritten: String) {
+        self.finalize_rewritten_response(
+            response,
+            rewritten,
+            "application/javascript; charset=utf-8",
+        );
+    }
+
+    fn rewrite_html_response(&self, response: &mut Response<EdgeBody>, rewritten: String) {
```
ChristianPavilonis: **Non-blocking: rewritten HTML responses now inherit the JS public-cache policy**

`rewrite_html_response()` uses the shared finalizer, which overwrites non-`Set-Cookie` responses with `Cache-Control: public, max-age=<ttl>`. That policy was appropriate for rewritten versioned JS/static assets, but the new HTML rewrite targets unversioned documents like `/us_pm/index.html`; it also bypasses the existing `forwarded_cookies`-aware `apply_cache_headers` path.

Suggested fix: split JS and HTML finalization/cache policy. For rewritten HTML, preserve upstream `Cache-Control` when present; if absent, apply the existing `apply_cache_headers(response, forwarded_cookies)` behavior. Add tests for preserving `no-store` and for private caching when cookies were forwarded.

ChristianPavilonis (approved):
## Review Summary

Approved. The change is narrowly scoped and the reported CI checks are passing. I left two non-blocking findings inline: both are worth addressing, but I do not think they need to block this PR if you prefer to follow up separately.

prk-Jr (approved):
👍 LGTM

crates/trusted-server-core/src/integrations/sourcepoint.rs
aram356: Addressed in 140ccd66b. Generalized the guard rewrite anchor from `.pmOrigin)` to `.(?:pmOrigin|msgOrigin))`, so it now also matches a standalone `<event>.origin===<obj>.msgOrigin)` guard — the `wrapperMessagingWithoutDetection.js` shape, whose guard compares only against the single `baseEndpoint`-derived `msgOrigin` value (per the PR description, `msgOrigin` is `baseEndpoint` used verbatim). Both `msgOrigin` and `pmOrigin` are Sourcepoint-specific field names, so generic origin checks stay untouched, and the trailing `)` requirement means the inner `msgOrigin` of the combined `…msgOrigin || …pmOrigin)` form (followed by `||`) is not matched a second time. Added `rewrites_standalone_msg_origin_guard_to_accept_same_origin` and `message_origin_guard_rewrite_does_not_double_match_combined_guard` covering both.

Note on scope: if a live wrapper guards on a bare `baseEndpoint` variable rather than a `.msgOrigin`/`.pmOrigin` property access, this anchor will not reach it — matching a bare `<event>.origin===<var>)` would over-match unrelated origin checks in this security-sensitive message-validation code, so I deliberately kept the anchor on the Sourcepoint field names. A captured minified sample of that exact guard would let us extend it safely if needed.

crates/trusted-server-core/src/integrations/sourcepoint.rs
aram356: Addressed in 140ccd66b. Split finalization into `finalize_rewritten_body` (body + content-encoding/length + `Vary` normalization, no cache policy) and per-type cache handling:

- `rewrite_javascript_response` keeps the fixed `public, max-age=<ttl>` policy for the versioned/hashed JS bundles (and stays private on `Set-Cookie`).
- `rewrite_html_response` now takes `forwarded_cookies` and defers to the existing `apply_cache_headers`, which preserves an upstream `Cache-Control` when present and otherwise applies the cookie-aware default (`private, max-age=0` when cookies were forwarded, `private, no-store` on `Set-Cookie`, else the public default).

Tests added: `rewrite_html_response_preserves_upstream_cache_control` (preserves `no-store`), `_uses_private_policy_when_cookies_were_forwarded`, `_uses_public_default_without_forwarded_cookies`, and `_uses_private_no_store_for_cookie_setting_responses`.

PR #624, opened 2026-04-08
Migrate handler layer to HTTP types (PR 12)

prk-Jr:
## Summary

- Move the PR 12 handler boundary to `http` request/response types so core handlers no longer depend on Fastly request/response APIs.
- Keep Fastly isolated to the adapter entry/exit path and the still-unmigrated integration/provider boundary, which keeps PR 13 scope separate.
- Lock the migrated surface with tests and migration guards so later changes do not accidentally reintroduce handler-layer Fastly types.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-adapter-fastly/src/main.rs` | Move the conversion boundary to the adapter entry/exit path and route core handlers with HTTP-native requests and responses. |
| `crates/trusted-server-core/src/auction/endpoints.rs` | Migrate the `/auction` handler to `http` types and rebuild a Fastly request only at the provider-facing `AuctionContext` boundary. |
| `crates/trusted-server-core/src/auction/formats.rs` | Consume HTTP requests directly and return HTTP auction responses. |
| `crates/trusted-server-core/src/auction/orchestrator.rs` | Keep provider parsing Fastly-based while making the provider response conversion explicit in the orchestrator. |
| `crates/trusted-server-core/src/compat.rs` | Add the minimal borrowed HTTP-to-Fastly request bridge needed for the remaining provider boundary. |
| `crates/trusted-server-core/src/geo.rs` | Make response header injection operate on HTTP responses. |
| `crates/trusted-server-core/src/integrations/google_tag_manager.rs` | Adapt the integration boundary so the shared proxy helper can stay HTTP-native while the integration trait remains Fastly-based. |
| `crates/trusted-server-core/src/integrations/gpt.rs` | Adapt the integration boundary so the shared proxy helper can stay HTTP-native while the integration trait remains Fastly-based. |
| `crates/trusted-server-core/src/integrations/testlight.rs` | Adapt the integration boundary so the shared proxy helper can stay HTTP-native while the integration trait remains Fastly-based. |
| `crates/trusted-server-core/src/migration_guards.rs` | Extend the guard to cover the handler modules migrated in PR 12. |
| `crates/trusted-server-core/src/proxy.rs` | Migrate first-party proxy/click/sign/rebuild handlers and response rewriting to HTTP request/response types. |
| `crates/trusted-server-core/src/publisher.rs` | Migrate publisher handlers to HTTP types and use the platform HTTP client for publisher-origin fetches. |
[...]

PR #831, opened 2026-06-30
Derive Lockr API Origin from Referer for first-party proxy

### LINKED ISSUE #832: Integration proxies forward publisher Authorization header to third-party APIs (breaks Lockr)
## Summary

The integration API proxies (`lockr`, `permutive`, `didomi`) forward the publisher site's **`Authorization`** header — e.g. a staging basic-auth credential — to the third-party upstream. This both **leaks the publisher credential** and **breaks the Lockr API**, which rejects any request carrying an unexpected `Authorization`.

## Impact (Lockr)

The `settings` call returns HTTP 200 with a failure body:

```json
{"success":false,"error":{"code":400,"message":"Invalid request"}}
```

The SDK's `getAIMSettings()` only proceeds on `success === true && data`, so init short-circuits: no settings data, no `pvStatus`, **no `page-view`**, and no identity tokens. This only manifests on deployments served behind site basic-auth (so the browser sends `Authorization` to the trusted server, which forwards it onward); a direct (no-auth) page works.

## Diagnosis (verified)

- Direct upstream replay: a fully-correct request (Origin + content-type + body) succeeds; **adding only an `Authorization` header reproduces the exact `400 Invalid request`**.
- Local `fastly compute serve` reproduction (legacy path, lockr enabled) reproduced the `400`; the proxy placed `authorization` on the outbound wire. Removing it cleared the `400`.

Earlier hypotheses (missing Origin, dropped POST body) were ruled out — both were forwarded correctly.

## Fix

Stop forwarding `Authorization` to the upstream in `lockr`, `permutive`, and `didomi` (matching `sourcepoint`, which already omits it intentionally). `datadome`, `prebid`, and `gpt` never forwarded it.

Fixed by #833.

aram356:
Closes #832

## Problem

Behind the first-party proxy, the Lockr `settings` call returns success-shaped HTTP 200 but a failure body:

```json
{"success":false,"error":{"code":400,"message":"Invalid request"}}
```

The SDK's `getAIMSettings()` only proceeds when `success === true && data`, so the entire init chain short-circuits — no settings data, no `pvStatus`, **no `page-view`**, and no identity tokens.

## Root cause

The Lockr identity API (`identity.loc.kr`) requires the publisher's `Origin` header. When a browser calls the API directly it is **cross-origin**, so the browser attaches `Origin` automatically. Behind the first-party proxy the browser calls `/integrations/lockr/api/...` on the **same origin** as the page, so it sends **no `Origin`**.

`copy_request_headers` only set `Origin` from a configured `origin_override` or a pre-existing request `Origin` — neither present in the first-party case — so the proxied request reached the upstream with no `Origin` and was rejected.

Verified by direct replay against the upstream: the same body with no `Origin` is rejected, while supplying the publisher `Origin` gets past that check.

## Fix

In the Lockr API proxy's `copy_request_headers`:

- Synthesize the publisher `Origin` from the page **`Referer`** when the request has no `Origin` and no `origin_override` is configured. The same-origin call still carries a `Referer` (the page URL), whose origin is exactly the value a direct cross-origin call would have sent.
- Forward the `Referer` header upstream so the proxied request matches a direct browser call.

Precedence: `origin_override` → request `Origin` → `Referer`-derived origin.

> Note: deriving the origin from the request `Host` is deliberately avoided — edge host-rewriting can make `Host` differ from the publisher's registered domain. `Referer` reflects the actual page origin.

## Testing

- New unit tests: derive Origin from Referer; prefer an explicit request Origin; prefer `origin_override`; and `origin_from_url` extraction (path/query/fragment stripped, scheme-less rejected).
- `cargo test -p trusted-server-core lockr` — 13 passed.
- `cargo fmt --all -- --check` and `cargo clippy -p trusted-server-core --all-features` — clean.

## Verification note

End-to-end browser verification requires a deployment of this change (the live first-party endpoint runs the deployed build); the fix is covered by unit tests and an upstream replay that isolates `Origin` as the cause.

### DIFF
--- a/crates/trusted-server-core/src/integrations/lockr.rs
+++ b/crates/trusted-server-core/src/integrations/lockr.rs
@@ -255,6 +255,7 @@ impl LockrIntegration {
             header::AUTHORIZATION,
             header::ACCEPT_LANGUAGE,
             header::ACCEPT_ENCODING,
+            header::REFERER,
         ];
 
         for header_name in &headers_to_copy {
@@ -266,13 +267,30 @@ impl LockrIntegration {
         // Always strip consent cookies — consent travels through the OpenRTB body
         self.copy_cookie_header(from, to)?;
 
-        // Use origin override if configured, otherwise forward original
-        let origin = self.config.origin_override.as_deref().or_else(|| {
-            from.get(header::ORIGIN)
-                .and_then(|value| value.to_str().ok())
-        });
+        // Present the publisher's `Origin` to the Lockr API.
+        //
+        // Behind the first-party proxy the browser's call to
+        // `/integrations/lockr/api` is same-origin, so it carries no `Origin`
+        // header — yet the Lockr API rejects requests that lack one. Fall back
+        // to the page origin derived from `Referer` (and finally to a configured
+        // override) so the upstream sees the same `Origin` a direct cross-origin
+        // browser call would have sent.
+        let origin = self
+            .config
+            .origin_override
+            .clone()
+            .or_else(|| {
+                from.get(header::ORIGIN)
+                    .and_then(|value| value.to_str().ok())
+                    .map(str::to_owned)
+            })
+            .or_else(|| {
+                from.get(header::REFERER)
+                    .and_then(|value| value.to_str().ok())
+                    .and_then(origin_from_url)
+            });
         if let Some(origin) = origin {
-            match HeaderValue::from_str(origin) {
+            match HeaderValue::from_str(&origin) {
                 Ok(value) => {
                     to.insert(header::ORIGIN, value);
                 }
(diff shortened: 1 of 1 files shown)


aram356:
Closing — the Origin theory was a misdiagnosis. Verified against the live upstream and the deployed server: the failure is the proxy not forwarding the POST **body** (upstream returns `400 Invalid request` for an empty body; a missing Origin returns `500`, which is not the observed symptom). Sending the deployed server a fully-correct request (valid Origin + content-type + body) still returns `400`, so the body is being dropped. POST-body forwarding is already fixed in main (#624); the deployed build predates it. `origin_override` is unrelated to this failure. Tracking the real cause in #832; regression test in the new PR.

PR #833, opened 2026-06-30
Stop forwarding Authorization header to third-party integration APIs

### LINKED ISSUE #832: Integration proxies forward publisher Authorization header to third-party APIs (breaks Lockr)

aram356:
## Root cause (verified with a local `fastly compute serve` reproduction)

The integration API proxies forward the publisher site's **`Authorization`** header (e.g. a staging basic-auth credential) to the third-party upstream. The Lockr API rejects any request carrying an unexpected `Authorization`:

```json
{"success":false,"error":{"code":400,"message":"Invalid request"}}
```

This short-circuits the SDK's `getAIMSettings()` init (which gates on `success && data`) — **no settings data, no `pvStatus`, no `page-view`, no identity tokens** — on any deployment served behind site auth. It also **leaks the publisher credential** to the third party.

### How it was confirmed

- Direct upstream replay: a fully-correct request (Origin + content-type + body) returns `success`; **adding only an `Authorization` header flips it to the exact `400 Invalid request`**. Missing Origin returns `500`, missing body/content-type returns `400` from a *different* cause — both ruled out.
- Local `fastly compute serve` (legacy path, lockr enabled) reproduced the `400`; debug logging showed the proxy putting `authorization` on the outbound wire. Removing it → upstream no longer returns the `400`.

This corrects two earlier misdiagnoses (Origin — PR #831, closed; POST body — the original framing of this PR). The body and content-type were always forwarded correctly.

## Fix

Stop forwarding `Authorization` in **lockr**, **permutive**, and **didomi**.

Audit of all integration proxies:
| Integration | Forwarded `Authorization`? | Action |
|---|---|---|
| lockr | yes | **stopped** |
| permutive | yes | **stopped** |
| didomi | yes | **stopped** |
| sourcepoint | no (already omitted, with comment) | — |
| datadome, prebid, gpt | no | — |

## Tests

- New `lockr_proxy_forwards_body_and_strips_authorization`: asserts the proxy forwards the POST body + content-type but **strips `Authorization`**. Extends the stub HTTP client to record outgoing request bodies.
- `cargo test -p trusted-server-core` — 1455 passed; `fmt` + `clippy` clean.

Fixes #832

### DIFF
--- a/crates/trusted-server-core/src/integrations/didomi.rs
+++ b/crates/trusted-server-core/src/integrations/didomi.rs
@@ -164,6 +164,10 @@ impl DidomiIntegration {
             );
         }
 
+        // `Authorization` is intentionally NOT forwarded: it carries the
+        // publisher site's own credential (e.g. staging basic-auth), which would
+        // leak to the third-party upstream and can break APIs that reject an
+        // unexpected `Authorization` header.
         for header_name in [
             header::ACCEPT,
             header::ACCEPT_LANGUAGE,
@@ -172,7 +176,6 @@ impl DidomiIntegration {
             header::USER_AGENT,
             header::REFERER,
             header::ORIGIN,
-            header::AUTHORIZATION,
         ] {
             if let Some(value) = original_headers.get(&header_name) {
                 proxy_headers.insert(header_name, value.clone());
--- a/crates/trusted-server-core/src/integrations/didomi.rs
+++ b/crates/trusted-server-core/src/integrations/didomi.rs
@@ -451,6 +451,46 @@ mod tests {
     }
 
     #[test]
+    fn copy_headers_strips_authorization() {
+        // Security regression guard: the publisher's Authorization header must
+        // not be forwarded to the Didomi upstream (credential leak).
+        let integration = DidomiIntegration::new(Arc::new(config(true)));
+        let backend = DidomiBackend::Api;
+        let original_req = http::Request::builder()
+            .method(Method::POST)
+            .uri("https://example.com/test")
+            .header(header::AUTHORIZATION, "Basic dXNlcjpwYXNz")
+            .header(header::USER_AGENT, "test-agent")
+            .body(EdgeBody::empty())
+            .expect("should build original request");
+        let mut proxy_req = http::Request::builder()
+            .method(Method::POST)
+            .uri("https://api.privacy-center.org/test")
+            .body(EdgeBody::empty())
+            .expect("should build proxy request");
+
+        integration.copy_headers(
+            &backend,
+            None,
+            original_req.headers(),
+            proxy_req.headers_mut(),
+        );
+
+        assert!(
+            proxy_req.headers().get(header::AUTHORIZATION).is_none(),
+            "should NOT forward the publisher's Authorization header to Didomi"
+        );
+        assert_eq!(
+            proxy_req
+                .headers()
+                .get(header::USER_AGENT)
+                .and_then(|v| v.to_str().ok()),
+            Some("test-agent"),
+            "should still forward required headers (user-agent)"
+        );
+    }
+
+    #[test]
     fn registers_custom_proxy_path() {
         let mut settings = create_test_settings();
         let custom_config = DidomiIntegrationConfig {
--- a/crates/trusted-server-core/src/integrations/lockr.rs
+++ b/crates/trusted-server-core/src/integrations/lockr.rs
@@ -248,11 +248,14 @@ impl LockrIntegration {
         from: &HeaderMap<HeaderValue>,
         to: &mut HeaderMap<HeaderValue>,
     ) -> Result<(), Report<TrustedServerError>> {
+        // NOTE: `Authorization` is intentionally NOT forwarded. It carries the
+        // publisher site's own credential (e.g. staging basic-auth), which the
+        // Lockr API rejects with `{"code":400,"message":"Invalid request"}` —
+        // and forwarding it would leak the publisher credential to a third party.
         let headers_to_copy = [
             header::CONTENT_TYPE,
             header::ACCEPT,
             header::USER_AGENT,
-            header::AUTHORIZATION,
             header::ACCEPT_LANGUAGE,
             header::ACCEPT_ENCODING,
         ];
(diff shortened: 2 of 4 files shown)


crates/trusted-server-core/src/integrations/lockr.rs
```diff
@@ -248,11 +248,14 @@ impl LockrIntegration {
         from: &HeaderMap<HeaderValue>,
         to: &mut HeaderMap<HeaderValue>,
     ) -> Result<(), Report<TrustedServerError>> {
+        // NOTE: `Authorization` is intentionally NOT forwarded. It carries the
```
ChristianPavilonis: **Lockr still forwards publisher cookies upstream**

This header allowlist now correctly avoids forwarding `Authorization`, but the same proxy path still calls `copy_cookie_header()` below, which forwards the inbound publisher `Cookie` header after removing only consent cookie names. Because `/integrations/lockr/api/...` is first-party, browsers attach publisher cookies automatically; session/auth/staging cookies can still be sent to Lockr, which is the same credential-leak class this PR is addressing.

**Suggestion:** Omit `Cookie` entirely for Lockr if Lockr does not require it, or replace `copy_cookie_header()` with an explicit Lockr cookie allowlist. Do not forward unparseable cookie headers unchanged, and add a regression test with `Cookie: session_id=secret; euconsent-v2=tcf` asserting unrelated publisher cookies do not reach the upstream.

crates/trusted-server-core/src/integrations/permutive.rs
```diff
@@ -257,11 +257,14 @@ impl PermutiveIntegration {
 
     /// Copy relevant request headers for proxying.
     fn copy_request_headers(&self, from: &HeaderMap<HeaderValue>, to: &mut HeaderMap<HeaderValue>) {
+        // `Authorization` is intentionally NOT forwarded: it carries the
```
ChristianPavilonis, line 260: **Add regression coverage for Permutive authorization stripping**

This security-sensitive behavior changed for Permutive too, but the new regression coverage only exercises Lockr. A future allowlist edit could reintroduce the `Authorization` credential leak here without failing tests.

**Suggestion:** Add a focused Permutive proxy test that sends `Authorization` to the first-party route and asserts the recorded upstream request omits it while preserving required headers/body.

crates/trusted-server-core/src/integrations/didomi.rs
```diff
@@ -164,6 +164,10 @@ impl DidomiIntegration {
             );
         }
 
+        // `Authorization` is intentionally NOT forwarded: it carries the
```
ChristianPavilonis, line 167: **Add regression coverage for Didomi authorization stripping**

This security-sensitive behavior changed for Didomi too, but the new regression coverage only exercises Lockr. A future allowlist edit could reintroduce the `Authorization` credential leak here without failing tests.

**Suggestion:** Add a focused Didomi proxy test that sends `Authorization` to the first-party route and asserts the recorded upstream request omits it while preserving required headers/body.

crates/trusted-server-core/src/platform/test_support.rs
```diff
@@ -391,6 +408,18 @@ impl PlatformHttpClient for StubHttpClient {
             .expect("should lock request_headers")
             .push(headers);
 
+        // Capture the outgoing request body so tests can assert it is forwarded.
+        let (_, body) = request.request.into_parts();
+        let body_bytes = body
```
ChristianPavilonis, line 415: **Do not hide request-body capture failures in StubHttpClient**

`into_bytes_bounded(...).await.map(...).unwrap_or_default()` records an empty body if collection fails or exceeds the limit, then still returns the queued response. Tests using `recorded_request_bodies()` cannot distinguish an intentionally empty body from a capture failure.

**Suggestion:** Propagate the collection error as `PlatformError::HttpClient` with context, e.g. `change_context(PlatformError::HttpClient).attach("failed to capture StubHttpClient request body")?`, instead of defaulting to `Vec::new()`.

ChristianPavilonis (approved):
## Summary

Reviewed the Trusted Server PR that strips `Authorization` from Lockr, Permutive, and Didomi integration proxies. The direction looks correct and CI is green. I am approving as requested, with inline notes for a remaining Lockr cookie-forwarding credential-leak risk and follow-up test coverage/test-infrastructure hardening.

crates/trusted-server-core/src/integrations/lockr.rs
aram356: Fixed. Lockr now forwards no `Cookie` header at all — `copy_cookie_header()` is removed (it also forwarded unparseable cookie headers verbatim). Verified the upstream does not need it: the successful direct request carried no `Cookie`, the SDK passes its first-party cookie data in the request body (`firstPartyCookies`), and Permutive/Didomi already forward no cookies. Regression test `lockr_proxy_forwards_body_and_strips_publisher_credentials` sends `Cookie: session_id=secret; euconsent-v2=tcf` and asserts no `Cookie` reaches the upstream.

crates/trusted-server-core/src/integrations/permutive.rs
aram356: Added `permutive_proxy_strips_authorization`: sends `Authorization` through the first-party route and asserts the recorded upstream request omits it while still forwarding `user-agent`.

crates/trusted-server-core/src/integrations/didomi.rs
aram356: Added `copy_headers_strips_authorization`: passes `Authorization` into `copy_headers` and asserts it is not present on the proxy request while `user-agent` is preserved.

crates/trusted-server-core/src/platform/test_support.rs
aram356: Fixed. Replaced `unwrap_or_default()` with `change_context(PlatformError::HttpClient).attach("failed to capture StubHttpClient request body")?`, so a capture failure now surfaces as a `send` error instead of being recorded as an empty body.

prk-Jr (approved):

Issue #790, opened 2026-06-21
Select a scenario configuration at startup, with a ready-to-run development.toml


jwrosewell:
**User story**

As a developer on a standard machine, I want to select a ready-to-run configuration at startup, so that I can run the native dev server without an edge compute provider or build-time config edits.

**Description**

Application settings are baked into the binary at build time from `trusted-server.toml`. Add a native runtime override. When the `TRUSTED_SERVER_CONFIG` environment variable names a config file, a native build reads and validates it at startup instead of the baked configuration. The Fastly and Cloudflare wasm targets cannot read files at runtime and keep the build-time configuration. Add a committed `development.toml` for a standard developer machine. It is vanilla, with no edge compute provider and real local-development secrets, so the dev server runs as-is, unlike `trusted-server.toml` which ships placeholder secrets that must be overridden.

**Done when**

- On native builds, `TRUSTED_SERVER_CONFIG` selects a config file that is read and validated at startup, failing loudly on a bad path.
- The wasm targets keep the build-time configuration.
- A committed `development.toml` parses, validates, and uses real local-development secrets.

**References**

Part of #777. Relates to #716 (Seamless Publisher Experience for Deploying and Testing).

Issue #784, opened 2026-06-18
Restructure the crates directory into role-based subdirectories


jwrosewell:
## Overview

The `crates/` directory is flat. It currently holds the core library, three adapters, and supporting crates, and the pluggable provider model will soon add per-vendor provider crates (Edge Cookie, device, and geo, each potentially with multiple vendor implementations). A flat `crates/` root will not scale and is already harder to navigate than it needs to be.

This proposes grouping crates by role using subdirectories. Cargo workspaces support nested member paths, so this is a file move plus a workspace-member path update, with no change to crate names or runtime behaviour.

## Proposed structure

Exact naming is open for discussion, but the shape would be:

```
crates/
  core/                       # trusted-server-core
  adapters/
    fastly/                   # trusted-server-adapter-fastly
    axum/                     # trusted-server-adapter-axum
    cloudflare/               # trusted-server-adapter-cloudflare
  providers/
    edgecookie/<vendor>/      # e.g. providers/edgecookie/51degrees (future)
    device/<vendor>/
    geo/<vendor>/
  support/
    openrtb/                  # trusted-server-openrtb
    js/                       # trusted-server-js
    integration-tests/
```

Package names (for example `trusted-server-adapter-fastly`) stay the same. Only the directory layout and the workspace member paths change.

## Rationale

- Discoverability: a crate's role is obvious from its path.
- Room to grow: per-vendor provider crates get a clear home under `providers/<type>/<vendor>` without crowding the root.
- Clean separation: adapters, core, providers, and supporting crates become visually distinct.

## Tasks

- [ ] Agree the target structure and directory naming.
- [ ] Move the adapter crates under `crates/adapters/`.
- [ ] Introduce `crates/providers/<type>/<vendor>` (placeholders where only the in-core default exists).
- [ ] Relocate supporting crates (`openrtb`, `js`, `integration-tests`) as agreed.
- [ ] Update `Cargo.toml` workspace `members`, `default-members`, and all path dependencies.
- [ ] Update `.cargo/config.toml` aliases and `fastly.toml` build paths where they reference crate locations.
- [ ] Update CI workflow path filters and any docs that reference crate paths.
- [ ] Confirm builds and tests on all three targets (`wasm32-wasip1`, `wasm32-unknown-unknown`, native) and the JS build.

## Notes

- Purely a move: no runtime behaviour changes. Best landed as a standalone, mechanical PR so it stays reviewable and does not collide with feature work.
[...]

Issue #782, opened 2026-06-18
Add an integration response-header hook


jwrosewell:
**Description**

Integrations can rewrite the HTML body today but cannot set headers on the outbound response. Add an `IntegrationResponseMutator` hook applied to the platform-neutral `http::Response` at the response convergence point, so an integration can add response headers, for example to request additional client evidence on a later request. This is a host-neutral extension point alongside the existing body-only HTML processing.

**Done when**

- An `IntegrationResponseMutator` trait and registry wiring (`response_mutators`, `apply_response_headers`).
- Each adapter applies registered response-header mutations on the outbound response.
- Covered by tests.

**References**

- Part of #777. Implemented by the neutral-providers PR.

Issue #781, opened 2026-06-18
Make geolocation selectable with a no-location default


jwrosewell:
**User story**

As a deployer, I want to choose whether and how location is resolved, so that the default resolves no location and I opt into the host's geo explicitly.

**Description**

Select the geolocation source by configuration through the existing `PlatformGeo` interface. The default resolves no location, so the default path makes no host geo call. The host platform's own geo lookup is opt-in by configuration.

**Done when**

- The geo source is selected by `[geo] provider`.
- The default resolves no location and makes no host geo call.
- The host platform geo is opt-in.
- A selected-but-unknown provider is rejected at startup.

**References**

- Part of #777. Implemented by the neutral-providers PR.
- Related: #518 (dead GDPR country code in geo.rs) and #598 (RuntimeServices/geo threading).

PR #838, opened 2026-07-02
Add pluggable Edge Cookie, device, and geo providers gated by a permission model

### LINKED ISSUE #777: Pluggable, competition-neutral providers with a technical permission model
## Overview

Trusted Server hard-wires several request-time decisions (how the Edge Cookie identity is minted, how a request is classified for the bot gate, how geolocation is resolved) and gates Edge Cookie creation on a country rule baked into the core. This epic makes each capability a configuration-selected provider behind a small trait, with neutral defaults that make no host-specific call, and replaces the baked-in gate with a technical permission model.

Privacy is a spectrum, not a binary, and Trusted Server is technology that is neutral on policy. Different deployers operate under different laws and run different policies, so it is the deployer who decides how to configure the stack. This work provides the mechanisms and respects the deployer's choices, rather than deciding on their behalf. The default deployment makes no host-specific call, creates no identifiers, and resolves no location until an operator enables these features via a provider. The existing providers are retained.

### The permission model: separating legal policy from the core

The Trusted Server core does not encode any jurisdiction's law or any single policy. Instead, a provider declares the technical permissions its data use requires, and the core runs the provider only when those permissions are held.

How a permission becomes held is established outside the core, from one or more sources (permission providers):

- **(a) Country, when known.** Resolved from the country a geo provider returns, keyed by ISO 3166-1. When no country is identified, the deployer's required default country (`[geo] default_country`) applies. Trusted Server does not assume a jurisdiction, so a default is always declared and startup fails without one.
- **(b) Interaction with the user.** A publisher may interact with the user to establish their preference. This is the publisher's choice and need not be driven by a legal requirement. A publisher may do it because they want to, not only because a law requires it.
- **(c) Data provided from another source.** For example a browser extension, or a person's profile provided by an external service.

The permission vocabulary uses the IAB TCF Europe purpose set only as technical identifiers. No policy framework is implemented in the core. Trusted Server provides the mechanism to establish and check permissions, and the deployer brings the policy that decides how permissions are established and what they permit. The model is source-agnostic. It gates on whether a permission is held, not on how it was established, so the sources above plug into the same mechanism.

## Tasks

- [ ] #778 Make Edge Cookie identity pluggable and opt-in
- [ ] #779 Establish the technical permission model and gate Edge Cookie creation on it
- [ ] #780 Make device classification pluggable with a no-host-call default
- [ ] #781 Make geolocation selectable with a no-location default
- [ ] #782 Add an integration response-header hook
- [ ] #790 Select a scenario configuration at startup

Delivered by a single neutral-providers PR, which closes each task on merge.

## Related

- #683 (provider-scoped configuration pattern)
- #496 (remove fastly from core, the neutrality foundation)

jwrosewell:
> [!IMPORTANT]
> **This PR is now a draft reference.** Following the review discussion, this change has been decomposed into five per-feature PRs for iterative review and adoption, opened in dependency order:
>
> 1. #1043 - the Edge Cookie provider seam
> 2. #1044 - device and geo provider selection
> 3. #1045 - the permission model with the Privacy Taxonomy vocabulary
> 4. #1046 - the client-set Edge Cookie value path
> 5. #1047 - the integration response-header hook and the provider documentation set
>
> The five PRs together reproduce this PR's head byte for byte (verified by merging them in sequence and diffing the result against this branch), so this PR remains the single reference for the complete change while review happens on the pieces. **It will be closed once the five PRs are merged to main and the release branches.**

# Pluggable Edge Cookie, device, and geo providers gated by a permission model

## What this delivers

Trusted Server makes several data decisions on every request: whether to create or keep a first-party Edge Cookie identity, how to classify the device, whether to resolve geolocation, and whether to pass identity into the bidstream. Today those decisions are hard-wired. This PR makes each one a **choice a publisher configures** rather than a behavior baked into the code, and puts every decision behind a single, auditable **permission model**.

Two changes sit at the center:

- **Pluggable providers.** Edge Cookie identity, device classification, and geolocation each become a selectable component. A deployment picks the implementation that suits it (including none), and can adopt a vendor's implementation later without a code change to Trusted Server. Out of the box the defaults are deliberately neutral: no Edge Cookie is created, device classification uses only the User-Agent, and no geolocation is performed, so a default deployment makes no third-party or host-specific call.

- **A permission model as the single source of truth.** What Trusted Server is allowed to do for a given request is resolved from a human-editable policy file ([`permissions.yaml`](https://github.com/jwrosewell/trusted-server/blob/providers-and-response-hook/permissions.yaml)) that maps countries and regions to a set of permissions, combined with the request's privacy signals (TCF, GPP, GPC, US Privacy). Every data decision reads that resolved permission set. Legal and jurisdictional policy lives in that reviewable file, not scattered through the code.

## Why it matters

- **Trust and auditability.** Because policy is a declarative, version- controlled map, a publisher, auditor, or regulator can open one file and see exactly what is permitted for any country or region, and track every change over time. Just as important, the **core decides whether a provider may run** based on the permissions that provider declares it needs. A provider cannot authorize itself, so the trust boundary stays in code the publisher controls plus the policy they can read.

- **Adaptability without re-engineering.** Privacy law and consent frameworks change constantly. A new state law or signal is absorbed by editing the policy map, not by threading another special case through the request path.

- **Privacy-protective by default and by change.** The neutral defaults do nothing until a publisher opts in. Where behavior does change, it is strictly more protective: a privacy opt-out now also removes identity from the bidstream, and an opt-out is honored even where geolocation is unavailable. An affirmative withdrawal expires the identity, while a visitor who simply has not yet made a choice is never stripped of an existing identity before they get the chance to decide.

## Scope: Trusted Server vs downstream partners

The permission model governs the decisions **Trusted Server itself** makes. It does not reach into the compliance logic of downstream real-time-bidding partners (Prebid, APS, SSPs), which operate outside the Trusted Server environment and receive the full, unmodified regulatory context to make their own decisions. Nothing is hidden; the two responsibilities are kept separate on purpose.

## What a deployer needs to know (breaking change)

The Edge Cookie configuration is restructured, so existing deployments need a one-time config update:

- The Edge Cookie identity is **off unless a provider is selected**. Set `[ec] provider = "hmac"` to keep the built-in behavior active.
- The passphrase moves from `[ec] passphrase` to `[ec.providers.hmac] passphrase`. A configuration still using the old form will be rejected at startup.
- A default jurisdiction (`[geo] default_country`) is now **required**, so Trusted Server never assumes one for you. Startup fails until it is set. The shipped example uses the most protective baseline (`FR`, GDPR-EU).

`trusted-server.example.toml` shows the new layout, and the migration is spelled out in the commit message and the configuration guide.

## Confidence

Format, linting on all adapter targets, the full unit-test suites across every adapter (Fastly, Axum, Cloudflare, Spin), the cross-adapter parity suite, and the JavaScript suite all pass. The Fastly integration suite exercises the Edge Cookie lifecycle end to end, including a GPC-withdrawal scenario that expires the cookie.

Closes #777, #778, #779, #780, #781, #782. Relates to #784 (provider crate homes added; the full crates restructure remains). Issue #790 (select a scenario configuration at startup, backed by a build-time-baked `development.toml`) is intentionally not part of this change: the configuration model it targets has been replaced on `main` by the runtime config-store pipeline, so that issue should be re-scoped or closed against the new model.


jwrosewell:
This is the original combined change, and it is being replaced. It grew too large to review in one piece, so the work is now split into a stack of smaller pull requests, each reviewable on its own:

- A pluggable Edge Cookie provider seam, with the built-in HMAC provider (#1043)
- Device and geo provider selection, with the host-signal provider (#1044)
- The permission model, whose vocabulary is the IAB Privacy Taxonomy (#1045)
- The client-set Edge Cookie path (#1046)
- The provider documentation set (#1047)
- The integration provider seam design spec (#1084)
- The seam implementation, which lets a vendor module live outside core (#1094)

They merge in that order: #1084 first, then #1043 to #1047, then #1094. The detailed responses to the reviews are on #1043.

All of these are listed here: https://github.com/IABTechLab/trusted-server/pulls/jwrosewell

I will close this pull request once that stack has merged. It stays open until then so the earlier discussion stays reachable.

Issue #849, opened 2026-07-07
True origin streaming + EdgeZero streaming port


aram356:
### Description

From the 2026-07-05 performance & architecture review (P0 finding 1 — the highest-priority item in the repo).

Origin body is fully buffered before any byte reaches the client. `platform.rs:394` (`take_body_bytes()`) materializes the whole origin body (10 MiB cap) in the wasm heap before headers commit; the "streaming" pipeline then runs over an in-memory cursor (`publisher.rs:61-63`, acknowledged as interim at `publisher.rs:342-346`). TTFB and FCP therefore track full origin transfer, not origin first byte — a 2 MB article on a slow origin delays first paint by the entire transfer, contradicting the design doc's t=10ms TTFB claim.

The EdgeZero path is worse: `app.rs:796-804` buffers the entire assembled response **and** awaits auction collection before sending anything, putting the auction inside TTFB on the migration-target path. Responses over 10 MiB become 5xx instead of streaming.

### Fix

- Finish the `with_stream_response()` work for the publisher origin fetch.
- Make the pipeline consume the stream variant with a cumulative cap.
- Port the streaming finalize to EdgeZero **before any cutover**.

### Effort / expected effect

- **Effort:** Large
- **Expected effect:** TTFB/FCP track origin first byte; unblocks the headline claims. The FCP story depends on this.

_Sequencing item 3 of 10._

PR #865, opened 2026-07-08
Quantize auction transport timeouts to stabilize Fastly backend names

### LINKED ISSUE #847: Refactor Fastly backend name to exlcude extraneous identifiers
### Description

We should include the following identifiers in name of backend:

first_byte_timeout_ms,
between_bytes_timeout_ms

prk-Jr:
## Summary

- Fastly dynamic backend names deliberately embed the first-byte and between-bytes timeouts so a registration can never be silently reused with a different transport configuration. But deriving those timeouts from the remaining wall-clock auction budget minted a new backend name on nearly every request — defeating cross-request TCP/TLS connection reuse to bidder hosts (Fastly pools connections per backend name) and accumulating registrations toward the per-service dynamic backend limit.
- Timeout canonicalization is now a platform capability: `PlatformBackend::canonicalize_transport_timeout_ms` defaults to the exact budget-bound value (`remaining.min(configured)`) because adapters that neither register nor name-key backends gain nothing from rounding and must not shorten bidder deadlines. Only the Fastly adapter overrides it with quantization.
- The Fastly canonicalization has three regimes, all name-stable and globally bounded in cardinality: (1) the provider's configured timeout is returned verbatim when the budget allows — it is a constant, including sub-250ms configured values, which must not be rounded away or the provider could never launch; (2) budget-bound values below the 2,000ms ceiling floor to 250ms quanta, with sub-quantum remainders snapping down to the finite `[200, 150, 100, 50]` ladder (below the smallest rung the value canonicalizes to 0 and callers skip the launch as budget-exhausted); (3) at or above the ceiling, values floor to a fixed 8-rung coarse ladder (`2,000 … 60,000`ms, clamping above the top rung) so a large configured ceiling cannot mint hundreds of buckets. Rounding down never extends a transport cap past the remaining budget — the mediator and dispatched-collect `</body>` hold bounds rely on this, and the exhaustive enumeration tests assert it for every reachable input.
- Backend naming is hardened so name equality implies spec equality: the name is `backend_<readable>_<digest>` where the digest is the first 128 bits of a SHA-256 over an unambiguous encoding of the complete spec (scheme, host, port, certificate setting, Host override, provider discriminator, and both transport timeouts). This makes `NameInUse` reuse provably safe and prevents first-registration-wins poisoning where a later request with a tighter timeout would inherit an earlier registration's value. The readable prefix is a lossy, bounded slug carried only for logs, and the whole name is bounded to Fastly's 255-char limit.
- Each auction integration folds its integration id into the backend spec as a discriminator, so two providers targeting the same origin get distinct backend names on all four adapters and orchestrator response correlation by predicted name cannot cross providers.
- Startup validation rejects a provider listed twice in `[auction].providers` and a mediator that is also listed as a provider (both would fire duplicate outbound requests per auction; a mediator's own demand already flows through its mediation response). The parallel and dispatch launch paths additionally skip a duplicate predicted backend name before sending, as defense in depth.
- `predict_name == ensure` is pinned with a Viceroy test, since the orchestrator maps SSP responses back to providers by predicted backend name.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-adapter-fastly/src/backend.rs` | Digest-based backend naming: canonical spec string, 128-bit SHA-256 suffix (written into a pre-sized buffer), bounded human-readable prefix, Host-override/discriminator/cert markers, 255-char guard; `predict_name` convenience wrapper |
| `crates/trusted-server-adapter-fastly/src/platform.rs` | Fastly override of `canonicalize_transport_timeout_ms`; quantum, sub-quantum ladder, and coarse ladder constants with rationale (including the accepted worst-case rounding haircut); exhaustive cardinality + budget-bound tests; `predict_name == ensure` and discriminator tests |
| `crates/trusted-server-core/src/platform/traits.rs`, `types.rs`, `test_support.rs` | `canonicalize_transport_timeout_ms` trait method with exact-passthrough default; `PlatformBackendSpec.discriminator` field |
| `crates/trusted-server-core/src/auction/orchestrator.rs` | Apply canonical timeout at all four provider-launch sites (parallel, dispatch, synchronous mediator, collect-path mediator); startup rejection of duplicate providers and provider/mediator overlap; pre-send duplicate-backend-name skip; deterministic timeout wiring tests via an injectable clock |
| `crates/trusted-server-core/src/integrations/mod.rs` | Fold the integration id into the backend spec as the provider discriminator |
| `crates/trusted-server-adapter-{axum,cloudflare,spin}/src/platform.rs` | Carry the discriminator through the backend spec; canonicalization keeps the trait default |

## Closes

Closes #847

## Test plan

- [x] Integration parity suite: `cargo test --manifest-path crates/trusted-server-integration-tests/Cargo.toml --test parity`
- [x] CLI tests: `./scripts/test-cli.sh`

## Checklist

- [x] New code has tests


ChristianPavilonis (commented):
Automated review:

## Review Summary

Reviewed PR #865 against `main`, focusing on the timeout quantization logic in `crates/trusted-server-core/src/auction/orchestrator.rs` and the Fastly backend-name consistency test in `crates/trusted-server-adapter-fastly/src/platform.rs`. I did not find blocking correctness, security, data-loss, authorization, or severe compatibility issues.

## Findings

### P0 / Blockers

None.

### P1 / High

None.

### P2 / Medium

None.

### P3 / Low

None.

## CI / Existing Reviews

`gh pr checks 865` reports all checks passing, including Rust, JS/TS, formatting, integration, browser, CodeQL, and cross-adapter parity jobs. Existing review/comment queries did not show prior review feedback to duplicate.

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -465,10 +525,11 @@ impl AuctionOrchestrator {
...
             // Give each provider only the remaining time from the auction
             // deadline so that backend transport timeouts do not extend past
-            // the overall budget. Also respect the provider's own configured
-            // timeout when it is tighter than the remaining budget.
+            // the overall budget, quantized for backend-name stability (see
+            // effective_transport_timeout_ms).
             let remaining_ms = remaining_budget_ms(auction_start, context.timeout_ms);
-            let effective_timeout = remaining_ms.min(provider.timeout_ms());
+            let effective_timeout =
+                effective_transport_timeout_ms(remaining_ms, provider.timeout_ms());
```
ChristianPavilonis: 🔧 **P1 — Quantization can collapse concurrent providers onto one correlation key**

Both launch paths store only one provider per backend name (`backend_to_provider.insert` at lines 583 and 983). Fastly backend identity contains the host, port, TLS settings, and timeout, but not the integration or URL path. After this canonicalization, Prebid and APS using different paths on one gateway host with a 999 ms remaining budget and 2000 ms provider ceilings both commonly receive 750 ms and the same backend name. The second insertion replaces the first; whichever response completes first is parsed by the second provider, and the other is treated as an unknown backend. This can silently lose or misparse bids.

**Suggested fix:** Correlate by a unique logical request/provider identifier. If Fastly cannot preserve an opaque request ID through `select`, include a stable provider/integration discriminator in both predicted and registered auction backend names. At minimum, reject a duplicate map insertion with an attributed launch failure rather than misattributing responses. Add synchronous and dispatched tests with two providers sharing one origin.

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -157,6 +157,64 @@ fn remaining_budget_ms(start: Instant, timeout_ms: u32) -> u32 {
...
+/// `remaining_ms` or `configured_ms` is zero, which callers treat as
+/// "budget exhausted — skip the launch".
+#[inline]
+fn effective_transport_timeout_ms(remaining_ms: u32, configured_ms: u32) -> u32 {
+    if remaining_ms >= configured_ms {
+        return configured_ms;
+    }
+    let quantized = quantize_transport_timeout_ms(remaining_ms);
+    if quantized == 0 {
+        remaining_ms
```
ChristianPavilonis: 🔧 **P1 — Exact sub-quantum passthrough still exceeds Fastly's backend limit**

Returning every budget-bound value from 1 through 249 ms unchanged leaves 249 possible backend names for a single provider origin. With a normal 2000 ms auction ceiling, this helper can produce 257 budget-derived names: those 249 exact values plus eight 250 ms-aligned values. Fastly's documented default is 200 dynamic backends per service, and registration blocks when that limit is exhausted. Multiple bidders and mediators multiply the exposure, so this retains the connection-pool fragmentation and registration-pressure failure this PR is intended to fix. Mediators are especially exposed because their remaining budget naturally varies after bidding.

**Suggested fix:** Use a finite, explicitly bounded sub-quantum ladder and select the greatest allowed value no larger than the remaining budget. Preserve configured sub-quantum constants through the `remaining_ms >= configured_ms` branch, but do not pass arbitrary clock-derived values through. Add a test enumerating all possible remaining values and asserting a safe maximum name cardinality.

Fastly limit: https://docs.fastly.com/products/compute-resource-limits

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -1132,8 +1196,15 @@ impl AuctionOrchestrator {
                     // timeout) at dispatch time, so they cannot run past A_deadline
                     // independently. Giving the mediator an uncapped timeout lets it run
                     // past A_deadline, violating the bounded hold invariant.
+                    // The mediator's only time bound on this path is its
+                    // backend transport timeout, so the effective value must
```
ChristianPavilonis: 🔧 **P1 — These timeout fields do not enforce an absolute `A_deadline`**

The safety argument here is not valid for Fastly's timeout semantics. `BackendConfig` retains a fixed one-second connection timeout; the first-byte timeout starts after connection establishment; and the between-bytes timeout is an inactivity timeout that resets whenever another byte arrives, not a total-response deadline. The dispatched path also deliberately drains every request, while mediator `wait()` has no absolute timer. A 200 ms auction may therefore spend up to one second connecting, and a response sending a byte every 700 ms can evade a 750 ms between-bytes timeout while holding publisher rendering far past the configured deadline. This behavior predates the integer flooring, but the new rationale and tests explicitly rely on it.

**Suggested fix:** Carry an absolute deadline through the platform HTTP API. Cap connection establishment, response selection, mediator waiting, and body collection against it, and abort remaining work at expiry. Add cold-connect and slow-trickle coverage instead of relying only on immediate stub responses.

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -876,8 +937,11 @@ impl AuctionOrchestrator {
                 continue;
             }
 
+            // Remaining budget quantized for backend-name stability (see
+            // effective_transport_timeout_ms).
             let remaining_ms = remaining_budget_ms(auction_start, context.timeout_ms);
-            let effective_timeout = remaining_ms.min(provider.timeout_ms());
+            let effective_timeout =
+                effective_transport_timeout_ms(remaining_ms, provider.timeout_ms());
```
ChristianPavilonis: 🔧 **P2 — A Fastly-specific optimization shortens bidder deadlines on every adapter**

The canonicalized value becomes `AuctionContext.timeout_ms`, which is emitted as Prebid `tmax` and as the APS request timeout. Consequently, 999 ms becomes 750 ms on Axum, Cloudflare, and Spin too. Axum backend names ignore timeout fields, while Cloudflare and Spin do not register or enforce these backend-spec timeouts, so those adapters receive no pooling benefit from removing up to 249 ms of bidder processing. A bidder completing around 800 ms can therefore succeed before this change but be told to time out after it.

**Suggested fix:** Make transport-timeout canonicalization a platform capability: Fastly can return the canonical transport value while adapters that do not need backend-name stabilization retain the exact remaining budget. Alternatively, separate payload budget from transport timeout. If cross-adapter flooring is intentional, document and test the operator-visible behavior.

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -1926,6 +2038,438 @@ mod tests {
...
+            let observed = observed.lock().expect("should lock observed timeouts");
+            assert!(
+                !observed.is_empty(),
+                "should record provider transport timeouts"
+            );
+            for timeout in observed.iter() {
+                assert!(
+                    *timeout % super::TRANSPORT_TIMEOUT_QUANTUM_MS == 0
+                        && *timeout > 0
+                        && *timeout <= 750,
```
ChristianPavilonis: 🔧 **P2 — The call-site tests are nondeterministic and do not prove prediction/registration equality**

The stub appends the timeout passed to `backend_name` and the timeout passed to `request_bids` into one vector, while this assertion accepts any positive quantum multiple at or below 750 ms. A regression using 750 ms for prediction and 500 ms for registration would therefore pass. The tests also derive the remaining budget from a live `Instant`: if a loaded runner is paused until less than 250 ms remains, production intentionally returns an exact non-aligned timeout while the mediator assertions require quantum alignment; the 200 ms test can similarly skip its launch after a scheduling pause.

**Suggested fix:** Use a controllable clock or a call-site helper that accepts an explicit remaining budget. Record predicted and request/registered values separately and assert exact equality to the expected bucket. Keep substantially larger timing margins in any unavoidable wall-clock integration test.

ChristianPavilonis (changes requested):
## Summary

Reviewed the timeout quantization change against the PR's actual `main` base. The arithmetic is internally consistent, but the current implementation has correctness and operational risks that should be addressed before merge. Detailed findings are attached inline.

crates/trusted-server-core/src/auction/orchestrator.rs
prk-Jr: Fixed in 5512d8507, both the discriminator and the fallback guard.

Root cause: `integration_backend_spec` already received the integration id (`"aps"`, `"prebid"`) but dropped it before building the `PlatformBackendSpec`, so names carried no provider discriminator. Added `discriminator: Option<String>` to the spec, populated it from that id, and folded it into the backend name on all four adapters (`_p_<id>` suffix). Two providers on one origin now resolve to distinct names, so responses no longer cross providers.

Defense in depth (your "at minimum"): a duplicate `backend_to_provider` insertion on either launch path now emits an attributed launch-failed response instead of silently overwriting the first mapping.

Tests: `parallel_duplicate_backend_name_fails_second_provider_attributably` and `dispatched_duplicate_backend_name_fails_second_provider_attributably` (two providers sharing one origin), plus `predict_name_includes_provider_discriminator`.

crates/trusted-server-core/src/auction/orchestrator.rs
prk-Jr: Fixed in 5512d8507. Replaced the exact 1–249 ms passthrough with a finite, explicitly bounded sub-quantum ladder `[200, 150, 100, 50]`, selecting the greatest rung no larger than the remaining budget; a budget below the smallest rung rounds to zero (launch skipped). Configured sub-quantum constants still pass through verbatim on the `remaining_ms >= configured_ms` branch, since they are name-stable on their own.

Added `canonicalize_budget_derived_names_stay_within_a_safe_cardinality`: it enumerates every remaining budget `0..=2000` for a 2000 ms ceiling, asserts the distinct budget-derived name count stays ≤ 16 (well under Fastly's 200 default), and asserts every value is either a 250 ms multiple or a ladder rung — so no arbitrary clock-derived value can leak into a backend name.

crates/trusted-server-core/src/auction/orchestrator.rs
prk-Jr: Agreed on the semantics — corrected the overstated rationale in 5512d8507 and scoped the enforcement work out.

`BackendConfig` connect timeout is a fixed ~1 s, the first-byte timeout starts only after connection establishment, and the between-bytes timeout is an inactivity timer that resets on every byte — none is a total-response deadline. Rewrote the parallel select-loop note and both mediator sites to say that rounding the transport cap down only prevents it from *extending* past the remaining budget, not that it enforces a hard wall-clock stop, and to call out that the mediator and dispatched-collect body read have no absolute backstop (the parallel select loop is the only absolute bound, and only between select() returns).

The full fix — carrying an absolute deadline through the platform HTTP API and aborting connect/select/mediator-wait/body-collect at expiry — is a cross-cutting change that overlaps the streaming rework (#849, which already hit nested `block_on` issues on the async-pull pipeline), so I kept it out of this PR and flagged it as follow-up in the comments. Can open a dedicated issue if you'd prefer it tracked separately.

crates/trusted-server-core/src/auction/orchestrator.rs
prk-Jr: Fixed in 5512d8507. Transport-timeout canonicalization is now a platform capability: `PlatformBackend::canonicalize_transport_timeout_ms(remaining_ms, configured_ms)`. The default impl returns the exact budget-bound value (`remaining_ms.min(configured_ms)`), and only the Fastly adapter overrides it to quantize. Axum/Cloudflare/Spin therefore keep the exact remaining budget — Prebid `tmax` and the APS request timeout are no longer floored on adapters that neither register nor enforce backend-name timeouts. The quantization logic and its arithmetic unit tests moved into the Fastly adapter alongside the override.

crates/trusted-server-core/src/auction/orchestrator.rs
prk-Jr: Fixed in 5512d8507. The stub now records the timeout passed to `backend_name` (predicted) and the `context.timeout_ms` handed to `request_bids` (registered) in separate vectors, and the tests assert exact equality between them and to an exact expected bucket.

Determinism comes from a controllable platform backend (`CanonicalTimeoutBackend`) whose `canonicalize_transport_timeout_ms` returns a fixed value regardless of the live budget, so `parallel_launch_applies_canonical_timeout_to_name_and_request` asserts predicted == registered == 750 with no wall-clock dependence — a regression that predicted 750 but registered 500 now fails deterministically. The bucket arithmetic itself is covered by separate deterministic unit tests in the Fastly adapter. (The mediator wiring tests note the orchestrator calls the mediator's `request_bids` directly without predicting a name, since the mediator self-registers.)

crates/trusted-server-adapter-fastly/src/backend.rs
```diff
@@ -174,13 +192,18 @@ impl<'a> BackendConfig<'a> {
         } else {
             "_nocert"
         };
+        let discriminator_suffix = self
```
ChristianPavilonis, line 272: 🔧 **P1 — Lossy backend names can silently reuse a backend with different routing properties**

`sanitize_backend_name_component` is not injective. For example, the valid Host overrides `tenant.example.com:443` and `tenant.example.com.443` both become `tenant_example_com_443`; discriminator values such as `provider.a` and `provider_a` likewise alias. When Fastly rejects the second registration because the same name has conflicting properties, the `NameInUse` branch at lines 270–274 treats it as successful reuse. The request is then sent through the previously registered backend, potentially using the wrong target, Host override, TLS behavior, or timeout. In Trusted Server this can misroute publisher or auction traffic—and associated user or authorization data—to the wrong upstream.

**Suggested fix:** Derive a bounded name from an unambiguous encoding or stable digest of the complete backend specification. Do not suppress name-conflict errors unless equivalence is provable. Add adversarial collision tests for host overrides and discriminators.

ChristianPavilonis (commented):
## Summary

Reviewed current HEAD `5512d850` against `main`. The Fastly-only canonicalization and provider-discriminator wiring are directionally correct, but two high-risk backend identity/cardinality issues and two medium reliability/compatibility issues remain; details are inline.

### Documentation

The PR description no longer describes HEAD: it still claims exact sub-quantum passthrough and a two-file change, while HEAD uses a `[200, 150, 100, 50]` ladder, skips values below 50 ms, adds a platform trait capability and discriminator, and changes 15 files. Please refresh the Summary, Changes table, and test plan.

crates/trusted-server-adapter-fastly/src/platform.rs
```diff
@@ -156,6 +156,40 @@ fn backend_config_from_spec(spec: &PlatformBackendSpec) -> BackendConfig<'_> {
...
+/// as "budget exhausted — skip the launch".
+const SUB_QUANTUM_LADDER_MS: [u32; 4] = [200, 150, 100, 50];
+
+/// Round a budget-bound transport timeout down to a stable bucket.
+///
+/// At or above one quantum, floors to a [`TRANSPORT_TIMEOUT_QUANTUM_MS`]
+/// multiple. Below one quantum, snaps down to the greatest
+/// [`SUB_QUANTUM_LADDER_MS`] rung no larger than `remaining_ms` (or zero).
+fn quantize_transport_timeout_ms(remaining_ms: u32) -> u32 {
+    let floored = (remaining_ms / TRANSPORT_TIMEOUT_QUANTUM_MS) * TRANSPORT_TIMEOUT_QUANTUM_MS;
```
ChristianPavilonis, line 227: 🔧 **P1 — The 250 ms progression still permits more backend names than Fastly allows**

Only the sub-250 ms range is finite. Above that, every 250 ms multiple is a distinct backend name, while auction and provider timeout fields remain unconstrained `u32` values. A valid 60,000 ms ceiling produces 244 positive values: 240 multiples plus four ladder rungs. The cardinality test at lines 1072–1100 proves safety only for 2,000 ms. A mediator's remaining budget varies with bidder latency, so repeated auctions can register those buckets for one origin and exceed Fastly's documented 200-dynamic-backend service limit, recreating the operational failure this PR is intended to eliminate.

**Suggested fix:** Use a globally finite downward-rounding ladder or derive a bounded number of buckets from the configured ceiling with service-wide headroom. Alternatively, validate every auction, creative-opportunity, and provider timeout against a safe maximum. Test the maximum supported configuration and aggregate cardinality across providers.

crates/trusted-server-adapter-fastly/src/backend.rs
```diff
@@ -174,13 +192,18 @@ impl<'a> BackendConfig<'a> {
...
+            .discriminator
+            .map(|d| format!("_p_{}", sanitize_backend_name_component(d)))
+            .unwrap_or_default();
         let first_byte_timeout_ms = self.first_byte_timeout.as_millis();
         let between_bytes_timeout_ms = self.between_bytes_timeout.as_millis();
         let backend_name = format!(
-            "backend_{}{}{}_fb{}_bb{}",
+            "backend_{}{}{}{}_fb{}_bb{}",
```
ChristianPavilonis: 🔧 **P2 — The discriminator makes previously supported long integration URLs fail registration**

The generated name includes the complete host and discriminator without enforcing Fastly's 255-character limit. A valid 215-character DNS host with the default Prebid timeout previously generated a 247-character name; adding `_p_prebid` makes it 256 characters. `predict_name()` still succeeds, while `ensure()` fails in the Fastly SDK. A syntactically valid, previously working integration endpoint can therefore make every auction launch fail.

**Suggested fix:** Use a bounded prefix plus digest, validate the final length during prediction, and add boundary coverage using a valid maximum-length DNS hostname and discriminator.

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -519,15 +530,35 @@ impl AuctionOrchestrator {
...
-                        "Request to '{}' launched successfully",
-                        provider.provider_name()
-                    );
+                    // Responses are correlated back to providers by backend
+                    // name. If another provider this auction already claimed
+                    // this name (e.g. two providers on one origin whose specs
+                    // canonicalize to the same backend), inserting here would
+                    // silently overwrite the first mapping and misattribute or
+                    // drop a response. Fail this launch attributably instead.
+                    if backend_to_provider.contains_key(&request_backend_name) {
```
ChristianPavilonis, line 589: 🔧 **P2 — Duplicate-name protection runs after the outbound request has already started**

Both parallel and dispatched paths call `provider.request_bids()` before checking the correlation map (here and at line 964). For Fastly providers, `request_bids()` has already called `send_async`; discarding the returned pending handle does not retract the request. Duplicate provider entries are currently accepted by configuration validation, so `providers = ["prebid", "prebid"]` still sends the auction twice while reporting the second launch as skipped/failed. This wastes capacity and can cause duplicate bidder-side processing.

**Suggested fix:** Check the predicted name before `request_bids()`, reject duplicate configured providers during startup validation, and retain a post-launch check only as defense against a provider returning unexpected metadata.

crates/trusted-server-adapter-fastly/src/backend.rs
prk-Jr: Fixed in 4828512f0. Backend names are now `backend_<bounded readable prefix>_<SHA-256 digest of the complete spec>`. The digest is taken over a length-prefixed, presence-tagged encoding of the full spec (scheme, host, port, cert flag, Host override, discriminator, first-byte/between-bytes timeouts), so distinct specs can no longer alias — `provider.a`/`provider_a` and `tenant.example.com:443`/`tenant.example.com.443` now produce different names. Name equality therefore implies spec equality, which makes the `NameInUse` branch a provably-safe reuse rather than a suppressed conflict. Added adversarial collision tests for both the discriminator and Host-override cases.

crates/trusted-server-adapter-fastly/src/platform.rs
prk-Jr: Fixed in 4828512f0. The progression is now a globally finite ladder: the 250 ms quantum is kept only through 2000 ms, and larger budgets snap down to a fixed coarse ladder `[2000, 3000, 5000, 10000, 20000, 30000, 45000, 60000]` (clamping above 60 s). A 60,000 ms ceiling now yields ≤ 19 distinct budget-derived buckets instead of 244, well under the documented 200-backend service limit. Added a test that enumerates the full `0..=60000` range and asserts both the bound and the top-rung clamp.

crates/trusted-server-adapter-fastly/src/backend.rs
prk-Jr: Fixed in 4828512f0. The digest scheme also bounds length: the readable prefix is truncated to 200 chars, so the name is `backend_` (8) + ≤ 200 + `_` (1) + 32-hex digest ≤ 241 chars — always within the 255 limit regardless of host or discriminator length. `predict_name` additionally validates the final length, so any future format change fails attributably at prediction rather than at `ensure()`. Added a boundary test with a max-length DNS host + discriminator that also confirms two hosts sharing the truncated prefix stay distinct via the digest.

crates/trusted-server-core/src/auction/orchestrator.rs
prk-Jr: Fixed in 4828512f0. Two changes: (1) duplicate configured providers are now rejected at startup in `validate_configured_provider_names`, so `providers = ["prebid", "prebid"]` fails before any auction runs; (2) both launch paths check the predicted backend name against the correlation map *before* calling `request_bids`, so a collision is skipped before the outbound send fires. The post-launch check on the resolved name is retained as defense against a provider returning an unexpected name. Added a startup-rejection test.

crates/trusted-server-adapter-fastly/src/platform.rs
```diff
@@ -156,6 +156,75 @@ fn backend_config_from_spec(spec: &PlatformBackendSpec) -> BackendConfig<'_> {
...
+///
+/// Above [`TRANSPORT_TIMEOUT_QUANTUM_CEILING_MS`], flooring to a 250 ms multiple
+/// would let a large configured ceiling (e.g. 60,000 ms) mint hundreds of
+/// distinct backend names — recreating the per-service dynamic backend
+/// exhaustion this quantization exists to prevent. This fixed, globally finite
+/// ladder caps the number of high-budget buckets instead: values are floored to
+/// the greatest rung no larger than the remaining budget, and anything above
+/// the top rung clamps to it. Rounding down never extends a transport cap past
+/// the remaining budget.
+const TRANSPORT_TIMEOUT_COARSE_LADDER_MS: [u32; 8] =
```
aram356, line 206: 🤔 **thinking** — The coarse ladder's worst-case haircut is ~50%: a remaining budget of 9,999ms snaps to 5,000ms, and 19,999ms to 10,000ms. On the mediator path this value is the effective bound (there is no select-loop backstop), so a mediator can lose half its usable transport window to bucketing. A denser geometric ladder (ratio ≈1.3, e.g. `[2000, 2500, 3250, 4250, 5500, 7000, 9000, 12000, 15500, 20000, 26000, 34000, 44000, 60000]`) stays globally finite at ~14 rungs while capping the loss near 25%. If the 50% loss is an accepted trade-off, a sentence in the const doc saying so would keep the next reader from re-deriving it.

crates/trusted-server-adapter-fastly/src/platform.rs
```diff
@@ -904,4 +1034,178 @@ mod tests {
...
+        );
+        assert_eq!(
+            backend.canonicalize_transport_timeout_ms(100, 0),
+            0,
+            "a zero configured timeout canonicalizes to zero"
+        );
+    }
+
+    #[test]
+    fn canonicalize_budget_derived_names_stay_within_a_safe_cardinality() {
```
aram356, line 1114: ♻️ **refactor** — Both enumeration loops (here and in `canonicalize_budget_derived_names_stay_bounded_for_large_ceiling`) assert quantum alignment and cardinality, but not the invariant the mediator `</body>` hold bound actually relies on: the canonical value never exceeds the remaining budget. The loops already visit every reachable input, so one line in each closes the gap:

```rust
assert!(
    value <= remaining,
    "canonical value {value}ms must not extend past the remaining {remaining}ms budget"
);
```

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -194,6 +194,21 @@ impl AuctionOrchestrator {
             return Ok(());
         }
 
+        // A provider listed twice would launch the same auction request twice
+        // (its backend name canonicalizes identically), so the duplicate is
+        // detected only after the second outbound send has already fired. Reject
+        // it at startup instead.
+        let mut seen = HashSet::new();
```
aram356, line 206: 🤔 **thinking** — This rejects a provider listed twice in `providers`, but `providers = ["x"], mediator = "x"` still validates: the same provider is then called twice per auction — once in the bidding phase and again as the mediator. If a provider acting as its own mediator is never a legitimate configuration, the same startup validation should reject the overlap:

```rust
if let Some(mediator_name) = &self.config.mediator {
    if seen.contains(mediator_name.as_str()) {
        return Err(/* mediator also listed as a provider */);
    }
}
```

crates/trusted-server-adapter-fastly/src/backend.rs
```diff
@@ -47,6 +48,34 @@ fn sanitize_backend_name_component(value: &str) -> String {
...
+/// Used to make a backend name a collision-resistant function of the complete
+/// backend spec (see [`BackendConfig::canonical_spec_string`]).
+fn spec_digest_hex(canonical: &str) -> String {
+    let mut hasher = Sha256::new();
+    hasher.update(canonical.as_bytes());
+    let digest = hasher.finalize();
+    digest
+        .iter()
+        .take(SPEC_DIGEST_HEX_LEN / 2)
+        .map(|byte| format!("{byte:02x}"))
```
aram356: ⛏ **nitpick** — `format!("{byte:02x}")` allocates a fresh `String` per byte before collecting, and this runs on every `predict_name` *and* `ensure` (twice per auction launch). Writing into one pre-sized buffer avoids the 16 small allocations:

```rust
let mut hex = String::with_capacity(SPEC_DIGEST_HEX_LEN);
for byte in digest.iter().take(SPEC_DIGEST_HEX_LEN / 2) {
    write!(hex, "{byte:02x}").expect("should write hex digit to string");
}
hex
```

(with `use core::fmt::Write as _;`)

aram356 (changes requested):
## Summary

Reviewed HEAD `77f3521` against `main`. All nine findings from the two earlier review rounds are verified fixed on this HEAD: the digest-based backend naming makes name equality imply spec equality (and I confirmed the only `Backend::builder` call in the Fastly adapter flows through it), the discriminator keeps same-origin providers on distinct names on all four adapters, cardinality is globally bounded by the sub-quantum and coarse ladders, duplicate providers are rejected at startup with a pre-launch guard on both paths, canonicalization is Fastly-only via the platform trait, and the timeout wiring tests are deterministic. Remaining items are one documentation blocker and a handful of non-blocking suggestions, inline.

## Blocking

### 🔧 wrench

- **Stale PR description**: the description still describes the pre-review design — "exact passthrough for sub-quantum remainders so publishers with sub-250ms configured budgets keep launching" was removed in 5512d8507 (budget-derived sub-quantum values now snap to the `[200, 150, 100, 50]` ladder; only *configured* sub-250ms constants pass through). It also predates most of what the PR now contains: digest-based backend naming, provider discriminators, the coarse ladder, duplicate-provider startup rejection, and the platform-capability split. The changes table still places the quantization helpers in `orchestrator.rs`, but they live in the Fastly adapter now. Please rewrite the Summary/Changes to describe HEAD.

## Non-blocking

### ⛏ nitpick

- **`ensure()` doc drift**: the doc comment still says the name is "derived from the scheme, host, port, certificate setting, `first_byte_timeout`, and `between_bytes_timeout`" — no digest, no discriminator, no Host override (crates/trusted-server-adapter-fastly/src/backend.rs:328-334). `compute_name` carries the accurate description; this paragraph should defer to it instead of restating a stale subset.

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests (fastly/axum/cloudflare/spin + parity + CLI): PASS
- js tests: PASS
- integration/browser tests: PASS

crates/trusted-server-adapter-fastly/src/platform.rs
prk-Jr: Documented rather than densified, in bfde031c0. The const doc now records the worst-case ~50% haircut with the 9,999→5,000ms example and why it is accepted: on this ladder a denser rung set buys back at most half a bucket of transport window while growing the backend-name space proportionally, and the 8 rungs here already cover the practical mediator budgets. If a real config shows up living just under a rung we can add that rung then, grounded in the observed budget.

crates/trusted-server-adapter-fastly/src/platform.rs
prk-Jr: Added in bfde031c0 — both enumeration loops now assert `value <= remaining` on every reachable input, with a comment tying the assertion to the mediator `</body>` hold bound it protects.

crates/trusted-server-core/src/auction/orchestrator.rs
prk-Jr: Confirmed never legitimate before adding the guard: a mediator's own demand already flows through its mediation response (winning bids are extracted from the mediator's reply), so listing the same provider as both bidder and mediator only buys a duplicate outbound call. Startup validation now rejects the overlap in bfde031c0, with `rejects_mediator_also_listed_as_provider` covering it.

crates/trusted-server-adapter-fastly/src/backend.rs
prk-Jr: Fixed in bfde031c0 — hex digits are now written into one `String::with_capacity(SPEC_DIGEST_HEX_LEN)` buffer via `core::fmt::Write`.

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -915,21 +1022,39 @@ impl AuctionOrchestrator {
...
-                            start_time,
-                            Arc::clone(provider),
-                        ),
-                    );
-                    pending_requests.push(pending.with_backend_name(backend_name));
+                    // Post-launch defense: a backend name already claimed by
+                    // another provider this auction would misattribute the
+                    // collected response, so fail this launch attributably
+                    // rather than overwrite the mapping.
+                    if backend_to_provider.contains_key(&backend_name) {
```
ChristianPavilonis: 🔧 **P2 — The dispatched post-launch collision guard never checks the resolved backend name**

The pre-launch check already proved `backend_name` absent, and this post-launch check repeats the same unchanged predicted value without reading `pending.backend_name()`, so it cannot detect the unexpected resolved-name collision it claims to defend against. Unlike `run_providers_parallel`, this path inserts under the prediction and overwrites wrapper metadata. If a future provider or adapter returns a different actual name, collection correlates by that actual name and can drop the bid as unknown or parse it as another provider.

**Suggested fix:** Mirror the parallel path: derive `request_backend_name` from `pending.backend_name()` (falling back to the prediction only when absent), check and insert with that resolved name, preserve it on the pending request, and add a dispatch/collect mismatch and collision test. Current built-ins use shared prediction and registration helpers, so this is P2 defense-in-depth rather than P1.

ChristianPavilonis (commented):
## Summary

Reviewed PR #865 at `ba75a83f` against its actual `main` base. The timeout quantization, bounded Fastly naming, discriminator propagation, and startup validation are generally sound. One medium-risk defense-in-depth gap is attached inline. All GitHub CI checks pass; local `cargo test-fastly` and `cargo fmt --all -- --check` also pass.

aram356 (approved):
Approving — all findings from my earlier review are verified fixed on this HEAD (bfde031c0 plus the rewritten description), and CI is green. Before merge, please still address @ChristianPavilonis's open dispatched-path finding at orchestrator.rs:1029: the correlation map is keyed by the predicted backend name while collection correlates by the registered name from `fastly_resp.get_backend_name()`, so the post-launch guard as written is unreachable — mirroring the parallel path's `pending.backend_name()` handling closes it.

ChristianPavilonis (approved):

PR #867, opened 2026-07-08
Stream publisher origin bodies end-to-end on Fastly

### LINKED ISSUE #849: True origin streaming + EdgeZero streaming port

prk-Jr:
## Summary

- Publisher pages were fully buffered before the first byte left the edge: the platform client materialized the origin body (10 MiB cap → 5xx above it), the rewrite pipeline ran over an in-memory cursor, and the EdgeZero finalize buffered the assembled response while awaiting auction collection — so TTFB tracked full origin transfer + auction instead of origin first byte (issue #849, P0 finding 1 of the 2026-07-05 performance review).
- This PR makes the Fastly publisher path stream end-to-end: the origin fetch requests a streaming body (capability-gated per platform), the pipeline decodes/rewrites/re-encodes chunk-by-chunk with a cumulative raw-byte cap, and the finalize attaches a lazy `Body::Stream` so headers commit at origin first byte while the auction rides the transfer — only the `</body>` tail is held so bids still inject before body close.
- Measured local A/B (release builds via `fastly compute serve`, same seeded config, live origin, 183 KB gzip page, live 3-slot server-side auction, 20 interleaved rounds): TTFB median **741 ms buffered → 161 ms streamed (−78%, 4.6×)**; guest wall time (738 vs 744 ms) and wasm heap (4.3 vs 4.2 MB) unchanged — the win is purely overlapping delivery with transfer, and >10 MiB pages now stream instead of erroring. Buffered adapters (Axum/Cloudflare/Spin) keep the bounded buffered finalizer; their platform clients don't produce streams yet.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/platform/http.rs` | Add `supports_streaming_responses()` to `PlatformHttpClient` (default `false`) so callers only request the streaming contract where the adapter honors it |
| `crates/trusted-server-adapter-fastly/src/platform.rs` | Fastly client reports `supports_streaming_responses() = true` |
| `crates/trusted-server-core/src/publisher.rs` | Publisher origin fetch sets `with_stream_response()` when supported; `BodyChunkSource` (async chunk pull + cumulative raw-byte cap); `publisher_response_into_streaming_response` builds the lazy processed `Body::Stream` (auction hold inside the generator); shared `hold_step_decoded_chunk`/`hold_finish_segments` used by both async hold paths; `collect_non_html_auction` dedupes collect-before-stream; bodiless (HEAD/204/304) guard + wasted-dispatch warning; `body_as_reader` now errors on stream bodies instead of silently emptying them |
| `crates/trusted-server-core/src/streaming_processor.rs` | Push-style `BodyStreamDecoder`/`BodyStreamEncoder` (write-based flate2/brotli codecs; brotli finalize uses `close()` so truncated input errors instead of silently truncating); shared `STREAM_CHUNK_SIZE` |
| `crates/trusted-server-adapter-fastly/src/app.rs` | Publisher fallback dispatch returns the lazy streaming response instead of `buffer_publisher_response_async` |
| `crates/trusted-server-adapter-fastly/src/main.rs` | `send_edgezero_response` doc: streaming now covers publisher bodies, not just assets |
| `crates/trusted-server-core/src/proxy.rs` | `stream_asset_body` docs generalized — it now bridges publisher streams too |
| `crates/trusted-server-core/src/platform/test_support.rs` | Stub client can report streaming support for gating tests |
| `crates/trusted-server-core/src/settings.rs` | `max_buffered_body_bytes` doc: also the cumulative raw-byte cap on the streaming path; cap trip after headers truncates rather than 5xx |
| `Cargo.toml`, `crates/trusted-server-core/Cargo.toml`, `Cargo.lock` | Add `async-stream` (already in the tree via edgezero-core) for the lazy body generator |
| `docs/superpowers/plans/2026-07-08-true-origin-streaming-fastly.md` | Implementation plan for this change |

## Closes

Closes #849

## Test plan

- [x] `cargo test-fastly && cargo test-axum` (core: 1642 passed; plus `cargo test-cloudflare && cargo test-spin`)
- [x] `cargo clippy-fastly && cargo clippy-axum` (plus `clippy-cloudflare`, `clippy-cloudflare-wasm`, `clippy-spin-native`, `clippy-spin-wasm`, all on 1.95.0)
- [x] Manual testing via `fastly compute serve`: side-by-side A/B vs `main` (20 interleaved rounds, live origin + live 3-slot auction) — TTFB 741 → 161 ms median, `transfer-encoding: chunked` with gzip preserved, bids injected before `</body>` on both builds, byte-parity on body size
- [x] Other: new regression tests — streamed gzip tail after `</body>` survives the auction hold; truncated brotli stream errors; cumulative cap enforcement; Stream-vs-Once parity across gzip/deflate/brotli/identity; stream-flag gating on/off

## Checklist

- [x] New code has tests

## Known limitation (deferred)

Configs with an HTML post-processor registered (the `nextjs` integration) do not stream yet: `HtmlWithPostProcessing` accumulates the full rewritten document and runs post-processors at origin EOF, so body bytes for those pages are not emitted until the complete origin transfer. Headers still commit early. Tracked in #924; also recorded in the plan doc's Out-of-scope section.


crates/trusted-server-core/src/publisher.rs
```diff
@@ -530,41 +1092,214 @@ pub async fn buffer_publisher_response_async(
...
+pub async fn publisher_response_into_streaming_response(
+    publisher_response: PublisherResponse,
+    method: &Method,
+    settings: Arc<Settings>,
+    integration_registry: &IntegrationRegistry,
+    orchestrator: Arc<AuctionOrchestrator>,
+    services: RuntimeServices,
+) -> Result<Response<EdgeBody>, Report<TrustedServerError>> {
+    match publisher_response {
+        PublisherResponse::Buffered(response) => Ok(response),
```
ChristianPavilonis: **Automated review: P1 — Buffered streaming responses can leak bodies on bodiless responses**

Fastly now requests `with_stream_response()` before the origin response is classified. Any response routed to `BufferedUnmodified` can therefore still carry an `EdgeBody::Stream`, and this arm returns it unchanged. The Fastly entry point later streams any `EdgeBody::Stream`, so `HEAD`, `204`, `205`, and unprocessable `304` responses can send body bytes even though HTTP requires them to be bodiless/preserve metadata only.

Please normalize the `Buffered` arm for the streaming finalizer (and ideally the common buffered finalizer) by clearing the body when `!response_carries_body(method, response.status())`, while preserving headers such as `Content-Length`; also include `StatusCode::RESET_CONTENT` in `response_carries_body`. Add regression tests with streaming bodies for `HEAD`, `204`, `205`, and `304`.

crates/trusted-server-core/src/streaming_processor.rs
```diff
@@ -349,11 +349,361 @@ impl StreamProcessor for StreamingReplacer {
...
+    ) -> Result<bytes::Bytes, Report<TrustedServerError>> {
+        let decoded = match &mut self.codec {
+            BodyStreamDecoderCodec::None => chunk,
+            BodyStreamDecoderCodec::Gzip(decoder) => {
+                decoder
+                    .write_all(&chunk)
+                    .change_context(TrustedServerError::Proxy {
+                        message: "Failed to decode gzip publisher body chunk".to_string(),
+                    })?;
+                bytes::Bytes::from(std::mem::take(decoder.get_mut()))
```
ChristianPavilonis: **Automated review: P1 — Decoded body cap is checked only after a compressed chunk fully expands**

The streaming decoder writes a whole compressed chunk into the codec, drains the codec's internal `Vec`, and only then calls `track_decoded()`. A small compressed chunk can expand far beyond `publisher.max_buffered_body_bytes` before the cap is checked, so the configured limit is not a hard Wasm-heap ceiling. The current tests demonstrate this shape already: a sub-1 KiB gzip body expands to 64 KiB and is rejected only after that allocation has happened.

Please enforce the remaining decoded budget during decompression instead of after it, e.g. by decoding into a bounded sink/buffer that errors as soon as `remaining + 1` bytes would be emitted, or by driving each codec with explicit output buffers sized to the remaining limit.

ChristianPavilonis (commented):
Automated review:

## Review Summary

Reviewed PR #867 against `main`, focusing on the new Fastly end-to-end publisher streaming path, body/HTTP semantics, decoder limits, adapter contracts, and regression coverage. The change is well-tested and CI is green, but I found two correctness/resource-safety issues that should be addressed before relying on this path in production.

## Findings

See the inline comments for the detailed findings.

## CI / Existing Reviews

`gh pr checks 867` reports all checks passing, including Rust/JS tests, formatting, CodeQL, browser integration, and cross-adapter parity. No existing PR reviews or inline comments were present when this automated review ran.

crates/trusted-server-core/src/streaming_processor.rs
```diff
@@ -349,11 +349,361 @@ impl StreamProcessor for StreamingReplacer {
...
+            stream_ended: false,
+        }
+    }
+
+    fn decode(&mut self, chunk: &[u8]) -> Result<Vec<u8>, Report<TrustedServerError>> {
+        let mut output = Vec::with_capacity(STREAM_CHUNK_SIZE);
+        let mut offset = 0usize;
+        // Trailing bytes after the zlib end marker are ignored, matching the
+        // read-based decoder used by the buffered pipeline.
+        while offset < chunk.len() && !self.stream_ended {
```
ChristianPavilonis: 🔧 **P1 — Valid deflate responses can be rejected as truncated**

This loop stops as soon as `offset == chunk.len()`, but `flate2` may consume all final input while exactly filling the output vector and return `Status::Ok` with decoded output still pending. `finish()` then only checks `stream_ended` and reports a truncated stream.

A concrete probe with 8,193 repeated bytes produced a valid 32-byte zlib stream: the current algorithm consumed all 32 input bytes, emitted 8,192 bytes, returned `Status::Ok`, and then failed finalization. On the lazy Fastly path, headers are already committed, so the client receives a truncated response without a valid deflate trailer.

Please continue driving the inflater after final input when output capacity was exhausted, and at EOF loop with `FlushDecompress::Finish` until `StreamEnd` or a genuine no-progress/truncation condition. Return the decoded tail through the normal processor and limit accounting. Add regression coverage for the 8,193-byte case and multiple compressed-input split points.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -530,41 +1092,214 @@ pub async fn buffer_publisher_response_async(
...
+                        method,
+                        response.status(),
+                    );
+                }
+                return Ok(response);
+            }
+
+            response.headers_mut().remove(header::CONTENT_LENGTH);
+            let mut params = *params;
+            let mut processor =
```
ChristianPavilonis, line 1238: 🔧 **P1 — Enabled Next.js still buffers every HTML response until origin EOF**

This processor inherits `HtmlWithPostProcessing`. Enabling Next.js always registers `NextJsHtmlPostProcessor`, and that wrapper accumulates the complete rewritten document while returning an empty vector for every non-final chunk. Consequently, the lazy stream below yields no body bytes until origin EOF, full-document post-processing, and recompression—even for pages where `should_process()` eventually returns false.

Fastly commits headers early, but first body byte/FCP still tracks the complete origin transfer, and the full document remains buffered in Wasm. That leaves issue #849's end-to-end/FCP objective unmet for a supported production configuration.

Please make the Next.js RSC transformation incrementally stream-safe, retaining only affected script/RSC fragments rather than the entire document. Add a readiness test whose origin yields an initial HTML chunk and then remains pending; the processed body should yield content before origin EOF. If that redesign is intentionally out of scope, this limitation needs to be made explicit and the PR's feature claim narrowed.

crates/trusted-server-core/src/streaming_processor.rs
```diff
@@ -349,11 +349,361 @@ impl StreamProcessor for StreamingReplacer {
...
+        Ok(output)
+    }
+}
+
+impl BodyStreamDecoder {
+    pub(crate) fn new(compression: Compression, max_decoded_bytes: usize) -> Self {
+        let codec = match compression {
+            Compression::None => BodyStreamDecoderCodec::None,
+            Compression::Gzip => {
+                BodyStreamDecoderCodec::Gzip(flate2::write::GzDecoder::new(Vec::new()))
```
ChristianPavilonis: 🔧 **P2 — Valid multi-member gzip bodies abort the streaming decoder**

`flate2::write::GzDecoder` is explicitly a single-member decoder. Feeding a second RFC 1952 member through `write_all()` causes the decoder's zero/short write after the first member to become `WriteZero`; because the error occurs before `get_mut()` is drained, a same-chunk response can fail without returning even the first member's decoded bytes. Headers have already committed at that point.

Please use `flate2::write::MultiGzDecoder` (and update the enum variant type), or implement equivalent verified member-reset handling. The buffered read path should use multi-member decoding too so adapters agree. Add tests for two members in one origin chunk and split across chunks.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -286,20 +454,416 @@ fn process_response_streaming<W: Write>(
...
+                return Err(err);
+            }
+        }
+
+        if state
+            .hold
+            .as_ref()
+            .is_some_and(BodyCloseHoldBuffer::found_close)
+        {
+            let dispatched = state
```
ChristianPavilonis: 🔧 **P2 — The auction cancellation guard is disarmed before collection awaits**

`take()` clears `DispatchedAuctionGuard` before `collect_stream_auction(...).await`. If the lazy body is dropped while collection is pending, the empty guard produces neither its promised warning nor terminal auction telemetry. The EOF path and non-HTML path have the same ordering.

This contradicts the guard's documented protection for drops at any await point and makes discarded SSP work/quota disappear from observability. Please retain a separate armed collection/drop sentinel through the await and disarm it only after collection reaches a terminal result. Async telemetry still cannot run from `Drop`, but the warning must survive cancellation. A test can poll into a deliberately pending platform `select()` and then drop the body.

ChristianPavilonis (changes requested):
## Review Summary

I found two high-impact issues and two medium-severity issues in the new Fastly streaming path. Valid deflate responses can be truncated, and enabling the supported Next.js integration still withholds every body byte until origin EOF, so issue #849's FCP objective is not met for that configuration. Valid multi-member gzip framing and cancellation-time auction observability also need correction.

I am requesting changes. The two earlier unresolved threads—bodiless streamed responses and decoded-limit enforcement only after a complete expansion—also remain current and are intentionally not duplicated here. All 19 CI checks are green, but the focused codec cases below are not covered by them.

ChristianPavilonis (commented):
## Review Summary

Reviewed PR #867 at `c8ae53f` against `main`. The lazy Fastly path is directionally sound, but two buffering points still undermine issue #849's FCP objective, and two HTTP/telemetry edge cases remain. All 19 CI checks currently pass; details are in the inline comments.

crates/trusted-server-core/src/streaming_processor.rs
```diff
@@ -349,11 +352,615 @@ impl StreamProcessor for StreamingReplacer {
...
+            )),
+            Compression::Deflate => Self::Deflate(flate2::write::ZlibEncoder::new(
+                Vec::new(),
+                flate2::Compression::default(),
+            )),
+            Compression::Brotli => Self::Brotli(Box::new(new_brotli_vec_encoder())),
+        }
+    }
+
+    pub(crate) fn encode_chunk(
```
ChristianPavilonis, line 1065: 🔧 **P1 — Recompression can withhold all browser-decodable content until origin EOF**

**Issue:** Each compressed `encode_chunk` arm calls `write_all()` and drains the inner `Vec`, but never flushes the codec. Gzip may emit only its 10-byte header, while deflate and Brotli may emit nothing until `finish()`. A focused probe using these writer APIs produced no usable compressed payload across roughly 1 MiB of compressible HTML before finalization.

**Why it matters:** Fastly commits response headers early, but browsers cannot render HTML from an empty stream or a gzip header alone. FCP can therefore still track the complete origin transfer for gzip, deflate, and Brotli pages. The existing tests drain the entire stream and prove final correctness, not progressive decodability.

**Suggested fix:** Flush each compressed encoder before draining its output, or use an explicit incremental/sync-flush policy. Add tests where the origin yields one compressed chunk and remains pending; the first response chunk must already decode to the rewritten document prefix without calling `finish()`.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -286,20 +454,443 @@ fn process_response_streaming<W: Write>(
...
+async fn hold_step_decoded_chunk<P: StreamProcessor>(
+    processor: &mut P,
+    encoder: &mut BodyStreamEncoder,
+    chunk: &[u8],
+    state: &mut AuctionHoldState,
+    collect_refs: &AuctionHoldCollectRefs<'_>,
+) -> Result<Vec<bytes::Bytes>, Report<TrustedServerError>> {
+    let mut segments = Vec::new();
+    if let Some(hold_buffer) = state.hold.as_mut() {
+        let ready = hold_buffer.push(chunk);
```
ChristianPavilonis: 🔧 **P1 — The prefix before `</body>` is retained across auction collection**

**Issue:** This prefix is processed into `segments`, but the function then awaits `collect_stream_auction()` before returning those segments to the lazy stream.

**Why it matters:** If `</body>` occurs in the first source chunk—common for small pages—the client receives no HTML until the auction finishes. For larger pages, the entire prefix from the close-containing chunk is delayed. This contradicts the contract that only the `</body>` tail is held and that the auction rides alongside transfer.

**Suggested fix:** Split the operation into phases: return/yield the ready prefix first, then await collection, then process and emit the held closing tail. Add a test with a deliberately pending auction and a first-chunk `</body>`; the prefix must become ready while collection remains pending.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -530,41 +1129,232 @@ pub async fn buffer_publisher_response_async(
...
+) -> Result<Response<EdgeBody>, Report<TrustedServerError>> {
+    match publisher_response {
+        PublisherResponse::Buffered(mut response) => {
+            // Fastly requests the origin body as a stream before the response is
+            // classified, so a buffered-unmodified response can still hold an
+            // `EdgeBody::Stream`. A bodiless response (HEAD, 204, 205, 304) must
+            // stay bodiless — `send_edgezero_response` streams any
+            // `EdgeBody::Stream` to the client — so drop the body while
+            // preserving metadata such as `Content-Length`.
+            if !response_carries_body(method, response.status()) {
```
ChristianPavilonis, line 1196: 🔧 **P2 — Bodiless normalization preserves invalid `Content-Length` framing**

**Issue:** Both finalizers clear bodies for 204/205 responses but preserve the origin's `Content-Length`, including a possible nonzero value.

**Why it matters:** RFC 9110 forbids `Content-Length` on 204, and a nonzero length on a now-empty 205 is known to be incorrect. Inconsistent downstream framing can cause interoperability failures and, across intermediaries, response-splitting risks.

**Suggested fix:** Remove `Content-Length` for 204 and remove it or normalize it to `0` for 205. Preserve representation length only where valid, such as HEAD and 304. Apply the same correction to the buffered finalizer and split the current combined regression test accordingly.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -530,41 +1129,232 @@ pub async fn buffer_publisher_response_async(
...
+                *response.body_mut() = body;
+            }
+            Ok(response)
+        }
+        PublisherResponse::Stream {
+            mut response,
+            body,
+            params,
+        } => {
+            if !response_carries_body(method, response.status()) {
```
ChristianPavilonis, line 1212: 🔧 **P2 — Bodiless streamed responses drop dispatched auctions without terminal telemetry**

**Issue:** A conditional navigation can dispatch an auction and receive a processable HTML 304. Classification selects `Stream`, but this branch only logs before returning; `params.dispatched_auction` is dropped without `emit_abandoned_auction`. The buffered equivalent has the same gap.

**Why it matters:** SSP work and quota consumption then have no terminal auction event, undermining Trusted Server's auction observability on a legitimate conditional-request path.

**Suggested fix:** Take the dispatched auction and await `emit_abandoned_auction(..., "bodiless_response")` using the retained observation before returning.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -233,6 +352,55 @@ struct ProcessResponseParams<'a> {
...
+    fn new(
+        params: &OwnedProcessResponseParams,
+        settings: &Settings,
+        integration_registry: &IntegrationRegistry,
+    ) -> Result<Self, Report<TrustedServerError>> {
+        let is_html = is_html_content_type(&params.content_type);
+        let is_rsc_flight =
+            content_type_contains_ascii_case_insensitive(&params.content_type, "text/x-component");
+        let inner: Box<dyn StreamProcessor> = if is_html {
+            Box::new(create_html_stream_processor(
```
aram356, line 370: ❓ **question — Next.js-enabled configs still deliver zero body bytes until origin EOF**

`PublisherBodyProcessor` wraps `create_html_stream_processor`, and when any HTML post-processor is registered (the `nextjs` integration), `HtmlWithPostProcessing::process_chunk` returns an empty vector for every non-final chunk. The lazy stream below then yields nothing until origin EOF, full-document post-processing, and recompression — even for pages where `should_process()` would return false. Headers commit early, but first body byte/FCP still tracks the complete origin transfer, so issue #849's objective is unmet for that configuration, and the PR description's "stream end-to-end" claim doesn't note the exception.

This was raised in the 2026-07-13 review and has no reply or change since. Is deferring it intentional?

If so, please: (1) call the limitation out in the PR body and the plan doc's "Out of scope" section, and (2) file a follow-up issue. The eventual fix is likely an up-front / streaming `should_process` gate so non-RSC pages skip accumulation entirely.

crates/trusted-server-core/src/streaming_processor.rs
```diff
@@ -144,7 +144,10 @@ impl<P: StreamProcessor> StreamingPipeline<P> {
         ) {
             (Compression::None, Compression::None) => self.process_chunks(input, output),
             (Compression::Gzip, Compression::Gzip) => {
-                let decoder = GzDecoder::new(input);
+                // Multi-member decoder: RFC 1952 permits concatenated gzip
+                // members, so a single-member reader would stop after the first.
+                // Matches the streaming `BodyStreamDecoder` gzip codec.
+                let decoder = MultiGzDecoder::new(input);
```
aram356: 🤔 **thinking — gzip trailing-garbage tolerance regressed, and gzip/deflate now diverge**

`main` used single-member `read::GzDecoder`, which stops at the end of the first member and silently ignores trailing bytes. `MultiGzDecoder` (correct per RFC 1952, and needed for concatenated members) instead treats trailing bytes as the start of a new member and errors when they aren't one — a 5xx on the buffered path, mid-body truncation after committed headers on the streaming path.

Meanwhile the deflate arm deliberately ignores trailing bytes ("matching the read-based decoder"), so the two codecs now differ in tolerance. Origins emitting junk after the gzip trailer are rare but exist (buggy padding/middleboxes). Consider tolerating trailing garbage once at least one member has fully decoded, or documenting the strictness change as intentional.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -286,20 +454,442 @@ fn process_response_streaming<W: Write>(
...
+    processor: &mut P,
+    encoder: &mut BodyStreamEncoder,
+    chunk: &[u8],
+    state: &mut AuctionHoldState,
+    collect_refs: &AuctionHoldCollectRefs<'_>,
+) -> Result<HoldStepSegments, Report<TrustedServerError>> {
+    let mut ready = Vec::new();
+    let bytes = match state.hold.as_mut() {
+        // Once the hold has been released the chunk streams straight through.
+        None => chunk.to_vec(),
```
aram356: ⛏ **nitpick — post-release chunks are copied needlessly**

Once the hold is released, every remaining decoded chunk goes through `chunk.to_vec()` even though `process_and_encode_chunk` only needs `&[u8]` — an extra 8 KiB allocation per chunk for the rest of the transfer on the hot streaming path.

**Fix**: borrow instead of copying, e.g.

```rust
let bytes: Cow<'_, [u8]> = match state.hold.as_mut() {
    None => Cow::Borrowed(chunk),
    Some(hold_buffer) => Cow::Owned(hold_buffer.push(chunk)),
};
```

aram356 (changes requested):
## Summary

Reviewed PR #867 at `5d6fc21f6` against `main`. All four findings from the 2026-07-14 review are fixed with one-to-one regression tests (per-chunk encoder flush, ready-prefix-before-collect, 204/205 framing, bodiless abandonment telemetry), and the earlier deflate/multi-gzip/decode-cap/guard findings all hold at head. One P1 from the 2026-07-13 review remains unaddressed and unanswered — post-processor configs still buffer the full document — plus one behavioral divergence and two smaller cleanups. All 19 CI checks pass.

## Blocking

### ❓ question

- **Next.js-enabled configs still deliver zero body bytes until origin EOF**: `HtmlWithPostProcessing` accumulates the whole rewritten document whenever a post-processor is registered, so the lazy stream emits nothing until EOF + post-process + recompress; #849's FCP objective is unmet for that configuration and the exception isn't documented (crates/trusted-server-core/src/publisher.rs:369 — see inline comment)

## Non-blocking

### 🤔 thinking

- **gzip trailing-garbage tolerance regressed vs main, and gzip/deflate now diverge**: `MultiGzDecoder` errors on trailing junk that single-member `GzDecoder` ignored, while the deflate arm deliberately ignores trailing bytes (crates/trusted-server-core/src/streaming_processor.rs:150 — see inline comment)

### ⛏ nitpick

- **Post-release chunks copied needlessly** in `hold_step_decoded_chunk` (crates/trusted-server-core/src/publisher.rs:731 — see inline comment)

### 🏕 camp site

- **`publisher.rs` is now ~7,300 lines**: the streaming machinery added here (`BodyChunkSource`, `AuctionHoldState`, `DispatchedAuctionGuard`, the `hold_*`/`passthrough_*` helpers, both finalizers) is a coherent unit that would read better as a `publisher/streaming.rs` submodule — not necessarily in this PR, but the file is past the point where navigation hurts

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests (fastly/axum/cloudflare/spin + parity): PASS
- js tests / format / docs format: PASS
- browser + integration tests: PASS

crates/trusted-server-core/src/publisher.rs
prk-Jr: Deferring is intentional. The post-processor path is inherently full-document today: the Next.js placeholder rewriter captures RSC payloads during the streaming pass and substitutes them at end-of-document, and `should_process()` consumes either that captured state or the complete document — a streaming gate is a design change rather than a patch for this PR.

Done as requested: the limitation is now called out in the PR description ("Known limitation" section) and in the plan doc's Out-of-scope section (666b3cb78), and #924 tracks the follow-up (up-front / streaming `should_process` gate so non-RSC pages skip accumulation entirely).

crates/trusted-server-core/src/streaming_processor.rs
prk-Jr: Implemented the suggested tolerance in 33384cd87. gzip decoding now goes through a shared `GzipStreamDecoder` used by both paths (`BodyStreamDecoder` directly, the buffered pipeline via a small `Read` adapter), so the two cannot drift. At each member boundary it sniffs the next bytes for the gzip magic number: a match starts the next member; anything else after at least one fully-decoded member is treated as trailing garbage and dropped — matching GNU gzip, main's single-member tolerance, and the deflate arm.

Integrity inside members is unchanged: truncated or corrupt members still error, and trailing bytes that happen to begin with the magic number are decoded as a member and fail if they are not one. Regression tests cover garbage in the same chunk, garbage split across chunks, a lone trailing `0x1f`, pseudo-member garbage (still errors), truncation (still errors), and the buffered pipeline path.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Fixed in 911268fb4 — post-release chunks are now `Cow::Borrowed`; only held chunks allocate via `Cow::Owned(hold_buffer.push(chunk))`.

prk-Jr:
Re the camp-site note on `publisher.rs` size: agreed on the shape — the streaming machinery (`BodyChunkSource`, `AuctionHoldState`, `DispatchedAuctionGuard`, the `hold_*`/`passthrough_*` helpers, both finalizers) is a coherent unit. Deferring the `publisher/streaming.rs` extraction so this PR stays reviewable; happy to do it as an immediate follow-up.

Summary of this round (33384cd87, 911268fb4, 666b3cb78):
- gzip trailing-garbage tolerance restored on both decode paths via a shared `GzipStreamDecoder` (33384cd87)
- post-release chunks borrowed instead of copied in the auction hold step (911268fb4)
- post-processor buffering exception documented in the plan doc Out-of-scope section (666b3cb78), the PR description, and tracked in #924

crates/trusted-server-core/src/streaming_processor.rs
```diff
@@ -347,11 +352,1032 @@ impl StreamProcessor for StreamingReplacer {
...
+    /// Decode as much of `chunk` as possible, draining any output the inflater
+    /// can still produce once all input is consumed.
+    ///
+    /// flate2 fills the output buffer up to its capacity, so a chunk that
+    /// exactly fills the buffer leaves decoded bytes (and possibly the
+    /// end-of-stream marker) pending with all input already consumed. The loop
+    /// keeps driving the inflater — reserving more output space — until it
+    /// makes no further progress, so those pending bytes are never stranded and
+    /// a valid stream is not mistaken for a truncated one at `finish`.
+    fn decode(&mut self, chunk: &[u8]) -> Result<Vec<u8>, Report<TrustedServerError>> {
```
ChristianPavilonis, line 852: 🔧 **P1 — Deflate decompression can allocate roughly twice the configured safety limit before rejecting the body**

`DeflateStreamDecoder::decode` grows `output`, lets `decompress_vec` fill the new capacity, and only then calls `self.charge(produced)`. Once output reaches the configured limit, the next `reserve` can double the `Vec` before the limit check rejects the produced bytes.

This defeats `max_buffered_body_bytes` as a hard decompression-bomb/Wasm-heap ceiling. A probe reproducing this exact loop with the current `flate2` grew the allocation to **33,554,432 bytes** before rejecting output at a **16,777,216-byte** limit.

Please inflate through a fixed-size scratch buffer bounded by the remaining budget, charge produced bytes before appending them, and error before reserving or producing beyond the limit. Add a deflate-bomb test that verifies peak buffered capacity, not merely eventual error reporting.

crates/trusted-server-core/src/streaming_processor.rs
```diff
@@ -347,11 +352,1032 @@ impl StreamProcessor for StreamingReplacer {
...
+    decoded: Vec<u8>,
+    position: usize,
+    finished: bool,
+}
+
+impl<R: Read> GzipDecodeReader<R> {
+    fn new(input: R) -> Self {
+        Self {
+            input,
+            decoder: GzipStreamDecoder::new(Rc::new(Cell::new(0)), usize::MAX),
```
ChristianPavilonis: 🔧 **P2 — The new buffered gzip reader materializes a large expansion before the downstream cap can act**

`GzipDecodeReader` gives `GzipStreamDecoder` an unlimited `usize::MAX` budget and decodes each 8 KiB compressed input block completely into `self.decoded` at line 668. `BoundedWriter` sees the data only afterward.

On buffered adapters, a few-kilobyte gzip body can therefore allocate several MiB even when an operator configured a much smaller `max_buffered_body_bytes`. The previous read-driven decoder emitted into the caller's small buffer rather than materializing the expansion.

Please make `GzipDecodeReader` produce decoded bytes incrementally through a fixed-size sink while retaining pending decoder/input state, or route buffered responses through the bounded `BodyStreamDecoder`. Add a low-limit compressed-bomb regression proving allocation does not precede rejection.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -1076,13 +1933,20 @@ async fn stream_html_with_auction_hold<W: Write, P: StreamProcessor>(
...
+            max_body_bytes,
+        )
+        .await;
+    }
 
-    let body = body_as_reader(body);
+    let body = body_as_reader(body)?;
     match compression {
         Compression::None => body_close_hold_loop(body, output, processor, ctx).await,
         Compression::Gzip => {
```
ChristianPavilonis, line 2039: 🔧 **P2 — Auction-enabled buffered adapters still discard valid concatenated gzip members**

The buffered HTML auction path immediately below still uses `flate2::read::GzDecoder`, which decodes only one gzip member. The new streaming and general buffered pipelines support RFC-valid concatenated members, but Axum/Cloudflare/Spin responses entering `stream_html_with_auction_hold` do not.

Everything after the first member is silently omitted, potentially including `</body>` and trailing markup. This creates valid-input data loss and adapter divergence. A focused probe confirmed that `read::GzDecoder` returned only the first member.

Please reuse the shared multi-member decoder in this branch—preferably by routing `Body::Once` through the same bounded hold pipeline as streamed bodies—and add an auction-enabled buffered-finalizer test with HTML split across two gzip members.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -3977,6 +5161,1157 @@ mod tests {
...
+                "should rewrite non-html stream after auction collection. Got: {css}"
+            );
+            assert!(
+                !css.contains("origin.example.com"),
+                "should not leave origin host after rewrite. Got: {css}"
+            );
+        });
+    }
+
+    fn drain_streaming_finalize_body(content_encoding: &str, body: EdgeBody) -> Vec<u8> {
```
ChristianPavilonis, line 6072: 🔧 **P2 — Tests do not prove that the real lazy response emits content before origin EOF**

This helper drains an immediately complete origin to EOF. The pre-finish codec test exercises `BodyStreamEncoder` in isolation. Neither test polls the actual `publisher_response_into_streaming_response` body while the origin remains pending.

The PR's primary acceptance criterion is improved first-byte/FCP behavior. A future await-before-yield or accidental rebuffering regression could pass every current unit test while restoring the original latency problem.

Please use an origin stream that yields one chunk and then stays `Pending`; assert the lazy response's first poll returns rewritten, browser-decodable content without EOF. Cover compressed HTML and the auction-hold path.

crates/trusted-server-core/src/settings.rs
```diff
@@ -50,15 +50,15 @@ pub struct Publisher {
...
-    /// On Fastly the *effective* ceiling for a publisher page is lower: the
-    /// platform HTTP client rejects any origin response whose raw (still
-    /// compressed) body exceeds 10 MiB before this buffer is ever filled, so
-    /// raising this value only helps highly compressible pages whose decoded
-    /// size exceeds the 16 MiB default while their compressed origin body stays
-    /// under 10 MiB. Raising it above ~10 MiB does not lift the platform cap for
-    /// uncompressed pages. That platform limit is removed once true streaming
-    /// lands (tracked for PR 15, issue #495), after which this setting becomes
-    /// the sole ceiling.
+    /// Fastly origin bodies are preserved as streams on the publisher path, so
```
ChristianPavilonis, line 53: 🔧 **P2 — Operator documentation still describes the old limit behavior**

`docs/guide/configuration.md` still says `max_buffered_body_bytes` applies only to fully buffered responses, that exceeding it becomes a 5xx, and that Fastly has a lower 10 MiB origin-body ceiling. All three statements are false after this change.

Operators may tune the setting without realizing it now caps raw and decoded Fastly streams and can truncate an already-committed production response. Please update the configuration table and detailed section to document the raw/decoded streaming limits, mid-stream truncation behavior, adapter differences, and removal of Fastly's old 10 MiB materialization ceiling.

ChristianPavilonis (commented):
## Review Summary

Reviewed PR #867 at `23f8e5be919ecd0ac6141936aef31f9531048018` against its actual `main` merge base. The end-to-end streaming design is directionally strong and all CI checks pass, but the new decoding paths still contain one high-severity memory-limit violation, plus buffered-adapter correctness, regression-coverage, and operator-documentation gaps. Details are inline.

crates/trusted-server-core/src/streaming_processor.rs
```diff
@@ -347,11 +352,1032 @@ impl StreamProcessor for StreamingReplacer {
...
+                    let consumed = decoder.write(input)?;
+                    if consumed > 0 {
+                        input = &input[consumed..];
+                        continue;
+                    }
+                    // A zero-byte write on non-empty input means the member
+                    // (deflate stream plus trailer) is complete: validate the
+                    // trailer and start sniffing for the next member.
+                    let GzipStreamState::Member(decoder) =
+                        std::mem::replace(&mut self.state, GzipStreamState::Poisoned)
```
aram356, line 560: ⛏ **nitpick — the `Poisoned` panic is reachable after a mid-transition error**

The `?` on `decoder.finish()` here (and on `write_all(&GZIP_MAGIC)` in the boundary→member transition below) returns while `state` is still `Poisoned`. If a caller ever invokes `decode`/`finish` again after that error, it hits `unreachable!` — which in Wasm aborts the instance mid-response — even though the panic message claims the state "is never left in place".

No current caller retries after an error, so this is latent, but the invariant lives entirely in caller convention. Returning an error from the `Poisoned` arms is strictly safer:

```rust
GzipStreamState::Poisoned => {
    return Err(io::Error::other(
        "gzip decoder previously failed; stream is unusable",
    ));
}
```

aram356 (changes requested):
## Summary

Re-reviewed at `e46085a84` against `main`. All three findings from my previous review are properly resolved: the post-processor buffering deferral is documented (PR body + plan doc) with follow-up #924 filed; trailing-garbage tolerance landed via the shared `GzipStreamDecoder` with member-boundary sniffing and thorough split-point tests (33384cd87); and the post-release chunk copy is gone (911268fb4). I'm resolving those threads.

The five findings from the 2026-07-21 review remain unaddressed at this head — the only pushes since are main merges. I verified each one independently against the current code rather than relaying them, and all five reproduce; they are restated below with my verification notes (no duplicate threads — please follow up on the existing comments). One new minor finding from this pass is inline.

## Blocking — open findings from the previous review, independently verified

- **Deflate decode can allocate ~2× the configured cap** ([existing thread](https://github.com/IABTechLab/trusted-server/pull/867#discussion_r3625118765)): confirmed at `streaming_processor.rs:738` and `:778` — when `output.len() == capacity` at the cap, `reserve(STREAM_CHUNK_SIZE)` doubles the `Vec` (16 MiB → 32 MiB) and `decompress_vec` fills the new spare capacity before `charge()` runs. A fixed-size scratch bounded by the remaining budget closes it.
- **`GzipDecodeReader` materializes unbounded per-block expansion** ([existing thread](https://github.com/IABTechLab/trusted-server/pull/867#discussion_r3625118769)): confirmed at `streaming_processor.rs:638` — `usize::MAX` budget, and each 8 KiB compressed block decodes fully into `self.decoded` (up to ~8 MiB at gzip's max ratio) before `BoundedWriter` sees a byte, where the replaced `read::MultiGzDecoder` decoded incrementally into 8 KiB caller buffers.
- **Buffered auction path still decodes single-member gzip** ([existing thread](https://github.com/IABTechLab/trusted-server/pull/867#discussion_r3625118773)): confirmed at `publisher.rs:1953` — `stream_html_with_auction_hold`'s buffered arm still uses `read::GzDecoder`, silently dropping every member after the first (potentially including `</body>`) on Axum/Cloudflare/Spin, while both other pipelines are now multi-member. `GzipDecodeReader` slots in directly.
- **No test proves early emission on the real lazy body** ([existing thread](https://github.com/IABTechLab/trusted-server/pull/867#discussion_r3625118778)): confirmed — `drain_streaming_finalize_body` drains completed origins only; nothing polls the `publisher_response_into_streaming_response` body while the origin stays `Pending`, so an await-before-yield regression would pass every current test despite being the PR's headline claim.
- **`docs/guide/configuration.md` still documents the old limit behavior** ([existing thread](https://github.com/IABTechLab/trusted-server/pull/867#discussion_r3625118784)): confirmed at lines 301–323 — buffered-only scope, 5xx-on-excess, and the 10 MiB Fastly origin ceiling are all inaccurate for the new streaming path (raw + decoded stream caps, mid-body truncation after committed headers, platform cap removed).

## Non-blocking

### ⛏ nitpick

- **`GzipStreamDecoder` poisoned-state panic reachable after a mid-transition error** (`streaming_processor.rs:515` — see inline comment)

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests (fastly/axum/cloudflare/spin + parity): PASS
- js tests / format / docs format: PASS
- browser + integration tests: PASS

crates/trusted-server-core/src/streaming_processor.rs
prk-Jr: Fixed in 69a2e7028.

`DeflateStreamDecoder::decode` and `finish` no longer inflate into a growing `Vec`. Both now decompress through a fixed `[u8; STREAM_CHUNK_SIZE]` scratch buffer whose write window is bounded by the remaining budget plus one byte (`decode_window`), charge the produced bytes *before* appending them, and append via `extend_capped`, which keeps amortized doubling but caps the reservation at `max_decoded_bytes + 1`. A bomb therefore produces at most one byte past the ceiling, and `charge` rejects it before that byte is buffered — nothing is reserved or filled past the limit.

Two tests, both asserting more than eventual error reporting:

- `deflate_decode_caps_buffer_capacity_at_decoded_limit` decodes a body sitting exactly at the cap and asserts the returned buffer's **capacity** stays `<= cap + STREAM_CHUNK_SIZE`, i.e. it never balloons toward ~2×.
- `deflate_decode_rejects_bomb_before_expanding_past_limit` decodes a member expanding to 64× the cap and asserts the cumulative-cap error.

crates/trusted-server-core/src/streaming_processor.rs
prk-Jr: Fixed in 69a2e7028.

`GzipDecodeReader::new` now takes a `max_decoded_bytes` budget instead of hardcoding `usize::MAX`, so each block's expansion is charged against `GzipStreamDecoder`'s bounded sink from inside the decompressor writes. `StreamingPipeline::with_max_decoded_bytes` threads it through, and `process_response_streaming` passes `settings.publisher.max_buffered_body_bytes` — the same ceiling `BoundedWriter` enforces downstream. A bomb is now rejected mid-decode instead of being fully materialized before the writer can act.

I kept the read-adapter shape rather than routing buffered responses through `BodyStreamDecoder`: `GzipDecodeReader` is also reached via `StreamingPipeline` from non-publisher callers (e.g. `proxy.rs`), so bounding the reader itself covers those too, and it let the buffered auction arm reuse the same decoder (see the concatenated-members thread).

Regression: `gzip_decode_reader_rejects_bomb_before_materializing_past_limit` reads a member expanding to 512× an 8 KiB cap, asserting the read fails with the decoded-cap error and that no more than the cap was emitted.

For the record, the deflate and brotli read decoders were already incremental — they emit into the caller's fixed-size buffer — so the gzip reader was the only one materializing ahead of the cap.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Fixed in 69a2e7028.

`stream_html_with_auction_hold`'s buffered arm now uses `GzipDecodeReader::new(body, max_body_bytes)` instead of `flate2::read::GzDecoder`, so it decodes concatenated RFC 1952 members and shares the streaming path's trailing-garbage tolerance. It also inherits the decode budget, wired to `ctx.settings.publisher.max_buffered_body_bytes` to match the streaming arm and the no-hold buffered path. `GzDecoder` is no longer imported outside tests.

Test: `stream_publisher_body_async_auction_hold_decodes_multi_member_gzip_buffered` drives an auction-enabled, buffered `Body::Once` body whose HTML is split across two gzip members, with `</body>` in the **second**. It asserts the second member's `</body></html>` survives and that bids inject at that close tag — both fail with a single-member decoder, so the test discriminates on exactly the dropped-member behavior.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Fixed in 69a2e7028. Added two tests that poll the real `publisher_response_into_streaming_response` body instead of draining an already-complete origin.

`origin_chunk_then_pending` yields exactly one chunk and then returns `Poll::Pending` forever, never signalling EOF. `first_lazy_body_chunk` polls the lazy `Body::Stream` once with a noop waker and **panics on `Pending`**, so an await-before-yield or accidental rebuffering regression fails the test rather than passing it.

- `streaming_finalize_emits_compressed_html_before_origin_eof` — gzip HTML; asserts the first poll yields gzip bytes that decode (flushed, not finished — the way a browser decodes an open response) to the streamed content.
- `streaming_finalize_auction_hold_emits_prefix_before_origin_eof` — auction-hold path; asserts the first poll already carries the injected head and body content, and does **not** yet contain `.bids=JSON.parse`, proving the prefix streams before the auction is collected rather than after.

One finding worth flagging: the compressed test needs a realistically sized page. `flate2::write::GzDecoder` buffers small members internally, so a few-hundred-byte gzip body genuinely emits nothing until `finish()`. The test uses a ~98 KiB page so the first decoded block flushes downstream, as a real page does.

crates/trusted-server-core/src/settings.rs
prk-Jr: Fixed in 69a2e7028, in `docs/guide/configuration.md`.

The `max_buffered_body_bytes` section now separates the two scopes: buffered adapters (Axum, Cloudflare, Spin) cap the decoded, post-rewrite output buffer, while the Fastly streaming path caps the stream twice over — cumulative *raw* (still compressed) bytes pulled from origin, and cumulative *decoded* bytes — with the decoded cap enforced during decompression so a bomb is rejected before its expansion is materialized.

A new "Behavior when exceeded" block records the adapter difference you flagged: buffered adapters fail before any bytes are committed, whereas on the streaming path headers are already committed, so the body is **truncated mid-stream and logged**, explicitly not a `5xx`, with a note to size the cap above the largest expected decoded page.

The 10 MiB Fastly origin-body ceiling is removed; the default note now states this setting is the sole ceiling on that path since origin bodies are streamed rather than materialized in full. The `[publisher]` table row was updated to match.

crates/trusted-server-core/src/streaming_processor.rs
prk-Jr: Fixed in 69a2e7028. Both `Poisoned` arms now return an error instead of panicking:

```rust
return Err(io::Error::other(
    "gzip decoder previously failed; stream is unusable",
));
```

in `decode`'s loop arm, and the equivalent `Err(...)` in `finish`.

Your read of the mechanism is right: the `?` on `decoder.finish()` and on `write_all(&GZIP_MAGIC)` both return while the state is still `Poisoned`, so the invariant held only by caller convention — a retry would have hit `unreachable!` and aborted the Wasm instance mid-response.

I left `take_decoded`'s arm as `unreachable!`: it is only called at the end of `decode`, and the loop's `Poisoned` arm now returns before it can be reached, so it is genuinely unreachable rather than convention-dependent.

crates/trusted-server-core/src/streaming_processor.rs
```diff
@@ -347,11 +375,1179 @@ impl StreamProcessor for StreamingReplacer {
...
+                io::Error::other("publisher origin body decoded byte count overflowed")
+            })?;
+        if next > self.max_decoded_bytes {
+            return Err(io::Error::other(format!(
+                "publisher origin body decoded size exceeded {}-byte streaming limit",
+                self.max_decoded_bytes
+            )));
+        }
+        self.decoded_bytes.set(next);
+        self.buffer.extend_from_slice(data);
```
aram356: 🤔 **thinking — the gzip/brotli decode sink can still commit ~2× the cap transiently**

`BoundedDecodeSink::write` charges correctly, but appends with plain `extend_from_slice`, whose amortized doubling can take the buffer's capacity to ~2× the ceiling (plus the realloc copy of the existing bytes) as a bomb approaches the cap — and reserved `Vec` capacity is committed Wasm heap. This is the same mechanism 69a2e7028 eliminated for deflate with `extend_capped`, so it's the last gap in the newly documented "rejected before its expansion is materialized" guarantee.

Transient and bounded at 2×, so not blocking — but it's a two-line change for consistency:

```rust
self.decoded_bytes.set(next);
extend_capped(&mut self.buffer, data, self.max_decoded_bytes);
Ok(data.len())
```

crates/trusted-server-core/src/streaming_processor.rs
```diff
@@ -347,11 +375,1179 @@ impl StreamProcessor for StreamingReplacer {
...
+
+    /// Take the decoded bytes accumulated in the sink so far.
+    fn take_decoded(&mut self) -> Vec<u8> {
+        match &mut self.state {
+            GzipStreamState::Member(decoder) => std::mem::take(&mut decoder.get_mut().buffer),
+            GzipStreamState::Boundary { sink, .. } | GzipStreamState::TrailingGarbage(sink) => {
+                std::mem::take(&mut sink.buffer)
+            }
+            GzipStreamState::Poisoned => {
+                unreachable!("poisoned state is never left in place across calls")
```
aram356: ⛏ **nitpick — `decode(&[])` on a poisoned decoder still panics**

The new error-returning `Poisoned` arms live inside the `while !input.is_empty()` loop, so an **empty** chunk skips the loop entirely and falls through to `take_decoded`, whose `Poisoned` arm is still `unreachable!`. That's the same re-entrant-caller premise 69a2e7028 just addressed: error → retry with an empty chunk → Wasm abort. (Empty stream chunks are legal for `BodyChunkSource` to yield.)

Returning `Vec::new()` from that arm finishes the job:

```rust
GzipStreamState::Poisoned => Vec::new(),
```

aram356 (approved):
## Summary

Re-reviewed at `75fd92793` against `main`. All six findings from the previous round are verified fixed in `69a2e7028` — checked in code, not just from the replies:

- Deflate decode/finish inflate through a fixed scratch window bounded by the remaining budget, charge before append, and cap the accumulator's reservation at the ceiling (`extend_capped`); at-cap capacity and bomb tests cover it.
- `GzipDecodeReader` takes a decode budget (threaded via `StreamingPipeline::with_max_decoded_bytes` from `max_buffered_body_bytes`), rejecting a bomb mid-decode with ≤ cap bytes emitted.
- The buffered auction hold path decodes multi-member gzip through the shared reader; the new test carries `</body>` in the second member and asserts bid injection.
- Two pending-origin tests poll the real lazy body once and fail on `Pending`, pinning the early-emission claim (compressed HTML and auction-hold prefix) against rebuffering regressions.
- `configuration.md` now accurately documents the dual raw/decoded stream caps, the buffered-vs-streaming failure modes, and the removal of the 10 MiB Fastly ceiling.
- The `Poisoned` arms in `decode`/`finish` return errors instead of panicking.

Two minor residuals from this pass are inline — both are one/two-line consistency follow-ups to the same hardening, fine to address here or immediately after.

## Non-blocking

### 🤔 thinking

- **Gzip/brotli decode sink can still commit ~2× the cap transiently**: `BoundedDecodeSink::write` appends via plain `extend_from_slice`, so amortized doubling can push committed capacity to ~2× the ceiling before the charge rejects — the mechanism `extend_capped` just eliminated for deflate (crates/trusted-server-core/src/streaming_processor.rs:460 — see inline comment)

### ⛏ nitpick

- **`decode(&[])` on a poisoned decoder still panics**: empty input skips the loop's error-returning `Poisoned` arms and reaches `take_decoded`'s `unreachable!` (crates/trusted-server-core/src/streaming_processor.rs:646 — see inline comment)

## CI Status

- fmt: PASS
- clippy: PASS
- rust/js/docs checks: 14 of 16 PASS; 2 still in progress at submission time (long-running test jobs; the identical tree passed fully at the pre-merge commit `69a2e7028`)

crates/trusted-server-core/src/proxy.rs
```diff
@@ -274,7 +277,7 @@ pub async fn stream_asset_body<W: Write>(
                 output
```
ChristianPavilonis, line 277: 🤔 **P1 — Fastly can buffer yielded publisher chunks instead of immediately sending them**

`stream_asset_body()` calls `write_all()` for each yielded chunk but never `flush()`. Fastly SDK 0.12.1 implements `StreamingBody` using `BufWriter<StreamingBodyHandle>` and documents that `flush()` is required to explicitly emit a client chunk.

Compressed publisher segments are commonly smaller than the writer's buffer. They can remain buffered while the generator awaits more origin data or auction collection. Headers commit early, but the browser receives no body content, defeating compressor sync-flushing and delaying FCP. Existing lazy-body tests drain into memory and bypass this real Fastly bridge.

Suggested follow-up: flush `output` after each stream chunk, with appropriate error context, and add a buffered-writer test proving the first yielded segment is flushed before polling the origin or auction again.

crates/trusted-server-core/src/streaming_processor.rs
```diff
@@ -347,11 +375,1242 @@ impl StreamProcessor for StreamingReplacer {
...
+    /// # Errors
+    ///
+    /// Returns an error if a member is corrupt, its trailer fails validation,
+    /// or the decoded budget is exceeded.
+    fn decode(&mut self, chunk: &[u8]) -> io::Result<Vec<u8>> {
+        let mut input = chunk;
+        while !input.is_empty() {
+            match &mut self.state {
+                GzipStreamState::Member(decoder) => {
+                    let consumed = decoder.write(input)?;
```
ChristianPavilonis, line 551: 🤔 **P1 — The gzip decoder does not release small decoded prefixes per origin chunk**

`GzipStreamDecoder::decode()` writes compressed input into `flate2::write::GzDecoder` and immediately drains the sink without flushing the decoder. `GzDecoder` can retain small decoded output internally until `flush()` or `finish()`.

A gzip origin chunk that expands to a modest HTML prefix can therefore produce an empty `Bytes`, causing the lazy response to poll the origin again instead of yielding renderable content. The progressive gzip test deliberately uses a large page that exceeds the decoder's internal buffering, so it misses this case.

Suggested follow-up: flush the active gzip member after consuming each source chunk before `take_decoded()`, while retaining `try_finish()` for integrity validation. Add a small gzip `origin_chunk_then_pending` test whose first poll must yield content.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -286,20 +461,438 @@ fn process_response_streaming<W: Write>(
...
+    let mut segments = Vec::new();
+
+    let decoded_tail = match decoder.finish() {
+        Ok(decoded_tail) => decoded_tail,
+        Err(err) => {
+            abandon_hold_auction(state, collect_refs.services, "stream_decode_error").await;
+            return Err(err);
+        }
+    };
+    if !decoded_tail.is_empty() {
```
ChristianPavilonis: 🤔 **P1 — Decoder-finalized prefixes are retained across auction collection**

`hold_finish_segments()` adds the ready prefix returned from `hold_step_decoded_chunk()` to `segments`, then awaits `hold_collect_close_tail()` before returning any segments to the lazy stream.

Whenever a codec emits document bytes only from `decoder.finish()`—notably small gzip bodies under the current decoder behavior—the client receives no body until both origin EOF and auction completion. This violates the intended invariant that only the closing `</body>` tail waits for auction collection.

Suggested follow-up: return the decoder-finalized ready prefix as a separate pre-collection stage so the lazy caller can yield it before awaiting collection. Add a compressed small-document test with a pending auction.

crates/trusted-server-core/src/streaming_processor.rs
```diff
@@ -144,7 +167,12 @@ impl<P: StreamProcessor> StreamingPipeline<P> {
         ) {
             (Compression::None, Compression::None) => self.process_chunks(input, output),
             (Compression::Gzip, Compression::Gzip) => {
-                let decoder = GzDecoder::new(input);
+                // Multi-member decoder: RFC 1952 permits concatenated gzip
+                // members, so a single-member reader would stop after the first.
+                // Shares `GzipStreamDecoder` with the streaming `BodyStreamDecoder`
+                // gzip codec, so both paths tolerate trailing garbage after the
+                // final member the same way.
+                let decoder = GzipDecodeReader::new(input, self.max_decoded_bytes);
```
ChristianPavilonis: 🤔 **P2 — The buffered gzip path turns the configured cap into a codec-specific decoded-input limit**

Buffered publisher processing passes `max_buffered_body_bytes` into `GzipDecodeReader` as a cumulative decoded-input ceiling. Buffered deflate and Brotli readers do not impose the equivalent limit; they stream decoded chunks to the processor and rely on the downstream output cap.

A valid gzip response whose decoded input exceeds the cap but whose rewrite shrinks below it can now fail only when gzip-encoded, while identity, deflate, and Brotli versions can succeed. This also differs from the configuration guide's description of the buffered-adapter setting as the post-rewrite output cap.

Suggested follow-up: bound the gzip reader's pending decode buffer rather than cumulative decoded input, or introduce and document a separate cumulative decoded-input limit applied consistently to every encoding.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -535,46 +1144,290 @@ pub async fn buffer_publisher_response_async(
...
+/// The response keeps no body, and its `Content-Length` is corrected where the
+/// origin's value would be invalid for the now-empty message:
+/// - **204 No Content**: RFC 9110 §8.6 forbids `Content-Length`; remove it.
+/// - **205 Reset Content**: the reset response carries nothing, so a nonzero
+///   origin length is wrong (RFC 9110 §15.4.6); normalize it to `0`.
+/// - **HEAD and 304 Not Modified**: `Content-Length` legitimately advertises
+///   the `GET` representation length, so it is preserved untouched.
+fn make_response_bodiless(response: &mut Response<EdgeBody>) {
+    *response.body_mut() = EdgeBody::empty();
+    match response.status() {
```
ChristianPavilonis, line 1435: 🤔 **P2 — Bodiless normalization can preserve invalid `Transfer-Encoding` framing**

`make_response_bodiless()` removes or rewrites `Content-Length` but leaves an origin `Transfer-Encoding` untouched. For 204, `Transfer-Encoding` is prohibited; for 205, inserting `Content-Length: 0` while retaining `Transfer-Encoding` creates conflicting framing. Fastly's compatibility conversion copies response headers verbatim.

Suggested follow-up: remove `Transfer-Encoding` for normalized 204/205 responses, along with inapplicable trailer metadata, and extend the framing tests with an origin `Transfer-Encoding: chunked`.

ChristianPavilonis (approved):
## Summary

Approving this change. The streaming architecture is directionally strong and all 19 CI checks pass at `18fe060f`. I have noted five concrete follow-up risks inline, following the repository's CREG review guidance; they are non-blocking for this approval.

PR #870, opened 2026-07-08
Document the config template and harden config validation

### LINKED ISSUE #869: Document trusted-server.example.toml and restore dropped-but-supported config sections
## Summary

`trusted-server.example.toml` is the source-controlled config template (`ts config init`, and embedded into the Cloudflare/Spin adapters via `include_str!`). Over time it drifted from the supported config schema: it dropped several sections that still exist in the Rust config structs, and it doesn't clearly distinguish the required minimum config from optional/vendor settings.

## Problems

- **No clear "minimum config".** Nothing in the file signals that only `[publisher]`, `[ec].passphrase`, and an admin-covering `[[handlers]]` are required to boot/validate. `[ec]` in particular is easy to assume optional, but an omitted/empty passphrase fails validation.
- **Dropped-but-still-supported sections.** The following map to live structs yet were missing from the template: `[tester_cookie]`, `[rewrite]`, the `[consent]` tree, `[image_optimizer]`, `[tinybird]`, `[integrations.osano]`, `publisher.max_buffered_body_bytes`, `sourcepoint.auth_cookie_name`, DataDome protection fields, Prebid override rules.
- **Sparse descriptions.** Most keys have no explanation of what they do or their defaults.

## Proposed change

- Keep the required trio active with per-key descriptions.
- Comment out every optional section/integration, each with a one-line description, so a fresh deploy starts minimal and safe.
- Re-add the dropped-but-supported sections as documented (commented) blocks.
- Keep the four `ts audit`-patched integrations (`gpt`, `didomi`, `datadome`, `google_tag_manager`) active with `enabled = false` so the audit CLI can still flip them.
- Keep all example hosts as `example.com` (no real vendor domains).

## Acceptance

- File deserializes and passes `Settings::from_toml` validation (verified with real secrets substituted for the intentional placeholders).
- `ts audit` draft-config patching still works (`run_audit_writes_selected_outputs_and_summary`).
- Config schema in the file matches the actual Rust structs.

### LINKED ISSUE #871: Harden deploy validation: reject fail-open publisher hosts, APS pub_id, and request-signing store ids
## Summary

Follow-up from #869 / PR #870. Several config fields **fail open** at `ts config validate` / deploy validation: template placeholder values (or trivially-fake values) pass validation and only fail later at runtime / key-rotation time. The secret placeholders (EC passphrase, proxy_secret, admin handler password, partner API tokens) are already rejected; these are the remaining gaps.

None of these are security holes on their own — a wrong domain or fake pub_id simply won't work at runtime — but rejecting them at config-push time gives operators a much faster, clearer failure.

## Gaps

1. **Reserved example publisher hosts pass validation** (High).
   `[publisher]` `domain = "example.com"`, `cookie_domain = ".example.com"`, `origin_url = "https://origin.example.com"` validate OK. An operator who changes the three secret placeholders but forgets the publisher block deploys against `example.com`.
   - Tension to resolve: CLAUDE.md mandates `example.com` in all examples, and the test suite uses `example.com` pervasively. Any rejection must distinguish "unedited template default" from legitimate test/example usage (e.g. only reject in `validate_settings_for_deploy`, and/or a dedicated reserved-host list mirroring the secret-placeholder approach). Needs its own tests.

2. **`ApsConfig.pub_id` has no non-empty/placeholder validation** (Medium).
   With `[integrations.aps] enabled = true`, `pub_id = "your-aps-publisher-id"` validates (only custom deserialization runs, no `#[validate]`). Add a non-empty + placeholder check.

3. **`RequestSigning` has no `Validate` derive / placeholder rejection** (Low).
   `config_store_id` / `secret_store_id` accept any string (e.g. `<management-config-store-id>`), so a misconfigured request-signing block validates and fails later at key rotation instead of at push.

## Acceptance

- `ts config validate` / deploy validation rejects each of the above with a clear message, without breaking the many `example.com`-based tests or the example config embedded in the Cloudflare/Spin adapters.
- Once rejection lands, update `trusted-server.example.toml` comments accordingly (the template currently notes these as fail-open).

## Context

- Placeholder-secret rejection lives in `Settings::reject_placeholder_secrets` / `validate_settings_for_deploy` (`config.rs`) and the per-type `*_PLACEHOLDERS` lists (`settings.rs`).
- Handler password rejection: `Handler::PASSWORD_PLACEHOLDERS`.
- GTM already validates its container id via `GTM_CONTAINER_ID_PATTERN` — a good model for pub_id.

aram356:
Closes #869. Closes #871.

Config-surface cleanup in four focused commits: document the app-config template, close deploy-validation gaps that let placeholder values through, and lean the config structs onto the `validator` crate's built-in validators instead of hand-rolled functions.

---

## 1 · Document `trusted-server.example.toml` and restore supported sections

Reworks the source-controlled app-config template (used by `ts config init`, embedded into the Cloudflare/Spin adapters via `include_str!`, and text-patched by `ts audit`).

- Required minimum stays active + documented: `[[handlers]]` (admin), `[publisher]`, `[ec]`.
- Every optional section/integration is commented out with a one-line description.
- Restores sections that still map to live config structs but had been dropped: `[tester_cookie]`, `[rewrite]`, `[consent]`, `[image_optimizer]`, `[tinybird]`, `[integrations.osano]`, `publisher.max_buffered_body_bytes`, `sourcepoint.auth_cookie_name`, DataDome protection fields, Prebid override rules.
- Keeps `gpt`/`didomi`/`datadome`/`google_tag_manager` as active `enabled = false` stubs so `ts audit` can flip them in place. All hosts are `example.com`.

## 2 · Reject fail-open placeholder config values at deploy validation

`ts config validate`/`push` route through `TrustedServerAppConfig::validate` → `validate_settings_for_deploy`. That path now rejects template placeholders that previously validated OK and only failed later at runtime:

- `publisher.domain` / `cookie_domain` / `origin_url` left at the `example.com` template defaults.
- `request_signing.config_store_id` / `secret_store_id` when the block is **enabled** (empty or the `<management-...>` placeholders).
- `integrations.aps.pub_id` when APS is **enabled** — non-empty via the built-in `length(min = 1)` validator, plus a custom validator for the reserved `your-aps-publisher-id` placeholder.

Why some checks use `#[validate]` and some don't: **APS** validates lazily through `get_typed` (enabled-gated), so an attribute fires only when APS is on. **Publisher / request_signing** stay in the deploy-only `reject_placeholder_secrets` — `Publisher::validate()` runs at *parse* time on the embedded `example.com` template (an attribute would reject the template itself), and request_signing rejection must be gated on the sibling `enabled` flag. This mirrors how the existing placeholder-*secret* rejection already works.

## 3 · Replace hand-rolled GTM/Prebid validators with the built-in `regex` validator

- **GTM `container_id`** and **Prebid `external_bundle_sha256`** used custom validator functions that just ran a regex / hex-format check. Both now use `#[validate(regex(...))]` (validator 0.20 ships regex support + an `AsRegex` impl for `LazyLock<Regex>`); operator-facing messages preserved via `message`.
- Added a test per validator asserting the accepted and rejected inputs.

## 4 · Bound `timeout_ms` with the built-in `range` validator

testlight/datadome already constrain their timeouts via `range`; `aps`, `prebid`, `adserver_mock`, and `auction` did not. Added `range(min = 1, max = 60000)` (catches `0` and absurd values) for consistency. `AuctionConfig` now derives `Validate` and is wired via `#[validate(nested)]`; all existing configs use 500–2000 ms.

---

## Intentionally left imperative

Tinybird, image-optimizer, proxy asset-route, creative-opportunities, and consent validators were **not** converted to attributes. Their logic interleaves `normalize()` mutation, charset checks, `enabled`-gating, cross-field rules, and contextual per-item error messages (e.g. "slot `X` must have positive width") with the one or two convertible checks. Converting them would fragment validation across `#[validate]` attrs and `prepare_runtime`/`validate_runtime`, cascade `Validate` derives up the parent chains, and downgrade contextual errors to generic ones — a regression, not a win. (`consent` additionally *clamps* rather than rejects, so a `range` attribute would change its behavior.) That code isn't reimplementing a built-in; it does things built-ins can't express.

## Verification

- `cargo test -p trusted-server-core --lib` → **1631 pass**, 0 fail (incl. new deploy-validation, GTM/Prebid regex, and timeout-range tests).
- `cargo test --package trusted-server-cli … audit` → pass (`ts audit` still patches the template).
- `Settings::from_toml` on the template → pass (deploy rejection is separate from parse).
- `cargo clippy -p trusted-server-core --lib` with `-D warnings` → clean.
- Integration fixture timeout values confirmed within range.

**Not yet run:** the full `clippy-fastly` / integration-parity CI gates — the `clippy-fastly` alias hits a pre-existing local env error compiling the JS build script, unrelated to these changes.

---

## Upgrade note

A present `[request_signing]` block now requires real store IDs **even when `enabled = false`**. The key rotate/deactivate admin routes are registered unconditionally and read these IDs regardless of `enabled`, so deploy validation now rejects placeholder or empty store IDs whenever the block exists. Operators who declared `[request_signing]` with `enabled = false` and left placeholder IDs will get a `ts config validate`/`push` failure naming `request_signing.config_store_id` / `secret_store_id`. To keep signing off with no store IDs, leave the entire `[request_signing]` block commented out (the template documents this).

### DIFF
--- a/crates/trusted-server-core/src/integrations/aps.rs
+++ b/crates/trusted-server-core/src/integrations/aps.rs
@@ -142,6 +142,7 @@ pub struct ApsConfig {
     pub endpoint: String,
     /// Timeout in milliseconds.
     #[serde(default = "default_timeout_ms")]
+    #[validate(range(min = 1, max = 60000))]
     pub timeout_ms: u32,
     /// Whether to include the APS HTTP exchange in auction response metadata.
     ///
--- a/crates/trusted-server-core/src/settings.rs
+++ b/crates/trusted-server-core/src/settings.rs
@@ -123,6 +123,51 @@ impl Publisher {
             .any(|p| p.eq_ignore_ascii_case(proxy_secret))
     }
 
+    /// Reserved example publisher values copied verbatim from the config
+    /// template. They deserialize fine but must be replaced before deploying.
+    const PLACEHOLDER_DOMAINS: &[&str] = &["example.com"];
+    const PLACEHOLDER_COOKIE_DOMAINS: &[&str] = &[".example.com"];
+    /// Reserved example origin hosts. Matched against the parsed URL host so a
+    /// spelling that resolves to the same host (an explicit `:443`, a trailing
+    /// slash, a different scheme) cannot slip past the placeholder check.
+    const PLACEHOLDER_ORIGIN_HOSTS: &[&str] = &["origin.example.com"];
+
+    /// Returns `true` if `domain` is the unedited template placeholder
+    /// (case-insensitive).
+    #[must_use]
+    pub fn is_placeholder_domain(domain: &str) -> bool {
+        Self::PLACEHOLDER_DOMAINS
+            .iter()
+            .any(|p| p.eq_ignore_ascii_case(domain.trim()))
+    }
+
+    /// Returns `true` if `cookie_domain` is the unedited template placeholder
+    /// (case-insensitive).
+    #[must_use]
+    pub fn is_placeholder_cookie_domain(cookie_domain: &str) -> bool {
+        Self::PLACEHOLDER_COOKIE_DOMAINS
+            .iter()
+            .any(|p| p.eq_ignore_ascii_case(cookie_domain.trim()))
+    }
+
+    /// Returns `true` if `origin_url` resolves to an unedited template
+    /// placeholder host (case-insensitive).
+    ///
+    /// The comparison is on the parsed URL host, not the raw string, so
+    /// equivalent spellings of the reserved host - an explicit default port, a
+    /// trailing slash, or a different scheme - are all rejected.
+    #[must_use]
+    pub fn is_placeholder_origin_url(origin_url: &str) -> bool {
+        Url::parse(origin_url.trim())
+            .ok()
+            .and_then(|url| url.host_str().map(str::to_owned))
+            .is_some_and(|host| {
+                Self::PLACEHOLDER_ORIGIN_HOSTS
+                    .iter()
+                    .any(|p| p.eq_ignore_ascii_case(&host))
+            })
+    }
+
     /// Extracts the host (including port if present) from the `origin_url`.
     ///
     /// # Examples
@@ -232,6 +277,15 @@ impl IntegrationSettings {
             },
         )?;
 
+        // Field validation runs only for integrations that resolve to enabled.
+        // An integration whose `enabled` flag is omitted falls back to its
+        // serde default, which the explicit-`false` fast path above cannot
+        // observe. Validating before this check would reject documented
+        // template placeholders in sections that are not actually turned on.
+        if !config.is_enabled() {
+            return Ok(None);
+        }
+
         config.validate().map_err(|err| {
             Report::new(TrustedServerError::Configuration {
                 message: format!(
@@ -239,10 +293,6 @@ impl IntegrationSettings {
                 ),
             })
         })?;
-
-        if !config.is_enabled() {
-            return Ok(None);
-        }
 
         Ok(Some(config))
     }
@@ -625,6 +675,34 @@ pub struct RequestSigning {
     pub enabled: bool,
     pub config_store_id: String,
     pub secret_store_id: String,
+}
+
+impl RequestSigning {
+    /// Reserved example store-id values from the config template, plus the
+    /// empty string, that must not be deployed while request signing is enabled.
+    pub const STORE_ID_PLACEHOLDERS: &[&str] = &[
+        "<management-config-store-id>",
+        "<management-secret-store-id>",
+    ];
+
+    /// Returns `true` if `store_id` is empty or a known template placeholder
+    /// (case-insensitive).
+    #[must_use]
+    pub fn is_placeholder_store_id(store_id: &str) -> bool {
+        let store_id = store_id.trim();
+        store_id.is_empty()
+            || Self::STORE_ID_PLACEHOLDERS
+                .iter()
+                .any(|p| p.eq_ignore_ascii_case(store_id))
+    }
+
+    /// Returns `true` if `store_id` cannot be deployed as-is: a placeholder, or
+    /// a value with surrounding whitespace that the key-management routes would
+    /// forward to the management API verbatim.
+    #[must_use]
+    pub fn is_unusable_store_id(store_id: &str) -> bool {
+        Self::is_placeholder_store_id(store_id) || store_id != store_id.trim()
+    }
 }
 
 fn default_request_signing_enabled() -> bool {
@@ -2618,6 +2696,7 @@ pub struct Settings {
     #[validate(nested)]
     pub rewrite: Rewrite,
     #[serde(default)]
+    #[validate(nested)]
     pub auction: AuctionConfig,
     #[serde(default)]
     pub consent: ConsentConfig,
@@ -2824,6 +2903,31 @@ impl Settings {
         for handler in &self.handlers {
             if Handler::is_placeholder_password(handler.password.expose()) {
                 insecure_fields.push(format!("handlers[{}].password", handler.path));
+            }
+        }
+        if Publisher::is_placeholder_domain(&self.publisher.domain) {
+            insecure_fields.push("publisher.domain".to_owned());
+        }
+        if Publisher::is_placeholder_cookie_domain(&self.publisher.cookie_domain) {
+            insecure_fields.push("publisher.cookie_domain".to_owned());
+        }
+        if Publisher::is_placeholder_origin_url(&self.publisher.origin_url) {
+            insecure_fields.push("publisher.origin_url".to_owned());
+        }
+        // Checked whenever the block is present, not just when it is enabled:
+        // the key rotate/deactivate admin routes are registered unconditionally
+        // and read these store IDs without consulting `enabled`, so placeholder
+        // IDs behind a disabled block would still reach key management at
+        // runtime. Surrounding whitespace is rejected too: the placeholder check
+        // trims for comparison but the raw value is what `signing_store_ids`
+        // forwards to `KeyRotationManager`, so a padded id would validate yet
+        // reach the management API unusable.
+        if let Some(request_signing) = &self.request_signing {
+            if RequestSigning::is_unusable_store_id(&request_signing.config_store_id) {
+                insecure_fields.push("request_signing.config_store_id".to_owned());
+            }
+            if RequestSigning::is_unusable_store_id(&request_signing.secret_store_id) {
+                insecure_fields.push("request_signing.secret_store_id".to_owned());
             }
         }
 
@@ -4403,6 +4507,79 @@ origin_host_header_overide = "www.example.com""#,
     }
 
     #[test]
+    fn is_placeholder_domain_rejects_known_placeholders_case_insensitively() {
+        for placeholder in Publisher::PLACEHOLDER_DOMAINS {
+            assert!(
+                Publisher::is_placeholder_domain(placeholder),
+                "should detect placeholder domain '{placeholder}'"
+            );
+        }
+        assert!(
+            Publisher::is_placeholder_domain(" Example.COM "),
+            "should detect trimmed, mixed-case placeholder domain"
+        );
+    }
+
+    #[test]
+    fn is_placeholder_domain_accepts_non_placeholder() {
+        assert!(
+            !Publisher::is_placeholder_domain("publisher.test"),
+            "should accept a real publisher domain"
+        );
+    }
+
+    #[test]
+    fn is_placeholder_cookie_domain_rejects_known_placeholders_case_insensitively() {
+        for placeholder in Publisher::PLACEHOLDER_COOKIE_DOMAINS {
+            assert!(
+                Publisher::is_placeholder_cookie_domain(placeholder),
+                "should detect placeholder cookie_domain '{placeholder}'"
+            );
+        }
+        assert!(
+            Publisher::is_placeholder_cookie_domain(" .Example.COM "),
+            "should detect trimmed, mixed-case placeholder cookie_domain"
+        );
+    }
+
+    #[test]
+    fn is_placeholder_cookie_domain_accepts_non_placeholder() {
+        assert!(
+            !Publisher::is_placeholder_cookie_domain(".publisher.test"),
+            "should accept a real cookie domain"
+        );
+    }
+
+    #[test]
+    fn is_placeholder_origin_url_rejects_equivalent_spellings_of_reserved_host() {
+        for reserved in [
+            "https://origin.example.com",
+            "https://origin.example.com/",
+            "https://origin.example.com:443",
+            "http://origin.example.com",
+            "https://Origin.Example.com",
+            " https://origin.example.com ",
+        ] {
+            assert!(
+                Publisher::is_placeholder_origin_url(reserved),
+                "should reject origin_url resolving to the reserved host: '{reserved}'"
+            );
+        }
+    }
+
+    #[test]
+    fn is_placeholder_origin_url_accepts_non_placeholder() {
+        assert!(
+            !Publisher::is_placeholder_origin_url("https://origin.publisher.test"),
+            "should accept a real origin url"
+        );
+        assert!(
+            !Publisher::is_placeholder_origin_url("https://cdn.example.com"),
+            "should accept a different host under the same example domain"
+        );
+    }
+
+    #[test]
     fn is_placeholder_handler_password_rejects_known_template_value() {
         assert!(
             Handler::is_placeholder_password("replace-with-admin-password-32-bytes"),
@@ -4425,6 +4602,26 @@ origin_host_header_overide = "www.example.com""#,
         assert!(
             format!("{err:?}").contains("handlers"),
             "error should mention handler password field"
+        );
+    }
+
+    #[test]
+    fn is_unusable_store_id_rejects_placeholders_empty_and_padded_values() {
+        for placeholder in RequestSigning::STORE_ID_PLACEHOLDERS {
+            assert!(
+                RequestSigning::is_unusable_store_id(placeholder),
+                "should reject placeholder store id '{placeholder}'"
+            );
+        }
+        for bad in ["", "   ", " 01GCFG ", "01GCFG "] {
+            assert!(
+                RequestSigning::is_unusable_store_id(bad),
+                "should reject unusable store id '{bad}'"
+            );
+        }
+        assert!(
+            !RequestSigning::is_unusable_store_id("01GCFG"),
+            "should accept a clean store id"
         );
     }
 
--- a/trusted-server.example.toml
+++ b/trusted-server.example.toml
@@ -1,132 +1,215 @@
+# =============================================================================
+# Trusted Server — application configuration template
+# =============================================================================
+#
+# This is the source-controlled starting point for an operator-owned
+# `trusted-server.toml`. Copy it (`ts config init`), fill in the required
+# values, and push it (`ts config push`) as an EdgeZero app-config blob.
+#
+# Only three sections are REQUIRED for the server to start and pass validation:
+#   1. [[handlers]] covering /_ts/admin  (admin authentication)
+#   2. [publisher]                       (domain + origin)
+#   3. [ec] passphrase                   (Edge Cookie identity secret)
+#
+# Everything below those is OPTIONAL. Most optional blocks are commented out —
+# uncomment and edit one to enable it — but a few integrations are kept as active
+# `enabled = false` stubs (see the Integrations section for why). All example
+# hosts use `example.com`; replace them with your real endpoints in your private
+# config, not in this template.
+#
+# `TRUSTED_SERVER__` env vars can override values when the `ts` CLI builds and
+# validates this config for push (e.g. TRUSTED_SERVER__PUBLISHER__DOMAIN=...;
+# nested keys use `__`). The overlay only replaces SCALAR leaves that already
+# exist in the parsed TOML: a commented-out or absent key is silently ignored,
+# and arrays or tables must be edited in TOML directly, not via env vars. The
+# deployed runtime reads the pushed config blob, so these env vars do not change
+# live behavior on their own.
+# =============================================================================
+
+
+# -----------------------------------------------------------------------------
+# REQUIRED — Admin authentication
+# -----------------------------------------------------------------------------
+# HTTP Basic-auth handler(s). At least one handler whose `path` regex covers
+# the /_ts/admin endpoints is mandatory; startup fails without it. Each handler
+# needs a non-placeholder username/password (deploy validation rejects the
+# sample password below).
 [[handlers]]
+# Regex matched against the request path. This one guards the admin surface.
 path = "^/_ts/admin"
 username = "admin"
 password = "replace-with-admin-password-32-bytes"
 
+# You can add more handlers to basic-auth-protect other path prefixes. The
+# sample password below is a known placeholder that deploy validation rejects,
+# so it forces a real secret before push:
+# [[handlers]]
+# path = "^/secure"
+# username = "user"
+# password = "replace-with-admin-password"
+
+
+# -----------------------------------------------------------------------------
+# REQUIRED — Publisher / origin
+# -----------------------------------------------------------------------------
 [publisher]
+# Public domain Trusted Server is fronting.
 domain = "example.com"
+# Cookie scope for first-party identity cookies (leading dot = include subdomains).
 cookie_domain = ".example.com"
+# Upstream origin to proxy publisher content from. No trailing slash.
 origin_url = "https://origin.example.com"
-# Optional: override outbound Host header while connecting to origin_url.
-# origin_host_header_override = "www.example.com"
+# HMAC secret for signing first-party proxy URLs. Replace before deploying.
 proxy_secret = "change-me-proxy-secret"
+# Optional: override the outbound Host header sent to origin_url.
+# origin_host_header_override = "www.example.com"
+# Optional: max bytes buffered when a response is post-processed in full (HTML
+# rewriting/injection) instead of streamed. Default 16 MiB; larger responses
+# return 502. Raise for deployments serving very large publisher pages.
+# max_buffered_body_bytes = 16777216  # 16 MiB
 
+
+# -----------------------------------------------------------------------------
+# REQUIRED — Edge Cookie (EC) identity
+# -----------------------------------------------------------------------------
 [ec]
+# Secret used to derive EC identifiers. Must be >= 32 chars and non-placeholder
+# in production (deploy validation rejects known placeholders).
 passphrase = "trusted-server-placeholder-secret"
+# KV store that persists EC identity state. This is the physical store name
+# bound per adapter (e.g. `ec_identity_store` in fastly.toml); edgezero.toml's
+# logical KV id is `trusted_server_kv`.
 ec_store = "ec_identity_store"
+# Max concurrent partner pull-sync requests.
 pull_sync_concurrency = 3
-# cluster_trust_threshold = 10
-# cluster_recheck_secs = 3600
+# Optional cluster-heuristic tuning (defaults shown):
+# cluster_trust_threshold = 10    # entries with cluster_size <= this are individual users
+# cluster_recheck_secs = 3600     # re-evaluate cluster_size after this many seconds
 
-# Example partner configuration. Replace the token before validating/pushing.
+# Optional identity partners (SSP/DSP/identity vendors). Each needs a real,
+# non-placeholder api_token (>= 32 bytes) at deploy. Configure real partners via
+# private config, not this template.
 # [[ec.partners]]
 # name = "Example Partner"
 # source_domain = "partner.example.com"
-# OpenRTB agent type; vendor-specific values are supported (PAIR uses 571187).
+# OpenRTB source.atype (default 3); vendor-specific values are supported
+# (PAIR uses 571187).
 # openrtb_atype = 3
+# include this partner's UIDs in auction user.eids
 # bidstream_enabled = true
 # api_token = "replace-with-partner-api-token-32-bytes-minimum"
-# batch_rate_limit = 60
-# pull_sync_enabled = false
+# batch_rate_limit = 60      # max batch-sync requests/min (default 60)
+# pull_sync_enabled = false  # default false
 
-# Custom headers to include in every response.
+
+# =============================================================================
+# OPTIONAL — Core features (disabled/omitted by default)
+# =============================================================================
+
+# Custom headers added to every response (e.g. X-Robots-Tag: noindex).
 # [response_headers]
 # X-Robots-Tag = "noindex"
 
-[request_signing]
-enabled = false
-config_store_id = "app_config"
-secret_store_id = "secrets"
+# Sign outbound OpenRTB/API requests. These are platform management-API store
+# IDs used for key-rotation WRITES; the runtime reads keys from fixed store names
+# (`jwks_store` / `signing_keys`), not these. To keep signing OFF, leave this
+# whole block commented out. If you uncomment it - even with `enabled = false` -
+# both store IDs are required and must be REAL values: the key rotate/deactivate
+# admin routes read them regardless of `enabled`, so deploy validation rejects
+# the placeholders below (replace them before pushing).
+# [request_signing]
+# enabled = false
+# config_store_id = "<management-config-store-id>"
+# secret_store_id = "<management-secret-store-id>"
 
-[integrations.prebid]
-enabled = false
-server_url = "https://prebid.example.com/openrtb2/auction"
-timeout_ms = 1000
-bidders = []
-debug = false
-client_side_bidders = []
-# Keep selected GAM inventory out of Trusted Server's Prebid refresh auctions.
-# Matching slots still refresh through GAM.
-# excluded_gam_ad_unit_path_suffixes = ["/trackingonly"]
-# Runtime bundle metadata. Set these after running `ts prebid bundle` and uploading the bundle.
-# external_bundle_url = "https://assets.example.com/prebid/trusted-prebid-<sha256>.js"
-# external_bundle_sha256 = "<manifest sha256>"
-# external_bundle_sri = "<manifest sri>"
+# First-party HTML/CSS rewriting controls.
+# [rewrite]
+# Domains left as-is (not proxied/rewritten). Supports "*.example.com" wildcards.
+# exclude_domains = ["*.cdn.example.com"]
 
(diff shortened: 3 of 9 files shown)


crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -201,6 +201,7 @@ pub struct ApsConfig {
 
     /// APS publisher ID (accepts both string and integer from config)
     #[serde(deserialize_with = "deserialize_pub_id")]
+    #[validate(length(min = 1), custom(function = validate_aps_pub_id))]
```
ChristianPavilonis: **P1 — APS validation runs even when the integration resolves to disabled**

`enabled` defaults to `false`, but `IntegrationSettings::get_typed` only short-circuits raw configs that explicitly contain `enabled = false`. Otherwise it calls `config.validate()` before checking `config.is_enabled()`. Consequently, an APS section that omits `enabled` but contains this documented placeholder now fails `ts config validate` even though APS resolves to disabled. This is a backward-incompatible config regression and can prevent application-state construction after an upgrade.

Please check the resolved `is_enabled()` value before field validation, while retaining the explicit-`false` fast path, and add regression coverage for APS and `adserver_mock` sections with omitted `enabled`.

ChristianPavilonis: **P2 — Whitespace-only APS publisher IDs pass the new non-empty check**

`length(min = 1)` counts whitespace, while `validate_aps_pub_id` trims only for comparison with the reserved placeholder and never rejects an empty trimmed value. I confirmed that enabled APS with `pub_id = "   "` passes `ts config validate`, leaving the invalid value to fail upstream.

Please reject `pub_id.trim().is_empty()` in the custom validator and test both empty and whitespace-only values through enabled deploy validation.

crates/trusted-server-core/src/settings.rs
```diff
@@ -2133,6 +2187,25 @@ impl Settings {
...
+            insecure_fields.push("publisher.domain".to_owned());
+        }
+        if Publisher::is_placeholder_cookie_domain(&self.publisher.cookie_domain) {
+            insecure_fields.push("publisher.cookie_domain".to_owned());
+        }
+        if Publisher::is_placeholder_origin_url(&self.publisher.origin_url) {
+            insecure_fields.push("publisher.origin_url".to_owned());
+        }
+        if let Some(request_signing) = &self.request_signing {
+            if request_signing.enabled {
```
ChristianPavilonis: **P2 — Disabled request-signing placeholders remain usable by live key-management routes**

The placeholder IDs are rejected only when `request_signing.enabled` is true. However, Fastly registers the rotate/deactivate admin routes unconditionally, and `signing_store_ids` consumes these IDs without checking `enabled`. I confirmed that a disabled request-signing block with both documented placeholders passes `ts config validate`, so authenticated key rotation still reaches invalid management-store IDs and fails later.

Please either reject placeholder/empty IDs whenever `[request_signing]` is present, or gate the key-management routes on `enabled` and document that rotation is unavailable while disabled.

trusted-server.example.toml
```diff
@@ -1,154 +1,373 @@
...
-[integrations.sourcepoint]
-enabled = false
-rewrite_sdk = true
-cdn_origin = "https://cdn.example.com"
-cache_ttl_seconds = 3600
+# Sourcepoint CMP first-party proxy.
+# [integrations.sourcepoint]
+# enabled = true
+# rewrite_sdk = true
+# cdn_origin = "https://cdn.example.com"
```
ChristianPavilonis: **P2 — The documented Sourcepoint block cannot validate as written**

`SourcepointConfig` requires `cdn_origin` to use exactly `cdn.privacy-mgmt.com`. Uncommenting this documented block therefore makes `ts config validate` fail with `cdn_origin host must be cdn.privacy-mgmt.com`; unlike deliberate invalid placeholders elsewhere, this override is neither required nor identified as invalid.

Please omit `cdn_origin` from the example so the valid pinned default is used, and document that the origin is fixed rather than operator-selectable.

ChristianPavilonis (commented):
## Summary

Reviewed the config hardening and template changes. I found one high-severity configuration compatibility regression and three medium validation/template gaps; details are inline.

crates/trusted-server-core/src/integrations/aps.rs
prk-Jr: 🔧 **wrench** — A whitespace-only `pub_id` still passes deploy validation. `length(min = 1)` counts the spaces, while `validate_aps_pub_id` trims only for the placeholder comparison. An enabled APS integration can therefore start with an unusable publisher ID.

Please reject `pub_id.trim().is_empty()` in the custom validator and add a whitespace-only regression test.

prk-Jr (changes requested):
## Summary

The configuration-template and validation hardening is well scoped, and all CI checks pass. One enabled APS configuration edge case still bypasses the new publisher-ID validation.

## Blocking

### 🔧 wrench

- **Reject whitespace-only APS publisher IDs**: the built-in length validator accepts non-empty whitespace, while the custom validator trims only for placeholder comparison (`crates/trusted-server-core/src/integrations/aps.rs:204`).

## CI Status

- fmt: PASS
- clippy / cargo checks: PASS
- rust tests: PASS
- js tests: PASS
- integration tests: PASS

crates/trusted-server-core/src/integrations/aps.rs
aram356: Confirmed and fixed in 717363f13.

`get_typed` now checks the resolved `is_enabled()` before running field validation, keeping the explicit-`false` fast path (which still avoids parsing entirely). Since `enabled` defaults to `false` for APS and `adserver_mock`, a section that omitted the flag was previously parsed and validated even though it resolved to disabled.

Regression coverage: `deploy_validation_skips_field_validation_for_integrations_with_omitted_enabled` asserts an omitted-`enabled` APS section carrying the documented placeholder plus an `adserver_mock` section with a validation-failing `endpoint` both pass deploy validation. I verified the test fails against the pre-fix ordering with exactly the reported error (`Integration 'aps' configuration failed validation: pub_id: ... aps_pub_id_placeholder`), so it is not vacuous.

aram356: Confirmed and fixed in 717363f13.

`validate_aps_pub_id` now trims first and rejects an empty result with an `aps_pub_id_blank` error before the placeholder comparison, so whitespace-only IDs no longer slip past the built-in `length(min = 1)` (which counts whitespace).

`deploy_validation_rejects_blank_aps_pub_id` covers both `""` and `"   "` through enabled deploy validation.

crates/trusted-server-core/src/settings.rs
aram356: Confirmed and fixed in 717363f13, taking the first of your two options.

The placeholder/empty check now runs whenever `[request_signing]` is present rather than only when `enabled` is true. I kept the key-management routes as they are: gating them on `enabled` is a runtime/routing behavior change across adapters, which is a wider blast radius than this config-validation PR should carry.

`is_placeholder_store_id` already trims and treats empty as a placeholder, so empty and whitespace-only IDs are covered by the same path. Two tests: `deploy_validation_rejects_placeholder_store_ids_while_request_signing_is_disabled` (the previous `..._allows_disabled_...` test asserted the old behavior and is inverted) and `deploy_validation_rejects_empty_request_signing_store_ids`.

Trade-off worth naming: an operator who declares the block but disables it must now supply real store IDs. That seems right given the routes read them regardless.

trusted-server.example.toml
aram356: Confirmed and fixed in 717363f13.

I dropped the `cdn_origin` line from the example so the pinned default applies, and documented that the origin is fixed rather than operator-selectable. I did not substitute the pinned host into the template: project convention keeps real vendor domains out of docs/examples, so naming it there would trade one problem for another.

Verified empirically both ways — the block as documented now passes `validate_settings_for_deploy`, and the old line reproduced exactly the error you predicted: `cdn_origin: cdn_origin host must be cdn.privacy-mgmt.com`.

crates/trusted-server-core/src/integrations/aps.rs
aram356: Confirmed and fixed in 717363f13. `validate_aps_pub_id` now trims and rejects a blank result before the placeholder check, so `length(min = 1)` counting whitespace no longer lets `"   "` through. Regression test `deploy_validation_rejects_blank_aps_pub_id` covers both empty and whitespace-only through enabled deploy validation.

trusted-server.example.toml
```diff
@@ -1,163 +1,384 @@
...
+
+# Server-side ad slot templates + creative-opportunity auction.
+# [creative_opportunities]
+# gam_network_id = "123456789"
+# price_granularity = "dense"
+# FCP is not affected by this value — body content above </body> has already
+# streamed and painted before the hold begins. What this caps is the slip on
+# DOMContentLoaded and window.load. 500 ms is the recommended default; raise
+# only if your SSPs need more headroom and analytics confirm the DCL slip is OK.
+# auction_timeout_ms = 500  # override via TRUSTED_SERVER__CREATIVE_OPPORTUNITIES__AUCTION_TIMEOUT_MS
```
ChristianPavilonis: **Documented environment overrides are silently ignored for the now-commented creative-opportunities config**

This PR comments out `[creative_opportunities]` and `auction_timeout_ms`, but lines 216–219 still advertise `TRUSTED_SERVER__CREATIVE_OPPORTUNITIES__AUCTION_TIMEOUT_MS` and `...__SLOT` overrides. The typed EdgeZero loader only overlays existing scalar leaves; it does not create absent tables/fields or arrays. The timeout override worked with the previous active template but is now silently ignored.

Operators can therefore validate and push a config believing auction timeout or slots were applied when the creative-opportunities configuration remains absent. I confirmed that an invalid override for existing `publisher.domain` is observed and rejected, while an invalid `CREATIVE_OPPORTUNITIES__AUCTION_TIMEOUT_MS` override validates successfully when the table or leaf is absent; the documented `...__SLOT` override is likewise ignored.

Please either preserve the seed table/scalar leaves needed by the overlay, add typed-loader support for absent tables and arrays, or remove/qualify these instructions and direct operators to edit private TOML. A CLI test should use these exact variables and assert the resulting typed values.

crates/trusted-server-core/src/settings.rs
```diff
@@ -114,6 +114,39 @@ impl Publisher {
...
+    pub fn is_placeholder_cookie_domain(cookie_domain: &str) -> bool {
+        Self::PLACEHOLDER_COOKIE_DOMAINS
+            .iter()
+            .any(|p| p.eq_ignore_ascii_case(cookie_domain.trim()))
+    }
+
+    /// Returns `true` if `origin_url` is the unedited template placeholder
+    /// (case-insensitive).
+    #[must_use]
+    pub fn is_placeholder_origin_url(origin_url: &str) -> bool {
```
ChristianPavilonis, line 160: **A semantically unchanged placeholder origin bypasses the new deploy check**

`is_placeholder_origin_url` compares the complete raw string, so `https://origin.example.com:443` passes even though it resolves to the same reserved template host as `https://origin.example.com`. I verified that `ts config validate` accepts this variant in a template-derived config. This defeats the new fail-fast protection and can publish a configuration that proxies production traffic to a domain the operator cannot own.

Please parse the URL and reject based on normalized `host_str()` matching `origin.example.com`, independent of explicit default ports or equivalent spellings, and add default-port regression coverage.

crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -285,6 +287,35 @@ impl Default for ApsConfig {
     }
 }
 
+/// Validator for [`ApsConfig::pub_id`]: rejects blank publisher IDs and the
+/// known template placeholder. The built-in `length` validator counts
+/// whitespace, so a whitespace-only `pub_id` has to be rejected here. This runs
+/// only when APS is enabled, because integration configs validate lazily via
+/// `get_typed`.
+fn validate_aps_pub_id(pub_id: &str) -> Result<(), ValidationError> {
```
ChristianPavilonis: **Identifier validation trims for comparison but forwards the untrimmed values**

APS validation checks `pub_id.trim()`, but the request uses the original value at `aps.rs:426`. Request-signing validation has the same pattern at `settings.rs:701`, while `request_signing/endpoints.rs:205-206` forwards the original store IDs. A config containing `pub_id = " 5128 "` or store IDs such as `" 01GCFG "` therefore passes deploy validation while retaining the whitespace sent to the APS or management API boundary. This is distinct from the previously fixed whitespace-only APS case.

Please reject surrounding whitespace (`value != value.trim()`) or normalize these identifiers during deserialization/finalization. Tests should ensure that accepted values are exactly what reaches APS payloads and `KeyRotationManager`.

trusted-server.example.toml
```diff
@@ -1,163 +1,384 @@
...
+# [[creative_opportunities.slot]]
+# id = "leaderboard"
+# gam_unit_path = "/123456789/leaderboard"
+# div_id = "div-gpt-ad-leaderboard"
+# page_patterns = ["/"]              # glob syntax (not regex); "/" matches only the homepage
+# formats = [{ width = 728, height = 90 }]
+
+# Direct Tinybird auction telemetry (all off / empty by default).
+# [tinybird]
+# enabled = false
```
ChristianPavilonis: **The restored Tinybird example omits the fields required to enable it**

The documented block exposes only `enabled`, but enabled Tinybird requires a non-empty `api_host`; its store, dataset, and token-key settings are also operationally relevant. Uncommenting this section and setting `enabled = true` therefore immediately fails validation, so the restored template does not provide enough guidance to configure the advertised feature.

Please show `api_host` as required, including its host-only format, and document the store/dataset/token fields and defaults. A template-derived validation test for enabled Tinybird would prevent this from regressing.

ChristianPavilonis (approved):
## Summary

The validator refactors and previously requested fixes are generally sound. I found one high-impact silent configuration regression and three medium-severity validation/documentation gaps; details and concrete fixes are included inline.

trusted-server.example.toml
prk-Jr: 🔧 **wrench** — This override is now silently ignored, and the array form below can never work.

Upstream `edgezero-core/src/app_config.rs` (`apply_env_overlay`) states it plainly: *"The overlay only overrides keys that already exist in the parsed tree (the existing TOML value's type drives coercion of the env string)."* And `coerce_env_value` rejects `Value::Array | Value::Table` with `env overlay supports scalar leaves only`.

So:

- `auction_timeout_ms = 500` was an **active** leaf on `main`, which is why `TRUSTED_SERVER__CREATIVE_OPPORTUNITIES__AUCTION_TIMEOUT_MS` worked. Commented out, the env var is dropped with no diagnostic.
- The `TRUSTED_SERVER__CREATIVE_OPPORTUNITIES__SLOT='[{...}]'` form on line 219 cannot work in either state: absent leaf → ignored, present array leaf → `EnvOverlay` error.
- The same silent drop now applies to every leaf this PR commented out that an operator or the dev loop might override: `auction.timeout_ms`, `integrations.prebid.server_url` / `timeout_ms`, `integrations.aps.pub_id`, `request_signing.*`, `debug.*`. `[proxy]` disappearing also takes `.env.dev`'s documented `TRUSTED_SERVER__PROXY__CERTIFICATE_CHECK` with it.

The header at lines 20-23 promises the mechanism without the caveat, which is what makes this land as a footgun rather than a preference: an operator can validate and push believing an override applied.

**Fix** — three small edits:

1. Add the caveat to the header block:

```toml
# `TRUSTED_SERVER__` env vars can override values when the `ts` CLI builds and
# validates this config for push (e.g. TRUSTED_SERVER__PUBLISHER__DOMAIN=...;
# nested keys use `__`). The overlay only replaces SCALAR leaves that already
# exist in the TOML — a commented-out key is silently ignored, and arrays or
# tables must be edited in TOML directly. The deployed runtime reads the pushed
# config blob, so these env vars do not change live behavior on their own.
```

2. Drop the `...__SLOT` example on lines 218-219 (or requalify it as "edit the `[[creative_opportunities.slot]]` blocks below").
3. Keep an active seed leaf for anything the local dev loop overrides today.

This overlaps @ChristianPavilonis's open comment on the same lines; the upstream quote and the wider set of affected leaves are the additions.

trusted-server.example.toml
```diff
@@ -1,163 +1,384 @@
...
-enabled = false
-app_id = ""
-api_endpoint = "https://identity.example.com"
-sdk_url = "https://identity.example.com/trusted-server.js"
-cache_ttl_seconds = 3600
-rewrite_sdk = true
+# Permutive DMP. `organization_id` and `workspace_id` required when enabled.
+# [integrations.permutive]
+# enabled = true
+# organization_id = ""
```
prk-Jr: 🔧 **wrench** — Uncommenting this block as written fails validation: `PermutiveConfig` marks both `organization_id` and `workspace_id` `#[validate(length(min = 1))]`, so `enabled = true` with `""` is rejected. Three more documented blocks share the defect:

- `[integrations.lockr]` (line 326): `app_id = ""`, also `length(min = 1)`.
- `[tinybird]` (lines 228-229): only `enabled` is shown, but `prepare_runtime` → `validate_tinybird_api_host` requires a non-empty host-only `api_host` when enabled, and `secret_store` / `auction_dataset` / `auction_token_secret` are operationally relevant.
- `[request_signing]` (lines 113-116): documented with `enabled = false` **plus** both `<management-...>` placeholders — which this PR's own always-present check now rejects. Uncommenting the block in order to explicitly disable signing fails `ts config validate`.

This is the same class as the sourcepoint `cdn_origin` finding already fixed in this PR: a block that reads as copy-and-edit but cannot be pushed as shown. In a PR whose stated purpose is making the template the operator-facing documentation, that matters more than it normally would.

**Fix** — use self-describing fill-me-in values (the APS block already does this well with `your-aps-publisher-id`) and mark required-when-enabled fields:

```toml
# Permutive DMP. `organization_id` and `workspace_id` required when enabled.
# [integrations.permutive]
# enabled = true
# organization_id = "your-permutive-organization-id"   # required
# workspace_id = "your-permutive-workspace-id"         # required
```

For `[request_signing]`, either comment out the two id lines or state that a present block requires real store IDs even when disabled.

Worth a regression test that uncomments each documented block individually and asserts parse + field validation, excluding the deliberately invalid placeholders (`ec.passphrase`, `handlers[].password`, GTM `container_id`, APS `pub_id`). That test is what would keep this from drifting again.

(The tinybird half of this overlaps @ChristianPavilonis's open comment.)

crates/trusted-server-core/src/integrations/aps.rs
prk-Jr: ♻️ **refactor** — `length(min = 1)` is now redundant: `validate_aps_pub_id` trims first and rejects the empty result, which subsumes the empty-string case the length rule was added for. Both validators run, so a blank `pub_id` reports two errors for one cause.

```rust
#[validate(custom(function = validate_aps_pub_id))]
pub pub_id: String,
```

The doc comment on `validate_aps_pub_id` already explains why the built-in cannot carry this check on its own, so dropping the attribute keeps the single source of truth.

crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -285,6 +287,35 @@ impl Default for ApsConfig {
...
+    if ApsConfig::PUB_ID_PLACEHOLDERS
+        .iter()
+        .any(|placeholder| placeholder.eq_ignore_ascii_case(pub_id))
+    {
+        return Err(ValidationError::new("aps_pub_id_placeholder"));
+    }
+    Ok(())
+}
+
+impl ApsConfig {
```
prk-Jr: ⛏ **nitpick** — This adds a second `impl ApsConfig` block for one associated const while `impl IntegrationConfig for ApsConfig` follows immediately below. Moving `PUB_ID_PLACEHOLDERS` next to the existing `ApsConfig` impl (or directly above `validate_aps_pub_id`, its only caller) keeps the type's inherent items in one place — matching how `RequestSigning::STORE_ID_PLACEHOLDERS` sits in its type's single impl block.

crates/trusted-server-core/src/settings.rs
```diff
@@ -114,6 +114,39 @@ impl Publisher {
             .any(|p| p.eq_ignore_ascii_case(proxy_secret))
     }
 
+    /// Reserved example publisher values copied verbatim from the config
+    /// template. They deserialize fine but must be replaced before deploying.
+    const PLACEHOLDER_DOMAINS: &[&str] = &["example.com"];
+    const PLACEHOLDER_COOKIE_DOMAINS: &[&str] = &[".example.com"];
```
prk-Jr, line 129: 🏕 **camp site** — The four pre-existing helpers in this file each have direct unit tests: `is_placeholder_passphrase_rejects_all_known_placeholders`, `..._is_case_insensitive`, `..._accepts_non_placeholder`, and the same trio for `api_token` / `proxy_secret` / `password`. The three new ones are only exercised indirectly through `validate_settings_for_deploy`.

Adding the mirror tests is cheap and would have surfaced the normalization gap @ChristianPavilonis flagged separately — a `..._accepts_non_placeholder` case naturally invites `https://origin.example.com:443` and `https://origin.example.com/`, both of which currently pass because the comparison is on the whole raw string.

crates/trusted-server-core/src/settings.rs
```diff
@@ -2139,6 +2198,28 @@ impl Settings {
...
+            insecure_fields.push("publisher.domain".to_owned());
+        }
+        if Publisher::is_placeholder_cookie_domain(&self.publisher.cookie_domain) {
+            insecure_fields.push("publisher.cookie_domain".to_owned());
+        }
+        if Publisher::is_placeholder_origin_url(&self.publisher.origin_url) {
+            insecure_fields.push("publisher.origin_url".to_owned());
+        }
+        // Checked whenever the block is present, not just when it is enabled:
+        // the key rotate/deactivate admin routes are registered unconditionally
```
prk-Jr, line 2917: 🤔 **thinking** — Agreed this is the right half to fix here, and the comment explaining why is exactly the sort of rationale that stops a future reader from "simplifying" it back. One consequence worth writing down outside the code: this rejects placeholder or empty store IDs in configs that are already pushed and working, where `[request_signing]` is present with `enabled = false`. Those operators get a `ts config push` failure on their next unrelated change, with an error naming a subsystem they deliberately turned off.

A CHANGELOG or release-note line ("a present `[request_signing]` block now requires real store IDs even when disabled, because the key-management routes read them regardless") turns that from a surprise into an expected upgrade step.

prk-Jr (changes requested):
## Summary

Re-review at `0c2b38cd1`. My earlier blocking finding is fixed: `validate_aps_pub_id` now trims before the placeholder comparison and `deploy_validation_rejects_blank_aps_pub_id` covers both empty and whitespace-only through enabled deploy validation. The `get_typed` reordering, the request-signing always-on check, and the sourcepoint `cdn_origin` removal are all in place. CI is green across all 19 checks.

Two findings remain that concern the template as documentation: a documented env-override mechanism that now silently does nothing, and four documented blocks that fail validation if uncommented as written.

## Blocking

### 🔧 wrench

- **Documented env-var overrides are silently ignored for the leaves this PR commented out** (`trusted-server.example.toml:216`, 218-219, 20-23) — see inline comment.
- **Four documented blocks cannot validate as written** (`trusted-server.example.toml:113-116`, 228-229, 317-318, 326) — see inline comment on the permutive block.

## Non-blocking

### 📌 out of scope

- **Gate key-management routes on `request_signing.enabled`**: rejecting placeholder store IDs while the block is disabled is the right call for this PR, but the underlying asymmetry is that rotate/deactivate are registered unconditionally and read the IDs without consulting `enabled`. Worth a follow-up issue so the config-side workaround does not become the permanent answer.

### 👍 praise

- **Removing the `example.com` origin overrides is a fail-open fix, not just documentation**: `main` shipped active `sdk_origin = "https://sdk.example.com"` / `api_origin` for didomi and datadome and `script_url = "https://ads.example.com/gpt.js"` for gpt. Since all three fields have real vendor defaults, `ts audit` flipping `enabled = true` on those sections previously produced a config that proxied to `example.com`. With the overrides gone the defaults apply. The PR description does not claim this, but it is the most valuable behavioral consequence of the template rework.
- **`GTM-REPLACE-ME` is fail-closed by construction**: it cannot match `^GTM-[A-Z0-9]{4,20}$` (unlike the previous `GTM-EXAMPLE`, which validated fine), and `audit/mod.rs:313-330` only flips `enabled` when `extract_gtm_container_id` actually returned an id — otherwise the integration goes to manual review. Nice pairing.
- **The `get_typed` reordering is covered by a test that would actually fail without it**: `deploy_validation_skips_field_validation_for_integrations_with_omitted_enabled` uses an `adserver_mock` endpoint that parses but fails the `url` validator, so the test cannot pass vacuously.

## CI Status

- fmt: PASS
- clippy / cargo check (cloudflare native + wasm, spin native + wasm): PASS
- rust tests: PASS (fastly, axum, cloudflare, spin, ts CLI, cross-adapter parity)
- js tests (vitest) / format-typescript / format-docs: PASS
- integration + browser + Fastly EC lifecycle: PASS
- CodeQL: PASS

crates/trusted-server-core/src/settings.rs
aram356: Fixed in e4e0e8f3e. `is_placeholder_origin_url` now parses the URL and compares `host_str()` against the reserved host (`origin.example.com`), so `:443`, a trailing slash, `http://`, and mixed case are all rejected. Regression coverage in `is_placeholder_origin_url_rejects_equivalent_spellings_of_reserved_host` (covers all those variants) plus an `..._accepts_non_placeholder` case.

crates/trusted-server-core/src/integrations/aps.rs
aram356: Fixed in e4e0e8f3e. APS `validate_aps_pub_id` now rejects a value that differs from its trimmed form (`aps_pub_id_whitespace`), and request-signing store IDs go through a new `RequestSigning::is_unusable_store_id` (placeholder/empty OR surrounding whitespace) in `reject_placeholder_secrets`. Since deploy validation now rejects padded values, an accepted id has no surrounding whitespace and is forwarded verbatim, so accepted == what reaches the APS payload / `KeyRotationManager`. Tests: `validate_aps_pub_id_accepts_valid_and_rejects_bad_values`, `is_unusable_store_id_rejects_placeholders_empty_and_padded_values`, and deploy-level `deploy_validation_rejects_blank_or_padded_aps_pub_id` / `deploy_validation_rejects_padded_request_signing_store_ids`. I went with reject rather than silent normalization to stay consistent with the PR's fail-fast theme.

trusted-server.example.toml
aram356: Fixed in e4e0e8f3e. The `[tinybird]` block now shows `api_host` as required (host-only format) plus the `secret_store` / `auction_dataset` / `auction_token_secret` fields with their defaults. `documented_tinybird_block_validates_when_uncommented` uncomments the block from the actual template and asserts it parses and passes the `validate_tinybird_api_host` check.

trusted-server.example.toml
aram356: Fixed in e4e0e8f3e, together with prk-Jr's overlapping comment. The header now states the overlay only replaces scalar leaves that already exist in the parsed TOML (commented/absent keys silently ignored; arrays and tables must be edited in TOML). The `...__AUCTION_TIMEOUT_MS` inline note and the `...__SLOT` array example are dropped/requalified to point at editing the `[[creative_opportunities.slot]]` blocks directly. Re your test suggestion: I added `documented_integration_blocks_validate_when_uncommented` for the copy-and-edit blocks; I did not add a CLI test asserting the timeout env var applies, because the correct behavior here is that it is NOT applied for a commented-out leaf - the fix is the doc caveat, not restoring an active leaf.

aram356: Fixed in e4e0e8f3e. Added the scalar-leaves-only caveat to the header block (commented/absent keys silently ignored; arrays and tables must be edited in TOML directly) and requalified the `...__SLOT` example to point at editing the `[[creative_opportunities.slot]]` blocks. On the seed-leaf point: `.env.dev`'s only active overrides are `PUBLISHER__ORIGIN_URL` (present, works) and the synthetic stores; `PROXY__CERTIFICATE_CHECK` is itself commented in `.env.dev` and `proxy.certificate_check` was already commented on main, so that overlay was inert before this PR too - not a regression I introduced, and the header caveat now documents why.

trusted-server.example.toml
aram356: Fixed in e4e0e8f3e. Permutive and lockr now use self-describing required values (`your-permutive-organization-id`, `your-permutive-workspace-id`, `your-lockr-app-id`) marked `# required (non-empty)`; tinybird shows `api_host` + the store/dataset/token fields. For `[request_signing]`: the store-id fields are required (no serde default), so commenting out just those two lines would break parsing - instead the block header now documents that a present block needs REAL store IDs even with `enabled = false`, and to keep signing off you leave the whole block commented. Added the regression test you suggested: `documented_integration_blocks_validate_when_uncommented` uncomments each block from the real template and asserts parse + field validation; I verified it fails if a required value is emptied back out.

crates/trusted-server-core/src/integrations/aps.rs
aram356: Done in e4e0e8f3e - dropped `length(min = 1)`, leaving `#[validate(custom(function = validate_aps_pub_id))]` as the single source of truth (it rejects blank, surrounding whitespace, and the placeholder).

crates/trusted-server-core/src/integrations/aps.rs
aram356: Done in e4e0e8f3e. Moved `PUB_ID_PLACEHOLDERS` to a module const directly above `validate_aps_pub_id`, its only caller, removing the one-item `impl ApsConfig` block.

crates/trusted-server-core/src/settings.rs
aram356: Done in e4e0e8f3e. Added mirror unit tests for all three helpers: `is_placeholder_domain_*`, `is_placeholder_cookie_domain_*`, and `is_placeholder_origin_url_*`. As you predicted, the origin test naturally invited `https://origin.example.com:443` and the trailing-slash form - both of which prompted the normalization fix in `is_placeholder_origin_url` and are now asserted rejected.

crates/trusted-server-core/src/settings.rs
aram356: Agreed. There is no CHANGELOG file in the repo, so I captured the upgrade note two ways: (1) an "Upgrade note" section added to the PR description, and (2) inline in the template - the `[request_signing]` block header now spells out that a present block needs real store IDs even when disabled, because the key-management routes read them regardless of `enabled`. Happy to move it into a CHANGELOG if the project adds one.

prk-Jr (approved):

PR #872, opened 2026-07-08
Implement parser-safe SSAT late binding

### LINKED ISSUE #857: Make SSAT publisher HTML assembly truly streaming and parser-safe
## Summary

Make the SSAT publisher HTML assembly path truly streaming and parser-safe.

The cache-control spec now defers origin-template caching and compression offload, but the architecture review found that the current publisher path still materializes origin bodies before client bytes can flow. This issue tracks the streaming/correctness work separately from the cache-header PR.

## Problem

Current behavior has several streaming/correctness gaps:

- Fastly publisher origin fetches can materialize the full origin response body in WASM heap before the client sees bytes.
- EdgeZero/non-Fastly paths buffer the assembled publisher response.
- The SSAT bid hold currently relies on a raw-byte `</body` search, which can false-positive inside inline scripts/JSON.
- HTML post-processors can force full-document buffering, which undermines the intended SSAT streaming model.

These issues affect TTFB/FCP and make the SSAT assembly path more fragile, but they are broader than the initial cache-header policy work.

## Proposed scope

- Add/finish true publisher origin streaming with a cumulative response cap.
- Ensure SSAT HTML can stream head/body while holding only the parser-confirmed body-close tail needed for bid injection.
- Replace raw-byte `</body` detection with parser-context-aware close-body handling.
- Review integrations/post-processors that force full-document buffering and either make them streaming-safe or explicitly route/document them as buffered mode.
- Port or document equivalent behavior for EdgeZero/non-Fastly adapters.

## Acceptance criteria

- [ ] Fastly SSAT HTML path no longer requires full origin body materialization before sending client bytes.
- [ ] Streaming path enforces a cumulative body cap without requiring a single full-body allocation.
- [ ] Body-close hold is parser-context-aware and does not trigger on `</body` literals inside inline scripts/JSON.
- [ ] EdgeZero/non-Fastly adapter behavior is either streaming-safe or explicitly documented/tested as buffered mode.
- [ ] Tests cover large HTML, inline-script `</body` literals, missing body close tags, and normal bid injection.
- [ ] Existing `private, max-age=0` SSAT privacy behavior remains unchanged.

## Out of scope

- Origin-template caching.
- Transformed-template caching.
- Compression offload via `Accept-Encoding: identity` / `X-Compress-Hint`.
- Auction backend timeout/name fixes.

## Reference

See `docs/superpowers/specs/2026-07-06-cache-control-header-design.md` for the cache-control scope split and deferred follow-up list.

ChristianPavilonis:
## Summary

- Implements parser-safe SSAT bid late binding using a `lol_html`-inserted opaque placeholder instead of raw `</body` byte scanning.
- Adds EOF fallback handling for malformed HTML, including a minimal executable tail when `<head>` and `</body>` are missing.
- Keeps full-document HTML post-processors on a bounded buffered path that sees final HTML, not placeholders.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/publisher_late_binding.rs` | Adds per-request bid placeholders, insertion tracking, and a streaming placeholder late binder with split-chunk detection, duplicate stripping, and held-tail cap enforcement. |
| `crates/trusted-server-core/src/html_processor.rs` | Adds bid injection modes, post-processing mode controls, reusable head bootstrap/post-processor helpers, and placeholder insertion from parser-confirmed body end tags. |
| `crates/trusted-server-core/src/publisher.rs` | Replaces raw body-close scanning with parser-safe late binding, EOF fallback, buffered post-processor late binding, cap enforcement, and auction collect/abandon handling. |
| `crates/trusted-server-core/src/lib.rs` | Exports the late-binding module. |
| `crates/trusted-server-core/benches/html_processor_bench.rs` | Updates benchmark config for the new HTML processor fields. |
| `docs/superpowers/specs/2026-07-08-ssat-publisher-streaming-parser-safe-design.md` | Adds the parser-safe SSAT streaming design spec. |
| `docs/superpowers/plans/2026-07-08-ssat-publisher-streaming-parser-safe-implementation-plan.md` | Adds the implementation plan for the design. |

## Closes

Closes #857

## Test plan

- [x] Other: `cargo test-cloudflare && cargo clippy-cloudflare`

## Checklist

- [x] New code has tests

### DIFF
--- a/crates/trusted-server-core/src/publisher.rs
+++ b/crates/trusted-server-core/src/publisher.rs
@@ -40,6 +40,9 @@ use crate::auction::telemetry::{
 use crate::auction::types::{
     AuctionContext, AuctionRequest, Bid, DeviceInfo, PublisherInfo, SiteInfo, UserInfo,
 };
+use crate::cache_policy::{
+    cache_control_headers_are_private_or_no_store, CachePolicy, EdgeCacheHeader,
+};
 use crate::consent::{consent_allows_server_side_auction, gate_eids_by_consent};
 use crate::constants::{COOKIE_TS_EIDS, HEADER_X_COMPRESS_HINT};
 use crate::cookies::handle_request_cookies;
@@ -165,6 +168,7 @@ fn accept_encoding_qvalue(header_value: &str, target: &str) -> Option<f32> {
 pub fn handle_tsjs_dynamic(
     req: &Request<EdgeBody>,
     integration_registry: &IntegrationRegistry,
+    edge_header: EdgeCacheHeader,
 ) -> Result<Response<EdgeBody>, Report<TrustedServerError>> {
     const PREFIX: &str = "/static/tsjs=";
     const UNIFIED_FILENAMES: &[&str] = &["tsjs-unified.js", "tsjs-unified.min.js"];
@@ -179,10 +183,8 @@ pub fn handle_tsjs_dynamic(
         // Serve core + immediate modules (excludes deferred like prebid)
         let module_ids = integration_registry.js_module_ids_immediate();
         let body = trusted_server_js::concatenate_modules(&module_ids);
-        let mut resp = serve_static_with_etag(&body, req, "application/javascript; charset=utf-8");
-        resp.headers_mut()
-            .insert(HEADER_X_COMPRESS_HINT, HeaderValue::from_static("on"));
-        return Ok(resp);
+        let hash = trusted_server_js::concatenated_hash(&module_ids);
+        return Ok(serve_tsjs_static(req, &body, &hash, edge_header));
     }
 
     if let Some(module_id) = parse_deferred_module_filename(filename) {
@@ -191,19 +193,46 @@ pub fn handle_tsjs_dynamic(
         if !deferred_ids.contains(&module_id) {
             return Ok(not_found_response());
         }
-        if let Some(content) = trusted_server_js::module_bundle(module_id) {
-            let mut resp =
-                serve_static_with_etag(content, req, "application/javascript; charset=utf-8");
-            resp.headers_mut()
-                .insert(HEADER_X_COMPRESS_HINT, HeaderValue::from_static("on"));
-            return Ok(resp);
+        if let (Some(content), Some(hash)) = (
+            trusted_server_js::module_bundle(module_id),
+            trusted_server_js::single_module_hash(module_id),
+        ) {
+            return Ok(serve_tsjs_static(req, content, hash, edge_header));
         }
     }
 
     Ok(not_found_response())
 }
 
-/// Extract a module ID from a deferred-module filename like `tsjs-sourcepoint.min.js`.
+fn serve_tsjs_static(
+    req: &Request<EdgeBody>,
+    body: &str,
+    expected_hash: &str,
+    edge_header: EdgeCacheHeader,
+) -> Response<EdgeBody> {
+    let mut resp = serve_static_with_etag(
+        body,
+        req,
+        "application/javascript; charset=utf-8",
+        edge_header,
+    );
+    if request_version_hash(req).is_some_and(|hash| hash == expected_hash) {
+        CachePolicy::public_immutable(Duration::from_secs(31_536_000))
+            .apply_to_headers(resp.headers_mut(), edge_header);
+    }
+    resp.headers_mut()
+        .insert(HEADER_X_COMPRESS_HINT, HeaderValue::from_static("on"));
+    resp
+}
+
+fn request_version_hash(req: &Request<EdgeBody>) -> Option<&str> {
+    req.uri().query()?.split('&').find_map(|pair| {
+        let (name, value) = pair.split_once('=')?;
+        (name == "v").then_some(value)
+    })
+}
+
+/// Extract a module ID from a deferred-module filename like `tsjs-prebid.min.js`.
 ///
 /// Returns `Some(&'static str)` if the filename matches a known JS module ID,
 /// `None` otherwise. The caller must additionally verify that the module is
@@ -425,6 +454,33 @@ pub(crate) fn classify_response_route(
     }
 
     ResponseRoute::Stream
+}
+
+fn response_cache_control_is_private_or_no_store(response: &Response<EdgeBody>) -> bool {
+    cache_control_headers_are_private_or_no_store(response.headers())
+}
+
+fn apply_publisher_asset_cache_policy(
+    settings: &Settings,
+    path: &str,
+    cache_rule_method: bool,
+    edge_header: EdgeCacheHeader,
+    response: &mut Response<EdgeBody>,
+) -> Result<(), Report<TrustedServerError>> {
+    if !cache_rule_method || response_cache_control_is_private_or_no_store(response) {
+        return Ok(());
+    }
+
+    let status = response.status();
+    if !(status.is_success() || status == StatusCode::NOT_MODIFIED) {
+        return Ok(());
+    }
+
+    if let Some(policy) = settings.asset_cache_policy_for_path(path)? {
+        policy.apply_to_headers(response.headers_mut(), edge_header);
+    }
+
+    Ok(())
 }
 
 /// Owned version of [`ProcessResponseParams`] for returning from
@@ -1306,6 +1362,7 @@ pub async fn handle_publisher_request(
     ec_context: &mut EcContext,
     auction: AuctionDispatch<'_>,
     mut req: Request<EdgeBody>,
+    edge_header: EdgeCacheHeader,
 ) -> Result<PublisherResponse, Report<TrustedServerError>> {
     log::debug!("Proxying request to publisher_origin");
 
@@ -1389,6 +1446,7 @@ pub async fn handle_publisher_request(
 
     let request_path = req.uri().path().to_string();
     let is_get = req.method() == http::Method::GET;
+    let cache_rule_method = req.method() == Method::GET || req.method() == Method::HEAD;
 
     let is_prefetch = is_prefetch_request(&req);
     let is_bot = is_bot_user_agent(&req);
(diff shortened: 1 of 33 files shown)


aram356:
@ChristianPavilonis Please resolve conflicts

aram356:
@ChristianPavilonis to investigate whether it is still relevant with ESI

crates/trusted-server-core/src/settings.rs
```diff
@@ -1889,6 +1892,322 @@ fn validate_tinybird_secret(value: &str, setting: &str) -> Result<(), Report<Tru
...
+            .iter()
+            .map(|value| value.trim().trim_start_matches('.').to_ascii_lowercase())
+            .filter(|value| !value.is_empty())
+            .collect();
+    }
+
+    fn prepare_runtime(&self) -> Result<(), Report<TrustedServerError>> {
+        self.validate_matcher_shape()?;
+        self.compiled_regex().map(|_| ())?;
+        self.compiled_globs().map(|_| ())?;
```
prk-Jr, line 2043: **🔧 [P2] Skip runtime matcher preparation for disabled cache rules**

`CacheAssetRule::matches_path` ignores rules with `enabled = false`, but startup unconditionally calls `validate_matcher_shape`, `compiled_regex` and `compiled_globs`. Keeping a disabled draft rule with `path_regex = "["` therefore prevents the entire service from starting; disabling a broken rule cannot restore service. This also conflicts with the documented “Disabled rules are ignored” behavior and the enabled/disabled config convention.

Verified scratch regression: deserialize a cache rule with `id = "disabled"`, `enabled = false`, `path_regex = "["`, assign it to test settings, and call `Settings::prepare_runtime()`. The expectation that disabled configuration is bypassed fails with `Configuration error: cache.asset_rules disabled path_regex [ failed to compile: ... unclosed character class`.

Proposed implementation:

```rust
    fn prepare_runtime(&self) -> Result<(), Report<TrustedServerError>> {
        if !self.enabled {
            return Ok(());
        }
        self.validate_matcher_shape()?;
        self.compiled_regex().map(|_| ())?;
        self.compiled_globs().map(|_| ())?;
        Ok(())
    }
```

Add paired enabled/disabled tests for invalid regex, invalid glob and incomplete matcher shape. Keep ID uniqueness checks if deliberately global. Prose-only: fix and regression tests span distinct hunks; no full release-gate verification of replacement bytes was performed. The displayed code is a proposal, not a verified GitHub suggestion.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -962,202 +1409,170 @@ async fn stream_html_with_auction_hold<W: Write, P: StreamProcessor>(
...
-                    },
-                )?;
-                if !final_out.is_empty() {
-                    writer
-                        .write_all(&final_out)
-                        .change_context(TrustedServerError::Proxy {
-                            message: "Failed to write finalized output".to_string(),
-                        })?;
+                        tail.as_bytes(),
+                        &mut state,
```
prk-Jr, line 1523: **🔧 [P2] Do not append executable fallback markup inside an open textarea**

EOF is not necessarily an HTML data-state insertion point. For an ad-eligible response ending with `<html><head></head><body><textarea>draft`, the new fallback appends the bids `<script>` directly into the still-open textarea. The browser displays the JavaScript as textarea content; it does not create or execute a bids script. This newly corrupts the publisher's visible form value, rather than merely retaining the old absence of ads on malformed input.

Verified by a temporary Rust test exercising `stream_publisher_body_async` with normal SSAT params and no dispatched auction, writing the actual output to `/tmp/pr872-rawtext.html`, then parsing that output with the existing JSDOM dependency. Result:

```json
{"bidsScripts":0,"textareaValue":"draft<script>(window.tsjs=window.tsjs||{}).bids=JSON.parse(\"{}\");(function(){var f=window.tsjs.adInit;if(typeof f===\"function\")f();})();</script>"}
```

Base comparison: at the merge base, `publisher.rs:654-657` returns `stream_publisher_body` immediately when no auction was dispatched, and `html_processor.rs:345-385` only inserts bids through the body end-tag handler. This exact fixture has no body end tag, so the base does not append the literal script to the textarea. Base behavior was established from source, not a separately compiled base checkout. The inability to render ads on arbitrary malformed HTML is preexisting and is not the finding.

Proposed fix: track whether EOF permits an HTML script insertion, or hold enough parser state to terminate the actual open context safely before insertion. If safe recovery is unavailable, skip the fallback rather than changing publisher text/form contents. Apply the same decision to buffered and streaming paths, and test the DOM result rather than `.contains(".bids=JSON.parse")`.

Conceptual code (requires new parser state/API):

```rust
if !found_placeholder && state.tracker.eof_allows_script_insertion() {
    // Resolve bids and append the EOF tail using the existing bounded writer.
}
```

Prose-only: parser tracking, both callers and regression tests need coordinated changes outside one hunk. This is an implementation direction, not compilable replacement code.

prk-Jr (changes requested):
## Summary

Parser-safe timing remains relevant to inline streaming despite the newer ESI path. This head introduces a disabled-configuration startup failure and an EOF fallback that can corrupt textarea contents; it also currently conflicts with main.

## Blocking

- **🔧 [P2] Skip runtime matcher preparation for disabled cache rules** — see inline at `crates/trusted-server-core/src/settings.rs:2043`.
- **🔧 [P2] Do not append executable fallback markup inside an open textarea** — see inline at `crates/trusted-server-core/src/publisher.rs:1523`.

## Validation and scope

Local formatting, native core clippy with warnings denied, four late-binder tests, and five SSAT tests passed. Scratch tests reproduced both findings; the malformed-output result was also parsed with JSDOM. Base behavior for the textarea case was verified from source, not a separate base build. Full adapter/parity gates were not rerun. Runtime diffs and relevant surrounding paths were reviewed, not every unchanged line of the large files/design documents.

GitHub reports merge conflicts. Current main has parser-owned ESI seams but still uses raw body-close scanning in inline paths, so selectively rebase this change rather than treating it as entirely superseded. Other release gates absent from the four reported checks below are not assumed to have passed.

## CI Status

- [integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/28979229221/job/85995211872): PASS
- [integration tests (Fastly EC lifecycle)](https://github.com/IABTechLab/trusted-server/actions/runs/28979229221/job/85995211858): PASS
- [browser integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/28979229221/job/85995211841): PASS
- [prepare integration artifacts](https://github.com/IABTechLab/trusted-server/actions/runs/28979229221/job/85993753043): PASS

crates/trusted-server-core/src/settings.rs
ChristianPavilonis: Addressed on the current branch by `42a34ae3`. `CacheAssetRule::prepare_runtime` now returns immediately for disabled rules, while enabled rules still run matcher and policy validation. The regression test `disabled_cache_asset_rules_defer_matcher_and_policy_validation` covers an invalid disabled draft rule. I also reran `cargo test-fastly` and the adapter test/lint matrix successfully.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Addressed in `3089b0da`. EOF is no longer treated as a safe markup insertion point: when no parser-confirmed body-end placeholder is emitted, Trusted Server preserves the rewritten publisher bytes, appends no bootstrap or bids markup, and abandons the pending auction with reason `unsafe_eof_insertion`.

DOM-level regressions now cover both an open `<textarea>` and an ordinary unclosed body in buffered and live-stream paths. The design and implementation plan were updated to document the content-preserving policy. `cargo test-fastly`, `cargo clippy-fastly`, the Axum/Cloudflare/Spin tests and lints, Rust formatting, and docs formatting all pass.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -1409,13 +1429,20 @@ fn create_html_stream_processor(
...
         .with_ad_state(params.ad_slots_script, params.ad_bids_state)
         .with_gpt_diagnostics(gpt_diagnostics)
         .with_body_close(body_close)
         .with_csp_nonce_observer(csp_nonce_observed)
         .with_datadome_client_tag_suppression(params.suppress_datadome_client_side_tag);
 
+    if let Some(late_binding) = late_binding {
+        config = config.with_bid_placeholder(
+            late_binding.placeholder.as_html().to_owned(),
+            late_binding.tracker,
```
prk-Jr, line 1442: 🔧 [P2] Run full-document post-processors after bid late binding

This config enables placeholder insertion but leaves full-document post-processing enabled. With Next.js registered, HtmlWithPostProcessing returns the entire rewritten document at EOF, and PlaceholderLateBinder then treats everything after the body placeholder as a streaming held tail. A page with a 70 KiB trailing HTML comment after </body> consequently fails with `SSAT placeholder held tail 71701 bytes exceeds 65536-byte limit`, even though it is below the configured body limit. The same fixture completes and contains the bids script with direct-state injection. Post-processors also receive the internal bid placeholder instead of final HTML.

Use the intended bounded buffered ordering: disable post-processing during rewriting, late-bind bids into a bounded buffer, then run the full-document post-processors with the same IntegrationDocumentState before encoding.

```text
decode → rewrite with post-processing disabled → late-bind bids
       → full-document post-processors → encode
```

Apply manually: the fix needs changes to both publisher orchestration and HTML post-processing helpers, so it cannot be one contiguous GitHub suggestion.

prk-Jr (changes requested):
## Summary

The PR adds parser-owned bid placeholders and preserves publisher content when no safe body-close insertion point exists. One reproduced regression remains in the buffered full-document post-processing path.

## Blocking

### 🔧 wrench

- **[P2] Run full-document post-processors after bid late binding** — see inline at `crates/trusted-server-core/src/publisher.rs:1437–1442`.

The inline finding is prose-only because the fix spans publisher orchestration and HTML post-processing helpers.

## Validation

A focused native integration test using the actual Next.js registry, HTML processor, and late binder reproduces the 64 KiB tail rejection. The direct-state control completes with a bids script; all four existing binder tests pass. The temporary test was removed and the review worktree is clean at `3089b0da93c9f68c5dbc1f11e344ede21d470d1a`.

## CI Status

- integration tests: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- cargo fmt: PASS (required)
- cargo test: PASS (required)
- vitest: PASS
- Analyze (actions): PASS

aram356:
Closing in place of other optimizations

PR #884, opened 2026-07-09
Add opt-in SSAT compression offload

### LINKED ISSUE #858: Measure and offload SSAT HTML compression from the Compute guest
## Summary

Measure and, if justified, offload SSAT HTML compression work from the Compute guest.

The current SSAT HTML rewrite path mirrors origin content encoding, which means gzip/deflate/brotli responses are decompressed and recompressed inside the guest. That CPU work is paid on every SSAT pageview because assembled HTML is private and not shared-cacheable.

## Problem

For large publisher documents, in-guest decompression/recompression can become a major vCPU cost and may increase risk of hitting Compute CPU limits. Fastly supports delivery compression via `X-Compress-Hint: on`, but using it with an `Accept-Encoding: identity` origin fetch changes wire-size behavior and needs measurement before becoming default.

## Proposed scope

- Measure current SSAT HTML compression CPU and wire sizes on representative pages.
- Test origin behavior for `Accept-Encoding: identity`.
- Test Fastly delivery compression with `X-Compress-Hint: on`.
- Compare current gzip/br-in + gzip/br-out behavior against identity-origin + delivery compression.
- Add a config flag or rollout guard if implementation proceeds.

## Measurement plan

For representative large and normal pages:

- Fetch origin with `Accept-Encoding: identity`, `gzip`, and `br`.
- Verify identity body equivalence after decompression.
- Run a flagged Fastly path using identity-origin + `X-Compress-Hint: on`.
- Compare:
  - `Content-Encoding`
  - transfer size
  - TTFB/FCP if available
  - guest CPU / vCPU-ms
  - error/truncation behavior

## Acceptance criteria

- [ ] Measurement compares current compression behavior vs identity-origin + delivery compression.
- [ ] Origin support for `Accept-Encoding: identity` is confirmed or failure modes are documented.
- [ ] Fastly `X-Compress-Hint: on` behavior is measured for representative HTML pages.
- [ ] Decision recorded: keep current behavior, enable offload by default, or expose as path/config-specific opt-in.
- [ ] If implemented, SSAT privacy behavior remains `private, max-age=0` and no shared cacheability is introduced.
- [ ] In-guest Brotli is not introduced as the default SSAT HTML path.

## Out of scope

- Origin-template caching.
- Parser-safe streaming/body-close hold.
- Static asset compression policy.

## Reference

See `docs/superpowers/specs/2026-07-06-cache-control-header-design.md` for why this is deferred from the initial cache-header PR.

ChristianPavilonis:
## Summary

- add a disabled-by-default `publisher.ssat_compression_offload_enabled` rollout guard
- explicitly declare delivery-compression capability at each adapter boundary so only Fastly can activate the offload path
- request identity HTML from the origin, process supported origin encodings to identity output, and let Fastly dynamically compress delivery with `X-Compress-Hint: on`
- preserve current mirror compression for disabled, non-SSAT, and non-Fastly paths
- preserve SSAT private cache behavior and document the origin bandwidth/10 MiB raw-body tradeoff

## Decision and measurement

The staging experiment found that identity-origin plus Fastly delivery compression reduced guest processing from approximately 44.6 ms to 9.0 ms and Fastly vCPU from 47 ms to 19 ms on the measured page. Browser delivery remained Brotli-compressed.

Because identity origin responses are substantially larger and still count against the current Fastly 10 MiB raw origin-response limit, this PR exposes the behavior as a publisher-level opt-in rather than changing the global default.

## Behavior

When all three conditions are true:

1. `publisher.ssat_compression_offload_enabled = true`
2. the adapter declares Fastly dynamic delivery compression
3. the existing server-side ad stack eligibility gates pass

Trusted Server sends `Accept-Encoding: identity` to the publisher origin. Supported gzip, deflate, or Brotli responses from origins that ignore the request are still decoded safely. Processed HTML leaves the guest without `Content-Encoding`, carries `X-Compress-Hint: on`, and varies on `Accept-Encoding` for Fastly delivery.

Axum, Cloudflare, and Spin explicitly declare the capability unavailable and retain existing compression behavior even when the shared setting is enabled.

## Validation

- `cargo test-fastly`
- `cargo test-axum`
- `cargo test-cloudflare`
- `cargo test-spin`
- `cargo clippy-fastly`
- `cargo clippy-axum`
- `cargo clippy-cloudflare`
- `cargo clippy-cloudflare-wasm`
- `cargo clippy-spin-native`
- `cargo clippy-spin-wasm`
- `cd crates/trusted-server-js/lib && npx vitest run`
- `cd crates/trusted-server-js/lib && npm run format`
- `cd crates/trusted-server-js/lib && node build-all.mjs`
- `npx --yes prettier@3.8.1 --check docs`

Closes #858

### DIFF
--- a/crates/trusted-server-adapter-fastly/src/app.rs
+++ b/crates/trusted-server-adapter-fastly/src/app.rs
@@ -119,6 +119,7 @@ use trusted_server_core::proxy::{
 use trusted_server_core::publisher::{
     buffer_publisher_response_async, handle_page_bids, handle_publisher_request,
     handle_tsjs_dynamic, page_bids_preflight_denied, AuctionDispatch,
+    DeliveryCompressionCapability, PublisherAdapterOptions,
 };
 use trusted_server_core::request_signing::{
     handle_deactivate_key, handle_rotate_key, handle_trusted_server_discovery,
@@ -795,7 +796,10 @@ async fn dispatch_fallback(
                             &mut ec.ec_context,
                             auction,
                             req,
-                            EdgeCacheHeader::SurrogateControl,
+                            PublisherAdapterOptions {
+                                edge_cache_header: EdgeCacheHeader::SurrogateControl,
+                                delivery_compression: DeliveryCompressionCapability::FastlyDynamic,
+                            },
                         )
                         .await
                         {
--- a/crates/trusted-server-adapter-fastly/src/app.rs
+++ b/crates/trusted-server-adapter-fastly/src/app.rs
@@ -96,6 +96,7 @@ use error_stack::Report;
 use trusted_server_core::auction::endpoints::handle_auction;
 use trusted_server_core::auction::AuctionTelemetrySink;
 use trusted_server_core::auction::{build_orchestrator, AuctionOrchestrator};
+use trusted_server_core::cache_policy::EdgeCacheHeader;
 use trusted_server_core::constants::{COOKIE_SHAREDID, COOKIE_TS_EIDS};
 use trusted_server_core::ec::batch_sync::handle_batch_sync;
 use trusted_server_core::ec::consent::ec_consent_withdrawn;
@@ -719,7 +720,7 @@ async fn dispatch_fallback(
     };
 
     let result = if uses_dynamic_tsjs_fallback(&method, &path) {
-        handle_tsjs_dynamic(&req, &state.registry)
+        handle_tsjs_dynamic(&req, &state.registry, EdgeCacheHeader::SurrogateControl)
     } else if state.registry.has_route(&method, &path) {
         // Integration-proxy responses are not bounded by publisher.max_buffered_body_bytes.
         // Only the handle_publisher_request branch below routes through
@@ -794,6 +795,7 @@ async fn dispatch_fallback(
                             &mut ec.ec_context,
                             auction,
                             req,
+                            EdgeCacheHeader::SurrogateControl,
                         )
                         .await
                         {
--- a/crates/trusted-server-core/src/publisher.rs
+++ b/crates/trusted-server-core/src/publisher.rs
@@ -40,6 +40,9 @@ use crate::auction::telemetry::{
 use crate::auction::types::{
     AuctionContext, AuctionRequest, Bid, DeviceInfo, PublisherInfo, SiteInfo, UserInfo,
 };
+use crate::cache_policy::{
+    cache_control_headers_are_private_or_no_store, CachePolicy, EdgeCacheHeader,
+};
 use crate::consent::{consent_allows_server_side_auction, gate_eids_by_consent};
 use crate::constants::{COOKIE_TS_EIDS, HEADER_X_COMPRESS_HINT};
 use crate::cookies::handle_request_cookies;
@@ -165,6 +168,7 @@ fn accept_encoding_qvalue(header_value: &str, target: &str) -> Option<f32> {
 pub fn handle_tsjs_dynamic(
     req: &Request<EdgeBody>,
     integration_registry: &IntegrationRegistry,
+    edge_header: EdgeCacheHeader,
 ) -> Result<Response<EdgeBody>, Report<TrustedServerError>> {
     const PREFIX: &str = "/static/tsjs=";
     const UNIFIED_FILENAMES: &[&str] = &["tsjs-unified.js", "tsjs-unified.min.js"];
@@ -179,10 +183,8 @@ pub fn handle_tsjs_dynamic(
         // Serve core + immediate modules (excludes deferred like prebid)
         let module_ids = integration_registry.js_module_ids_immediate();
         let body = trusted_server_js::concatenate_modules(&module_ids);
-        let mut resp = serve_static_with_etag(&body, req, "application/javascript; charset=utf-8");
-        resp.headers_mut()
-            .insert(HEADER_X_COMPRESS_HINT, HeaderValue::from_static("on"));
-        return Ok(resp);
+        let hash = trusted_server_js::concatenated_hash(&module_ids);
+        return Ok(serve_tsjs_static(req, &body, &hash, edge_header));
     }
 
     if let Some(module_id) = parse_deferred_module_filename(filename) {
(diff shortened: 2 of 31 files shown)


ChristianPavilonis:
Remove the opt in

crates/trusted-server-core/src/publisher.rs
```diff
@@ -2303,14 +2400,29 @@ pub async fn handle_publisher_request(
                 content_encoding
             );
 
+            let offload_stream = compression_offload_active && is_html_content_type(&content_type);
+            let input_compression = Compression::from_content_encoding(&content_encoding);
+            let output_compression = if offload_stream {
```
prk-Jr, line 2405: 🤔 **thinking** — Offload changes what `max_buffered_body_bytes` actually bounds.

Every delivery path funnels through `buffer_publisher_response_async`, which wraps a `BoundedWriter::new(settings.publisher.max_buffered_body_bytes)` (`publisher.rs:623`) — `stream_publisher_body` has no callers outside this module, so nothing streams straight to the client today.

With mirrored compression the 16 MiB cap applies to the **gzip/brotli** output. With `output_compression = None` it applies to the **identity** output. A page whose origin response is, say, 1 MiB gzip but decodes to 20 MiB of HTML passes today (re-encoded output is ~1 MiB) and starts returning an error once the flag is on (`publisher body exceeded maximum buffered size`).

The origin-side 10 MiB pressure the PR description worries about went away when origin negotiation was kept; this is where the size pressure actually landed. Worth a sentence in the rollout warning in `docs/guide/configuration.md` — something like "when enabled, `max_buffered_body_bytes` bounds the uncompressed HTML; raise it if pages are large or highly compressible" — so an operator flipping the flag knows which dial to check.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -893,10 +910,9 @@ async fn buffer_html_late_binding_with_postprocessors<W: Write>(
     };
     use crate::integrations::IntegrationDocumentState;
 
-    let compression = Compression::from_content_encoding(&ctx.params.content_encoding);
     let decoded = match decode_body_to_vec(
         body,
-        compression,
+        ctx.params.input_compression,
```
prk-Jr, line 915: 🤔 **thinking** — The buffered post-processor branch is offload-capable but untested.

The new tests all land in `stream_html_with_auction_hold`, because `crate_test_settings_str()` sets `[integrations.nextjs] enabled = false` and NextJS is the only integration registering an HTML post-processor — so `has_html_post_processors()` is false and `stream_publisher_body_async` never takes this branch.

For a Next.js publisher (a primary target for SSAT) the offload path is this function instead: decode with `input_compression`, then `encode_processed_html_to_writer(..., output_compression)`. The wiring reads correct, it just has no test asserting identity output on this path.

A cheap addition: reuse `ssat_compression_test_settings(true)` with `[integrations.nextjs] enabled = true` and assert the buffered output is identity and still contains the injected `tsjs` state.

prk-Jr (commented):
## Summary

Clean, well-gated rollout flag. The three-way gate (setting + adapter capability + existing SSAT eligibility), the explicit `DeliveryCompressionCapability` declaration at each adapter boundary, and the forward-compatible config serialization are all the right shape, and the test coverage on the new paths is genuinely good. One question about the PR description's stated behavior/measurement, plus a few non-blocking notes.

## Blocking

### ❓ question

- **PR description and measurement no longer describe the shipped behavior**: The description says "request identity HTML from the origin" and "Trusted Server sends `Accept-Encoding: identity` to the publisher origin". The code does not do this — `restrict_accept_encoding` is unchanged, and the new tests assert `gzip, br` is still forwarded to the origin (`crates/trusted-server-core/src/publisher.rs:534`, `:620`, `:756`). `docs/guide/configuration.md` describes the actual behavior correctly ("continues to negotiate only supported encodings with the publisher origin"), so only the PR body is stale — presumably from before "Fix SSAT compression offload rollout safety".

  Two consequences worth confirming before merge:

  1. The staging measurement (guest 44.6 ms → 9.0 ms, vCPU 47 ms → 19 ms) was taken with identity origin responses. What ships still decodes gzip/deflate/br in the guest and only drops the re-encode. The re-encode is likely the dominant cost, so most of the win probably survives — but the number in the description does not describe this build. Has it been re-measured with negotiated origin responses?
  2. The stated tradeoff ("identity origin responses are substantially larger and still count against the current Fastly 10 MiB raw origin-response limit") no longer applies, since origin responses stay compressed. The remaining size pressure moved to the in-guest buffer instead — see the `max_buffered_body_bytes` note below.

## Non-blocking

### 🤔 thinking

- **Offload shifts `max_buffered_body_bytes` from bounding compressed output to bounding uncompressed output** (`crates/trusted-server-core/src/publisher.rs:2405`)
- **No coverage for the HTML post-processor buffered path under offload** (`crates/trusted-server-core/src/publisher.rs:915`)

### 🌱 seedling

- **Delivery capability is a per-call-site literal** (`crates/trusted-server-adapter-fastly/src/app.rs:801`)

### 📌 out of scope

- **New env-override block documents a mechanism that is test-only at runtime** (`docs/guide/configuration.md:342`)

### 👍 praise

- **Capability decoupled from `EdgeCacheHeader`** (`crates/trusted-server-core/src/publisher.rs:1885`)
- **Forward-compatible config serialization with a test and a documented rollback order** (`crates/trusted-server-core/src/settings.rs:56`)
- **Fastly delivery billing warning in the docs** — uncompressed-size billing is exactly the kind of thing that gets discovered after rollout, not before.

## CI Status

`test.yml` and `format.yml` trigger only on pull requests to `main` or `feature/**`, so this PR (base `refactor/ssat-streaming`) ran only the four integration-test checks — all passing. The fmt/clippy/unit-test gates were run locally at `68c67c9`:

- clippy: PASS (`clippy-fastly`, `clippy-axum`, `clippy-cloudflare`, `clippy-cloudflare-wasm`, `clippy-spin-native`, `clippy-spin-wasm`)
- rust tests: PASS (core 1660 under Viceroy; axum, cloudflare, spin adapters)
- GitHub integration tests: PASS (4/4)

crates/trusted-server-adapter-fastly/src/app.rs
```diff
@@ -795,7 +796,10 @@ async fn dispatch_fallback(
                             &mut ec.ec_context,
                             auction,
                             req,
-                            EdgeCacheHeader::SurrogateControl,
+                            PublisherAdapterOptions {
+                                edge_cache_header: EdgeCacheHeader::SurrogateControl,
+                                delivery_compression: DeliveryCompressionCapability::FastlyDynamic,
```
prk-Jr, line 801: 🌱 **seedling** — The capability is declared as a literal at each call site, so it lives or dies by whoever adds the next publisher route.

That is a real improvement over inferring it from `EdgeCacheHeader` (the doc comment on the enum explains why, and it's the right call). But a second Fastly publisher dispatch path added later would default to whatever the author types, and the failure is silent — you just lose the offload, or gain it somewhere unintended.

Not for this PR: consider hanging the capability off the platform layer (`RuntimeServices`) so it is derived once per adapter rather than restated per call site.

docs/guide/configuration.md
```diff
@@ -299,6 +303,45 @@ openssl rand -base64 32
...
+Deploy a Trusted Server binary that supports this setting before enabling it.
+Before rolling back to an older binary, set the option to `false` and publish
+the configuration with the current CLI so the disabled field is omitted, then
+roll back the binary.
+:::
+
+**Environment Override**:
+
+```bash
+TRUSTED_SERVER__PUBLISHER__SSAT_COMPRESSION_OFFLOAD_ENABLED=true
```
prk-Jr, line 342: 📌 **out of scope** — This env-override block follows the existing convention in this document, but `TRUSTED_SERVER__*` overrides are test-only at runtime: `Settings::from_toml_and_env` is `#[cfg(test)]` and its own doc comment says "Runtime loading does not use this legacy helper". Nothing in `trusted-server-cli` reads `TRUSTED_SERVER__*` either.

That is pre-existing and affects every setting in this page, so it is not this PR's problem to fix. It matters a little more here than elsewhere, though: this is a rollout flag, so it is exactly the setting an operator will try to toggle from the environment during a canary, and it will silently do nothing. Worth either dropping the block from this one section or filing a docs follow-up for the page as a whole.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -1821,6 +1877,41 @@ async fn collect_stream_auction(
     }
 }
 
+/// Delivery-compression behavior explicitly supported by an adapter.
+///
+/// This capability is independent of [`EdgeCacheHeader`]. Adapters must declare
+/// it directly so cache-header selection cannot accidentally enable Fastly-only
+/// response delivery behavior.
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
```
prk-Jr, line 1885: 👍 **praise** — Declaring the capability separately from `EdgeCacheHeader` rather than inferring `SurrogateControl ⇒ Fastly` is the right call, and the doc comment says why. Non-Fastly adapters get `Unavailable` explicitly, so a shared config with the flag on cannot leak Fastly-only delivery behavior into Cloudflare/Spin/Axum — and `unavailable_delivery_compression_keeps_mirrored_gzip_behavior` pins that.

crates/trusted-server-core/src/settings.rs
```diff
@@ -46,6 +50,11 @@ pub struct Publisher {
     /// Keep this secret stable to allow existing links to decode.
     #[validate(custom(function = validate_redacted_not_empty))]
     pub proxy_secret: Redacted<String>,
+    /// Enables Fastly dynamic delivery compression for eligible server-side ad
+    /// stack HTML responses. Defaults to `false` and has no effect on runtimes
+    /// that do not explicitly declare support for Fastly dynamic delivery.
+    #[serde(default, skip_serializing_if = "is_false")]
```
prk-Jr, line 56: 👍 **praise** — `skip_serializing_if = "is_false"` against `Publisher`'s `deny_unknown_fields` is a nice bit of rollout care: a published config with the flag off stays loadable by an older binary. The paired test in `config.rs` asserts both directions, and the docs spell out the rollback order (disable → push → roll back the binary). Easy detail to miss; good that it's covered.

PR #885, opened 2026-07-10
Thread EC KV read through the request and recover orphaned cookies

### LINKED ISSUE #851: Thread the KV EID read through the request and fix the EC TTL decay bug
### Description

From the 2026-07-05 performance & architecture review (P0 finding 5 + KV/EC section).

**Blocking KV read gates the hot path.** `apply_auction_eids_and_device` does a synchronous KV get before anything is dispatched (`publisher.rs:1350-1362` -> `auction/endpoints.rs:309`). Every SSAT navigation pays that round trip in TTFB.

Fix: issue the origin fetch async first, then KV + dispatch, then await origin; or carry the entry from the EC context that already read it. Thread the `(KvEntry, generation)` read once per request through `RouteResult` into finalize and pull sync — saves 1-2M reads per 1M PV and removes a pre-send RMW read.

**EC TTL decay bug (directly undermines the Safari/Firefox CPM bet).** The `ts-ec` cookie lives 1 year, but the KV row expires 1 year after last *write*, and `upsert_partner_ids` fails closed on a missing key while nothing recreates the row when the cookie still exists. Long-lived quiet users silently lose their identity graph and all future partner enrichment.

Fix: recreate-on-missing in the upsert path, or a sampled TTL-refresh write (sampling keeps write cost bounded).

### Effort / expected effect

- **Effort:** Medium
- **Expected effect:** -1-2M reads/1M PV; protects the Safari/Firefox CPM bet.

_Sequencing item 5 of 10._

prk-Jr:
## Summary

- Remove the redundant per-request EC identity-graph reads that gated the publisher-navigation hot path, and overlap the Fastly origin fetch with the KV lookup + auction dispatch. One request-scoped snapshot now feeds auction EID resolution, finalize, and pull sync (was up to three reads).
- Recover an orphaned `ts-ec` cookie (valid cookie, missing KV row) by rotating to a freshly generated, KV-backed EC ID on consent-granted real-browser navigations — fixing the TTL-decay identity loss where quiet long-lived users silently lost their identity graph.
- Collapse pull sync's per-partner read-modify-write into one request-wide bulk CAS, and make withdrawal tombstones existing-key-only so a forged cookie can never create a KV root.
- Preserve publisher response streaming while starting the Fastly origin request before EC KV preload and auction dispatch. Pending-stream completion is an explicit platform capability and remains independent from ordinary eager-send streaming support.

## Changes

| File | Change |
| ---- | ------ |
| `ec/mod.rs` | `EcKvSnapshot` request-scoped state (NotRead/Missing/Present/Failed), EC-ID binding, recovery-eligibility flag; `generate_if_needed` uses bounded Add-only creation |
| `ec/kv.rs` | `load_snapshot`, `create_if_absent`, snapshot-aware bulk upsert, existing-key-only conditional tombstone; CAS/refresh semantics + tests |
| `ec/finalize.rs` | Consume/return snapshot, orphan recovery (browser-nav gated, atomic add-then-cookie), two-ID existing-only withdrawal |
| `ec/prebid_eids.rs` | Split validated update collection (`collect_eid_cookie_updates`) from KV persistence |
| `ec/pull_sync.rs` | Reuse finalized snapshot for eligibility; aggregate partner results across batches into one bulk write |
| `auction/endpoints.rs` | Resolve auction EIDs from a snapshot instead of performing KV I/O |
| `platform/http.rs` | Add explicit pending-response streaming capability, direct single-handle wait contract, request-method metadata, and buffered `select` safeguards |
| `platform/test_support.rs` | Model eager streaming and pending streaming independently; add direct-wait, bodiless-response, and capability-matrix coverage |
| `publisher.rs` | Origin/auction scheduling (pending-stream capable: origin start, KV preload, auction dispatch, direct wait; fallback: preload, dispatch, eager send); preserve request snapshot, rewrite, cache bypass, DataDome privacy, and failure telemetry |
| `http_util.rs` | Comment documenting the `fastly-ssl` scheme-signal layering wart (no behavior change) |
| `adapter-fastly/platform.rs` | Implement Fastly pending streaming with direct `PendingRequest::wait()`, retain method/stream metadata, preserve HEAD and bodiless-status framing, and reject stream-marked handles from auction `select` |
| `adapter-fastly/{app,main}.rs` | Carry snapshot + navigation eligibility through `EcRequestState`/`EcFinalizeState`; mutable finalize outcome into post-send pull sync |

## Closes

Closes #851

## Test coverage

- Concurrent pending path proves `origin start -> EC KV lookup -> auction dispatch -> origin wait`.
- Pending-stream support is tested independently from ordinary eager-send streaming support.
- Origin-start failure performs no KV read or auction dispatch.
- Origin-wait failure preserves the existing single `origin_proxy_error` abandonment outcome.
- Pending path verifies origin URI/Host rewriting, internal-header removal, conditional/range removal, cache bypass, the original client snapshot used for auction construction, configured publisher domain, DataDome suppression/privacy, and streamed response bodies.
- Direct pending wait, buffered auction `select`, HEAD, and `1xx`/`204`/`205`/`304` bodiless semantics are covered.

## Verification

- [x] `cargo test-fastly`
- [x] `cargo test-axum`
- [x] `cargo test-cloudflare`
- [x] `cargo test-spin`
- [x] `cargo clippy-fastly`
- [x] `cargo clippy-axum`
- [x] `cd crates/trusted-server-js/lib && node build-all.mjs`
- [x] `cd crates/trusted-server-js/lib && npx vitest run` — 45 files, 893 tests, no type errors
- [x] `cd crates/trusted-server-js/lib && npm run format`
- [x] `cd docs && npm run format`

## Merge readiness

- Integrate current main while preserving both forced-stream test fixtures and request-driven pending streaming.
- Update three image-proxy tests to consume streamed bodies while retaining their preflight, response-policy, and request-metadata assertions.
- Address review feedback on CAS warning formatting, missing-snapshot pull-sync diagnostics, explicit orphan-ID handling, and pull-sync KV-read accounting.
- Add an active-EC ESI cold/warm regression covering one snapshot per navigation and no origin fetch on a warm cache hit.

## PR #1013 compatibility

The current branch incorporates PR #1013. Cache-eligible ESI requests retain the origin request until the shared-template lookup: a warm hit performs no origin fetch, and a cold miss fetches the origin once. The origin-first pending optimization applies only when no template-cache key is available.

A regression test covers cold and warm ESI navigations with an active EC, a request-scoped KV lookup, and a client advertising concurrent fanout and pending streaming. It asserts one EC lookup per navigation, one origin request across both navigations, and preserved streaming and cache-bypass flags.

## Hardening note

N/A — no config-derived regex or pattern compilation touched. Fail-closed invariants covered by tests: KV errors never become authoritative misses (no rotation), tombstones are never revived, cookies are emitted only after a backing row exists, store/CAS failures return a failed snapshot rather than claiming request-local IDs were persisted, and stream-marked pending requests cannot enter the buffered auction-selection path.

## Checklist

- [x] New code has tests


crates/trusted-server-adapter-fastly/src/app.rs
```diff
@@ -407,6 +407,7 @@ fn build_ec_request_state(
         match EcContext::read_from_request_with_geo(settings, req, services, geo_info.as_ref()) {
             Ok(mut context) => {
                 context.set_device_signals(device_signals);
+                context.set_recovery_eligible(is_real_browser && is_navigation_request(req));
```
ChristianPavilonis: 🔧 **P1 — Orphan recovery is authorized for non-publisher and short-circuited requests**

Recovery eligibility is enabled during generic pre-routing for every browser-shaped document request, before route selection and request filters. Named routes and filter short circuits still reach EC finalization. With a missing format-valid `ts-ec` and a valid `ts-eids` or `sharedId` update, finalization performs its lazy lookup and rotates the identity.

**Why it matters:** A discovery/API/integration navigation or DataDome challenge can create a persistent EC root and cookie—and subsequently trigger pull sync—even though no publisher page was served. This exceeds the stated publisher-navigation boundary and introduces identity side effects on blocked responses.

**Suggested fix:** Default recovery eligibility to false and authorize it only inside the publisher fallback after filters and a successful origin start, or return an explicit publisher-recovery eligibility outcome from `handle_publisher_request`. Add lifecycle tests for named routes, filter short circuits, and origin-start failures.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -186,6 +195,30 @@ impl KvIdentityGraph {
...
+    /// Loads one request-scoped snapshot, preserving miss versus failure at the caller boundary.
+    #[must_use]
+    pub fn load_snapshot(&self, ec_id: &str) -> EcKvSnapshot {
+        match self.get(ec_id) {
+            Ok(Some((entry, generation))) => EcKvSnapshot::Present {
+                ec_id: ec_id.to_owned(),
+                entry: Box::new(entry),
+                generation: Some(generation),
+            },
+            Ok(None) => EcKvSnapshot::Missing {
```
ChristianPavilonis, line 207: 🔧 **P1 — A single eventually-consistent KV miss can rotate a valid identity**

`load_snapshot` classifies every `Ok(None)` as authoritative `Missing`, and `ec_finalize_response` immediately rotates that identity. Fastly edge data stores are eventually consistent, so a recently created live key can transiently appear absent at another POP. Additionally, `publisher.rs` can overwrite an Add-confirmed `Present` snapshot from `generate_if_needed` with such a miss.

**Why it matters:** Fresh, valid identities can be rotated unnecessarily, fragmenting partner mappings and leaving multiple year-lived roots. If the immediate post-Add preload misses, finalization can also suppress the newly generated cookie despite the successful write.

**Suggested fix:** Never downgrade an in-request Add-confirmed `Present` snapshot on a refresh miss/failure. For incoming cookies, require a strongly consistent or cross-request confirmation before orphan recovery rather than treating one edge miss as authoritative. Test transient Add→Missing→Present behavior.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -1343,7 +1392,11 @@ pub async fn handle_publisher_request(
     );
 
     let consent_context = ec_context.consent().clone();
-    let ec_id = ec_context.ec_value().filter(|_| ec_allowed);
+    let ec_id_owned = ec_context
```
ChristianPavilonis: 🔧 **P2 — Consent-withdrawn navigations skip the origin-overlapped snapshot read**

The ID used by `should_preload_ec_snapshot` is filtered through `ec_allowed()`. An existing EC under explicit withdrawal therefore becomes `None`, so the handler neither starts the concurrent origin path nor preloads the row.

**Why it matters:** Finalization waits until after the origin completes and then performs the KV lookup/CAS needed for withdrawal. This places new latency directly on the privacy-critical response path and contradicts the design requirement to preload active-EC navigations even when consent prevents auction dispatch.

**Suggested fix:** Maintain separate unfiltered active and consent-filtered auction EC IDs. Use the active ID only for internal snapshot preload; continue using the filtered ID for auction identity and EIDs.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -445,6 +495,86 @@ impl KvIdentityGraph {
...
+        ec_id: &str,
+        updates: &[PartnerIdUpdate],
+        snapshot: EcKvSnapshot,
+    ) -> EcKvSnapshot {
+        if updates.is_empty() {
+            return snapshot;
+        }
+
+        let mut current = snapshot;
+        for _attempt in 0..MAX_CAS_RETRIES {
```
ChristianPavilonis, line 529: 🔧 **P2 — Snapshot refreshes consume one of the advertised CAS attempts**

In both partner upsert and conditional tombstoning, an initial `NotRead` or generation-unavailable snapshot spends one loop iteration performing only a lookup. That leaves four write attempts instead of the five attempts existing CAS paths receive.

**Why it matters:** Generation-unavailable snapshots are normal after successful creation or mutation. Under four conflicts, enrichment or a withdrawal tombstone fails even though the fifth configured CAS attempt could succeed.

**Suggested fix:** Resolve the initial usable snapshot outside the retry counter, or increment the counter only after a CAS precondition failure. Add a test with four conflicts followed by a successful fifth write.

crates/trusted-server-core/src/auction/endpoints.rs
```diff
@@ -247,7 +246,11 @@ pub async fn handle_auction(
 
     // Resolve partner EIDs from the KV identity graph when the user has
     // a valid EC and both KV and partner stores are available.
-    let eids = resolve_auction_eids(kv, registry, ec_context);
+    let auction_kv_snapshot = match (kv, ec_id) {
```
ChristianPavilonis: 🔧 **P2 — Auction endpoints perform KV reads when the snapshot cannot be used**

`/auction` loads a snapshot before determining whether a partner registry exists. `handle_page_bids` similarly loads at `publisher.rs:2323` before checking whether the auction is enabled, slots matched, or the request is a bot/prefetch.

**Why it matters:** Requests that cannot consume server-side EIDs now incur billable KV reads that the previous resolver avoided, working against this PR’s core read-reduction objective.

**Suggested fix:** Defer these endpoint-local loads until the auction branch actually runs and a non-empty registry is available.

ChristianPavilonis (changes requested):
## Summary

Changes requested. The snapshot/recovery design is promising, but I found two high-severity identity/privacy issues and three medium correctness, latency, and resource-usage issues. Details are attached inline.

crates/trusted-server-adapter-fastly/src/app.rs
prk-Jr: Fixed in 81896a916.

Recovery eligibility now defaults to `false` in `build_ec_request_state`. It is authorized only inside the publisher fallback, in the `Ok(pub_response)` arm — after request filters pass and the origin start succeeds — gated on `is_real_browser && is_navigation_request`. Named routes, integration proxies, and request-filter short circuits never reach that point, so they can no longer rotate identity.

Lifecycle tests added in `app.rs`: `named_route_response_is_not_recovery_eligible`, `filter_short_circuit_response_is_not_recovery_eligible`, and `publisher_navigation_origin_start_failure_is_not_recovery_eligible` (the harness has no live backend, so a real-browser navigation exercises the origin-start-failure path directly).

crates/trusted-server-core/src/ec/kv.rs
prk-Jr: Fixed in 81896a916 with two guards:

1. No-downgrade: the origin-overlapped preload in `publisher.rs` never overwrites an in-request Add-confirmed `Present` snapshot with a `Missing`/`Failed` refresh. A fresh create that reads back absent on an eventually-consistent store no longer suppresses the new cookie or rotates.

2. Confirm-before-rotate: `confirm_then_recover_orphaned_ec` in `finalize.rs` performs a second authoritative `load_snapshot` — separated from the preload by the full origin round trip — before rotating. A now-visible row is adopted (pending updates merged); only a confirmed authoritative miss rotates; a read failure never rotates. Test: `finalize_transient_missing_row_confirms_present_and_does_not_rotate`.

On "strongly consistent": `EcKvStore::lookup` has no consistency parameter and re-exports from `edgezero_core`, so a true strong-consistency read would require a cross-adapter trait change (and likely an upstream change). The confirming re-read is the strongest guard available within the current abstraction. Happy to open a follow-up if a strongly-consistent read path is wanted.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Fixed in 81896a916.

`handle_publisher_request` now keeps two IDs: an unfiltered `active_ec_id` (drives the origin-overlapped snapshot preload and consent-withdrawal tombstoning) and the consent-filtered `ec_id` (auction identity/EIDs). `should_preload_ec_snapshot` and the preload use `active_ec_id`, so a consent-withdrawn navigation still reads the row while the origin is in flight — the withdrawal CAS stays off the post-origin latency path. Auction dispatch continues to see no EC under withdrawal because it still uses the filtered `ec_id`.

crates/trusted-server-core/src/ec/kv.rs
prk-Jr: Fixed in 81896a916.

Both `upsert_partner_ids_from_snapshot` and `tombstone_existing_from_snapshot` now resolve the initial usable snapshot outside the retry counter. A `NotRead`, generation-unavailable, or foreign-ID snapshot no longer spends a loop iteration on a lookup-only pass, so all five write attempts remain.

Tests added: `snapshot_upsert_gen_unavailable_survives_four_conflicts_then_writes` and `tombstone_gen_unavailable_survives_four_conflicts_then_writes` — each starts from a generation-unavailable snapshot, hits four CAS conflicts, and succeeds on the fifth write. Both fail under the previous accounting.

crates/trusted-server-core/src/auction/endpoints.rs
prk-Jr: Fixed in 81896a916.

`/auction` now gates the snapshot load on `registry.is_some()` — without a registry `resolve_auction_eids` returns `None`, so the read was unused. `handle_page_bids` moves its load inside the live-auction branch, so it runs only when the auction is enabled, consent-granted, slots matched, not a bot/prefetch, and a partner registry exists. Kill-switch, no-slot, bot/prefetch, and no-registry requests now incur no identity-graph read.

aram356:
@prk-Jr Please confirm that is still issue with SSAT and ESI

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -638,6 +783,93 @@ impl KvIdentityGraph {
...
+        // `MAX_CAS_RETRIES` iterations stay available for the tombstone write.
+        let mut current = match snapshot {
+            EcKvSnapshot::Present {
+                ec_id: ref snapshot_id,
+                generation: Some(_),
+                ..
+            } if snapshot_id == ec_id => snapshot,
+            EcKvSnapshot::Missing {
+                ec_id: ref snapshot_id,
+            } if snapshot_id == ec_id => return snapshot,
```
ChristianPavilonis, line 810: 🔧 **P1 — A preloaded miss can drop an explicit withdrawal**

A publisher preload can observe `Missing` at an eventually-consistent Fastly POP. This return treats that stale result as final, so finalization deletes the browser cookie but never writes the authoritative tombstone; the live row can remain `consent.ok = true` and accept later identity/EID operations.

Please revalidate a preloaded `Missing` after the origin round trip and tombstone only when that confirmation is `Present`. That preserves existing-key-only behavior for forged cookies. Add a `Missing -> Present` withdrawal test.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -445,6 +495,101 @@ impl KvIdentityGraph {
...
+                generation: Some(_),
+                ..
+            } if snapshot_id == ec_id => snapshot,
+            EcKvSnapshot::Missing {
+                ec_id: ref snapshot_id,
+            }
+            | EcKvSnapshot::Failed {
+                ec_id: ref snapshot_id,
+            } if snapshot_id == ec_id => return snapshot,
+            _ => self.load_snapshot(ec_id),
```
ChristianPavilonis, line 526: 🔧 **P1 — EID ingestion re-downgrades a successful Add**

`generate_if_needed` records a successful `Add` as `Present { generation: None }`. When `ts-eids` or `sharedId` has an update, this branch refreshes that proven snapshot; a transient miss/failure then reaches finalization as `Missing`/`Failed`, which suppresses the new `ts-ec` cookie despite the successfully persisted root. The publisher preload guard does not cover this finalization refresh.

Retain explicit Add-confirmation provenance and do not replace it with a best-effort EID-write refresh failure. The update can be skipped/logged, but cookie issuance should continue from the successful Add. Please add an Add-success + lookup-miss test with a valid EID update.

crates/trusted-server-core/src/ec/pull_sync.rs
```diff
@@ -89,16 +92,12 @@ pub fn dispatch_pull_sync(
...
-    let kv_entry = match kv.get(context.ec_id()) {
-        Ok(entry) => entry.map(|(entry, _)| entry),
-        Err(err) => {
-            log::warn!(
-                "Pull sync: failed to read identity graph for '{}': {err:?}",
-                super::log_id(context.ec_id())
-            );
-            return;
-        }
+    let Some(kv_entry) = context.snapshot.entry_for(context.ec_id()) else {
```
ChristianPavilonis, line 95: 🔧 **P1 — Post-send pull sync can disclose an ID after withdrawal**

This authorizes post-send partner calls from the earlier request snapshot without checking whether a concurrent request has since tombstoned the EC. After the page response reaches the browser, a CMP withdrawal can complete before this callback dispatches pull sync, yet the stale live snapshot still sends the raw `ec_id` to partners.

Revalidate live, non-tombstoned consent immediately before dispatching post-send pulls (through a consistency-safe revocation check). Add a concurrency regression test that tombstones after snapshot capture and asserts no pull request is sent.

crates/trusted-server-core/src/auction/endpoints.rs
```diff
@@ -253,9 +253,15 @@ pub async fn handle_auction(
...
     };
 
-    // Resolve partner EIDs from the KV identity graph when the user has
-    // a valid EC and both KV and partner stores are available.
-    let eids = resolve_auction_eids(kv, registry, ec_context);
+    // Resolve partner EIDs from the KV identity graph when the user has a valid
+    // EC and both KV and partner stores are available. Gate the read on a
+    // present registry: without one, `resolve_auction_eids` yields no
+    // server-side EIDs, so the snapshot would be an unused billable KV read.
+    let auction_kv_snapshot = match (kv, ec_id, registry) {
```
ChristianPavilonis, line 260: 🔧 **P2 — Endpoint snapshots are not shared with finalization**

This snapshot is local, while Fastly passes an unchanged immutable `EcContext` into response finalization. If `ts-eids`/`sharedId` has updates, finalization sees `NotRead` and performs a second billable lookup. The same issue exists in the page-bids local snapshot at `publisher.rs:4117`.

Thread a mutable `EcContext` through these handlers and store the loaded snapshot before returning (or otherwise return it to the Fastly caller). Add a counting-KV test covering endpoint EID resolution plus finalization.

crates/trusted-server-core/src/ec/finalize.rs
```diff
@@ -97,13 +114,130 @@ pub fn ec_finalize_response(
...
+) {
+    let confirmed = graph.load_snapshot(ec_id);
+    match confirmed {
+        EcKvSnapshot::Present { .. } => {
+            // The row became visible after the origin round trip: adopt it and
+            // merge any pending updates rather than rotating a valid identity.
+            let merged = graph.upsert_partner_ids_from_snapshot(ec_id, updates, confirmed);
+            ec_context.set_kv_snapshot(merged);
+        }
+        EcKvSnapshot::Missing { .. } => {
```
ChristianPavilonis, line 235: 🔧 **P2 — Two eventually-consistent misses are still treated as authoritative absence**

The confirmation read reduces the chance of a false rotation but cannot prove absence on an eventually-consistent store. If both reads are stale, this creates a replacement ID and abandons a still-live identity graph, fragmenting its accumulated EIDs.

Do not rotate unless the backend can provide a consistency-safe absence proof; otherwise fail closed and retry recovery on a later navigation. Add a two-miss test with the original row becoming visible afterward.

ChristianPavilonis (changes requested):
## Summary

Request changes. I found three P1 identity/consent races and two P2 snapshot-contract regressions in the request-scoped EC snapshot implementation. All findings are inline.

crates/trusted-server-core/src/ec/kv.rs
prk-Jr: Fixed in af56a10f.

`tombstone_existing_from_snapshot` no longer short-circuits on a `Missing` snapshot. It falls through to the re-read, which on the publisher path is separated from the preload by the full origin round trip, so a `Missing -> Present` withdrawal now writes the tombstone. Test: `tombstone_revalidates_preloaded_missing_and_writes_when_row_is_present`.

The second miss is also covered. A point read can't prove absence, so a re-read that still misses is escalated to `KvIdentityGraph::key_exists_confirmed`, which lists the key prefix — Fastly's KV list reads the primary data source unless `eventual_consistency()` is requested, so an empty page is a real absence proof. Proven-absent stays a no-op, preserving existing-key-only behavior for forged cookies. A key the store still lists gets an unconditional tombstone: no CAS generation survives a missed read, and a withdrawal must win over a concurrent write. A failed existence check returns `Failed` rather than reporting a completed withdrawal. Tests: `tombstone_writes_unconditionally_when_both_point_reads_miss_a_listed_row`, `tombstone_fails_closed_when_the_existence_check_fails`.

crates/trusted-server-core/src/ec/kv.rs
prk-Jr: Fixed in af56a10f.

`upsert_partner_ids_from_snapshot` now captures the existence proof carried by an incoming `Present` snapshot for this EC ID — an `Add`-confirmed create from `generate_if_needed` or an earlier authoritative read — and keeps it across the refresh. A refresh that reads absent or unreadable skips the update, logs it, and returns the proven snapshot, so cookie issuance continues from the successful `Add`.

Write failures deliberately still return `Failed`: those are real operation failures rather than ambiguous reads, and `dispatch_pull_sync` keys its persistence warning on them. Tests: `snapshot_upsert_keeps_add_confirmed_present_when_refresh_misses` (Add success + lookup miss + a valid EID update), `snapshot_upsert_keeps_add_confirmed_present_when_refresh_fails`, and `snapshot_upsert_without_proof_still_reports_a_refresh_miss` to pin that an unproven miss still reaches orphan recovery as `Missing`.

crates/trusted-server-core/src/ec/pull_sync.rs
prk-Jr: Fixed in af56a10f.

`dispatch_pull_sync` re-reads the live row immediately before the first partner request. Anything short of a live, consenting entry cancels the dispatch, and the loop's per-partner eligibility check now runs against the live entry rather than the request snapshot. Test: `dispatch_pull_sync_skips_dispatch_when_ec_is_tombstoned_after_snapshot_capture` tombstones the row after snapshot capture and asserts no backend request is recorded and no UID is written back.

Two notes on cost and limits. The read sits behind the snapshot-based eligibility filter, so a dispatch with nothing to pull still costs no KV operation (`dispatch_pull_sync_skips_revalidation_read_when_no_partner_is_eligible`), and the fresh snapshot feeds the write-back instead of the stale one, which removes a CAS conflict round. On "consistency-safe": the strong path here is the list API, which proves existence but not entry contents, and a tombstone is a written row rather than a deletion — so revocation still has to be read through `lookup`. That narrows the window from the whole page-response duration to the instant before dispatch, but does not close it. If you want it fully closed I'd rather do it as a follow-up on a strongly consistent value read than approximate it here.

crates/trusted-server-core/src/auction/endpoints.rs
prk-Jr: Fixed in 65180a7e.

`handle_auction` and `handle_page_bids` now take `&mut EcContext` and store the loaded snapshot before returning, so finalization ingests `ts-eids`/`sharedId` updates from the read the endpoint already performed. `ec_id` and the consent context are taken by value in both handlers so the store-back does not conflict with a live borrow of the context. The Fastly router already held `&mut EcRequestState`; the other adapters own their `EcContext` locally and only needed the binding made mutable.

Test: `auction_endpoint_snapshot_is_reused_by_response_finalization` runs `/auction` then `ec_finalize_response` against a counting KV store and asserts exactly one lookup across both, plus that the `sharedId` update still reaches the row. It counts two without the fix.

crates/trusted-server-core/src/ec/finalize.rs
prk-Jr: Fixed in af56a10f.

You were right that a second point read is not a proof. There is a consistency-safe absence proof available, though — it just isn't on `lookup`. Fastly's KV list defaults to `ListMode::Strong` and reads the primary data source (`eventual_consistency()` is opt-in), and `EcKvStore::count_keys_with_prefix` already goes through it on every adapter. `KvIdentityGraph::key_exists_confirmed` lists the full EC ID as a prefix with a limit of 1; EC IDs are fixed-width `{64hex}.{6alnum}`, so it matches at most the key itself.

`confirm_then_recover_orphaned_ec` now rotates only on `Ok(false)`. A key the store still lists means both point reads were stale, so the identity is left alone and recovery is retried on a later navigation; a failed existence check does not rotate either. Tests: `finalize_two_missing_reads_do_not_rotate_a_row_the_store_still_lists` (two misses, row still live and visible to list, asserts no rotation and that the original row survives) and `finalize_does_not_rotate_when_the_existence_check_fails`.

This does add one list operation per rotation candidate, but only after two misses on a recovery-eligible publisher navigation, so it is off the common path.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -464,6 +514,153 @@ impl KvIdentityGraph {
...
+        // once; an authoritative miss or failure for this EC ID is returned
+        // as-is (the hot path never retries a failed lookup). This keeps all
+        // `MAX_CAS_RETRIES` iterations available for actual writes.
+        let mut current = match snapshot {
+            EcKvSnapshot::Present {
+                ec_id: ref snapshot_id,
+                generation: Some(_),
+                ..
+            } if snapshot_id == ec_id => snapshot,
+            EcKvSnapshot::Missing {
```
ChristianPavilonis: :wrench: **P2: A transient endpoint miss suppresses EID persistence**

Fastly point lookups are eventually consistent, but a matching `Missing` snapshot returns without revalidation. `/auction` and `/_ts/page-bids` save their first lookup into `EcContext`; finalization then drops valid `ts-eids` or `sharedId` updates when that lookup was a stale miss. These named routes cannot run orphan recovery.

A live EC row can therefore fail to receive a newly collected partner ID until a later request retries. Before this PR, finalization's `ingest_eid_cookies` performed its own read, which could observe the row after replication converged. Existing tests cover a present snapshot and a genuinely missing root, but not `Missing -> Present` during endpoint finalization.

Please revalidate a matching `Missing` snapshot once when updates exist, apply the CAS update if the row becomes visible, and retain the no-create behavior for confirmed absence. A `Missing -> Present` endpoint/finalization test would cover the regression.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -657,6 +854,170 @@ impl KvIdentityGraph {
...
+                        "conditional withdrawal tombstone failed for '{}': {err:?}",
+                        log_id(ec_id)
+                    );
+                    return EcKvSnapshot::Failed {
+                        ec_id: ec_id.to_owned(),
+                    };
+                }
+            }
+        }
+        EcKvSnapshot::Failed {
```
ChristianPavilonis, line 1038: :wrench: **P2: CAS-exhausted withdrawals fail without telemetry**

After five tombstone CAS conflicts, this returns `EcKvSnapshot::Failed` without logging the exhaustion. Finalization does not inspect or log that outcome, and completely discards it when withdrawing a non-active cookie ID.

The browser cookie is cleared while the KV row can remain live with `consent.ok = true`. Server-side users of the retained EC ID can continue treating it as consented, with no log telling operators that withdrawal enforcement failed. Tests cover four conflicts followed by a successful fifth write, but not exhaustion. The previous unconditional tombstone path logged write failures.

Please log CAS exhaustion with the redacted EC ID and have finalization report every failed withdrawal outcome. Response handling can remain best-effort, but a five-conflict regression test and an operational counter or warning would make this failure observable.

ChristianPavilonis (approved):
## Summary

Reviewed `7b71f83331a5506e078537c749d283ff7fc9324a` against `616b74051a4ec4147160d882ae8ac77bfe6785dd`. Two medium-severity findings are included inline. All reported CI checks pass, and the existing review threads were checked to avoid duplicate feedback.

crates/trusted-server-core/src/ec/kv.rs
prk-Jr: Confirmed and fixed in ab291a70.

Traced it before changing anything: `/auction` stores its lookup into the context (`auction/endpoints.rs:272`), `recovery_eligible` is set only for publisher navigations (`adapter-fastly/src/app.rs:873`), and `main`'s `upsert_partner_ids` performed its own `get` at finalize time. So the short-circuit was a real regression window on exactly those routes.

`upsert_partner_ids_from_snapshot` now revalidates a matching `Missing` once before the updates are dropped. Only a `Failed` lookup still short-circuits — the hot path does not retry a read that already errored. A refresh that still misses returns `Missing` and creates nothing, so the no-create behavior and the downstream orphan-recovery branch in `ec_finalize_response` are unchanged.

Tests added:
- `snapshot_upsert_revalidates_transient_missing_and_persists`
- `snapshot_upsert_confirmed_missing_still_never_creates`
- `snapshot_upsert_failed_snapshot_is_not_revalidated`
- `finalize_named_route_transient_miss_still_persists_eid_updates` and `finalize_named_route_confirmed_miss_does_not_create_a_row`, both with `recovery_eligible = false`, covering the endpoint/finalization `Missing -> Present` case you asked for.

One cost note: on a publisher navigation where the preload missed *and* there are pending updates, `confirm_then_recover_orphaned_ec` now repeats a read the upsert already did. That is the orphan path only, and deduping it means threading the refreshed snapshot out of the upsert and into the confirm step, which widens the change well past the fix. Left as-is — say the word if you would rather have that plumbed.

crates/trusted-server-core/src/ec/kv.rs
prk-Jr: Confirmed and fixed in ab291a70. CAS exhaustion was the only tombstone exit without a log — the write-error and existence-check-error paths already warn.

- `tombstone_existing_from_snapshot` warns on exhaustion with the redacted EC ID, naming the actual hazard: the identity-graph row may still be live with consent granted.
- `upsert_partner_ids_from_snapshot`'s exhaustion tail was silent in the same way, so it warns too (partner updates not persisted). Slightly beyond the comment, same defect class, one log line.
- `ec_finalize_response` now warns on every `EcKvSnapshot::Failed` withdrawal outcome, including the non-active cookie ID whose outcome was discarded entirely — only the active ID's outcome is stored back on the context. Response handling stays best effort.

Test added: `tombstone_existing_from_snapshot_returns_failed_after_cas_exhaustion` — `MAX_CAS_RETRIES` conflicts, asserts `Failed` *and* that the stored row is still `consent.ok = true`, which is the exact divergence the warning exists to surface.

No counter: there is no metrics facility in the repo, so a `log::warn!` is the observable signal rather than inventing infrastructure to satisfy the review. Happy to wire one up if you want to point me at where it should live.

jwrosewell:
No objection to the change itself. One sequencing note before it lands.

This PR and the provider stack in #1043–#1047 and #1094 both change the Edge Cookie path and overlap on `ec/kv.rs`, `ec/finalize.rs`, `ec/pull_sync.rs` and `publisher.rs`, so whichever merges second will need to resolve conflicts there. Could this merge behind that stack?

If the stack can't be merged quickly and #885 needs to go first, the alternative we put to Rowena, Jason and Shailley is a release branch model, keeping the current live and next release branches in sync so neither blocks the other.

Either order works for us. Flagging it now so whoever merges first knows the other is coming.

jwrosewell:
**Sequencing note on this PR and the provider stack.**

This PR and the open provider stack (#1043 to #1047, #1084, #1094) rework the same Edge Cookie code. A merge simulation of this PR's head (d414740a8) against the stack's tip (35584dbf3) conflicts in 8 files, and against #1043 alone, the first PR of the stack, in 6 files, so whichever lands second re-resolves this by hand across the chain:

- `crates/trusted-server-adapter-axum/src/app.rs`
- `crates/trusted-server-adapter-cloudflare/src/app.rs`
- `crates/trusted-server-adapter-spin/src/app.rs`
- `crates/trusted-server-core/src/auction/endpoints.rs`
- `crates/trusted-server-core/src/ec/finalize.rs`
- `crates/trusted-server-core/src/ec/mod.rs`
- `crates/trusted-server-core/src/ec/pull_sync.rs`
- `crates/trusted-server-core/src/publisher.rs`

Reproduce:

```
git fetch upstream main refs/pull/885/head:pr-885
git merge-tree --write-tree --name-only pr-885 <stack head>
```

The request is the one we have made on #940 and #1094. The stack has been open since 19 August, carrying work that has been under review since 2 July as #838, is green on required CI, and its branches are kept rebased close to `main`, so please land the stack first, or say here that this PR goes first so we rebase once against a known base. The KV read-through and orphaned-cookie recovery this PR does are close to the stack's read-back and finalize changes, so sequencing them deliberately will also make both reviews easier than untangling a surprise conflict later.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -464,6 +514,166 @@ impl KvIdentityGraph {
...
+                    );
+                    return EcKvSnapshot::Failed {
+                        ec_id: ec_id.to_owned(),
+                    };
+                }
+            }
+        }
+
+        log::warn!(
+            "snapshot partner upsert for '{}': CAS conflict after {MAX_CAS_RETRIES} retries;              {} partner updates were not persisted",
```
aram356: ⛏ **nitpick** — This string literal contains a run of spaces where a `\` line-continuation was evidently intended, so the emitted warning reads `…retries;              3 partner updates were not persisted`. Neighboring multi-line log messages in this file use the `\` continuation.

```suggestion
            "snapshot partner upsert for '{}': CAS conflict after {MAX_CAS_RETRIES} retries; \
             {} partner updates were not persisted",
```

_(compile-verified: `cargo fmt --all -- --check` and `cargo clippy-fastly` pass with this applied — pure log-output change)_

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -657,6 +867,177 @@ impl KvIdentityGraph {
...
+                        ec_id: ec_id.to_owned(),
+                    };
+                }
+            }
+        }
+        // Withdrawal enforcement lost every CAS race, so the row can still be
+        // live with consent granted while the browser cookie is cleared. That
+        // divergence is only visible to operators if it is logged here.
+        log::warn!(
+            "withdrawal tombstone for '{}': CAS conflict after {MAX_CAS_RETRIES} retries; the              identity-graph row may still be live with consent granted",
```
aram356: ⛏ **nitpick** — Same embedded-spaces defect as the partner-upsert warning above. This is the operator-facing message the adjacent comment says exists precisely so the cookie/KV divergence is visible — worth keeping it legible.

```suggestion
            "withdrawal tombstone for '{}': CAS conflict after {MAX_CAS_RETRIES} retries; the \
             identity-graph row may still be live with consent granted",
```

_(compile-verified: `cargo fmt --all -- --check` and `cargo clippy-fastly` pass with this applied — pure log-output change)_

crates/trusted-server-core/src/ec/pull_sync.rs
```diff
@@ -89,16 +104,12 @@ pub fn dispatch_pull_sync(
...
-        Err(err) => {
-            log::warn!(
-                "Pull sync: failed to read identity graph for '{}': {err:?}",
-                super::log_id(context.ec_id())
-            );
-            return;
-        }
+    let Some(request_entry) = context.snapshot.entry_for(context.ec_id()) else {
+        return;
     };
```
aram356, line 114: 🤔 **thinking** — This exit narrows pull sync relative to main, and it does so silently. On main, `dispatch_pull_sync` read KV itself, so any real-browser request (subresources included) could dispatch. Now dispatch requires a `Present` request snapshot, which only navigations (preload), `/auction`, `/_ts/page-bids`, and requests whose EID-cookie ingestion loaded the row produce — a real-browser subresource request with a valid EC and no EID cookies, previously a dispatch opportunity, now exits here with no log at all, while every other ineligibility exit in this function logs. The narrowing itself looks like the intended KV-read saving from the design doc; the silent exit is the actionable part.

```suggestion
    let Some(request_entry) = context.snapshot.entry_for(context.ec_id()) else {
        log::debug!(
            "Pull sync: skipping dispatch for '{}' because the request captured no usable \
             identity snapshot",
            super::log_id(context.ec_id())
        );
        return;
    };
```

_(compile-verified: `cargo fmt --all -- --check` and `cargo clippy-fastly` pass with this applied)_

crates/trusted-server-core/src/ec/finalize.rs
```diff
@@ -97,13 +126,144 @@ pub fn ec_finalize_response(
...
+        let snapshot = graph.upsert_partner_ids_from_snapshot(
+            &ec_id,
+            &updates,
+            ec_context.kv_snapshot().clone(),
+        );
+        ec_context.set_kv_snapshot(snapshot);
+        if ec_context.kv_snapshot().entry_for(&ec_id).is_some() {
+            set_ec_cookie_on_response(settings, ec_context, response);
+        } else {
+            log::warn!("Skipping generated EC cookie because backing row is not authoritative");
```
aram356, line 144: 🤔 **thinking** — When EID updates are present and the partner-ID CAS *write* fails (or exhausts its five attempts), `upsert_partner_ids_from_snapshot` returns `Failed` and this branch suppresses the cookie — even though the `Add` in `generate_if_needed` already proved the backing row exists. That is asymmetric with `keep_proven`, which correctly refuses to let a failed *read* retract the Add-confirmed proof: a failed write doesn’t disprove existence either. Net effect: the user churns to a fresh ID on the next visit and the just-created row becomes KV garbage, with no privacy gain (the row is live either way). Low probability, and the fail-closed choice is documented as deliberate — raising it as a design question, not a bug. A fix would carry the existence proof through the write-failure path for cookie issuance while still logging the enrichment failure.

crates/trusted-server-core/src/ec/mod.rs
```diff
@@ -297,20 +363,45 @@ impl EcContext {
...
-                    "Failed to create or revive EC entry for id '{}' after generation: {err:?}",
-                    log_id(ec_value),
-                );
-                self.ec_value = None;
-                self.ec_generated = false;
-                return Err(err.change_context(TrustedServerError::EdgeCookie {
-                    message: "Failed to persist generated EC ID to KV identity graph".to_string(),
-                }));
+            if let Some(graph) = kv {
+                match graph.create_if_absent(&ec_id, &entry) {
```
aram356, line 367: ♻️ **refactor** — This migration (and the pull-sync/finalize ones) leaves several `pub` entry points with zero production callers — they’re reachable only from their own tests, and no dead-code lint fires because they’re `pub` in a lib crate. Verified across all workspace crates:

- `KvIdentityGraph::create_or_revive` (`ec/kv.rs:362`) — last caller was this function, now on `create_if_absent`. (Generation no longer reviving tombstones is a behavior improvement worth keeping — a withdrawn ID stays withdrawn.)
- `KvIdentityGraph::upsert_partner_id` (`ec/kv.rs:689`) — last caller was pull sync’s `drain_pull_batch`, now aggregated into the bulk snapshot path.
- `ingest_eid_cookies` (`ec/prebid_eids.rs:158`) plus wrappers `ingest_prebid_eids` (`:197`) and `ingest_sharedid_cookie` (`:322`) — finalize now uses `collect_eid_cookie_updates` + snapshot upserts.

Apply manually — the deletion spans multiple files and their tests, so it can’t be a one-click suggestion. Removing them (here or in a fast follow-up) makes the snapshot-aware paths the only entry points, so future callers can’t reach for a non-snapshot write path that bypasses the new invariants.

crates/trusted-server-core/src/ec/finalize.rs
```diff
@@ -97,13 +126,144 @@ pub fn ec_finalize_response(
...
+fn recover_orphaned_ec(
+    settings: &Settings,
+    ec_context: &mut EcContext,
+    graph: &KvIdentityGraph,
+    updates: &[super::kv::PartnerIdUpdate],
+    response: &mut Response<EdgeBody>,
+) {
+    // Snapshot the orphaned ID once so every fail-closed exit binds the failed
+    // snapshot to the same key.
+    let orphan_id = ec_context.ec_value().unwrap_or_default().to_owned();
```
aram356: ⛏ **nitpick** — `unwrap_or_default()` can’t fire on the current call path (recovery is only reachable with an active EC value), but if it ever did, every fail-closed exit below would store `EcKvSnapshot::Failed { ec_id: "" }`, which `belongs_to`/`entry_for` then silently never match. A `let Some(orphan_id) = ec_context.ec_value().map(str::to_owned) else { return; }` (or an `expect("should have an active EC during orphan recovery")` per house style) would make the invariant explicit.

crates/trusted-server-core/src/platform/http.rs
```diff
@@ -301,6 +331,17 @@ pub trait PlatformHttpClient: Send + Sync {
...
     }
 
+    /// Whether a request started by [`send_async`](Self::send_async) can be
+    /// completed directly by [`wait`](Self::wait) while preserving a streamed
+    /// response body.
+    ///
+    /// Supporting adapters must override `wait` with direct single-handle
+    /// completion. A pending request marked for streaming must never be passed
+    /// to [`select`](Self::select), whose fan-out contract remains buffered.
+    fn supports_pending_streaming_responses(&self) -> bool {
```
aram356, line 341: 🌱 **seedling** — The default `wait` routes through `select`, and both implemented `select`s now reject stream-marked handles — so an adapter that returns `true` here without overriding `wait` fails at runtime on every publisher navigation. The doc comment states the requirement, but a future adapter author could still miss it. Not for this PR: consider a shared conformance test (or a `debug_assert`) tying this capability to an overridden `wait`.

aram356 (approved):

docs/superpowers/specs/2026-07-10-kv-eid-request-snapshot-ec-recovery-design.md
```diff
@@ -0,0 +1,268 @@
...
+
+Valid partner responses are collected across every HTTP concurrency batch into
+one request-wide set of `PartnerIdUpdate` values and merged in one final bulk
+CAS operation. Draining a network batch never writes KV. When the finalized snapshot still
+has a usable generation, the uncontended case performs one write and no
+additional read. When finalization already wrote the row, pull sync uses its
+updated entry for eligibility, collects responses, then performs one refresh
+lookup to obtain the new generation before its bulk CAS. A conflict rereads the
+latest entry, rejects a tombstone, re-merges every collected update, and retries
+within the existing bound. Pull sync never creates a missing root.
```
aram356: 📝 **note** — This paragraph predates the consent-revalidation fix (`af56a10f`, added for the round-2 “post-send pull sync can disclose an ID after withdrawal” finding): the shipped code always re-reads the live row before the first partner request and cancels on anything short of a live, consenting entry, so “the uncontended case performs one write and no additional read” no longer holds — every dispatching request pays exactly one revalidation read, and the bulk CAS writes from that refreshed snapshot rather than the finalized one. Since this spec ships in this PR and reads as authoritative, the stale sentence misstates exactly the KV-read accounting the PR is about.

```suggestion
Valid partner responses are collected across every HTTP concurrency batch into
one request-wide set of `PartnerIdUpdate` values and merged in one final bulk
CAS operation. Draining a network batch never writes KV. Because pull sync runs
after the response is flushed and discloses the raw EC ID to partners, the live
row is re-read once before the first partner request; anything short of a live,
consenting entry cancels the dispatch, and that refreshed snapshot supplies the
generation for the bulk CAS. A conflict rereads the latest entry, rejects a
tombstone, re-merges every collected update, and retries within the existing
bound. Pull sync never creates a missing root.
```

_(format-verified: `prettier --check` passes with this applied)_

aram356 (approved):

PR #887, opened 2026-07-10
Fix Prebid User ID diagnostics

### LINKED ISSUE #886: Fix Prebid User ID bundle diagnostics
## Description

Publisher-specific external Prebid bundles compile an explicit set of User ID submodules, but runtime diagnostics currently compare publisher configuration against the default preset instead of the modules actually included in the generated bundle. Diagnostics also run before late publisher `userSync.userIds` configuration is available.

As a result, a missing module can reduce `pbjs.getUserIdsAsEids()` output while diagnostics incorrectly claim that the module is present. The supported module registry also lacks PAIR ID and lockr AIM entries needed by publisher configurations.

## Reproduction steps

1. Generate an external Prebid bundle with a custom `integrations.prebid.bundle.user_id_modules` list.
2. Load a publisher page that configures additional User ID modules after the bundle initializes.
3. Inspect `window.__tsjs_prebid_diagnostics.userIdModules` during an auction.
4. Observe that `includedModules` reflects the default preset and late configured names are absent.

## Expected behavior

- Generated bundles expose their exact selected User ID module names.
- Runtime diagnostics compare publisher configuration against that exact selection.
- Diagnostics refresh when `requestBids()` observes late configuration.
- Missing-module warnings are emitted once rather than on every auction.
- PAIR ID and lockr AIM are valid explicit bundle selections.

## Affected area

- Integrations (Prebid)
- JS build pipeline

## Version

Prebid.js 10.26.0 external bundle pipeline.

## Logs

No crash is produced. The symptom is incomplete EID output combined with inaccurate runtime diagnostics.

## Acceptance criteria

- [ ] The generated User ID import module exports the exact selected module list.
- [ ] Runtime diagnostics use the generated selection instead of the default preset.
- [ ] Diagnostics capture publisher configuration available by `requestBids()` time.
- [ ] Repeated auctions do not repeat identical missing-module warnings.
- [ ] `pairIdSystem` and `lockrAIMIdSystem` are supported registry entries.
- [ ] Unit tests cover generated metadata, late configuration, warning deduplication, and registry mappings.

ChristianPavilonis:
## Summary

- Report the exact User ID modules compiled into each publisher-specific Prebid bundle.
- Refresh diagnostics when auctions begin so late publisher configuration is visible without repeating warnings.
- Add PAIR ID and lockr AIM to the supported User ID module registry.
- Preserve vendor-specific OpenRTB EID `atype` values end-to-end, including PAIR's required `571187`.
- Exercise the production external-bundle wiring in an isolated smoke test.

## Why `atype` changed from `u8` to `i32`

OpenRTB EID agent types are not limited to the standard single-byte values. The canonical Prebid PAIR EID uses `source: "google.com"` and `atype: 571187`, which cannot fit in a Rust `u8` (maximum `255`). The previous JS and Rust sanitizers therefore silently removed PAIR's `atype` before the EID reached the downstream Prebid request.

`Uid.atype` and the EC partner configuration/runtime path now use `i32`, matching the generated OpenRTB model and preserving vendor-specific values such as `571187`. Browser input, auction payloads, cookies, configured EC partners, KV-resolved EIDs, and downstream serialization all enforce a non-negative signed 32-bit range; negative or oversized values are rejected or omitted rather than propagated.

## Changes

| Area | Change |
| ---- | ------ |
| External Prebid build | Exports the exact selected User ID module list from generated bundle imports and verifies the production builder/alias wiring with an isolated smoke test. |
| Prebid browser integration | Uses generated module metadata, refreshes diagnostics at `requestBids()`, deduplicates warnings, and preserves non-negative `atype` values through `i32::MAX`. |
| User ID registry | Registers `pairIdSystem` and `lockrAIMIdSystem`. |
| Rust OpenRTB and auction parsing | Widens UID `atype` from `u8` to `i32` and preserves PAIR's `571187`. |
| EC configuration and cookies | Widens configured, structured-cookie, legacy-cookie, and KV-resolved agent types while validating or sanitizing negative values. |
| Regression coverage | Covers JS auction payloads, both cookie formats, EC configuration and KV continuity, Rust model serialization, downstream Prebid serialization, and the real external bundle. |

## Closes

Closes #886

## Test plan

- [x] JS build and tests: `cd crates/trusted-server-js/lib && npm run build && npx vitest run`
- [x] JS lint and format: `cd crates/trusted-server-js/lib && npm run lint && npm run format`
- [ ] Release WASM build: `cargo build --package trusted-server-adapter-fastly --release --target wasm32-wasip1`

## Checklist

- [x] No `unwrap()` added
- [x] Uses the existing JS logging wrapper; no `println!`-style output added
- [x] New code has tests


crates/trusted-server-core/src/ec/prebid_eids.rs
```diff
@@ -33,7 +33,7 @@ struct LegacyCookieEid {
         dead_code,
         reason = "legacy cookie field is deserialized for compatibility but not emitted"
     )]
-    atype: u8,
+    atype: i32,
```
aram356, line 32: ♻️ **refactor** — The `#[allow(dead_code, reason = "legacy cookie field is deserialized for compatibility but not emitted")]` above this field is now stale: `entry.atype` **is** read and emitted into the OpenRTB `Uid` in `legacy_cookie_eids_to_openrtb` (line 342, `atype: (entry.atype >= 0).then_some(entry.atype)`). Since this PR touches both sites, drop the `#[allow(dead_code)]` block entirely — the field is genuinely used, so the allow is unnecessary and its reason contradicts the code.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -139,33 +142,40 @@ function recordUserIdModuleDiagnostics(): PrebidUserIdDiagnostics {
...
+  const previouslyMissingConfiguredUserIdNames = new Set<string>();
   if (typeof window !== 'undefined') {
     const tsjsWindow = window as typeof window & {
       __tsjs_prebid_diagnostics?: { userIdModules?: PrebidUserIdDiagnostics };
     };
+    for (const name of tsjsWindow.__tsjs_prebid_diagnostics?.userIdModules
+      ?.missingConfiguredUserIdNames ?? []) {
+      previouslyMissingConfiguredUserIdNames.add(name);
+    }
     tsjsWindow.__tsjs_prebid_diagnostics = {
```
aram356, line 167: 🤔 **thinking** — `window.__tsjs_prebid_diagnostics.userIdModules` is a single un-namespaced global. If two Prebid bundles with different `INCLUDED_PREBID_USER_ID_MODULES` sets run on the same page, they clobber each other's snapshot and cross-contaminate the warning-dedup set: one bundle's write can suppress or resurface the other's warnings, and `includedModules` reflects whichever ran last. Related: dedup only compares against the immediately-preceding snapshot, so a name that oscillates missing→resolved→missing re-warns on each reappearance. If this ever matters in practice, keying the snapshot by bundle identity (or merging rather than overwriting) would fix it.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -25,16 +25,19 @@ import 'prebid.js/modules/userId.js';
...
-import './_user_ids.generated';
 
 import { log } from '../../core/log';
 import { buildAdRequest, parseAuctionResponse } from '../../core/auction';
 import type { AuctionBid, AuctionEid } from '../../core/auction';
 import type { AuctionSlot } from '../../core/types';
 
-import { DEFAULT_PREBID_USER_ID_MODULES, PREBID_USER_ID_MODULE_REGISTRY } from './user_id_modules';
+import { INCLUDED_PREBID_USER_ID_MODULES } from './_user_ids.generated';
+import { PREBID_USER_ID_MODULE_REGISTRY } from './user_id_modules';
```
aram356, line 35: 🏕 **camp site** — With this import gone, `DEFAULT_PREBID_USER_ID_MODULES` in `user_id_modules.ts` has no external consumers left — its only remaining reference is the internal `MODULE_SORT_ORDER` construction in the same file. The `export` keyword can be dropped (keep the const).

crates/trusted-server-js/lib/test/build-prebid-external.test.mjs
```diff
@@ -18,6 +27,39 @@ describe('build-prebid-external metadata', () => {
     });
   });
 
+  it('renders the exact selected User ID modules for runtime diagnostics', () => {
+    expect(renderIncludedUserIdModulesExport(['liveIntentIdSystem', 'pairIdSystem'])).toBe(
+      'export const INCLUDED_PREBID_USER_ID_MODULES = ["liveIntentIdSystem","pairIdSystem"];'
+    );
+  });
+
+  it('includes generated User ID metadata in the production external bundle', async () => {
```
aram356, line 36: 🤔 **thinking** — This runs a full Vite/Rollup production build of prebid.js inside the vitest suite (120s budget), which dominates the job's wall clock. esbuild also spawns a child process, and `vitest.config.ts` disables threads with a note that child processes are blocked in some local sandboxes — so this test can hang or fail locally where CI passes. The isolation itself is done well (tmpdir output, cleanup in `finally`, `.external-generated-*` gitignored). If local runs become painful, consider moving it to a separate script/job or gating it behind a tag.

crates/trusted-server-js/lib/test/build-prebid-external.test.mjs
```diff
@@ -18,6 +27,39 @@ describe('build-prebid-external metadata', () => {
...
+        outputDirectory,
+      ]);
+
+      const manifest = JSON.parse(
+        fs.readFileSync(path.join(outputDirectory, 'manifest.json'), 'utf8')
+      );
+      const bundle = fs.readFileSync(path.join(outputDirectory, manifest.filename), 'utf8');
+
+      expect(manifest.userIdModules).toEqual(['pairIdSystem', 'lockrAIMIdSystem']);
+      expect(bundle).toContain('["pairIdSystem","lockrAIMIdSystem"]');
```
aram356: ⛏ **nitpick** — `toContain('["pairIdSystem","lockrAIMIdSystem"]')` relies on esbuild preserving that exact array literal in minified output; a bundler upgrade could break the assertion without a real regression. The `manifest.userIdModules` assertion above already covers selection — matching each module name individually would keep the wiring check while being less brittle.

aram356 (approved):

PR #890, opened 2026-07-10
Dump full SSAT prebid responses in ts-debug auction comment

### LINKED ISSUE #889: For SSAT acution write responses from prebid server
### Description

should match response we from /auction endpoint

prk-Jr:
## Summary

- The server-side auction (SSAT) stream path only emitted a summary line in the `ts-debug` HTML comment (`ssp`/`mediator`/`winning`/`time`), so an operator seeing `winning=0` could not tell whether prebid returned nothing, errored, or bid below the floor.
- Dump the full `provider_responses` (and `mediator_response`) into the comment so the SSAT surfaces the same prebid server response detail an operator gets from the `/auction` endpoint — status, every bid, and `metadata` (which carries PBS `ext.errors` / `ext.debug.httpcalls` when prebid `debug=true`).
- Bid creative and provider metadata are attacker/partner-influenced, so the embedded JSON has its HTML-comment terminators neutralized to keep the dump inside the comment and out of the live DOM.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/publisher.rs` | `prepend_auction_debug_comment` now serializes `provider_responses` + `mediator_response` as JSON into the `ts-debug` comment; neutralizes `-->` / `--!>` in the payload; adds a unit test asserting status is surfaced and terminators are neutralized (exactly one closing `-->`). |

## Closes

Closes #889

## Test plan

- [x] `cargo test-fastly` (new test `auction_debug_comment_dumps_provider_status_and_neutralises_terminators` passes; `cargo check-fastly` clean)
- [ ] `cargo test-axum` — deferred to CI
- [ ] `cargo clippy-fastly && cargo clippy-axum` — deferred to CI
- [ ] JS tests / format — n/a (no JS change)
- [ ] Docs format — n/a (no docs change)

## Notes

- The dump is gated behind the existing `[debug].auction_html_comment` flag — off by default.
- On a winning auction the dump includes `Bid.creative` (full ad markup) inline and can be large; intended for debugging, not steady-state production.

## Checklist

- [x] New code has tests


crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -2124,14 +2124,23 @@ impl AuctionProvider for PrebidAuctionProvider {
...
+            // Surface the HTTP status and a body snippet on the response metadata
+            // so the ts-debug auction dump shows *why* prebid errored (e.g. a 4xx
+            // from a PBS that rejects the unsigned server-side request) without
+            // needing log access. A bare `AuctionResponse::error` yields empty
+            // metadata, which is indistinguishable from other failures in the dump.
+            let body_snippet = body_preview[..body_preview.floor_char_boundary(512)].to_string();
+            return Ok(AuctionResponse::error("prebid", response_time_ms)
+                .with_metadata("error_type", serde_json::json!("http_status"))
+                .with_metadata("status", serde_json::json!(status.as_u16()))
+                .with_metadata("body", serde_json::json!(body_snippet)));
```
aram356: 🔧 **wrench** — The upstream PBS response body is echoed into the **public `POST /auction` response**, which this codebase explicitly forbids.

`AuctionResponse.metadata` is cloned verbatim by `ProviderSummary::from` (`auction/types.rs:252`) into `ext.orchestrator.provider_details[].metadata` of the OpenRTB response (`auction/formats.rs:305`). There is **no debug gate** on that path — it is not covered by `[debug].auction_html_comment`.

`auction/orchestrator.rs:100` states the invariant this breaks:

> `// SECURITY:` … Providers MUST NOT interpolate upstream-controlled content (response bodies, parse errors, headers) into their `TrustedServerError::*.message` fields. Use static text and log details server-side with `log::warn!` instead.

Verified by serializing a `ProviderSummary` built from this response:

```json
{"name":"prebid","status":"error","bid_count":0,"bidders":[],"time_ms":12,
 "metadata":{"error_type":"http_status","status":403,
              "body":"PBS internal: upstream-secret-detail"}}
```

Any caller of `POST /auction` now gets PBS internal error text (hostnames, stack traces, auth hints) reflected back.

**Fix** — keep `status` (a `u16`, safe and genuinely useful); drop `body` and log it instead. The existing body snippet is only emitted at `trace` level, so raising it to `warn` gives the operator the same detail without publishing it:

```rust
log::warn!(
    "Prebid returned non-success status {status}: {}",
    &body_preview[..body_preview.floor_char_boundary(512)]
);
return Ok(AuctionResponse::error("prebid", response_time_ms)
    .with_metadata("error_type", serde_json::json!(ERROR_TYPE_HTTP_STATUS))
    .with_metadata("status", serde_json::json!(status.as_u16())));
```

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -2124,14 +2124,23 @@ impl AuctionProvider for PrebidAuctionProvider {
...
             }
-            return Ok(AuctionResponse::error("prebid", response_time_ms));
+            // Surface the HTTP status and a body snippet on the response metadata
+            // so the ts-debug auction dump shows *why* prebid errored (e.g. a 4xx
+            // from a PBS that rejects the unsigned server-side request) without
+            // needing log access. A bare `AuctionResponse::error` yields empty
+            // metadata, which is indistinguishable from other failures in the dump.
+            let body_snippet = body_preview[..body_preview.floor_char_boundary(512)].to_string();
+            return Ok(AuctionResponse::error("prebid", response_time_ms)
+                .with_metadata("error_type", serde_json::json!("http_status"))
```
aram356: 🔧 **wrench** — `"http_status"` is not a recognised `error_type`, so PBS HTTP errors are mislabelled as transport errors in telemetry.

The legal values are the four consts in `auction/orchestrator.rs:95-98` (`parse_response`, `launch_failed`, `transport`, `timeout`). `provider_status()` in `auction/telemetry.rs:795` matches on exactly those and falls through:

```rust
_ => "transport_error",
```

So every PBS 4xx/5xx lands in the same telemetry bucket as a genuine connection failure — which defeats this PR's own goal ("an operator seeing `winning=0` could not tell whether prebid returned nothing, errored, or bid below the floor") on the telemetry path.

**Fix** — register the new type rather than using a bare literal:

```rust
// auction/orchestrator.rs — make these pub(crate) and add:
pub(crate) const ERROR_TYPE_HTTP_STATUS: &str = "http_status";

// auction/telemetry.rs — provider_status():
Some("http_status") => "http_status_error",
```

crates/trusted-server-core/src/publisher.rs
```diff
@@ -871,8 +871,28 @@ pub(crate) fn prepend_auction_debug_comment(
...
+    // entry, and `metadata` (which carries PBS `ext.errors` / `ext.debug.httpcalls`
+    // when prebid `debug=true`) — without needing log access.
+    //
+    // `Bid.creative` and provider metadata are attacker/partner-influenced and
+    // may contain `-->` (or the `--!>` variant), which would terminate the HTML
+    // comment early and leak the remaining markup into the live DOM. Neutralise
+    // both terminators before embedding so the dump stays inside the comment.
+    let neutralise_comment_terminators =
+        |json: String| -> String { json.replace("-->", "-- >").replace("--!>", "-- !>") };
+    let providers_dump = serde_json::to_string_pretty(&result.provider_responses)
```
aram356: 🤔 **thinking** — Unbounded dump size on every page render.

`provider_responses` carries `Bid.creative` for **every** bid, including losing ones, and `to_string_pretty` roughly doubles it with indentation. A 20-bid PBS response with full ad markup can add megabytes to each HTML response on the streaming path — and this is prepended to every page while the flag is on.

The PR notes this in the description, but it is cheap to bound: use `to_string` instead of `to_string_pretty`, and either omit `creative` from the dump or cap the serialized payload at a fixed byte budget (with a `…truncated` marker) so an operator can't accidentally take a site down by leaving the flag on.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -871,8 +871,28 @@ pub(crate) fn prepend_auction_debug_comment(
...
+    // each SSP returned — `status` (nobid vs error vs success), every `bids`
+    // entry, and `metadata` (which carries PBS `ext.errors` / `ext.debug.httpcalls`
+    // when prebid `debug=true`) — without needing log access.
+    //
+    // `Bid.creative` and provider metadata are attacker/partner-influenced and
+    // may contain `-->` (or the `--!>` variant), which would terminate the HTML
+    // comment early and leak the remaining markup into the live DOM. Neutralise
+    // both terminators before embedding so the dump stays inside the comment.
+    let neutralise_comment_terminators =
+        |json: String| -> String { json.replace("-->", "-- >").replace("--!>", "-- !>") };
```
aram356: ♻️ **refactor** — One escaping rule instead of a two-entry blacklist.

To confirm this is airtight I fuzzed 11 comment-escape vectors through a real HTML5 parser (jsdom), including the nested-comment paths that reach `comment end state` without a literal `-->`: `<!-->`, `<!--!>`, `<!--->`, `<!---->`, `---!>`, `----!>`, `--!-->`. **All stayed contained** — the implementation is correct.

But establishing that requires walking the HTML5 tokenizer through the `comment less-than sign bang dash dash` → `comment end` → `comment end bang` states. A single rule is equivalent and trivially auditable, since *every* comment-end path requires a literal `--`:

```rust
let neutralise_comment_terminators = |json: String| -> String { json.replace("--", "- -") };
```

Bonus: it also keeps the output well-formed if the page is ever served as `application/xhtml+xml`, where XML forbids `--` inside a comment outright (today a creative containing `--` would be a fatal parse error → blank page).

crates/trusted-server-core/src/publisher.rs
```diff
@@ -2439,6 +2459,76 @@ mod tests {
...
+            "should surface the no-bid provider status: {comment}"
+        );
+        assert!(
+            comment.contains("provider_responses="),
+            "should dump the provider_responses payload"
+        );
+        // The creative's `-->` must be neutralised so the only comment terminator
+        // is the trailing one — otherwise embedded markup would leak into the DOM.
+        assert_eq!(
+            comment.matches("-->").count(),
```
aram356: ♻️ **refactor** — Test coverage gap: only the `-->` variant is pinned.

The code also handles `--!>`, and the HTML5 nested-comment paths (`<!-->`, `<!--!>`, `<!--->`) reach comment-end state without containing a bare `-->`. All of these are handled correctly today, but nothing in the suite would catch a regression if the `--!>` replace were dropped.

Suggest a table-driven case over the vectors that actually exercise each tokenizer path:

```rust
for creative in [
    "<div>evil-->break</div>",
    "--!><img src=x onerror=alert(1)>",
    "<!--><img src=x onerror=alert(1)>",
    "<!--!><img src=x onerror=alert(1)>",
    "----!><img src=x onerror=alert(1)>",
] { /* assert exactly one `-->` and no `--!>` survives */ }
```

crates/trusted-server-core/src/publisher.rs
```diff
@@ -871,8 +871,28 @@ pub(crate) fn prepend_auction_debug_comment(
...
+    // `Bid.creative` and provider metadata are attacker/partner-influenced and
+    // may contain `-->` (or the `--!>` variant), which would terminate the HTML
+    // comment early and leak the remaining markup into the live DOM. Neutralise
+    // both terminators before embedding so the dump stays inside the comment.
+    let neutralise_comment_terminators =
+        |json: String| -> String { json.replace("-->", "-- >").replace("--!>", "-- !>") };
+    let providers_dump = serde_json::to_string_pretty(&result.provider_responses)
+        .map(neutralise_comment_terminators)
+        .unwrap_or_else(|e| format!("<provider_responses serialize error: {e}>"));
+    let mediator_dump = serde_json::to_string_pretty(&result.mediator_response)
```
aram356: ⛏ **nitpick** — When there is no mediator this dumps `mediator_response=null`, which the summary line already says as `mediator=none`. Skipping the dump when `result.mediator_response.is_none()` keeps the comment tighter.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -2439,6 +2459,76 @@ mod tests {
     use super::*;
     use crate::auction::types::{AdFormat, AdSlot, MediaType};
     use crate::integrations::IntegrationRegistry;
+
+    #[test]
+    fn auction_debug_comment_dumps_provider_status_and_neutralises_terminators() {
+        use crate::auction::orchestrator::OrchestrationResult;
```
aram356: ♻️ **refactor** — CLAUDE.md: *"No local imports within functions or blocks"* (`use super::*` is the only exception, and only in `#[cfg(test)]` modules). Move `OrchestrationResult` / `AuctionResponse` up to the test module's import block.

Relatedly, `make_test_bid_with_creative` (line 2512) is inserted *between* the module's `use` lines, splitting the import block in two — please move the helper below the imports.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -2341,6 +2350,44 @@ mod tests {
         );
     }
 
+    #[test]
+    fn parse_response_attaches_status_and_body_metadata_on_http_error() {
+        use crate::auction::types::BidStatus;
```
aram356: ♻️ **refactor** — Same as in `publisher.rs`: CLAUDE.md forbids local imports inside function bodies. Hoist `BidStatus` into the test module's import block.

aram356 (changes requested):
## Summary

The HTML-comment side of this PR is sound — I fuzzed the terminator neutralisation against a real HTML5 parser (11 escape vectors, including the `<!-->` / `<!--!>` nested-comment paths) and nothing escaped the comment. The problem is the other half: the new prebid error metadata is **not** confined to the debug comment. It flows into the public `POST /auction` response, which breaks an invariant the codebase documents explicitly.

## Blocking

### 🔧 wrench

- **Upstream PBS error body is echoed into the public `/auction` response** (`integrations/prebid.rs:2143`): `.with_metadata("body", …)` lands in `ProviderSummary.metadata` (`auction/types.rs:252`) → `ext.orchestrator.provider_details[]` (`auction/formats.rs:305`), ungated by any debug flag. `auction/orchestrator.rs:100` carries a `// SECURITY:` comment forbidding exactly this: *"Providers MUST NOT interpolate upstream-controlled content (response bodies, parse errors, headers)… Use static text and log details server-side with `log::warn!` instead."* Verified by serializing the resulting summary — the body snippet appears on the wire. See inline comment for the fix.
- **`error_type: "http_status"` is mislabelled as a transport error in telemetry** (`integrations/prebid.rs:2141`): it isn't one of the four `ERROR_TYPE_*` consts (`auction/orchestrator.rs:95-98`), so `provider_status()` (`auction/telemetry.rs:795`) hits its `_ => "transport_error"` fallback. Every PBS 4xx/5xx is now bucketed with genuine connection failures — which defeats this PR's own goal on the telemetry path.
- **`[debug].auction_html_comment` doc no longer describes what the flag does** (`settings.rs:1904`): it still promises *"auction pipeline stats (SSP count, mediator status, winning bid count)"*, but the comment now embeds every bid's full raw SSP creative markup. Operators make the enable/disable call from that sentence. The sibling `adm_in_bids` flag already carries the right warning (*"Never enable in production — injects raw HTML from SSPs"*) — please mirror it here, and update the `prepend_auction_debug_comment` doc at `publisher.rs:858`, which still says "summarising".

## Non-blocking

### 🤔 thinking

- **Unbounded dump size** (`publisher.rs:885`): every bid's `creative` — including losers — pretty-printed into every page render. See inline.

### ♻️ refactor

- **A single `replace("--", "- -")` would replace the two-terminator blacklist** (`publisher.rs:884`) — equivalent, trivially auditable, and XHTML-safe. Inline comment has the fuzz results.
- **Test coverage gap**: `--!>` and the nested-comment paths are handled but unpinned (`publisher.rs:2502`).
- **Local imports inside test fn bodies** violate CLAUDE.md (`publisher.rs:2465`, `integrations/prebid.rs:2355`).

### ⛏ nitpick

- `mediator_response=null` is dumped when there is no mediator, duplicating `mediator=none` (`publisher.rs:888`).

### 🌱 seedling

- `path_label` only ever receives `"stream"` (`publisher.rs:1265`), though the doc says it differentiates `stream` from `buffered`. The buffered path produces no dump at all, so an operator debugging that path gets nothing from this feature. Pre-existing, not introduced here — worth a follow-up.

## CI Status

All 19 checks pass on `08de9c4` (fmt, clippy, cargo test across fastly/axum/cloudflare/spin/parity/CLI, vitest, docs format, integration + browser tests, CodeQL).

crates/trusted-server-core/src/auction/telemetry.rs
```diff
@@ -800,6 +800,7 @@ fn provider_status(response: &AuctionResponse) -> &'static str {
             Some("parse_response") => "parse_error",
             Some("transport") => "transport_error",
             Some("timeout") => "timeout",
+            Some("http_status") => "http_status_error",
```
ChristianPavilonis, line 803: 🔧 **P1 / High — PBS HTTP failures are excluded from provider-health error metrics**

The new `http_status_error` value is emitted here, but `tinybird/pipes/auction_provider_stats_mv.pipe:15` only counts `launch_error`, `parse_error`, and `transport_error` as errors. These requests therefore increase `requests` without increasing `errors`, `timeouts`, or another outcome counter, and `provider_health.error_rate` under-reports PBS 4xx/5xx failures. The documented status contract at `docs/superpowers/specs/2026-06-22-auction-prebid-metrics-tinybird-grafana-design.md:347` is also stale.

**Suggested fix:** Add `http_status_error` to the Tinybird rollup's error set, update the documented status list, and add coverage for both the Rust mapping and downstream aggregation classification.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -871,8 +882,48 @@ pub(crate) fn prepend_auction_debug_comment(
...
+    // comment-end-bang variant), which would end the comment early and leak the
+    // remaining markup into the live DOM. Neutralise both, then cap the size so
+    // a large SSP response cannot bloat the page.
+    //
+    // NB: a single `replace("--", …)` is deliberately NOT used — because
+    // `str::replace` is non-overlapping, it re-forms a live `-->` / `--!>` at
+    // the junction of an odd dash-run (`--->` -> `- -->`, `----->` -> `- -- -->`),
+    // reintroducing exactly the terminator we are trying to remove. The two
+    // targeted replacements below cannot re-form either sequence.
+    let render_dump = |json: String| -> String {
```
ChristianPavilonis, line 1004: 🤔 **P2 / Medium — The 256 KiB cap does not bound serialization work or transient allocations**

`serde_json::to_string` materializes the complete provider response before this closure runs, and the chained `replace` calls then allocate and scan full-size strings before truncation. Provider responses can each reach 2 MiB, and mediator serialization repeats the process separately. On Wasm edge runtimes this can create substantial transient heap pressure and latency despite the final output cap; with a mediator, the final comment can also exceed 512 KiB because each section receives its own budget.

**Suggested fix:** Serialize through a bounded writer, apply one explicit total-comment budget, then neutralize the bounded buffer. Add an oversized multibyte provider-plus-mediator test verifying bounded output, valid UTF-8, the truncation marker, and comment containment.

ChristianPavilonis (approved):
## Summary

Approved after reviewing the current head against `main`. The HTML-comment containment and public-response protections appear correct. I noted two follow-up findings inline: one telemetry aggregation gap and one edge-runtime resource concern.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -855,12 +855,23 @@ pub(crate) fn write_bids_to_state(
...
+/// never enable in production.
 ///
 /// `path_label` differentiates the streaming-with-auction-hold path (`stream`)
-/// from the buffered path (`buffered`) in the resulting `<!-- ts-debug: -->`
-/// marker so on-page debugging can tell which code path produced the bids.
+/// from the buffered path (`buffered`) in the marker so on-page debugging can
+/// tell which code path produced the bids.
 pub(crate) fn prepend_auction_debug_comment(
     path_label: &str,
     result: &crate::auction::orchestrator::OrchestrationResult,
```
aram356, line 952: 🔧 **wrench** — This comment advertises `ext.debug.httpcalls` as the feature, and that is exactly the problem: it renders the **visitor's identity graph into the publisher's page source**.

`integrations/prebid.rs:1831-1836` copies PBS's entire `ext.debug` subtree into `AuctionResponse.metadata` verbatim when `[integration.prebid].debug = true`:

```rust
if self.config.debug {
    if let Some(debug) = ext.and_then(|e| e.get("debug")) {
        auction_response.metadata.insert("debug".to_string(), debug.clone());
```

That subtree is `resolvedrequest` (the full OpenRTB request Trusted Server sent) and `httpcalls` (the outbound body sent to each bidder). That request is built at `integrations/prebid.rs:1514-1600` and carries:

| Field | Source | Content |
| --- | --- | --- |
| `user.id` | `prebid.rs:1515` | **EC ID** |
| `user.ext.eids` | `prebid.rs:1528` | **the consent-gated identity graph** |
| `user.consent` | `prebid.rs:1516-1519` | **raw TCF TC string** |
| `device.ip` | `prebid.rs:1583` | **real client IP** |
| `device.geo.lat/lon` | `prebid.rs:1584-1600` | **precise geo** |

Before this PR the comment held four counters. Now it serializes `metadata`, so all of the above lands in a DOM comment node — readable by **every third-party tag already running on the page** via `document.documentElement.innerHTML`, and present in view-source, scrapers, and archives.

This inverts the purpose of server-side EID resolution: the identity graph is resolved and consent-gated on the edge specifically so it does *not* reach the client script environment, and this hands it straight back. Note the codebase already redacts in the equivalent situation — `auction/telemetry.rs:115-116` ships `consent_present: bool`, never the raw TC string.

I verified this rather than inferring it; a runnable reproduction is in the review body — it passes on this branch.

**Fix** — whitelist keys instead of dumping the map. Keep `error_type` / `status` / `responsetimemillis` / `errors` / `warnings` / `bidstatus`; exclude `debug` entirely, or scrub `resolvedrequest` and `httpcalls[].requestbody` before embedding.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -871,8 +882,48 @@ pub(crate) fn prepend_auction_debug_comment(
...
         None => "none".to_string(),
     };
+    // Full per-provider dump so the operator can see exactly what each SSP
+    // returned — `status` (nobid vs error vs success), every `bids` entry, and
+    // `metadata` — without needing log access.
+    //
+    // `Bid.creative` and provider metadata are attacker/partner-influenced and
+    // may contain an HTML5 comment terminator (`-->`, or the `--!>`
+    // comment-end-bang variant), which would end the comment early and leak the
+    // remaining markup into the live DOM. Neutralise both, then cap the size so
```
aram356: 🔧 **wrench** — The unbounded dump can turn a working page into a **500**.

The whole comment is emitted in a single `end_tag.before(&bids_script, ContentType::Html)` (`html_processor.rs:365`), and that write lands in a `BoundedWriter`:

```rust
// publisher.rs:507
let mut output = BoundedWriter::new(settings.publisher.max_buffered_body_bytes);

// publisher.rs:574-583
fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
    if self.inner.len() + buf.len() > self.limit {
        return Err(std::io::Error::other("publisher body exceeded maximum buffered size"));
```

Default limit is 16 MiB (`settings.rs:71-73`). The error propagates through `write_processed_chunk` as `TrustedServerError::Proxy` → 5xx. So enabling this flag can **break the very page it was enabled to debug** — and it fails at the `</body>` hold, after the page has already begun streaming.

The size is realistic, not theoretical: `UPSTREAM_RTB_MAX_RESPONSE_BYTES` allows **2 MiB per provider** (`integrations/mod.rs:163`), `Bid.creative` is the raw `adm` retained for *every* bid from *every* provider (losers included), `to_string_pretty` adds indentation and escapes every `"` in that HTML, and with prebid debug on, `metadata.debug.httpcalls` carries the full request **and** response body for every bidder. The two chained `String::replace` calls then keep ~3 full copies of the payload live simultaneously, against a Wasm heap this repo's own design notes put at ~16 MB with a 3-4 MB baseline.

Every other partner-influenced payload in this codebase is bounded — `MAX_AUCTION_BODY_SIZE` (256 KiB, `auction/endpoints.rs:42`), `MAX_CREATIVE_SIZE` (1 MiB, `creative.rs:310`), `SIGN_MAX_BODY_BYTES` (64 KiB, `proxy.rs:38`) — and this PR itself caps the error snippet at 512 bytes. This dump is the one thing that isn't capped.

**Fix** — truncate `Bid.creative` per bid and cap the total serialized dump (256 KiB, matching `MAX_AUCTION_BODY_SIZE`) with a `<truncated N bytes>` marker.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -2341,6 +2352,50 @@ mod tests {
...
+        );
+        assert_eq!(
+            result.metadata["error_type"],
+            json!("http_status"),
+            "should tag the error path so telemetry buckets it as an http status error"
+        );
+        assert_eq!(
+            result.metadata["status"],
+            json!(403),
+            "should surface the upstream HTTP status code"
```
aram356, line 2390: 🔧 **wrench** — Heads-up: this assertion **pins the security bug in place**.

It asserts that the raw upstream response body is retained in `metadata` — and `metadata` is cloned verbatim into `ext.orchestrator.provider_details[].metadata` of the public `POST /auction` response (see my comment on line 2143). So fixing that exposure requires deleting or inverting this assertion.

Calling it out so the change doesn't read like a regression when you make it. The replacement test worth having is the inverse:

```rust
#[test]
fn parse_response_omits_upstream_body_from_public_metadata() {
    // … 403 with a body …
    assert!(
        !result.metadata.contains_key("body"),
        "upstream response body must not reach ProviderSummary.metadata — \
         it is serialized into the public /auction response"
    );
}
```

Worth noting the surrounding module already has the gating pattern this path should have followed: `enrich_response_metadata_includes_debug_fields_when_enabled` (`prebid.rs:5877`) tests exactly that shape for the `config.debug`-gated keys.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -2123,15 +2123,25 @@ impl AuctionProvider for PrebidAuctionProvider {
...
+            // `AuctionResponse.metadata` (cloned verbatim into
+            // `ext.orchestrator.provider_details[].metadata`). Log the snippet
+            // server-side and surface only the numeric HTTP status — enough for an
+            // operator to tell an error from a no-bid without publishing the body.
+            log::warn!(
+                "Prebid returned non-success status {status}: {}",
+                &body_preview[..body_preview.floor_char_boundary(512)]
+            );
+            return Ok(AuctionResponse::error("prebid", response_time_ms)
+                .with_metadata(
```
aram356, line 2139: ♻️ **refactor** — This is the only provider error response in the codebase that omits `message`.

All four orchestrator error constructors set it (`auction/orchestrator.rs:123, 129, 141, 147`), so every consumer of `provider_details[].metadata` can currently rely on `error_type` + `message` being present. This path breaks that shape.

Adding a **static** message (per the `SECURITY:` note at `auction/orchestrator.rs:100-105` — static text only, no upstream content) restores parity and gives the operator a human-readable line without the leak:

```rust
.with_metadata("message", serde_json::json!("Prebid returned non-success status"))
```

crates/trusted-server-core/src/publisher.rs
```diff
@@ -871,8 +882,48 @@ pub(crate) fn prepend_auction_debug_comment(
         Some(r) => format!("ok({}_bids)", r.bids.len()),
         None => "none".to_string(),
     };
+    // Full per-provider dump so the operator can see exactly what each SSP
+    // returned — `status` (nobid vs error vs success), every `bids` entry, and
+    // `metadata` — without needing log access.
```
aram356: ⛏ **nitpick** — `.map(neutralise_comment_terminators)` applies only to the `Ok` branch, so the `unwrap_or_else` fallback string on the next line bypasses the neutraliser entirely.

Unreachable in practice (`serde_json` won't fail on these types), so this is not a live bug — but it's an un-neutralised seam in a path whose entire safety property is "nothing reaches the DOM un-neutralised." Applying the neutraliser after the fallback instead of inside the `Ok` branch removes the seam and the need to reason about it:

```rust
let providers_dump = neutralise_comment_terminators(
    serde_json::to_string_pretty(&result.provider_responses)
        .unwrap_or_else(|e| format!("<provider_responses serialize error: {e}>")),
);
```

aram356 (changes requested):
## Summary

Second, deeper pass across the surfaces the first review didn't reach (privacy/PII content of the dump, every egress channel, runtime/memory/caching, and the test surface). It turned up two problems that outrank everything in my previous review.

The short version: with prebid `debug = true`, this dump renders the **visitor's identity graph — EC ID, EIDs, TC consent string, client IP, and geo lat/lon — into the publisher's page source**, where any third-party tag on the page can read it. And because the dump is unbounded, it can push the page over the buffered-body cap and turn a working page into a 500.

## Blocking

### 🔧 wrench

- **The dump renders the visitor's identity graph into the page HTML** (`publisher.rs:874-890`). `AuctionResponse.metadata` carries PBS's entire `ext.debug` subtree, copied verbatim at `integrations/prebid.rs:1831-1836` when `[integration.prebid].debug = true` — that is `resolvedrequest` (the full OpenRTB request TS sent) and `httpcalls` (the outbound body sent to each bidder). That request is built at `integrations/prebid.rs:1514-1600` and contains `user.id` (EC ID), `user.ext.eids`, the raw TC string, `device.ip`, and `device.geo.lat/lon`. Serializing `metadata` into a DOM comment node hands the consent-gated identity graph straight back to the client-side script environment that server-side EID resolution exists to keep it out of. Reproduction below. See inline.
- **The unbounded dump can convert a working page into a 500** (`publisher.rs:892`). Output is written through a `BoundedWriter` capped at `publisher.max_buffered_body_bytes` (default 16 MiB, `settings.rs:71-73`); `write()` returns `Err` past the cap (`publisher.rs:574-583`) and that surfaces as `TrustedServerError::Proxy` → 5xx. The whole comment goes out in one `end_tag.before()` write. With `UPSTREAM_RTB_MAX_RESPONSE_BYTES` at 2 MiB *per provider* (`integrations/mod.rs:163`) and `httpcalls` carrying full per-bidder request/response bodies, multi-MB dumps are realistic — so the debug flag can break the page it was enabled to debug. Every other partner-influenced payload here is bounded (`MAX_AUCTION_BODY_SIZE` 256 KiB, `MAX_CREATIVE_SIZE` 1 MiB — and this PR's own 512-byte snippet). This dump is the one that isn't. See inline.
- **The new prebid test asserts the leak as intended behaviour** (`integrations/prebid.rs:2383-2389`). Fixing the `/auction` body exposure from my previous review means this assertion has to be deleted or inverted — flagging so it isn't mistaken for a regression. See inline.

## Non-blocking

### ♻️ refactor

- **The new error response omits `message`** (`integrations/prebid.rs:2140`), the one key every other error path sets (`auction/orchestrator.rs:123, 129, 141, 147`). The prebid HTTP-error shape now diverges from every other provider error. See inline.
- **`--!>` neutralisation is untested.** Delete `.replace("--!>", "-- !>")` from `publisher.rs:884` and the suite stays green — the fixture only plants `-->` in `Bid.creative`. Half the claimed protection is unguarded. Also worth planting a terminator in `metadata` (which the code comment at `publisher.rs:879-882` explicitly calls out as attacker-influenced, but the test never exercises) and using a `Some(_)` mediator so the second `neutralise` call at `publisher.rs:888-890` is covered at all.

### ⛏ nitpick

- **The serialize-error fallback bypasses the neutraliser** (`publisher.rs:887, 890`) — `.map()` applies only to the `Ok` branch. Unreachable in practice, but it's a latent seam in a security-relevant path. See inline.
- **Non-deterministic output.** `AuctionResponse.metadata` and `winning_bids` are `HashMap`s, so key order varies per render and two page sources of the same auction won't diff cleanly.

### 📌 out of scope

- **`auction_html_comment` is undocumented.** Zero hits across `docs/`, and it is absent from the `[debug]` block in `trusted-server.example.toml` (which lists only `ja4_endpoint_enabled`). There is no operator-facing warning anywhere for a flag that now publishes creatives and, with prebid debug on, user identifiers.

## Reproduction — identity graph in page source

Drop into `publisher.rs`'s test module and run `cargo test -p trusted-server-core --target wasm32-wasip1 --lib -- probe_identity_graph --nocapture`. It passes on this branch — every needle is present in the injected HTML:

```rust
#[test]
fn probe_identity_graph_lands_in_page_html() {
    // Shape of PBS `ext.debug` as prebid.rs:1831-1836 stores it verbatim.
    let prebid = crate::auction::types::AuctionResponse::error("prebid", 12).with_metadata(
        "debug",
        serde_json::json!({
            "resolvedrequest": {
                "user": {
                    "id": "EC-ID-abc123",
                    "consent": "CPtc-TCSTRING-xyz",
                    "ext": { "eids": [{ "source": "example.com",
                                        "uids": [{ "id": "EID-USER-999" }] }] }
                },
                "device": { "ip": "203.0.113.77",
                            "geo": { "lat": 37.7749, "lon": -122.4194 } }
            }
        }),
    );
    let result = crate::auction::orchestrator::OrchestrationResult {
        provider_responses: vec![prebid],
        mediator_response: None,
        winning_bids: std::collections::HashMap::new(),
        total_time_ms: 12,
        metadata: std::collections::HashMap::new(),
    };
    let state = Arc::new(Mutex::new(Some("<script>BIDS</script>".to_string())));
    prepend_auction_debug_comment("stream", &result, &state);
    let page = state.lock().expect("should lock").clone().expect("should have html");
    for needle in ["EC-ID-abc123", "EID-USER-999", "CPtc-TCSTRING-xyz", "203.0.113.77", "37.7749"] {
        assert!(page.contains(needle), "leaked into page HTML: {needle}");
    }
}
```

## Checked and *not* a problem

Cross-user cache poisoning. `publisher.rs:1671-1683` forces `Cache-Control: private, max-age=0` and strips `Surrogate-Control` on any ad-stack HTML, which is exactly the condition under which this comment can exist — so one visitor's dump cannot be shared-cached and served to another. Flagging explicitly so nobody spends time on it. The exposure is same-page JS, not the CDN.

Also clean: `GET /__ts/page-bids`, `window.tsjs.bids`, the telemetry event rows, and the mediator request all read `Bid` fields rather than `AuctionResponse.metadata`, so none of them carry the new data.

## Why CI's 19 green checks caught none of this

Both this and the `/auction` exposure from my previous review live in cross-module data-flow contracts with no test at any boundary. Every `formats.rs` fixture hardcodes `metadata: HashMap::new()` and `BidStatus::Success` (`auction/formats.rs:418, 448`), so no error response with metadata is ever pushed through `convert_to_openrtb_response`; the one exact-JSON wire test hardcodes `provider_details: vec![]` (`openrtb.rs:212`); and `provider_status()` has a `_ =>` catch-all (`auction/telemetry.rs:803`) which by construction cannot fail on an unrecognised `error_type`. It's an assertion gap, not an execution gap — the tests do run.

## Suggested minimal fix

Whitelist what goes into the dump instead of serializing the whole `metadata` map: keep `error_type`, `status`, `responsetimemillis`, `errors`, `warnings`, `bidstatus`; exclude `debug` outright (or scrub `resolvedrequest` + `httpcalls[].requestbody`). Then cap the serialized payload — 256 KiB matching `MAX_AUCTION_BODY_SIZE`, with a `<truncated N bytes>` marker, and truncate `Bid.creative` per bid. That closes the identity exposure, the 5xx risk, and the memory pressure in one change, and it is consistent with the 512-byte cap this same PR already applies at `integrations/prebid.rs:2139`.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Fixed in 9d96b722d. The dump now runs through a fail-closed allowlist (`redact_response_for_dump` in `publisher.rs`): only `error_type / status / message / responsetimemillis / errors / warnings / bidstatus` survive, so `metadata["debug"]` (`resolvedrequest` + `httpcalls`) is dropped and the identity graph never reaches the DOM. Full debug detail stays available server-side via `log::trace!`.

Added regression test `auction_debug_comment_never_leaks_provider_debug_metadata` — your reproduction, inverted: it plants the same `resolvedrequest` shape (EC ID, EID, TC string, IP, geo) and asserts every needle is absent from the rendered comment while an allowlisted key still surfaces.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Fixed in 9d96b722d. Single serialize → single neutralise → single 256 KiB total budget (`MAX_AUCTION_DEBUG_DUMP_BYTES`), instead of the previous per-section budget whose worst case was 512 KiB with a mediator. Per-bid creative is previewed to 512 B (`redact_bid_for_dump`), and dropping `metadata.debug` removes the largest contributor (the full per-bidder `httpcalls` request/response bodies). The truncation marker now reports `…(truncated N bytes)`.

Added `auction_debug_comment_truncates_oversized_creative` covering the per-bid cap.

crates/trusted-server-core/src/integrations/prebid.rs
prk-Jr: This one is already resolved on the current branch head. The test now asserts the inverse — `parse_response_attaches_status_metadata_without_leaking_body_on_http_error` checks `!result.metadata.contains_key("body")` and that no metadata value contains the upstream body. The body-into-metadata path was removed in d5871da84; the error response carries only `error_type` + `status`, plus a static `message` as of 9d96b722d. Nothing from the upstream body reaches `ProviderSummary.metadata`, so there's no assertion left to invert.

crates/trusted-server-core/src/integrations/prebid.rs
prk-Jr: Fixed in 9d96b722d. The http-error response now sets a static `message` (`"Prebid returned non-success status"` — no upstream content, per the SECURITY note in `orchestrator.rs`), restoring the `error_type` + `message` shape every other provider error path uses. The test asserts the message is present alongside the existing no-body-leak checks.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Fixed in 9d96b722d. The dump is now built once, then `render_dump` (neutralise + cap) wraps the whole `serde_json::to_string(...).unwrap_or_else(...)` result — so the serialize-error fallback string is neutralised on the same path as the `Ok` branch. There's no longer a branch that reaches the DOM un-neutralised.

crates/trusted-server-core/src/auction/telemetry.rs
prk-Jr: Fixed in 9d96b722d. `http_status_error` is now in the rollup error set in `auction_provider_stats_mv.pipe`, and the documented status list in the design spec is updated.

Coverage on both sides:
- Rust mapping: `provider_call_maps_http_status_errors_to_their_own_bucket` asserts `error_type = "http_status"` → row status `http_status_error` (distinct from `transport_error`).
- Downstream aggregation: `provider_health.yaml` case `http_status_error_counts_as_error`, backed by an isolated fixture row (`2026-06-23 12:03:00`), asserts `errors:1, error_rate:1`.

Caveat: I could not run the Tinybird suite locally (no `tb` workspace here) — the fixture row + test are correct-by-construction on an isolated minute, but please confirm with `tb test run` before merge.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Addressed in 9d96b722d alongside the size/500 fix. There's now one explicit total budget (256 KiB) shared across `provider_responses` + `mediator_response`, rather than a per-section budget. More importantly, the payload is bounded *before* serialization: `redact_response_for_dump` drops `metadata.debug` (the multi-MB per-bidder `httpcalls` bodies) and previews each creative to 512 B, so the transient serialize + `replace` allocations are bounded, not just the final output. Serializing through a fully streaming bounded writer would be a larger change — flagging in case you'd like it tracked as a follow-up.

aram356 (commented):
Superseded — see the approval below, which carries the findings inline.

aram356 (dismissed):

crates/trusted-server-core/src/publisher.rs
```diff
@@ -876,8 +955,68 @@ pub(crate) fn prepend_auction_debug_comment(
...
+    // targeted replacements below cannot re-form either sequence. Applied to the
+    // serialize-error fallback too, so nothing reaches the DOM un-neutralised.
+    let render_dump = |json: String| -> String {
+        let neutralised = json.replace("-->", "-- >").replace("--!>", "-- !>");
+        if neutralised.len() > MAX_AUCTION_DEBUG_DUMP_BYTES {
+            let end = neutralised.floor_char_boundary(MAX_AUCTION_DEBUG_DUMP_BYTES);
+            format!(
+                "{}…(truncated {} bytes)",
+                &neutralised[..end],
+                neutralised.len() - end
```
aram356, line 1011: 📝 **note** — Confirming this rebuttal is correct, and retracting my earlier suggestion.

In my first review I proposed collapsing these two replacements into a single `replace("--", "- -")`, calling it equivalent and easier to audit. That was **wrong**, and this comment explains exactly why. I tested it against the current implementation:

```
input      | replace("--", "- -")    leaks? | this PR's two-replace   leaks?
---------- | ----------------------- | ------ | ---------------------- | ------
"--->"     | "- -->"                 | YES    | "--- >"                | no
"---!>"    | "- --!>"                | YES    | "--- !>"               | no
"----->"   | "- -- -->"              | YES    | "----- >"              | no
```

Because `str::replace` is non-overlapping, on an odd-length dash run the replacement's trailing `-` rejoins the untouched `->` and re-forms a live terminator. My "simplification" would have reintroduced the exact escape vector this code exists to prevent.

No action needed — the code is right. Leaving this on the record so the comment isn't "simplified away" by a future reader (or by me) on the same faulty reasoning.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -860,12 +860,91 @@ pub(crate) fn write_bids_to_state(
...
+    if let Some(creative) = &bid.creative {
+        value["creative"] =
+            serde_json::Value::String(truncate_with_marker(creative, MAX_BID_CREATIVE_DUMP_BYTES));
+    }
+    value
+}
+
+/// Prepend a `<!-- ts-debug: ... -->` HTML comment carrying the full auction
+/// result — pipeline stats plus every provider response, including each bid's
+/// raw `adm` creative markup — onto the shared `ad_bids_state` so it lands
```
aram356: ⛏ **nitpick** — This doc now describes the behaviour as it was *before* the redaction landed in `9d96b722d`.

It says the comment carries "the **full** auction result … including each bid's **raw** `adm` creative markup". As of this head, neither is true: `redact_response_for_dump` filters metadata to `DEBUG_DUMP_METADATA_ALLOWLIST`, and `redact_bid_for_dump` previews each creative to `MAX_BID_CREATIVE_DUMP_BYTES` (512 bytes).

Same drift in two other places, which matters more because operators read them to decide whether to switch the flag on:
- `settings.rs:1905-1908` — "including each bid's raw `adm` creative markup **and provider metadata**"
- `trusted-server.example.toml` `[debug]` — "including each bid's creative markup and allowlisted provider metadata" (the metadata half is right; the creative half still implies full markup)

The drift is in the safe direction — it overstates the exposure — but someone reading `settings.rs` today would reasonably conclude the identity leak is still live, which is the thing you just fixed. Suggest something like:

```rust
/// Prepend a `<!-- ts-debug: ... -->` HTML comment carrying a redacted view of
/// the auction result — pipeline stats plus, per provider, its status, bids
/// (each creative previewed to 512 bytes), and allowlisted metadata. Identity-
/// bearing metadata (notably prebid's `debug` subtree) is dropped; see
/// `DEBUG_DUMP_METADATA_ALLOWLIST`.
```

crates/trusted-server-core/src/publisher.rs
```diff
@@ -860,12 +860,91 @@ pub(crate) fn write_bids_to_state(
...
+        "response_time_ms": response.response_time_ms,
+        "bids": bids,
+        "metadata": metadata,
+    })
+}
+
+/// Build a redacted JSON view of a single bid: every field except `creative`,
+/// which is previewed to [`MAX_BID_CREATIVE_DUMP_BYTES`].
+fn redact_bid_for_dump(bid: &crate::auction::types::Bid) -> serde_json::Value {
+    let mut value = serde_json::to_value(bid).unwrap_or(serde_json::Value::Null);
```
aram356, line 930: 🌱 **seedling** — Bid-level metadata is fail-open, while response-level metadata is fail-closed. Not a live leak; worth a follow-up rather than a change here.

`serde_json::to_value(bid)` serializes the whole `Bid` and only `creative` is subsequently narrowed, so `Bid.metadata`, `nurl`, and `burl` reach the DOM unfiltered. Meanwhile `redact_response_for_dump` puts `AuctionResponse.metadata` through a strict allowlist. Two metadata maps, two opposite defaults.

I checked whether it's exploitable today and it isn't: the only writer of `Bid.metadata` is `integrations/aps.rs:453-467` (`amzniid` / `amznbid` / `amznp` / `amznsz` / `amznactt`), all opaque targeting keys, and prebid never populates it. So this is purely defence-in-depth.

The reason it's worth a note: the `SECURITY:` comment above states that "`Bid.creative` and provider metadata are attacker/partner-influenced" and that layer 1 "drops all non-allowlisted metadata" — which is true of the response map but not the bid map. A future provider that mirrors a bidder's `ext` into `Bid.metadata` would leak into page source silently, with the comment above reading as though it were covered. An allowlist on the bid map (or a note narrowing the comment's claim) would close the gap.

`nurl`/`burl` are the same shape of concern — win/billing notification URLs become readable, and firable, by any script on the page. Debug-flag-gated and documented as never-in-production, so low severity.

aram356 (approved):

crates/trusted-server-core/src/publisher.rs
prk-Jr: Confirmed — leaving the two-replace neutraliser and its comment as-is. The odd-dash-run re-formation you documented (`--->` → `- -->`) is exactly why a single `replace("--", …)` is unsafe. No change here.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Fixed — updated all three sites to describe the redacted behaviour:

- `prepend_auction_debug_comment` doc: now "a redacted view of the auction result … each creative previewed to `MAX_BID_CREATIVE_DUMP_BYTES`, and allowlisted metadata", with identity-bearing metadata (prebid's `debug` subtree) called out as dropped (see `DEBUG_DUMP_METADATA_ALLOWLIST`).
- `settings.rs` `auction_html_comment` doc: creative "previewed (not the full `adm` markup)", metadata "filtered to a fail-closed allowlist that drops identity-bearing keys".
- `trusted-server.example.toml` `[debug]`: creative "previewed (not the full markup)", metadata already described as allowlisted.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Agreed — defence-in-depth, not a live leak. Narrowed the `SECURITY:` comment above `prepend_auction_debug_comment` so it no longer implies the bid map is covered: it now states the allowlist applies to *response-level* metadata, while `Bid.metadata`/`nurl`/`burl` pass through today (safe because `integrations/aps.rs` is the only writer and emits opaque targeting keys). Tracked the fail-closed bid allowlist (plus dropping `nurl`/`burl`) as a follow-up: #925.

PR #896, opened 2026-07-12
Optimize dev proxy upstream performance

### LINKED ISSUE #895: Optimize dev proxy upstream performance
## Problem

`ts dev proxy` opens a new upstream TCP connection and TLS session for every forwarded request. Realistic browser page loads therefore repeat connection setup and add avoidable latency, especially against remote or staging upstreams.

## Scope

- Establish repeatable sequential, concurrent, and remote-latency benchmarks.
- Reuse HTTP/1.1 upstream connections with bounded per-origin and global capacity.
- Preserve request and response streaming, backpressure, trailers, cancellation, and hop-by-hop header semantics.
- Add bounded DNS caching and buffered CONNECT-head parsing only when their performance gates are met.
- Keep shutdown, retry, TLS identity, `--resolve`, and insecure-mode behavior safe and deterministic.
- Evaluate HTTP/2 and `TCP_NODELAY` experimentally and remove them if their retention gates are not met.

## Acceptance criteria

- One hundred sequential zero-body GETs reuse one upstream TCP/TLS connection.
- Concurrent traffic respects six live connections per origin and 64 globally.
- Streaming bodies remain byte-identical and are never replayed unsafely.
- Stale reused connections retry at most once and only for proven replayable requests.
- DNS, waiters, idle connections, and lifecycle state remain bounded.
- Ctrl-C restores system proxy state before the bounded manager drain.
- CLI tests, E2E tests, formatting, Clippy, docs formatting, and performance workloads pass.

aram356:
Closes #895

## Summary

- Add repeatable dev-proxy performance metrics and benchmark workloads for sequential, matched-concurrency, saturation, DNS, parser, and remote-latency scenarios.
- Add a bounded HTTP/1.1 upstream connection manager with safe streaming lease return, cancellation, stale-connection retry, connector ownership, and bounded shutdown.
- Retain evidence-backed CONNECT-head parsing and DNS caching while rejecting HTTP/2 and `TCP_NODELAY` experiments that did not satisfy their retention gates.
- Precompute upstream identities and preserve TLS, Host, `--resolve`, Basic authentication, hop-by-hop sanitation, and blind-tunnel isolation.

## Final review corrections

- Count complete responses as successful independently from whether their upstream connection is reusable, including already-empty and `Connection: close` responses.
- Force stale-readiness recovery through a fresh reservation instead of selecting another idle sender.
- Remove the rejected HTTP/2 application-mode key state; retained TLS configurations advertise only HTTP/1.1.
- Forward response trailers before lease reconciliation and preserve every downstream trailer declaration.
- Backfill connector cancellation/unwind, early-response, browser-cancellation, truncated-body, trailer, credential-isolation, insecure-warning, forwarding-capacity, over-read, shutdown-order, and origin-key integration coverage.

## Performance results

- Sequential TLS: 100 connections and handshakes reduced to 1, with a 59.7% median duration improvement in the recorded alternating runs.
- DNS churn workload: 13.6% median improvement with one miss and 99 cache hits.
- Concurrency-20 saturation: bounded to six upstream connections and handshakes.
- Remote saturation diagnostic: six pooled connections reduce connection and handshake churn by 94%; the six-connection safety cap intentionally trades 9.0% duration versus the unbounded baseline at concurrency 20.

Full measurements and experiment decisions are recorded in `docs/superpowers/implementation-notes/2026-07-10-dev-proxy-performance.md`.

## Verification

- [x] CLI `cargo check --tests` and Clippy with `-D warnings`
- [x] `./scripts/test-cli.sh` — 147 unit + 26 E2E tests
- [x] All six ignored `proxy_perf` workloads with one test thread
- [x] Fastly, Axum, Cloudflare, Spin, and 13 cross-adapter parity tests
- [x] Native/Wasm Clippy matrices for Fastly, Axum, Cloudflare, and Spin with zero warnings
- [x] JS build + 405 Vitest tests + JS/docs Prettier checks

### DIFF
--- a/crates/trusted-server-cli/src/commands/dev/proxy/metrics.rs
+++ b/crates/trusted-server-cli/src/commands/dev/proxy/metrics.rs
@@ -1,0 +1,265 @@
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::time::Duration;
+
+const DURATION_BUCKET_UPPER_BOUNDS_MICROS: [u64; 9] = [
+    100, 500, 1_000, 5_000, 10_000, 50_000, 100_000, 500_000, 1_000_000,
+];
+const DURATION_BUCKET_COUNT: usize = DURATION_BUCKET_UPPER_BOUNDS_MICROS.len() + 1;
+
+/// Low-overhead process-local metrics for `ts dev proxy`.
+#[derive(Default)]
+pub struct ProxyMetrics {
+    initial_heads_parsed: AtomicU64,
+    initial_heads_rejected: AtomicU64,
+    tcp_attempts: AtomicU64,
+    tcp_established: AtomicU64,
+    initial_head_parse_latency: DurationHistogram,
+    connect_latency: DurationHistogram,
+    pool_acquisition_latency: DurationHistogram,
+    queue_wait_latency: DurationHistogram,
+}
+
+impl ProxyMetrics {
+    pub fn record_initial_head_parsed(&self, duration: Duration) {
+        self.initial_heads_parsed.fetch_add(1, Ordering::Relaxed);
+        self.initial_head_parse_latency.record(duration);
+    }
+
+    pub fn record_initial_head_rejected(&self, duration: Duration) {
+        self.initial_heads_rejected.fetch_add(1, Ordering::Relaxed);
+        self.initial_head_parse_latency.record(duration);
+    }
+
+    pub fn record_tcp_attempt(&self) {
+        self.tcp_attempts.fetch_add(1, Ordering::Relaxed);
+    }
+
+    pub fn record_tcp_established(&self, duration: Duration) {
+        self.tcp_established.fetch_add(1, Ordering::Relaxed);
+        self.connect_latency.record(duration);
+    }
+
+    pub fn record_pool_acquisition(&self, duration: Duration) {
+        self.pool_acquisition_latency.record(duration);
+    }
+
+    pub fn record_queue_wait(&self, duration: Duration) {
+        self.queue_wait_latency.record(duration);
+    }
+
+    #[must_use]
+    pub fn debug_summary(&self) -> String {
+        let snapshot = self.snapshot();
+        format!(
+            "dev proxy metrics: initial_heads_parsed={} initial_heads_rejected={} \
+             initial_head_parse_samples={} tcp_attempts={} tcp_established={} \
+             connect_samples={} pool_acquisition_samples={} queue_wait_samples={} \
+             duration_bounds_us={:?} initial_head_parse_us_buckets={:?} \
+             connect_us_buckets={:?} pool_acquisition_us_buckets={:?} \
+             queue_wait_us_buckets={:?}",
+            snapshot.initial_heads_parsed,
+            snapshot.initial_heads_rejected,
+            snapshot.initial_head_parse_latency.total(),
+            snapshot.tcp_attempts,
+            snapshot.tcp_established,
+            snapshot.connect_latency.total(),
+            snapshot.pool_acquisition_latency.total(),
+            snapshot.queue_wait_latency.total(),
+            DURATION_BUCKET_UPPER_BOUNDS_MICROS,
+            snapshot.initial_head_parse_latency.buckets(),
+            snapshot.connect_latency.buckets(),
+            snapshot.pool_acquisition_latency.buckets(),
+            snapshot.queue_wait_latency.buckets(),
+        )
+    }
+
+    #[must_use]
+    pub fn snapshot(&self) -> ProxyMetricsSnapshot {
+        ProxyMetricsSnapshot {
+            initial_heads_parsed: self.initial_heads_parsed.load(Ordering::Relaxed),
+            initial_heads_rejected: self.initial_heads_rejected.load(Ordering::Relaxed),
+            tcp_attempts: self.tcp_attempts.load(Ordering::Relaxed),
+            tcp_established: self.tcp_established.load(Ordering::Relaxed),
+            initial_head_parse_latency: self.initial_head_parse_latency.snapshot(),
+            connect_latency: self.connect_latency.snapshot(),
+            pool_acquisition_latency: self.pool_acquisition_latency.snapshot(),
+            queue_wait_latency: self.queue_wait_latency.snapshot(),
+        }
+    }
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct ProxyMetricsSnapshot {
+    pub initial_heads_parsed: u64,
+    pub initial_heads_rejected: u64,
+    pub tcp_attempts: u64,
+    pub tcp_established: u64,
+    pub initial_head_parse_latency: DurationHistogramSnapshot,
+    pub connect_latency: DurationHistogramSnapshot,
+    pub pool_acquisition_latency: DurationHistogramSnapshot,
+    pub queue_wait_latency: DurationHistogramSnapshot,
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct DurationHistogramSnapshot {
+    buckets: [u64; DURATION_BUCKET_COUNT],
+}
+
+impl DurationHistogramSnapshot {
+    #[must_use]
+    pub fn buckets(&self) -> &[u64; DURATION_BUCKET_COUNT] {
+        &self.buckets
+    }
+
+    #[must_use]
+    pub fn total(&self) -> u64 {
+        self.buckets.iter().sum()
+    }
+}
+
+struct DurationHistogram {
+    buckets: [AtomicU64; DURATION_BUCKET_COUNT],
+}
+
+impl Default for DurationHistogram {
+    fn default() -> Self {
+        Self {
+            buckets: std::array::from_fn(|_| AtomicU64::new(0)),
+        }
+    }
+}
+
+impl DurationHistogram {
+    fn record(&self, duration: Duration) {
+        let micros = u64::try_from(duration.as_micros()).unwrap_or(u64::MAX);
+        let bucket = DURATION_BUCKET_UPPER_BOUNDS_MICROS
+            .iter()
+            .position(|upper_bound| micros <= *upper_bound)
+            .unwrap_or(DURATION_BUCKET_COUNT - 1);
+        self.buckets[bucket].fetch_add(1, Ordering::Relaxed);
+    }
+
+    #[must_use]
+    fn snapshot(&self) -> DurationHistogramSnapshot {
+        DurationHistogramSnapshot {
+            buckets: std::array::from_fn(|index| self.buckets[index].load(Ordering::Relaxed)),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn snapshot_separates_attempts_from_established() {
+        let metrics = ProxyMetrics::default();
+
+        metrics.record_tcp_attempt();
+        metrics.record_tcp_attempt();
+        metrics.record_tcp_established(Duration::from_millis(7));
+        let snapshot = metrics.snapshot();
+
+        assert_eq!(snapshot.tcp_attempts, 2, "should count every attempt");
+        assert_eq!(
+            snapshot.tcp_established, 1,
+            "should count only established connections"
+        );
+        assert_eq!(
+            snapshot.connect_latency.total(),
+            1,
+            "should record one successful connect duration"
+        );
+    }
+
+    #[test]
+    fn histogram_uses_fixed_bounded_buckets() {
+        let metrics = ProxyMetrics::default();
+
+        for duration in [
+            Duration::ZERO,
+            Duration::from_micros(100),
+            Duration::from_micros(101),
+            Duration::from_secs(1),
+            Duration::from_micros(1_000_001),
+        ] {
+            metrics.record_tcp_established(duration);
+        }
+        let snapshot = metrics.snapshot();
+
+        assert_eq!(
+            snapshot.connect_latency.buckets(),
+            &[2, 1, 0, 0, 0, 0, 0, 0, 1, 1],
+            "should place timings into fixed buckets with a final overflow bucket"
+        );
+    }
+
+    #[test]
+    fn snapshot_records_initial_head_and_pool_phase_timings() {
+        let metrics = ProxyMetrics::default();
+
+        metrics.record_initial_head_parsed(Duration::from_micros(250));
+        metrics.record_initial_head_rejected(Duration::from_micros(500));
+        metrics.record_pool_acquisition(Duration::from_millis(2));
+        metrics.record_queue_wait(Duration::from_millis(3));
+        let snapshot = metrics.snapshot();
+
+        assert_eq!(
+            snapshot.initial_heads_parsed, 1,
+            "should count parsed heads"
+        );
+        assert_eq!(
+            snapshot.initial_heads_rejected, 1,
+            "should count rejected heads"
+        );
+        assert_eq!(
+            snapshot.initial_head_parse_latency.total(),
+            2,
+            "should time both accepted and rejected parsing"
+        );
+        assert_eq!(
+            snapshot.pool_acquisition_latency.total(),
+            1,
+            "should time pool acquisition"
+        );
+        assert_eq!(
+            snapshot.queue_wait_latency.total(),
+            1,
+            "should time queue waits"
+        );
+    }
+
+    #[test]
+    fn debug_summary_contains_only_aggregate_metrics() {
+        let metrics = ProxyMetrics::default();
+        metrics.record_tcp_attempt();
+        metrics.record_tcp_established(Duration::from_millis(7));
+        metrics.record_initial_head_parsed(Duration::from_micros(250));
+
+        let summary = metrics.debug_summary();
+
+        assert!(
+            summary.contains("tcp_attempts=1"),
+            "should report aggregate attempt count"
+        );
+        assert!(
+            summary.contains("connect_samples=1"),
+            "should report aggregate timing count"
+        );
+        assert!(
+            summary.contains("initial_heads_parsed=1"),
+            "should report aggregate parse count"
+        );
+        assert!(
+            summary.contains("duration_bounds_us=[100, 500")
+                && summary.contains("connect_us_buckets=[0, 0, 0, 0, 1"),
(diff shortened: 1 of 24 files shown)


crates/trusted-server-cli/src/commands/dev/proxy/upstream/connect.rs
```diff
@@ -0,0 +1,421 @@
...
+    fn verify_tls13_signature(
+        &self,
+        _: &[u8],
+        _: &rustls::pki_types::CertificateDer<'_>,
+        _: &rustls::DigitallySignedStruct,
+    ) -> Result<rustls::client::danger::HandshakeSignatureValid, rustls::Error> {
+        Ok(rustls::client::danger::HandshakeSignatureValid::assertion())
+    }
+    fn supported_verify_schemes(&self) -> Vec<rustls::SignatureScheme> {
+        rustls::crypto::aws_lc_rs::default_provider()
```
prk-Jr: ♻️ **refactor** — `aws_lc_rs::default_provider()` builds a fresh `CryptoProvider` (cloning every cipher-suite/signature algorithm) on each call, and rustls can invoke `supported_verify_schemes()` per handshake. Consider caching the scheme list once:

```rust
fn supported_verify_schemes(&self) -> Vec<rustls::SignatureScheme> {
    static SCHEMES: std::sync::OnceLock<Vec<rustls::SignatureScheme>> = std::sync::OnceLock::new();
    SCHEMES
        .get_or_init(|| {
            rustls::crypto::aws_lc_rs::default_provider()
                .signature_verification_algorithms
                .supported_schemes()
        })
        .clone()
}
```

Only reached in `--insecure` mode (loud-warned), so purely cosmetic — but on-theme for a PR about avoiding handshake cost.

crates/trusted-server-cli/src/commands/dev/proxy/upstream/mod.rs
```diff
@@ -0,0 +1,446 @@
...
+
+        let (parts, body) = request.into_parts();
+        let (body, upload_state) =
+            RequestUploadBody::new(body, metadata.upload_initially_complete());
+        let request = Request::from_parts(parts, body);
+        let mut request = request;
+        let response = loop {
+            match lease.connection.value.send_request(request).await {
+                Ok(response) => break response,
+                Err(_error) if lease.reused && !stale_retry && replay_template.is_some() => {
```
prk-Jr, line 290: 🌱 **seedling** — The stale-connection `send_request` retry is intentionally gated on `replay_template.is_some()` (provably-empty GET/HEAD/OPTIONS). A POST that races an upstream idle-close after `ready()` already passed therefore returns 502 with no replay. This is the **correct** safe choice — replaying a non-idempotent request could double-submit — noting it only so the trade-off is documented and deliberate rather than read as an oversight.

crates/trusted-server-cli/src/commands/dev/proxy/upstream/body.rs
```diff
@@ -0,0 +1,442 @@
...
+        }
+        if reusable {
+            self.manager.return_idle(lease.connection);
+        } else {
+            lease.connection.abort.abort();
+        }
+    }
+}
+
+fn can_reuse(
```
prk-Jr, line 177: 👍 **praise** — The `can_reuse` matrix (response complete AND upload complete AND no close-intent AND driver alive) is exactly the right set of conditions, and counting a complete-but-unpoolable response as a success independent of reuse is the correct semantics. The exhaustive test alongside it makes the invariant hard to regress.

prk-Jr (approved):
## Summary

Adds a bounded, reusable HTTP/1.1 upstream connection manager (plus DNS caching, precomputed upstream identities, and a repeatable benchmark suite) to `ts dev proxy`. The design is careful and thoroughly tested: the manager is an actor with a **bounded ordinary lane + unbounded control lane** so lifecycle events (cancel / driver-closed / shutdown) can never deadlock behind a saturated acquire lane, and capacity reconciliation is airtight — every path leads back to `release(id)` (reservation drop → `ConnectFailed`, driver-guard drop → `DriverClosed`, shutdown connector abort → finish error → reservation drop). The `AcquireTicket` single-terminal-transition invariant closes the resolve/cancel race. No blocking issues found; approving.

## Non-blocking

### ♻️ refactor
- **Crypto provider rebuilt per call**: `NoVerifier::supported_verify_schemes` calls `aws_lc_rs::default_provider()` (a non-trivial clone) on every invocation — cache in a `OnceLock`. `--insecure`-only path, cosmetic. (connect.rs:282)

### 🌱 seedling
- **Non-idempotent 502 on stale reuse**: the `send_request` retry is gated on provably-empty idempotent requests, so a POST racing an upstream idle-close returns 502 rather than replaying. This is the correct safe behavior; flagged only so it's a documented trade-off. (upstream/mod.rs:246)

### ⛏ nitpick
- **Doc comments on new `pub` items**: several new public items (`ProxyMetrics::record_*`, `PooledResponseBody::new`, etc.) lack doc comments. CLAUDE.md asks for one per public item; `missing_docs` isn't enforced here and it matches surrounding CLI style, so optional.

### 👍 praise
- Lifecycle & concurrency design: the `can_reuse` matrix + exhaustive test, DNS miss-coalescing that survives owner cancellation without poisoning the in-flight key, the trailer-frame deferral in `PooledResponseBody`, and the `priority_shutdown_overtakes_a_saturated_ordinary_lane` test are all excellent.
- Security hardening: Basic-auth refused on non-loopback binds, hop-by-hop + `Connection`-token stripping, `Forwarded` / `Fastly-SSL` stripped and authoritatively re-stamped, strict host validation before PAC / Host-header embedding, and redacted `BasicAuth` Debug.

## CI Status

- fmt: PASS
- clippy (fastly/axum/cloudflare/spin native+wasm): PASS
- rust tests (fastly/axum/cloudflare/spin/CLI/parity): PASS
- js build + vitest + docs/ts prettier: PASS

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
```diff
@@ -603,9 +641,7 @@ fn rewrite_headers(
     // Strip hop-by-hop headers first, so a client cannot flag the authoritative
     // headers we stamp below as connection-specific and have them dropped.
     strip_hop_by_hop(headers);
-    if let Ok(value) = HeaderValue::from_str(&outcome.host_header) {
-        headers.insert(hyper::header::HOST, value);
-    }
+    headers.insert(hyper::header::HOST, outcome.host_header.clone());
```
ChristianPavilonis, line 714: 🔧 **P1 — Preserve request-trailer semantics on the upstream leg**

`rewrite_headers` removes both `Trailer` and `TE` through `strip_hop_by_hop` without regenerating them. `RequestUploadBody` still forwards trailer frames, but Hyper's HTTP/1 encoder only serializes fields declared by the outgoing `Trailer` header; otherwise it discards them. A chunked upload can therefore arrive intact except for silently lost trailer fields, and removing `TE: trailers` prevents compliant origins from knowing the proxy accepts response trailers.

Capture the browser's trailer declarations and `TE: trailers` before sanitation, then regenerate safe upstream-leg `Trailer` and `TE`/`Connection: TE` metadata. Please add an E2E test proving request trailers reach the origin and a compliant origin can conditionally return response trailers.

crates/trusted-server-cli/src/commands/dev/proxy/upstream/connect.rs
```diff
@@ -0,0 +1,421 @@
...
+            .task
+            .as_mut()
+            .expect("should retain connector task")
+            .await;
+        self.task.take();
+        result.map_err(|error| Report::new(io::Error::other(error.to_string())))?
+    }
+}
+
+impl Drop for PendingConnection {
```
ChristianPavilonis, line 143: 🔧 **P1 — Release connector capacity only after cancellation completes**

Dropping `PendingConnection` only calls `AbortHandle::abort()`, while the independently dropped `Reservation` immediately sends `ConnectFailed`; the manager can then release capacity before Tokio has actually dropped the DNS/TCP/TLS task. Replacement admission or shutdown acknowledgement can occur while the old connector remains live, violating the hard six-per-origin/64-global bound. The current test yields until `abort.is_finished()` before reacquiring, which masks this ordering.

Transfer reconciliation ownership to an exactly-once guard inside the connector task, and release the reservation only when that task actually exits (disarming the connector guard after successful driver registration). Add a test proving replacement admission and shutdown remain blocked until cancellation completion.

crates/trusted-server-cli/src/commands/dev/proxy/upstream/manager.rs
```diff
@@ -0,0 +1,919 @@
...
+        }
+        result
+    }
+
+    fn return_connection(&mut self, connection: IdleConnection<T>) {
+        if self.closing || !self.live.contains_key(&connection.id) {
+            connection.abort.abort();
+            return;
+        }
+        if let Some(index) = self.waiters.iter().position(|waiter| {
```
ChristianPavilonis: 🔧 **P2 — Continue handoff after a waiter loses the cancellation race**

A waiter can cancel after this `position` check but before `ticket.resolve()`. The actor then removes that waiter, stores the returned connection as idle, and never checks the next eligible same-origin waiter. With no further connection event, that waiter can time out after 30 seconds while a reusable connection sits idle for 60 seconds; newer requests may also take it first.

Make handoff loop over matching waiters until delivery succeeds or none remain. If the oneshot send fails, recover the lease payload and retry the next waiter. A deterministic two-waiter cancellation-race test would protect the FIFO invariant.

crates/trusted-server-cli/src/commands/dev/proxy/upstream/mod.rs
```diff
@@ -0,0 +1,446 @@
...
+        outcome: &RewriteOutcome,
+    ) -> Result<Response<BoxBody<Bytes, hyper::Error>>, Report<ProxyError>> {
+        let request_started = tokio::time::Instant::now();
+        let replay_template = ReplayTemplate::capture(&request, metadata);
+        let mut stale_retry = false;
+        let mut acquisition_mode = AcquisitionMode::Normal;
+        let mut lease = loop {
+            let acquisition_started = tokio::time::Instant::now();
+            let acquired =
+                acquire_for_mode(&self.manager, rule.origin_key().clone(), acquisition_mode)
```
ChristianPavilonis: 🔧 **P2 — Count setup and acquisition failures as failed requests**

Acquisition errors here and connector errors at lines 209–212 return before `record_request_failed`; stale-replacement acquisition/connection has the same omission. `forward_request` converts these errors to visible 502 responses, but `requests_failed` remains zero, so the new diagnostics under-report common DNS/TCP/TLS/pool failures. Failed acquisition latency is also omitted.

Ensure every terminal `UpstreamClient::send` error records exactly one failure—preferably with a scoped outcome guard or centralized error boundary—and record acquisition duration on both success and failure. Add an unreachable-loopback-origin E2E assertion for `502`, zero completions, and one failure.

crates/trusted-server-cli/src/commands/dev/proxy/upstream/dns.rs
```diff
@@ -0,0 +1,610 @@
...
+    resolver: Arc<dyn Resolver>,
+    key: CacheKey,
+    generation: u64,
+    deadline: Instant,
+    metrics: Arc<ProxyMetrics>,
+    signal: watch::Sender<Option<SharedResult>>,
+) {
+    let result = resolve(resolver, &key, deadline, &metrics).await;
+    let shared = result.as_ref().map(Arc::clone).map_err(OwnedError::from_io);
+    let _ = signal.send(Some(shared.clone()));
```
ChristianPavilonis: 🔧 **P2 — Retain the completed lookup when all original receivers cancel**

`watch::Sender::send` does not retain the value when there are no receivers. A new caller can subscribe to the still-present `Loading` entry between this failed send and cache-state replacement, then observe channel closure and return `Interrupted` even though the lookup completed.

```suggestion
    signal.send_replace(Some(shared.clone()));
```

Please add a multithreaded regression test where every initial waiter cancels and a late subscriber arrives during completion.

crates/trusted-server-cli/tests/proxy_perf.rs
```diff
@@ -0,0 +1,215 @@
...
+
+#[tokio::test]
+#[ignore = "manual performance workload"]
+async fn perf_pooled_saturation_concurrency_twenty() {
+    run_concurrent_pooled("saturation_concurrency_20", 20).await;
+}
+
+#[tokio::test]
+#[ignore = "manual performance workload"]
+async fn perf_http1_remote_model() {
```
ChristianPavilonis, line 78: 🔧 **P2 — Make the named remote benchmark self-validating**

This test labels output `http1_remote_model_30_30_25`, but the 30 ms connect and TLS delays are enabled only when `TS_PERF_VARIANT` starts with `remote_`. Under the default `pooled` variant, the named remote workload silently runs with only the 25 ms response delay, allowing mislabeled benchmark evidence.

Require/assert `remote_baseline` or `remote_pooled` for this test, or construct typed remote options directly. Please also document the exact alternating commands used for the recorded remote results.

crates/trusted-server-cli/src/commands/dev/proxy/metrics.rs
```diff
@@ -0,0 +1,416 @@
...
+        if unexpected {
+            self.unexpected_ca_mints.fetch_add(1, Ordering::Relaxed);
+        }
+    }
+
+    #[must_use]
+    pub fn debug_summary(&self) -> String {
+        let snapshot = self.snapshot();
+        format!(
+            "dev proxy metrics: browser_connections={} initial_heads_parsed={} initial_heads_rejected={} \
```
ChristianPavilonis, line 162: 🔧 **P2 — Include the latency values advertised by the troubleshooting guide**

The debug summary prints only sample counts for DNS, TLS, HTTP handshake, request-to-headers, and CA-mint timing. It emits latency buckets only for parsing, TCP connect, pool acquisition, and queue wait. Consequently the guide's claim that this summary can distinguish DNS and TLS latency is not currently true.

Include totals or histogram buckets for every advertised latency phase, or narrow the documentation claim to the phases actually emitted.

ChristianPavilonis (approved):
## Summary

Approved with seven actionable follow-ups noted inline. The pooled HTTP/1 architecture and coverage are strong; the highest-priority changes are preserving request trailers end-to-end and tying connector-capacity release to actual task termination.

crates/trusted-server-cli/src/commands/dev/proxy/upstream/connect.rs
aram356: Addressed in 3764bcb9b. The supported signature-scheme list is now cached in a process-wide OnceLock, while each TLS config still receives its own Vec. The TLS client-config regression coverage and full CLI gates pass.

crates/trusted-server-cli/src/commands/dev/proxy/upstream/mod.rs
aram356: Addressed in 3764bcb9b. I documented the deliberate safety choice at the retry branch: non-idempotent or streaming requests return 502 after a post-dispatch failure instead of being replayed. The existing POST non-retry regression test remains green.

crates/trusted-server-cli/src/commands/dev/proxy/upstream/body.rs
aram356: Thanks. I retained the explicit can_reuse matrix and its coverage while applying the surrounding lifecycle fixes; the full CLI suite remains green.

crates/trusted-server-cli/src/commands/dev/proxy/server.rs
aram356: Addressed in 3764bcb9b. Request Trailer and TE metadata are captured before hop-by-hop sanitation, unsafe trailer names are filtered, and safe metadata is regenerated for the upstream leg. Added unit coverage plus the request_and_response_trailers_cross_the_upstream_leg end-to-end test.

crates/trusted-server-cli/src/commands/dev/proxy/upstream/connect.rs
aram356: Addressed in 3764bcb9b with connector-owned lifecycle completion. Pending connectors now own the reservation; cancellation and errors reconcile from inside that task. Successful-but-unclaimed connections use a guarded result that retains capacity until DriverClosed. Added cancellation, shutdown-drain, and delayed-driver-exit regression tests.

crates/trusted-server-cli/src/commands/dev/proxy/upstream/manager.rs
aram356: Addressed in 3764bcb9b. A failed waiter delivery now returns the Lease to the manager, which continues scanning same-origin reusable waiters before idling the connection. Added failed_waiter_delivery_hands_connection_to_next_waiter.

crates/trusted-server-cli/src/commands/dev/proxy/upstream/mod.rs
aram356: Addressed in 3764bcb9b. A scoped send-outcome guard now records exactly one pre-response failure, and acquisition latency is recorded on both success and failure paths, including stale replacement. Added unreachable-origin and failed-acquisition end-to-end metric tests.

crates/trusted-server-cli/src/commands/dev/proxy/upstream/dns.rs
aram356: Addressed in 3764bcb9b. DNS completion now uses send_replace so the completed result remains in the watch channel even when all initial waiters cancel. Added a deterministic multi-threaded cancellation regression test.

crates/trusted-server-cli/tests/proxy_perf.rs
aram356: Addressed in 3764bcb9b. Remote benchmark execution now validates and accepts only remote_baseline and remote_pooled. Added unit coverage and documented the exact alternating warmup and ten-pair recorded command loop.

crates/trusted-server-cli/src/commands/dev/proxy/metrics.rs
aram356: Addressed in 3764bcb9b. The debug summary now reports sample count, total microseconds, and histogram buckets for every tracked latency, with the summary test extended accordingly.

PR #899, opened 2026-07-13
Render SSAT winning creative inline without PBS Cache round trip

### LINKED ISSUE #898: SSAT should return creatives to render without additional round trip to server side
### Description

SSAT should return creatives to render without additional round trip to server side

### LINKED ISSUE #926: Pin and browser-test the GAM-served PUC render contract for inline SSAT creatives
## Context

PR #899 ships the winning creative inline in `window.tsjs.bids[...].adm`, rendered
by the Prebid Universal Creative (PUC) via the pbRender bridge
(`installTsRenderBridge` → `TS_DISPLAY_RENDERER`, `f.srcdoc = d.ad`) inside GAM's
iframe. Several open questions about the **GAM-served PUC render contract** were
deferred from that PR; this task tracks pinning and testing it.

Why load-bearing: the render context (which origin the `srcdoc` document's base
URL inherits, whether PUC honors a custom `renderer`) determines whether the
inline `adm`'s asset/click URLs resolve. PR #899 mitigated the immediate risk by
emitting **absolute** first-party URLs (`crate::creative::rewrite_inline_creative_html`
→ `https://<publisher.domain>/first-party/…`), which resolve regardless of the
document base URL and regardless of whether the custom renderer is honored.
Confirming the actual contract closes the remaining questions below.

## Tasks

- [ ] **Pin the PUC version/template** used by the GAM-served bundle. Confirm
  whether `rendererVersion >= 3` (so the custom `renderer` field —
  `TS_DISPLAY_RENDERER` — is honored) or whether PUC's own fallback path renders
  `d.ad`. Record the pinned version/template.
- [ ] **Add a browser contract test** against the GAM-served PUC bundle that
  exercises the end-to-end inline render (mocktioneer bid winning so the bridge
  actually fires). Per PR #899's test plan this path was never exercised —
  GAM usually awards its own demand, so the bridge never fired. The test should
  assert the absolute first-party URLs in the inline `adm` resolve in the render
  context.
- [ ] **Confirm cookie context**: verify whether first-party proxy requests
  issued from GAM's iframe are a third-party cookie context, so EC cookies may
  not attach as they would first-party. Assess impact on measurement/attribution.
- [ ] **JS decline-fast** (`gpt/index.ts` `installTsRenderBridge`): once the PUC
  version is pinned, reply to a declined "Prebid Request" with `adRenderFailed`
  (PUC 1.18.0 emits it) so the slot fails fast / collapses instead of staying
  blank after `stopImmediatePropagation()`. Applies to both the inline-adm
  decline and the PBS Cache `.catch`/no-`adm` decline.
- [ ] **Aggregate inline-`adm` byte cap**: the 1 MiB cap in
  `sanitize_creative_html` is per-creative. `build_bid_map` inlines the creative
  for every winning slot into the page's `<script>`, so N slots can inline up to
  N MiB with no aggregate bound — pure document-weight/TTFB cost in the common
  case where GAM awards its own demand and the bridge never fires. Add an
  aggregate budget in `build_bid_map` (omit `adm` past the budget → fall back to
  PBS Cache coordinates), or a narrower condition for inlining.

## References

- PR #899 (feat/ssat-render-inline-creative)
- `crates/trusted-server-core/src/publisher.rs` — `build_bid_map`
- `crates/trusted-server-core/src/creative.rs` — `rewrite_inline_creative_html`
- `crates/trusted-server-js/lib/src/integrations/gpt/index.ts` — `installTsRenderBridge`, `TS_DISPLAY_RENDERER`, `extractCachedAdm`

prk-Jr:
## Summary

- On the server-side auction (SSAT / streaming path), the winning creative `adm`
  is now always included in `window.tsjs.bids`, so the existing render bridge
  serves it **locally** when GAM's Prebid line item fires — eliminating the
  render-time PBS Cache round trip and matching how client-side `/auction`
  renders from the copy the browser already holds.
- GAM stays in the loop: the header bid still competes via `hb_pb`, and
  `hb_cache_host`/`hb_cache_path` remain as the fallback for an **absent** `adm`.
  No GAM bypass in production.
- The verbose `debug_bid` blob stays behind `inject_adm_for_testing`. The
  GAM-bypass path (`injectAdmIntoSlot`) is now gated on the per-bid `debug_bid`
  field instead of `bid.adm` — no new global flag, no `TsjsApi` change, correct
  across SPA auction responses.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/publisher.rs` | `build_bid_map` always inserts `adm` when `bid.creative` is `Some`; param `include_adm` renamed to `include_debug_bid` (now gates only the `debug_bid` blob); `write_bids_to_state` param renamed to match. Added a hostile-`adm` regression test proving `</script>` / `U+2028` / `U+2029` cannot break out of the emitted `<script>`. |
| `crates/trusted-server-js/lib/src/integrations/gpt/index.ts` | Gate the direct GAM-replacement bypass on `bid.adm && bid.debug_bid`; render-bridge log message updated to "from inline adm". |
| `crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts` | Added bypass-gate tests (bypass fires only when `debug_bid` is present); renamed "debug adm" terminology to "inline/local adm"; bridge fixtures now carry both `hb_cache_*` coords and inline `adm`, proving the bridge prefers local `adm`. |
| `docs/superpowers/specs/2026-07-13-ssat-render-inline-creative-design.md` | Design spec. |
| `docs/superpowers/plans/2026-07-13-ssat-render-inline-creative.md` | Implementation plan. |

## Closes

Closes #898

## Test plan

- [x] `cargo test-fastly && cargo test-axum` (also `cargo test-cloudflare && cargo test-spin` — all green)
- [ ] `cargo clippy-fastly && cargo clippy-axum` (only `clippy-fastly` run locally; full matrix delegated to CI)
- [x] JS tests: `cd crates/trusted-server-js/lib && npx vitest run` (407 passed)
- [ ] WASM build: `cargo build --package trusted-server-adapter-fastly --release --target wasm32-wasip1` (delegated to CI)
- [x] Other: manual local dev confirmed `adm` now present in `window.tsjs.bids` for all slots. The full inline-render path (GAM awards the Prebid line item → creative renders with **no** PBS Cache fetch) is **not yet exercised** locally — mocktioneer's $0.20 bid loses to GAM's own demand, so GAM serves its own creative and the bridge never fires. Requires a winning Prebid line item in GAM to verify end-to-end.

## Checklist

- [x] Uses `tracing` macros (not `println!`) — no new Rust logging introduced
- [x] New code has tests


crates/trusted-server-core/src/publisher.rs
```diff
@@ -1983,12 +1983,16 @@ pub(crate) fn build_bid_map(
...
-                // The pbRender bridge can use it while PBS Cache is unavailable.
-                if include_adm {
-                    if let Some(ref adm) = bid.creative {
-                        obj.insert("adm".to_string(), serde_json::Value::String(adm.clone()));
-                    }
+                // Always include the winning creative so the pbRender bridge can
+                // render it locally when GAM serves the Prebid Universal Creative
+                // — no PBS Cache round trip. The `hb_cache_*` coordinates above
+                // remain as the fallback for an absent `adm`.
+                if let Some(ref adm) = bid.creative {
```
aram356: 🔧 **Blocking: preserve the production creative-processing boundary.**

This copies raw `bid.creative` into the client-facing bid map, while the established `/auction` path first calls `sanitize_creative_html` and `rewrite_creative_html`. That pipeline removes dangerous markup and enforces the deliberate 1 MiB limit; the browser renderer also explicitly documents server-side sanitization as its primary defense.

The JSON/script-context escaping test is valuable, but it only prevents breaking out of the containing script. It does not apply the render-time processing or resource limit before PUC receives the creative. This also changes the old testing-only raw-`adm` behavior into an unconditional production path.

Please share the existing sanitize/rewrite preparation with SSAT before building `window.tsjs.bids`. If PUC intentionally requires unsanitized creatives, preserve an explicit size cap, document the PUC isolation boundary that makes raw execution acceptable, and add hostile render-time plus oversized-creative coverage before enabling it unconditionally.

docs/superpowers/specs/2026-07-13-ssat-render-inline-creative-design.md
```diff
@@ -0,0 +1,164 @@
...
+- **Always** insert `adm` (from `bid.creative`) for a winner when present —
+  there is no runtime reason to withhold it, so it is not parameterized.
+- Insert the verbose `debug_bid` blob **only** when the testing flag is set. The
+  single `include_adm` param becomes `include_debug_bid`.
+
+`hb_cache_host`/`hb_cache_path` remain inserted unconditionally.
+
+### 2. Bridge renders local `adm`; cache is the fallback for an _absent_ `adm`
+
+`installTsRenderBridge` already prefers `matchedBid.adm` and falls back to PBS
```
aram356, line 111: 🔧 **Blocking: decode a cached bid before passing it to PUC.**

The current fallback does `res.text()` and sends that entire string as the PUC response's `ad`. PBS Cache bid entries are cached bid objects, and PUC's own cache path parses the returned JSON and extracts `bidObject.adm`. The test currently mocks the cache response as raw `<div>…</div>` markup, so it does not exercise the production payload shape. See the [Prebid caching contract](https://docs.prebid.org/prebid-server/features/pbs-caching.html) and [PUC's cache renderer](https://github.com/prebid/prebid-universal-creative/blob/1.18.0/src/mobileAndAmpRender.js).

When `adm` is absent—exactly when this fallback is needed—the bridge can therefore send serialized JSON to be rendered as the ad. Please parse the cached bid response, validate/extract its string `adm` (optionally retaining explicitly tested raw-markup compatibility), and add a realistic `returnCreative=false` cache-response test plus malformed/missing-`adm` handling.

docs/superpowers/specs/2026-07-13-ssat-render-inline-creative-design.md
```diff
@@ -0,0 +1,164 @@
...
+Cache. Once `adm` is present in production, the local render becomes the default
+and the round trip disappears.
+
+**Fallback scope (corrected):** the bridge posts the markup to the PUC and
+returns; it receives **no render-success signal**. So the PBS Cache fallback
```
aram356, line 125: 🤔 **Non-blocking: pin and test the actual PUC protocol.**

PUC 1.18.0 emits `Prebid Event` messages for `adRenderSucceeded` and `adRenderFailed`, so the protocol does expose a signal even though this bridge does not currently consume it. Its dynamic-renderer path also requires `rendererVersion >= 3`; the response here supplies `renderer` without a version, so that renderer is ignored by this PUC version. See [`messaging.js`](https://github.com/prebid/prebid-universal-creative/blob/1.18.0/src/messaging.js), [`renderingManager.js`](https://github.com/prebid/prebid-universal-creative/blob/1.18.0/src/renderingManager.js), and [`dynamicRenderer.js`](https://github.com/prebid/prebid-universal-creative/blob/1.18.0/src/dynamicRenderer.js).

Please confirm and pin the PUC version/template used by GAM, add a browser contract test against that real bundle, and either send the renderer protocol it expects or remove the unused renderer field. The render-failure and beacon-timing claims in this design should then be updated accordingly.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -593,10 +593,13 @@ export function installTsAdInit(): void {
...
-          // Adapted from PR #241 — uses window.tsjs.bids[slotId].adm instead of pbjs.
-          // Only active when inject_adm_for_testing injects adm into bids server-side.
-          if (bid.adm) {
+          // GAM interceptor (testing bypass): directly replace the GAM creative.
+          // `adm` is now always injected in production, so it can no longer gate
+          // this path. `debug_bid` is present only when inject_adm_for_testing is
+          // on, so it is the per-bid signal that the testing bypass is enabled.
+          // In production the render bridge serves the creative and GAM stays in
+          // the loop; this direct replace stays testing-only.
+          if (bid.adm && bid.debug_bid) {
```
aram356, line 606: 👍 **Non-blocking:** Using the current bid's `debug_bid` as the testing-bypass gate is a good choice. It keeps production GAM rendering intact now that `adm` is always present and avoids a stale global flag across SPA navigations; the new tests cover both branches.

aram356 (changes requested):
## Review summary

The inline `adm` path is a useful latency improvement, and the per-bid debug gate is a good way to avoid stale SPA-global state. I am requesting changes for two production contract gaps before this ships:

1. The new SSAT path bypasses the existing creative sanitization, URL-rewrite, and size-limit boundary.
2. The advertised PBS Cache fallback forwards a cached bid JSON document as creative markup instead of extracting its `adm`.

## Additional non-blocking finding

♻️ **Update the public contracts to match the new semantics.** Several comments outside this diff still say `adm` is present only under `inject_adm_for_testing`: [`Settings::inject_adm_for_testing`](https://github.com/IABTechLab/trusted-server/blob/1159407f05b905359fd1ff224a3ac43e4126c6f8/crates/trusted-server-core/src/settings.rs#L1911-L1918), [`AuctionBidData.adm`](https://github.com/IABTechLab/trusted-server/blob/1159407f05b905359fd1ff224a3ac43e4126c6f8/crates/trusted-server-js/lib/src/core/types.ts#L60-L63), and the GPT integration comments at [lines 279–285](https://github.com/IABTechLab/trusted-server/blob/1159407f05b905359fd1ff224a3ac43e4126c6f8/crates/trusted-server-js/lib/src/integrations/gpt/index.ts#L279-L285) and [825–835](https://github.com/IABTechLab/trusted-server/blob/1159407f05b905359fd1ff224a3ac43e4126c6f8/crates/trusted-server-js/lib/src/integrations/gpt/index.ts#L825-L835). Please update them to distinguish production inline `adm` from the testing-only direct GAM replacement gated by `debug_bid`.

## Verification

- Reviewed all five changed files against base `0e5dbb211da5e61598681df3abfda239484afa5f`.
- Current GitHub CI is green (19 reported checks).
- Focused GPT Vitest suite: 26/26 passing.
- Focused Rust regressions: 2/2 passing.
- `git diff --check`: clean.
- Local `cargo test-fastly` compiled, but Viceroy could not access the macOS certificate keychain (`No keychain is available`); the corresponding GitHub Fastly job passes, so this is a local environment limitation rather than a PR failure.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Fixed in 1c983027c. `build_bid_map` now runs the same boundary as the `/auction` path — `sanitize_creative_html` then `rewrite_creative_html` — before inserting `adm`, so the 1 MiB `MAX_CREATIVE_SIZE` cap and first-party URL rewriting both apply to the inline path. When sanitization rejects the creative (oversized or unparseable), `adm` is omitted and the bridge falls back to the PBS Cache coordinates. Threaded `&Settings` through `build_bid_map` and `write_bids_to_state` for `rewrite_creative_html`. New coverage: `build_bid_map_sanitizes_hostile_adm` (asserts `<script>`/`on*`/`javascript:` are stripped) and `build_bid_map_omits_oversized_adm` (adm omitted above the cap).

docs/superpowers/specs/2026-07-13-ssat-render-inline-creative-design.md
prk-Jr: Fixed in 1c983027c. The fallback no longer forwards `res.text()` verbatim. A new `extractCachedAdm()` `JSON.parse`s the cache response and extracts the string `adm` (matching PUC's `bidObject.adm`), retains raw-markup compatibility for caches that return the creative directly, and declines to render — no `Prebid Response`, no win/billing beacons — when the payload carries no usable `adm`, rather than shipping a serialized bid document to PUC. Tests: a realistic `returnCreative=false` JSON payload (asserts the extracted `adm` is forwarded), a missing-`adm` decline, and a non-JSON raw-markup compatibility case.

docs/superpowers/specs/2026-07-13-ssat-render-inline-creative-design.md
prk-Jr: Deferring this to a follow-up rather than guessing. Pinning the `rendererVersion`/template and adding a browser contract test needs the actual PUC bundle GAM serves in production, which I can't verify from here — and sending an unconfirmed `rendererVersion` or dropping the `renderer` field could break rendering on whatever version is live. The design spec already documents that the bridge consumes no `adRenderSucceeded`/`adRenderFailed` signal and that fallback fires only on absent/empty inline `adm`. Happy to open a tracking issue to confirm the GAM-served PUC version and then reconcile the `renderer` field + render-failure claims — let me know and I'll file it.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -1983,12 +1987,27 @@ pub(crate) fn build_bid_map(
...
+                //
+                // Run the same creative-processing boundary as the `/auction`
+                // path (see `auction::formats`): sanitize dangerous markup first,
+                // then rewrite URLs to first-party proxies. `sanitize_creative_html`
+                // also enforces the 1 MiB creative cap, returning an empty string
+                // for oversized or unparseable markup — in which case the entry is
+                // omitted and the bridge falls back to the PBS Cache coordinates.
+                if let Some(ref raw_creative) = bid.creative {
+                    let sanitized = crate::creative::sanitize_creative_html(raw_creative);
+                    let adm = crate::creative::rewrite_creative_html(settings, &sanitized);
```
aram356: 🔧 **Blocking: the rewritten inline `adm` emits root-relative URLs that will not resolve where the bridge renders it.**

Sanitization here is right, and the 1 MiB cap now applies — thank you. But `rewrite_creative_html` was built for creatives rendered from the **first-party origin**, and reusing it on this path changes what ships to the client. Running a typical creative through `sanitize_creative_html` → `rewrite_creative_html` produces:

```html
<script src="/static/tsjs=tsjs-unified.min.js?v=5e45a1c0…" id="trustedserver-js"></script>
<img src="/first-party/proxy?tsurl=https%3A%2F%2Fcdn.example-ssp.com%2Fcreative.png&tstoken=…">
<iframe src="/first-party/proxy?tsurl=https%3A%2F%2Fcdn.example-ssp.com%2Ff.html&tstoken=…">
```

Every URL is **root-relative** — `build_signed_url_for` returns `format!("{}?{}", base_path, …)` with `base_path = "/first-party/proxy"`, and the injected script tag is `/static/tsjs=…`.

That matters because this same `adm` string is consumed in two different origin contexts:

- `injectAdmIntoSlot` (the testing bypass, gated on `debug_bid`) injects into the publisher's own div — first-party, so root-relative resolves. **This is the only path exercised in local dev.**
- `installTsRenderBridge` hands it to PUC, which renders it via `f.srcdoc=d.ad` (`TS_DISPLAY_RENDERER`, `index.ts:826`). A `srcdoc` document's base URL is inherited from its browsing-context container's document — GAM's origin, not the publisher's. The proxied URLs and the injected script then resolve against GAM's origin and 404: broken creative, no impression. This holds whether or not PUC honors the custom `renderer` (see the earlier `rendererVersion >= 3` note), because PUC's own fallback also writes into the GAM-origin document.

Before `1c983027c` the inline `adm` was raw with absolute SSP URLs, which render in any origin. As written, the fix trades the XSS gap for a render regression in the exact production path this PR exists to add.

To be explicit about what I have **not** verified: if GAM renders the Prebid line item in a same-origin *friendly* iframe, the base URL is the publisher page and these URLs resolve fine. Which case applies is publisher/SafeFrame-dependent. I am raising this as blocking rather than asserting it is broken, because the downside is "no creative renders in production", the bridge hooks PUC's `"Prebid Request"` postMessage protocol (its cross-domain path), and this PR's own test plan notes the end-to-end inline render was never exercised — mocktioneer's bid loses to GAM's demand, so the bridge never fired.

**Fix** — emit absolute first-party URLs for the inline path (`settings.publisher.domain` is available server-side), or have `TS_DISPLAY_RENDERER` prepend `<base href="https://<first-party>/">` to the srcdoc before the ad. The server-side absolute URLs are preferable — they do not depend on the renderer being honored. Worth confirming separately that first-party proxy requests issued from GAM's iframe are a third-party cookie context, so EC cookies may not attach as they would first-party.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -1983,12 +1987,27 @@ pub(crate) fn build_bid_map(
...
+                // remain as the fallback for an absent `adm`.
+                //
+                // Run the same creative-processing boundary as the `/auction`
+                // path (see `auction::formats`): sanitize dangerous markup first,
+                // then rewrite URLs to first-party proxies. `sanitize_creative_html`
+                // also enforces the 1 MiB creative cap, returning an empty string
+                // for oversized or unparseable markup — in which case the entry is
+                // omitted and the bridge falls back to the PBS Cache coordinates.
+                if let Some(ref raw_creative) = bid.creative {
+                    let sanitized = crate::creative::sanitize_creative_html(raw_creative);
```
aram356: 🤔 **Non-blocking: this also injects the full tsjs bundle into every creative.**

Beyond URL rewriting, `rewrite_creative_html` prepends `<script src="/static/tsjs=tsjs-unified.min.js?v=…" id="trustedserver-js">` to the creative's `<body>` — it is documented as safeguarding click URLs inside first-party proxied creatives. On this path that ships core plus every integration bundle into each ad creative iframe, which I suspect is not intended here.

It is currently moot because that script tag is root-relative and 404s (see the blocking comment), but it will become live once that is fixed. Consider a rewrite variant that performs the sanitize/URL-rewrite work without the body-script injection for the inline path.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -3940,8 +3960,16 @@ mod tests {
         };
         use crate::http_util::RequestInfo;
         use crate::price_bucket::PriceGranularity;
+        use crate::settings::Settings;
         use std::collections::HashMap;
 
+        // Default settings are enough for the creative boundary: the sanitize
+        // pass needs no config, and `rewrite_creative_html` only signs URLs it
```
aram356, line 4374: ♻️ **Non-blocking: nothing asserts the rewrite half of the boundary.**

`build_bid_map_sanitizes_hostile_adm` and `build_bid_map_omits_oversized_adm` prove sanitization and the cap — both good additions. But `test_settings()` is `Settings::default()`, and this comment concedes the fixtures carry no proxyable URLs, so `rewrite_creative_html` never actually rewrites anything in any test. The second half of the boundary the last review asked for is therefore unproven, which is why the blocking finding above ships CI-green across all 19 checks.

A fixture with a proxyable absolute URL (e.g. ``) asserting the shape of the emitted `adm` URL would both close the gap and pin down the render-context decision.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -921,7 +960,18 @@ export function installTsRenderBridge(): void {
 
     fetch(cacheUrl, { mode: 'cors' })
       .then((res) => (res.ok ? res.text() : Promise.reject(res.status)))
-      .then((ad) => {
+      .then((body) => {
+        // PBS Cache returns the cached bid as a JSON object; extract its `adm`
+        // the same way the Prebid Universal Creative does before rendering.
+        const ad = extractCachedAdm(body);
+        if (!ad) {
```
aram356: 🤔 **Non-blocking: declining here leaves the slot blank rather than failing fast.**

`e.stopImmediatePropagation()` has already fired at line 952, so returning without a `port.postMessage` means PUC keeps waiting on the MessageChannel and Prebid.js was suppressed from handling it — the slot stays empty. The `.catch` path below has the same shape and predates this PR, so this is not a regression, and declining is clearly better than shipping a serialized bid document to PUC.

Still, consider replying with something PUC can act on so it can fail fast or collapse the slot. This ties into the PUC-protocol question deferred to a follow-up: PUC 1.18.0 does emit `adRenderFailed`, so there is a signal available once the version is pinned.

aram356 (changes requested):
## Summary

Re-reviewed against base `0e5dbb211`, focusing on `1c983027c` and the two blocking findings from the last round.

The PBS Cache decode finding is resolved: `extractCachedAdm` mirrors PUC's `bidObject.adm` extraction, keeps raw-markup compatibility, declines on malformed payloads, and the new tests use realistic `returnCreative=false` shapes. The doc updates also resolve the earlier contract-drift finding.

The sanitize/rewrite finding is only half resolved. Sanitization and the 1 MiB cap are correctly applied, but `rewrite_creative_html` was designed for creatives rendered from the first-party origin, and reusing it here rewrites the inline `adm` into root-relative URLs that will not resolve in the context the pbRender bridge hands them to. Details inline on `publisher.rs:2003`.

## Blocking

### 🔧 wrench

- **Rewritten inline `adm` emits root-relative URLs that will not resolve in PUC's render context** (`crates/trusted-server-core/src/publisher.rs:2003`)

## Non-blocking

### 🤔 thinking

- **The rewriter injects the full tsjs bundle into every creative** (`crates/trusted-server-core/src/publisher.rs:2002`)
- **Declining after `stopImmediatePropagation()` leaves the slot blank** (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts:967`)
- **The 1 MiB cap is per-creative, not aggregate.** `write_bids_to_state` inlines the creative for every winning slot into the page's `<script>` on every page load, so N slots can inline up to N MiB with no aggregate bound. Per this PR's own test plan, GAM usually awards its own demand and the bridge never fires — in that common case the inlined bytes are pure document-weight/TTFB cost. Worth considering an aggregate cap, or a narrower condition for inlining.

### ♻️ refactor

- **No test covers the rewrite half of the boundary** (`crates/trusted-server-core/src/publisher.rs:3967`)

### 📌 out of scope

- **Please do file the PUC tracking issue you offered.** Deferring the PUC version/template pin was reasonable on its own terms, but the blocking finding above makes it load-bearing: the render context that pin would establish is exactly what determines whether the inline `adm`'s rewritten URLs resolve. Confirming the GAM-served PUC bundle and adding a browser contract test against it would settle both that finding and the earlier `rendererVersion`/`renderer`-field question.

## CI Status

- fmt: PASS (GitHub)
- clippy: PASS (GitHub)
- rust tests: PASS (GitHub; `cargo test`, axum native, cloudflare, spin, parity, CLI all green). Local `publisher::tests` 84/84 pass.
- js tests: PASS (GitHub vitest; local GPT suite 77/77 pass)
- All 19 reported checks are green.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Resolved in d0b18b402. The inline path now uses `rewrite_inline_creative_html`, which emits absolute first-party URLs — `https://<settings.publisher.domain>/first-party/proxy?tsurl=…` and `/first-party/click` — instead of root-relative ones. They resolve against the publisher origin regardless of the `srcdoc` document base URL, so they hold whether or not PUC honors the custom renderer (its fallback writes into the same GAM-origin document). The first-party `/auction` path keeps the root-relative rewriter (`rewrite_creative_html`), which is correct there.

The third-party-cookie question for proxy requests issued from GAM's iframe is tracked in #926, alongside pinning/browser-testing the actual PUC contract.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Resolved in d0b18b402. `rewrite_inline_creative_html` skips the `<body>` tsjs bundle injection (`inject_tsjs = false`); the click-URL safeguard it provided is unnecessary now that click URLs are absolute. The `/auction` path (`rewrite_creative_html`) keeps the injection — the two share one rewriter behind a single `inject_tsjs` knob.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Added in d0b18b402:
- `build_bid_map_rewrites_inline_adm_to_absolute_first_party_urls` (publisher.rs) — sets `publisher.domain`, feeds a creative with a proxyable absolute ``, and asserts the emitted `adm` contains `https://example.com/first-party/proxy?tsurl=`, contains no root-relative `src="/first-party/proxy`, proxies the original CDN URL, and carries no tsjs bundle.
- `inline_rewrite_emits_absolute_urls_and_omits_tsjs` and `inline_rewrite_preserves_relative_urls` (creative.rs) pin the function-level contract (absolute proxy + click URLs; relative URLs untouched).

So the rewrite half of the boundary is now exercised with a proxyable fixture rather than URL-free markup.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
prk-Jr: A fail-fast signal is the right call, and it depends on pinning the PUC version first. Tracked in #926: once the version is pinned, reply to a declined "Prebid Request" with `adRenderFailed` (PUC 1.18.0) for both the inline-adm and PBS Cache decline paths so the slot collapses instead of staying blank. Leaving the current decline as-is here since it is not a regression and declining still beats shipping a serialized bid document to PUC.

docs/superpowers/specs/2026-07-13-ssat-render-inline-creative-design.md
prk-Jr: Filed #926. It pins the GAM-served PUC version/template, adds a browser contract test exercising the end-to-end inline render (mocktioneer winning so the bridge actually fires), and settles the `rendererVersion >= 3` / custom-renderer question. It also carries the third-party-cookie check and the aggregate inline-`adm` cap raised in the summary.

The absolute-URL fix in d0b18b402 makes the inline render independent of whether the custom renderer is honored; the browser test in #926 will confirm it end to end.

crates/trusted-server-core/src/creative.rs
```diff
@@ -492,14 +515,53 @@ pub fn sanitize_creative_html(markup: &str) -> String {
...
+/// Differs from [`rewrite_creative_html`] in the two ways that context requires:
+/// - Proxy/click URLs are emitted **absolute** (`https://<publisher-domain>/first-party/…`)
+///   so they resolve against the publisher origin regardless of the document's
+///   base URL, and independently of whether the custom renderer is honored.
+/// - The `tsjs` bundle is **not** injected into `<body>`: its only job is to
+///   safeguard click URLs, which are already absolute here, and shipping the full
+///   core-plus-integrations bundle into every creative iframe is pure weight.
+#[must_use]
+pub fn rewrite_inline_creative_html(settings: &Settings, markup: &str) -> String {
+    let base_origin = format!("https://{}", settings.publisher.domain);
```
ChristianPavilonis: 🔧 **High: build inline URLs from the actual request origin.**

This hard-codes `https://{settings.publisher.domain}`, discarding the incoming scheme, subdomain, and port. `publisher.domain` cannot contain a port, deployments may receive traffic on a subdomain, and Axum/Viceroy development runs over HTTP with a port. Affected creatives therefore request proxy/click endpoints from the wrong origin, with no render-time fallback.

Please pass the trusted request origin through `build_bid_map` and `write_bids_to_state`. Initial navigation already retains `request_scheme`/`request_host`, and SPA page-bids already has `RequestInfo`. Add coverage for HTTP localhost with a port, a request subdomain differing from `publisher.domain`, and a non-default HTTPS port.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -1983,12 +1987,33 @@ pub(crate) fn build_bid_map(
...
+                // against GAM and 404. `rewrite_inline_creative_html` emits
+                // absolute first-party URLs and omits the tsjs bundle injection.
+                // `sanitize_creative_html` also enforces the 1 MiB creative cap,
+                // returning an empty string for oversized or unparseable markup —
+                // in which case the entry is omitted and the bridge falls back to
+                // the PBS Cache coordinates.
+                if let Some(ref raw_creative) = bid.creative {
+                    let sanitized = crate::creative::sanitize_creative_html(raw_creative);
+                    let adm = crate::creative::rewrite_inline_creative_html(settings, &sanitized);
+                    if !adm.is_empty() {
```
ChristianPavilonis, line 2226: 🔧 **High: include the winning creative dimensions with production `adm`.**

The bridge currently sizes every response from `slot.formats[0]`, while the actual winning `Bid.width`/`height` are emitted only inside testing-only `debug_bid`. For a multi-size slot, a 300×250 winner can therefore be rendered as another configured size, causing clipping, whitespace, or broken layout.

Please emit winning `width` and `height` in the normal bid map and `AuctionBidData`, then prefer those values in the inline response. A regression where the winner matches the second configured format would pin the contract.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -1983,12 +1987,33 @@ pub(crate) fn build_bid_map(
...
+                // Universal Creative inside GAM's iframe (`f.srcdoc = d.ad`), a
+                // foreign origin where root-relative `/first-party/…` URLs resolve
+                // against GAM and 404. `rewrite_inline_creative_html` emits
+                // absolute first-party URLs and omits the tsjs bundle injection.
+                // `sanitize_creative_html` also enforces the 1 MiB creative cap,
+                // returning an empty string for oversized or unparseable markup —
+                // in which case the entry is omitted and the bridge falls back to
+                // the PBS Cache coordinates.
+                if let Some(ref raw_creative) = bid.creative {
+                    let sanitized = crate::creative::sanitize_creative_html(raw_creative);
```
ChristianPavilonis: 🔧 **High: resolve `${AUCTION_PRICE}` before sanitizing, rewriting, and signing.**

The new inline path forwards `adm` without PUC-equivalent auction-price preparation. URL rewriting serializes query pairs, turning `${AUCTION_PRICE}` into `%24%7BAUCTION_PRICE%7D`; the JS bridge then forwards the markup unchanged. Tracking or billing URLs can consequently receive a literal/encoded macro instead of the clearing price, and substitution after signing would invalidate the signed value.

Please expand the macro from the exact winning CPM before `sanitize_creative_html` and `rewrite_inline_creative_html`. Add tests for both ordinary markup and an absolute URL query containing the macro, asserting neither literal nor encoded macro survives.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -819,6 +826,36 @@ const TS_DISPLAY_RENDERER =
...
+ * renders `bidObject.adm`, not the raw response body. This mirrors that
+ * extraction so the bridge hands PUC a creative string rather than a serialized
+ * bid document.
+ *
+ * A non-JSON body is treated as raw creative markup for backward compatibility
+ * with caches that store the creative directly. Returns `undefined` when the
+ * JSON payload carries no usable string `adm` (missing or malformed), so the
+ * caller can decline to render instead of injecting a serialized object.
+ */
+export function extractCachedAdm(body: string): string | undefined {
```
ChristianPavilonis: 🤔 **Medium: preserve the cached bid fields needed for rendering.**

Reducing the PBS Cache object to `adm` discards its price, `width`/`height` or `w`/`h`, and `wurl`. The fallback then uses the first slot format and performs no cached-price macro preparation. PUC's cache path uses those fields, so fallback creatives can be mis-sized, retain unresolved price macros, or omit the cached win notification.

Please decode a validated cached-bid structure rather than only a string, retaining price, dimensions, and notification metadata. Prepare the creative before responding and deduplicate notifications; keep raw-markup compatibility as a separate response variant. A realistic cache-object test should cover these fields.

docs/superpowers/specs/2026-07-13-ssat-render-inline-creative-design.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Design
+
+### 1. Always include the render `adm`; keep `debug_bid` gated
+
+`build_bid_map`:
+
+- **Always** insert `adm` for a winner when present — there is no runtime reason
+  to withhold it, so it is not parameterized. The creative is first run through
+  the **same sanitize/rewrite boundary as the `/auction` path** (`auction::formats`):
```
ChristianPavilonis, line 96: 📝 **Low: reconcile the design and implementation plan with the final implementation.**

The spec still names `rewrite_creative_html`, although the code now relies on `rewrite_inline_creative_html` with absolute URLs and no TSJS injection. The accompanying plan also omits changed components, shows obsolete signatures/raw-`adm` insertion, and leaves every task unchecked.

Please update the design to describe the final inline rewriter and render metadata requirements, and reconcile the plan as completed or mark it superseded.

ChristianPavilonis (commented):
## Summary

Re-reviewed the current head after removing the earlier first-party iframe-proxy blocker. The previous raw-creative, cache-JSON, root-relative URL, and unwanted TSJS-injection findings are resolved. CI is green, but the new default inline-render path still has three high-confidence compatibility/correctness gaps, plus one cache-fallback contract gap and stale design records. Details are inline.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -2128,12 +2186,42 @@ pub(crate) fn build_bid_map(
...
+                // `sanitize_creative_html` also enforces the 1 MiB creative cap,
+                // returning an empty string for oversized or unparseable markup —
+                // in which case the entry is omitted and the bridge falls back to
+                // the PBS Cache coordinates.
+                if let Some(ref raw_creative) = bid.creative {
+                    // Resolve ${AUCTION_PRICE} from the exact winning CPM BEFORE
+                    // sanitizing, rewriting, and signing — URL rewriting would
+                    // otherwise encode the literal macro into the signed proxy/click
+                    // URL, and signing would lock that wrong value.
+                    let priced = crate::creative::expand_auction_price_macro(raw_creative, cpm);
```
aram356, line 2219: 🔧 **Blocking: `nurl`/`burl` are emitted with an unexpanded `${AUCTION_PRICE}`.**

Expanding the macro before sanitize/rewrite/sign is the right call, and the cached-price expansion in the JS fallback mirrors it well. But the expansion is applied only to `bid.creative`. The `nurl` and `burl` notification URLs are inserted raw a few lines above (`:2183-2188`) and fired verbatim by `fireWinBillingBeacons` → `queueWinBillingBeacon`. Per OpenRTB, those two fields are the canonical place `${AUCTION_PRICE}` appears — the win notice is the substitution's primary consumer.

I verified this rather than inferring it, with a temporary probe through `build_bid_map` on a bid whose `nurl`/`burl` carry the macro (probe since reverted):

```
PROBE_NURL=https://ssp.example/win?p=${AUCTION_PRICE}
PROBE_BURL=https://ssp.example/bill?p=${AUCTION_PRICE}
```

The `adm` on that same bid was expanded correctly — so the two fields disagree about the clearing price within a single response.

**Why this is in scope rather than pre-existing.** In the base, the bridge's inline branch already called `fireWinBillingBeacons`, but that branch only ran when `adm` was present, which pre-PR meant `inject_adm_for_testing` only. This PR ships `adm` unconditionally and so promotes that branch to the primary production path. Beacons that previously fired only under the testing flag now fire on every TS-won impression, each carrying a literal macro to the SSP — an unresolved clearing price in win/billing notifications, which is a revenue-reporting correctness issue and is the kind of malformed notification some SSPs reject outright.

**Fix** — expand from the same `cpm` already in hand at this scope, before inserting either field:

```rust
if let Some(ref nurl) = bid.nurl {
    let nurl = crate::creative::expand_auction_price_macro(nurl, cpm);
    obj.insert("nurl".to_string(), serde_json::Value::String(nurl));
}
if let Some(ref burl) = bid.burl {
    let burl = crate::creative::expand_auction_price_macro(burl, cpm);
    obj.insert("burl".to_string(), serde_json::Value::String(burl));
}
```

Worth a regression asserting no literal or encoded macro survives in either field, alongside the existing `adm` one. Two adjacent cases to consider while you are here: the `debug_bid` mirror carries raw `nurl`/`burl` too (fine, it is diagnostic, but worth a deliberate decision), and the PBS Cache fallback in the bridge expands the cached `adm` from `cached.price` while still firing `matchedBid`'s unexpanded beacons.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -1351,6 +1379,10 @@ async fn emit_abandoned_auction(
     .await;
 }
 
+// Private orchestration helper called only from `body_close_hold_loop`, whose
+// arguments mirror the fields of `AuctionCollectCtx` it destructures; a separate
```
aram356: ♻️ **Non-blocking: the parameter struct this defers to already exists.**

The commit message reasons that a parameter struct "would just duplicate that context" — but `AuctionCollectCtx` (`:1083`) already holds exactly these eight fields, and the caller destructures it at `:1213` only to pass the pieces straight back in. So the struct is not hypothetical; it is being unpacked and re-threaded.

The real constraint is narrower than the message suggests: `dispatched` and `telemetry.take()` are moved per loop iteration, so the whole context cannot be passed by value twice. That still leaves a clean shape — borrow the context and pass the two moved values explicitly:

```rust
async fn collect_stream_auction(
    dispatched: DispatchedAuction,
    telemetry: AuctionTelemetryCarry,
    ctx: &AuctionCollectCtx<'_>,
)
```

Three arguments, no `allow`, and the request-origin threading that triggered the lint stops widening the signature. CLAUDE.md's "never more than 7 arguments — use a struct instead" points the same way. Not blocking, and the `allow` is honestly documented if you would rather keep it.

aram356 (changes requested):
## Summary

Re-reviewed at `f4e0a79fa` against base `0f8f93921`, focusing on the work since `1c983027c`.

Both of my earlier blocking findings are resolved, and the four raised in the interim review land correctly. One new blocking issue: this PR introduces `${AUCTION_PRICE}` expansion for the creative but not for the `nurl`/`burl` notification URLs, which is where that macro most commonly appears — and this PR is what promotes those beacons from a testing-only path to the primary production one.

## Blocking

### 🔧 wrench

- **`nurl`/`burl` fire with an unexpanded `${AUCTION_PRICE}`** (`crates/trusted-server-core/src/publisher.rs:2183-2188`, expansion at `:2211`)

## Non-blocking

### ♻️ refactor

- **`#[allow(clippy::too_many_arguments)]` sidesteps a struct that already exists** (`crates/trusted-server-core/src/publisher.rs:1383`)

### 📝 note

- **Earlier findings verified as resolved.** `rewrite_inline_creative_html` genuinely fixes the foreign-origin URL problem. I checked the part that could have silently broken: `build_signed_url_for` computes the token over `tsurl` + params only and excludes `base_path`, so absolute inline URLs validate through the proxy identically to root-relative ones. `base_origin` derives from `RequestInfo.host`, already the established origin source for URL rewriting across the codebase, so it is not a new input surface. The dimension, origin-threading, macro-ordering, and cache-decode fixes all check out, and the slot-scoped bid lookup correctly resolves the duplicate-`hb_adid` collision. Thanks for filing #926 with the PUC pin, the third-party-cookie check, and the aggregate-`adm` cap.

## CI Status

- fmt: PASS (GitHub)
- clippy: PASS (GitHub)
- rust tests: PASS — 19/19 GitHub checks green; local `trusted-server-core` lib suite 1659/1659 pass
- js tests: PASS — local full Vitest run 425/425 pass across 30 files

crates/trusted-server-core/src/publisher.rs
prk-Jr: Fixed in 2ad899fbe. Both fields now expand from the same `cpm` before insertion:

```rust
if let Some(ref nurl) = bid.nurl {
    let nurl = crate::creative::expand_auction_price_macro(nurl, cpm);
    obj.insert("nurl".to_string(), serde_json::Value::String(nurl));
}
```

Regression added — `build_bid_map_expands_auction_price_macro_in_notification_urls` asserts no literal or encoded `AUCTION_PRICE` survives in either field and that both resolve to the exact CPM, alongside the existing adm test.

On the two adjacent cases:

**`debug_bid` mirror** — keeping it verbatim, now stated in the code as a deliberate call. Nothing renders or fires from that blob, and it only exists under `inject_adm_for_testing`; showing what the bidder actually sent is its purpose, so expanding it would cost the raw view without changing any outbound request. Checked that the client only reads `bid.debug_bid` as a presence gate for the testing-only GAM-replace path and never touches its `nurl`/`burl`.

**PBS Cache fallback** — resolved by this same server-side change rather than a client one. That branch fires `matchedBid`'s beacons, which now arrive already expanded from the auction clearing price, so the disagreement is gone. Left a note at the fire site that the auction result is the authoritative price and `cached.price` is only the render source, so the expansion does not get duplicated client-side later.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Right on both counts — the struct exists and the constraint is only that `dispatched` and `telemetry.take()` are moved per iteration. Allow removed in 2ad899fbe.

One deviation from the exact shape you sketched: the borrowed third argument cannot also own the two moved values, so `AuctionCollectCtx` is split rather than passed by reference whole —

```rust
struct AuctionCollectCtx<'a> {
    dispatched: DispatchedAuction,
    telemetry: AuctionTelemetryCarry,
    deps: AuctionCollectDeps<'a>,
}

async fn collect_stream_auction(
    dispatched: DispatchedAuction,
    telemetry: AuctionTelemetryCarry,
    deps: &AuctionCollectDeps<'_>,
)
```

Three arguments and no allow, as proposed. Splitting rather than adding parameters upstream also keeps `stream_html_with_auction_hold` and `body_close_hold_loop` at their existing arity, so request-origin threading stops widening anything.

aram356 (approved):

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -1041,6 +1041,18 @@ impl IntegrationHeadInjector for PrebidIntegration {
     }
 }
 
+/// Returns `true` when `params` is not a usable PBS bidder-params object — an
```
aram356, line 1044: 📌 **Non-blocking (out of scope): #910 is bundled into this PR via a stacked-branch merge.**

`Suppress fabricated empty Prebid bidder params (#910)` was opened against `feat/ssat-render-inline-creative` — this PR's own head branch — and merged into it, so GitHub counts this file (+405/−24) as part of PR 899's diff even though it is unrelated to inline SSAT creative rendering. When 899 merges, #910's content rides in with it under this PR's changelog entry rather than its own.

I reviewed the change on its merits since it is in scope, and it is sound: the merge/drop logic is deterministic regardless of `slot.bidders` HashMap order, the override pass correctly runs before the drop so a rule-populated fabricated-empty survives, the residual param-loss edge is honestly documented, and it ships eight new tests including an order-independence loop. No defect — this is only a note that two unrelated concerns landed in one PR, in case that bundling was not deliberate.

aram356 (approved):
## Summary

Approving. Both blocking findings from the earlier rounds are resolved and verified at `07890a99`; CI is 19/19 green, and locally 1671 Rust core tests and 425 JS tests pass.

## Resolved since the last review

- **🔧 `nurl`/`burl` `${AUCTION_PRICE}` expansion** — fixed in `2ad899fbe`. Both fields now expand from the same `cpm` as the `adm`, so the price disagreement I flagged is gone, not merely hidden. `build_bid_map_expands_auction_price_macro_in_notification_urls` asserts no literal or encoded macro survives and both resolve to the exact CPM.
- **♻️ `too_many_arguments` allow** — fixed in `2ad899fbe`. `AuctionCollectCtx` was split into moved per-auction state plus a borrowed `AuctionCollectDeps`, returning `collect_stream_auction` to three arguments with no `allow`.

## Non-blocking

### 📌 out of scope

- **#910 is bundled into this PR** (`crates/trusted-server-core/src/integrations/prebid.rs`) — details inline.

## CI Status

- fmt / clippy / rust tests / js tests: PASS (19/19 GitHub checks green)
- Local: `trusted-server-core` lib 1671/1671, full Vitest 425/425

Note: `cargo clippy-fastly` fails locally with a `clippy-driver` "Unrecognized option: 'p'" misparse of the unchanged `trusted-server-js/build.rs` (edition-2024 build script) — an environment artifact on my toolchain, not a PR defect; CI clippy is green on this commit.

PR #900, opened 2026-07-13
Avoid no-op EC KV reads in post-send pull sync

### LINKED ISSUE #880: Avoid no-op EC KV reads in post-send pull sync
## Summary

Avoid EC identity-graph KV reads when post-send pull sync has no work to perform.

`dispatch_pull_sync` currently reads the EC row before it checks whether any partners have pull sync enabled. When pull-enabled partners do exist, it reads the row on every eligible real-browser response even when every partner UID is already present. The work happens after response delivery, so it does not gate response bytes, but it still consumes KV operations and extends Compute instance time.

## Proposed direction

1. Resolve pull-enabled partners before reading KV and return immediately when the set is empty.
2. Add a bounded completeness mechanism, such as a short-lived browser marker, so fully synced users can skip no-op KV checks.
3. Preserve the current consent gate, missing-UID-only behavior, partner rate limits, and best-effort post-send failure policy.

## Out of scope

Request-scoped reuse of `(KvEntry, generation)` across auction, finalize, and pull sync belongs exclusively to #851. This issue must not introduce another request-scoped EC cache or duplicate that work.

## Acceptance criteria

- No pull-enabled partners results in zero EC KV operations.
- A user known to be complete for the current pull-partner set can skip the initial EC KV read independently of request-scoped reuse from #851.
- Missing, expired, or invalid completeness state falls back safely to KV.
- Partner configuration changes cannot leave a user permanently marked complete against the old partner set.
- Incomplete users continue to dispatch only eligible partners and persist returned UIDs with CAS protection.
- Unit tests cover no-partner, complete, stale-marker, partial, and KV-failure cases.

## Relevant code

- `crates/trusted-server-core/src/ec/pull_sync.rs`

ChristianPavilonis:
## Summary

- Skip post-send identity-graph construction when pull sync has no eligible work.
- Add a signed, one-hour browser completeness marker bound to the active EC ID and current pull-partner set.
- Preserve shared snapshot reads required by auctions, EID ingestion, and withdrawal while bounding marker-delayed orphan detection to the one-hour marker lifetime.

This PR builds on #885, which is now in `main`. The zero-operation guarantee is scoped to KV work caused solely by pull sync; other identity lifecycle consumers may still require the shared EC snapshot. A valid marker acts as recent existence proof, so deletion of a previously complete row is detected after marker expiry rather than on every marked request.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/ec/pull_sync_marker.rs` | Add the signed marker protocol, partner-set fingerprinting, cookie lifecycle, and focused tests |
| `crates/trusted-server-core/src/ec/{mod,finalize,pull_sync}.rs` | Carry marker state, reconcile it before response delivery, and prepare no-op pull paths before graph construction |
| `crates/trusted-server-core/src/publisher.rs` | Validate markers before snapshot preload while preserving reads required by other identity consumers |
| `crates/trusted-server-adapter-fastly/src/main.rs` | Construct the post-send identity graph only after pull-sync preparation succeeds |
| `docs/guide/edge-cookies.md` | Document the host-only marker and its bounded behavior |
| `docs/superpowers/plans/2026-07-13-issue-880-no-op-pull-sync-kv-reads.md` | Record the reviewed implementation plan and validation contract |

## Closes

Closes #880

## Test plan

- [x] Other: `cargo test-cloudflare`, `cargo test-spin`, cross-adapter parity, and all Cloudflare/Spin native + WASM clippy targets

## Checklist

- [x] New code has tests

### DIFF
--- a/crates/trusted-server-adapter-fastly/src/main.rs
+++ b/crates/trusted-server-adapter-fastly/src/main.rs
@@ -21,7 +21,6 @@ use trusted_server_core::ec::registry::PartnerRegistry;
 use trusted_server_core::error::TrustedServerError;
 use trusted_server_core::integrations::RequestFilterEffects;
 use trusted_server_core::platform::PlatformGeo as _;
-use trusted_server_core::platform::RuntimeServices;
 use trusted_server_core::proxy::{stream_asset_body, AssetProxyCachePolicy};
 use trusted_server_core::settings::Settings;
 
@@ -312,11 +311,44 @@ fn run_edgezero_pull_sync_after_send(
     partner_registry: &PartnerRegistry,
     ec_state: &EcFinalizeState,
 ) {
-    if ec_state.is_real_browser {
-        if let Some(context) = build_pull_sync_context(&ec_state.ec_context) {
-            run_pull_sync_after_send(settings, partner_registry, &context, &ec_state.services);
-        }
+    if !ec_state.is_real_browser {
+        return;
     }
+
+    let prepared_context = build_pull_sync_context(&ec_state.ec_context, partner_registry);
+    let Some((context, kv)) =
+        prepare_pull_sync_after_send(prepared_context, || require_identity_graph(settings))
+    else {
+        return;
+    };
+
+    let limiter = FastlyRateLimiter::new(RATE_COUNTER_NAME);
+    dispatch_pull_sync(
+        settings,
+        &kv,
+        partner_registry,
+        &limiter,
+        &context,
+        &ec_state.services,
+    );
+}
+
+fn prepare_pull_sync_after_send<F>(
+    context: Option<PullSyncContext>,
+    graph_factory: F,
+) -> Option<(PullSyncContext, KvIdentityGraph)>
+where
+    F: FnOnce() -> Result<KvIdentityGraph, Report<TrustedServerError>>,
+{
+    let context = context?;
+    let kv = match graph_factory() {
+        Ok(kv) => kv,
+        Err(err) => {
+            log::debug!("Pull sync: identity graph unavailable, skipping: {err:?}");
+            return None;
+        }
+    };
+    Some((context, kv))
 }
 
 /// Sends a finalized `EdgeZero` response to the client.
@@ -421,24 +453,6 @@ pub(crate) fn maybe_identity_graph(settings: &Settings) -> Option<KvIdentityGrap
         .map(|store_name| KvIdentityGraph::new(FastlyEcKvStore::new(store_name)))
 }
 
-fn run_pull_sync_after_send(
-    settings: &Settings,
-    partner_registry: &PartnerRegistry,
-    context: &PullSyncContext,
-    services: &RuntimeServices,
-) {
-    let kv = match require_identity_graph(settings) {
-        Ok(kv) => kv,
-        Err(err) => {
-            log::debug!("Pull sync: identity graph unavailable, skipping: {err:?}");
-            return;
-        }
-    };
-
-    let limiter = FastlyRateLimiter::new(RATE_COUNTER_NAME);
-    dispatch_pull_sync(settings, &kv, partner_registry, &limiter, context, services);
-}
-
 /// Constructs a `KvIdentityGraph` from settings, or returns an error if the
 /// `ec_store` config is not set.
 pub(crate) fn require_identity_graph(
@@ -511,6 +525,26 @@ mod tests {
             "#,
         )
         .expect("should parse test settings")
+    }
+
+    #[test]
+    fn pull_sync_noop_states_skip_post_send_graph_factory() {
+        for reason in ["no partners", "complete snapshot", "unread marker state"] {
+            let calls = std::cell::Cell::new(0);
+            let result = prepare_pull_sync_after_send(None, || {
+                calls.set(calls.get() + 1);
+                Err(Report::new(TrustedServerError::KvStore {
+                    store_name: "unexpected".to_owned(),
+                    message: "graph factory should not run".to_owned(),
+                }))
+            });
+            assert!(result.is_none(), "{reason} preparation should return none");
+            assert_eq!(
+                calls.get(),
+                0,
+                "{reason} should not invoke the graph factory"
+            );
+        }
     }
 
     #[test]
--- a/crates/trusted-server-core/src/ec/finalize.rs
+++ b/crates/trusted-server-core/src/ec/finalize.rs
@@ -16,6 +16,7 @@ use super::generation::{generate_ec_id, is_valid_ec_id};
 use super::kv::{apply_partner_id_updates, CreateIfAbsentOutcome, KvIdentityGraph};
 use super::kv_types::KvEntry;
 use super::prebid_eids::collect_eid_cookie_updates;
+use super::pull_sync_marker::{expire_marker, reconcile_marker};
 use super::registry::PartnerRegistry;
 use super::{current_timestamp, EcContext, EcKvSnapshot};
 
@@ -47,6 +48,7 @@ pub fn ec_finalize_response(
     sharedid_cookie: Option<&str>,
     response: &mut Response<EdgeBody>,
 ) {
+    ec_context.validate_pull_sync_marker(settings, registry);
     let consent_allows_ec = ec_consent_granted(ec_context.consent());
     let consent_withdrawn = ec_consent_withdrawn(ec_context.consent());
 
@@ -57,8 +59,12 @@ pub fn ec_finalize_response(
         // consent input.
         clear_ec_headers_on_response(response, Some(registry));
 
-        // Only expire the browser cookie and tombstone the identity-graph row
-        // when the request carries an explicit withdrawal signal.
+        if consent_withdrawn {
+            expire_marker(ec_context.pull_sync_marker_mut(), response);
+        }
+
+        // Only expire the EC cookie and tombstone the identity-graph row when
+        // explicit withdrawal accompanies an EC cookie.
         if consent_withdrawn && ec_context.cookie_was_present() {
             expire_ec_cookie(settings, response);
 
@@ -102,6 +108,8 @@ pub fn ec_finalize_response(
             }
         }
 
+        reconcile_pull_sync_marker(settings, registry, ec_context, response);
+
         // Ordinary returning-user page views no longer refresh the browser
         // cookie, emit the EC header, or update KV TTL.
         return;
@@ -113,6 +121,7 @@ pub fn ec_finalize_response(
     if ec_context.ec_generated() {
         let (Some(graph), Some(ec_id)) = (kv, ec_context.ec_value().map(str::to_owned)) else {
             log::info!("Skipping generated EC response write because KV graph is unavailable");
+            reconcile_pull_sync_marker(settings, registry, ec_context, response);
             return;
         };
 
@@ -129,6 +138,26 @@ pub fn ec_finalize_response(
             log::warn!("Skipping generated EC cookie because backing row is not authoritative");
         }
     }
+
+    reconcile_pull_sync_marker(settings, registry, ec_context, response);
+}
+
+fn reconcile_pull_sync_marker(
+    settings: &Settings,
+    registry: &PartnerRegistry,
+    ec_context: &mut EcContext,
+    response: &mut Response<EdgeBody>,
+) {
+    let ec_id = ec_context.ec_value().map(str::to_owned);
+    let snapshot = ec_context.kv_snapshot().clone();
+    reconcile_marker(
+        settings,
+        registry,
+        ec_id.as_deref(),
+        &snapshot,
+        ec_context.pull_sync_marker_mut(),
+        response,
+    );
 }
 
 fn recover_orphaned_ec(
@@ -979,6 +1008,104 @@ mod tests {
         assert!(
             graph.get(&cookie_ec).expect("should read store").is_none(),
             "a missing second ID must never be created by withdrawal"
+        );
+    }
+
+    #[test]
+    fn finalize_sets_marker_for_complete_pull_partner_snapshot() {
+        let settings = create_test_settings();
+        let ec_id = sample_ec_id("compl1");
+        let mut partner = make_partner("ssp.example.com");
+        partner.pull_sync_enabled = true;
+        partner.pull_sync_url = Some("https://sync.example.com/pull".to_owned());
+        partner.pull_sync_allowed_domains = vec!["sync.example.com".to_owned()];
+        partner.ts_pull_token = Some(Redacted::new("pull-token".to_owned()));
+        let registry = PartnerRegistry::from_config(&[partner]).expect("should build registry");
+        let mut ec_context = make_context(
+            Some(&ec_id),
+            Some(&ec_id),
+            true,
+            false,
+            Jurisdiction::NonRegulated,
+        );
+        let mut entry = live_entry();
+        entry.ids.insert(
+            "ssp.example.com".to_owned(),
+            crate::ec::kv_types::KvPartnerId {
+                uid: "partner-uid".to_owned(),
+            },
+        );
+        ec_context.set_kv_snapshot(EcKvSnapshot::Present {
+            ec_id,
+            entry: Box::new(entry),
+            generation: Some(1),
+        });
+        let mut response = empty_response();
+
+        ec_finalize_response(
+            &settings,
+            &mut ec_context,
+            None,
+            &registry,
+            None,
+            None,
+            &mut response,
+        );
+
+        let cookies = response
+            .headers()
+            .get_all(http::header::SET_COOKIE)
+            .iter()
+            .filter_map(|value| value.to_str().ok())
+            .collect::<Vec<_>>();
+        assert!(
+            cookies
+                .iter()
+                .any(|cookie| cookie.starts_with("ts-ec-pull-complete=v1.")),
+            "complete snapshot should issue the marker"
+        );
+    }
+
+    #[test]
+    fn explicit_withdrawal_expires_marker_without_ec_cookie() {
+        let settings = create_test_settings();
+        let consent = ConsentContext {
+            jurisdiction: Jurisdiction::UsState("CA".to_owned()),
+            gpc: true,
+            source: ConsentSource::Cookie,
+            ..Default::default()
+        };
+        let mut ec_context = make_context_with_consent(None, None, false, false, consent);
+        ec_context.set_pull_sync_marker_for_test(
+            crate::ec::pull_sync_marker::PullSyncMarkerState::Invalid,
+        );
+        let mut response = empty_response();
+
+        ec_finalize_response(
+            &settings,
+            &mut ec_context,
+            None,
+            &PartnerRegistry::empty(),
+            None,
+            None,
+            &mut response,
+        );
+
+        let cookies = response
+            .headers()
+            .get_all(http::header::SET_COOKIE)
+            .iter()
+            .filter_map(|value| value.to_str().ok())
+            .collect::<Vec<_>>();
+        assert!(
+            cookies.iter().any(|cookie| {
+                cookie.starts_with("ts-ec-pull-complete=;") && cookie.contains("Max-Age=0")
+            }),
+            "withdrawal should expire the marker independently of EC cookie state"
+        );
+        assert!(
+            cookies.iter().all(|cookie| !cookie.starts_with("ts-ec=;")),
+            "missing EC cookie should not add an EC-cookie expiry"
         );
     }
 
--- a/crates/trusted-server-core/src/ec/mod.rs
+++ b/crates/trusted-server-core/src/ec/mod.rs
@@ -45,6 +45,7 @@ pub mod kv_types;
 pub mod partner;
 pub mod prebid_eids;
 pub mod pull_sync;
+pub(crate) mod pull_sync_marker;
 pub mod rate_limiter;
 pub mod registry;
 
@@ -65,7 +66,7 @@ use error_stack::Report;
 use http::Request;
 
 use crate::consent::{self as consent_mod, ConsentContext, ConsentPipelineInput};
-use crate::constants::COOKIE_TS_EC;
+use crate::constants::{COOKIE_TS_EC, COOKIE_TS_EC_PULL_COMPLETE};
 use crate::cookies::handle_request_cookies;
 use crate::ec::cookies::ec_id_has_only_allowed_chars;
 use crate::error::TrustedServerError;
@@ -76,6 +77,7 @@ use device::DeviceSignals;
 
 use self::kv::{CreateIfAbsentOutcome, KvIdentityGraph};
 use self::kv_types::KvEntry;
+use self::pull_sync_marker::{validate_marker_state, PullSyncMarkerState};
 
 /// Request-scoped view of one EC identity-graph lookup.
 ///
@@ -147,6 +149,8 @@ pub use generation::{
 struct RequestEc {
     /// EC ID from the `ts-ec` cookie, if present.
     cookie_ec: Option<String>,
+    /// Pull-sync completeness marker, if present.
+    pull_sync_marker: Option<String>,
     /// The parsed cookie jar (retained for consent pipeline input).
     jar: Option<CookieJar>,
 }
@@ -163,8 +167,17 @@ fn parse_ec_from_request(req: &Request<EdgeBody>) -> Result<RequestEc, Report<Tr
         .and_then(|j| j.get(COOKIE_TS_EC))
         .map(cookie::Cookie::value)
         .and_then(|value| request_ec_id_if_allowed(value, "ts-ec cookie"));
+    let pull_sync_marker = jar
+        .as_ref()
+        .and_then(|j| j.get(COOKIE_TS_EC_PULL_COMPLETE))
+        .map(cookie::Cookie::value)
+        .map(str::to_owned);
 
-    Ok(RequestEc { cookie_ec, jar })
+    Ok(RequestEc {
+        cookie_ec,
+        pull_sync_marker,
+        jar,
+    })
 }
 
 fn request_ec_id_if_allowed(value: &str, source: &str) -> Option<String> {
@@ -226,6 +239,8 @@ pub struct EcContext {
     kv_snapshot: EcKvSnapshot,
     /// Whether this request may rotate an orphaned EC identity.
     recovery_eligible: bool,
+    /// Browser-carried proof of recent pull-partner completeness.
+    pull_sync_marker: PullSyncMarkerState,
 }
 
 impl EcContext {
@@ -307,6 +322,7 @@ impl EcContext {
             device_signals: None,
             kv_snapshot: EcKvSnapshot::NotRead,
             recovery_eligible: false,
+            pull_sync_marker: PullSyncMarkerState::from_cookie(parsed.pull_sync_marker),
         })
     }
 
@@ -393,6 +409,7 @@ impl EcContext {
 
             self.ec_value = Some(ec_id);
             self.ec_generated = true;
+            self.pull_sync_marker.invalidate_for_replaced_ec();
             return Ok(());
         }
 
@@ -495,11 +512,43 @@ impl EcContext {
         self.recovery_eligible
     }
 
+    /// Validates a browser completeness marker against the active EC and partner set.
+    pub(crate) fn validate_pull_sync_marker(
+        &mut self,
+        settings: &Settings,
+        registry: &registry::PartnerRegistry,
+    ) {
+        validate_marker_state(
+            &mut self.pull_sync_marker,
+            settings,
+            registry,
+            self.ec_value.as_deref(),
+        );
+    }
+
+    /// Returns the current pull-sync marker state.
+    #[must_use]
+    pub(crate) fn pull_sync_marker(&self) -> &PullSyncMarkerState {
+        &self.pull_sync_marker
+    }
+
+    /// Returns mutable pull-sync marker state for response reconciliation.
+    pub(crate) fn pull_sync_marker_mut(&mut self) -> &mut PullSyncMarkerState {
+        &mut self.pull_sync_marker
+    }
+
+    /// Sets pull-sync marker state in focused unit tests.
+    #[cfg(test)]
+    pub(crate) fn set_pull_sync_marker_for_test(&mut self, state: PullSyncMarkerState) {
+        self.pull_sync_marker = state;
+    }
+
     /// Replaces an orphaned active ID after its new backing row is persisted.
     pub(crate) fn replace_with_generated(&mut self, ec_id: String, snapshot: EcKvSnapshot) {
         self.ec_value = Some(ec_id);
         self.ec_generated = true;
         self.kv_snapshot = snapshot;
+        self.pull_sync_marker.invalidate_for_replaced_ec();
     }
 
     /// Returns whether EC creation is permitted by consent for this request.
@@ -549,6 +598,7 @@ impl EcContext {
             device_signals: None,
             kv_snapshot: EcKvSnapshot::NotRead,
             recovery_eligible: false,
+            pull_sync_marker: PullSyncMarkerState::Absent,
         }
     }
 
@@ -571,6 +621,7 @@ impl EcContext {
             device_signals: None,
             kv_snapshot: EcKvSnapshot::NotRead,
             recovery_eligible: false,
+            pull_sync_marker: PullSyncMarkerState::Absent,
         }
     }
 
@@ -596,6 +647,7 @@ impl EcContext {
             device_signals: None,
             kv_snapshot: EcKvSnapshot::NotRead,
             recovery_eligible: false,
+            pull_sync_marker: PullSyncMarkerState::Absent,
         }
     }
 }
--- a/crates/trusted-server-core/src/ec/pull_sync.rs
+++ b/crates/trusted-server-core/src/ec/pull_sync.rs
@@ -22,6 +22,7 @@ use crate::settings::Settings;
 use super::generation::{ec_hash, is_valid_ec_id};
 use super::kv::{KvIdentityGraph, PartnerIdUpdate};
 use super::kv_types::KvEntry;
+use super::pull_sync_marker::entry_is_pull_complete;
 use super::rate_limiter::RateLimiter;
 use super::registry::{PartnerConfig, PartnerRegistry};
 
@@ -59,14 +60,22 @@ struct PullSyncResponse {
 ///
 /// Returns `None` when consent denies EC or there is no active EC ID.
 #[must_use]
-pub fn build_pull_sync_context(ec_context: &EcContext) -> Option<PullSyncContext> {
-    if !ec_context.ec_allowed() {
+pub fn build_pull_sync_context(
+    ec_context: &EcContext,
+    registry: &PartnerRegistry,
+) -> Option<PullSyncContext> {
+    if registry.pull_enabled_partners().is_empty() || !ec_context.ec_allowed() {
         return None;
     }
 
     let ec_id_ref = ec_context.ec_value()?;
     if !is_valid_ec_id(ec_id_ref) {
         log::debug!("Pull sync: skipping dispatch because active EC ID is invalid format");
+        return None;
+    }
+
+    let entry = ec_context.kv_snapshot().entry_for(ec_id_ref)?;
+    if !entry.consent.ok || entry_is_pull_complete(entry, registry) {
         return None;
     }
 
@@ -92,14 +101,18 @@ pub fn dispatch_pull_sync(
     services: &RuntimeServices,
 ) {
     let now = current_timestamp();
+    let mut pull_partners = registry.pull_enabled_partners();
+
+    if pull_partners.is_empty() {
+        return;
+    }
+
     let Some(kv_entry) = context.snapshot.entry_for(context.ec_id()) else {
         return;
     };
     if !kv_entry.consent.ok {
         return;
     }
-
-    let mut pull_partners = registry.pull_enabled_partners();
 
     // Sort by source domain for deterministic ordering, then apply a rotating
     // hourly offset so that different partners get dispatch priority (§10.3).
@@ -109,10 +122,6 @@ pub fn dispatch_pull_sync(
         "Pull sync: {} pull-enabled partners after filtering",
         pull_partners.len(),
     );
-
-    if pull_partners.is_empty() {
-        return;
-    }
 
     // Rotate the partner list so that the starting partner changes each
     // hour. This ensures fair distribution when max_concurrency limits
@@ -493,9 +502,13 @@ mod tests {
             ..ConsentContext::default()
         };
         let ec_id = format!("{}.ABC123", "a".repeat(64));
-        let ec_context = EcContext::new_for_test(Some(ec_id), consent);
+        let mut ec_context = EcContext::new_for_test(Some(ec_id.clone()), consent);
+        let graph = KvIdentityGraph::in_memory("pull_store");
+        ec_context.set_kv_snapshot(seed_present_snapshot(&graph, &ec_id));
+        let registry = PartnerRegistry::from_config(&[pull_enabled_ec_partner("ssp.example.com")])
+            .expect("should build registry");
 
-        let context = build_pull_sync_context(&ec_context)
+        let context = build_pull_sync_context(&ec_context, &registry)
             .expect("should build pull sync context for valid EC");
         assert_eq!(
             context.ec_id(),
@@ -511,11 +524,50 @@ mod tests {
             ..ConsentContext::default()
         };
         let ec_context = EcContext::new_for_test(Some("invalid-ec".to_owned()), consent);
+        let registry = PartnerRegistry::from_config(&[pull_enabled_ec_partner("ssp.example.com")])
+            .expect("should build registry");
 
-        let context = build_pull_sync_context(&ec_context);
+        let context = build_pull_sync_context(&ec_context, &registry);
         assert!(
             context.is_none(),
             "should reject pull sync context when EC ID format is invalid"
+        );
+    }
+
+    #[test]
+    fn build_pull_sync_context_skips_empty_registry_and_complete_snapshot() {
+        let consent = ConsentContext {
+            jurisdiction: crate::consent::jurisdiction::Jurisdiction::NonRegulated,
+            ..ConsentContext::default()
+        };
+        let ec_id = format!("{}.ABC123", "a".repeat(64));
+        let mut ec_context = EcContext::new_for_test(Some(ec_id.clone()), consent);
+        let graph = KvIdentityGraph::in_memory("pull_store");
+        let mut snapshot = seed_present_snapshot(&graph, &ec_id);
+        let registry = PartnerRegistry::from_config(&[pull_enabled_ec_partner("ssp.example.com")])
+            .expect("should build registry");
+
+        assert!(
+            build_pull_sync_context(&ec_context, &PartnerRegistry::empty()).is_none(),
+            "no pull partners should skip before graph construction"
+        );
+        assert!(
+            build_pull_sync_context(&ec_context, &registry).is_none(),
+            "an unread snapshot should skip before graph construction"
+        );
+
+        if let EcKvSnapshot::Present { entry, .. } = &mut snapshot {
+            entry.ids.insert(
+                "ssp.example.com".to_owned(),
+                crate::ec::kv_types::KvPartnerId {
+                    uid: "uid".to_owned(),
+                },
+            );
+        }
+        ec_context.set_kv_snapshot(snapshot);
+        assert!(
+            build_pull_sync_context(&ec_context, &registry).is_none(),
+            "a complete snapshot should skip post-send work"
         );
     }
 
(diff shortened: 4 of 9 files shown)


crates/trusted-server-core/src/ec/finalize.rs
```diff
@@ -58,8 +60,12 @@ pub fn ec_finalize_response(
         // consent input.
         clear_ec_headers_on_response(response, Some(registry));
 
-        // Only expire the browser cookie and tombstone the identity-graph row
-        // when the request carries an explicit withdrawal signal.
+        if consent_withdrawn {
+            expire_marker(ec_context.pull_sync_marker_mut(), response);
+        }
```
prk-Jr: 🔧 **wrench** — Withdrawal always emits a marker-expiry `Set-Cookie`, downgrading cache privacy for all withdrawn traffic.

`expire_marker` has no presence guard — it appends the cookie whatever the state, including `Absent`. So every explicitly-withdrawn request (GPC=true, US-state opt-out) now returns `Set-Cookie: ts-ec-pull-complete=; …Max-Age=0`, even for the common case of a withdrawn visitor who has no EC cookie and never had a marker. On the base branch this path emitted no `Set-Cookie` at all, because `expire_ec_cookie` is gated on `cookie_was_present()`.

Downstream that matters: `apply_terminal_response_effects` (`crates/trusted-server-adapter-fastly/src/main.rs:418`) runs `enforce_set_cookie_cache_privacy` *after* finalize, and for any cookie-bearing response that strips every edge-cache header and forces `Cache-Control: private, max-age=0`. `template_cache_ttl` (`crates/trusted-server-core/src/publisher.rs:6168`) also bypasses the shared template cache on `Set-Cookie`. Withdrawn requests never run the ad stack, so they never get `with_cache_bypass` — those responses were shareable before this change and are not after it, for an entire traffic class, as a side effect of a KV-read optimization.

Gating on `was_present()` keeps the plan's guarantee intact: it reads only the request cookie, never KV, so withdrawal still clears the marker with no KV dependency — exactly how `ts-ec` expiry is already gated on `cookie_was_present()`. `explicit_withdrawal_expires_marker_without_ec_cookie` still passes, since it seeds `PullSyncMarkerState::Invalid` (present).

```suggestion
        if consent_withdrawn && ec_context.pull_sync_marker().was_present() {
            expire_marker(ec_context.pull_sync_marker_mut(), response);
        }
```

_Verified in an isolated worktree at this head: `cargo fmt --all -- --check`, `cargo clippy-fastly`, `cargo check-axum`, `cargo check-cloudflare`, and 332 native `trusted-server-core` `ec::` tests all pass with this applied._

crates/trusted-server-core/src/ec/finalize.rs
```diff
@@ -144,6 +153,26 @@ pub fn ec_finalize_response(
             log::warn!("Skipping generated EC cookie because backing row is not authoritative");
         }
     }
+
+    reconcile_pull_sync_marker(settings, registry, ec_context, response);
+}
+
+fn reconcile_pull_sync_marker(
```
prk-Jr, line 131: 🤔 **thinking** — Marker issuance re-adds a `Set-Cookie` to responses PR #885 deliberately made cookie-free.

Same mechanism as the wrench finding above, but inherent to the design rather than a bug. The returning-user path comments that "ordinary returning-user page views no longer refresh the browser cookie" — yet a marker issue or expire makes exactly those responses cookie-bearing, so `enforce_set_cookie_cache_privacy` downgrades them to `private, max-age=0` with edge-cache headers stripped, and Esi-mode responses stop being eligible for template-cache storage.

Bounded to roughly one response per hour per user per host, against one saved KV read per navigation, so it is plausibly a clear win — but the trade is not in the plan's risk table or in `docs/guide/edge-cookies.md`, and the two effects land in different budgets (KV operations vs. edge cache hit rate). Worth stating explicitly so it isn't rediscovered from a cache-hit-rate graph later.

crates/trusted-server-core/src/ec/pull_sync_marker.rs
```diff
@@ -0,0 +1,673 @@
...
+    let pull_partners = sorted_pull_partner_domains(registry);
+    !pull_partners.is_empty() && entry_has_all_pull_partner_ids(entry, &pull_partners)
+}
+
+fn entry_has_all_pull_partner_ids(entry: &KvEntry, pull_partners: &[String]) -> bool {
+    entry.consent.ok
+        && pull_partners
+            .iter()
+            .all(|source_domain| entry.ids.contains_key(source_domain))
+}
```
prk-Jr: ♻️ **refactor** — Completeness duplicates dispatch eligibility; the two can drift.

`entry_has_all_pull_partner_ids` reimplements the inverse of `is_partner_pull_eligible` (`crates/trusted-server-core/src/ec/pull_sync.rs:286-290`, which is just `entry.ids.get(&partner.source_domain).is_none()`). They agree today only because eligibility ignores `pull_sync_ttl_sec`. If per-partner TTL refresh is ever wired into eligibility, completeness silently over-claims, and the marker then suppresses both the snapshot read and pull sync for up to an hour — the failure would be invisible, because both sides would still "agree" in their own tests.

Deriving one from the other makes that drift impossible:

```rust
// pull_sync.rs
pub(crate) fn is_partner_pull_eligible(partner: &PartnerConfig, kv_entry: Option<&KvEntry>) -> bool

// pull_sync_marker.rs
fn entry_has_all_pull_partner_ids(entry: &KvEntry, registry: &PartnerRegistry) -> bool {
    entry.consent.ok
        && !registry
            .pull_enabled_partners()
            .iter()
            .any(|partner| is_partner_pull_eligible(partner, Some(entry)))
}
```

Apply manually — can't be auto-applied as a suggestion because it changes `is_partner_pull_eligible`'s visibility in `pull_sync.rs` and replaces the domain-slice parameter threaded through `reconcile_marker`, so it spans two files.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -536,6 +550,26 @@ mod tests {
...
+                }))
+            });
+            assert!(result.is_none(), "{reason} preparation should return none");
+            assert_eq!(
+                calls.get(),
+                0,
+                "{reason} should not invoke the graph factory"
+            );
+        }
+    }
```
prk-Jr, line 572: ⛏ **nitpick** — The three-state loop asserts the same thing three times.

`for reason in ["no partners", "complete snapshot", "unread marker state"]` runs three byte-identical `prepare_pull_sync_after_send(None, …)` calls. None of the three named states is exercised here — all three are decided inside `build_pull_sync_context`, which this test never calls — so the labels read as coverage that doesn't exist. What the test actually proves (a `None` plan never reaches the graph factory) is worth keeping; the loop is not.

```suggestion
    #[test]
    fn pull_sync_noop_states_skip_post_send_graph_factory() {
        let calls = std::cell::Cell::new(0);
        let result = prepare_pull_sync_after_send(None, || {
            calls.set(calls.get() + 1);
            Err(Report::new(TrustedServerError::KvStore {
                store_name: "unexpected".to_owned(),
                message: "graph factory should not run".to_owned(),
            }))
        });
        assert!(
            result.is_none(),
            "a skipped pull-sync plan should return none"
        );
        assert_eq!(calls.get(), 0, "should not invoke the graph factory");
    }
```

The three states themselves are already covered in core by `build_pull_sync_context_skips_empty_registry_and_complete_snapshot`.

_Verified in an isolated worktree at this head: `cargo fmt --all -- --check` and `cargo clippy-fastly` (which compiles `--all-targets`, including this test) both pass with this applied._

crates/trusted-server-core/src/ec/pull_sync_marker.rs
```diff
@@ -0,0 +1,673 @@
...
+
+    let payload = marker_payload(ec_id, expires_at, fingerprint);
+    let key = marker_key(settings);
+    let mut mac = HmacSha256::new_from_slice(&key).expect("should create marker HMAC");
+    mac.update(payload.as_bytes());
+    mac.verify_slice(&tag).ok()?;
+    Some(expires_at)
+}
+
+fn marker_key(settings: &Settings) -> [u8; 32] {
```
prk-Jr, line 292: 👍 **praise** — Marker crypto and framing.

The details here are the ones that usually get skipped: labelled key derivation off `ec.passphrase` rather than using the passphrase directly as the MAC key; a NUL-delimited payload that puts the version and the fingerprint *inside* the MAC input, so no field can be shifted into another; constant-time `verify_slice`; a length cap applied before any parsing; a host-only cookie, deliberately tighter than `ts-ec`'s `Domain=.<publisher>`; and `Redacted` plus a hand-written `Debug` so the value can't leak through a log line. No UID or EC ID in the cookie itself. The order-independent, set-sensitive fingerprint test covering add/remove/enable/disable is the right shape of test for this.

prk-Jr (changes requested):
## Summary

Solid, well-scoped implementation of the #880 contract: the marker protocol is cryptographically careful, the preload gate enumerates its consumers explicitly instead of guessing, and the fallbacks all fail toward the KV path. One blocking issue: the withdrawal path emits the marker-expiry `Set-Cookie` unconditionally, which downgrades cache privacy for an entire traffic class that previously had no `Set-Cookie` at all.

> 2 of the inline comments below carry a one-click GitHub `suggestion` — use **Commit suggestion** (or **Add suggestion to batch**) to apply them as commits on the PR branch. Both were applied in an isolated worktree at this head and verified with `cargo fmt --all -- --check`, `cargo clippy-fastly`, `cargo check-axum`, `cargo check-cloudflare`, and the native `trusted-server-core` `ec::` + publisher marker tests (332 and 28 passing respectively), with a byte-exact pre/post drift check. The remaining comments describe the fix in prose because it spans two files.

## Blocking

### 🔧 wrench

- **Withdrawal always emits a marker-expiry `Set-Cookie`, downgrading cache privacy for all withdrawn traffic** — see inline at `crates/trusted-server-core/src/ec/finalize.rs:63`

## Non-blocking

### 🤔 thinking

- **Marker issuance re-adds a `Set-Cookie` to responses PR #885 deliberately made cookie-free** — see inline at `crates/trusted-server-core/src/ec/finalize.rs:160`

### ♻️ refactor

- **Completeness duplicates dispatch eligibility; the two can drift** — see inline at `crates/trusted-server-core/src/ec/pull_sync_marker.rs:170`

### ⛏ nitpick

- **The three-state loop asserts the same thing three times** — see inline at `crates/trusted-server-adapter-fastly/src/main.rs:553`

### 👍 praise

- **Marker crypto and framing** — see inline at `crates/trusted-server-core/src/ec/pull_sync_marker.rs:299`

- 📝 **Verified as correct, for the record** — several things that look risky on first read hold up: the `expect` in `reconcile_marker` is unreachable because the `belongs_to` check precedes the `Present` match and `entry_for` ignores `generation`; the partner-set fingerprint is consistent across the validate and issue sites because every caller builds the registry from `PartnerRegistry::from_config(&settings.ec.partners)`; `upsert_partner_ids_from_snapshot` returns early on empty updates, so a marker-valid navigation really does perform zero reads rather than moving the read later; stored KV EIDs reach only the auction path, which `auction_needs_row` forces the read for; tombstones are excluded via `consent.ok`; and per-partner `pull_sync_ttl_sec` is not wired into eligibility, so presence-only completeness matches dispatch today (see the ♻️ finding for the drift risk that creates).

## CI Status

- cargo fmt: PASS
- cargo test: PASS
- cargo test (cloudflare native + wasm32-unknown-unknown): PASS
- vitest: PASS
- format-typescript: PASS
- format-docs: PASS
- integration tests: PASS

All 14 reported checks pass. `gh pr checks --required` returned no names for this PR, so no check is annotated as branch-protection-required (the base branch is `fix/kv-eid-request-snapshot-ec-ttl`, not `main`).

crates/trusted-server-core/src/ec/finalize.rs
ChristianPavilonis: Addressed in 963c331c. Withdrawal now emits the marker-expiration cookie only when the request carried a marker. I added a regression proving that withdrawal with neither an EC cookie nor marker remains cookie-free; the existing present-marker test still proves deletion is independent of EC cookie and KV state.

crates/trusted-server-core/src/ec/finalize.rs
ChristianPavilonis: Documented in 963c331c. The Edge Cookie guide and implementation plan now state that issuing or expiring the marker adds Set-Cookie and can make an otherwise shareable outgoing response private, in exchange for avoiding later KV reads. They also note that a valid marker is not refreshed on each request.

crates/trusted-server-core/src/ec/pull_sync_marker.rs
ChristianPavilonis: Addressed in 963c331c. The shared completeness predicate now lives in pull_sync.rs and derives completeness by negating is_partner_pull_eligible across the pull-enabled partners. pull_sync_marker.rs consumes that predicate, so dispatch eligibility has one owner without creating a mutual module dependency.

crates/trusted-server-adapter-fastly/src/main.rs
ChristianPavilonis: Addressed in 963c331c. I removed the loop and kept one assertion that a None plan never invokes the graph factory. The distinct no-partner, unread-snapshot, and complete-snapshot states remain covered in the core pull-sync test.

crates/trusted-server-core/src/ec/pull_sync_marker.rs
ChristianPavilonis: Thanks for the careful review of the marker protocol and its tests. No change was needed for this thread.

aram356:
@ChristianPavilonis please assign issue for this PR

crates/trusted-server-core/src/ec/pull_sync.rs
```diff
@@ -59,8 +59,11 @@ struct PullSyncResponse {
 ///
 /// Returns `None` when consent denies EC or there is no active EC ID.
```
prk-Jr: ♻️ **refactor** — This is a `pub` function and the contract widened in this PR: it now also returns `None` when the registry has no pull-enabled partners, when the request snapshot has no entry for the active EC ID (`NotRead` / `Missing` / `Failed`), when the entry's consent is not ok, and when the entry is already pull-complete. The current line describes only two of those.

```suggestion
/// Returns `None` when there are no pull-enabled partners, when consent denies
/// EC, when there is no valid active EC ID, or when the request snapshot holds
/// no consented row that is still missing at least one pull-partner UID.
```

_(verified in a scratch worktree: `cargo fmt --all -- --check` and `cargo check-fastly` clean)_

crates/trusted-server-core/src/ec/pull_sync.rs
```diff
@@ -274,6 +282,17 @@ pub fn dispatch_pull_sync(
     }
 }
 
+/// Returns whether a live entry contains every pull-enabled partner ID.
```
prk-Jr: ♻️ **refactor** — The body also requires `entry.consent.ok` and a non-empty pull-partner set, so a consent-revoked entry and an empty registry both return `false`. The empty-registry case in particular is load-bearing — it is what stops `reconcile_marker` from issuing a marker for a configuration with no pull partners — so it is worth stating.

```suggestion
/// Returns whether a consented live entry holds a UID for every pull-enabled
/// partner. Always `false` when the registry has no pull-enabled partners.
```

_(verified in a scratch worktree: `cargo fmt --all -- --check` and `cargo check-fastly` clean)_

crates/trusted-server-core/src/ec/pull_sync_marker.rs
```diff
@@ -0,0 +1,644 @@
...
+            &settings,
+            &registry,
+            Some(EC_ID),
+            &EcKvSnapshot::NotRead,
+            &mut state,
+            &mut response,
+        );
+
+        assert_expired(&state, &response);
+    }
```
prk-Jr, line 616: 🤔 **thinking** — `docs/guide/edge-cookies.md` states that "A valid marker is not refreshed on each request, so this cost is limited to responses that establish or clear marker state in exchange for avoiding later KV reads." That is a cache-privacy claim: every marker write forces `private, max-age=0` through `enforce_set_cookie_cache_privacy` and strips the edge-cache headers. The only thing enforcing it is the `if !state.is_valid()` guard on line 139, and no test covers it.

The mirror case matters just as much: a user whose row will never complete (a partner that never returns a UID) sits at `Absent` with an incomplete `Present` snapshot on every navigation, and `expire_if_present` must stay a no-op there or every one of those responses becomes uncacheable.

Both are one test:

```suggestion
        assert_expired(&state, &response);
    }

    #[test]
    fn steady_states_leave_the_marker_cookie_untouched() {
        let (settings, registry) = settings_and_registry(&["a.example.com"]);
        let complete = live_snapshot(EC_ID, &["a.example.com"]);
        let incomplete = live_snapshot(EC_ID, &[]);
        let cases = [
            (PullSyncMarkerState::Valid { expires_at: 4_600 }, &complete),
            (PullSyncMarkerState::Absent, &incomplete),
        ];

        for (mut state, snapshot) in cases {
            let mut response = empty_response();
            reconcile_marker(
                &settings,
                &registry,
                Some(EC_ID),
                snapshot,
                &mut state,
                &mut response,
            );
            assert!(
                marker_cookies(&response).is_empty(),
                "a steady marker state should not add a Set-Cookie header"
            );
        }
    }
```

_(verified in a scratch worktree: `cargo fmt --all -- --check`, `cargo clippy-fastly`, and `cargo test -p trusted-server-core --lib ec::` all clean — 334 passed, including the new case)_

crates/trusted-server-core/src/publisher.rs
```diff
@@ -4388,8 +4405,26 @@ pub async fn handle_publisher_request(
 
     let request_method = req.method().clone();
     let mut origin_request = Some(req);
-    let should_preload_ec =
-        should_preload_ec_snapshot(is_navigation, is_get, active_ec_id.is_some(), kv.is_some());
+    let eid_cookie_may_need_persistence = cookie_jar
+        .as_ref()
+        .is_some_and(|jar| jar.get(COOKIE_TS_EIDS).is_some() || jar.get(COOKIE_SHAREDID).is_some());
+    let should_preload_ec = should_preload_ec_snapshot(&EcSnapshotPreloadInput {
```
prk-Jr, line 4414: 🤔 **thinking** — `should_preload_ec` does two jobs: it decides the KV read, and it also gates the concurrent origin `send_async` on line 4434. Before this PR the two were the same question. Now a marker skip flips both.

I convinced myself it is currently latency-neutral, but only through a three-step chain: `marker_valid` requires `partner_set_fingerprint` to be `Some`, which requires a non-empty pull set, which makes `registry.is_empty()` false, which makes `auction_needs_row == should_run_auction` — so `marker_can_skip` implies `!should_run_auction`, so neither the `dispatch_auction` await on line 4538 nor the template-cache lookup on line 4637 (already excluded by the `template_cache_key.is_none()` guard on the concurrent path) sits between here and the origin send on line 4724.

That is a real invariant, but nothing in the code states it. If `auction_needs_row` is ever loosened — say to skip the read for an auction with no server-side EIDs — the origin fetch would silently move behind auction dispatch, and the parity suite would not catch it. Worth naming so the next editor sees the coupling:

```suggestion
    // Also gates the concurrent origin `send_async` below. A marker-driven skip
    // stays latency-neutral only because `marker_valid` implies a non-empty
    // registry, so `auction_needs_row == should_run_auction` and a skip means no
    // auction dispatch runs between here and the origin send.
    let should_preload_ec = should_preload_ec_snapshot(&EcSnapshotPreloadInput {
```

_(verified in a scratch worktree: `cargo fmt --all -- --check` and `cargo check-fastly` clean)_

prk-Jr (approved):
## Summary

Adds a signed, host-only browser marker that lets an eligible navigation skip the EC identity-graph read when pull-sync completeness is the only reason to inspect the row, and returns from post-send pull sync before the graph is constructed when there is nothing to pull. The design is sound and the skip is well fenced: I traced each path rather than only reading the diff.

> 4 of the inline comments below carry a one-click GitHub `suggestion` — use **Commit suggestion** (or **Add suggestion to batch** for several at once) to apply them as commits on the PR branch. All four were applied and verified in an isolated worktree at `963c331` (`cargo fmt --all -- --check`, `cargo clippy-fastly`, `cargo check-fastly`, and `cargo test -p trusted-server-core --lib ec::` — 334 passed), individually and as a batch, with a byte-exact pre/post drift check.

### What I verified beyond the diff

- **`marker_valid` implies a non-empty registry.** `partner_set_fingerprint` returns `None` for an empty pull set, so a valid marker requires at least one pull-enabled partner, so `auction_needs_row == should_run_auction`. A marker skip can therefore never suppress a read the auction needs.
- **A generated EC never carries a valid marker.** `generate_if_needed` and `replace_with_generated` both call `invalidate_for_replaced_ec()`, so the generated-cookie path at `ec/finalize.rs:150` always has a real snapshot. No phantom-cookie regression from `entry_for(&ec_id).is_none()`.
- **The EID gate matches its consumer exactly.** `collect_eid_cookie_updates` reads only `ts-eids` and `sharedId` — the same two constants `eid_cookie_may_need_persistence` checks — and `upsert_partner_ids_from_snapshot` short-circuits on empty updates, so a `NotRead` snapshot costs no hidden read in finalize.
- **No shared-cache leak from the new `Set-Cookie`.** `enforce_set_cookie_cache_privacy` runs in `apply_terminal_response_effects`, i.e. after `ec_finalize_response`, and the template cache stores reader-neutral bytes rather than response headers.
- **The `expect()` in `reconcile_marker` cannot fire.** The `!belongs_to(ec_id)` early return plus the `Present` match arm together guarantee `entry_for` is `Some`.
- **`reconcile_marker` runs exactly once per finalize call.** Every branch either returns before it or falls through to a single call, so no response can pick up a duplicate marker `Set-Cookie`.
- Only the Fastly adapter calls `build_pull_sync_context`; the Cloudflare, Spin, and Axum adapters are unaffected by the signature change.

## Non-blocking

### ♻️ refactor

- **`build_pull_sync_context` doc no longer describes its return contract** — see inline at `crates/trusted-server-core/src/ec/pull_sync.rs:60`
- **`entry_is_pull_complete` doc omits the consent and empty-registry conditions** — see inline at `crates/trusted-server-core/src/ec/pull_sync.rs:285`

### 🤔 thinking

- **No test for the two "steady state implies no `Set-Cookie`" invariants** — see inline at `crates/trusted-server-core/src/ec/pull_sync_marker.rs:580`
- **`should_preload_ec` is double-duty and its safety argument is implicit** — see inline at `crates/trusted-server-core/src/publisher.rs:4411`

- 🤔 **`prepare_pull_sync_after_send`'s test asserts a property that already held** — `crates/trusted-server-adapter-fastly/src/main.rs:553`. `pull_sync_noop_states_skip_post_send_graph_factory` proves that a `None` plan does not build the graph, but the pre-PR shape (`if is_real_browser && let Some(context) = build_pull_sync_context(..)` wrapping `run_pull_sync_after_send`, which built the graph inside) already guaranteed that. The behaviour that actually changed — `build_pull_sync_context` returning `None` for a complete row or an empty pull registry — is covered by `build_pull_sync_context_skips_empty_registry_and_complete_snapshot`. The case this helper uniquely adds and does not cover is `Some(context)` with a failing `graph_factory`. Worth either covering that arm or noting that the helper exists purely to make the ordering testable.

- 🌱 **Allocation churn on the finalize path** — `sorted_pull_partner_domains` clones every `source_domain` `String`, and a single `reconcile_marker` walks it twice on the issue path (once directly, once through `create_marker` -> `partner_set_fingerprint`), with `entry_is_pull_complete` calling `pull_enabled_partners()` again. That is three to four partner-set traversals per response in Wasm, where CLAUDE.md asks to minimise allocations. Small in absolute terms and it does not belong in this PR, but a cached fingerprint on `PartnerRegistry` would remove all of it. Follow-up.

- 📝 **Orphan-recovery detection is deferred by up to `MARKER_MAX_AGE_SECS`** — a marker-skipped navigation never observes a `Missing` snapshot, so `recovery_eligible` cannot fire until the marker expires. This is explicitly accepted in the plan doc and `docs/guide/edge-cookies.md`; recording it here so the tradeoff is visible on the PR thread rather than only in the plan.

- 👍 **The crypto hygiene here is genuinely good.** Key separation through `MARKER_KEY_LABEL` rather than using `ec.passphrase` as the HMAC key directly; length-prefixed domain hashing in `partner_set_fingerprint`, which removes the concatenation ambiguity a naive join would introduce; constant-time `verify_slice` for the tag; a both-sided expiry window (`expires_at <= now` and `expires_at > now + MARKER_MAX_AGE_SECS`) so a forged far-future expiry is rejected; and `Redacted` plus a hand-written `Debug` so the marker cannot leak into logs. `EcSnapshotPreloadInput` naming all four "another consumer needs the row" guards is the right call too — it turns an implicit condition into something reviewable.

## CI Status

- `cargo fmt`: PASS
- `cargo test`: PASS
- `cargo test (axum native)`: PASS
- `cargo test (cross-adapter parity)`: PASS
- `cargo test (ts CLI, native)`: PASS
- `cargo check (cloudflare native + wasm32-unknown-unknown)`: PASS
- `cargo check/build/test (spin native + wasm32-wasip1)`: PASS
- `integration tests`: PASS
- `integration tests (Fastly EC lifecycle)`: PASS
- `browser integration tests`: PASS
- `prepare integration artifacts`: PASS
- `vitest`: PASS
- `format-typescript`: PASS
- `format-docs`: PASS

No check reported as required by branch protection on this branch; all 14 reported checks pass.

crates/trusted-server-core/src/ec/finalize.rs
```diff
@@ -86,6 +94,8 @@ pub fn ec_finalize_response(
             }
         }
 
+        reconcile_pull_sync_marker(settings, registry, ec_context, response);
```
aram356, line 97: 🔧 **wrench** — Orphan recovery silently stops firing on the marker-skip path.

The PR summary says snapshot reads required by "orphan recovery" are preserved. They aren't. Recovery is gated four lines above this one on the snapshot being `Missing`:

```rust
if matches!(ec_context.kv_snapshot(), EcKvSnapshot::Missing { .. })
    && ec_context.recovery_eligible()
```

On the marker-skip path the preload never runs, so the snapshot is `NotRead`, not `Missing`. And `updates` is necessarily empty there — a `ts-eids`/`sharedId` cookie would have tripped `eid_cookie_may_need_persistence` and forced the preload — so `upsert_partner_ids_from_snapshot` returns at its `if updates.is_empty() { return snapshot; }` short-circuit (`ec/kv.rs:557`) with `NotRead` intact. `matches!(NotRead, Missing { .. })` is false, and `confirm_then_recover_orphaned_ec` is never called.

I verified this rather than inferring it. Taking the existing `finalize_rotates_orphaned_cookie_to_new_backed_ec` test and changing only the seeded snapshot from `Missing` to `NotRead`:

```
SCRATCH: rotated=false set_cookie_present=false
```

The orphan is not rotated and no replacement cookie is issued. With `Missing` the same test rotates correctly.

Two things make this worse than a one-request delay:

1. **It is self-perpetuating.** `reconcile_marker`'s `Failed { .. } | EcKvSnapshot::NotRead` arm leaves a `Valid` marker untouched (it only expires an `Invalid` one). So the orphaned user skips the read, keeps the marker, and skips again on every subsequent navigation for the remaining hour.
2. **The affected population is exactly the at-risk one.** `ts-ec` is year-lived while the marker is `Max-Age=3600`, and a marker holder had a complete row an hour ago — precisely the user whose row may since have been deleted or evicted. This is the case #885 was merged to handle.

The direct fix is a fifth hatch:

```rust
// in EcSnapshotPreloadInput
recovery_eligible: bool,

// in should_preload_ec_snapshot
let marker_can_skip = input.marker_valid
    && !input.auction_needs_row
    && !input.eid_cookie_may_need_persistence
    && !input.privacy_needs_row
    && !input.recovery_eligible
    && !input.snapshot_already_read;
```

But note that `recovery_eligible` is set for real-browser navigations, which is close to the whole target population — so combined with the separate finding on `publisher.rs:4417`, it may leave the optimization with very little left. If you'd rather keep the deferral as a deliberate trade-off, that's defensible, but then please correct the PR description and the `docs/guide/edge-cookies.md` sentence that lists "detected orphan recovery" among the paths that "continue to use KV" — detection *is* the read being skipped — and add a test pinning the chosen behaviour.

Apply manually — can't be auto-applied as a suggestion because the fix spans the struct definition, the gate body, and the call site across two separate hunks.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -4389,8 +4406,26 @@ pub async fn handle_publisher_request(
...
+        is_navigation,
+        is_get,
+        has_ec_id: active_ec_id.is_some(),
+        has_kv: kv.is_some(),
+        marker_valid: ec_context.pull_sync_marker().is_valid(),
```
aram356, line 4419: 🔧 **wrench** — The `auction_needs_row` hatch cancels the optimization on exactly the traffic it targets.

`auction_needs_row` is `should_run_auction && registry non-empty`, and `should_run_auction` is `should_run_ad_stack`, which resolves to:

```rust
is_server_side_ad_eligible_navigation(is_get, is_navigation, is_prefetch, is_bot, consent_allows_auction)
    && config.ad_templates_enabled
    && has_matched_slots
    && config.auction_enabled
```

That is a real-browser, non-bot, non-prefetch navigation GET with matched slots and consent — essentially the same population as preload eligibility (`is_navigation && is_get && has_ec_id && has_kv`). So on any page that actually serves ads, the skip is cancelled and the KV read happens anyway.

The new tests don't surface this because they all pass `slots: &[]` with `create_test_settings()`, so `matched_slots` is empty and `should_run_auction` is false. I re-ran the same `run_marker_lookup_probe` shape using `scheduling_settings()` (auction enabled, `creative_opportunities` configured) and a single real `scheduling_slot()`:

```
valid_marker_lookups=1   invalid_marker_lookups=1
```

Identical — a valid marker saves nothing once one ad slot matches.

That leaves the marker's benefit confined to navigations with no matched slots. Worth confirming that population is large enough to justify a new signed browser cookie, the HMAC path, and the orphan-recovery regression flagged at `finalize.rs:97`.

If it isn't, the pull-sync half of this PR already delivers the #880 win on its own — `entry_is_pull_complete` in `build_pull_sync_context` plus `prepare_pull_sync_after_send` skip the post-send graph construction with no cookie involved. Worth considering as the smaller change.

Apply manually — this is a design question about the gate rather than a specific edit.

crates/trusted-server-core/src/ec/pull_sync_marker.rs
```diff
@@ -0,0 +1,646 @@
...
+    let version = segments.next()?;
+    let expires = segments.next()?;
+    let fingerprint = segments.next()?;
+    let tag = segments.next()?;
+    if segments.next().is_some() || version != MARKER_VERSION {
+        return None;
+    }
+
+    let expires_at = expires.parse::<u64>().ok()?;
+    if expires_at <= now || expires_at > now.saturating_add(MARKER_MAX_AGE_SECS) {
```
aram356: 🤔 **thinking** — Zero clock-skew tolerance on the upper expiry bound.

`set_marker` issues `expires_at = now + MARKER_MAX_AGE_SECS`, and this line rejects anything above `now + MARKER_MAX_AGE_SECS`. The accepted window is therefore exactly the issuance width, leaving no room for skew between edge POPs: a POP even one second behind the issuing POP rejects a freshly issued marker.

It fails closed — an `Invalid` marker just restores the normal KV read — so this is efficiency, not safety. But the cost isn't only a missed skip: `reconcile_marker` then expires the cookie, so a user bouncing between two slightly-skewed POPs re-mints and re-clears the marker repeatedly, and each `Set-Cookie` makes the response private via `enforce_set_cookie_cache_privacy`.

Widening the bound is safe because `expires_at` is inside the HMAC payload — a client cannot inflate it, so the upper bound only guards against a server that issued a bad expiry:

```rust
    let expires_at = expires.parse::<u64>().ok()?;
    // Allow a small clock-skew margin on the upper bound: `expires_at` is
    // covered by the HMAC, so a client cannot inflate it, and without this a
    // validating POP even one second behind the issuing POP rejects a freshly
    // issued marker and falls back to the KV read it was meant to avoid.
    let max_future = now.saturating_add(MARKER_MAX_AGE_SECS.saturating_add(MARKER_CLOCK_SKEW_SECS));
    if expires_at <= now || expires_at > max_future {
        return None;
    }
```

with `const MARKER_CLOCK_SKEW_SECS: u64 = 60;` alongside `MARKER_MAX_AGE_SECS`.

Apply manually — can't be auto-applied as a suggestion because it needs a new constant outside this hunk, and it also requires updating `marker_rejects_expired_overlong_and_tampered_values`. I verified that: with the tolerance applied, clippy is clean and 2686 tests pass, but that test's `validate_marker(..., 999)` case now succeeds, since a marker 3601s in the future falls inside the widened window. Its assertion message ("marker more than one hour in the future should fail") would need to move past the new bound.

crates/trusted-server-core/src/ec/pull_sync_marker.rs
```diff
@@ -0,0 +1,646 @@
...
+        Ok(value) => {
+            response.headers_mut().append(header::SET_COOKIE, value);
+        }
+        Err(err) => {
+            log::warn!("Skipping pull-sync marker cookie: invalid header value: {err}");
+        }
+    }
+}
+
+fn format_marker_cookie(value: &str, max_age: u64) -> String {
```
aram356, line 197: 🤔 **thinking** — Host-only scope is a load-bearing security invariant, but only an implicit one.

The marker cookie is host-only (no `Domain`), while `ts-ec` is domain-wide — `cookies.rs` emits `Domain={domain}` with `ec_cookie_domain()` returning `.{publisher.domain}`, so `ts-ec` is shared across all subdomains.

Since the request host is **not** part of `marker_payload` (only version, `ec_id`, `expires_at`, and the partner fingerprint), the absence of `Domain=` is the *only* thing preventing a marker minted on `a.pub.example` from validating on `b.pub.example` — the same `ts-ec` reaches both hosts and the passphrase is shared, so the tag verifies. It's correct today, and the `marker_cookie_is_host_only_and_secure` test pinning `!cookie.contains("Domain=")` is good defensive testing.

The risk is a future edit. Someone noticing that subdomain visitors always miss the marker could "fix" it by adding `Domain=` to match `ts-ec` — and silently create a cross-host replay primitive, with the existing test the only thing standing in the way.

Two options, in order of robustness:

```rust
// 1. Bind the host into the payload, which makes a later Domain= widening safe:
fn marker_payload(host: &str, ec_id: &str, expires_at: u64, fingerprint: &str) -> String {
    format!("{MARKER_VERSION}\0{host}\0{ec_id}\0{expires_at}\0{fingerprint}")
}

// 2. Or, at minimum, document the constraint here:
// Host-only by design: the request host is not part of the signed payload, so
// omitting `Domain` is what prevents a marker minted on one host from
// validating on a sibling host that shares the domain-wide `ts-ec` cookie.
```

`SameSite=Lax`, `Secure`, and `HttpOnly` all look right — `Strict` would break the feature on the top-level cross-site navigations it targets.

Apply manually — option 1 touches `create_marker`, `validate_marker`, and the call sites together; option 2 is a comment whose placement you may prefer to choose.

crates/trusted-server-core/src/ec/mod.rs
```diff
@@ -170,8 +174,17 @@ fn parse_ec_from_request(req: &Request<EdgeBody>) -> Result<RequestEc, Report<Tr
         .and_then(|j| j.get(COOKIE_TS_EC))
         .map(cookie::Cookie::value)
         .and_then(|value| request_ec_id_if_allowed(value, "ts-ec cookie"));
+    let pull_sync_marker = jar
+        .as_ref()
+        .and_then(|j| j.get(COOKIE_TS_EC_PULL_COMPLETE))
+        .map(cookie::Cookie::value)
+        .map(str::to_owned);
```
aram356, line 181: ♻️ **refactor** — The marker cookie is stored unvalidated, unlike its sibling one line above.

`ts-ec` on the preceding lines is filtered through `request_ec_id_if_allowed`, but the marker value goes straight to `.map(str::to_owned)` with no length or charset check. `MAX_MARKER_LENGTH` (256) is enforced only much later, inside `validate_marker`, so the unbounded value is allocated and stored on `EcContext` first.

The security impact is genuinely low, and I'd rather say so than overstate it: the memory is the attacker's own `Cookie` header, which the platform already caps, and it's a single `to_owned` of a slice the jar has already parsed — no amplification. Log injection is ruled out too, since the manual `Debug` impl prints `Unvalidated(<redacted>)` and nothing logs the raw value.

The argument is consistency. A reader of these two adjacent bindings sees one validated and one not, and has to open a different module to learn the second is bounded later:

```rust
    let pull_sync_marker = jar
        .as_ref()
        .and_then(|j| j.get(COOKIE_TS_EC_PULL_COMPLETE))
        .map(cookie::Cookie::value)
        .filter(|value| value.len() <= pull_sync_marker::MAX_MARKER_LENGTH)
        .map(str::to_owned);
```

which needs `MAX_MARKER_LENGTH` made `pub(crate)` (or a small `value_is_within_length_bound` helper exported from `pull_sync_marker`).

Apply manually — can't be auto-applied as a suggestion because it needs the visibility change in `pull_sync_marker.rs`. I verified the shape compiles and passes: clippy clean, 2687 core tests pass. (A single-file variant with a literal `256` also passes, but duplicating the constant seemed worse than a two-file change.)

crates/trusted-server-core/src/publisher.rs
```diff
@@ -4389,8 +4406,26 @@ pub async fn handle_publisher_request(
...
+        auction_needs_row: should_run_auction
+            && auction
+                .registry
+                .is_some_and(|registry| !registry.is_empty()),
+        eid_cookie_may_need_persistence,
+        privacy_needs_row: crate::ec::consent::ec_consent_withdrawn(&consent_context),
+        snapshot_already_read: !matches!(
```
aram356: ⛏ **nitpick** — `snapshot_already_read` is constant-false at this call site.

Nothing writes the EC snapshot between `EcContext` construction and this point on the navigation path, so `ec_context.kv_snapshot()` is always `NotRead` here and this field is always `false`. The hatch never fires in production; only the unit test exercises the `true` case.

Harmless, and defensible as a guard against a future reordering. But combined with `privacy_needs_row` — which is also redundant, since `finalize_unusable_consent` writes the tombstone directly and never reads the snapshot — it suggests the hatch list was assembled by enumerating plausible consumers rather than by tracing which states actually reach this line. That's the same process gap behind the orphan-recovery hole flagged at `ec/finalize.rs:97`, which is the one that does have teeth.

No change needed if the redundancy is deliberate; a short comment saying so would save the next reader the trace.

aram356 (changes requested):
## Summary

Adds a signed, one-hour browser marker asserting that the EC row already holds every pull-partner UID, and uses it to skip the KV snapshot preload and the post-send identity-graph construction. The cryptographic construction is solid — a proper two-level HMAC KDF with domain separation, constant-time tag verification, length-prefixed fingerprint hashing, a host-only `Secure`/`HttpOnly` cookie, and every rejection path failing closed into a full KV read. I found no forgery, replay, or injection vector.

The blocking findings are about behavior and payoff rather than crypto: orphan recovery stops firing on the skip path (a regression against `main`, and one the PR description says is preserved), and the `auction_needs_row` hatch cancels the skip on exactly the ad-serving navigations the change targets. Both were verified with scratch tests in a worktree at this head, not by inspection alone.

> None of the inline comments below carry a one-click `suggestion`. Scratch verification demoted both suggestion candidates to prose: the clock-skew fix breaks an existing assertion in a different range, and the cookie-bound fix needs a shared constant that lives outside the hunk. Each comment gives the proposed code in a plain block to apply manually.

## Blocking

### 🔧 wrench

- **Orphan recovery silently stops firing on the marker-skip path** — see inline at `crates/trusted-server-core/src/ec/finalize.rs:97`
- **The optimization does not fire on ad-serving navigations** — see inline at `crates/trusted-server-core/src/publisher.rs:4417`

### ❓ question

- **What is the measured KV-read reduction in production?** — see Cross-cutting below

## Non-blocking

### 🤔 thinking / ♻️ refactor / ⛏ nitpick

- **Zero clock-skew tolerance on the marker expiry upper bound** — see inline at `crates/trusted-server-core/src/ec/pull_sync_marker.rs:263`
- **Host-only cookie scope is an unstated security invariant** — see inline at `crates/trusted-server-core/src/ec/pull_sync_marker.rs:196`
- **Marker cookie is stored unvalidated at the parse site** — see inline at `crates/trusted-server-core/src/ec/mod.rs:181`
- **`snapshot_already_read` is constant-false at the call site** — see inline at `crates/trusted-server-core/src/publisher.rs:4424`

- ❓ **What is the measured KV-read reduction in production?** — Findings 1 and 2 together mean the change trades an orphan-recovery regression for a benefit that, in the tests as written, only materialises on navigations with no matched ad slots. Before merging ~1,100 lines and a new signed browser cookie, could you share the measured before/after KV read counts on real traffic?

  Note that the pull-sync half of this PR stands on its own and looks right independently: the `entry_is_pull_complete` short-circuit in `build_pull_sync_context` (`crates/trusted-server-core/src/ec/pull_sync.rs:76-81`) plus `prepare_pull_sync_after_send` (`crates/trusted-server-adapter-fastly/src/main.rs:355`) already skip the post-send graph construction when there is nothing to pull, with no cookie, no HMAC, and no snapshot-skip involved. If the marker's measured win is small, closing #880 with just that part would avoid both blocking findings entirely.

- 📝 **Verification performed for this review** — Ran in a clean worktree at `4bd7adad`: `cargo fmt --all -- --check` (clean), `cargo clippy-fastly`, `cargo clippy-axum`, `cargo test-axum` (26 passed), `cargo test -p trusted-server-core --lib` (2687 passed). Two scratch reproductions confirmed findings 1 and 2; both were reverted and the worktree restored to the PR head.

- 📝 **Angles checked and cleared** — For the record, these were investigated and found sound, so they need no action: the `expect()` at `pull_sync_marker.rs:135` is provably unreachable (the guards at lines 121 and 126 make `entry_for` total on that arm); the three `HmacSha256::new_from_slice(...).expect(...)` calls cannot fail (HMAC accepts any key length, and a ≥32-character passphrase is separately enforced at config load in `settings.rs:611-631`); the `\0`-separated marker payload is not ambiguous because `ec_id` is `is_valid_ec_id`-filtered before it reaches `ec_value`; the non-constant-time fingerprint comparison before `verify_slice` leaks only a public config digest, not keying material; no new dependencies were added (`hex`, `hmac`, `sha2` were already present); and the doc's cache-privacy claim is accurate — `enforce_set_cookie_cache_privacy` does downgrade any `Set-Cookie`-bearing response.

## CI Status

- cargo fmt: PASS (required)
- cargo test: PASS (required)
- cargo test (cloudflare native + wasm32-unknown-unknown) / cargo check: PASS
- vitest: PASS
- Analyze (actions): PASS
- Analyze (rust): PENDING
- integration tests: PENDING
- integration tests (Fastly EC lifecycle): PENDING
- CodeQL: SKIPPED

No failing checks at review time.

crates/trusted-server-core/src/ec/pull_sync.rs
ChristianPavilonis: Addressed in ae1c472f. The public documentation now lists every `None` case: no pull-enabled partners, denied consent, no valid active EC ID, and no consented snapshot row with a missing pull-partner ID.

crates/trusted-server-core/src/ec/pull_sync.rs
ChristianPavilonis: Addressed in ae1c472f. The documentation now states that completeness requires a consented live row with every pull-enabled partner ID and always returns false for an empty pull registry.

crates/trusted-server-core/src/ec/pull_sync_marker.rs
ChristianPavilonis: Addressed in ae1c472f. Added `steady_states_leave_the_marker_cookie_untouched`, covering both a valid marker with a complete row and an absent marker with an incomplete row. Both cases assert that reconciliation emits no marker `Set-Cookie`.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Addressed in ae1c472f. Added a comment at the preload decision documenting that it also gates the concurrent origin send and why a marker skip cannot currently sit in front of an auction. The new matched-slot test also pins the auction side of that invariant.

crates/trusted-server-core/src/ec/finalize.rs
ChristianPavilonis: Addressed in ae1c472f by making the bounded deferral explicit rather than adding the proposed hatch. The PR body, Edge Cookie guide, and plan now state that a valid marker acts as recent existence proof and defers detection of a subsequently deleted row until marker expiry. Added `valid_marker_with_unread_snapshot_defers_orphan_recovery` to pin that behavior; the existing expired-marker lookup and missing-row recovery tests cover restoration after expiry. A direct `recovery_eligible` hatch is not viable at this decision point because Fastly authorizes recovery only after the publisher handler succeeds, and applying equivalent navigation eligibility here would cancel the marker skip for its target traffic.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Addressed in ae1c472f by making the boundary executable. Added `valid_marker_does_not_skip_auction_snapshot_lookup` with an enabled auction and a matched slot; it proves the valid marker still performs one lookup when the auction needs the row. The hatch remains intentional because auction EID enrichment consumes the snapshot. The marker optimization is therefore explicitly limited to requests where pull-sync completeness is the only row consumer.

crates/trusted-server-core/src/ec/pull_sync_marker.rs
ChristianPavilonis: Addressed in ae1c472f. Marker validation now permits 60 seconds of issuer clock skew while retaining the signed upper bound. Added a failing-before regression for the 60-second boundary and moved the overlong-future assertion beyond the combined lifetime and skew allowance.

crates/trusted-server-core/src/ec/pull_sync_marker.rs
ChristianPavilonis: Addressed in ae1c472f with a comment beside cookie formatting explaining that the payload omits the request host, so the missing `Domain` attribute prevents replay across sibling hosts sharing the EC cookie and passphrase. The existing exact cookie test continues to enforce host-only scope. I kept host binding out of this PR because it would require request-host plumbing and a payload-version decision.

crates/trusted-server-core/src/ec/mod.rs
ChristianPavilonis: No code change here. I am keeping the length check at marker validation because an overlong present cookie must become `Invalid`, allowing reconciliation to expire it. Filtering it to `None` while parsing would turn it into `Absent` and leave the bad browser cookie in place. The copied value remains request-header bounded and redacted. Avoiding that copy without changing semantics would require a broader parser-state refactor, which is deferred.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Addressed in ae1c472f. Removed `snapshot_already_read` from `EcSnapshotPreloadInput`, the gate, and its unit test. It was constant-false at this call site and would have caused another preload rather than reusing a future preexisting snapshot.

aram356 (approved):

PR #901, opened 2026-07-13
Make EC withdrawal tombstoning idempotent across request bursts

ChristianPavilonis:
## Summary

- Make existing EC withdrawal tombstones a true no-op across repeated and bursty requests.
- Preserve the first tombstone's 24-hour TTL by avoiding backend work once authoritative tombstone state is observed.
- Keep existing-key-only privacy, CAS race handling, two-ID withdrawal, and best-effort browser response behavior intact.
- Use an exact completion-marker check to suppress stale-read rewrites without adding marker I/O to normal fresh-ID creation.

This PR is stacked on #900.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/ec/kv.rs` | Return matching tombstone snapshots before KV work, use exact marker-first stale-miss checks, keep fresh creation independent of markers, narrow tombstone APIs, and add operation/race/namespace coverage |
| `crates/trusted-server-core/src/ec/finalize.rs` | Add two-present-ID, repeated-withdrawal, and KV-failure cookie-deletion integration coverage |
| `crates/trusted-server-core/src/ec/mod.rs` | Document that snapshot generations do not independently authorize writes |
| `docs/guide/edge-cookies.md` | Document repeated-withdrawal TTL behavior |
| `docs/superpowers/plans/2026-07-13-issue-881-idempotent-withdrawal-tombstones.md` | Record the reviewed lifecycle semantics and verification contract |

## Closes

Closes #881

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy-fastly && cargo clippy-axum && cargo clippy-cloudflare && cargo clippy-cloudflare-wasm && cargo clippy-spin-native && cargo clippy-spin-wasm`
- [x] `cargo test-fastly && cargo test-axum && cargo test-cloudflare && cargo test-spin`
- [x] `./scripts/test-cli.sh`
- [x] `cargo test --manifest-path crates/trusted-server-integration-tests/Cargo.toml --test parity`
- [x] JS tests: `cd crates/trusted-server-js/lib && npx vitest run`
- [x] JS format: `cd crates/trusted-server-js/lib && npm run format`
- [x] JS build: `cd crates/trusted-server-js/lib && node build-all.mjs`
- [x] Docs format: `cd docs && npm run format`
- [x] WASM build: `cargo build --package trusted-server-adapter-fastly --release --target wasm32-wasip1`
- [ ] Manual testing via `fastly compute serve` — not run

## Checklist

[...]

PR #902, opened 2026-07-13
Group batch-sync mappings by EC ID before KV updates

### LINKED ISSUE #882: Group S2S batch-sync mappings by EC ID before KV updates
## Summary

Reduce S2S batch-sync KV operations by grouping validated mappings by normalized EC ID before updating the identity graph.

The endpoint accepts up to 1,000 mappings. Each valid mapping currently performs its own read-modify-write, so repeated mappings for one EC can produce repeated KV reads, writes, and CAS contention. In the worst changed-value case, one request performs roughly 2,000 KV operations before retries.

## Proposed direction

Validate mappings individually, group valid mappings by normalized EC ID, and apply the final partner UID for each distinct EC with one CAS-protected read-modify-write. Preserve per-input response accounting and deterministic last-write-wins behavior.

## Acceptance criteria

- KV read-modify-write count is bounded by the number of distinct valid EC IDs rather than mapping count.
- Validation errors remain associated with their original input indexes.
- For multiple valid mappings targeting the same EC and partner, the last valid mapping in request order is the persisted value.
- Missing and withdrawn entries reject every affected mapping consistently.
- Accepted/rejected counts and HTTP status behavior remain compatible with the current endpoint contract.
- Infrastructure-failure and abort semantics are explicitly documented and tested after grouping.
- Tests cover duplicate EC IDs, conflicting UIDs, mixed valid/invalid mappings, tombstones, missing entries, and CAS conflicts.

## Relevant code

- `crates/trusted-server-core/src/ec/batch_sync.rs`
- `crates/trusted-server-core/src/ec/kv.rs`

ChristianPavilonis:
## Summary

- Validate the full batch first, then group valid mappings by normalized EC ID in deterministic first-occurrence order.
- Apply only the last valid partner UID for each distinct EC through the existing CAS-protected writer while preserving per-input outcomes.
- Define and test grouped eligibility, error ordering, and infrastructure-abort accounting.

This PR is stacked on #901.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/ec/batch_sync.rs` | Group valid mappings before KV work, fan group outcomes back to original indexes, and add ordering/accounting/HTTP coverage |
| `crates/trusted-server-core/src/ec/kv.rs` | Add regression coverage for the existing conditional writer's CAS retry behavior |
| `docs/guide/api-reference.md` | Document normalized grouping, last-valid-wins, outcome fan-out, sorted errors, and groupwise failure behavior |
| `docs/guide/ec-setup-guide.md`, `docs/guide/edge-cookies.md` | Align operational guidance with the grouped contract |
| `docs/superpowers/plans/2026-07-13-issue-882-group-batch-sync-by-ec-id.md` | Record the reviewed design and verification contract |

## Closes

Closes #882

## Test plan

- [x] Other: focused grouped batch-sync/CAS tests, `cargo test-cloudflare`, `cargo test-spin`, cross-adapter parity, and all Cloudflare/Spin native + WASM clippy targets

## Checklist

- [x] New code has tests

### DIFF
--- a/crates/trusted-server-core/src/ec/batch_sync.rs
+++ b/crates/trusted-server-core/src/ec/batch_sync.rs
@@ -2,15 +2,18 @@
 //!
 //! Partners send authenticated batch ID sync requests via Bearer token.
 //! Each mapping associates an `ec_id` (`{64hex}.{6alnum}`)
-//! with the partner's user ID. Mappings are individually validated and
-//! written to the KV identity graph, with per-mapping rejection reasons
-//! reported in the response.
+//! with the partner's user ID. Mappings are individually validated, then valid
+//! mappings are grouped by normalized EC ID before one call to the KV update
+//! path per group.
+//! Responses still report outcomes per original mapping index.
 //!
 //! Mapping timestamps are retained in the request schema for client
 //! compatibility, but the EC identity graph no longer stores per-partner sync
-//! timestamps. Valid mappings therefore use idempotent last-write-wins
-//! semantics: unchanged UIDs are accepted without a write; different UIDs
-//! replace the stored value regardless of timestamp.
+//! timestamps. The last valid mapping in request order supplies each group's
+//! UID; unchanged UIDs are accepted without a write, and different UIDs replace
+//! the stored value regardless of timestamp.
+
+use std::collections::HashMap;
 
 use edgezero_core::body::Body as EdgeBody;
 use error_stack::{Report, ResultExt};
@@ -86,6 +89,17 @@ struct BatchSyncResponse {
 struct MappingError {
     index: usize,
     reason: &'static str,
+}
+
+/// Valid mappings sharing one normalized EC ID.
+///
+/// `indexes` stays in request order, while the groups themselves stay ordered
+/// by first valid occurrence. The lookup map used to locate this structure is
+/// never used for processing order.
+struct MappingGroup {
+    ec_id: String,
+    partner_uid: String,
+    indexes: Vec<usize>,
 }
 
 // ---------------------------------------------------------------------------
@@ -179,19 +193,28 @@ fn content_length_exceeds_limit(req: &Request<EdgeBody>, max_body_size: usize) -
         .is_some_and(|content_length| content_length > max_body_size)
 }
 
+/// Validates all mappings, then processes each normalized EC ID once.
+///
+/// Successful and eligibility outcomes fan out to every valid input in a
+/// group. On infrastructure failure, the failing group and all unprocessed
+/// valid groups are rejected as unavailable; already validated invalid inputs
+/// keep their specific errors. Errors are sorted by original input index.
 fn process_mappings(
     writer: &dyn BatchSyncWriter,
     partner_id: &str,
     mappings: &[SyncMapping],
 ) -> (usize, Vec<MappingError>) {
-    let mut accepted: usize = 0;
     let mut errors = Vec::new();
+    let mut groups: Vec<MappingGroup> = Vec::new();
+    let mut group_indexes: HashMap<String, usize> = HashMap::new();
 
-    for (idx, mapping) in mappings.iter().enumerate() {
+    // Validate all inputs before beginning KV work. The vector preserves group
+    // order; the map only locates an existing group in constant time.
+    for (index, mapping) in mappings.iter().enumerate() {
         let ec_id = normalize_ec_id_for_kv(&mapping.ec_id);
         if !is_valid_ec_id(&ec_id) {
             errors.push(MappingError {
-                index: idx,
+                index,
                 reason: REASON_INVALID_EC_ID,
             });
             continue;
@@ -199,42 +222,57 @@ fn process_mappings(
 
         if mapping.partner_uid.trim().is_empty() || mapping.partner_uid.len() > MAX_UID_LENGTH {
             errors.push(MappingError {
-                index: idx,
+                index,
                 reason: REASON_INVALID_PARTNER_UID,
             });
             continue;
         }
-        match writer.upsert_partner_id_if_exists(&ec_id, partner_id, &mapping.partner_uid) {
+
+        if let Some(&group_index) = group_indexes.get(&ec_id) {
+            let group = &mut groups[group_index];
+            group.partner_uid.clone_from(&mapping.partner_uid);
+            group.indexes.push(index);
+        } else {
+            group_indexes.insert(ec_id.clone(), groups.len());
+            groups.push(MappingGroup {
+                ec_id,
+                partner_uid: mapping.partner_uid.clone(),
+                indexes: vec![index],
+            });
+        }
+    }
+
+    let mut accepted = 0;
+    for (group_index, group) in groups.iter().enumerate() {
+        match writer.upsert_partner_id_if_exists(&group.ec_id, partner_id, &group.partner_uid) {
             Ok(UpsertResult::Written | UpsertResult::Unchanged) => {
-                accepted += 1;
+                accepted += group.indexes.len();
             }
             Ok(UpsertResult::NotFound | UpsertResult::ConsentWithdrawn) => {
-                errors.push(MappingError {
-                    index: idx,
+                errors.extend(group.indexes.iter().map(|&index| MappingError {
+                    index,
                     reason: REASON_INELIGIBLE,
-                });
+                }));
             }
             Err(err) => {
                 log::warn!(
-                    "Batch sync KV write failed for index {idx} (ec_id '{}'): {err:?}",
-                    log_id(&mapping.ec_id),
+                    "Batch sync KV write failed for group starting at index {} (ec_id '{}'): {err:?}",
+                    group.indexes[0],
+                    log_id(&group.ec_id),
                 );
-                errors.push(MappingError {
-                    index: idx,
-                    reason: REASON_KV_UNAVAILABLE,
-                });
-                // Abort remaining mappings on infrastructure failure.
-                for remaining_idx in (idx + 1)..mappings.len() {
-                    errors.push(MappingError {
-                        index: remaining_idx,
+                for unavailable_group in &groups[group_index..] {
+                    errors.extend(unavailable_group.indexes.iter().map(|&index| MappingError {
+                        index,
                         reason: REASON_KV_UNAVAILABLE,
-                    });
+                    }));
                 }
                 break;
             }
         }
     }
 
+    errors.sort_by_key(|error| error.index);
+    debug_assert_eq!(accepted + errors.len(), mappings.len());
     (accepted, errors)
 }
 
@@ -301,29 +339,47 @@ mod tests {
         }
     }
 
+    #[derive(Clone, Debug, PartialEq, Eq)]
+    struct WriterCall {
+        ec_id: String,
+        partner_id: String,
+        uid: String,
+    }
+
     struct MockWriter {
         results: std::cell::RefCell<VecDeque<Result<UpsertResult, Report<TrustedServerError>>>>,
+        calls: std::cell::RefCell<Vec<WriterCall>>,
     }
 
     impl MockWriter {
         fn new(results: Vec<Result<UpsertResult, Report<TrustedServerError>>>) -> Self {
             Self {
                 results: std::cell::RefCell::new(results.into()),
+                calls: std::cell::RefCell::new(Vec::new()),
             }
+        }
+
+        fn calls(&self) -> Vec<WriterCall> {
+            self.calls.borrow().clone()
         }
     }
 
     impl BatchSyncWriter for MockWriter {
         fn upsert_partner_id_if_exists(
             &self,
-            _ec_id: &str,
-            _partner_id: &str,
-            _uid: &str,
+            ec_id: &str,
+            partner_id: &str,
+            uid: &str,
         ) -> Result<UpsertResult, Report<TrustedServerError>> {
+            self.calls.borrow_mut().push(WriterCall {
+                ec_id: ec_id.to_owned(),
+                partner_id: partner_id.to_owned(),
+                uid: uid.to_owned(),
+            });
             self.results
                 .borrow_mut()
                 .pop_front()
-                .expect("should provide mock result for each mapping")
+                .expect("should provide mock result for each group")
         }
     }
 
@@ -359,6 +415,14 @@ mod tests {
             .header("authorization", "Bearer test-token-32-bytes-minimum-value")
             .body(EdgeBody::from(body.to_owned()))
             .expect("should build authorized batch request")
+    }
+
+    fn response_json(response: Response<EdgeBody>) -> serde_json::Value {
+        let body = response
+            .into_body()
+            .into_bytes()
+            .expect("should contain batch-sync response");
+        serde_json::from_slice(&body).expect("should serialize batch-sync response")
     }
 
     fn test_registry() -> PartnerRegistry {
@@ -578,8 +642,12 @@ mod tests {
             Ok(UpsertResult::NotFound),
             Ok(UpsertResult::ConsentWithdrawn),
         ]);
-        let ec_id = format!("{}.ABC123", "a".repeat(64));
-        let mappings = vec![mapping(&ec_id, "uid-1", 100), mapping(&ec_id, "uid-2", 101)];
+        let missing_ec_id = format!("{}.ABC123", "a".repeat(64));
+        let withdrawn_ec_id = format!("{}.ABC123", "b".repeat(64));
+        let mappings = vec![
+            mapping(&missing_ec_id, "uid-1", 100),
+            mapping(&withdrawn_ec_id, "uid-2", 101),
+        ];
 
         let (accepted, errors) = process_mappings(&writer, "partner", &mappings);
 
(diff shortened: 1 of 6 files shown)


crates/trusted-server-core/src/ec/batch_sync.rs
```diff
@@ -179,62 +193,86 @@ fn content_length_exceeds_limit(req: &Request<EdgeBody>, max_body_size: usize) -
...
                 reason: REASON_INVALID_PARTNER_UID,
             });
             continue;
         }
-        match writer.upsert_partner_id_if_exists(&ec_id, partner_id, &mapping.partner_uid) {
+
+        if let Some(&group_index) = group_indexes.get(&ec_id) {
+            let group = &mut groups[group_index];
+            group.partner_uid.clone_from(&mapping.partner_uid);
+            group.indexes.push(index);
```
prk-Jr, line 234: ♻️ **refactor** — Conflicting UIDs inside one group are discarded with no signal.

This is where a batch carrying two different UIDs for the same EC ID silently loses the earlier one. The final KV state matches the pre-grouping code, so it isn't a regression — but two different UIDs for one EC in one request is almost always a partner-side bug, and right now the response reports `accepted: 2` with nothing in the logs to notice it by. `debug` rather than `warn` so a partner that legitimately repeats mappings can't flood the log.

```suggestion
        if let Some(&group_index) = group_indexes.get(&ec_id) {
            let group = &mut groups[group_index];
            if group.partner_uid != mapping.partner_uid {
                log::debug!(
                    "Batch sync index {index} supplies a conflicting UID for ec_id '{}'; keeping the last valid value",
                    log_id(&ec_id),
                );
            }
            group.partner_uid.clone_from(&mapping.partner_uid);
            group.indexes.push(index);
```

_Scratch-verified in a worktree at `3a1e5de3`: `cargo fmt --all -- --check` clean, `cargo clippy -p trusted-server-core --target wasm32-wasip1 --all-targets --all-features -- -D warnings` clean, `cargo test -p trusted-server-core --lib ec::` 349 pass, no post-verification drift._

crates/trusted-server-core/src/ec/batch_sync.rs
```diff
@@ -361,6 +417,14 @@ mod tests {
             .expect("should build authorized batch request")
     }
 
+    fn response_json(response: Response<EdgeBody>) -> serde_json::Value {
+        let body = response
+            .into_body()
+            .into_bytes()
+            .expect("should contain batch-sync response");
+        serde_json::from_slice(&body).expect("should serialize batch-sync response")
```
prk-Jr: ⛏ **nitpick** — This `expect` guards a deserialize, but the message says "serialize".

```suggestion
        serde_json::from_slice(&body).expect("should deserialize batch-sync response")
```

The same swap appears twice more, at `:906-907` (`"should serialize grouped success response"`) and `:927-928` (`"should serialize grouped multi-status response"`) — both also on `serde_json::from_slice`. Those need the same one-word fix applied manually; they're separate ranges and can't ride this suggestion.

_Scratch-verified in a worktree at `3a1e5de3`: `cargo fmt --all -- --check` clean, clippy clean on `wasm32-wasip1 --all-targets`, 22 batch-sync tests pass, no post-verification drift._

docs/guide/api-reference.md
```diff
@@ -158,10 +158,36 @@ re-evaluated within the recheck window).
...
+  Missing and withdrawn EC entries reject every valid mapping in their group as
+  `ineligible`.
+- If a KV infrastructure failure occurs, every valid mapping in the failing
+  group and each unprocessed valid group is rejected as `kv_unavailable`; no
+  later group is updated. Already processed groups keep their outcomes, and
+  validation errors are preserved.
+- `errors` is sorted by original input index. Therefore each input has exactly
+  one outcome and `accepted + rejected` equals the number of submitted
+  mappings. The endpoint returns `200 OK` only when all mappings are accepted;
+  otherwise it returns `207 Multi-Status`.
```
prk-Jr: ⛏ **nitpick** — "Therefore" points the implication the wrong way: sort order doesn't establish one-outcome-per-input. The exactly-one-outcome property is what the code guarantees (each index reaches either a validation error or exactly one group), and `accepted + rejected == mappings.len()` follows from that; the sort is an independent, separate guarantee about `errors` ordering.

```suggestion
- Each input receives exactly one outcome, so `accepted + rejected` equals the
  number of submitted mappings. `errors` is sorted by original input index. The
  endpoint returns `200 OK` only when all mappings are accepted; otherwise it
  returns `207 Multi-Status`.
```

_Scratch-verified in a worktree at `3a1e5de3`: the repo's pinned docs prettier 3.8.1 `--check guide/api-reference.md` passes, no post-verification drift._

prk-Jr (commented):
## Summary

Grouping S2S batch-sync mappings by normalized EC ID before touching KV is correct as implemented, and the accounting invariant holds by construction: every input index lands in exactly one bucket (a validation error, or a single group that is accepted or fanned out as an error), and the abort slice `groups[group_index..]` includes the failing group, so `accepted + errors.len() == mappings.len()`. I traced the duplicate-UID orderings by hand and the final KV state is identical to the pre-grouping code — fewer writes, same result. No blocking findings.

> 3 of the inline comments below carry a one-click GitHub `suggestion` — use **Commit suggestion** (or **Add suggestion to batch** for several at once) to apply them as commits on the PR branch. The remaining comment describes the fix in prose because splitting the bundled test spans two ranges and adds a new test function, which can't be auto-applied.

## Non-blocking

### ♻️ refactor

- **Conflicting UIDs inside one group are discarded with no signal** — see inline at `crates/trusted-server-core/src/ec/batch_sync.rs:231`
- **`handle_batch_sync_reports_grouped_success_and_rejection_counts` bundles two tests** — see inline at `crates/trusted-server-core/src/ec/batch_sync.rs:883`

### ⛏ nitpick

- **`expect` message says "serialize" on a deserialize** — see inline at `crates/trusted-server-core/src/ec/batch_sync.rs:425`
- **"Therefore" is a non-sequitur in the docs contract** — see inline at `docs/guide/api-reference.md:182`

- ⛏ **Stale third mock result in `process_mappings_aborts_on_kv_unavailable`** (`crates/trusted-server-core/src/ec/batch_sync.rs:544-572`) — three distinct EC IDs now produce two writer calls, so the queued third `Ok(UpsertResult::Written)` is dead. The test still catches a missing `break` (a third call would pop it and push `accepted` to 2, failing the `accepted == 1` assertion), but the leftover misrepresents the new call shape. Drop it and add `assert_eq!(writer.calls().len(), 2, "should stop after the failing group")` — this PR added call recording precisely for that. These lines are unchanged by the diff, so there is no hunk to anchor an inline comment to.

- ♻️ **No test at the batch size the issue is about** — issue #882's acceptance criterion is "work bounded by distinct valid normalized EC IDs", but the largest duplicate case covered is 3 mappings. A `MAX_BATCH_SIZE`-sized batch of one repeated EC ID asserting `writer.calls().len() == 1` would pin the actual performance contract rather than inferring it from the 2-3 mapping cases. The existing grouping tests prove the mechanism; this would prove the bound.

- 🤔 **No CHANGELOG entry for a partner-visible contract change** — `CHANGELOG.md` is actively maintained (59 bullets under `[Unreleased]`, including endpoint-level behavior entries such as the admin Basic-auth coverage change) and follows Keep a Changelog's "all notable changes". This PR changes the documented `/_ts/api/v1/batch-sync` contract in two partner-observable ways: `accepted` now counts group members rather than individual writes, and the abort boundary moved from positional to groupwise, so `kv_unavailable` and accepted indexes can interleave (a partner retrying "from the first `kv_unavailable` index onward" will now re-send already-accepted mappings — harmless because the write is idempotent, but a behavior change to reason about). Neither this PR nor #901 touches the file. Worth a `### Changed` bullet, or an explicit decision that batch-sync internals stay out of the changelog.

- 📝 **Three-deep stack** — 902 → 901 (`fix/idempotent-ec-withdrawal-tombstones`, open) → 900 (`fix/no-op-kv-reads`) → `main`. The diff reviewed here is the top delta only; merge order matters.

- 👍 **Load-bearing details that were easy to get wrong** — the explicit `errors.sort_by_key` plus the `debug_assert_eq!` accounting invariant: the sort is genuinely required now that abort fan-out appends group indexes out of order, and the assertion documents the one-outcome-per-input property at the point it could break. The new `upsert_partner_id_if_exists_retries_cas_conflict` test in `kv.rs` pins the CAS boundary the grouping now leans on and was previously uncovered by any `upsert_partner_id_if_exists` test. And documenting the `A, B, A` boundary change as intentional in both the API reference and the plan, rather than leaving it implicit, is the right call for a contract partners integrate against.

## CI Status

All 14 reported checks pass. `gh pr checks --required` reports no required checks on this branch, so none of these is merge-blocking under branch protection.

- cargo fmt: PASS
- cargo test: PASS
- vitest: PASS
- format-typescript: PASS
- format-docs: PASS
- integration tests: PASS

Locally, in a worktree at `3a1e5de3`: `cargo test -p trusted-server-core --lib ec::` (349 pass), `cargo clippy -p trusted-server-core --target wasm32-wasip1 --all-targets --all-features -- -D warnings` (clean), `cargo fmt --all -- --check` (clean), and the pinned docs prettier 3.8.1 `--check` (clean).

crates/trusted-server-core/src/ec/batch_sync.rs
```diff
@@ -622,5 +692,342 @@ mod tests {
...
+        );
+        assert_eq!(
+            writer.calls().len(),
+            2,
+            "should stop after the failing group"
+        );
+    }
+
+    #[test]
+    fn handle_batch_sync_reports_grouped_success_and_rejection_counts() {
```
prk-Jr: ♻️ **refactor** — This test bundles two independent scenarios into one `#[test]`: a written group returning `200 OK`, and an ineligible group returning `207 Multi-Status`. Two arrange-act-assert cycles in one test means a failure in the first half hides the second, and `success_body` / `rejected_body` are each shadowed from request JSON into response bytes partway through.

Both halves also hand-roll the `response_json` helper this PR adds at `:420` — which `handle_batch_sync_reports_validation_errors_without_writer_calls` does use. Splitting them makes the helper pay for itself:

```rust
#[test]
fn handle_batch_sync_accepts_every_member_of_a_written_group() {
    let registry = test_registry();
    let limiter = MockRateLimiter {
        should_exceed: false,
    };
    let ec_id = format!("{}.ABC123", "a".repeat(64));
    let writer = MockWriter::new(vec![Ok(UpsertResult::Written)]);
    let body = format!(
        r#"{{"mappings":[{{"ec_id":"{ec_id}","partner_uid":"one","timestamp":1}},{{"ec_id":"{ec_id}","partner_uid":"two","timestamp":2}}]}}"#
    );

    let response =
        handle_batch_sync_with_writer(&writer, &registry, &limiter, authorized_batch_request(&body))
            .expect("should return success response");

    assert_eq!(response.status(), StatusCode::OK);
    let json = response_json(response);
    assert_eq!(json["accepted"], 2);
    assert_eq!(json["rejected"], 0);
}

#[test]
fn handle_batch_sync_rejects_every_member_of_an_ineligible_group() {
    // … same shape with Ok(UpsertResult::NotFound), MULTI_STATUS, and the
    // existing errors assertion.
}
```

Apply manually — can't be auto-applied as a suggestion because the change spans two separate ranges in this test and introduces a second test function.

prk-Jr (approved):
Approving. Nothing in the earlier review was blocking: the grouping logic is correct, every input index resolves to exactly one outcome by construction, and the final KV state matches the pre-grouping code for every duplicate-UID ordering — with one writer call per distinct normalized EC ID instead of one per mapping.

The findings in https://github.com/IABTechLab/trusted-server/pull/902#pullrequestreview-5114965721 stand as non-blocking follow-ups, at the author's discretion: three one-click suggestions (a `debug` log when a batch supplies conflicting UIDs for one EC, an `expect` message that says "serialize" on a deserialize, and a docs sentence whose "Therefore" points the implication backwards), plus test-hygiene and CHANGELOG notes in the review body.

CI is green across all 14 reported checks.

aram356:
@ChristianPavilonis please assign issue to this PR

crates/trusted-server-core/src/ec/batch_sync.rs
```diff
@@ -179,62 +193,86 @@ fn content_length_exceeds_limit(req: &Request<EdgeBody>, max_body_size: usize) -
...
-                    });
+                    }));
                 }
                 break;
             }
         }
     }
 
+    errors.sort_by_key(|error| error.index);
+    debug_assert_eq!(accepted + errors.len(), mappings.len());
```
aram356: ⛏ **nitpick** — CLAUDE.md asks for descriptive assertion messages, and every other `debug_assert_eq!` in the codebase carries one (`publisher.rs:3021`, `openrtb.rs:480`, `registry.rs:823`). This is the only bare instance, and it encodes this PR's central accounting invariant — that each submitted mapping yields exactly one outcome — so it is worth naming for whoever trips it.

```suggestion
    debug_assert_eq!(
        accepted + errors.len(),
        mappings.len(),
        "should report exactly one outcome per submitted mapping"
    );
```

_Scratch-verified at the PR head: `cargo fmt --all -- --check` clean, compiles, and all 22 `batch_sync` tests pass (with the inherited `kv.rs:1591` import error patched out locally so the test target builds). No drift between the approved bytes and the verified tree._

aram356 (changes requested):
## Summary

The grouping rework is well-executed: validation-before-write, the vector-plus-map split for deterministic first-occurrence ordering, group outcome fan-out, and the per-input accounting invariant are all correct, and the 22 tests cover the contract thoroughly. The docs and the recorded design plan match the implementation exactly.

The blocking issue is CI state rather than the algorithm: the required `cargo test` and `cargo fmt` checks are both red, and as a consequence none of this PR's new tests have actually executed in CI.

> 1 of the inline comments below carries a one-click GitHub `suggestion` — use **Commit suggestion** to apply it as a commit on the PR branch. The remaining findings are prose because they concern CI state or span more than the diff.

## Blocking

### 🔧 wrench

- **CI is red: duplicate `InMemoryEcKv` import breaks the test build** — see "Cross-cutting / body-level findings" below

## Non-blocking

### ⛏ nitpick

- **`debug_assert_eq!` missing the descriptive message CLAUDE.md requires** — see inline at `crates/trusted-server-core/src/ec/batch_sync.rs:275`

- 🔧 **CI is red: duplicate `InMemoryEcKv` import breaks the test build** — `cargo test` and `cargo fmt` are both FAIL, and both are required checks under branch protection. Both trace to a single compile error:

  ```
  error[E0252]: the name `InMemoryEcKv` is defined multiple times
      --> crates/trusted-server-core/src/ec/kv.rs:1591:9
       |
  1401 |     use crate::ec::kv_backend::test_support::InMemoryEcKv;
       |         ------------------------------------------------- previous import of the type `InMemoryEcKv` here
  ...
  1591 |     use crate::ec::kv_backend::test_support::InMemoryEcKv;
       |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `InMemoryEcKv` reimported here
  ```

  Two things worth separating:

  1. **This is inherited from the base branch, not introduced here.** This PR's diff touches zero `InMemoryEcKv` lines, and PR #901 shows the identical two failing checks. The one-line fix (delete `kv.rs:1591`) belongs in #901.
  2. **It means this PR's new tests have never run in CI.** The whole `trusted-server-core` lib-test target fails to compile, so all 22 new `batch_sync` tests plus the new `upsert_partner_id_if_exists_retries_cas_conflict` test were skipped. Verified locally: with line 1591 deleted, all 22 `batch_sync` tests pass.

  Also inherited from #901 and surfacing only once compilation is unblocked: `ec::finalize::tests::finalize_withdrawal_keeps_cookie_deletion_on_kv_failure` fails on the base branch as well. Both need to clear in #901 before this PR can merge green.

  Finally, the PR description checks off `cargo test-fastly && cargo test-axum` and `cargo fmt --all -- --check` as passing. Those gates are currently red, so the test plan should be corrected.

- 👍 **Infrastructure-abort accounting is precisely handled** — `crates/trusted-server-core/src/ec/batch_sync.rs:263-269`. `groups[group_index..]` includes the failing group and then breaks, so no input is counted twice, and pre-validated inputs keep their specific error reasons instead of being flattened to `kv_unavailable` the way the old positional abort did. The interaction with the trailing `errors.sort_by_key` is subtle, and `process_mappings_aborts_by_group_and_preserves_sorted_accounting` pins it exactly.

## CI Status

- cargo test: FAIL (required)
- cargo fmt: FAIL (required)
- vitest: PASS
- integration tests: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS

aram356 (approved):

crates/trusted-server-core/src/ec/batch_sync.rs
```diff
@@ -179,62 +193,90 @@ fn content_length_exceeds_limit(req: &Request<EdgeBody>, max_body_size: usize) -
...
             }
         }
     }
 
+    errors.sort_by_key(|error| error.index);
+    debug_assert_eq!(
+        accepted + errors.len(),
+        mappings.len(),
+        "should report exactly one outcome per submitted mapping"
+    );
```
aram356, line 279: Resolved — the bare `debug_assert_eq!` flagged in the previous pass now carries a descriptive message, matching CLAUDE.md and the other `debug_assert_eq!` sites in the codebase (`publisher.rs:3021`, `openrtb.rs:480`, `registry.rs:823`).

Verified at `e9557c86`: `cargo fmt --all -- --check` clean, `cargo clippy-fastly` clean, and `cargo test -p trusted-server-core --lib` passes 2729 tests with 0 failures.

crates/trusted-server-core/src/ec/batch_sync.rs
```diff
@@ -505,6 +572,11 @@ mod tests {
         assert_eq!(errors[0].reason, REASON_KV_UNAVAILABLE);
         assert_eq!(errors[1].index, 2);
         assert_eq!(errors[1].reason, REASON_KV_UNAVAILABLE);
+        assert_eq!(
+            writer.calls().len(),
+            2,
+            "should stop after the failing group"
```
aram356, line 578: Resolved — this stale third queued result (`Ok(UpsertResult::Written)`) could never be consumed, so the test would still have passed if the abort had failed to stop group processing. Removing it plus the new `writer.calls().len() == 2` assertion now pins the abort boundary directly.

This was a latent gap in the previous revision that the prior review pass did not catch.

PR #903, opened 2026-07-13
Complete removal of the legacy consent KV path

### LINKED ISSUE #883: Complete removal of the legacy consent KV persistence path
## Summary

Complete the legacy `consent_store` removal described and closed in #666.

The current code still contains consent KV load/save/delete helpers and `ConsentPipelineInput` plumbing, but production `EcContext` construction passes `ec_id: None` and `kv_store: None`, so those operations are not used. On the Fastly EdgeZero path, a configured `consent_store` can still be opened as a fail-closed route dependency even though the opened store is not wired into consent load/save.

This leaves dead persistence code, misleading documentation/configuration, and a store availability dependency that does not provide the advertised consent continuity.

## Proposed direction

Follow the decision already recorded in #666: request cookies/headers/geo are the live consent source, while EC identity lifecycle and withdrawal state live in the EC identity store. Remove the remaining separate consent-KV path rather than accidentally wiring its read-before-write behavior onto the request hot path.

If persisted consent continuity has become a product requirement since #666, stop and define that contract explicitly before implementation instead of preserving the current half-wired state.

## Acceptance criteria

- Remove `consent_store` from active configuration, examples, manifests, and adapter route dependencies.
- Remove unused consent KV fields and branches from `ConsentPipelineInput` and `build_consent_context`.
- Remove `crates/trusted-server-core/src/storage/kv_store.rs` and tests when no live caller remains.
- EdgeZero consent-dependent routes no longer open an otherwise-unused consent KV store.
- Documentation clearly states the authoritative sources for live consent and withdrawal enforcement.
- Existing consent gating and EC tombstone behavior remain unchanged.
- Migration/config parsing behavior for deployments that still contain `consent_store` is documented and tested.

## Relevant code

- `crates/trusted-server-core/src/consent/mod.rs`
- `crates/trusted-server-core/src/consent_config.rs`
- `crates/trusted-server-core/src/storage/kv_store.rs`
- `crates/trusted-server-core/src/ec/mod.rs`

ChristianPavilonis:
## Summary

- Remove the inactive legacy consent KV fallback/write path, storage module, source metadata, and Fastly route dependency.
- Keep live consent request-local and preserve `ec.ec_store` as the sole KV-backed EC identity/withdrawal store.
- Fail fast when stale TOML or JSON/app-config still contains `consent_store`, with explicit operator migration guidance.

This PR is stacked on #902.

## Changes

| Area | Change |
| ---- | ------ |
| Consent core | Remove KV fields/branches from `ConsentPipelineInput`, delete the dead storage module, and remove `ConsentSource::KvStore` |
| Configuration | Remove `ConsentConfig::consent_store` and add strict TOML/runtime-JSON rejection tests |
| Fastly | Stop opening an unused consent store for auction, page-bids, and publisher routes; remove the local fixture and named opener |
| Other adapters | Correct generic platform-KV comments that still advertised consent persistence |
| Documentation | Explain request-local consent authority, strict upgrade failure, binding cleanup, rollback retention, and why legacy records are not migrated into `ec.ec_store` |

### Migration note

Deployments that still configure `[consent].consent_store` must remove the field before upgrading. Strict config loading intentionally rejects it. The old store may be retained unchanged for a rollback window, then unlinked and deleted; its records must not be copied into `ec.ec_store`.

Before this change, Fastly returned 503 on auction, page-bids, and publisher routes when a configured legacy store could not be opened, even though consent code never read that store. After this removal, a stale `consent_store` field instead prevents normal startup, and Fastly's startup-error router returns 500 on registered user routes until the configuration is corrected.

## Closes

Closes #883

## Test plan

- [x] Other: strict TOML/JSON migration tests, consent and Fastly dispatch tests, `cargo test-cloudflare`, `cargo test-spin`, parity, `./scripts/test-cli.sh`, all Cloudflare/Spin clippy targets, and all-features core rustdoc

## Checklist

- [x] New code has tests


aram356:
@ChristianPavilonis please assign issue for this PR

crates/trusted-server-adapter-spin/src/platform.rs
```diff
@@ -209,14 +209,9 @@ fn spin_secret_variable_name(
...
-/// record — the privacy-safe failure mode. Consequently TTL-backed consent
-/// persistence (`save_consent_to_kv`) is unavailable on Spin: each write returns
-/// the error, which the core caller logs and treats as non-fatal (consistent
-/// with all adapters — failing to persist consent never breaks the request, and
-/// not persisting is the safe direction). Operators configuring
-/// `settings.consent.consent_store` on Spin should expect stored-consent
-/// fallback not to function.
+/// returns `KvError::Validation` rather than silently writing a non-expiring
+/// record. Callers of the generic platform KV interface must handle that
+/// capability difference explicitly.
```
prk-Jr: ♻️ **refactor** — This sentence was rewritten in this PR but kept the wrong variant. The pinned edgezero `v0.0.4` (`9e661ae`) Spin KV store returns `KvError::Unsupported { operation: "put_bytes_with_ttl" }` for TTL writes (`edgezero-adapter-spin/src/key_value_store.rs:138`), and `KvError::Unsupported`'s own doc comment names "TTL writes on Spin" as its motivating example.

`KvError::Validation` is a different condition entirely: `KvHandle::put_bytes_with_ttl` runs `validate_key` / `validate_ttl` / `validate_value` before delegating, and those are what produce `Validation` — for a malformed key, TTL, or value, on any backend. `KvHandleAdapter` itself plain-delegates, so it never originates either error.

The distinction matters to callers beyond documentation accuracy: edgezero maps `KvError::Unsupported` to HTTP 501, and `Validation` does not, so a caller branching on the error kind to detect the capability gap would be looking for the wrong variant.

```suggestion
/// returns `KvError::Unsupported` (operation `put_bytes_with_ttl`) rather than
/// silently writing a non-expiring record. `KvError::Validation` is a different
/// condition: it comes from `KvHandle`'s key/TTL/value validators, not from the
/// Spin capability gap. Callers of the generic platform KV interface must handle
/// that difference explicitly.
```

_Verified in an isolated worktree at this head: `cargo fmt --all -- --check`, `cargo clippy-spin-native`, and `cargo clippy-spin-wasm` all clean. Worth checking rather than assuming, since `doc_markdown` is set to `warn` in `[workspace.lints.clippy]` and clippy runs with `-D warnings`._

docs/guide/configuration.md
```diff
@@ -498,7 +498,13 @@ TRUSTED_SERVER__TESTER_COOKIE__ENABLED=true
 
 ## EC Configuration
 
-Settings for Edge Cookie identifier generation. The `ec_store` KV store is the only KV-backed EC lifecycle store. It holds identity graph state, minimal consent metadata, source-domain keyed partner UIDs, and withdrawal tombstones. Consent configuration controls request-local interpretation and forwarding, not separate KV persistence.
+Settings for generating privacy-preserving Edge Cookie identifiers. The `ec_store` KV store is the only KV-backed EC lifecycle store; it holds identity graph state, minimal consent metadata, source-domain keyed partner UIDs, and withdrawal tombstones. Live consent is interpreted from request cookies, headers, geolocation, and policy defaults, not separate KV persistence.
+
+### Migrating from `consent_store`
+
+The legacy `[consent].consent_store` setting has been removed. Trusted Server uses a strict configuration schema, so TOML and JSON/app-config that still contain `consent_store` fail during configuration loading. Remove the field before upgrading; it is not accepted as an ignored or deprecated option.
```
prk-Jr: ♻️ **refactor** — "fail during configuration loading" reads like a scoped, consent-feature degradation, but the actual failure is total. Tracing it: `Settings::from_json_value` returns `TrustedServerError::Configuration`, whose `status_code()` is `INTERNAL_SERVER_ERROR` (`crates/trusted-server-core/src/error.rs:120`); that propagates out of `build_state`, which hands off to `startup_error_router`; and that router registers both `/` and `/{*rest}` across all seven `publisher_fallback_methods()` (`crates/trusted-server-adapter-fastly/src/app.rs:251`). So a leftover `consent_store` field means every request on every path returns 500 until the field is removed — not a degraded consent path.

That is worth stating plainly, because it changes how an operator sequences the upgrade. It also makes the pre-flight worth naming: `ts config validate` deserializes through the same strict `Settings` path (`TrustedServerAppConfig::deserialize` delegates to `Settings::deserialize`), so it catches a stale field before `ts config push` ever publishes the blob.

```suggestion
The legacy `[consent].consent_store` setting has been removed. Trusted Server uses a strict configuration schema, so TOML and JSON/app-config that still contain `consent_store` fail during configuration loading. Remove the field before upgrading; it is not accepted as an ignored or deprecated option. Configuration loading happens at startup, so a stale field is not a partial degradation: the service answers every request with an HTTP 500 startup error until the field is removed. Run `ts config validate` before `ts config push` to catch it ahead of deployment.
```

_Verified with the repo-pinned Prettier 3.8.1 `--check`; the single-line shape is required by `proseWrap: "preserve"` in `docs/.prettierrc`._

docs/guide/fastly.md
```diff
@@ -302,7 +302,9 @@ Verify stores are linked to your active service version:
 fastly resource-link list --service-id <service-id> --version <active-version>
 ```
 
-If EC sync returns `kv_unavailable` or identify responses are degraded, first check that the identity store is present and linked to the active version. Legacy partner/consent KV bindings can be removed once no deployment-specific tooling depends on them.
+If EC sync returns `kv_unavailable` or identify responses are degraded, first check that the identity store is present and linked to the active version.
+
+Before upgrading a deployment that used the legacy consent store, remove `[consent].consent_store` from TOML or JSON/app-config; strict configuration loading rejects the removed field. Remove its local Viceroy fixture and active Fastly resource link as well. The old records are not read or migrated into `ec.ec_store` and must not be copied there. You may retain the old store unchanged for a defined rollback window, then delete the store and its records.
```
prk-Jr: ♻️ **refactor** — This paragraph recommends retaining the old store for a rollback window, which is reasonable. Worth making explicit that nothing cleans it up in the meantime: during that window the store still holds consent records keyed by EC ID, and no code path deletes them when a user withdraws consent. `delete_consent_from_kv` existed for exactly that purpose but was never wired to a caller (I checked the base commit as well as this head), and this PR removes it along with the rest of the module. So withdrawal writes its tombstone to `ec.ec_store` and the retained legacy records are simply untouched.

Nothing in this release reads those records, so this is not a live exposure, and the property predates the PR. But since this is the paragraph telling operators to keep the store around, one sentence bounding the window and warning against re-pointing a service at it seems worth the space in a privacy-focused product.

```suggestion
Before upgrading a deployment that used the legacy consent store, remove `[consent].consent_store` from TOML or JSON/app-config; strict configuration loading rejects the removed field. Remove its local Viceroy fixture and active Fastly resource link as well. The old records are not read or migrated into `ec.ec_store` and must not be copied there. You may retain the old store unchanged for a defined rollback window, then delete the store and its records. Edge Cookie withdrawal does not delete anything from the retained legacy store, so keep the window short and do not re-point a service at it.
```

_Verified with the repo-pinned Prettier 3.8.1 `--check`._

prk-Jr (approved):
## Summary

Removes the legacy consent KV persistence path — config field, dead storage module, `ConsentSource::KvStore`, the Fastly consent-route store opener, and the local fixture — and replaces it with strict fail-fast config rejection plus operator migration guidance. I verified the central claim independently: at the base commit the sole production `ConsentPipelineInput` call site (`crates/trusted-server-core/src/ec/mod.rs:298`) already passed `ec_id: None, kv_store: None`, every `Some(...)` was a test, and `delete_consent_from_kv` had no callers at all. The path really was dead, so this is a pure-subtraction change with no live behavior removed.

A whole-tree residual sweep for `consent_store`, `consent_kv`, `ConsentSource::KvStore`, `storage::kv_store`, `open_kv_store`, and `runtime_services_for_consent_route` comes back clean — matches are limited to the strict-rejection tests, the migration docs, the new plan, and one spec already marked obsolete. Dropping the `ConsentSource::KvStore` variant carries no stored-record hazard (`ConsentSource` has no serde derives), no intra-doc link references a removed item, and the absence of `Cargo.toml`/`Cargo.lock` churn is correct since `hex` and `sha2` remain widely used. The `migration_guards.rs` entry removal is required, not optional — `include_str!` on the deleted paths would fail to compile. The deleted `edgezero_missing_consent_store_breaks_only_consent_routes` test also asserted admin-401; that coverage survives at `app.rs:1597` and `app.rs:2286`, so nothing of value was lost.

No blocking findings. Four non-blocking, three of them one-click suggestions.

> 3 of the inline comments below carry a one-click GitHub `suggestion` — use **Commit suggestion** (or **Add suggestion to batch** for several at once) to apply them as commits on the PR branch. All three were applied in an isolated worktree at this head and verified there: `cargo fmt --all -- --check`, `cargo clippy-spin-native`, and `cargo clippy-spin-wasm` clean for the Rust one; the repo-pinned Prettier 3.8.1 `--check` clean for the two Markdown ones; and a byte-exact before/after patch comparison confirmed verification did not mutate the tree beyond the reviewed bytes.

## Non-blocking

### ♻️ refactor

- **Spin TTL doc names the wrong `KvError` variant** — see inline at `crates/trusted-server-adapter-spin/src/platform.rs:212`
- **Migration doc understates the blast radius and omits the pre-flight command** — see inline at `docs/guide/configuration.md:505`
- **Retained legacy store is not covered by EC withdrawal deletion** — see inline at `docs/guide/fastly.md:307`

- 📌 **Generic `RuntimeServices` KV slot now has zero production consumers** — after this PR, `RuntimeServices::kv_store()`, `kv_handle()`, and `with_kv_store()` have no production callers; only tests reach them. The plan's contract item 7 states that retaining this infrastructure is deliberate and out of scope, so this is a follow-up rather than a change request. Two consequences seem worth tracking. Cloudflare (`crates/trusted-server-adapter-cloudflare/src/platform.rs:616`) and Spin (`crates/trusted-server-adapter-spin/src/platform.rs:726`) still resolve a real KV binding on every request to populate a slot nothing reads. And Fastly permanently populates it with `UnavailableKvStore` (`crates/trusted-server-adapter-fastly/src/app.rs:189`), so the first future consumer of the generic slot will silently degrade on the production adapter rather than fail loudly. A tracked issue would keep that from being rediscovered the hard way.

- 👍 **Test-first ordering on a deliberately breaking config change** — writing `settings_rejects_removed_consent_store_toml` / `_json` and confirming they failed *while the field still existed* is exactly the right sequencing for a strict-schema removal; it proves the guard tests the removal rather than merely passing alongside it. Covering both `Settings::from_toml` and `Settings::from_json_value` matters too, since the runtime blob path goes through the latter via `config_payload.rs:39`. The residual audit also holds up under an independent sweep, and dropping `runtime_services_for_consent_route` removed an entire nesting level from `dispatch_fallback` for free.

## CI Status

- cargo fmt: PASS (this job also runs all 8 target-matched clippy invocations plus the CLI and OpenRTB-codegen clippy steps)
- cargo test: PASS
- integration tests: PASS
- vitest: PASS
- format-typescript: PASS
- format-docs: PASS
- Analyze (rust) / CodeQL: not run (absent from `gh pr checks` output for this PR)

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -1587,9 +1587,6 @@ mod tests {
     // CAS-conflict injection tests
     // -----------------------------------------------------------------------
```
aram356, line 1586: 🌱 **seedling** — This import removal is unrelated to consent-store removal: it fixes a base-branch defect. On `perf/group-batch-sync-by-ec-id` the core test target fails to compile with `error[E0252]: InMemoryEcKv reimported here` (duplicate of the import at line 1401). Carrying the fix here is pragmatic since this PR can't be green otherwise, but it belongs in #902 where the duplicate was introduced. At minimum, mention it in the PR body so reviewers aren't surprised by an unexplained hunk in an otherwise-pure deletion.

crates/trusted-server-core/src/ec/finalize.rs
```diff
@@ -1638,6 +1638,9 @@ mod tests {
         };
         let mut ec_context =
             make_context_with_consent(Some(&ec_id), Some(&ec_id), true, false, consent);
+        ec_context.set_pull_sync_marker_for_test(
+            crate::ec::pull_sync_marker::PullSyncMarkerState::Invalid,
+        );
```
aram356, line 1695: 🌱 **seedling** — Like the `ec/kv.rs` import fix, this is a base-branch repair rather than consent-store work. I verified it's load-bearing: removing these three lines makes `finalize_withdrawal_keeps_cookie_deletion_on_kv_failure` fail on `KV failure should not prevent marker deletion`. The defect traces to the pull-sync-marker commits earlier in the stack (`a9551632e` / `be328a237`), not to this PR. Worth pushing down to #902, or calling out in the PR body — a test-only change inside a deletion PR reads as scope creep until you trace it.

crates/trusted-server-adapter-fastly/src/app.rs
```diff
@@ -147,7 +147,7 @@ use trusted_server_core::tester_cookie::{handle_clear_tester, handle_set_tester}
 use crate::middleware::{AuthMiddleware, FinalizeResponseMiddleware};
 use crate::platform::{
     FastlyPlatformBackend, FastlyPlatformConfigStore, FastlyPlatformGeo, FastlyPlatformHttpClient,
-    FastlyPlatformSecretStore, UnavailableKvStore, open_kv_store,
+    FastlyPlatformSecretStore, UnavailableKvStore,
```
aram356, line 150: 🤔 **thinking** — With `open_kv_store` and `runtime_services_for_consent_route` gone, every Fastly request now gets `UnavailableKvStore` in the generic `RuntimeServices` KV slot, and that slot has no production consumer left anywhere in the workspace — the only remaining `.kv_store()` call is an assertion in `adapter-spin/src/platform.rs:1064`. The plan doc's contract item 7 scopes this out as platform infrastructure, which is defensible, but a KV abstraction nothing reads tends to become permanent. A follow-up issue to either wire it to a real consumer or delete `with_kv_store` / `kv_store()` would keep it from ossifying.

crates/trusted-server-adapter-fastly/src/app.rs
```diff
@@ -691,7 +657,7 @@ async fn run_named_route(
                 ec.kv_graph.as_ref(),
                 registry_ref,
                 &mut ec.ec_context,
-                &consent_services,
+                services,
```
aram356, line 660: 📝 **note** — Worth spelling out in the PR body: this hunk removes a real behavioral guarantee, not just dead code. Auction, page-bids, and publisher routes previously returned **503 fail-closed** when a configured `consent_store` couldn't be opened, and two deleted tests asserted exactly that. Removing it is correct — the store was opened and never read, so failing closed protected nothing, and the path becomes unreachable once strict config rejects the field. But “stop opening an unused consent store” undersells it for anyone later bisecting an availability change.

aram356 (approved):

crates/trusted-server-adapter-fastly/src/app.rs
aram356, line 150: 🤔 **thinking** — Re-verified at `240f0f8e2`: with `open_kv_store` and `runtime_services_for_consent_route` gone, the generic `RuntimeServices` KV slot still has no production consumer anywhere in the workspace. The only remaining `.kv_store()` call is a test assertion in `adapter-spin/src/platform.rs:1065`, and `with_kv_store` has no non-test caller — every Fastly request now carries `UnavailableKvStore` in that slot.

The plan doc's contract item 7 scopes this out as platform infrastructure, which is a fair deferral. But I searched open and closed issues and found no follow-up filed, so nothing currently tracks it. A quick issue to either wire the slot to a real consumer or delete `with_kv_store` / `kv_store()` would keep it from ossifying as permanent dead plumbing. Not a merge blocker.

aram356 (approved):

PR #905, opened 2026-07-13
Inherit shared package metadata from the workspace

### LINKED ISSUE #904: Inherit shared package metadata from the workspace
Each crate currently repeats its own `version`, `edition`, `license`, `publish`, and `authors`, and the values have drifted: three crates sit on edition 2021 while seven are on 2024, some declare a license and some do not, and `trusted-server-js` is missing `publish = false`.

Consolidate all five keys into a `[workspace.package]` table and have every crate inherit them, so the package sections share one shape and one source of truth.

Two related cleanups:

- `trusted-server-openrtb-codegen` is not a workspace member (it is absent from `members` and nothing path-depends on it), so it cannot inherit anything or use workspace dependencies. Add it as a member while keeping it out of default builds.
- Inheriting `edition` unifies the workspace on 2024, which requires renaming a `gen` binding (reserved keyword) and collapsing the nested `if` / `if let` sites that clippy's `collapsible_if` flags once let-chains are stable.

aram356:
Closes #904.

## What

Adds a `[workspace.package]` table with `authors`, `edition`, `license`, `publish`, and `version`, and has all ten crates inherit every one of them. Each `[package]` section now follows the same shape:

```toml
[package]
name = "..."
description = "..."
authors = { workspace = true }
edition = { workspace = true }
license = { workspace = true }
publish = { workspace = true }
version = { workspace = true }
```

`trusted-server-openrtb-codegen` joins the workspace `members` so it shares those settings and can use workspace dependencies (`prost-build`). It stays out of default builds: `default-members` is still just the Fastly adapter, and every alias in `.cargo/config.toml` selects packages explicitly with `-p`, so no `build` / `check` / `test` / `clippy` command reaches it.

Nine crates gained a `description`; they had none before.

## Why the diff is 82 files

Inheriting `edition` moves `trusted-server-core`, `-adapter-fastly`, and `-js` from 2021 to 2024, joining the seven crates already there. Two mechanical consequences:

- **`gen` is a reserved keyword in edition 2024.** One binding renamed in `ec/kv.rs`.
- **Edition 2024 stabilizes let-chains**, so clippy `collapsible_if` now fires on nested `if` / `if let`. 41 sites collapsed via `cargo clippy --fix`, then reviewed.

The rest of the churn is `rustfmt` reordering imports, because its style edition follows the crate edition. No semantic change. (A `rustfmt.toml` with `style_edition = "2021"` would suppress that churn, at the cost of holding formatting deliberately out of step with the edition — happy to add it if preferred.)

## Verification

All green locally:

- clippy across all seven target/feature combos (fastly wasm, axum, cloudflare native + wasm, spin native + wasm, cli)
- `cargo test` for fastly (1,624 core tests + 99 + 21 + 2), axum, cloudflare, spin, and cli — 0 failures

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,9 @@ members = [
     "crates/trusted-server-integration-tests",
     "crates/trusted-server-js",
     "crates/trusted-server-openrtb",
+    # Manual protobuf codegen tool — a member so it shares workspace settings,
+    # but no default member or alias builds it. Run it explicitly with `-p`.
+    "crates/trusted-server-openrtb-codegen",
 ]
 
 # Viceroy (cargo test-fastly runner) calls `cargo run --bin trusted-server-adapter-fastly`
@@ -24,6 +27,13 @@ default-members = [
 
 [profile.release]
 debug = 1
+
+[workspace.package]
+authors = []
+edition = "2024"
+license = "Apache-2.0"
+publish = false
+version = "0.1.0"
 
 [workspace.dependencies]
 anyhow = "1"
@@ -72,6 +82,7 @@ log-fastly = "0.12"
 lol_html = "2.7.2"
 matchit = "0.9"
 mime = "0.3"
+prost-build = "0.13"
 rand = "0.8"
 rcgen = { version = "0.13", features = ["x509-parser"] }
 regex = "1.12.3"
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -53,6 +53,11 @@ jobs:
       - name: Run host-target CLI clippy
         run: cargo clippy --package trusted-server-cli --target x86_64-unknown-linux-gnu --all-targets --all-features -- -D warnings
 
+      # The codegen crate is a workspace member but is not a default member and
+      # is not covered by any adapter-scoped alias, so lint it explicitly.
+      - name: Run host-target OpenRTB codegen clippy
+        run: cargo clippy --package trusted-server-openrtb-codegen --target x86_64-unknown-linux-gnu --all-targets -- -D warnings
+
   format-typescript:
     runs-on: ubuntu-latest
     defaults:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -82,6 +82,12 @@ jobs:
       - name: Run host-target CLI tests
         run: ./scripts/test-cli.sh
 
+      # The codegen crate is not a default member and no adapter alias builds it,
+      # so its tests only run if invoked explicitly on the host target.
+      - name: Run host-target OpenRTB codegen tests
+        run: |
+          cargo test --package trusted-server-openrtb-codegen --target "$(rustc -vV | sed -n 's/host: //p')"
+
       - name: Verify Fastly WASM release build
         run: cargo build --package trusted-server-adapter-fastly --release --target wasm32-wasip1
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -454,8 +454,8 @@ version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "85a885520bf6249ab931a764ffdb87b0ceef48e6e7d807cfdb21b751e086e1ad"
 dependencies = [
- "prost",
- "prost-types",
+ "prost 0.14.4",
+ "prost-types 0.14.4",
  "tonic",
  "tonic-prost",
  "ureq",
@@ -471,7 +471,7 @@ dependencies = [
  "bollard-buildkit-proto",
  "bytes",
  "chrono",
- "prost",
+ "prost 0.14.4",
  "serde",
  "serde_json",
  "serde_repr",
@@ -767,7 +767,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.48.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -1677,7 +1677,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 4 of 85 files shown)


Cargo.toml
```diff
@@ -25,6 +28,13 @@ default-members = [
 [profile.release]
 debug = 1
 
+[workspace.package]
+authors = []
+edition = "2024"
```
ChristianPavilonis, line 33: 🌱 **P2 — Complete the Rust 2024 semantic-preservation pass**

Running the canonical `cargo fix --edition` migration from the clean base produces preservation changes that are absent here: `if let`/`else` rewrites in the Fastly middleware, auction orchestrator, APS, Sourcepoint, and settings paths, plus an explicit `use<>` capture bound on `Result<impl StreamProcessor>`. It also reports 23 distinct `tail-expr-drop-order` locations requiring review.

**Why it matters:** Rust 2024 changes `if let` temporary scope, tail-expression drop order, and return-position `impl Trait` lifetime capture. The affected code includes production auction, configuration, middleware, proxy, and streaming paths. The existing tests exercise only the new semantics, so they do not establish parity with Edition 2021.

**Suggested fix:** Run the official `cargo fix --edition` sequence for the three former Edition 2021 crates, retain or deliberately document the generated preservation changes, and audit the remaining drop-order warnings.

Cargo.toml
```diff
@@ -10,6 +10,9 @@ members = [
     "crates/trusted-server-integration-tests",
     "crates/trusted-server-js",
     "crates/trusted-server-openrtb",
+    # Manual protobuf codegen tool — a member so it shares workspace settings,
+    # but no default member or alias builds it. Run it explicitly with `-p`.
+    "crates/trusted-server-openrtb-codegen",
```
ChristianPavilonis, line 15: 🌱 **P2 — Add CI coverage for the new codegen workspace member**

The new `trusted-server-openrtb-codegen` member remains outside every scoped alias and CI test/Clippy command. The uncovered command currently fails:

```sh
cargo clippy --locked \
  -p trusted-server-openrtb-codegen \
  --target "$(rustc -vV | sed -n 's/host: //p')" \
  --all-targets -- -D warnings
```

It exits 101 with `clippy::items-after-test-module` at `crates/trusted-server-openrtb-codegen/src/main.rs:17`, because the `include!` at line 15 brings in a test module before `main`. Its seven tests pass, so the issue is specifically hidden lint coverage.

**Suggested fix:** Move the `include!` below `main`, then add native host-target Clippy and test steps for this member. Keeping it outside `default-members` is compatible with explicit CI coverage.

ChristianPavilonis (approved):
## Summary

Approved. The workspace metadata consolidation and let-chain rewrites look sound, and all 19 checks are green. I left two non-blocking follow-up notes inline.

prk-Jr (approved):

aram356:
Thanks — both notes were fair, and the first one caught a real gap. Addressed in 601c2675c.

## P2 — Rust 2024 semantic-preservation pass

You were right that I never ran `cargo fix --edition`; I went straight to the edition bump and leaned on the test suite, which (as you noted) only exercises the new semantics. I re-ran the canonical migration from clean `main` for the three former-2021 crates and worked through everything it reported.

**Applied:** the `use<>` capture bound on `create_html_stream_processor`. This is the one change with real teeth — the function takes five references and returns `impl StreamProcessor`, so under 2024 the opaque type implicitly captures every input lifetime, which would force callers to keep the settings and registry alive for as long as the processor. `+ use<>` restores the 2021 capture-nothing signature, with a doc comment explaining why.

**Audited and deliberately not taken:** the six `if let`/`else` rewrites. `cargo fix` turns them into `match x { Ok(v) => {..} _ => {..} }` to guard against the scrutinee temporary dropping before the `else` block. That only matters if the temporary has a side-effectful destructor. Here the temporaries are `Result<HeaderValue, InvalidHeaderValue>`, `Result<JsonValue, serde_json::Error>`, `Result<ApsBidResponse, _>`, and `Option<(String, Instant, Provider)>` — plain owned data — and the `else` blocks only log or build a fallback value. No locks, guards, or channels, so the rescope is unobservable, and the generated `match` blocks are strictly worse to read. Documented in the commit message rather than merged.

**Audited:** all 23 `tail-expr-drop-order` sites. The only destructors implicated are `Bytes` (38 occurrences) and `anyhow::Error` (8) — both pure deallocation — plus one `fastly::Body` at `platform.rs:345`, whose `BufWriter::drop` does flush. That one is the only side-effectful destructor in the set, and the temporary being reordered against it is the `Result<usize, _>` returned by `body.read(..)`, which merely deallocates. The flush still happens; only its ordering relative to freeing an unrelated `Result` changes, which nothing observes.

## P2 — CI coverage for the codegen member

Reproduced exactly as described: `clippy::items-after-test-module`, exit 101. Moved the `include!` below `main` (with a comment noting why it has to sit there) and added host-target clippy and test steps so the crate cannot silently rot again:

- `.github/workflows/format.yml` — host-target clippy for `trusted-server-openrtb-codegen`
- `.github/workflows/test.yml` — host-target `cargo test` for it in the native job

It stays out of `default-members`, as you suggested.

## Verification

`cargo fmt --check`; clippy across all eight target/feature combos including the new codegen one; and the full suite — fastly 1746, axum 32, cloudflare 32, spin 72, cli 108, codegen 7, parity 13. Zero failures.

Cargo.toml
aram356: Addressed in 601c2675c.

Ran the canonical `cargo fix --edition` from clean `main` for the three former-2021 crates.

**Applied:** `+ use<>` on `create_html_stream_processor`. It takes five references and returns `impl StreamProcessor`, so under 2024 the opaque type would capture every input lifetime and force callers to keep the settings and registry alive as long as the processor. `use<>` restores the 2021 capture-nothing signature.

**Audited, not taken:** the six `if let`/`else` rewrites. The generated `match` guards against the scrutinee temporary dropping before the `else` block, which only matters for a side-effectful destructor. These temporaries are `Result<HeaderValue, _>`, `Result<JsonValue, _>`, `Result<ApsBidResponse, _>`, and `Option<(String, Instant, Provider)>` — plain owned data — and the `else` blocks only log or build a fallback. No locks or guards, so the rescope is unobservable. Reasoning recorded in the commit message.

**Audited:** all 23 `tail-expr-drop-order` sites. Destructors involved are `Bytes` (38) and `anyhow::Error` (8), both pure deallocation, plus one `fastly::Body` at `platform.rs:345` whose `BufWriter::drop` does flush — but the temporary reordered against it is the `Result<usize, _>` from `body.read(..)`, which only deallocates. The flush still happens; nothing observes the reordering.

Cargo.toml
aram356: Addressed in 601c2675c. Reproduced exactly as described — `clippy::items-after-test-module`, exit 101.

Moved the `include!` below `main` (with a comment noting why it must sit there), and added host-target coverage so the crate cannot silently rot again:

- `.github/workflows/format.yml` — host-target clippy for `trusted-server-openrtb-codegen`
- `.github/workflows/test.yml` — host-target `cargo test` in the native job

Both now pass locally (clippy clean, 7 tests). The crate stays out of `default-members`, as you suggested.

PR #680, opened 2026-05-06
Implement server-side ad slot templates with PBS and APS auction

prk-Jr:
## Summary

- Adds server-side ad slot templates under `[creative_opportunities]` in `trusted-server.toml`. Matching slots are selected from the incoming document URL at the edge, and `window.tsjs.adSlots` is injected at `<head>` open so initial GPT setup does not need a separate slot-discovery request.
- Runs the server-side auction in parallel with the origin fetch for eligible document navigations. Configured providers such as Prebid Server and APS race under the creative-opportunity auction deadline; winning bids are price-bucketed and injected as `window.tsjs.bids` before `</body>`.
- Adds the `window.tsjs.adInit` GPT runtime path. It reads `window.tsjs.adSlots` and `window.tsjs.bids` synchronously, defines or reuses GPT slots, applies slot-level and `hb_*` targeting, sets `ts_initial=1`, refreshes the initial slots, and fires `nurl`/`burl` only after `slotRenderEnded` confirms the TS bid won via `hb_adid`.
- Adds PBS stored-request fallback keyed by slot ID when no inline PBS bidder params are present, filters non-PBS provider keys from PBS requests, and keeps bidder-param override rules for per-bidder/zone params.
- Adds per-bidder PBS win/billing suppression with `[integrations.prebid].suppress_nurl_bidders`, while retaining the deployment-wide `[integrations.prebid].suppress_nurl` compatibility switch.
- Improves the deferred slim-Prebid refresh path so GPT refresh auctions derive ad unit sizes and zone from injected `window.tsjs.adSlots` / GPT metadata instead of placeholder refresh sizes.
- Implements EID forwarding for the server-side auction path: reads the `ts-eids` cookie written by TSJS, gates EIDs by consent, and forwards them as OpenRTB `user.ext.eids` to PBS.
- Forwards real client IP and geo from Fastly `ClientInfo` into the server-side PBS request so bidders see the browser client context rather than the Fastly edge context.
- Keeps SPA/CSR route changes covered by the existing `GET /__ts/page-bids` hook, which updates `window.tsjs.adSlots` / `window.tsjs.bids` and re-runs `window.tsjs.adInit()` after `pushState`, `replaceState`, or `popstate` navigations.

## What the server-side auction sends to PBS

| Field | Value | Source |
| ----- | ----- | ------ |
| `user.id` | EC ID | `ts-ec` cookie or generated EC identity |
[...]

PR #910, opened 2026-07-14
Suppress fabricated empty Prebid bidder params

prk-Jr:
## Summary

Stacked on #899 (`feat/ssat-render-inline-creative`) — base branch is that PR's head, not `main`.

A configured Prebid bidder with no inline params and no matching bid-param override was expanded into `"bidder": {}` in the OpenRTB request. PBS rejects an empty bidder object, surfacing as the intermittent auction error.

## Fix

In `to_openrtb` ([`crates/trusted-server-core/src/integrations/prebid.rs`](crates/trusted-server-core/src/integrations/prebid.rs)):

- Track the bidders the publisher **explicitly** supplied (including an explicit empty `{}`).
- Reorder to: build bidder map → apply overrides → drop fabricated empties → decide stored-request.
- Drop any bidder still an empty object after overrides **unless** it was explicitly supplied.
- Recompute the PBS stored-request fallback *after* the drop, so a slot whose configured bidders all resolve to fabricated empties falls back to the stored-request/no-bid path instead of shipping invalid bidder objects.

Behavior contract:

- Absent bidder → dropped (never `"bidder": {}`).
- Publisher-supplied `{}` → preserved (misconfiguration stays visible).
- Bidder populated by a global/zone override → kept.
- Valid bidders still ship even when a sibling has no params.
- No eligible bidders → stored-request fallback.

The browser half of this defense (`prevBidderParams` retention on refresh) already shipped; this completes the server half.

## Tests

New unit tests in `prebid.rs`:

- `to_openrtb_drops_fabricated_empty_bidder_params`
- `to_openrtb_preserves_an_explicitly_empty_bidder`
- `to_openrtb_keeps_a_fabricated_bidder_that_an_override_populates`
- `to_openrtb_falls_back_to_stored_request_when_all_bidders_are_fabricated_empty`

## Verification

- `cargo fmt --all -- --check` ✓
- `cargo clippy-fastly` ✓
- `cargo test-fastly` ✓ (1628 core + 99 fastly, 0 failed)
- Other adapters (axum/cloudflare/spin) + clippy variants left to CI.

## Test plan

- [ ] CI green across all adapters
- [ ] Manual: slot with a configured bidder lacking params no longer emits `"bidder": {}` to PBS

### DIFF
--- a/crates/trusted-server-core/src/integrations/prebid.rs
+++ b/crates/trusted-server-core/src/integrations/prebid.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -1424,10 +1424,22 @@ impl PrebidAuctionProvider {
                 // Only pass through keys that are known PBS bidders — skip provider-specific
                 // keys like "aps" which belong to their own separate auction provider.
                 let mut bidder: HashMap<String, Json> = HashMap::new();
+                // Bidders the publisher explicitly supplied — including an
+                // explicit empty `{}`. A configured bidder that exists only
+                // because `expand_trusted_server_bidders` fabricated an empty
+                // params object is NOT explicit and must not ship as
+                // `"bidder": {}` (which PBS rejects).
+                let mut explicit_bidders: HashSet<String> = HashSet::new();
                 for (name, params) in &slot.bidders {
                     if name == TRUSTED_SERVER_BIDDER {
+                        if let Some(per_bidder) =
+                            params.get(BIDDER_PARAMS_KEY).and_then(Json::as_object)
+                        {
+                            explicit_bidders.extend(per_bidder.keys().cloned());
+                        }
                         bidder.extend(expand_trusted_server_bidders(&self.config.bidders, params));
                     } else if self.config.bidders.iter().any(|b| b == name) {
+                        explicit_bidders.insert(name.clone());
                         bidder.insert(name.clone(), params.clone());
                     } else if name != "aps" {
                         // `aps` is intentionally handled by its own provider. Any
@@ -1443,16 +1455,32 @@ impl PrebidAuctionProvider {
                     }
                 }
 
-                // When no inline PBS bidder params exist (e.g. creative-opportunity slots
-                // whose PBS params live in stored requests), tell PBS to resolve bidder
-                // config from the stored request keyed by this slot ID.
+                // Apply canonical and compatibility-derived rules in normalized
+                // order. An override rule can populate a bidder that arrived with
+                // empty params, promoting a fabricated empty into a valid bidder.
+                for (name, params) in &mut bidder {
+                    self.bid_param_override_engine
+                        .apply(BidParamOverrideFacts { bidder: name, zone }, params);
+                }
+
+                // Drop bidders that are still an empty object after overrides and
+                // were not explicitly supplied. Shipping `"bidder": {}` makes PBS
+                // reject the imp; an explicit empty object is preserved so genuine
+                // publisher misconfiguration stays visible.
+                bidder.retain(|name, params| {
+                    let is_empty_object = params.as_object().is_some_and(serde_json::Map::is_empty);
+                    !is_empty_object || explicit_bidders.contains(name)
+                });
+
+                // When no eligible PBS bidder params remain (e.g. creative-opportunity
+                // slots whose PBS params live in stored requests, or a slot whose
+                // configured bidders all resolved to fabricated empties), tell PBS to
+                // resolve bidder config from the stored request keyed by this slot ID.
                 //
-                // This cannot fire for the client /auction path: the JS adapter
-                // injects a `trustedServer` entry into every ad unit, so `bidder`
-                // is only empty for server-side creative-opportunity slots with
-                // no inline provider params (or when `config.bidders` is empty,
-                // where PBS previously received an empty bidder map and returned
-                // no bids — a stored-request miss is the same no-bid outcome).
+                // This cannot fire for a client /auction slot that carries real
+                // inline params: the JS adapter injects a `trustedServer` entry, and
+                // any bidder with params survives the drop above. It falls back only
+                // when nothing eligible remains — the same no-bid outcome as before.
                 let storedrequest = if bidder.is_empty() {
                     Some(ImpStoredRequest {
                         id: slot.id.clone(),
@@ -1460,12 +1488,6 @@ impl PrebidAuctionProvider {
                 } else {
                     None
                 };
-
-                // Apply canonical and compatibility-derived rules in normalized order.
-                for (name, params) in &mut bidder {
-                    self.bid_param_override_engine
-                        .apply(BidParamOverrideFacts { bidder: name, zone }, params);
-                }
 
                 Some(Imp {
                     id: Some(slot.id.clone()),
@@ -5947,6 +5969,131 @@ set = { placementId = "explicit_header" }
         assert_eq!(
             prebid["bidder"]["kargo"]["placementId"], "client_123",
             "should use inline bidder params from trustedServer expansion"
+        );
+    }
+
+    #[test]
+    fn to_openrtb_drops_fabricated_empty_bidder_params() {
+        // config.bidders lists three, but the slot supplies inline params only
+        // for kargo. Without a matching override, triplelift and criteo would
+        // expand to empty `{}` objects — invalid bidder entries PBS rejects.
+        // They must be dropped; the valid kargo bidder must still ship.
+        let config = parse_prebid_toml(
+            r#"
+[integrations.prebid]
+enabled = true
+server_url = "https://prebid.example"
+bidders = ["kargo", "triplelift", "criteo"]
+"#,
+        );
+
+        let slot = make_ts_slot(
+            "ad-header-0",
+            &json!({ "kargo": { "placementId": "kn1" } }),
+            None,
+        );
+        let request = make_auction_request(vec![slot]);
+
+        let ortb = call_to_openrtb(config, &request);
+        let params = bidder_params(&ortb);
+
+        assert_eq!(
+            params["kargo"]["placementId"], "kn1",
+            "should keep the valid inline bidder"
+        );
+        assert!(
+            !params.contains_key("triplelift"),
+            "should drop a fabricated empty bidder with no inline params or override"
+        );
+        assert!(
+            !params.contains_key("criteo"),
+            "should drop a fabricated empty bidder with no inline params or override"
+        );
+    }
+
+    #[test]
+    fn to_openrtb_preserves_an_explicitly_empty_bidder() {
+        // A publisher-supplied empty `{}` is a real (if misconfigured) signal and
+        // must survive so the misconfiguration stays visible — unlike a fabricated
+        // empty, which is dropped.
+        let config = parse_prebid_toml(
+            r#"
+[integrations.prebid]
+enabled = true
+server_url = "https://prebid.example"
+bidders = ["kargo"]
+"#,
+        );
+
+        let slot = make_ts_slot("ad-header-0", &json!({ "kargo": {} }), None);
+        let request = make_auction_request(vec![slot]);
+
+        let ortb = call_to_openrtb(config, &request);
+        let params = bidder_params(&ortb);
+
+        assert_eq!(
+            params["kargo"],
+            json!({}),
+            "should preserve an explicitly supplied empty bidder object"
+        );
+    }
+
+    #[test]
+    fn to_openrtb_keeps_a_fabricated_bidder_that_an_override_populates() {
+        // criteo has no inline params (fabricated empty), but an override rule
+        // fills it — so it is valid and must ship, not be dropped.
+        let config = parse_prebid_toml(
+            r#"
+[integrations.prebid]
+enabled = true
+server_url = "https://prebid.example"
+bidders = ["criteo"]
+
+[integrations.prebid.bid_param_overrides.criteo]
+networkId = 99999
+"#,
+        );
+
+        let slot = make_ts_slot("ad-header-0", &json!({}), None);
+        let request = make_auction_request(vec![slot]);
+
+        let ortb = call_to_openrtb(config, &request);
+        let params = bidder_params(&ortb);
+
+        assert_eq!(
+            params["criteo"]["networkId"], 99999,
+            "override should populate the fabricated empty bidder, keeping it"
+        );
+    }
+
+    #[test]
+    fn to_openrtb_falls_back_to_stored_request_when_all_bidders_are_fabricated_empty() {
+        // config.bidders present, but the slot supplies no inline params and no
+        // override matches — every configured bidder resolves to a fabricated
+        // empty and is dropped, leaving PBS to resolve via the stored request.
+        let config = parse_prebid_toml(
+            r#"
+[integrations.prebid]
+enabled = true
+server_url = "https://prebid.example"
+bidders = ["kargo", "triplelift"]
+"#,
+        );
+
+        let slot = make_ts_slot("ad-header-0", &json!({}), None);
+        let request = make_auction_request(vec![slot]);
+
+        let ortb = call_to_openrtb(config, &request);
+        let ext = ortb.imp[0].ext.as_ref().expect("should have imp ext");
+        let prebid = ext.get("prebid").expect("should have prebid in ext");
+
+        assert!(
+            prebid.get("bidder").is_none(),
+            "should drop all fabricated empty bidders"
+        );
+        assert_eq!(
+            prebid["storedrequest"]["id"], "ad-header-0",
+            "should fall back to stored request when no eligible bidders remain"
         );
     }
 
--- a/crates/trusted-server-core/src/integrations/prebid.rs
+++ b/crates/trusted-server-core/src/integrations/prebid.rs
@@ -1437,7 +1437,18 @@ impl PrebidAuctionProvider {
                         {
                             explicit_bidders.extend(per_bidder.keys().cloned());
                         }
-                        bidder.extend(expand_trusted_server_bidders(&self.config.bidders, params));
+                        // Fill in trusted-server-expanded bidders without
+                        // clobbering a direct bidder entry. `slot.bidders` is a
+                        // HashMap, so this branch can run after a direct entry;
+                        // `extend` would then overwrite valid direct params with a
+                        // fabricated empty `{}`, which `explicit_bidders` would
+                        // wrongly preserve past the drop below. `or_insert` makes
+                        // direct params win regardless of iteration order.
+                        for (bidder_name, bidder_params) in
+                            expand_trusted_server_bidders(&self.config.bidders, params)
+                        {
+                            bidder.entry(bidder_name).or_insert(bidder_params);
+                        }
                     } else if self.config.bidders.iter().any(|b| b == name) {
                         explicit_bidders.insert(name.clone());
                         bidder.insert(name.clone(), params.clone());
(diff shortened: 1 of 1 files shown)


crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -1424,10 +1424,22 @@ impl PrebidAuctionProvider {
...
+                // `"bidder": {}` (which PBS rejects).
+                let mut explicit_bidders: HashSet<String> = HashSet::new();
                 for (name, params) in &slot.bidders {
                     if name == TRUSTED_SERVER_BIDDER {
+                        if let Some(per_bidder) =
+                            params.get(BIDDER_PARAMS_KEY).and_then(Json::as_object)
+                        {
+                            explicit_bidders.extend(per_bidder.keys().cloned());
+                        }
                         bidder.extend(expand_trusted_server_bidders(&self.config.bidders, params));
```
ChristianPavilonis: 🔧 **Mixed bidder representations have nondeterministic precedence.** `slot.bidders` is a `HashMap`, so when a request has both a direct `kargo` entry and `trustedServer.bidderParams` that omits `kargo`, this expansion can run after the direct entry and overwrite valid params with fabricated `{}`. Because the direct entry marks `kargo` explicit, the subsequent retention step preserves the empty object and PBS can reject the impression.

Prefer direct bidder params deterministically, regardless of iteration order, and add a regression test containing both representations.

```suggestion
                        for (bidder_name, bidder_params) in
                            expand_trusted_server_bidders(&self.config.bidders, params)
                        {
                            bidder.entry(bidder_name).or_insert(bidder_params);
                        }
```

ChristianPavilonis (approved):
## Review Summary

Approved with one non-blocking medium-severity finding documented inline. The primary fabricated-empty suppression, override ordering, and stored-request fallback behavior are covered by focused tests.

Please address the inline mixed-representation precedence case in follow-up work before relying on this path for clients that can send both direct bidder params and `trustedServer.bidderParams`.

crates/trusted-server-core/src/integrations/prebid.rs
aram356: 🔧 **wrench** — Mixed bidder representations have nondeterministic precedence. `slot.bidders` is a `HashMap`, so when a request contains both a direct `kargo` entry and `trustedServer.bidderParams` that omits `kargo`, this expansion can run after the direct entry and overwrite its valid params with fabricated `{}`. Because the direct entry already marked `kargo` explicit, the retention step then preserves the empty object and PBS can reject the impression—the failure this PR aims to eliminate.

Make direct bidder params win deterministically regardless of iteration order, and add a regression test containing both representations. For example:

```rust
for (bidder_name, bidder_params) in
    expand_trusted_server_bidders(&self.config.bidders, params)
{
    bidder.entry(bidder_name).or_insert(bidder_params);
}
```

aram356 (changes requested):
## Summary

The fabricated-empty suppression, override ordering, stored-request fallback, and focused tests are otherwise sound, but mixed bidder representations can still recreate the invalid empty bidder object nondeterministically.

## Blocking

### 🔧 wrench

- **Mixed bidder representations have nondeterministic precedence**: direct bidder params can be overwritten by a fabricated empty object depending on `HashMap` iteration order (`crates/trusted-server-core/src/integrations/prebid.rs:1440`).

## CI Status

- GitHub integration checks: PASS (4/4)
- fmt: PASS locally
- focused native core tests: PASS (44/44)
- Fastly test runtime: BLOCKED locally by unavailable macOS certificate keychain after successful compilation
- clippy-fastly: BLOCKED locally by a build-script tooling error (`Unrecognized option: 'p'`)
- JS tests: not run; no JS/TS files changed

crates/trusted-server-core/src/integrations/prebid.rs
prk-Jr: Fixed in bfea103d9. Replaced the `extend` with the suggested `entry().or_insert()` loop, so trustedServer-expanded bidders only fill in absent entries — a direct bidder entry's valid params are never overwritten by a fabricated empty, and direct wins regardless of HashMap iteration order. Added `to_openrtb_prefers_direct_bidder_params_over_fabricated_empty`: a looped regression test with a slot carrying both a direct `kargo` entry and a `trustedServer.bidderParams` that omits it. It fails against the old overwrite path (kargo comes back as an empty object) and passes with the fix.

prk-Jr: Fixed in bfea103d9. Switched the `extend` to the `entry().or_insert()` loop you suggested, so a direct bidder entry's params win deterministically over a trustedServer-fabricated empty regardless of iteration order. Added `to_openrtb_prefers_direct_bidder_params_over_fabricated_empty` — a looped regression test over a slot with both a direct `kargo` entry and a `trustedServer.bidderParams` that omits kargo; it fails against the old overwrite behavior and passes with the fix.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -1442,16 +1465,32 @@ impl PrebidAuctionProvider {
...
+                        .apply(BidParamOverrideFacts { bidder: name, zone }, params);
+                }
+
+                // Drop bidders that are still an empty object after overrides and
+                // were not explicitly supplied. Shipping `"bidder": {}` makes PBS
+                // reject the imp; an explicit empty object is preserved so genuine
+                // publisher misconfiguration stays visible.
+                bidder.retain(|name, params| {
+                    let is_empty_object = params.as_object().is_some_and(serde_json::Map::is_empty);
+                    !is_empty_object || explicit_bidders.contains(name)
```
aram356: 🔧 **wrench** — Preserving an explicit `{}` still ships the payload PBS rejects.

PBS cannot see provenance: `"bidder": {"kargo": {}}` is byte-identical on the wire whether the empty object was fabricated by `expand_trusted_server_bidders` or supplied by the publisher. This retain rule drops the first and ships the second, but PBS rejects both — so the PR's stated goal ("never emit `\"bidder\": {}`") is only half-enforced. The test at `prebid.rs:6077` asserts this shape ships.

This is reachable from the **primary shipped path**, not a corner case. The Prebid adapter folds every server-side bid into `bidderParams[bid.bidder] = bid.params ?? {}` (`crates/trusted-server-js/lib/src/integrations/prebid/index.ts:584`), so any publisher ad unit with `bids: [{bidder: 'kargo', params: {}}]` marks `kargo` explicit and ships `{}`. The JS test at `index.test.ts:541` pins that shape as `{ appnexus: {} }`.

The blast radius is wider than one imp: `parse_response` (`prebid.rs:2157-2166`) collapses **any** non-2xx into `AuctionResponse::error`, zeroing bids for every slot in the auction. One publisher's empty params can take down otherwise-healthy slots.

The "misconfiguration stays visible" rationale doesn't hold in practice. When PBS rejects, the operator sees only `Prebid returned non-success status: 400` — the response body is discarded unless trace logging is enabled. That is precisely the opaque intermittent error this PR set out to eliminate. Dropping it and logging makes the misconfiguration visible where someone can actually act on it:

```rust
bidder.retain(|name, params| {
    if is_empty_object(params) {
        log::warn!(
            "prebid: dropping bidder '{}' on slot '{}' — empty params",
            name,
            slot.id
        );
        return false;
    }
    true
});
```

If preserving explicit empties is a deliberate contract with PBS that I'm missing, could you spell out what makes the explicit case safe on the wire when the fabricated one isn't?

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -1423,10 +1423,33 @@ impl PrebidAuctionProvider {
...
+                        // wrongly preserve past the drop below. `or_insert` makes
+                        // direct params win regardless of iteration order.
+                        for (bidder_name, bidder_params) in
+                            expand_trusted_server_bidders(&self.config.bidders, params)
+                        {
+                            bidder.entry(bidder_name).or_insert(bidder_params);
+                        }
                     } else if self.config.bidders.iter().any(|b| b == name) {
+                        explicit_bidders.insert(name.clone());
                         bidder.insert(name.clone(), params.clone());
```
aram356: 🔧 **wrench** — The merge is asymmetric: an empty direct entry still wins over real params.

`bfea103d9` fixed the `trustedServer` branch with `or_insert`, but this direct branch still uses an unconditional `insert`. A slot carrying a direct `kargo: {}` **and** a `trustedServer` entry whose `bidderParams` supplies real `kargo` params deterministically discards the real params and ships `"bidder": {"kargo": {}}` — the exact invalid payload this PR exists to remove, now on every run rather than intermittently. I verified it against this head:

```
PROBE direct-empty + ts-real => {"bidder":{"kargo":{}}}
PROBE LOST PARAMS on iter 0: {}
```

Note the comment at line 1444-1446 claims `or_insert` makes direct params win "regardless of iteration order". That's true but incomplete, and no reordering of the current single loop fixes it: if the direct branch runs first it inserts `{}` into a vacant entry, and `or_insert` then skips. The two sources have to be separated before merging:

```rust
// Merge: start from the expansion, then let direct entries win — but never
// let an empty `{}` clobber real params from the other source.
let mut bidder = expanded;
for (name, params) in direct {
    let keep_existing = is_empty_object(&params)
        && bidder.get(&name).is_some_and(|existing| !is_empty_object(existing));
    if !keep_existing {
        bidder.insert(name, params);
    }
}
```

In fairness on reachability: shipped JS never emits a direct entry and a `trustedServer` entry on the same slot — the adapter folds and filters at `index.ts:576-591` — so this needs a hand-crafted POST to the public `/auction` endpoint. That is the same reachability as the mirror-image case already fixed in this PR, which is why I'd rather see both closed than one.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -1442,16 +1465,32 @@ impl PrebidAuctionProvider {
...
+                // configured bidders all resolved to fabricated empties), tell PBS to
+                // resolve bidder config from the stored request keyed by this slot ID.
                 //
-                // This cannot fire for the client /auction path: the JS adapter
-                // injects a `trustedServer` entry into every ad unit, so `bidder`
-                // is only empty for server-side creative-opportunity slots with
-                // no inline provider params (or when `config.bidders` is empty,
-                // where PBS previously received an empty bidder map and returned
-                // no bids — a stored-request miss is the same no-bid outcome).
+                // This cannot fire for a client /auction slot that carries real
```
aram356: 🤔 **thinking** — This "cannot fire" claim isn't accurate.

A client `/auction` slot **can** carry real inline params and still hit the stored-request fallback: when the params name a bidder absent from `config.bidders`, `expand_trusted_server_bidders` fabricates `{}` for each configured bidder (`prebid.rs:945-954`), the retain drops them all, and `bidder` ends up empty. Verified against this head with `config.bidders = ["appnexus"]` and inline `{"kargo": {"placementId": "kn1"}}`:

```
PROBE real-inline-params-unconfigured-bidder => {"storedrequest":{"id":"ad-header-0"}}
```

The real params are silently discarded and the slot falls back to a stored request. The param loss itself is pre-existing (the `log::debug!` drop above), so this isn't a regression — but the comment is new and load-bearing, since it's the stated justification that the fallback can't over-broaden. Worth correcting and pinning with a test, as `to_openrtb_falls_back_to_stored_request_when_all_bidders_are_fabricated_empty` covers only the empty-`bidderParams` variant.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -1442,16 +1465,32 @@ impl PrebidAuctionProvider {
...
+                    self.bid_param_override_engine
+                        .apply(BidParamOverrideFacts { bidder: name, zone }, params);
+                }
+
+                // Drop bidders that are still an empty object after overrides and
+                // were not explicitly supplied. Shipping `"bidder": {}` makes PBS
+                // reject the imp; an explicit empty object is preserved so genuine
+                // publisher misconfiguration stays visible.
+                bidder.retain(|name, params| {
+                    let is_empty_object = params.as_object().is_some_and(serde_json::Map::is_empty);
```
aram356: 🤔 **thinking** — Non-object params bypass the drop entirely.

`as_object().is_some_and(is_empty)` only catches empty **objects**. A `Json::Null` params value is not an object, so it survives and ships as `"bidder": {"kargo": null}` — presumably just as invalid to PBS as `{}`. Verified against this head:

```
PROBE direct-null-params => {"bidder":{"kargo":null}}
```

This is reachable without a malformed request: `BidConfig.params` carries `#[serde(default)]` (`crates/trusted-server-core/src/auction/formats.rs:71-72`), so a POST that simply omits `params` deserializes to `Null` rather than `{}`. If the intent is "anything that isn't a usable params object shouldn't ship", the predicate probably wants to treat a non-object as ineligible too — though that's a slightly wider behavior change, so I'd rather flag it than prescribe it.

aram356 (changes requested):
## Summary

The `or_insert` fix in `bfea103d9` does resolve the case flagged in the previous round — I confirmed the new 64-iteration test genuinely fails against the old `extend` and passes now. The reordering (build → override → drop → decide stored-request) is sound, and tracing the history shows this PR is finishing a migration #680 already intended rather than inventing a new policy: `e06af4b0f` abandoned "fill `config.bidders` with empty `{}`" in favour of the stored-request path a year ago, and that intent only leaked because `expand_trusted_server_bidders` fabricates the same empties one layer up. That evidence retired my initial concern that broadening the fallback was risky.

What holds the PR up is that the fix is one-directional, and that the central design choice re-creates the payload the PR exists to eliminate. PBS cannot distinguish a fabricated `{}` from an explicitly supplied one — they are identical bytes on the wire — so preserving explicit empties keeps emitting `"bidder": {}` through the primary shipped JS path, and `parse_response` turns the resulting PBS rejection into an auction-wide bid wipeout rather than a visible misconfiguration signal.

## Blocking

### 🔧 wrench

- **Preserving an explicit `{}` still ships the payload PBS rejects**: provenance is invisible to PBS; reachable via the normal Prebid adapter path (`params: {}` → `bidderParams: {kargo: {}}`), and a PBS non-2xx zeroes every slot in the auction, not just the offending imp (`prebid.rs:1480-1483`, `prebid.rs:2157-2166`).
- **The merge is asymmetric — an empty direct entry still wins over real params**: the direct branch keeps an unconditional `insert`, so a direct `{}` deterministically discards `trustedServer`-supplied params and emits `"bidder": {"kargo": {}}` (`prebid.rs:1453`).

### ❓ question

- **The premise underpinning the whole design is unverified in-repo**: there is no test, fixture, or PBS mock anywhere backing either "PBS rejects `"bidder": {}`" or "an unknown stored-request ID yields a per-imp no-bid rather than a request-wide 400". Both exist only as prose in comments and commit messages. If a stored-request miss is in fact a 400, this trades one whole-auction failure mode for another instead of fixing it. Related: `auction/endpoints.rs:60-63` already requires a stored request to exist for every slot code that omits inline params, and this PR widens that set to include all-fabricated-empty slots — operators with `config.bidders` set but no stored requests provisioned move from one broken state to another. Has the empty-bidder rejection been confirmed against the deployed PBS?

## Non-blocking

### 🤔 thinking

- **The new "cannot fire" comment is inaccurate**: real inline params for a bidder outside `config.bidders` do fall back to the stored request, silently dropping those params (`prebid.rs:1490-1493`).
- **Non-object params bypass the drop**: `null` params ship as `"bidder": {"kargo": null}`; `#[serde(default)]` on `BidConfig.params` makes this reachable from a POST that omits `params` (`prebid.rs:1481`).

### ♻️ refactor

- **Extract an `is_empty_object` helper**: the emptiness predicate is inlined in the retain at `prebid.rs:1481` and is needed again by both suggested fixes above. A shared `fn is_empty_object(params: &Json) -> bool { params.as_object().is_some_and(serde_json::Map::is_empty) }` keeps the rule in one place.

### 📌 out of scope

- **`zone` leaks into every bidder's params**: when a `trustedServer` entry has no `bidderParams`, `expand_trusted_server_bidders` (`prebid.rs:942-957`) hands each configured bidder the entire `trustedServer` object as its params, so a zone-only slot ships `{"kargo": {"zone": "header"}, "triplelift": {"zone": "header"}}` — a TS-internal orchestration key sent to PBS as a bidder param. These are non-empty, so the new drop never catches them; it's the third route by which fabricated params reach PBS. Pre-existing and not this PR's job, but it limits the "fabricated params never ship" contract and is worth a follow-up issue.

## CI Status

- GitHub checks: PASS (4/4)
- fmt: PASS (verified locally)
- clippy (`trusted-server-core`, native host, `-D warnings`): PASS (verified locally)
- rust tests (`trusted-server-core`, native host): PASS (1631/1631, verified locally)
- js tests: not run — no JS/TS files changed

Note: #899 is still open, so this stays stacked and cannot merge first.

crates/trusted-server-core/src/integrations/prebid.rs
prk-Jr: Resolved in 6a75cc499. Explicit empties are no longer preserved — the drop is now unconditional. `is_unusable_bidder_params` (empty object **or** non-object) runs after overrides and drops every remaining unusable value regardless of provenance, `log::warn`-ing each drop with slot + bidder so the misconfiguration is visible where an operator can act; the slot then falls back to its stored request. `to_openrtb_preserves_an_explicitly_empty_bidder` is now `to_openrtb_drops_an_explicitly_empty_bidder` (asserts drop + stored-request fallback).

On the premise question in the summary — whether a stored-request miss is a per-imp no-bid or a request-wide 400: I can't verify that from the repo; it is deployed-PBS behavior. What this change verifiably fixes is removing the definitely-rejected `"bidder": {}` / `null` from the wire, routed through the same stored-request fallback #680 already established for slots omitting inline params (`endpoints.rs` documents a stored request must exist for any such slot code). If a stored-request miss is itself a 400 on deployed PBS, that is a separate pre-existing exposure this does not widen for correctly-provisioned operators — but confirming it against deployed PBS would settle it. Do you have that confirmation, or should it be tracked?

Separately, the `zone`-leak out-of-scope item from the summary is filed as #927.

crates/trusted-server-core/src/integrations/prebid.rs
prk-Jr: Resolved in 6a75cc499. The two sources are now collected separately — `expanded` (the trustedServer expansion) and `direct` (config-bidder entries) — then merged: a direct entry wins, except when an unusable direct value (`{}` or a non-object) would overwrite real params the expansion supplied. That closes the direct-`{}`-clobbers-real case (and, via the same predicate, direct-`null`-clobbers-real). Added `to_openrtb_direct_empty_does_not_clobber_trusted_server_params` — a 64-iteration loop mirroring the existing prefers-direct test.

crates/trusted-server-core/src/integrations/prebid.rs
prk-Jr: Corrected in 6a75cc499. The comment no longer claims the fallback cannot fire for slots with real inline params; it now spells out that a slot whose only inline params name a bidder absent from `config.bidders` also lands in the stored-request fallback — that bidder is never expanded, the configured bidders fabricate empties, the drop clears them, and the real params are lost. Pinned by `to_openrtb_falls_back_to_stored_request_for_real_params_of_unconfigured_bidder`.

crates/trusted-server-core/src/integrations/prebid.rs
prk-Jr: Resolved in 6a75cc499. The predicate now treats a non-object as unusable too — `params.as_object().is_none_or(serde_json::Map::is_empty)` — so `null` (reachable via a POST that omits `params`, since `BidConfig.params` is `#[serde(default)]`) is dropped rather than shipped as `"bidder": {"kargo": null}`. Added `to_openrtb_drops_non_object_bidder_params`. This also folds the `is_empty_object` helper the summary asked for into a single shared `is_unusable_bidder_params` used by both the merge and the drop.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -1442,16 +1462,65 @@ impl PrebidAuctionProvider {
...
+                // explicitly, and shipping either gets the imp rejected; a single
+                // non-2xx then collapses into an auction-wide bid wipeout (see
+                // `parse_response`), so one misconfigured slot must never reach the
+                // wire. Log each drop at warn so the misconfiguration is visible
+                // where an operator can act on it — the PBS 400 body is otherwise
+                // discarded unless trace logging is enabled.
+                bidder.retain(|name, params| {
+                    if is_unusable_bidder_params(params) {
+                        log::warn!(
+                            "prebid: dropping bidder '{}' on slot '{}' — empty or non-object params; slot falls back to its stored request",
```
aram356: 🤔 **thinking** — This now warns on a designed path, not just a misconfigured one.

Making the drop unconditional was the right call, but it widened what reaches this `warn`. `creative_opportunities.rs:294-305` *deliberately* emits `bidderParams: {}` — its own comment says it gives "each bidder an empty params object that the override engine then fills with zone-aware rules". When no override matches that bidder/zone, every configured bidder is dropped here and each one logs a warning. I confirmed the shape lands in the fallback:

```
N2 creative-opp-empty-bidderParams => {"storedrequest":{"id":"s"}}
```

So a routine creative-opportunity slot emits N warnings (one per configured bidder) describing its *intended* behavior, and the message tells the operator it's a misconfiguration falling back to a stored request. Warn-level noise on an expected path erodes the signal this log exists to provide — the genuine publisher misconfiguration you added it for gets buried under the designed case.

Since the two are distinguishable at this point, they could log differently — e.g. keep `warn` when the slot supplied *some* inline params (a real publisher mistake), and drop to `debug` when the expansion fabricated everything (creative-opportunity / stored-request path). Even just collapsing to one line per slot rather than one per bidder would help.

Non-blocking — behavior is correct either way, this is purely about operability.

aram356 (approved):
## Summary

All four findings from the previous round are resolved, and I verified each against `6a75cc499` rather than taking the fix on description:

| Finding | Before | After |
| --- | --- | --- |
| 🔧 Explicit `{}` ships the rejected payload | `{"bidder":{"kargo":{}}}` | `{"storedrequest":{"id":"s"}}` |
| 🔧 Asymmetric merge drops real params | `LOST PARAMS on iter 0` | **0/64 bad** |
| 🤔 `null` params bypass the drop | `{"bidder":{"kargo":null}}` | `{"storedrequest":{"id":"s"}}` |
| 🤔 Inaccurate "cannot fire" comment | inaccurate | corrected, pinned by test |

The merge rewrite is the right shape. Collecting `expanded` and `direct` separately and merging with an explicit `clobbers_real` guard is what makes this order-independent — no reordering of the previous single loop could have achieved it, since whichever branch ran first would win. I also checked the null-clobbers-real variant I hadn't explicitly raised (0/64 bad), and confirmed valid params still ship unchanged. `is_unusable_bidder_params` folds in the shared predicate, and #927 tracks the `zone` leak.

Verification on this head: `cargo fmt --all -- --check` PASS; clippy (`trusted-server-core`, `--all-targets`, `-D warnings`) clean; **1634/1634** core tests pass (+3 new); GitHub checks green 4/4.

## Non-blocking

### 🤔 thinking

- **The unconditional drop now warns on a designed path**: creative-opportunity slots intentionally emit `bidderParams: {}`, so each one logs a `warn` per configured bidder calling its intended behavior a misconfiguration (`prebid.rs:1498-1503`, `creative_opportunities.rs:294-305`). Detail inline.

### 📌 out of scope

- **The PBS premise remains unverified**: your answer is fair — whether a stored-request miss is a per-imp no-bid or a request-wide 400 is deployed-PBS behavior and genuinely not knowable from this repo, and this PR does verifiably remove definitely-rejected payloads via the fallback #680 already established. It shouldn't hold the PR. But it stays an untested assumption underpinning the design, and there's still no PBS mock or fixture anywhere in the repo to pin it. Worth an issue alongside #927 so it's tracked rather than dropped — happy to file it if you'd prefer.

Approving. Note this stays stacked on #899, which is still open.

## CI Status

- GitHub checks: PASS (4/4)
- fmt: PASS (verified locally)
- clippy (`trusted-server-core`, native host, `-D warnings`): PASS (verified locally)
- rust tests (`trusted-server-core`, native host): PASS (1634/1634, verified locally)
- js tests: not run — no JS/TS files changed

PR #916, opened 2026-07-15
Make auction creative rewriting optional

### LINKED ISSUE #914: Make creative rewriting optional
## Problem
Creative HTML is always sanitized and rewritten before it is returned from the auction path. Operators cannot opt out when a deployment needs the original creative markup.

## Proposal
Add an operator-facing configuration option that controls creative rewriting. It should default to the current enabled behavior for backwards compatibility.

When disabled, winning creative HTML should be delivered without the creative URL-rewriting pass. Define and document whether sanitization remains mandatory or is controlled separately, with security implications made explicit.

## Acceptance criteria
- A documented config option controls creative rewriting.
- The default preserves current behavior.
- The auction creative delivery path honors the option.
- Tests cover enabled (current behavior) and disabled behavior.
- Configuration and creative-processing documentation are updated.

ChristianPavilonis:
## Summary

- Make winning-bid creative rewriting configurable while keeping server-side sanitization mandatory.
- Preserve rollback compatibility and make the operator migration, privacy, and rendering consequences explicit.
- Fix protocol-relative rewrite-exclusion matching and record the associated first-party signing validation behavior.

## Changes

| File | Change |
| ---- | ------ |
| `CHANGELOG.md` | Document optional rewriting and protocol-relative exclusion behavior. |
| `crates/trusted-server-cli/tests/config_env_overlay.rs` | Cover migrated overlays, default blob omission, and clean local config diffs. |
| `crates/trusted-server-core/src/auction/README.md` | Distinguish mandatory sanitization from configurable rewriting. |
| `crates/trusted-server-core/src/auction/endpoints.rs` | Document `/auction` sanitization and rewrite behavior. |
| `crates/trusted-server-core/src/auction/formats.rs` | Use the centralized auction creative processing policy and cover both modes. |
| `crates/trusted-server-core/src/auction/orchestrator.rs` | Update the exhaustive configuration test literal. |
| `crates/trusted-server-core/src/auction_config_types.rs` | Add the default-true setting and document rollback-safe serialization. |
| `crates/trusted-server-core/src/config_payload.rs` | Test and label the legacy blob schema snapshot. |
| `crates/trusted-server-core/src/creative.rs` | Normalize protocol-relative exclusions and centralize sanitize-then-rewrite processing. |
| `crates/trusted-server-core/src/proxy.rs` | Prove proxy response rewriting remains independent and reject excluded signing targets. |
| `crates/trusted-server-core/src/publisher.rs` | Apply the setting to production SSAT inline creatives and cover sanitize-only delivery. |
| `crates/trusted-server-core/src/settings.rs` | Parse the setting and warn when enabled auctions disable rewriting. |
| `docs/guide/auction-orchestration.md` | Document sanitize-only and sanitize-and-rewrite modes. |
| `docs/guide/cli.md` | Document typed config validation, diff, and push flow. |
| `docs/guide/configuration.md` | Document setting scope, migration, rollback, and environment overrides. |
| `docs/guide/creative-processing.md` | Explain privacy, direct-resource, runtime, and rendering effects. |
| `trusted-server.example.toml` | Add and explain the operator setting. |

## Closes

Closes #914

## Test plan

- [x] Other: `cargo test-cloudflare`, `cargo test-spin`, all target-matched clippy aliases, and `./scripts/test-cli.sh`

## Checklist

- [ ] Uses `tracing` macros (not `println!`) — N/A; this project requires `log` macros.
- [x] New code has tests


crates/trusted-server-core/src/auction_config_types.rs
```diff
@@ -11,6 +11,10 @@ pub struct AuctionConfig {
     #[serde(default)]
     pub enabled: bool,
 
+    /// Rewrite sanitized winning-bid creative HTML to first-party endpoints.
+    #[serde(default = "default_rewrite_creatives")]
```
aram356: 🔧 **wrench — Preserve rollback compatibility for default-valued blobs**

This field is always serialized, including when it has the default value `true`. The base `AuctionConfig` has `#[serde(deny_unknown_fields)]` and no `rewrite_creatives` field, so after `ts config push` a deployment rollback to the base binary cannot deserialize the current blob and can fail during configuration loading. The new legacy-blob test covers old → new compatibility, but not new → old rollback compatibility.

Please omit the field when it is `true` (for example, with a `skip_serializing_if` predicate) and add a compatibility test proving that a default head payload is accepted by the legacy schema. An explicitly configured `false` necessarily needs an operator rollback step that restores the previous blob, which should also be documented.

docs/guide/configuration.md
```diff
@@ -1191,19 +1191,30 @@ Settings for the auction orchestrator that coordinates multiple bid providers.
...
+| ------------------- | ------------- | ------------------ | ----------------------------------------------------------------- |
+| `enabled`           | Boolean       | `false`            | Enable the auction orchestrator                                   |
+| `rewrite_creatives` | Boolean       | `true`             | Rewrite sanitized winning-bid `adm` through first-party endpoints |
+| `providers`         | Array[String] | `[]`               | Provider names that participate (e.g., `["prebid", "aps"]`)       |
+| `mediator`          | String        | Optional           | Mediator provider name (runs parallel mediation when set)         |
+| `timeout_ms`        | Integer       | `2000`             | Auction timeout in milliseconds                                   |
+| `creative_store`    | String        | `"creative_store"` | Deprecated; creatives are now delivered inline                    |
+
+Creative markup returned by `POST /auction` is always server-sanitized. With
+`rewrite_creatives = true` (the default), eligible absolute or protocol-relative
```
aram356: 🔧 **wrench — Protocol-relative URLs do not currently honor rewrite exclusions**

This new sentence groups absolute and protocol-relative URLs under the `rewrite.exclude_domains` contract, but `creative::to_abs()` calls `settings.rewrite.is_excluded(t)` before converting `//host/path` to `https://host/path`. `Rewrite::is_excluded()` uses `Url::parse()`, which rejects the scheme-relative input, so an excluded `//cdn.example.com/ad.png` is still rewritten through the first-party proxy.

Please normalize protocol-relative candidates before exclusion matching and add exact/wildcard regression cases. Alternatively, narrow this documentation so it does not publish a privacy guarantee the implementation does not provide.

docs/guide/configuration.md
```diff
@@ -1221,6 +1232,7 @@ server_url = "https://prebid-server.example.com/openrtb2/auction"
 
 ```bash
 TRUSTED_SERVER__AUCTION__ENABLED=true
+TRUSTED_SERVER__AUCTION__REWRITE_CREATIVES=true
```
aram356, line 1298: 🔧 **wrench — Make this environment override actionable for existing configs**

This override currently has two silent failure modes. First, current production configuration is loaded from the config-store blob; the core `build.rs` is empty apart from `rerun-if-changed`, so the guide's build-time/baked-in environment instructions are obsolete. The supported overlay happens during `ts config validate/push`. Second, EdgeZero v0.0.4 only overlays scalar leaves already present in the parsed TOML. A valid pre-PR config omits `rewrite_creatives`, so setting this variable to `false` is ignored and serde defaults rewriting back to `true`. I reproduced this with the legacy integration fixture: even `not-a-boolean` was ignored and validation returned OK, while the same value correctly failed against the updated example where the leaf exists.

Please update the environment sections to describe the typed CLI push flow and explicitly migrate existing `[auction]` tables by adding this leaf before relying on the override (or change the loader to materialize defaults before overlays). Add a CLI-path regression test using a legacy TOML file so this cannot silently regress.

docs/guide/creative-processing.md
```diff
@@ -1,10 +1,10 @@
 # Creative Processing
 
-Learn how Trusted Server automatically rewrites ad creative HTML and CSS to route all resources through first-party domains.
+Learn how Trusted Server rewrites ad creative HTML and CSS to route resources through first-party domains.
 
 ## Overview
 
-Creative processing transforms third-party ad creatives by rewriting URLs to go through your first-party domain. This provides:
+Creative processing has separate auction-response and proxied-response paths. When rewriting is enabled for a path, it transforms third-party ad creatives by rewriting URLs to go through your first-party domain. This provides:
```
aram356, line 7: 🤔 **thinking — Qualify the all-resources privacy claim below**

The following bullet says all resources load through the publisher domain, but enabled rewriting intentionally leaves URLs matching `rewrite.exclude_domains` direct. Consider changing it to: “Eligible, non-excluded external resources load through your domain.”

aram356 (changes requested):
## Summary

The runtime branch itself is well scoped: sanitization remains mandatory, rewriting is gated after sanitization, both modes have substantive tests, and `/first-party/proxy` remains independent. I am requesting changes for rollback compatibility and two operator-facing configuration/privacy contracts described in the inline comments.

## Non-blocking

### 🏕 camp site

- **Update the internal auction README**: `crates/trusted-server-core/src/auction/README.md:139`, `:251`, and `:382` still describe creative rewriting as unconditional. Please consider updating those passages to distinguish mandatory sanitization from default-on, configurable rewriting.

### 📌 out of scope

- **Track the pre-existing `iframe[srcdoc]` sanitizer gap**: the generic handler at `crates/trusted-server-core/src/creative.rs:385` neither removes nor recursively sanitizes `srcdoc`, while the normal renderer grants `allow-scripts` and `allow-same-origin` at `crates/trusted-server-js/lib/src/core/render.ts:14`. This exists at the base SHA and should not block this PR, but it deserves a security follow-up that removes `srcdoc` and adds regression coverage through both rewrite modes.

### 👍 praise

- The implementation keeps sanitization strictly before the configuration branch, avoids logging creative contents, covers default/disabled behavior and legacy blob loading, and verifies that proxy HTML/CSS rewriting is independent. No new dependency, OS API, or WASM-incompatible construct is introduced.

## CI Status

- fmt and all adapter/target clippy checks: PASS
- Rust tests and builds (Fastly, Axum, Cloudflare, Spin, parity, CLI): PASS
- JS formatting and Vitest: PASS
- integration, browser, Fastly EC lifecycle, and CodeQL checks: PASS

crates/trusted-server-core/src/auction_config_types.rs
ChristianPavilonis: Addressed. `rewrite_creatives = true` now uses `skip_serializing_if`, while explicit `false` remains serialized. Added tests for omission/defaulting, explicit-false round trips, and deserializing the current default auction payload with a test-only legacy `deny_unknown_fields` schema. Focused tests and the full adapter test suite pass.

docs/guide/configuration.md
ChristianPavilonis: Addressed by normalizing protocol-relative candidates before exclusion matching and reusing that absolute value for rewriting. Added exact-domain, wildcard base/subdomain, non-excluded, and HTML-level regression coverage. `cargo test-fastly` passes.

docs/guide/configuration.md
ChristianPavilonis: Addressed. The configuration and CLI guides now describe the typed validate/diff/push flow, EdgeZero v0.0.4 existing-leaf constraint, migration steps, and rollback sequence. Added a host CLI regression based on the legacy integration fixture: it migrates the leaf, proves `false` appears in the locally pushed blob, and proves an invalid boolean fails coercion. CLI tests pass.

docs/guide/creative-processing.md
ChristianPavilonis: Addressed. The privacy bullet now says: “Eligible, non-excluded external resources load through your domain.”

ChristianPavilonis:
Implemented and pushed the requested review fixes in `6fcbee0d`.

Also addressed the non-blocking internal auction README cleanup. The pre-existing nested `iframe[srcdoc]` sanitizer gap remains out of scope for this PR and is now tracked in #929.

Validation completed locally:
- all Fastly/Axum/Cloudflare/Spin test aliases
- native CLI tests and cross-adapter parity
- all target-matched clippy aliases plus CLI clippy
- Rust/JS/docs formatting and Vitest

crates/trusted-server-core/src/creative.rs
```diff
@@ -56,30 +56,21 @@ pub(super) fn to_abs(settings: &Settings, u: &str) -> Option<String> {
...
-    let lower = t.to_ascii_lowercase();
-    if lower.starts_with("data:")
-        || lower.starts_with("javascript:")
-        || lower.starts_with("mailto:")
-        || lower.starts_with("tel:")
-        || lower.starts_with("blob:")
-        || lower.starts_with("about:")
-    {
+    // Match exclusions against the same absolute URL used for rewriting.
+    if settings.rewrite.is_excluded(&absolute) {
```
prk-Jr, line 69: 📌 **out of scope** — This changes `to_abs()` to build the absolute URL *before* calling `settings.rewrite.is_excluded()`, instead of checking exclusion on the raw token. That's a real behavior fix: `is_excluded()` calls `url::Url::parse()`, which fails on a schemeless string like `//trusted-cdn.example.com/x.js`, so previously an `exclude_domains` entry never actually excluded a creative URL that used the protocol-relative form — it got rewritten anyway despite being configured as excluded. It's well tested (new cases around line 1317+), but it's an independent bug fix bundled into the "make rewriting optional" PR and isn't mentioned in the Change table. Worth calling out explicitly in the PR description since it changes existing exclusion behavior on `main` for protocol-relative URLs.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -250,21 +251,34 @@ pub fn convert_to_openrtb_response(
         let width = to_openrtb_i32(bid.width, "width", &bid_context);
         let height = to_openrtb_i32(bid.height, "height", &bid_context);
 
-        // Process creative HTML if present - — sanitize dangerous markup first, then rewrite URLs.
+        // Process creative HTML if present — always sanitize dangerous markup first.
         let creative_html = if let Some(ref raw_creative) = bid.creative {
             let sanitized = creative::sanitize_creative_html(raw_creative);
-            let rewritten = creative::rewrite_creative_html(settings, &sanitized);
+            let sanitized_len = sanitized.len();
+            let rewrite_creatives = settings.auction.rewrite_creatives;
```
prk-Jr: 🤔 **thinking** — `settings.auction.rewrite_creatives` is read fresh on every iteration of the per-bid loop; it's loop-invariant for the whole response. Trivial cost, not blocking — just flagging that hoisting it above the `for` loop would be marginally cleaner.

prk-Jr (commented):
## Summary

Adds a default-true `[auction].rewrite_creatives` setting: creative sanitization for `POST /auction` stays mandatory, while first-party resource/click rewriting and TSJS injection become optional. Backward-compatibility handling (default omitted from serialized/legacy blobs, explicit `false` preserved, legacy schema round-trip, EdgeZero env-overlay pre-existing-leaf quirk) is thorough and well tested. No blocking issues found.

## Non-blocking

### ⛏ nitpick

- **"unre-written" wording**: `docs/guide/configuration.md` and `docs/guide/creative-processing.md` repeat the hyphenation "unre-written" (e.g. "sanitized but unre-written HTML") several times. Suggest "not rewritten" or "unrewritten" for readability — no behavior impact.

### 👍 praise

- **Backward-compatibility test coverage**: the `skip_serializing_if` default-omission design plus tests across `settings.rs` (TOML omitted/explicit-false), `config_payload.rs` (legacy JSON blob round-trip and legacy-schema deserialization), and the new CLI integration test (`config_env_overlay.rs`, proving the EdgeZero "env overlay only overrides pre-existing leaves" quirk is handled) directly cover the real rollback/migration edge cases operators will hit. The `proxy.rs` test proving `/first-party/proxy` rewriting stays independent of the new flag is a good isolation check too.

### 📝 note

- **Change table is stale**: the PR description's file table lists 12 files; the actual diff touches 16, including `creative.rs` (see inline comment — an independent exclusion-matching bug fix bundled here), `crates/trusted-server-cli/tests/config_env_overlay.rs`, `crates/trusted-server-core/src/auction/README.md`, and `docs/guide/cli.md`. Worth syncing the table with the real diff before merge.

## CI Status

- fmt: PASS
- clippy (fastly/axum/cloudflare native+wasm/spin native+wasm): PASS
- rust tests (fastly/axum/cloudflare/spin/cross-adapter parity/CLI): PASS
- js tests (vitest): PASS
- js/docs format: PASS

crates/trusted-server-core/src/creative.rs
```diff
@@ -56,30 +56,21 @@ pub(super) fn to_abs(settings: &Settings, u: &str) -> Option<String> {
...
     }
 
-    if t.starts_with("//") {
-        Some(format!("https:{t}"))
-    } else if lower.starts_with("http://") || lower.starts_with("https://") {
-        Some(t.to_owned())
-    } else {
-        None
-    }
+    Some(absolute)
```
aram356, line 73: 🔧 **wrench** — Undeclared behavior change: this reorder is a real fix, but it is unrelated to `rewrite_creatives` and is not listed in the PR description or the CHANGELOG.

`RewriteConfig::is_excluded` (`settings.rs:553`) parses with `url::Url::parse`, which fails on `//host/path` and returns `false` — so protocol-relative URLs were **never** excluded before. Matching against the normalized absolute URL corrects that, and changes two user-visible behaviors:

1. **Creatives** — `//excluded-cdn.example/x.png` is now left direct instead of proxied.
2. **`/first-party/proxy`** — `to_abs` is also the `tsurl` validator at `proxy.rs:1633`, so URLs on excluded domains that previously signed successfully now fail with `unsupported url`.

**Fix**: add a CHANGELOG `### Fixed` entry and a PR-description line covering both effects, or split this into its own PR. The new tests at `creative.rs:1317+` cover the creative path; nothing covers the proxy-validator side effect.

crates/trusted-server-core/src/auction_config_types.rs
```diff
@@ -11,6 +11,13 @@ pub struct AuctionConfig {
     #[serde(default)]
     pub enabled: bool,
 
+    /// Rewrite sanitized winning-bid creative HTML to first-party endpoints.
+    #[serde(
+        default = "default_rewrite_creatives",
+        skip_serializing_if = "is_default_rewrite_creatives"
+    )]
+    pub rewrite_creatives: bool,
```
aram356, line 24: 🔧 **wrench** — `skip_serializing_if` here is load-bearing for binary rollback, but nothing in the code says so.

`AuctionConfig` is `#[serde(deny_unknown_fields)]` (line 8), so a pushed blob carrying `rewrite_creatives` hard-fails config load on a binary that predates the field. `skip_serializing_if` is the only thing preventing that — and the rationale currently lives solely in a docs page and a TOML comment. The next person who "simplifies" this attribute silently breaks rollback for every deployment.

**Fix**:

```rust
/// Rewrite sanitized winning-bid creative HTML to first-party endpoints.
///
/// `skip_serializing_if` is deliberate, not cosmetic: [`AuctionConfig`] is
/// `deny_unknown_fields`, so a pushed blob carrying this key fails config load
/// on a binary predating it. Omitting the default keeps blobs
/// rollback-compatible; an explicit `false` must serialize and therefore pins
/// the deployment to a binary that knows the field.
#[serde(
    default = "default_rewrite_creatives",
    skip_serializing_if = "is_default_rewrite_creatives"
)]
pub rewrite_creatives: bool,
```

Worth repeating a one-line pointer on `is_default_rewrite_creatives` (line 69) too, since that is where the deletion would start.

trusted-server.example.toml
```diff
@@ -112,6 +112,11 @@ rewrite_script = true
 
 [auction]
 enabled = false
+# Defaults to true. Keep this leaf present when using the EdgeZero v0.0.4
+# environment override. Set false to return sanitized but unre-written winning-bid
+# adm, skipping proxy/click URL conversion and creative TSJS injection.
+# Sanitization is always applied. Restore and push true before an older-binary rollback.
+rewrite_creatives = true
```
aram356, line 119: ❓ **question** — Does `ts config diff` stay clean once this leaf is present?

The example now ships `rewrite_creatives = true` explicitly (needed for the EdgeZero v0.0.4 overlay, which does not create missing leaves), but `config push` serializes it away via `skip_serializing_if`. If `config diff` compares the typed loader's output on both sides this is a no-op; if it compares the TOML document against the stored blob, every operator sees a phantom diff on every run.

Could you confirm which it is? If it is the latter, an assertion in `crates/trusted-server-cli/tests/config_env_overlay.rs` covering `config diff` would pin it.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -250,21 +251,34 @@ pub fn convert_to_openrtb_response(
...
+            let processed = if rewrite_creatives {
+                creative::rewrite_creative_html(settings, &sanitized)
+            } else {
+                sanitized
```
aram356: ♻️ **refactor** — "Always sanitize, conditionally rewrite" is a policy, but it lives as an inline `if` inside one converter. A second auction-creative emitter would silently miss it, and the invariant cannot be unit-tested where the functions actually live.

**Fix**: move it into `creative` so it has a single home:

```rust
// creative.rs
/// Sanitize auction creative markup, then rewrite it to first-party endpoints
/// unless `[auction].rewrite_creatives` is disabled. Sanitization is mandatory
/// in both modes.
pub fn process_auction_creative(settings: &Settings, raw: &str) -> String {
    let sanitized = sanitize_creative_html(raw);
    if settings.auction.rewrite_creatives {
        rewrite_creative_html(settings, &sanitized)
    } else {
        sanitized
    }
}
```

That also collapses the `sanitized_len` and `rewrite_mode` temporaries at the call site.

crates/trusted-server-core/src/auction/formats.rs
aram356: ♻️ **refactor** — Disabling this flag turns off the product's core first-party mediation promise, but the only signal is a `debug!` line per bid — invisible at production log levels.

**Fix**: emit a one-shot `log::warn!` at config load when `auction.enabled && !auction.rewrite_creatives`, e.g. `"Auction creative rewriting disabled; creative assets and clicks will contact third-party hosts directly"`. That makes an accidental `false` visible without trawling debug logs.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -250,21 +251,34 @@ pub fn convert_to_openrtb_response(
...
+            };
+            let rewrite_mode = if rewrite_creatives {
+                "enabled"
+            } else {
+                "disabled"
+            };
```
aram356: ⛏ **nitpick** — `rewrite_mode` as `"enabled"`/`"disabled"` can just be the bool: `rewrite={}` with `rewrite_creatives` reads the same and drops the four-line `let`.

crates/trusted-server-core/src/config_payload.rs
```diff
@@ -46,6 +46,24 @@ mod tests {
...
+        _providers: Vec<String>,
+        #[serde(rename = "mediator")]
+        _mediator: Option<String>,
+        #[serde(rename = "timeout_ms")]
+        _timeout_ms: u32,
+        #[serde(rename = "creative_store")]
+        _creative_store: String,
+        #[serde(rename = "allowed_context_keys")]
+        _allowed_context_keys: std::collections::HashSet<String>,
+    }
```
aram356, line 69: 🤔 **thinking** — This hand-mirrored copy of the pre-`rewrite_creatives` schema is the right idea, but nothing keeps it in sync. Every future `AuctionConfig` field needs a matching field added here, or the test quietly starts passing for the wrong reason (it would still accept the payload — it just would no longer be asserting what its name claims).

A comment naming the release this snapshot represents (e.g. `// Mirrors AuctionConfig as of v0.x — do not add fields introduced after that.`) would at least make the drift obvious in review.

crates/trusted-server-cli/tests/config_env_overlay.rs
```diff
@@ -0,0 +1,122 @@
...
+
+fn validate_with_overlay(project: &MigratedProject, raw_value: &str) -> Output {
+    Command::new(env!("CARGO_BIN_EXE_ts"))
+        .args(["config", "validate", "--manifest"])
+        .arg(&project.manifest_path)
+        .arg("--app-config")
+        .arg(&project.config_path)
+        .env(REWRITE_ENV, raw_value)
+        .output()
+        .expect("should run ts config validate")
```
aram356, line 64: 🤔 **thinking** — Neither this helper nor the `config push` invocation pins a working directory, so `Command` inherits the crate's cwd while the assertion reads `.edgezero/local-config-trusted_server_config.json` from the tempdir. It passes on CI, so the store evidently resolves relative to `--manifest` — but that is an edgezero-cli implementation detail, and if it ever changes the test either fails confusingly or writes `.edgezero/` into the repo.

**Fix**: add `.current_dir(project.directory.path())` to both commands.

aram356 (changes requested):
## Summary

Narrowly-scoped, backward-compatible feature with strong test coverage (default, explicit `false`, TOML parse, blob round-trip, legacy-schema acceptance, CLI env overlay, and a regression test proving `/first-party/proxy` rewriting is unaffected) and careful documentation. Two items need attention before merge: an undeclared drive-by behavior change in `to_abs`, and the fact that the rollback-critical `skip_serializing_if` carries no in-code rationale. Both are inline.

## Blocking

### 🔧 wrench

- **Undeclared behavior change in `to_abs`**: moving the exclusion check onto the normalized URL is a real fix — `is_excluded` never matched protocol-relative URLs because `url::Url::parse` fails on `//host/path` — but it is unrelated to `rewrite_creatives` and appears in neither the PR description nor the CHANGELOG. It also silently tightens the `/first-party/proxy` `tsurl` validator (`proxy.rs:1633`), which is untested. (`crates/trusted-server-core/src/creative.rs:53-73`)
- **`skip_serializing_if` is load-bearing but unexplained**: `AuctionConfig` is `deny_unknown_fields`, so omitting the default is what keeps pushed blobs readable by an older binary during rollback. That rationale exists only in docs and a TOML comment, so a future cleanup of the attribute silently breaks rollback. (`crates/trusted-server-core/src/auction_config_types.rs:15-19`)

### ❓ question

- **Client-side behavior with `rewrite_creatives = false`**: disabling also drops `data-tsclick` and the `tsjs-unified.min.js` runtime injection. The docs frame this purely as a privacy trade-off. Is the creative render bridge tolerant of a missing runtime, or does the creative fail to size/render? If rendering depends on it, that is a functional consequence and belongs in the warning block in `docs/guide/creative-processing.md`.
- **`ts config diff` with the leaf present**: raised inline on `trusted-server.example.toml:119`.

## Non-blocking

### ♻️ refactor

- **Gate the flag inside `creative`**: `crates/trusted-server-core/src/auction/formats.rs:256-268` (inline).
- **Warn when the privacy default is off**: `crates/trusted-server-core/src/auction/formats.rs:258` (inline).

### 🤔 thinking

- **Hand-mirrored `LegacyAuctionConfig` will drift**: `crates/trusted-server-core/src/config_payload.rs:52-72` (inline).
- **CLI test does not pin the working directory**: `crates/trusted-server-cli/tests/config_env_overlay.rs:36-63` (inline).

### 📌 out of scope

- **`/first-party/proxy` sign handler bypasses `to_abs` for protocol-relative URLs**: the `//`-prefixed branch builds the absolute URL inline and never calls `to_abs`, so it skips the exclusion check entirely — the exact inconsistency this PR fixes on the creative path. Pre-existing; worth a follow-up issue now that the sibling path was corrected. (`crates/trusted-server-core/src/proxy.rs:1624-1632`)
- **`[debug].inject_adm_for_testing` embeds raw, unsanitized `adm`**: correctly documented here as debug-only, but it remains the one creative path with neither sanitization nor rewriting. Follow-up issue. (`crates/trusted-server-core/src/publisher.rs:2144-2147`)

### ⛏ nitpick

- **`rewrite_mode` string is unnecessary**: `crates/trusted-server-core/src/auction/formats.rs:264-268` (inline).

## Verification notes

- The docs change from "env vars baked at build time" to "typed CLI overlay" was verified accurate: no `TRUSTED_SERVER__` handling remains in `crates/trusted-server-core/build.rs`.
- `[debug].inject_adm_for_testing` referenced in the new docs exists as described (`settings.rs:1923`).
- `convert_to_openrtb_response` has exactly two production callers, both in `auction/endpoints.rs` (219, 340), so the new flag's blast radius matches what the docs claim.

## CI Status

Taken from the PR's own checks — not re-run locally.

- fmt: PASS
- clippy / CodeQL (`Analyze (rust)`, `Analyze (javascript-typescript)`, `Analyze (actions)`): PASS
- rust tests (`cargo test`, axum native, cloudflare, spin, ts CLI native, cross-adapter parity): PASS
- js tests (vitest), format-typescript, format-docs: PASS
- integration tests, Fastly EC lifecycle, browser integration tests: still pending at review time

crates/trusted-server-core/src/creative.rs
ChristianPavilonis: Addressed in 51a374dc. The PR description and CHANGELOG now call out the protocol-relative exclusion fix and its excluded absolute `/first-party/sign` validation consequence.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in 51a374dc. `rewrite_creatives` is now read once before the bid loop and logged directly as a boolean.

crates/trusted-server-core/src/creative.rs
ChristianPavilonis: Addressed in 51a374dc. Added the Unreleased `Fixed` entry, updated the PR description, and added `proxy_sign_rejects_excluded_absolute_url` coverage for the validator behavior.

crates/trusted-server-core/src/auction_config_types.rs
ChristianPavilonis: Addressed in 51a374dc. The field docs and `is_default_rewrite_creatives` predicate now explain the `deny_unknown_fields` rollback contract.

trusted-server.example.toml
ChristianPavilonis: Confirmed and pinned in 51a374dc. `config diff` compares typed serialized payloads: the new CLI integration test pushes a TOML config with explicit default `true`, verifies the blob omits it, then asserts `config diff --local` reports `# no changes`.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in 51a374dc. `creative::process_auction_creative` now owns mandatory sanitization plus optional rewriting, with direct enabled/disabled unit coverage.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in 51a374dc. Shared settings finalization now emits one warning when enabled auctions disable rewriting, including the direct third-party contact consequence.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in 51a374dc. The derived mode string was removed; the existing debug log now records the hoisted boolean directly.

crates/trusted-server-core/src/config_payload.rs
ChristianPavilonis: Addressed in 51a374dc. `LegacyAuctionConfig` is now documented as the intentional pre-`rewrite_creatives` rollback snapshot and warns maintainers not to add later fields.

crates/trusted-server-cli/tests/config_env_overlay.rs
ChristianPavilonis: Addressed in 51a374dc. Both existing spawned commands, plus the new diff command, explicitly run in the temporary project directory.

aram356 (changes requested):
## Summary

Re-review after the "Address creative rewriting review feedback" commits. Every finding from the previous review was addressed and verified (including that the new config-load `log::warn!` fires on the production blob path, not just TOML parse). One new blocking gap remains: the `/first-party/sign` protocol-relative branch does not honor `exclude_domains`, which is inconsistent with this PR's own "Fixed" changelog entry. Details inline on `crates/trusted-server-core/src/proxy.rs`.

## Blocking

### 🔧 wrench

- **`/first-party/sign` bypasses `exclude_domains` for protocol-relative URLs** (`crates/trusted-server-core/src/proxy.rs:1642-1648`)

  The CHANGELOG adds: *"Protocol-relative creative URLs now honor `rewrite.exclude_domains` … excluded absolute URLs submitted to `/first-party/sign` are rejected."* But the sign handler's `//` branch builds `abs` inline (`format!("{}:{}", default_scheme, trimmed)`) and never runs the exclusion check, while the `else` branch routes through `to_abs`, which now does. Net result:

  - `https://cdn.example/asset.js` on an excluded host → **rejected** (`proxy_sign_rejects_excluded_absolute_url` proves it);
  - `//cdn.example/asset.js` on the **same** excluded host → **still signed and proxied**.

  `build_proxy_url_with_extras` → `build_signed_url_for` performs no exclusion check either, so nothing downstream catches it. The two changelog clauses are each individually accurate (one is scoped to "absolute"), but the asymmetry undercuts the protocol-relative headline.

  The `//` branch can't simply call `to_abs` — it intentionally preserves the request scheme rather than hardcoding `https:` — so the fix is a uniform post-check after `abs` is computed, before signing:

  ```rust
  if settings.rewrite.is_excluded(&abs) {
      return Err(Report::new(TrustedServerError::Proxy {
          message: "unsupported url".to_string(),
      }));
  }
  ```

  Either apply that (and add a `//`-host case to `proxy_sign_rejects_excluded_absolute_url`), or narrow the changelog to state the sign path enforces exclusions only for absolute URLs and explain why protocol-relative is intentionally exempt.

## Non-blocking

### 🌱 seedling

- **`[debug].inject_adm_for_testing` remains the one creative path with neither sanitization nor rewriting**: pre-existing and correctly documented here as debug-only. Now that auction `adm` is uniformly sanitized, this raw-`adm` debug path is the remaining asymmetry — worth a follow-up issue so it is not mistaken for a production-safe toggle. (`crates/trusted-server-core/src/publisher.rs:2144-2147`)

### 📝 note

- **Prior review resolutions confirmed**: `to_abs` behavior change now carries a `### Fixed` CHANGELOG entry plus proxy-validator test coverage; `skip_serializing_if` has an explanatory doc comment; `config diff` no-op is asserted; the sanitize+conditional-rewrite policy is centralized in `creative::process_auction_creative`; the disabled-rewrite `warn` fires via `finalize_deserialized` on both TOML and blob load paths; `LegacyAuctionConfig` carries a drift-guard comment; the CLI test pins its working directory.
- **New public API surface is intentional**: `creative` is `pub mod`, so `process_auction_creative` is public — consistent with its `pub` siblings `sanitize_creative_html` / `rewrite_creative_html`.

## CI Status

Taken from the PR's own checks (full run, all green):

- fmt: PASS
- clippy / CodeQL (`Analyze (rust)`, `Analyze (javascript-typescript)`, `Analyze (actions)`): PASS
- rust tests (`cargo test`, axum native, cloudflare, spin, ts CLI native, cross-adapter parity): PASS
- integration tests, Fastly EC lifecycle, browser integration tests: PASS
- js tests (vitest), format-typescript, format-docs: PASS

crates/trusted-server-core/src/proxy.rs
```diff
@@ -2408,6 +2408,28 @@ mod tests {
...
+            let req = build_http_post_json_request("https://edge.example/first-party/sign", &body);
+            let err: Report<TrustedServerError> =
+                handle_first_party_proxy_sign(&settings, &noop_services(), req)
+                    .await
+                    .expect_err("should reject excluded URL");
+
+            assert_eq!(
+                err.current_context().status_code(),
+                StatusCode::BAD_GATEWAY,
+                "should reject excluded absolute URLs as unsupported"
```
aram356: 🔧 **wrench** — This test covers the *absolute* excluded URL, but the sibling **protocol-relative** case is not handled and slips through.

In `handle_first_party_proxy_sign` (proxy.rs:1642-1648) the `//` branch builds `abs` inline (`format!("{}:{}", default_scheme, trimmed)`) and never runs the exclusion check, while the `else` branch routes through `to_abs`, which now does. So:

- `https://cdn.example/asset.js` on an excluded host → **rejected** (this test);
- `//cdn.example/asset.js` on the **same** excluded host → **still signed and proxied**.

`build_proxy_url_with_extras` → `build_signed_url_for` performs no exclusion check either, so nothing downstream catches it. This contradicts the PR's own `### Fixed` headline (*"Protocol-relative creative URLs now honor rewrite.exclude_domains"*).

**Fix** — a uniform post-check after `abs` is computed, before signing (the `//` branch can't just call `to_abs` because it intentionally preserves the request scheme instead of hardcoding `https:`):

```rust
if settings.rewrite.is_excluded(&abs) {
    return Err(Report::new(TrustedServerError::Proxy {
        message: "unsupported url".to_string(),
    }));
}
```

Then extend this test with a `//cdn.example/asset.js` case. Alternatively, narrow the changelog to state the sign path enforces exclusions only for absolute URLs and explain why protocol-relative is intentionally exempt.

aram356 (commented):
Inline version of the blocking finding from the prior change-request review (that finding could not be anchored inline because the sign-handler lines are unchanged by this PR; anchoring it here on the new test, which is what needs extending).

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -250,21 +252,21 @@ pub fn convert_to_openrtb_response(
         let width = to_openrtb_i32(bid.width, "width", &bid_context);
         let height = to_openrtb_i32(bid.height, "height", &bid_context);
 
-        // Process creative HTML if present - — sanitize dangerous markup first, then rewrite URLs.
+        // Process creative HTML if present — always sanitize dangerous markup first.
         let creative_html = if let Some(ref raw_creative) = bid.creative {
-            let sanitized = creative::sanitize_creative_html(raw_creative);
-            let rewritten = creative::rewrite_creative_html(settings, &sanitized);
+            let processed = creative::process_auction_creative(settings, raw_creative);
```
prk-Jr, line 257: 🔧 **wrench** — This is the only creative path the new setting reaches. The server-side auction (SSAT) inline path in `build_bid_map` still rewrites unconditionally:

```rust
// crates/trusted-server-core/src/publisher.rs:3259-3274
if let Some(ref raw_creative) = bid.creative {
    let priced = crate::creative::expand_auction_price_macro(raw_creative, cpm);
    let sanitized = crate::creative::sanitize_creative_html(&priced);
    let adm = crate::creative::rewrite_inline_creative_html(
        settings,
        &base_origin,
        &sanitized,
    );
    if !adm.is_empty() {
        obj.insert("adm".to_string(), serde_json::Value::String(adm));
    }
}
```

That has been the production render path for server-side auctions since #899, and it is what populates `window.tsjs.bids`. So with `[auction].rewrite_creatives = false`, `POST /auction` returns sanitized-only markup while page-rendered SSAT creatives still get first-party proxy/click URLs and `data-tsclick` — the opposite of what an operator setting an `[auction].*` key would expect.

**Fix**: gate the `rewrite_inline_creative_html` call on `settings.auction.rewrite_creatives` (keeping `sanitize_creative_html` unconditional), or scope the setting name and docs explicitly to `POST /auction`. Either way, add a `build_bid_map` test for the disabled case — none exists today.

docs/guide/creative-processing.md
```diff
@@ -41,16 +41,47 @@ Creative processing transforms third-party ad creatives by rewriting URLs to go
...
+style blocks, forms, event handlers, or other content removed by sanitization.
+The client render bridge still creates and sizes its sandboxed iframe from bid
+or ad-unit dimensions, but the creative TSJS runtime and its click/resource
+protections are not injected. Sanitizer acceptance is not a host allowlist;
+ordinary accepted HTTP(S) URLs may cause the browser to contact external
+creative hosts directly.
+:::
+
+This setting does not affect HTML/CSS response rewriting under
+`/first-party/proxy`. It also does not change the separate debug-only
```
prk-Jr: 🔧 **wrench** — Both claims here are inaccurate.

1. The page-bids `adm` path is not gated by `[debug].inject_adm_for_testing`. That flag is passed as `include_debug_bid` (`publisher.rs:3698`) and gates only the `debug_bid` diagnostic blob (`publisher.rs:3281`). The `adm` insertion at `publisher.rs:3259-3274` is unconditional.
2. That `adm` is never raw — it is `sanitize_creative_html` followed by `rewrite_inline_creative_html`.

As written, this tells operators that the primary SSAT render path is a non-production diagnostic, which it is not. Please correct it alongside the gating decision for `build_bid_map`.

docs/guide/auction-orchestration.md
```diff
@@ -553,9 +554,36 @@ EC identity is maintained with the `ts-ec` cookie; auction responses do not emit
...
+Disabling rewriting does not disable or reverse sanitization. Scripts,
+stylesheets, style blocks, forms, event handlers, dangerous URL schemes, and
+other rejected content remain removed. It also removes the injected creative
+runtime and first-party proxy/click mediation from the resulting `adm`.
+Sanitizer-accepted hosts are not allowlisted or trusted merely because their
+URLs remain in the output.
+
+The setting applies only to the shared `POST /auction` response converter.
+HTML/CSS returned by `/first-party/proxy` continues to be rewritten. The
+separate debug-only `[debug].inject_adm_for_testing` publisher and page-bids
```
prk-Jr: 🔧 **wrench** — Same inaccuracy as `docs/guide/creative-processing.md:80-82`: the publisher page-bids `adm` path is not gated by `[debug].inject_adm_for_testing` (that flag only controls the `debug_bid` blob at `publisher.rs:3281`), and its `adm` is always sanitized and rewritten via `rewrite_inline_creative_html` (`publisher.rs:3259-3274`) rather than raw. Both guides need the same correction.

crates/trusted-server-core/src/proxy.rs
```diff
@@ -2408,6 +2408,28 @@ mod tests {
         });
     }
 
+    #[test]
+    fn proxy_sign_rejects_excluded_absolute_url() {
```
prk-Jr: 🔧 **wrench** — Good test, but it only covers the absolute form. The protocol-relative form still bypasses the exclusion check entirely:

```rust
// crates/trusted-server-core/src/proxy.rs:1642-1655
let abs = if trimmed.starts_with("//") {
    // builds `<scheme>:<trimmed>` — never calls `to_abs`, so `is_excluded` never runs
    format!("{}:{}", default_scheme, trimmed)
} else {
    crate::creative::to_abs(settings, trimmed).ok_or_else(|| { /* ... */ })?
};
```

So `//excluded.example/a.png` is signed while `https://excluded.example/a.png` is rejected. Creative HTML now correctly leaves the protocol-relative URL direct, but the signing endpoint will still mint a proxy token for the same asset — the exclusion fix is incomplete at exactly the boundary the CHANGELOG entry calls out.

**Fix**: run the exclusion check on the constructed absolute URL in that branch too, and add a `proxy_sign_rejects_excluded_protocol_relative_url` test.

crates/trusted-server-core/src/proxy.rs
```diff
@@ -2408,6 +2408,28 @@ mod tests {
...
+            });
+            let req = build_http_post_json_request("https://edge.example/first-party/sign", &body);
+            let err: Report<TrustedServerError> =
+                handle_first_party_proxy_sign(&settings, &noop_services(), req)
+                    .await
+                    .expect_err("should reject excluded URL");
+
+            assert_eq!(
+                err.current_context().status_code(),
+                StatusCode::BAD_GATEWAY,
```
prk-Jr: 🤔 **thinking** — `unsupported url` maps to `BAD_GATEWAY` for a client-supplied URL; `400` would fit better. Pre-existing behavior, but this test now pins 502 as the contract, so it is worth a deliberate decision rather than an inherited one.

crates/trusted-server-core/src/creative.rs
```diff
@@ -56,30 +56,21 @@ pub(super) fn to_abs(settings: &Settings, u: &str) -> Option<String> {
...
-    // Skip non-network schemes commonly found in creatives
-    let lower = t.to_ascii_lowercase();
-    if lower.starts_with("data:")
-        || lower.starts_with("javascript:")
-        || lower.starts_with("mailto:")
-        || lower.starts_with("tel:")
-        || lower.starts_with("blob:")
-        || lower.starts_with("about:")
-    {
+    // Match exclusions against the same absolute URL used for rewriting.
```
prk-Jr, line 68: 👍 **praise** — Right root-cause fix. `Rewrite::is_excluded` parses with `url::Url::parse`, which fails on `//host/path` and silently returned `false`, so protocol-relative creative URLs bypassed `exclude_domains` entirely. Normalizing before the check is the correct ordering.

Also worth noting for other reviewers: dropping the explicit `data:` / `javascript:` / `mailto:` / `tel:` / `blob:` / `about:` skip-list is behavior-preserving — none of those match `//` or `http(s)://`, so they fall through to the final `else` and still return `None`.

crates/trusted-server-core/src/auction_config_types.rs
```diff
@@ -11,6 +11,18 @@ pub struct AuctionConfig {
...
 
+    /// Rewrite sanitized winning-bid creative HTML to first-party endpoints.
+    ///
+    /// The default must stay omitted from serialized config blobs: older
+    /// [`AuctionConfig`] schemas reject unknown fields during binary rollback.
+    /// An explicit `false` remains serialized and requires restoring a
+    /// compatible blob before rolling back.
+    #[serde(
+        default = "default_rewrite_creatives",
+        skip_serializing_if = "is_default_rewrite_creatives"
```
prk-Jr, line 22: 👍 **praise** — Correct handling of a defaulted field on a `deny_unknown_fields` struct: omitting the default keeps stored blobs readable by a previous binary, while an explicit `false` stays serialized. The doc comment states the rollback constraint precisely, and `LegacyAuctionConfig` in `config_payload.rs` pins it with a test rather than leaving it as prose.

crates/trusted-server-core/src/config_payload.rs
```diff
@@ -46,6 +46,27 @@ mod tests {
...
     use crate::redacted::Redacted;
     use crate::test_support::tests::crate_test_settings_str;
+    use serde::Deserialize;
+
+    // Intentionally mirrors `AuctionConfig` before `rewrite_creatives` existed.
+    // Do not add fields introduced after that snapshot: this test proves a
+    // default payload remains readable by the previous binary schema.
+    #[derive(Deserialize)]
+    #[serde(deny_unknown_fields)]
+    struct LegacyAuctionConfig {
```
prk-Jr, line 56: 👍 **praise** — A `deny_unknown_fields` mirror of the pre-change schema is the right way to prove rollback compatibility: any future `AuctionConfig` field added without `skip_serializing_if` fails this test instead of failing in production during a rollback. The comment warning against extending the struct is the part that keeps it useful.

prk-Jr (changes requested):
## Summary

Makes winning-bid creative rewriting configurable via a default-true `[auction].rewrite_creatives`, keeps sanitization mandatory, and fixes protocol-relative URL matching against `rewrite.exclude_domains`. The serialization design (default omitted, explicit `false` retained) and its rollback rationale are solid and well covered by tests. Three blocking issues: the setting does not reach the server-side auction (SSAT) inline `adm` render path, two guides describe that path incorrectly, and the protocol-relative exclusion fix does not reach `/first-party/sign`.

## Blocking

### 🔧 wrench

- **`[auction].rewrite_creatives` is not honored on the SSAT inline `adm` path**: `build_bid_map` calls `sanitize_creative_html` then `rewrite_inline_creative_html` unconditionally (`crates/trusted-server-core/src/publisher.rs:3259-3274`). That is the production server-side-auction render path since #899 and it is what populates `window.tsjs.bids`. With `rewrite_creatives = false`, `POST /auction` returns sanitized-only markup while page-rendered SSAT creatives still receive first-party proxy/click URLs and `data-tsclick`. Either gate that rewrite call on the setting (keeping sanitization unconditional), or scope the setting name and docs explicitly to `POST /auction`. No test covers the SSAT path under the new flag.

- **Two guides describe the page-bids `adm` path incorrectly** (`docs/guide/creative-processing.md:80-82`, `docs/guide/auction-orchestration.md:583-584`): both state the path is "debug-only `[debug].inject_adm_for_testing`" and "may include raw `adm`". `inject_adm_for_testing` is passed as `include_debug_bid` (`publisher.rs:3698`) and gates only the `debug_bid` diagnostic blob (`publisher.rs:3281`). The `adm` insertion at `publisher.rs:3259-3274` is unconditional, and the value is always sanitized and rewritten — never raw.

- **Protocol-relative exclusion fix does not reach `/first-party/sign`** (`crates/trusted-server-core/src/proxy.rs:1642-1655`): the `trimmed.starts_with("//")` branch builds the absolute URL from the request scheme and bypasses `to_abs` entirely, so no `is_excluded` check runs. `https://excluded.example/a.png` is now rejected (covered by the new test), but `//excluded.example/a.png` is still signed. Creative HTML leaves that URL direct while the signing endpoint mints a proxy token for it. Add the exclusion check to that branch and a matching protocol-relative test.

## Non-blocking

### 🤔 thinking

- **502 for client-supplied URLs at `/first-party/sign`**: `unsupported url` maps to `BAD_GATEWAY`; `400` fits client input better. Pre-existing, but the new test now pins 502 as the contract.

### 👍 praise

- Rollback-safe serialization: `skip_serializing_if` plus the `deny_unknown_fields` `LegacyAuctionConfig` mirror test is the right way to prove a defaulted field stays readable by the previous binary schema.
- `to_abs` exclusion fix addresses the actual root cause — `is_excluded` parses with `url::Url::parse`, which fails on `//host/path` and silently returned `false`. Dropping the explicit `data:`/`javascript:`/`blob:` skip-list is safe: the final `else` returns `None` for all of them.

## CI Status

All 19 checks pass (fmt, clippy/check across fastly, axum, cloudflare, spin, parity, CLI tests, vitest, docs format, integration and browser suites).

crates/trusted-server-core/src/proxy.rs
ChristianPavilonis: Addressed in `710bf493`. After either URL form is normalized, `/first-party/sign` now checks `rewrite.exclude_domains` before parsing or signing. `proxy_sign_rejects_excluded_urls` covers both absolute and protocol-relative inputs.

crates/trusted-server-core/src/auction/formats.rs
ChristianPavilonis: Addressed in `710bf493`. `creative::process_inline_auction_creative` now centralizes mandatory sanitization plus optional foreign-origin rewriting, and `build_bid_map` uses it after expanding the auction price macro. The new disabled-mode regression proves SSAT `adm` stays sanitized while accepted resource and click URLs remain direct.

docs/guide/creative-processing.md
ChristianPavilonis: Addressed in `710bf493`. The guide now describes SSAT/page-bids as a production path that always sanitizes and follows `rewrite_creatives`. It also clarifies that `inject_adm_for_testing` controls only `debug_bid` and the testing-only direct GAM replacement, not `adm` delivery.

docs/guide/auction-orchestration.md
ChristianPavilonis: Addressed in `710bf493`. The orchestration guide now describes the production SSAT/page-bids path accurately, including mandatory sanitization, conditional absolute URL rewriting, and the separate role of `inject_adm_for_testing`.

crates/trusted-server-core/src/proxy.rs
ChristianPavilonis: Addressed in `710bf493`. The sign handler now checks exclusions after constructing the absolute target, and the regression covers both `https://cdn.example/asset.js` and `//cdn.example/asset.js`.

crates/trusted-server-core/src/proxy.rs
ChristianPavilonis: Deliberately left unchanged. `TrustedServerError::Proxy` maps to `BAD_GATEWAY` across the existing proxy validation paths; changing the assertion would not change endpoint behavior, while reclassifying that broader error contract is outside this PR. The test therefore preserves the current status contract.

ChristianPavilonis:
Implemented and pushed the remaining review fixes in `710bf493`.

- `[auction].rewrite_creatives` now controls the production SSAT/page-bids inline `adm` path while sanitization remains mandatory.
- `/first-party/sign` now rejects excluded protocol-relative targets as well as excluded absolute targets.
- The creative-processing, configuration, orchestration, auction README, and changelog wording now describes both auction delivery contexts accurately.
- The pre-existing `TrustedServerError::Proxy` → 502 mapping is intentionally unchanged in this PR.

Validation completed locally:
- `cargo test-fastly`, `cargo test-axum`, `cargo test-cloudflare`, `cargo test-spin`
- cross-adapter parity and native CLI tests
- all target-matched Fastly/Axum/Cloudflare/Spin clippy aliases
- Rust, JS, and docs formatting
- Vitest (425 tests)

crates/trusted-server-core/src/creative.rs
```diff
@@ -515,6 +506,46 @@ pub fn sanitize_creative_html(markup: &str) -> String {
     String::from_utf8(out).unwrap_or_default()
 }
 
+/// Sanitize auction creative HTML, then optionally rewrite it to first-party endpoints.
+///
+/// Sanitization is mandatory in both modes. Rewriting is controlled by
+/// [`crate::auction_config_types::AuctionConfig::rewrite_creatives`].
+#[must_use]
+pub fn process_auction_creative(settings: &Settings, raw: &str) -> String {
```
aram356: 🌱 **seedling** — `process_auction_creative` is `pub` but has no caller outside `trusted-server-core` (only `auction::formats` uses it), while its sibling `process_inline_auction_creative` just below is deliberately `pub(crate)`. CLAUDE.md advises avoiding unnecessary `pub`, so `pub(crate)` would be tighter and match the inline sibling.

Non-blocking: the other creative entry points (`sanitize_creative_html`, `rewrite_creative_html`, `rewrite_inline_creative_html`) are all `pub` too, so leaving this as-is is a defensible consistency call. Flagging only for your judgment.

aram356 (approved):

prk-Jr (approved):

PR #922, opened 2026-07-16
Trace winning auction bids to rendered creatives on the page

### LINKED ISSUE #920: Design how to trace winning TS auction to corresponding creatives on the page
### Description

Design how to trace winning TS auction to corresponding creatives on the page

prk-Jr:
## Summary

- Verifying that a winning server-side auction bid actually rendered on the page was impossible: no TS-owned identifier survived from the auction to the DOM on either render path. This PR mints a trace tuple — `(auction_id, slot_id, bidder, bid_id, crid, adm_hash)` — server-side and carries it end-to-end.
- Server-side ad templates (`creative_opportunities`) and SPA `/__ts/page-bids` now inject `hb_auction_id`/`hb_crid`/`hb_adm_hash` per slot; `/auction` responses carry `bid.ext.ts = {auction_id, adm_hash}` and pass through the upstream `crid`/bid `id` (with the legacy synthetic fallback when absent).
- The client stamps rendered slot elements and iframes with `data-ts-*` attributes, keeps a per-slot registry at `window.tsjs.renders`, and fires a `tsjs:adRendered` CustomEvent — so a creative in the DOM joins back to the server-side `auction winner:` / `auction delivered creative:` log lines by inspection, and e2e tooling can await and assert renders. The `adm_hash` (16-hex SHA-256 prefix of the delivered markup) enables byte-exact content verification on the `/auction` iframe path and the `inject_adm_for_testing` path.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/auction/types.rs` | Add `Bid.crid` and `Bid.bid_id` (serde-default, skip-if-none); add `adm_trace_hash()` and `Bid::creative_trace_hash()` with tests |
| `crates/trusted-server-core/src/auction/mod.rs` | Re-export `adm_trace_hash` |
| `crates/trusted-server-core/src/auction/orchestrator.rs` | Log one `auction winner:` trace line per winning bid on the sync (`run_auction`) and dispatched-collect paths |
| `crates/trusted-server-core/src/auction/formats.rs` | `/auction` bids carry `ext.ts = {auction_id, adm_hash}` (hash of post-sanitize/rewrite markup), pass through upstream `crid`/bid `id` with legacy fallback; `auction delivered creative:` info log; tests |
| `crates/trusted-server-core/src/publisher.rs` | `build_bid_map` emits `hb_auction_id`/`hb_crid`/`hb_adm_hash`; auction ID threaded through `write_bids_to_state` (both streaming call sites) and `handle_page_bids`; `debug_bid` gains `bid_id`/`crid`; tests |
| `crates/trusted-server-core/src/integrations/prebid.rs` | Extract `crid` and bid `id` from PBS bid responses; tests |
| `crates/trusted-server-core/src/integrations/adserver_mock.rs` | Restore `crid`/`bid_id` from the original SSP bid when parsing mediation responses |
| `crates/trusted-server-core/src/integrations/aps.rs` | New `Bid` fields set to `None` (APS carries no creative pre-decode) |
| `crates/trusted-server-core/src/auction/telemetry.rs` | Test fixture updated for new `Bid` fields |
| `crates/trusted-server-js/lib/src/core/trace.ts` | New: `window.tsjs.renders` registry, `data-ts-*` stamping (stale attributes removed on re-render), `tsjs:adRendered` event |
| `crates/trusted-server-js/lib/src/core/auction.ts` | `parseAuctionResponse` retains `auctionId` (bid `ext.ts` first, top-level `id` fallback) and `admHash` |
| `crates/trusted-server-js/lib/src/core/request.ts` | `renderCreativeInline` stamps container + iframe and records renders (including `rendered:false` for rejected/missing-slot) |
| `crates/trusted-server-js/lib/src/core/types.ts` | `AuctionBidData` trace fields, `RenderRecord`, `TsjsApi.renders`, debug-bid mirror fields |
| `crates/trusted-server-js/lib/src/integrations/gpt/index.ts` | `slotRenderEnded` stamps + records; pbRender bridge records `debug-adm`/`pbs-cache` serves with the slot element captured at message-receipt time (SPA-navigation safe); debug adm iframe marked |
| `crates/trusted-server-js/lib/test/core/trace.test.ts` | New: registry, count/overwrite, event, stamping, stale-attribute removal |
| `crates/trusted-server-js/lib/test/core/auction.test.ts` | Auction-ID retention and `ext.ts` precedence tests |
| `crates/trusted-server-js/lib/test/core/request.test.ts` | Stamping/registry/event tests, rejected-render marker test |
| `crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts` | `slotRenderEnded` stamping/registry test incl. empty-render and re-render count |

## Closes

Closes #920

## Test plan

- [x] `cargo test-fastly && cargo test-axum` (also `cargo test-cloudflare && cargo test-spin` and the integration parity suite)
- [x] `cargo clippy-fastly && cargo clippy-axum` (also `clippy-cloudflare`, `clippy-cloudflare-wasm`, `clippy-spin-native`, `clippy-spin-wasm`)
- [x] JS tests: `cd crates/trusted-server-js/lib && npx vitest run` (415 passed)
- [x] Other: new unit tests for `adm_trace_hash`, bid-map trace fields, `/auction` `ext.ts`/crid/bid-id passthrough, PBS crid/bid-id extraction, and JS trace registry/stamping on both render paths

## Checklist

- [x] New code has tests


crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -921,10 +927,97 @@ function syncPrebidEidsCookie(): void {
...
+ */
+export function installPrebidRenderTrace(): void {
+  if (typeof window === 'undefined') return;
+  const p = pbjs as unknown as {
+    onEvent?: (event: string, handler: (bid: PrebidWonBid) => void) => void;
+    __tsRenderTraceInstalled?: boolean;
+  };
+  if (typeof p.onEvent !== 'function' || p.__tsRenderTraceInstalled) return;
+  p.__tsRenderTraceInstalled = true;
+  p.onEvent('bidWon', (bid) => {
```
ChristianPavilonis: **P1 — `bidWon` is not proof that the creative rendered successfully**

This listener treats Prebid's `bidWon` event as authoritative and immediately records `rendered: true` and `injected: true`. In Prebid 10.26, `bidWon` is emitted when the bid is marked as the winner; an asynchronous renderer can subsequently fail. Prebid exposes a separate `adRenderSucceeded` event after the render function succeeds ([source](https://github.com/prebid/Prebid.js/blob/a5197e75d8f40212f3407e57ba81d039d0a687f1/src/adRendering.ts#L309-L335)). The current code can therefore emit `tsjs:adRendered`, stamp the DOM, and show a confirmed green render for a creative that failed.

Please record confirmation from `adRenderSucceeded` and read its `event.bid`. Optionally handle `adRenderFailed` to record a failed attempt, and add a test proving `bidWon` followed by render failure does not produce a confirmed record.

crates/trusted-server-js/lib/src/core/auction.ts
```diff
@@ -146,6 +156,12 @@ export function parseAuctionResponse(body: any): AuctionBid[] {
         seat,
         creativeId: b.crid ?? `${seat}-${b.impid ?? ''}`,
         adomain: Array.isArray(b.adomain) ? b.adomain : [],
+        auctionId:
```
ChristianPavilonis, line 166: **P1 — The advertised `bid_id` never reaches the browser trace tuple**

Although Rust places the upstream bid ID in OpenRTB `bid.id`, `parseAuctionResponse` ignores `b.id`, `RenderRecord` has no `bidId`, and no `data-ts-bid-id` is stamped. On the SSAT path, `bid_id` becomes `hb_adid` only as a last-resort fallback; when a cache ID or ad ID exists, the upstream bid ID is lost. This means the promised `(auction_id, slot_id, bidder, bid_id, crid, adm_hash)` contract is not met on either normal path, preventing exact correlation with the new `auction winner:` log's `bid_id`.

Please propagate a distinct bid-ID field through both paths: `hb_bid_id`/`AuctionBid.bidId` → Prebid `meta.tsBidId` → `RenderRecord.bidId` → `data-ts-bid-id`. It should not be overloaded onto `hb_adid`.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -445,6 +477,73 @@ function installInitialLoadDetector(ts: TsjsApi): void {
...
+ * later *replaced*. So instead of delaying the initial ad request, this detects
+ * the swap after the fact and re-runs `adInit()`, which destroys the orphaned
+ * slots and re-binds against the live DOM (reusing the publisher's own slot for
+ * that div when they have since defined one).
+ *
+ * Bounded by {@link MAX_ORPHAN_RECONCILE_ATTEMPTS} and
+ * {@link ORPHAN_RECONCILE_WINDOW_MS} so a page whose DOM never settles cannot
+ * spin on ad requests.
+ */
+function watchForOrphanedSlots(ts: TsjsApi): void {
```
ChristianPavilonis, line 703: **P1 — Hydration recovery can replay the previous route's targeting during SPA navigation**

This orphan observer remains active when `onNavigate` begins. If the new DOM commits quickly but `/__ts/page-bids` takes longer than the 250 ms debounce, the observer calls `ts.adInit()` while `ts.adSlots` and `ts.bids` still describe the previous route. `adInit()` can then bind the old auction to the new route's matching div prefix and issue another GAM request, risking stale creatives, duplicate billable requests, and consumption of the recovery budget during ordinary navigation.

Please stop or invalidate the orphan watch immediately when navigation to a different pathname starts. Capture an ad-init/navigation generation in the watcher and only rebind if it still matches. A regression test should defer the page-bids response while performing a fast DOM route swap.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -933,6 +1124,7 @@ export function installTsRenderBridge(): void {
           })
         );
         fireWinBillingBeacons(slotId, matchedBid);
+        recordBridgeRender(slotId, matchedBid, 'pbs-cache', slotEl);
```
ChristianPavilonis: **P1 — A stale PBS Cache completion can overwrite the new route's render state**

After the asynchronous cache fetch resolves, this records the old creative without revalidating the source iframe, route generation, or current bid. `recordBridgeRender` writes the registry entry before checking `el.isConnected`; that check only suppresses DOM stamping. After SPA navigation, an old auction can therefore become the current `window.tsjs.renders[slotId]` and emit a false render event even when its element was detached.

Before recording, verify that the captured source is still connected to the same slot and that the current bid's `hb_adid`/`hb_auction_id` matches the captured bid. Abort outstanding cache requests on navigation where possible.

crates/trusted-server-js/lib/src/core/trace.ts
```diff
@@ -0,0 +1,485 @@
...
+ * history is trimmed from the front rather than growing without limit.
+ */
+const MAX_RENDER_LOG_ENTRIES = 200;
+
+/**
+ * Page-global render counter backing [`RenderRecord.seq`]. Module-scoped
+ * rather than stored on `window.tsjs` so a re-executed bundle cannot resume
+ * mid-sequence and hand two different renders the same number.
+ */
+let renderSeq = 0;
```
ChristianPavilonis: **P2 — The page-global sequence number is not global across generated bundles**

`build-all.mjs` builds core, GPT, and the other integrations as separate self-contained IIFEs, so each inlined copy of this module starts at `seq = 1`. Normal cross-path tracing can produce duplicate `#1` rows and badges. For the same slot, `isCurrentRender` compares only `seq`, so multiple history rows can even be marked current.

Please keep the counter on shared `window.tsjs`, and preferably use object identity as the current-record check. Add a built-bundle test that loads more than one IIFE.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -596,9 +702,46 @@ export function installTsAdInit(): void {
...
+            ? {
+                path: 'ssat' as const,
+                auctionId: bid.hb_auction_id,
+                bidder: bid.hb_bidder,
+                adId: bid.hb_adid,
+                creativeId: bid.hb_crid,
+                admHash: bid.hb_adm_hash,
+              }
+            : { path: 'gam-refresh' as const };
+          const record = recordRender({
```
ChristianPavilonis: **P2 — One Universal Creative impression is counted as two renders**

`slotRenderEnded` appends a record here, and the bridge later appends another record for the same impression. This inflates `count`, history length, sequence numbers, and panel totals. Depending on event ordering, the weaker GAM-only signal can also overwrite the confirmed bridge signal.

Please create one impression record and enrich or replace it when bridge confirmation arrives without incrementing `count`, `seq`, or history.

crates/trusted-server-js/lib/src/core/trace.ts
```diff
@@ -0,0 +1,485 @@
...
+      } else {
+        el.removeAttribute(name);
+      }
+    }
+    // Badge any slot that actually shows something, carrying its honest status
+    // colour: green ✓ for a confirmed TS render, blue ◐ for `gam-only` (GAM
+    // rendered, TS cannot confirm it as its own). Slots with nothing on screen
+    // (`empty`) or nothing visible (`hidden`) stay unbadged — there is no
+    // creative there to label. Never badge the iframe itself.
+    const status = panelStatus(record);
```
ChristianPavilonis: **P2 — Old badges survive empty or hidden re-renders**

Existing badges are removed only inside `attachTraceBadge`, which is never called when the new status is `empty` or `hidden`. A slot can therefore retain a green or blue badge after its current record says it is empty or invisible, directly contradicting the overlay's status.

Please remove the element's existing direct trace badge before deciding whether the new record merits one, and add filled→empty and visible→hidden transition tests.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -323,11 +349,14 @@ function injectAdmIntoSlot(divId: string, adm: string): void {
...
           f.height = String(slotEl!.offsetHeight || 90);
           f.setAttribute('sandbox', ADM_IFRAME_SANDBOX);
+          f.setAttribute('data-ts-injected-adm', 'true');
           f.src = innerSrc;
           slotEl!.appendChild(f);
           log.debug(`[tsjs-gpt] gam-intercept: inserted src iframe for '${divId}'`);
         }
       });
+      // Placement deferred to the animation frame — not confirmed yet.
+      return false;
```
ChristianPavilonis, line 532: **P2 — Successful deferred ADM injection remains recorded as unconfirmed**

This branch returns `false` immediately, but the animation-frame callback can subsequently insert or update the iframe successfully. Nothing then corrects the render record, so the panel permanently reports `gam-only` even though Trusted Server placed the creative.

Please make injection completion asynchronous or update the same record after the animation-frame callback, guarded by a render generation so a newer render cannot be overwritten.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -541,6 +640,13 @@ export function installTsAdInit(): void {
           if (bid[key]) gptSlot.setTargeting(key, String(bid[key]!));
         });
         gptSlot.setTargeting(TS_INITIAL_TARGETING_KEY, '1');
+        // Arm SSAT attribution for the next render of this slot only. Without
+        // this, every later publisher refresh would still read the page-load
+        // bid out of ts.bids and claim the (long finished) server-side auction
+        // rendered it.
+        (ts.ssatTargetingFresh ??= {})[slot.id] = TS_BID_TARGETING_KEYS.some((key) =>
```
ChristianPavilonis: **P2 — A late render event can consume the next route's SSAT attribution**

Attribution is a mutable boolean keyed only by logical slot ID. The listener later reads the current `ts.bids` and `divToSlotId` instead of immutable state associated with the request that produced the event. A delayed event from route A can arrive after route B's `adInit`, consume route B's freshness flag, and be stamped with route B's auction tuple; route B's real response is then mislabeled `gam-refresh`.

Please associate slot/request objects with an immutable auction/navigation generation and reject stale events. Add a rapid-navigation test where the old event arrives after the new `adInit`.

ChristianPavilonis (changes requested):
## Summary

Requesting changes due to four high-severity and five medium-severity correctness issues in the browser-side trace implementation. The Rust propagation is generally sound, but the current GPT/Prebid integration can report failed or stale creatives as confirmed, replay prior-route targeting, and does not propagate the promised bid ID.

The PR is also currently unmergeable (`DIRTY`), with conflicts in the Prebid integration test and `trusted-server.example.toml`. Local formatting, target-specific tests/clippy, JavaScript tests, and integration parity all passed.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
prk-Jr: Fixed in 03aa5ec06. The trace hook now listens to `adRenderSucceeded` and `adRenderFailed`, reads the rendered bid from `event.bid`, and no longer subscribes to `bidWon`. Hook-level tests verify that only a successful render is confirmed and a render failure remains `rendered: false` / `injected: false`.

crates/trusted-server-js/lib/src/core/auction.ts
prk-Jr: Fixed in 03aa5ec06. The upstream OpenRTB bid ID now travels independently through Rust `hb_bid_id`, `AuctionBid.bidId`, Prebid `meta.tsBidId`, `RenderRecord.bidId`, and `data-ts-bid-id`. `hb_adid` remains the Universal Creative bridge identifier and `hb_bid_id` is trace-only rather than GAM targeting. Both SSAT and client-side auction paths have regression coverage.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
prk-Jr: Fixed in 03aa5ec06. Navigation now calls `beginNavigation` before awaiting page-bids, which immediately disconnects the orphan observer, clears its timers, and bumps the render generation. The watcher also captures and revalidates its generation before rebinding. A regression test delays page-bids while swapping the route DOM and verifies that old bids are not replayed.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
prk-Jr: Fixed in 03aa5ec06. PBS Cache requests now capture the route generation, bid snapshot, slot element, and message source, use an `AbortController`, and revalidate the full bid fingerprint plus source/element ownership before posting the response, firing beacons, or recording the render. Navigation aborts outstanding cache requests. Regression coverage replaces the live bid before completion and verifies the stale result is dropped.

crates/trusted-server-js/lib/src/core/trace.ts
prk-Jr: Fixed in 03aa5ec06. Sequence allocation now uses the shared `window.tsjs.renderSeq`, with the module counter only as a fallback, and current-row detection uses render-record object identity. The regression test builds and executes two separate IIFEs and verifies that their sequence values remain globally monotonic.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
prk-Jr: Fixed in 03aa5ec06. GAM and Universal Creative bridge signals now share an open impression record and enrich it in place via `updateRender`, preserving `seq`, `count`, `at`, and history length. Confirmation fields cannot be downgraded by a later weaker signal. Tests cover both GAM-first and bridge-first arrival orders.

crates/trusted-server-js/lib/src/core/trace.ts
prk-Jr: Fixed in 03aa5ec06. `stampCreativeTrace` now removes the existing direct badge before evaluating the new status, so empty and hidden renders leave no stale badge. Added explicit filled-to-empty and visible-to-hidden transition tests.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
prk-Jr: Fixed in 03aa5ec06. Deferred ADM placement reports completion back to the original record and updates it in place after a successful animation-frame retry. Before any deferred DOM query or mutation, the callback verifies the captured generation, connected element, and exact element identity, preventing an old route from overwriting a reused slot. Tests cover both successful promotion and stale same-element navigation.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
prk-Jr: Fixed in 03aa5ec06. SSAT attribution is now stored as immutable per-request context in a FIFO keyed by the GPT slot object, including the captured bid and render generation. A stale event retires only its own old entry and is rejected, leaving the new route attribution queued for its actual event. The rapid-navigation regression test exercises route A arriving after route B ad initialization.

crates/trusted-server-core/src/auction/formats.rs
```diff
@@ -275,15 +276,57 @@ pub fn convert_to_openrtb_response(
             String::new()
         };
 
+        // Trace hash over the exact markup delivered to the client (post
+        // sanitize/rewrite) so the client can stamp the rendered creative with
+        // a value that matches this response byte-for-byte. Logged at info so
+        // server logs join against the DOM markers without debug logging.
+        let adm_hash = (!creative_html.is_empty()).then(|| adm_trace_hash(&creative_html));
```
aram356, line 283: 🔧 **wrench** — `adm_hash` is computed over two different byte-streams, so the documented log join silently fails on the `/auction` path.

This line hashes `creative_html`, i.e. the creative **after** `sanitize_creative_html` + `rewrite_creative_html`. But `log_winning_bids` (`orchestrator.rs:174`) and `build_bid_map` (`publisher.rs:2146`) both hash the **raw** `bid.creative` via `Bid::creative_trace_hash()`.

One winning bid therefore emits two different `adm_hash` values across the two new info logs, and the value the client stamps into the DOM depends on which path served it. The doc comments on `stampCreativeTrace`, `build_bid_map`, and the `trace.ts` header all explicitly promise that the DOM marker "joins back to the server-side `auction winner:` log lines" — on the `/auction` path that join can never match.

I checked the SSAT / `inject_adm_for_testing` path and it *is* self-consistent (the bid map's `adm` is also raw `bid.creative`), so this is scoped to `/auction`.

**Fix** — either hash one canonical form on both paths, or name the two distinctly and log both on the winner line so the relationship is explicit (e.g. `adm_hash_raw` on `auction winner:`, `adm_hash_delivered` here).

---

**Additionally: this PR should be stacked on top of #916.** That PR ("Make auction creative rewriting optional") rewrites this exact block — it replaces the `rewritten` binding with a `processed` one gated on `settings.auction.rewrite_creatives`. Two consequences:

1. **Textual conflict** — #916 and #922 both edit these ~30 lines; whichever lands second will conflict here.
2. **It makes this finding config-dependent** — once rewriting is optional, `adm_hash` becomes a hash of sanitized-only output *or* sanitized+rewritten output depending on `rewrite_creatives`. The same creative would then produce different `adm_hash` values across deployments, on top of already differing from the raw hash on the `auction winner:` line.

#916 is smaller and already `MERGEABLE`, so the cheap ordering is to land it first and rebase this PR onto it, defining `adm_hash` against #916's final `processed` value with the raw/delivered naming split above.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -445,6 +650,88 @@ function installInitialLoadDetector(ts: TsjsApi): void {
...
+function watchForOrphanedSlots(ts: TsjsApi): void {
+  if (typeof MutationObserver === 'undefined' || typeof document === 'undefined') return;
+  // Re-arm: a previous window may still be open from an earlier adInit().
+  stopOrphanWatch();
+  if (orphanReconcileAttempts >= MAX_ORPHAN_RECONCILE_ATTEMPTS) return;
+
+  const generation = currentGeneration(ts);
+  orphanObserver = new MutationObserver(() => {
+    clearTimeout(orphanDebounceTimer);
+    orphanDebounceTimer = setTimeout(() => {
```
aram356, line 712: 🔧 **wrench** — the orphan-recovery debounce has no max-wait, so it can never fire on a busy page.

Every observed mutation calls `clearTimeout(orphanDebounceTimer)` and restarts the 250 ms timer, and the observer watches `document.documentElement` with `subtree: true`. Any page that mutates more often than every 250 ms during the 5 s window starves this trailing debounce indefinitely — the orphan check never runs at all.

That is the common case on the pages this targets: ad iframes resizing, lazy images, publisher animations. It also includes this PR's own trace layer — `renderTracePanel()` does `panel.replaceChildren()` on *every* render and `attachTraceBadge` appends a child into slot elements, both of which feed this observer.

So the recovery silently no-ops on exactly the busy hydrating pages it was built for, and the tests only exercise a quiet DOM.

**Fix** — add a max-wait so the check runs at least once per ~1 s regardless of churn, or replace the debounce with a bounded interval poll across the 5 s window.

Related, same block: `orphanReconcileAttempts` is module-scoped and never reset, so two recoveries consumed on the first route leave every later SPA navigation with no recovery for the rest of the page's lifetime.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -2107,10 +2120,47 @@ pub(crate) fn build_bid_map(
...
+                // always present per spec but only unique per bid instance, not a
+                // creative identifier. It still satisfies what hb_adid needs here —
+                // a stable value GAM's Universal Creative echoes back verbatim so
+                // the render bridge can find this exact winning bid — for bidders
+                // (e.g. Kargo) that return neither a cache UUID nor `adid`.
+                let hb_adid = bid
+                    .cache_id
+                    .as_deref()
+                    .or(bid.ad_id.as_deref())
+                    .or(bid.bid_id.as_deref());
```
aram356, line 2163: ❓ **question** — now that `hb_bid_id` exists, is the `hb_adid` → `bid_id` fallback still meant to ride in this PR?

The new `hb_bid_id` field added in `03aa5ec` carries the bid ID for tracing purposes separately, which is the right call. That leaves this fallback serving only its *other* purpose — the bridge's reverse lookup and GAM line-item matching — which is a serving-behaviour change, not a tracing one.

Concretely: bidders returning neither `cache_id` nor `ad_id` previously got **no** `hb_adid`, so they could not match `hb_adid`-targeted GAM line items at all. With this fallback they now can. I traced the downstream paths and the PBS Cache branch is safe (it returns early unless `hb_cache_host`/`hb_cache_path` are set, and those only come from `cache_id`). But if the pbRender bridge is not intercepting — bundle not loaded on that page, or a race before `installTsRenderBridge` runs — GAM's Universal Creative resolves an adId that maps to nothing and the slot goes blank, where previously the bid simply never won.

That may well be the intent, and the commit message argues it convincingly for Kargo. But it deserves to be flagged as a serving change rather than landing inside an observability PR — ideally as its own PR that can be reverted independently of the tracing work.

crates/trusted-server-js/lib/src/core/trace.ts
```diff
@@ -0,0 +1,582 @@
...
+
+/**
+ * Whether an element is effectively visible: connected, non-zero box, and no
+ * ancestor hiding it via `display:none`, `visibility:hidden`, or `opacity:0`.
+ *
+ * The ancestor walk is what catches a slot the publisher holds at `opacity:0`
+ * on a wrapper until its own ad code reveals it — the slot's own computed
+ * opacity is `1`, so only walking up exposes the gate.
+ */
+export function isEffectivelyVisible(el: Element | null): boolean {
```
aram356, line 91: 🤔 **thinking** — this forces synchronous layout in the ad-render hot path, in production, unconditionally.

`getBoundingClientRect()` plus `getComputedStyle()` walked up the entire ancestor chain, called on every render from `slotRenderEnded`, `renderCreativeInline`, the bridge, and the Prebid `bidWon` hook — none of them gated on the trace cookie. Both APIs flush pending layout/style, so this is a forced reflow per render to produce a field only the debug panel consumes.

**Suggestion** — compute `visible` only when the overlay is armed and leave it `undefined` otherwise. `RenderRecord.visible` is already optional and `stateSummary` already renders `vis:?` for the absent case, so nothing downstream needs to change:

```ts
visible: traceOverlayEnabled() ? isEffectivelyVisible(slotEl) : undefined,
```

crates/trusted-server-core/src/auction/orchestrator.rs
```diff
@@ -162,6 +162,29 @@ fn remaining_budget_ms(start: Instant, timeout_ms: u32) -> u32 {
...
+/// Log one structured trace line per winning bid.
+///
+/// Emits the full trace tuple — auction ID, slot, bidder, ad/cache/creative
+/// IDs, and the creative trace hash — so a rendered creative on the page
+/// (carrying the same tuple in its DOM markers) can be joined back to this
+/// auction in server logs.
+fn log_winning_bids(auction_id: &str, winning_bids: &HashMap<String, Bid>) {
+    for (slot_id, bid) in winning_bids {
+        log::info!(
+            "auction winner: auction_id={} slot_id={} bidder={} price={:?} bid_id={:?} ad_id={:?} cache_id={:?} crid={:?} adm_hash={:?}",
```
aram356, line 174: 🤔 **thinking** — this adds two permanent `info` lines per winning bid, per auction.

Between this and `auction delivered creative:` in `formats.rs`, a multi-slot page emits a per-request multiplier on log volume, always on, in production. The comment in `formats.rs` explains the `info` choice as "so server logs join against the DOM markers without debug logging" — reasonable during a debugging session, but it is a standing cost on every auction on every request, forever.

**Suggestion** — emit at `debug`, or gate on the same `[debug] trace_route_enabled` flag that already gates the overlay route, so the verbose join is available exactly when someone is tracing.

crates/trusted-server-core/src/trace_cookie.rs
```diff
@@ -0,0 +1,218 @@
...
+/// Returns `404 Not Found` while `[debug] trace_route_enabled` is false. When
+/// enabled, sets (or with `?enabled=false` clears) the `ts-trace` cookie
+/// scoped to `publisher.cookie_domain` and returns `302 Found` redirecting to
+/// `/` — landing back on the homepage confirms the toggle round-trip worked.
+///
+/// # Errors
+///
+/// Returns [`TrustedServerError::InvalidHeaderValue`] if the configured cookie
+/// domain cannot be rendered as an HTTP header value.
+pub fn handle_trace_mode(
```
aram356, line 72: 🤔 **thinking** — `GET /_ts/trace` sets a cookie with no CSRF protection.

A `GET` that mutates state via `Set-Cookie` can be triggered cross-site — `` on any third-party page arms the overlay for that visitor. `SameSite=Lax` restricts cookie *sending*, not *setting*, so it does not prevent this.

Impact is genuinely bounded: config-gated, off by default, and the overlay only surfaces data already on `window.tsjs` (as the module docs correctly note). But for an operator who enables it in production, the consequence is a floating panel and badges drawn over live ad slots for any visitor an attacker can reach — visible defacement of the ad experience.

**Suggestion** — require the existing tester cookie alongside the flag, or make the toggle a `POST`.

crates/trusted-server-core/src/trace_cookie.rs
```diff
@@ -0,0 +1,218 @@
...
+        q.split('&')
+            .any(|pair| pair == "enabled=false" || pair == "enabled=0")
+    })
+}
+
+/// Handles `GET /_ts/trace`.
+///
+/// Returns `404 Not Found` while `[debug] trace_route_enabled` is false. When
+/// enabled, sets (or with `?enabled=false` clears) the `ts-trace` cookie
+/// scoped to `publisher.cookie_domain` and returns `302 Found` redirecting to
```
aram356, line 65: ⛏ **nitpick** — two small doc issues in this block.

1. This line says the cookie is "scoped to `publisher.cookie_domain`", but `format_trace_cookie` is deliberately host-only — and its own doc comment says so explicitly and explains why. The two docs in this file contradict each other; this is the stale half.

2. The `# Errors` section below documents `TrustedServerError::InvalidHeaderValue` "if the configured cookie domain cannot be rendered as an HTTP header value" — but the cookie string is built entirely from constants (`COOKIE_TS_TRACE` and a `const u32` max-age), so `HeaderValue::from_str` cannot fail and nothing configurable feeds it. The documented error path is unreachable; consider constructing the header infallibly and dropping the `# Errors` section.

crates/trusted-server-js/lib/src/core/trace.ts
```diff
@@ -0,0 +1,582 @@
...
+  ok: { color: '#3fb950', mark: '✓', label: 'ok' },
+  hidden: { color: '#d29922', mark: '⚠', label: 'hidden' },
+  'gam-only': { color: '#58a6ff', mark: '◐', label: 'gam-only' },
+  empty: { color: '#f85149', mark: '✗', label: 'empty' },
+};
+
+/**
+ * Attach (or replace) the per-slot confirmation badge on a slot element.
+ *
+ * Only called for `ok` slots — a TS creative that actually placed and is
```
aram356, line 146: ⛏ **nitpick** — stale doc comment.

This says "Only called for `ok` slots" and "Hidden / gam-only / empty slots deliberately get none", but the caller in `stampCreativeTrace` badges `gam-only` as well (`status === 'ok' || status === 'gam-only'`) — a later commit in this series changed the behaviour and the badge colour/mark table alongside it, but not this comment. As written it documents behaviour that commit deliberately removed.

aram356 (changes requested):
## Summary

Deep review of `03aa5ec`, focused on ground the previous review did not cover — the Rust propagation, and the build/bundle and hot-path costs of the client trace layer. Two blocking correctness findings, both new.

`03aa5ec` fixes the per-bundle `renderSeq` problem properly (shared `ts.renderSeq` plus an object-identity `isCurrentRender`), and `hb_bid_id` now carries the bid ID as its own trace field. I verified both. The remaining nine findings from the existing review are not restated here.

**Ordering: this should be stacked on top of #916.** That PR rewrites the exact block in `auction/formats.rs` where this one computes `adm_hash`, and makes rewriting config-gated — which both conflicts textually and turns the `adm_hash` finding below into a config-dependent one. #916 is smaller and already `MERGEABLE`; landing it first and rebasing this onto it is the cheaper order. Detail on the `formats.rs` comment.

## Blocking

### 🔧 wrench

- **`adm_hash` is two different hashes**: `/auction` hashes post-sanitize/rewrite markup, while `auction winner:` and `build_bid_map` hash raw `bid.creative`. The documented "join the DOM marker back to `auction winner:`" contract cannot hold on the `/auction` path. Compounded by #916 — see the inline comment. (`auction/formats.rs:283`)
- **Orphan-recovery debounce has no max-wait**: any page mutating more often than every 250 ms starves the check for the whole 5 s window — including via this PR's own panel and badge DOM writes. The recovery silently no-ops on exactly the busy hydrating pages it targets, and the tests only cover a quiet DOM. (`integrations/gpt/index.ts:712`)

### ❓ question

- **Is the `hb_adid` → `bid_id` fallback still meant to ride in this PR?** Now that `hb_bid_id` carries the bid ID for tracing separately, this fallback serves only GAM line-item matching and the bridge lookup — a serving-behaviour change, not a tracing one. (`publisher.rs:2163`)

## Non-blocking

### 🤔 thinking

- **~20 KB of debug-overlay code ships on every page view.** Measured `03aa5ec` against its merge base: `tsjs-core.js` 10,031 → 18,516 bytes (**+85%**), `tsjs-gpt.js` 18,908 → 30,638 (**+62%**). The panel is runtime-gated by a cookie so it cannot be tree-shaken, and it is inlined into both bundles. This repo already has the machinery to avoid that — `.with_deferred_js()` and per-integration IIFEs. A separate `tsjs-trace.js` fetched only when `ts-trace` is armed would cost nothing on the overwhelming majority of page views that never trace.
- **`isEffectivelyVisible()` forces synchronous layout on every render, in production**, ungated by the trace cookie, from all four render paths. (`core/trace.ts:91`)
- **Two new permanent `info` log lines per winning bid, per auction.** (`auction/orchestrator.rs:174`)
- **`GET /_ts/trace` sets a cookie with no CSRF protection**; `SameSite=Lax` does not prevent cross-site *setting*. (`trace_cookie.rs:72`)

### 📌 out of scope

- **This is five features in one PR**: trace-tuple propagation, the `/_ts/trace` route + overlay, GPT orphan recovery, Prebid `bidWon` tracing, and the `hb_adid` fallback. The two riskiest — re-issuing GAM ad requests and changing `hb_adid` — are production serving changes inside an observability PR, and they are where both blockers above and several of the earlier P1s land. Splitting them out would let the tracing work land quickly and keep the risky parts independently revertable.

### ⛏ nitpick

- Stale doc comment on `attachTraceBadge` — still says `ok`-only after a later commit added `gam-only` badging. (`core/trace.ts:146`)
- `trace_cookie.rs` doc/impl mismatch (cookie described as domain-scoped but deliberately host-only) plus an unreachable documented `# Errors` path. (`trace_cookie.rs:65`)

## CI Status

Run locally against `03aa5ec` in a clean worktree:

- fmt (`cargo fmt --all -- --check`): **PASS**
- clippy (`cargo clippy-fastly`): **PASS**
- rust tests (`cargo test-axum`): **PASS** — 20 tests
- js tests (`npx vitest run`): **PASS** — 461 tests across 31 files
- GitHub checks: only `Analyze (javascript-typescript)` reported — **SUCCESS**

No new Rust dependencies — `sha2` and `hex` are already workspace deps of `trusted-server-core`. The conflicts noted in the previous review are resolved; the PR is `MERGEABLE` against `main` today, though see the #916 stacking note above.

PR #767, opened 2026-06-12
Apply neutral language framing across documentation and code comments

jwrosewell:
## Background

Privacy is a spectrum, not a binary. Trusted Server is technology, and it is neutral on policy. Different deployers operate under different laws and run different policies, and it is the deployer who decides how to configure the stack accordingly. Trusted Server's role is to give deployers the flexibility to make those choices and to respect those choices when made, rather than to limit them in the deployer's name. This change updates documentation and code comments to reflect that framing.

Wording such as "privacy-preserving" identifiers, "built-in GDPR compliance", and "privacy-first" design presents privacy as a property of the technology itself. That overstates what software can promise, understates the role of the deployer, and asks the reader to accept a policy position in order to accept a technical description.

## What changed

- Loaded terms are replaced with factual descriptions of behavior. Compliance capability claims become descriptions of the consent signals read (TCF v2 format, GPP, US Privacy, GPC) and the configuration the publisher controls.
- Consent frameworks are described neutrally. TCF is one framework among others, and "TCF v2 format" is used when the encoded string schema is meant rather than the policy framework.
- Code comments and string literals receive the same treatment, covering the ec module, integration modules, auction types, settings, and TOML comments. Client TLS and H2 values are named for what they are (probabilistic identifier, JA4 string, H2 SETTINGS string). Third-party SDK function names are unchanged.
- Statements that did not match the code are corrected. The architecture page claimed user data is never persisted while EC identity state is persisted in KV stores. The edge cookie page described `x-ts-ec` as an inbound transport when it is a response header. The placeholder secret causes a settings load failure rather than a logged warning. Consent gating is built in per jurisdiction, with configuration tuning jurisdiction lists, GPC interpretation, and conflict resolution.

## Scope

No runtime behavior changes. Code edits are limited to comments, doc comments, and test assertion message strings.

An audit worksheet and a followups plan under `docs/superpowers/plans/` record the findings, the rewrites, and the items intentionally deferred to a separate code cycle (behavior questions and symbol renames that are out of scope for a wording change).

PR #352, opened 2026-02-20
Add CLAUDE.md and consolidate AI agent conventions

aram356:
## Summary

- Establish `CLAUDE.md` as the single source of truth so all AI agents (Claude Code, Codex, Cursor) share identical project conventions
- Add slash commands, subagents, issue templates, and PR template to standardize team workflows
- Rewrite `AGENTS.md` as a thin redirect to eliminate duplication and drift

## Changes

| File | Change |
| ---- | ------ |
| `CLAUDE.md` | Rewritten — project overview, workspace layout, toolchain, build commands, coding conventions, error handling, testing, documentation, tracing, git conventions, integration system, JS pipeline, CI gates, workflow, verification, subagents, slash commands, key files, guardrails |
| `AGENTS.md` | Rewritten — redirects to `CLAUDE.md` with 8-rule fallback summary for Codex |
| `.agents/` | Deleted — all 8 `.mdc` rule files and `mcp.json` removed; content inlined into `CLAUDE.md` |
| `.claude/settings.json` | New — team-shared permission allowlist organized by category (read, cargo, npm, git) |
| `.claude/commands/` | New — 5 slash commands: `/check-ci`, `/test-all`, `/test-crate`, `/review-changes`, `/verify` |
| `.claude/agents/` | New — 7 subagents: `build-validator`, `code-architect`, `code-simplifier`, `verify-app`, `pr-creator`, `issue-creator`, `repo-explorer` |
| `.github/ISSUE_TEMPLATE/` | New — 3 form-based issue templates: Bug Report, Story, Task (blank issues disabled) |
| `.github/pull_request_template.md` | New — PR template with summary, changes table, closes section, test plan, checklist |
| `.github/dependabot.yml` | Normalized YAML quote style |
| `.github/workflows/codeql.yml` | Normalized YAML quote style and indentation |
| `.gitignore` | Updated — track `.claude/settings.json`, `.claude/commands/`, `.claude/agents/` via negation pattern |
| `README.md` | Fixed clippy command to match CI (`--no-deps` → `-- -D warnings`) |
| `docs/guide/onboarding.md` | Updated stale `AGENTS.md` reference to point to `CLAUDE.md` |
| `docs/guide/testing.md` | Fixed clippy command to match CI |

## Closes

Closes #230

## Test plan

- [x] `cargo test --workspace`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo fmt --all -- --check`
[...]

PR #923, opened 2026-07-17
Make AGENTS.md the source of truth with CLAUDE.md as a symlink

### LINKED ISSUE #1091: Guard against CLAUDE.md symlink checking out as a plain text file
## Problem

PR #923 replaces `CLAUDE.md` with a symlink to `AGENTS.md`. When a clone has
`core.symlinks=false` — the Git for Windows default unless Developer Mode is
enabled or the shell is elevated — Git materializes the symlink as a **plain
text file whose entire contents are the target path**.

Reproduced locally against the PR head:

```
$ git config core.symlinks false
$ git checkout <pr-head>
$ ls -la CLAUDE.md
-rw-r--r--  1 user  staff  9 CLAUDE.md
$ cat CLAUDE.md
AGENTS.md
```

The result is a 9-byte `CLAUDE.md` containing the literal string `AGENTS.md`.

## Why this matters

Claude Code loads `CLAUDE.md` from the repository root as its project
instructions. On such a checkout it loads the string `AGENTS.md` and nothing
else, so the agent operates with **zero** project conventions: no
`error-stack` rule, no `expect("should ...")` rule, no target-matched clippy
or test aliases, no WASM constraints.

The failure is silent. Nothing errors, no check fails, and the contributor has
no signal that their agent is unconfigured. The resulting PRs simply violate
conventions for reasons that are hard to trace back to the cause.

## Proposed fix

Add a cheap CI guard that fails when the root `CLAUDE.md` is not a working
symlink. Either check is sufficient:

- Assert the tree entry mode is `120000`:
  `git ls-tree HEAD CLAUDE.md | grep -q '^120000'`
- Or assert the checked-out file resolves and is non-trivial in size
  (a real load would be ~19KB; the broken form is 9 bytes).

Documenting the workaround (as PR #923 does in its description) helps someone
who already suspects the problem, but does not detect it.

## Context

Found while reviewing #923. The PR body documents the Windows behavior and the
manual `core.symlinks=true` workaround, but ships no automated detection.

### LINKED ISSUE #1092: Document that bare cargo test fails at the workspace root
## Problem

The agent guide documents that bare `cargo build` and bare `cargo check` fail
at the workspace root:

```
# Build (per-target aliases — bare `cargo build` fails at the workspace root)
# Check compilation (per-target aliases — bare `cargo check` fails at the workspace root)
```

There is no equivalent warning for bare `cargo test`, even though the same
constraint applies: the workspace spans adapters for `wasm32-wasip1`,
`wasm32-unknown-unknown`, and native targets, so a bare `cargo test` /
`cargo test --workspace` tries to compile the Fastly adapter for the host
target.

## Why it surfaced now

Until PR #923, the warning existed in exactly one place: the "Fallback
Summary" in the old `AGENTS.md` redirect stub, which read:

> Do NOT use bare `cargo test --workspace` — it will attempt to compile the
> Fastly adapter for the host target.

That stub is removed by #923 (correctly — it was a duplicated, drift-prone
copy of the rules). The rest of its content is redundant with the main guide,
but this one sentence has no counterpart there, so the guidance is lost.

Note this is a pre-existing gap in the main guide rather than a defect
introduced by #923 — the removal just makes it visible.

## Proposed fix

Add the bare-`cargo test` caveat alongside the existing aliases in the
"Testing & Quality" section, matching the phrasing already used for
`cargo build` and `cargo check`:

```
# Run all Rust tests — use workspace aliases (bare `cargo test` fails at the
# workspace root: it compiles the Fastly adapter for the host target).
```

## Context

Found while reviewing #923.

billadunblock:
## Summary

- Make `AGENTS.md` the single source of truth for AI coding agent instructions, replacing `CLAUDE.md` with a symlink to it — `AGENTS.md` is the tool-agnostic industry convention (used by Codex, Cursor, and others; see [agents.md](https://agents.md/)), while the symlink keeps Claude Code working by its conventional filename with zero duplication or drift.
- Drop the old `AGENTS.md` redirect stub and its fallback summary, so there is only one copy of the rules.
- Update all in-repo references (PR template, onboarding docs, agent and command definitions) to point at `AGENTS.md`.

Rebased onto current `main` (was 49 commits behind) to resolve the review feedback — see **Review feedback** below.

## Changes

| File | Change |
| ---- | ------ |
| `AGENTS.md` | Now holds the full agent guide (moved from `CLAUDE.md` via `git mv`, so `git log --follow` preserves history); redirect stub removed |
| `CLAUDE.md` | Replaced with a symlink to `AGENTS.md` (mode `120000`) |
| `.github/pull_request_template.md` | Checklist link now points at `AGENTS.md` |
| `docs/guide/onboarding.md` | Docs table entry now references `AGENTS.md` |
| `.claude/agents/code-simplifier.md` | Reference updated to `AGENTS.md` |
| `.claude/agents/pr-reviewer.md` | All 5 references updated to `AGENTS.md` (the file grew from 1 to 5 references in #707, which landed after this branch forked) |
| `.claude/commands/review-changes.md` | Reference updated to `AGENTS.md` |

Historical `docs/superpowers/` plans and specs still say `CLAUDE.md`. That is deliberate — they are dated records of past decisions, and the symlink keeps those references resolving.

## Closes

No linked issue. Three follow-ups filed during review are intentionally **not** closed by this PR:

- #1091 — CI guard for the symlink on `core.symlinks=false` clones
- #1092 — document that bare `cargo test` fails at the workspace root
- #1093 — root Markdown is not covered by any format gate (pre-existing on `main`)

## Test plan

- [x] **Content parity** — `git diff main:CLAUDE.md AGENTS.md` shows only the intentional header change, nothing else.
- [x] **Merge cleanliness** — rebased onto `main`; `git merge-base --is-ancestor main HEAD` confirms a clean fast-forward, so the earlier "distinct types" conflict is gone.
- [x] **Symlink** — tree entry mode is `120000` with a relative target that does not escape the repo root; `head -2 CLAUDE.md` resolves through to the `AGENTS.md` content.
- [x] **History** — `git log --follow AGENTS.md` still reaches the original #352 commit.
- [x] **Docs format** — `cd docs && npm run format` passes. *Scope note:* that script is `prettier --check .` run from `docs/`, so it exercises the `docs/guide/onboarding.md` edit but does **not** inspect root-level `AGENTS.md`. A green `format-docs` check is not evidence that the moved file is well-formed; root Markdown is covered by no gate today (#1093).

No Rust or JS code is touched, so the cargo and vitest gates don't apply to this change.

## Review feedback

| Finding | Status |
| ------- | ------ |
| 🔧 **Content regression: the move reverts #767** | **Fixed.** Rebased onto current `main` and carried #767's neutral-language wording into the `AGENTS.md` Project Overview. Verified by the content-parity check above. |
| 🤔 **"distinct types" conflict can silently re-drop the fix** | **Fixed, both parts.** `CLAUDE.md` keeps the symlink, *and* `main`'s post-fork edit was ported into `AGENTS.md`. The branch is now a fast-forward on `main`, so no conflict resolution happens at merge time. |
| 📝 **"Docs format" does not cover the changed file** | **Corrected** in the test plan above, with the scope caveat spelled out. |
| 📝 **Description wrongly said issues are disabled** | **Corrected** — the Closes section now names the three follow-up issues. |
| 🤔 **Windows symlink failure is silent and undetected** | **No change here** — inherent to the symlink approach this PR chose. Tracked in #1091; the workaround stays documented under Notes. |
| 🌱 **Bare `cargo test` warning lost with the stub** | **No change here**, per review — pre-existing gap in the guide, tracked in #1092. |

## Notes

- **Windows**: Git for Windows defaults to `core.symlinks=false` unless symlink support is available (Windows Developer Mode enabled, or an elevated shell). With symlinks disabled, git checks out `CLAUDE.md` as a plain 9-byte text file containing the path `AGENTS.md`, so an agent reading `CLAUDE.md` gets that string instead of the conventions — silently, with no failing check (see #1091). To get a real symlink: enable Developer Mode (Settings → System → For developers), then clone with `git clone -c core.symlinks=true`, or in an existing clone run `git config core.symlinks true` followed by `git checkout -f CLAUDE.md` to re-materialize it.

## References

- [agents.md](https://agents.md/) — the open AGENTS.md specification: a simple, tool-agnostic Markdown convention for giving coding agents project instructions.
- [openai/agents.md](https://github.com/openai/agents.md) — the spec's source repository, with examples and the list of supporting tools.
- [Claude Code memory docs](https://docs.claude.com/en/docs/claude-code/memory) — Claude Code loads `CLAUDE.md` from the project root, which the symlink satisfies while keeping `AGENTS.md` as the single source.


billadunblock:
@jevansnyc as we expand agent harnesses and enrich agentic workflows, this change can be pretty handy (prioritize AGENT.md wth CLAUDE.md as symlink). Consider for corporate users, they can only use Claude Code Enterprise plan (data not for training) which is not very practical for the token price, multi harnesses would be ideal in lots of cases.

AGENTS.md
```diff
@@ -1,29 +1,458 @@
...
 
-This file exists because Codex looks for `AGENTS.md` by convention. All shared
-rules are maintained in `CLAUDE.md` to avoid duplication and drift. If you
-cannot access `CLAUDE.md`, the critical rules are summarized below as a
-fallback.
+## Project Overview
+
+Rust-based edge computing application targeting **Fastly Compute**. Handles
+privacy-preserving Edge Cookie (EC) ID generation, ad serving with GDPR compliance,
+real-time bidding integration, and publisher-side JavaScript injection.
```
aram356: 🔧 **wrench** — Content regression: this reverts #767 on `main`.

This branch forked 49 commits ago. Commit `69c6309ec` ("Apply neutral language framing across documentation and code comments", #767) rewrote this paragraph on `main`, but the moved copy still carries the pre-#767 wording. Merging as-is silently reintroduces "privacy-preserving" and "GDPR compliance".

I verified this is the **only** content drift between the two files — after this change, `AGENTS.md` matches `origin/main:CLAUDE.md` byte-for-byte except the intentional header.

Note the "distinct types" merge conflict makes this easy to lose again: resolving `CLAUDE.md` in favour of the symlink does not carry `main`'s text edit into `AGENTS.md`. See the body-level note on conflict resolution.

```suggestion
Rust-based edge computing application targeting **Fastly Compute**. Handles
Edge Cookie (EC) ID generation, ad serving with consent signal extraction
and enforcement, real-time bidding integration, and publisher-side
JavaScript injection.
```

aram356 (changes requested):
## Summary

The direction is sound and the mechanics are mostly clean: `git mv` preserved
history, the symlink is well-formed (mode `120000`, relative target, does not
escape the repo root), all five in-repo reference updates are correct, and the
deleted fallback summary is genuinely redundant with the moved content. All 18
CI checks pass.

One blocking issue: the branch forked 49 commits ago and the moved copy is
stale, so merging silently reverts a change that already landed on `main`.

> 1 of the inline comments below carries a one-click GitHub `suggestion` — use
> **Commit suggestion** to apply it as a commit on the PR branch. The remaining
> findings are body-level because they concern the merge, the PR description,
> or repo-wide tooling rather than a single line.

## Blocking

### 🔧 wrench

- **Content regression: the move reverts #767** — see inline at `AGENTS.md:9`

## Non-blocking

### 🤔 thinking

- **"Distinct types" merge conflict can silently re-drop the fix** — see below
- **Windows symlink failure mode is documented but unmitigated** — see below

### 📝 note

- **"Docs format" in the test plan does not cover the changed file** — see below
- **PR description states issues are disabled; they are not** — see below

### 🌱 seedling

- **Bare `cargo test` warning was lost with the fallback summary** — see below

- 🤔 **The merge conflict is a "distinct types" conflict, and naive resolution
  re-drops the blocking fix.** `git merge-tree origin/main <head>` reports:

  ```
  CONFLICT (distinct types): CLAUDE.md had different types on each side;
  renamed one of them so each can be recorded somewhere.
  ```

  Git cannot auto-merge a regular-file-vs-symlink change, so it stages both
  sides rather than combining them. Resolving `CLAUDE.md` in favour of the
  symlink does **not** carry `main`'s text edit into `AGENTS.md` — that edit
  lives in a different file after this PR. The resolution therefore has two
  parts, and skipping the second reintroduces the 🔧 finding even if the
  suggestion is applied now:

  1. `CLAUDE.md` → keep the symlink (this PR's side).
  2. `AGENTS.md` → port `main`'s post-fork edit to the Project Overview.

  Worth rebasing on `main` and re-checking `git diff origin/main:CLAUDE.md`
  against `AGENTS.md` before merge: after the fix the two should differ only in
  the intentional header.

- 🤔 **The Windows failure mode is silent, and nothing detects it.** The PR
  description documents the `core.symlinks=false` behavior and the manual
  workaround, which is good. But I reproduced the failure and the outcome is
  worse than "tools see the pointer text": with symlinks disabled, `CLAUDE.md`
  checks out as a 9-byte file whose entire contents are the string
  `AGENTS.md`. Claude Code loads that as the project instructions, so a
  contributor on such a clone gets **zero** conventions — no `error-stack`
  rule, no `expect("should ...")`, no target-matched clippy/test aliases — with
  no error and no failing check. Documentation only helps someone who already
  suspects the problem. Filed as #1091 for a cheap CI guard; not a merge
  blocker for this PR.

- 📝 **The "Docs format" test-plan item does not exercise this change.**
  `docs/package.json`'s `format` script is `prettier --check .` run from
  `docs/`, so it never inspects root-level `AGENTS.md`. The green `format-docs`
  check is not evidence that the moved file is well-formed. Separately: root
  `AGENTS.md` does fail `prettier --check` — but so does `main`'s `CLAUDE.md`
  (identical table-alignment drift), so this is pre-existing and **not** a
  defect of this PR. Filed as #1093.

- 📝 **The PR description says "issues are disabled on this repository" — they
  are not.** The repository has `has_issues: true` and issues are actively
  filed (#1085–#1089 within the last two days). Minor, but it changes what the
  "Closes" section should say, and follow-ups from this review were filed as
  #1091, #1092, #1093.

- 🌱 **One line of guidance is genuinely lost with the fallback summary.**
  Removing the duplicated summary is the right call, and I checked every rule
  in it against the moved guide — all are covered except one: "Do NOT use bare
  `cargo test --workspace` — it will attempt to compile the Fastly adapter for
  the host target." The main guide warns about bare `cargo build` and bare
  `cargo check` but not bare `cargo test`. Pre-existing gap in the guide rather
  than something this PR breaks; filed as #1092.

## CI Status

- cargo fmt: PASS (required)
- cargo test: PASS (required)
- vitest: PASS
- integration tests: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS
- CodeQL: PASS

All checks pass, which is expected — no code changed. Worth noting that no
gate compares documentation content across branches, so CI cannot catch the
🔧 finding above.

CLAUDE.md
```diff
@@ -1,457 +0,0 @@
-# CLAUDE.md
-
-> Single source of truth for all AI coding agents (Claude Code, Codex, Cursor,
-> etc.). If you're reading `AGENTS.md`, it redirects here.
```
aram356, line 4: 🤔 **thinking** — The Windows failure mode here is silent, and nothing detects it.

The PR description documents the `core.symlinks=false` behavior and the manual workaround, which is good. But the outcome is worse than "tools see the pointer text". I reproduced a symlinks-disabled checkout of this head:

```
$ git config core.symlinks false && git checkout <head>
$ ls -la CLAUDE.md
-rw-r--r--  1 user  staff  9 CLAUDE.md
$ cat CLAUDE.md
AGENTS.md
```

These 457 lines become a 9-byte file whose entire contents are the string `AGENTS.md`. Claude Code loads that as its project instructions, so a contributor on such a clone gets **zero** conventions — no `error-stack` rule, no `expect("should ...")`, no target-matched clippy/test aliases, no WASM constraints — with no error raised and no failing check.

Documentation only helps someone who already suspects the problem. Filed #1091 for a cheap CI guard (assert the tree entry mode is `120000`, or that the checked-out file is non-trivial in size).

Not a merge blocker — it's inherent to the symlink approach this PR deliberately chose, and the tradeoff looks correct. Flagging so the silent-failure characteristic is on record.

CLAUDE.md
```diff
@@ -1,457 +0,0 @@
-# CLAUDE.md
```
aram356, line 1: 🤔 **thinking** — This type change is a "distinct types" merge conflict, and resolving it naively re-drops the blocking fix.

`git merge-tree origin/main <head>` reports:

```
CONFLICT (distinct types): CLAUDE.md had different types on each side;
renamed one of them so each can be recorded somewhere.
```

Git cannot auto-merge a regular file against a symlink, so it stages both sides rather than combining them. The trap: `main` also edited this file's *content* after the fork (commit `69c6309ec`, #767), and resolving `CLAUDE.md` in favour of the symlink does **not** carry that edit anywhere — after this PR the text lives in `AGENTS.md`, a different file.

So the resolution has two parts, and skipping the second silently reintroduces the 🔧 finding on `AGENTS.md` even if that suggestion is committed first:

1. `CLAUDE.md` → keep the symlink (this PR's side).
2. `AGENTS.md` → port `main`'s post-fork edit into the Project Overview.

Suggested check after rebasing, which should print only the intentional header difference:

```bash
git diff origin/main:CLAUDE.md AGENTS.md
```

AGENTS.md
```diff
@@ -1,29 +1,458 @@
...
 
-**Before doing anything else, read `CLAUDE.md` in this repository root.** It
-contains all project conventions, coding standards, build commands, workflow
-rules, and CI requirements. Everything in `CLAUDE.md` applies to you.
+> Single source of truth for all AI coding agents (Claude Code, Codex, Cursor,
+> etc.). `CLAUDE.md` is a symlink to this file, kept for tools that look it up
+> by name.
 
-This file exists because Codex looks for `AGENTS.md` by convention. All shared
-rules are maintained in `CLAUDE.md` to avoid duplication and drift. If you
```
aram356, line 8: 🌱 **seedling** — One line of guidance is genuinely lost with this stub.

Removing the duplicated fallback summary is the right call — it was a drift-prone second copy of the rules, which is exactly what this PR is fixing. I checked every rule in it against the moved guide, and all are covered except one:

> Do NOT use bare `cargo test --workspace` — it will attempt to compile the Fastly adapter for the host target.

The main guide warns about bare `cargo build` and bare `cargo check` ("per-target aliases — bare `cargo build` fails at the workspace root") but has no equivalent for `cargo test`, even though the same multi-target constraint applies.

This is a pre-existing gap in the guide rather than something this PR breaks — the deletion just removes the only place it was written down. Filed as #1092; no change needed here.

billadunblock:
@aram356 Thanks for your review (and precious tokens of course 🤖). I've forced a sync from latest CLAUDE.md and updated the PR. Please assign #1091 to me and I am happy to address in separate PR.

In addition, I'll open separate PR for symlink `.claude/skills|agents|etc.` to `.agents/skills|agents|etc.` but want to monitor first symlink a bit in case we receive complaints.

aram356:
> @aram356 Thanks for your review (and precious tokens of course 🤖). I've forced a sync from latest CLAUDE.md and updated the PR. Please assign #1091 to me and I am happy to address in separate PR.
> 
> In addition, I'll open separate PR for symlink `.claude/skills|agents|etc.` to `.agents/skills|agents|etc.` but want to monitor first symlink a bit in case we receive complaints.


We will fix how to assign tickets to you but in meantime please work on #1091 and #1092

billadunblock:
> > @aram356 Thanks for your review (and precious tokens of course 🤖). I've forced a sync from latest CLAUDE.md and updated the PR. Please assign #1091 to me and I am happy to address in separate PR.
> > In addition, I'll open separate PR for symlink `.claude/skills|agents|etc.` to `.agents/skills|agents|etc.` but want to monitor first symlink a bit in case we receive complaints.
> 
> We will fix how to assign tickets to you but in meantime please work on #1091 and #1092

Addressed both issue in the PR.

aram356 (approved):
## Summary

Second review pass, against `3d98a8f09` (the branch was rebased and force-pushed since the first pass at `62d178b3f`). Every finding from that pass is resolved, and the new CI guard does what it claims. Approving.

## Previous findings — resolved

| Finding | Status |
|---|---|
| 🔧 Content regression reverting #767 | Fixed — `AGENTS.md` now carries main's post-#767 wording |
| 🤔 "Distinct types" merge conflict | Gone — rebased from 49 commits behind to 1; now `MERGEABLE` |
| 🤔 Windows symlink silent failure | Mitigated by the new CI guard (#1091) |
| 📝 Docs-format claim / issues-disabled claim | Corrected in the PR description |
| 🌱 Bare `cargo test` guidance lost | Restored at `AGENTS.md:96` (#1092) |

I verified the content fix directly: the only remaining difference between `origin/main:CLAUDE.md` and this branch's `AGENTS.md` is the intentional header change plus the deliberate `cargo test` addition.

## New surface — the CI guard

I exercised `check-claude-md-symlink` against constructed broken trees rather than reading it:

- Regular file (mode `100644`) containing `AGENTS.md` → correctly rejected
- Symlink pointing at a wrong target → correctly rejected
- `CLAUDE.md` deleted → correctly rejected
- Current head → passes

It triggers on `pull_request` and pushes to `main` with no path filter, so it cannot be skipped, and the error annotations name the remedy. Good guard.

One incidental finding while testing: with `core.symlinks=false`, Git ignores the type difference entirely, so the materialized text file does not stage on its own — I had to construct the broken tree entry with `update-index --cacheinfo` to trigger it. The broken form is therefore harder to commit accidentally than the original concern assumed, which makes this guard cheap insurance rather than a fix for something imminent.

## Non-blocking

- 🤔 **The guard runs but does not block merge.** Branch protection currently requires only `cargo fmt`, `format-typescript`, `format-docs`, and `cargo test`. `CLAUDE.md symlink guard` is not in that list, so a PR that breaks the symlink would show a red check and still be mergeable. Worth adding to the required set — a repository-settings change, not a code change, so nothing to do in this PR.
- 📌 **#1093 remains open** — root Markdown is covered by no format gate, and `AGENTS.md` fails `prettier --check` (as does `main`'s `CLAUDE.md` — identical table-alignment drift). Pre-existing and correctly out of scope here.

## CI Status

All 19 checks PASS, including the newly added guard:

- cargo fmt: PASS (required)
- cargo test: PASS (required)
- vitest: PASS
- integration tests: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS
- CodeQL: PASS

Note the branch is `BEHIND` main; a merge or rebase before landing will re-run these.

ChristianPavilonis (approved):

PR #928, opened 2026-07-17
Add TS admin endpoint to look up EC by id

### LINKED ISSUE #921: Add TS admin endpoint to look up EC by id
### Description

Endpoint: /_ts/admin/ec/<id>

prk-Jr:
## Summary

- Adds `GET /_ts/admin/ec/{id}` (explicit EC ID) and `GET /_ts/admin/ec` (EC ID from the caller's `ts-ec` cookie) so operators can inspect EC identity graph entries without KV console access.
- The EC response includes a derived **auction view**: the exact EIDs the auction would attach for the entry, plus the reason each stored partner ID was skipped (`empty_uid`, `not_in_registry`, `bidstream_disabled`) — making KV-to-auction EID propagation debuggable in one call. Corrupt or legacy-schema entries stay inspectable via a new raw lookup that reports `entry_error` + `raw_body` instead of failing closed.
- Adds `GET /_ts/admin/eids`, the client-side half of the picture: it echoes the request's `ts-eids` and `sharedId` cookies and previews what cookie ingestion would write into the EC entry's `ids` map (matched partner UIDs, deduplicated exactly like the ingestion path, plus dropped sources). Pure request inspection with no KV access, so every adapter serves the real handler; always responds 200.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/ec/admin.rs` | New `handle_admin_ec_lookup` + `handle_admin_eids_lookup` handlers, response payload types, 14 unit tests |
| `crates/trusted-server-core/src/ec/kv.rs` | New `KvIdentityGraph::lookup_raw` returning raw body/metadata/generation without validation |
| `crates/trusted-server-core/src/ec/prebid_eids.rs` | Expose cookie-ingestion collectors `pub(crate)` so the EIDs preview reuses them instead of reimplementing |
| `crates/trusted-server-core/src/ec/mod.rs` | Register `admin` module |
| `crates/trusted-server-core/src/settings.rs` | Add all three routes to `Settings::ADMIN_ENDPOINTS`; update coverage tests |
| `crates/trusted-server-adapter-fastly/src/app.rs` | Route EC + EIDs paths to the real handlers; route-table tests. The EC arm deliberately builds its own identity graph instead of reusing the bot-gated `ec.kv_graph`, so operator curl requests work |
| `crates/trusted-server-adapter-axum/src/app.rs` | EC routes → local 501 (`AdminEcNotSupported`); EIDs route → real handler |
| `crates/trusted-server-adapter-axum/tests/routes.rs` | Route-registration, authenticated-501 (EC) and authenticated-200 (EIDs) tests |
| `crates/trusted-server-adapter-cloudflare/src/app.rs` | EC routes → local 501; EIDs route → real handler |
| `crates/trusted-server-adapter-cloudflare/tests/routes.rs` | Route-registration, authenticated-501 (EC) and authenticated-200 (EIDs) tests |
| `crates/trusted-server-adapter-spin/src/app.rs` | EC routes → local 501; EIDs route → real handler; extend fallback-methods table |
| `crates/trusted-server-adapter-spin/tests/routes.rs` | Authenticated-501 (EC) and authenticated-200 (EIDs) tests |

## Closes

Closes #921

## Test plan

- [x] `cargo test-fastly && cargo test-axum` (also `cargo test-cloudflare && cargo test-spin`)
- [x] `cargo clippy-fastly && cargo clippy-axum` (also `clippy-cloudflare`, `clippy-cloudflare-wasm`, `clippy-spin-native`, `clippy-spin-wasm`)
- [ ] JS tests: `cd crates/trusted-server-js/lib && npx vitest run` (JS untouched — left to CI)
- [ ] JS format: `cd crates/trusted-server-js/lib && npm run format` (JS untouched — left to CI)
- [ ] Docs format: `cd docs && npm run format` (docs untouched — left to CI)
- [x] WASM build: `cargo check -p trusted-server-adapter-fastly --target wasm32-wasip1`
- [x] Manual testing via `fastly compute serve`: verified auth gating, the EC lookup against the seeded Viceroy entry, and the EIDs echo against real browser cookies; this surfaced (and the branch fixes) an initial bug where the bot-gated KV graph made curl requests return 501
- [x] Other: integration parity suite (`cargo test --manifest-path crates/trusted-server-integration-tests/Cargo.toml --test parity`)

## Hardening note

All three paths were added to `Settings::ADMIN_ENDPOINTS`, so startup validation fails closed (a `Configuration` error via `Report`, no `panic!`/`unwrap()`/`expect()`) for any config whose basic-auth handler regex does not cover them; prefix-style regexes like the example config's `^/_ts/admin` remain valid. Covered by the updated `uncovered_admin_endpoints_*` and `admin_endpoints_match_fastly_router` tests. Routes are registered explicitly on every adapter so they never fall through to the publisher fallback, which would forward the caller's `Authorization` header to the origin.

## Checklist

- [x] New code has tests


crates/trusted-server-core/src/settings.rs
```diff
@@ -2197,9 +2197,19 @@ impl Settings {
...
+    ///
+    /// The `/_ts/admin/ec/{id}` entry is the literal router pattern; handler
+    /// path regexes are matched against it verbatim, so prefix-style admin
+    /// regexes (e.g. `^/_ts/admin`) cover it while regexes too narrow to
+    /// cover the parameterized route are rejected fail-closed.
+    pub(crate) const ADMIN_ENDPOINTS: &[&str] = &[
+        "/_ts/admin/keys/rotate",
+        "/_ts/admin/keys/deactivate",
+        "/_ts/admin/ec",
+        "/_ts/admin/ec/{id}",
```
ChristianPavilonis, line 2217: 🔧 **P0 — A route-template match does not secure real IDs**

Startup validation matches the literal `/_ts/admin/ec/{id}`, but `enforce_basic_auth` matches the concrete request URI. A configuration with a handler such as `^/_ts/admin/ec/[{]id[}]$` passes startup validation while a real `/_ts/admin/ec/<valid-id>` matches no handler and is served without Basic Auth. This response deliberately includes raw consent, partner-ID, geo/device, and publisher metadata.

Please fail closed at runtime for recognized admin paths with no matching handler, rather than treating them as public. Also add a regression test for a template-only handler configuration and a concrete valid EC ID request.

crates/trusted-server-adapter-fastly/src/app.rs
```diff
@@ -1053,6 +1071,25 @@ const NAMED_ROUTES: &[NamedRoute] = &[
         primary_methods: &[Method::POST],
         handler: NamedRouteHandler::DeactivateKey,
     },
+    // Admin EC lookup: the bare route reads the EC ID from the caller's
+    // `ts-ec` cookie; the parameterized route takes an explicit EC ID.
+    NamedRoute {
+        path: "/_ts/admin/ec",
+        primary_methods: &[Method::GET],
```
ChristianPavilonis, line 1099: 🔧 **P1 — Non-GET admin requests still reach the publisher fallback**

Declaring only `GET` here causes the generic route loop below to bind every other fallback verb to the publisher proxy. After `^/_ts/admin` Basic Auth succeeds, `POST`/`PUT`/etc. requests to these new routes retain `Authorization` and their bodies when `handle_publisher_request` forwards them to the origin. The same pattern affects `/ _ts/admin/eids` and the Axum, Cloudflare, and Spin adapters.

Route every non-GET supported method, plus malformed/trailing admin-EC paths, to a local 405/deny handler across all adapters and add authenticated wrong-method regression coverage.

crates/trusted-server-adapter-fastly/src/app.rs
```diff
@@ -574,6 +578,18 @@ async fn run_named_route(
...
         NamedRouteHandler::DeactivateKey => handle_deactivate_key(&state.settings, services, req),
+        NamedRouteHandler::AdminEcLookup => {
+            // Deliberately NOT `ec.kv_graph`: that copy is bot-gated (None for
+            // non-browser clients), and operators hit this auth-gated endpoint
+            // with curl. Build the graph directly from settings instead.
+            let kv = crate::maybe_identity_graph(&state.settings);
+            let partner_registry = PartnerRegistry::from_config(&state.settings.ec.partners)?;
+            handle_admin_ec_lookup(kv.as_ref(), &partner_registry, &req)
+        }
+        NamedRouteHandler::AdminEidsLookup => {
```
ChristianPavilonis: 🔧 **P1 — The EIDs diagnostic is not actually read-only on Fastly**

This branch runs only after `execute_named` has built and attached the normal EC finalization state. For a browser request with a valid `ts-ec` plus EID cookies, the entry point finalizer invokes `ingest_eid_cookies` and writes the IDs to KV. That contradicts the endpoint's stated pure/no-KV inspection behavior and differs from the portability adapters.

Dispatch this diagnostic before EC setup/finalization (like the existing early batch-sync branch) so it cannot attach `EcFinalizeState`, then add a Fastly regression test that proves the request performs no KV write.

crates/trusted-server-core/src/ec/admin.rs
```diff
@@ -0,0 +1,925 @@
...
+        tombstone: None,
+        entry: None,
+        entry_error: None,
+        raw_body: None,
+        metadata: None,
+        metadata_error: None,
+        auction: None,
+    };
+
+    match serde_json::from_slice::<KvEntry>(&lookup.body) {
```
ChristianPavilonis: 🔧 **P1 — The claimed raw record is lossy for parseable legacy/forward data**

`KvEntry` accepts unknown fields, but this typed deserialize followed by `serde_json::to_value` removes them. It also normalizes supported legacy representations such as map-shaped `seen_domains`; the same loss occurs for parseable metadata. `raw_body` is only exposed when deserialization fails, so operators cannot inspect the omitted data.

Parse the stored bytes separately as `serde_json::Value` for display (adding only the derived ISO fields), while retaining typed parsing solely for validation and auction derivation. Please cover unknown fields and legacy `seen_domains` in tests.

crates/trusted-server-core/src/ec/admin.rs
```diff
@@ -0,0 +1,925 @@
...
+/// A cookie-derived partner UID that ingestion would store.
+#[derive(Debug, Serialize)]
+struct MatchedPartnerId {
+    /// Partner namespace key in the EC entry's `ids` map.
+    source_domain: String,
+    /// The UID that would be stored.
+    uid: String,
+}
+
+/// Handles `GET /_ts/admin/eids`.
```
ChristianPavilonis, line 563: 🔧 **P2 — Document this operator-facing API**

The API reference's protected-endpoint list still contains only the two key-management routes. It does not document these three endpoints, their payloads/statuses, the Fastly-only EC lookup support, or the portability-adapter 501 responses.

Please update `docs/guide/api-reference.md` with the authentication requirement, request forms, response/error contracts, `no-store` behavior, and adapter limitations.

ChristianPavilonis (changes requested):
## Summary

Request changes. This adds sensitive admin diagnostics, but the current route/auth wiring permits an authentication bypass and can forward admin credentials to the publisher origin. The inline comments also cover Fastly-only write side effects, loss of raw KV data, and the missing operator API contract.

crates/trusted-server-core/src/settings.rs
ChristianPavilonis, line 2217: **P1: Validate a concrete parameterized EC path for auth coverage**

This is a router template, but runtime auth calls `handler_for_path(req.uri().path())` with the concrete request path. Consequently, a configuration with handlers such as `^/_ts/admin/(keys/rotate|keys/deactivate|ec|eids)$` and `^/_ts/admin/ec/\{id\}$` passes startup coverage, even though neither handler matches `/_ts/admin/ec/<actual-id>`. An unauthenticated caller can then reach the Fastly lookup handler and read the raw EC entry. Validate coverage against a representative concrete ID path (and add a regression test), or otherwise make the coverage model distinguish router parameters from literal braces.

ChristianPavilonis (commented):
**Automated Review:**

## Summary

Reviewed the admin EC/EID lookup routes, cross-adapter registration, KV handling, response behavior, and authorization validation. I found one high-severity authorization-coverage gap for the parameterized EC lookup route.

## Findings by severity

### P0 / Blockers

None.

### P1 / High

One inline finding.

### P2 / Medium

None.

### P3 / Low

None.

## CI / Existing Reviews

The displayed CI check runs for head `866bcd4c24be67ca5434fff82727375fbdfc5a45` are successful. Immediately before submission, the PR had no existing reviews or inline review comments. Local tests were not run during this review.

crates/trusted-server-adapter-fastly/src/app.rs
```diff
@@ -525,6 +532,17 @@ async fn execute_named(
         return Ok(run_batch_sync(&state, &services, req));
     }
 
+    // This diagnostic only previews request cookies. Running the normal EC
+    // lifecycle would attach finalization state and could ingest those cookies
+    // into KV after the handler returns, violating the endpoint's read-only
+    // contract.
+    if matches!(handler, NamedRouteHandler::AdminEidsLookup) {
```
ChristianPavilonis: **P1 — Admin EC GET lookups are not read-only.** This early return exempts only `AdminEidsLookup`; `AdminEcLookup` continues through `build_ec_request_state` and response finalization. A browser-shaped authenticated lookup carrying `ts-eids`/`sharedId` can therefore ingest those values into the caller's EC row, and a withdrawal signal can write a tombstone and trigger pull sync. A diagnostic GET should not mutate identity state.

Please dispatch `AdminEcLookup` before EC setup as well (while constructing its KV graph directly), and add a regression test that a cookie-bearing lookup response carries no `EcFinalizeState` and produces no KV write.

crates/trusted-server-core/src/settings.rs
```diff
@@ -2191,15 +2191,45 @@ impl Settings {
...
+    /// continue to report this operator-facing route template.
+    pub(crate) const ADMIN_ENDPOINTS: &[&str] = &[
+        "/_ts/admin/keys/rotate",
+        "/_ts/admin/keys/deactivate",
+        "/_ts/admin/ec",
+        "/_ts/admin/ec/{id}",
+        "/_ts/admin/eids",
+    ];
+
+    const ADMIN_EC_ID_AUTH_PROBE: &str = concat!(
```
ChristianPavilonis: **P2 — A single fixed probe does not prove coverage of the dynamic route.** For example, a handler matching `^/_ts/admin/ec/[a-f0-9]{64}[.][a-z]{6}$` accepts this `.abc123` probe and lets startup succeed, but valid EC IDs whose alphanumeric suffix contains a digit or uppercase character later fail closed as a configuration `500`.

Require coverage of the complete dynamic ID language (or at least validate a character-class corpus including uppercase and numeric suffixes) and add a mixed-case/digit regression case.

docs/guide/api-reference.md
```diff
@@ -580,6 +581,67 @@ curl -X POST https://edge.example.com/_ts/admin/keys/deactivate \
 
 ---
 
+## Admin Diagnostic Endpoints
+
+These endpoints expose sensitive identity and cookie data and require HTTP Basic Authentication. Configure a handler that covers the entire `/_ts/admin` namespace; startup rejects configurations that do not protect every admin route. All responses are JSON with `Cache-Control: no-store`.
```
ChristianPavilonis: **P3 — The documented response contract excludes authentication failures.** The shared Basic-auth middleware returns a plaintext `401 Unauthorized` response without `Cache-Control: no-store`, so it is not JSON/no-store as this sentence says. Please either qualify this statement to successful diagnostic-handler responses or update the auth response to meet the documented contract.

ChristianPavilonis (changes requested):
## Summary

Requesting changes: the Fastly admin EC GET lookup currently runs the mutating EC finalization lifecycle. I also found an incomplete dynamic-route auth-coverage validation and a documentation mismatch for authentication failures.

The currently displayed PR checks are passing. I reviewed existing inline feedback and did not repeat items resolved at the current head.

crates/trusted-server-adapter-fastly/src/app.rs
```diff
@@ -574,6 +592,17 @@ async fn run_named_route(
         }
         NamedRouteHandler::RotateKey => handle_rotate_key(&state.settings, services, req),
         NamedRouteHandler::DeactivateKey => handle_deactivate_key(&state.settings, services, req),
+        NamedRouteHandler::AdminEcLookup => {
```
aram356: 🔧 **wrench** — The EC lookup GET is not read-only: it still runs the full mutating EC lifecycle. `execute_named` early-exits only `AdminEidsLookup`; this arm runs after `build_ec_request_state`, the request-filter pipeline, and gets `EcFinalizeState` attached via `attach_dispatch_extensions`. For a browser-shaped authenticated request, `ec_finalize_response` ingests the caller's `ts-eids`/`sharedId` cookies into KV, can mint a new EC entry and set a `ts-ec` cookie on the admin JSON response, and fires pull-sync after send. The worst case is this endpoint's own core workflow: an operator who copies a user's `ts-ec` cookie into a browser to debug that user's entry ingests their **own** `ts-eids`/`sharedId` values into that user's identity record — the inspection contaminates the record being inspected. Request filters (e.g. DataDome) also still run here and can challenge the exact curl workflow the un-bot-gated KV graph in this arm was built for.

**Fix**: dispatch `AdminEcLookup` in `execute_named` before EC setup, exactly like `AdminEidsLookup` — this handler already builds its own KV graph and never reads `EcRequestState`:

```rust
if matches!(
    handler,
    NamedRouteHandler::AdminEidsLookup | NamedRouteHandler::AdminEcLookup
) {
    let response = PartnerRegistry::from_config(&state.settings.ec.partners)
        .and_then(|registry| match handler {
            NamedRouteHandler::AdminEcLookup => {
                let kv = crate::maybe_identity_graph(&state.settings);
                handle_admin_ec_lookup(kv.as_ref(), &registry, &req)
            }
            _ => handle_admin_eids_lookup(&registry, &req),
        })
        .unwrap_or_else(|error| http_error(&error));
    return Ok(response);
}
```

And add a regression test mirroring `admin_eids_diagnostic_skips_ec_finalization`: a cookie-bearing, browser-shaped EC lookup response must carry no `EcFinalizeState`.

crates/trusted-server-core/src/settings.rs
aram356: 🔧 **wrench** — This single probe under-represents the EC ID language. The suffix `abc123` is lowercase, but real suffixes are mixed-case: `generate_random_suffix` draws from the full alphanumeric alphabet and `is_valid_ec_id` accepts `Ab12z9`. A handler regex like `^/_ts/admin/ec/[a-f0-9]{64}[.][a-z0-9]{6}$` passes startup validation yet fails at runtime for every uppercase-suffix ID.

To be clear about severity: this is not an auth bypass — the new `is_admin_path` fail-closed check in `enforce_basic_auth` turns the miss into a 500, never a credential forward. But it converts the startup lint into intermittent, ID-dependent runtime 500s, which will be miserable to diagnose in production.

**Fix**: probe a small corpus instead of one value and require all to match, e.g. add a second probe with an uppercase-and-digit suffix (`...aaaa.ZZ99Zz`), then add a mixed-case regression test alongside `from_toml_rejects_literal_parameter_template_auth_coverage`.

docs/guide/api-reference.md
aram356: 🔧 **wrench** — "All responses are JSON with `Cache-Control: no-store`" is not true for authentication failures: the shared basic-auth middleware emits the 401 challenge as plaintext without `no-store`. Qualify this sentence to successful diagnostic-handler responses (changing the shared auth response instead would be out of scope for this PR).

crates/trusted-server-core/src/ec/admin.rs
```diff
@@ -0,0 +1,1195 @@
...
+    }
+    let matched = dedupe_partner_updates(updates)
+        .into_iter()
+        .map(|update| MatchedPartnerId {
+            source_domain: update.partner_id,
+            uid: update.uid,
+        })
+        .collect();
+
+    let unmatched = eids
```
aram356: 🤔 **thinking** — The preview silently drops the most debuggable case. A `ts-eids` source that matches a configured partner but has no valid UID (all empty or over the size limit — `first_valid_uid` returns `None` in `collect_prebid_eid_updates`) appears in neither `ingest.matched` nor `ingest.unmatched`. "My partner is configured but ingestion stores nothing" is precisely the situation an operator would bring to this endpoint, and the payload would show nothing at all for that source. Consider reason-tagged drops, e.g. `unmatched: [{"source": ..., "reason": "no_partner" | "no_valid_uid"}]`.

crates/trusted-server-core/src/ec/admin.rs
```diff
@@ -0,0 +1,1195 @@
...
+    };
+
+    let body =
+        serde_json::to_string(&payload).change_context(TrustedServerError::Configuration {
+            message: "failed to serialize admin EIDs response".to_owned(),
+        })?;
+    Ok(json_response(StatusCode::OK, body))
+}
+
+fn extract_cookie_value(req: &Request<EdgeBody>, name: &str) -> Option<String> {
```
aram356: ♻️ **refactor** — This is a byte-identical third copy of `extract_cookie_value` in core (`auction/endpoints.rs` has the same function, and the Fastly adapter carries a fourth). Worth hoisting into `cookies.rs` or `http_util.rs` while this PR touches the area, and having the existing copies call it.

crates/trusted-server-core/src/ec/admin.rs
```diff
@@ -0,0 +1,1195 @@
...
+    }
+    None
+}
+
+fn json_error(status: StatusCode, message: &str) -> Response<EdgeBody> {
+    let body = serde_json::json!({ "error": message });
+    json_response(status, body.to_string())
+}
+
+fn json_response(status: StatusCode, body: String) -> Response<EdgeBody> {
```
aram356, line 664: ⛏ **nitpick** — Consider adding `X-Content-Type-Options: nosniff` here. `raw_body` reflects arbitrary KV bytes to an authenticated browser, and the APS diagnostic endpoints already set nosniff on their responses. One header on this shared builder covers every admin diagnostic response.

crates/trusted-server-core/src/ec/admin.rs
```diff
@@ -0,0 +1,1195 @@
...
+#[derive(Debug, Serialize)]
+struct AdminEcLookupResponse {
+    /// The EC ID that was looked up.
+    ec_id: String,
+    /// Platform KV store name the entry was read from.
+    store: String,
+    /// Store generation marker for the entry.
+    generation: u64,
+    /// `true` when the entry is a consent-withdrawal tombstone
+    /// (`consent.ok = false`). Absent when the body failed to parse.
```
aram356: ⛏ **nitpick** — "Absent when the body failed to parse" is incomplete: the field is also absent when the JSON parses but the `KvEntry` schema deserialization fails (only the `Ok(entry)` branch of `build_lookup_response` sets it). Suggest "Absent when the body failed to parse or deserialize."

aram356 (changes requested):
## Summary

Adds three basic-auth-gated diagnostics (`GET /_ts/admin/ec[/{id}]`, `GET /_ts/admin/eids`) plus admin-namespace hardening (fail-closed startup coverage, runtime fail-closed auth for unmatched admin paths, publisher-fallback denial). The hardening work and cross-adapter test coverage are in good shape, but the Fastly EC lookup still runs the mutating EC finalization lifecycle, the auth-coverage probe under-represents the EC ID language, and the documented response contract is inaccurate for 401s.

## Blocking

### 🔧 wrench

- **EC lookup GET is not read-only**: `AdminEcLookup` still flows through `build_ec_request_state`, request filters, and EC finalization; a browser-shaped authenticated lookup ingests the caller's `ts-eids`/`sharedId` cookies into KV and can contaminate the very record being inspected (crates/trusted-server-adapter-fastly/src/app.rs:595)
- **Auth-coverage probe misses mixed-case EC IDs**: a handler regex matching only lowercase suffixes passes startup but produces ID-dependent runtime 500s (crates/trusted-server-core/src/settings.rs:2221)
- **Docs overstate the response contract**: the 401 challenge is plaintext without `Cache-Control: no-store` (docs/guide/api-reference.md:586)

## Non-blocking

### 🤔 thinking

- **Ingest preview drops partner-matched sources with no valid UID**: they appear in neither `matched` nor `unmatched`, hiding the exact failure mode operators would debug (crates/trusted-server-core/src/ec/admin.rs:489)

### ♻️ refactor

- **Third copy of `extract_cookie_value` in core**: hoist to a shared helper (crates/trusted-server-core/src/ec/admin.rs:516)

### ⛏ nitpick

- **Missing `X-Content-Type-Options: nosniff`** on the shared admin JSON builder (crates/trusted-server-core/src/ec/admin.rs:537)
- **`tombstone` doc comment incomplete**: also absent on schema-deserialization failure (crates/trusted-server-core/src/ec/admin.rs:114)
- **No unauthenticated 401 test for the new routes**: every new-route adapter test sends valid credentials; the existing 401 tests cover only the keys routes. Given this PR's hardening story is auth coverage, one `GET /_ts/admin/ec` → 401 test per adapter would pin it.

### 📝 note

- **Deliberate breaking config change**: configs whose admin handler covers only the keys endpoints (e.g. `^/_ts/admin/keys/`) now fail startup. Correct fail-closed behavior, but worth an explicit upgrade/release note so operators are not surprised at deploy time.

## CI Status

- fmt: PASS
- clippy: PASS
- js tests: PASS
- docs/format checks: PASS

crates/trusted-server-adapter-fastly/src/app.rs
prk-Jr: Fixed in f3776a9a, with the regression assertion tightened in 436e25f2. execute_named now dispatches both admin diagnostics before request filters, EC setup, and finalization, and constructs the EC lookup graph directly. The browser-shaped cookie regression confirms that the handler runs, no EcFinalizeState is attached, and no Set-Cookie mutation is emitted; without finalization state, cookie ingestion, tombstone writes, and pull sync cannot run.

crates/trusted-server-core/src/settings.rs
prk-Jr: Fixed in de61c43f. Dynamic EC auth coverage now probes both a lowercase suffix (.abc123) and a mixed-case/digit suffix (.Ab12Z9), and startup requires every probe to match. The added regression rejects a lowercase-only handler while the valid alphanumeric handler fixture covers both probes.

docs/guide/api-reference.md
prk-Jr: Fixed in be434c29 and refined in 436e25f2. The documentation now limits the JSON and no-store contract to normal diagnostic-handler responses after successful authentication, and explicitly documents the shared plaintext 401 challenge plus unexpected plaintext 5xx responses as outside that contract.

crates/trusted-server-adapter-fastly/src/app.rs
prk-Jr: Fixed in f3776a9a, with the regression assertion tightened in 436e25f2. AdminEcLookup now shares the pre-EC early-dispatch branch with AdminEidsLookup, builds its KV graph directly, and skips request filters and EC finalization. The browser-shaped cookie regression confirms the lookup handler runs, carries no EcFinalizeState, and emits no Set-Cookie header, so post-response ingestion, minting, tombstones, and pull sync cannot occur.

crates/trusted-server-core/src/settings.rs
prk-Jr: Fixed in de61c43f. The startup check now uses a two-value corpus with .abc123 and .Ab12Z9 and requires all probes to be covered. A regression test verifies that a lowercase-only suffix regex is rejected instead of becoming an ID-dependent runtime 500.

docs/guide/api-reference.md
prk-Jr: Fixed in be434c29 and refined in 436e25f2. The API reference now scopes JSON plus Cache-Control: no-store to normal authenticated diagnostic-handler responses and separately calls out the shared plaintext 401 challenge and unexpected plaintext 5xx path.

crates/trusted-server-core/src/ec/admin.rs
prk-Jr: Implemented in 8649b43e. ingest.unmatched is now reason-tagged with no_partner and no_valid_uid. The diagnostic and production ingestion paths share the same Prebid EID analysis, and regressions cover missing partners, empty or oversized UIDs, and duplicate sources where a valid UID suppresses the invalid drop.

crates/trusted-server-core/src/ec/admin.rs
prk-Jr: Implemented in 1198eb97. The duplicated core implementations were replaced by the generic trusted_server_core::cookies::extract_cookie_value helper, and both the auction endpoint and admin diagnostic call it. Helper-level tests preserve trimming, embedded-equals, missing-header, and non-UTF-8 behavior. The Fastly-local helper remains separate because it operates on fastly::Request rather than http::Request<B>.

crates/trusted-server-core/src/ec/admin.rs
prk-Jr: Added in a3a04795. The shared admin JSON response builder now sets X-Content-Type-Options: nosniff, covering successful and JSON error responses from both diagnostics. The response-header test pins the behavior.

crates/trusted-server-core/src/ec/admin.rs
prk-Jr: Fixed in a3a04795. The tombstone field documentation now says it is absent when the body fails either JSON parsing or typed KvEntry deserialization, matching the branches in build_lookup_response.

ChristianPavilonis (commented):
**Automated Review:**

## Review Summary

Reviewed the admin EC/EID diagnostic endpoints, their authorization and reserved-namespace fallback handling, Fastly KV lookup path, cross-adapter routing, and the related cookie/EID parsing changes. I found no high-confidence correctness, security, or compatibility defects in the reviewed change.

## Findings

No findings.

## CI / Existing Reviews

GitHub CI is green, including core and adapter tests, cross-adapter parity, Fastly EC lifecycle integration coverage, Cloudflare and Spin checks, formatting, CodeQL, and Vitest. I also reviewed the current-head inline feedback; the identified admin authorization and publisher-fallback concerns are addressed by the runtime/startup coverage checks and local fallback denials.

crates/trusted-server-core/src/ec/admin.rs
```diff
@@ -0,0 +1,1288 @@
...
+    }
+
+    if path.starts_with("/_ts/admin/eids/") {
+        return Some(AdminDiagnosticShape::Malformed);
+    }
+
+    // Reserve the complete admin namespace at the publisher-fallback boundary.
+    // A successfully authenticated malformed or future admin path must never
+    // forward its Authorization header or body to the publisher origin.
+    (path == "/_ts/admin" || path.starts_with("/_ts/admin/"))
```
ChristianPavilonis: **Percent-encoded admin separators leak Basic credentials**: `/_ts/admin%2Fec` matches a normal `^/_ts/admin` Basic-auth handler but misses this literal-slash namespace check. It consequently reaches publisher fallback, which forwards its original `Authorization` header and body. This affects all adapters.

**Suggestion**: Treat case-insensitive encoded separators immediately after `/_ts/admin` as reserved and deny them before fallback. Add cross-adapter regressions for `%2F` and `%2f`.

crates/trusted-server-core/src/settings.rs
```diff
@@ -2191,15 +2191,52 @@ impl Settings {
...
+    /// continue to report this operator-facing route template.
+    pub(crate) const ADMIN_ENDPOINTS: &[&str] = &[
+        "/_ts/admin/keys/rotate",
+        "/_ts/admin/keys/deactivate",
+        "/_ts/admin/ec",
+        "/_ts/admin/ec/{id}",
+        "/_ts/admin/eids",
+    ];
+
+    const ADMIN_EC_ID_AUTH_PROBES: [&str; 2] = [
```
ChristianPavilonis, line 2235: **Sampled probes do not establish EC route coverage**: A handler can match only these two IDs, so startup accepts a configuration even though other valid six-character alphanumeric suffixes are uncovered. The same gap lets a weak-password handler for an unprobed valid ID evade placeholder-password validation.

**Suggestion**: Do not infer dynamic-route coverage from samples. Require an explicit handler binding or a prefix-level matcher for `/_ts/admin/ec/`, and validate the selected handler's credentials.

ChristianPavilonis (dismissed):
## Summary

Approved with two inline high-severity findings.

### Finding folded into the review body

**Medium - Retired alias descendants still fall through**: Only exact `/admin/keys/rotate` and `/admin/keys/deactivate` are denied. Trailing-slash and descendant forms reach publisher fallback with any supplied `Authorization` header and body; Axum, Cloudflare, and Spin have the same gap. Deny the complete retired `/admin/keys` namespace, including trailing, descendant, and encoded-separator forms, at the shared fallback boundary.

CI is green; local `cargo test-fastly` and `git diff --check` also passed.

crates/trusted-server-core/src/ec/admin.rs
ChristianPavilonis: **P0 - Percent-encoded admin separators leak Basic credentials**: `/_ts/admin%2Fec` matches a normal `^/_ts/admin` Basic-auth handler but misses this literal-slash namespace check. It reaches publisher fallback, which forwards its original `Authorization` header and body. This affects all adapters and must block merge.

**Required fix**: Treat case-insensitive encoded separators immediately after `/_ts/admin` as reserved and deny them before fallback. Add cross-adapter regressions for `%2F` and `%2f`.

ChristianPavilonis (dismissed):
## Blocking finding

This review supersedes my prior approval. The percent-encoded admin-path credential leak is a P0: an authenticated request such as `/_ts/admin%2Fec` passes the normal Basic-auth matcher but bypasses the fallback reservation, forwarding the admin `Authorization` header and request body to the publisher origin. This must be fixed before merge.

crates/trusted-server-core/src/ec/admin.rs
ChristianPavilonis: **P0 - Percent-encoded admin separators leak Basic credentials**: `/_ts/admin%2Fec` matches a normal `^/_ts/admin` Basic-auth handler but misses this literal-slash namespace check. It reaches publisher fallback, which forwards its original `Authorization` header and body. This affects all adapters and must block merge.

**Required fix**: Treat case-insensitive encoded separators immediately after `/_ts/admin` as reserved and deny them before fallback. Add cross-adapter regressions for `%2F` and `%2f`.

crates/trusted-server-core/src/settings.rs
ChristianPavilonis, line 2235: **P1 - Sampled probes do not establish EC route coverage**: A handler can match only these two IDs, so startup accepts a configuration even though other valid six-character alphanumeric suffixes are uncovered. The same gap lets a weak-password handler for an unprobed valid ID evade placeholder-password validation.

**Required fix**: Do not infer dynamic-route coverage from samples. Require an explicit handler binding or a prefix-level matcher for `/_ts/admin/ec/`, and validate the selected handler's credentials.

ChristianPavilonis (changes requested):
## Review Summary

Requesting changes. The percent-encoded admin-path credential leak is a P0 and must be resolved before merge. One additional P1 is inline.

### P2 / Medium

**Retired alias descendants still fall through**: Only exact `/admin/keys/rotate` and `/admin/keys/deactivate` are denied. Trailing-slash and descendant forms reach publisher fallback with any supplied `Authorization` header and body; Axum, Cloudflare, and Spin have the same gap. Deny the complete retired `/admin/keys` namespace, including trailing, descendant, and encoded-separator forms, at the shared fallback boundary.

This finding is in the body because GitHub could not resolve the affected Fastly diff line for an inline comment.

## CI

GitHub CI is green; local `cargo test-fastly` and `git diff --check` also passed.

crates/trusted-server-core/src/ec/admin.rs
prk-Jr: Fixed in caf44220.

The reservation now spans the whole `/_ts/admin` prefix, mirroring the `^/_ts/admin` auth-handler regex rather than requiring a literal `/` after `admin`, and it is evaluated on the percent-decoded path as well. `/_ts/admin%2Fec`, `%2f`, double-encoded `%252F`, `%5C`, and `/%5Fts/admin/ec` are all answered locally with 404 before publisher fallback, so no `Authorization` header or body reaches the origin.

Regressions: `admin_diagnostic_fallback_reserves_encoded_admin_separators` in core, plus the encoded paths added to `authenticated_admin_diagnostic_fallback_is_denied_locally` in all four adapters. `auth::tests::encoded_admin_separator_path_is_auth_gated` pins the premise that `^/_ts/admin` matches the encoded path and consumes the credentials.

The retired-alias finding from the review body is in the same commit: `/admin/keys` is reserved by prefix on raw and decoded paths, so `/admin/keys`, `/admin/keys/rotate/`, `/admin/keys/rotate/extra`, `/admin/keys%2Frotate`, and `/admin%2fkeys/rotate` are denied at the shared boundary. `/admin/login` and `/admin/keyboards` still fall through.

crates/trusted-server-core/src/settings.rs
prk-Jr: Fixed in 8684e1b0.

Coverage is no longer inferred from ID samples. `ADMIN_EC_ID_AUTH_PROBES` is now the bare prefix `/_ts/admin/ec/` plus one concrete ID, and both must match the same configuration: the prefix probe rejects handlers anchored to an ID grammar, the concrete probe rejects handlers anchored to the prefix itself (`^/_ts/admin/ec/$`), so only prefix-level matchers such as `^/_ts/admin` or `^/_ts/admin/ec/` pass.

Note this rejects even a grammar-exact regex like `^/_ts/admin/ec/[a-f0-9]{64}[.][A-Za-z0-9]{6}$`. That is deliberate: the router accepts any single segment after `/_ts/admin/ec/` and basic auth runs on the raw path before routing, so malformed IDs still reach the admin handler (400) and must be authenticated. Documented as breaking in CHANGELOG.md and docs/guide/configuration.md.

On credentials: the placeholder and weak-password check now runs on every handler instead of on handlers inferred to cover an admin endpoint. Handler selection is first-match-wins, so a narrow handler ahead of the admin pattern governs the IDs it matches and no probe set can enumerate them; scoping the check by inferred coverage would reproduce the same sampling gap. Also breaking for a config whose non-admin handler uses `changeme`/`password`/`admin`.

Tests: `from_toml_rejects_sampled_id_only_dynamic_admin_ec_auth_coverage`, `from_toml_rejects_prefix_anchored_admin_ec_auth_coverage`, `from_toml_accepts_prefix_matcher_admin_ec_auth_coverage`, `from_toml_rejects_placeholder_password_on_shadowing_admin_handler`, `from_toml_rejects_weak_password_on_non_admin_handler`.

crates/trusted-server-core/src/ec/admin.rs
```diff
@@ -0,0 +1,1400 @@
...
+
+    reserved.then_some(AdminDiagnosticShape::Malformed)
+}
+
+/// Returns whether `path` sits in a namespace that must never reach publisher
+/// fallback, because doing so would forward Trusted Server admin credentials
+/// and request bodies to the publisher origin.
+fn is_reserved_admin_path(path: &str) -> bool {
+    path.starts_with(ADMIN_NAMESPACE_PREFIX) || path.starts_with(RETIRED_ADMIN_KEYS_PREFIX)
+}
```
aram356, line 151: ♻️ **refactor** — The retired `/admin/keys` reservation over-matches unrelated publisher paths. The bare `starts_with(RETIRED_ADMIN_KEYS_PREFIX)` also denies `/admin/keystore`, `/admin/keystone-...`, and any other publisher path that happens to start with `/admin/keys` — the test above even pins `/admin/keyboards` as falling through, which differs from `/admin/keystore` only by its fourth letter. The bare prefix is required for `/_ts/admin` (it mirrors the unanchored `^/_ts/admin` auth regex, which consumes credentials for paths like `/_ts/adminec`), but no auth handler ever matches `/admin/keys*`, so only the exact retired alias and its separator descendants represent the stale-script risk. Encoded forms stay covered by the percent-decoded evaluation: every path in `admin_diagnostic_fallback_reserves_retired_admin_keys_namespace` and in the four adapter deny tests is still denied with the separator-aware form below.

```suggestion
fn is_reserved_admin_path(path: &str) -> bool {
    path.starts_with(ADMIN_NAMESPACE_PREFIX)
        || path == RETIRED_ADMIN_KEYS_PREFIX
        || path
            .strip_prefix(RETIRED_ADMIN_KEYS_PREFIX)
            .is_some_and(|remainder| remainder.starts_with('/'))
}
```

Verified in a scratch worktree with this change applied: `cargo fmt --all -- --check`, all six clippy aliases, the full `cargo test-fastly` / `test-axum` / `test-cloudflare` / `test-spin` suites, and the cross-adapter parity suite all pass. Consider also adding `"/admin/keystore"` to `admin_diagnostic_fallback_ignores_unrelated_publisher_paths` to pin the boundary.

aram356 (approved):
## Summary

All findings from the prior review rounds are resolved and verified in code: the EC lookup is dispatched before EC setup and is genuinely read-only (regression tests assert no `EcFinalizeState` and no `Set-Cookie`), the percent-encoded admin-separator leak is closed by reserving the bare `/_ts/admin` prefix on both the raw and percent-decoded path (a credential leak requires the `^/_ts/admin` auth regex to match the raw path, and every such path is caught by the raw prefix check; decode failures safely fall back to the raw check), dynamic-route auth coverage is no longer inferred from ID samples, and the breaking configuration changes are documented in CHANGELOG.md and the configuration guide. Approving with one non-blocking suggestion and two notes.

> 1 of the inline comments below carries a one-click GitHub `suggestion` — use **Commit suggestion** to apply it as a commit on the PR branch.

## Non-blocking

### ♻️ refactor

- **Retired `/admin/keys` reservation over-matches unrelated publisher paths** — see inline at `crates/trusted-server-core/src/ec/admin.rs:98`

- 🏕 **Leftover `extract_cookie_value` copy in the Fastly adapter** — `crates/trusted-server-adapter-fastly/src/main.rs:456` still carries its own copy (outside this diff); core's new generic `cookies::extract_cookie_value<B>` can replace it in a follow-up.
- 📝 **Residual coverage-inference gap is closed by the runtime guard** — the probe pair requires each probe to match *some* handler, not the *same* handler, so a config pairing `^/_ts/admin/ec/$` with a grammar-anchored regex still passes startup while leaving malformed IDs to the runtime fail-closed 500 in `enforce_basic_auth`. That runtime guard is the actual guarantee (no credential forward is possible), so this is informational only — full regex-containment checking would not be practical.

## CI Status

- integration tests: PASS
- CodeQL: PASS
- cargo test: PASS (required)
- cargo fmt: PASS (required)
- Analyze (actions): PASS
- Analyze (rust): PASS
- vitest: PASS

crates/trusted-server-core/src/ec/admin.rs
```diff
@@ -0,0 +1,1410 @@
...
+
+/// Percent-decodes `path` once, returning `None` when the path contains no
+/// escape sequence or decodes to invalid UTF-8.
+///
+/// Routers and the basic-auth matcher both operate on the raw path, so an
+/// encoded separator can shift a request out of the literal admin namespace
+/// while still matching the admin auth handler. Checking the decoded form as
+/// well keeps the reservation closed for `%2F`, `%2f`, and their
+/// double-encoded variants.
+fn percent_decoded_path(path: &str) -> Option<String> {
```
ChristianPavilonis, line 160: **Double-encoded retired admin paths bypass fallback reservation**

`POST /admin%252Fkeys/rotate` is neither recognized as the retired `/admin/keys` namespace nor reserved after one decode. It reaches publisher fallback in every adapter.

The fallback then forwards the original `Authorization` header and body to the publisher origin. A proxy or origin that decodes the path again can receive `/admin/keys/rotate`.

Please apply bounded iterative percent-decoding to a fixed point before checking reserved namespaces, and add cross-adapter regressions for double-encoded separators.

crates/trusted-server-core/src/ec/admin.rs
```diff
@@ -0,0 +1,1410 @@
...
+fn requested_ec_id(req: &Request<EdgeBody>) -> Result<String, Box<Response<EdgeBody>>> {
+    let remainder = req
+        .uri()
+        .path()
+        .strip_prefix(ADMIN_EC_PATH)
+        .unwrap_or("")
+        .trim_matches('/');
+
+    let ec_id = if remainder.is_empty() {
+        match extract_cookie_value(req, COOKIE_TS_EC) {
```
ChristianPavilonis: **Bare EC lookup chooses a different duplicate cookie than normal EC processing**

`extract_cookie_value` selects the first `ts-ec` cookie, while `EcContext` parses cookies into `CookieJar`, whose existing test confirms the last duplicate wins. With duplicate cookies, `GET /_ts/admin/ec` can inspect a different EC record than the normal request lifecycle uses.

Please resolve the bare route's cookie through the same `CookieJar` path as `EcContext`, and add a duplicate-cookie regression test.

ChristianPavilonis (approved):
## Review summary

Reviewed PR #928 at head `6fb4238` against `main`. GitHub CI is green, local `cargo test-fastly` passed, and `git diff --check` is clean.

I am approving this change with two follow-up findings noted inline. They do not block this approval, but both are worth addressing before relying on the new fallback hardening and cookie-based lookup in production.

## Existing reviews

Earlier review findings around auth coverage, read-only lookup behavior, raw KV fidelity, and single-encoded fallback paths appear addressed at the current head.

crates/trusted-server-core/src/ec/admin.rs
ChristianPavilonis, line 160: **Double-encoded retired admin paths bypass fallback reservation**

`POST /admin%252Fkeys/rotate` is neither recognized as the retired `/admin/keys` namespace nor reserved after one decode. It reaches publisher fallback in every adapter.

The fallback then forwards the original `Authorization` header and body to the publisher origin. A proxy or origin that decodes the path again can receive `/admin/keys/rotate`.

Please apply bounded iterative percent-decoding to a fixed point before checking reserved namespaces, and add cross-adapter regressions for double-encoded separators.

crates/trusted-server-core/src/ec/admin.rs
ChristianPavilonis: **Bare EC lookup chooses a different duplicate cookie than normal EC processing**

`extract_cookie_value` selects the first `ts-ec` cookie, while `EcContext` parses cookies into `CookieJar`, whose existing test confirms the last duplicate wins. With duplicate cookies, `GET /_ts/admin/ec` can inspect a different EC record than the normal request lifecycle uses.

Please resolve the bare route's cookie through the same `CookieJar` path as `EcContext`, and add a duplicate-cookie regression test.

crates/trusted-server-core/src/ec/admin.rs
prk-Jr: Closed. `is_reserved_admin_path` now matches `/_ts/admin` as a bare prefix, so every path the unanchored `^/_ts/admin` auth regex authenticates is also reserved at the publisher-fallback boundary — encoded separator or not. Commit 464acd8e additionally decodes the check to a bounded fixed point, so multi-encoded forms such as `/admin%252Fkeys/rotate` cannot survive to fallback either.

Covered by `admin_diagnostic_fallback_reserves_encoded_admin_separators` and `admin_diagnostic_fallback_reserves_multi_encoded_separators` in core, plus the `authenticated_admin_diagnostic_fallback_is_denied_locally` route tests on the axum, Cloudflare, and Spin adapters.

crates/trusted-server-core/src/settings.rs
prk-Jr: Closed. Coverage is no longer inferred from ID-shaped samples. `ADMIN_EC_ID_AUTH_PROBES` requires two probes to match: the bare prefix `/_ts/admin/ec/` and a concrete ID. The bare prefix rejects handlers anchored to the EC ID grammar; the concrete ID rejects handlers anchored to the prefix itself (`^/_ts/admin/ec/$`). Together they admit only prefix-level matchers such as `^/_ts/admin` or `^/_ts/admin/ec/`.

Covered by `from_toml_rejects_sampled_id_only_dynamic_admin_ec_auth_coverage`. The residual case — a config pairing `^/_ts/admin/ec/$` with a grammar-anchored regex, where each probe matches a different handler — is caught at runtime by the fail-closed 500 in `enforce_basic_auth`, so no credential forward is possible.

crates/trusted-server-core/src/ec/admin.rs
prk-Jr: Closed — same fix as the earlier thread on this file. `is_reserved_admin_path` matches `/_ts/admin` as a bare prefix, so every path the unanchored `^/_ts/admin` auth regex authenticates is reserved before publisher fallback can forward its `Authorization` header or body. Commit 464acd8e extends the check to decode to a bounded fixed point for multi-encoded separators.

Regressions live in core (`admin_diagnostic_fallback_reserves_encoded_admin_separators`, `admin_diagnostic_fallback_reserves_multi_encoded_separators`) and in the axum, Cloudflare, and Spin route suites.

Issue #834, opened 2026-07-01
Origin cache-header audit: per-content-type cacheability diagnostics for publisher origins


aram356:
Trusted Server serves many content types through one edge hostname; HTML, JS bundles, creatives/images, static assets, RTB/JSON — each with a different optimal caching posture, all driven by the cache directives the origin returns (Cache-Control, Surrogate-Control, Surrogate-Key, Vary, ETag). Today TS mostly passes these through untouched and exposes no cache diagnostics, so a publisher developer can't tell whether each content type is cached correctly. The result is silent over-caching (personalized/consent-sensitive HTML served stale or shared across users) and under-caching (immutable assets re-fetched from origin every request).

This epic delivers ts dev audit headers, a CLI command that audits origin response cache directives grouped by content type and returns a per-type pass/warn/fail verdict naming the responsible header and recommending the correct value; so misconfigured caching is caught before it hurts hit ratio or leaks personalized content.

Out of scope: auto-fixing headers, request-side cache-key/hit-ratio analysis, and the ts dev proxy tool.

Note: `ts dev audit headers` slots into the same ts dev CLI group

PR #860, opened 2026-07-07
Implement configurable cache header policies

ChristianPavilonis:
## Summary

- Standardizes cache policy rendering across Fastly, Cloudflare, CDN, and `s-maxage` fallback headers.
- Adds safe, configurable caching for hash-validated TSJS, publisher-origin static assets, and rehosted asset proxy responses.
- Hardens privacy handling so private, no-store, and cookie-bearing responses strip shared edge-cache headers.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/cache_policy.rs` | Adds typed cache-policy rendering for browser and edge headers, including no-store/private cleanup. |
| `crates/trusted-server-core/src/settings.rs` | Adds `cache.asset_rules` config, matchers/presets, validation, runtime prep, and path-to-policy resolution. |
| `trusted-server.example.toml` | Documents disabled operator-controlled static/fingerprinted asset cache-rule examples. |
| `crates/trusted-server-core/src/http_util.rs` | Routes static ETag responses through the cache-policy renderer. |
| `crates/trusted-server-core/src/tsjs.rs` | Uses exact module-set hashes when available and avoids unverifiable fallback hashes. |
| `crates/trusted-server-js/Cargo.toml` | Makes hashing dependencies available to the build script. |
| `crates/trusted-server-js/build.rs` | Generates per-module SHA-256 metadata for bundled JS modules. |
| `crates/trusted-server-js/src/bundle.rs` | Exposes per-module hashes and caches concatenated bundle hashes. |
| `crates/trusted-server-core/src/publisher.rs` | Applies hash-validated immutable TSJS caching and configured publisher asset cache rules. |
| `crates/trusted-server-core/src/proxy.rs` | Applies normalized cache policies to rehosted asset proxy responses and reapplies them after finalization. |
| `crates/trusted-server-core/src/response_privacy.rs` | Removes shared-cache headers from private/no-store/cookie-bearing responses. |
| `crates/trusted-server-core/src/integrations/prebid.rs` | Makes the neutralized Prebid shim `no-store, private` instead of long-lived public cache. |
| `crates/trusted-server-core/src/integrations/testlight.rs` | Updates the default TSJS fallback comment/source behavior for registry-free configuration. |
| `crates/trusted-server-core/src/lib.rs` | Exports the new `cache_policy` module. |
| `crates/trusted-server-adapter-axum/src/app.rs` | Passes the portable `s-maxage` fallback edge header mode to TSJS and publisher handlers. |
[...]

PR #823, opened 2026-06-25
Add ts CLI ad-template config diagnostics and browser audit

prk-Jr:
## Summary

- Adds operator tooling to verify server-side ad-template (`[creative_opportunities]`) configuration: static path/slot diagnostics via `ts config ad-templates …`, and browser-backed live verification via `ts audit …` (local Chrome/Chromium over CDP).
- Adds `ts audit ad-templates generate <url>` to **bootstrap** `[creative_opportunities]` from a live site. One run **crawls the publisher's sections** (sitemap via `robots.txt`, else navigation links), samples a landing page and an article per section, reconciles each slot across the pages it appeared on, and writes the result into an existing `trusted-server.toml` **in place**, preserving every other section and comment.
- Where the evidence proves it, the run **infers a `{network_id}`/`{section}` ad-unit template** plus the `section_root`/`section_segment` policy it depends on, instead of pinning each slot to the one literal path it happened to be scraped from. A wrong template makes a publisher bid against inventory that does not exist, so inference refuses rather than guesses — see the table below.
- Everything the run writes is **validated before it replaces the file**, on the `--dry-run` path too. The candidate is deserialized as `TrustedServerAppConfig` and checked through `TrustedServerAppConfig::new`, the push-time validation path. This checks slot/template compilation, provider configuration, and secret-store key references without I/O. Resolved secret values are checked separately at runtime. Parse/schema failures report safe location or context without source excerpts or invalid values. Deploy-validation failures report a generic phase-level error because their detailed messages can contain literal values. This applies to both candidate refusal and pre-existing-baseline warnings.
- Lets publishers catch slot/config drift before deploy — match a path to configured slots, assert expected slots in CI, explain runtime ad-stack gating, and confirm configured slots actually render (DOM/GPT/APS evidence) on live pages.
- Keeps the CLI host-only: browser deps (`chromiumoxide`) are excluded from the `wasm32-wasip1` build, and the runtime ad-stack gate is shared with `publisher.rs` so the CLI cannot drift from server behavior.

> Stacks on the EdgeZero CLI base (`server-side-ad-templates-impl`). After merging the base back in, the diff against it is feature-only.

> **Behavior change**: bare `ts audit <url>` remains a hidden alias for `ts audit generate <url>`; the base branch's draft-artifact generation moved to the explicit `ts audit generate <url>`.

[...]

PR #930, opened 2026-07-17
Add cache-header audit design spec for ts dev audit headers

### LINKED ISSUE #835: Create spec and plan to audit publsher origin headers
### Description

1. Use superpowers skill to create a spec and plan

Example spec & plan:

Spec: https://github.com/IABTechLab/trusted-server/blob/worktree-review-ts-dev-proxy-spec/docs/superpowers/specs/2026-06-22-ts-dev-proxy-design.md
Plan: https://github.com/IABTechLab/trusted-server/blob/worktree-review-ts-dev-proxy-spec/docs/superpowers/plans/2026-06-22-ts-dev-proxy.md

2. Breakdown the work into tasks in the parent Epic after completing spec and plan.

vasujain00:
Technical spec for epic #834: origin cache-header audit with per-content-type cacheability diagnostics. Covers content-type taxonomy, cacheability rules for all five directives (Cache-Control, Surrogate-Control, Surrogate-Key, Vary, ETag), per-type verdict rollup, CLI integration into the ts dev group, and a 7-task implementation breakdown.

Closes #835

## Summary

-

## Changes

| File | Change |
| ---- | ------ |
|      |        |

## Closes

Closes #

## Test plan

- [ ] `cargo fmt --all -- --check`
- [ ] Other: 

## Checklist

- [ ] Changes follow [CLAUDE.md](/CLAUDE.md) conventions
- [ ] No `unwrap()` in production code — use `expect("should ...")`
- [ ] New code has tests
- [ ] No secrets or credentials committed


docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
...
+
+## 2. Command Surface
+
+```
+ts dev audit headers [OPTIONS] [URLS...]
+```
+
+### Arguments
+
+| Argument | Description |
```
aram356: 🔧 **wrench** — This file is what's failing the `format-docs` CI check: `npx prettier --check` rejects every markdown table in the doc (unpadded cells).

**Fix**: `cd docs && npm run format:write` and commit the result.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
...
+
+**Acceptance criteria:**
+- Guide covers all user scenarios with copy-pasteable examples
+- `ts dev audit headers --help` is clear and useful
+
+---
+
+## 11. Dependency Graph
+
+```
```
aram356: 🔧 **wrench** — The dependency graph should be a mermaid diagram, not ASCII box art. The docs site uses vitepress with `vitepress-plugin-mermaid`, and existing specs (including the ts-dev-proxy spec referenced by #835 as the template) use mermaid; this block will render as a dead code block.

**Fix**:
````markdown
```mermaid
flowchart TD
    T1["Task 1: CLI restructure"] --> T6["Task 6: E2E wiring"]
    T2["Task 2: Rules engine"] --> T3["Task 3: Origin fetch"] --> T4["Task 4: Analysis"] --> T5["Task 5: Output"] --> T6
    T2 --> T4
    T6 --> T7["Task 7: Docs"]
```
````

File: docs/superpowers/specs/2026-07-14-cache-header-audit-design.md [hunk 3]
```diff
@@ -0,0 +1,426 @@
...
+}
+```
+
+---
+
+## 8. CLI Integration
+
+### 8.1 `ts dev` restructure
+
+`ts dev` is currently a flat command wrapping `fastly compute serve`. It must be restructured into a subcommand tree:
```
aram356: 🔧 **wrench** — Stale premise: `ts dev` is not "a flat command wrapping `fastly compute serve`". On main, `DevCommand` is already a `clap::Subcommand` enum with a macOS-only `Proxy` variant (`crates/trusted-server-cli/src/commands/dev/mod.rs`), and local serving lives in the separate top-level `ts serve` (edgezero). There is no `DevArgs` and no `Serve` variant to default to — implementing this section as written would invent a `Serve` behavior `ts dev` never had.

**Fix**: rewrite Section 8.1 (and Task 1, line 289) against current main: add an `Audit(DevAuditCommand)` variant to the existing `DevCommand` enum and drop the `Serve` default / backward-compat requirement. Task 1's "modifies existing code on `feature/ts-cli`" is also stale — the CLI landed on main.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
...
+enum DevAuditCommand {
+    Headers(AuditHeadersArgs),
+}
+```
+
+`ts dev` (bare, no subcommand) defaults to `Serve` for backward compatibility.
+
+### 8.2 Error handling
+
+Add `CliError::HeaderAudit` variant to `error.rs`.
```
aram356: 🔧 **wrench** — There is no `CliError` enum to add a variant to. `crates/trusted-server-cli/src/error.rs` defines `CliResult<T> = Result<T, String>` plus helper functions.

**Fix**: either follow the existing String-error pattern, or explicitly propose introducing a typed error enum for the CLI — which is a much larger change than this one-liner implies and should be called out as such.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
...
+- Unit tests cover all classification + evaluation logic
+
+---
+
+### Task 3: Origin request engine
+
+**Type:** Net-new (`dev_audit/fetch.rs`)
+**Dependencies:** Task 2 (uses `ContentTypeGroup`)
+
+Fetch responses from origin using `reqwest` (already a dep). Support explicit URL list and config-derived discovery.
```
aram356: 🔧 **wrench** — `reqwest` is not "already a dep" of `trusted-server-cli`; it's defined at the workspace level but not consumed by the CLI crate. More importantly, the workspace's default build target is `wasm32-wasip1`, which is exactly why the dev-proxy's tokio/rustls/ring stack is cfg-scoped to macOS in the CLI's `Cargo.toml`. This spec's own design decision says explicit URLs exist "for CI" (i.e. Linux) — but today `DevCommand` has zero variants off macOS.

**Fix**: state where the HTTP-client dependencies get scoped (presumably `cfg(not(target_arch = "wasm32"))`) and that `ts dev audit`, unlike `ts dev proxy`, must be available on all host platforms.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
...
+---
+
+## 5. URL Discovery
+
+When no explicit URLs are provided:
+
+1. Read `publisher.origin_url` from `trusted-server.toml`
+2. Probe well-known paths per content type:
+   - HTML: `/`
+   - JS: discover from `<script>` tags on `/`, or use `/_ts/` internal paths
```
aram356: 🔧 **wrench** — These discovery probes target edge-only routes against the origin. `/_ts/api/v1/identify` and `/_ts/` JS paths are served by the trusted server edge, not by `publisher.origin_url` — probing the publisher origin for them will 404. The epic scopes this tool as an *origin* audit ("audits origin response cache directives").

**Fix**: drop the `/_ts` probes, or explicitly define a two-surface audit (origin vs. edge) with separate expectations. This also resolves Open Question 1 (line 424): the epic already says origin-direct.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
+# Technical Specification: Origin Cache-Header Audit (`ts dev audit headers`)
+
+**Status:** Draft
+**Author:** @vasujai
+**Epic:** [#834](https://github.com/IABTechLab/trusted-server/issues/834)
+**Planning task:** [#835](https://github.com/IABTechLab/trusted-server/issues/835)
+**Related:** [#293](https://github.com/IABTechLab/trusted-server/issues/834) (cache header refactoring), [#428](https://github.com/IABTechLab/trusted-server/issues/428) (ETag multi-value), PR #860
```
aram356: 🔧 **wrench** — Wrong link: the text says #293 but the URL points at `/issues/834`.

**Fix**: `[#293](https://github.com/IABTechLab/trusted-server/issues/293)`.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
...
+
+## 4. Cacheability Rules
+
+Each content-type group has an expected caching posture. Rules are evaluated against: `Cache-Control`, `Surrogate-Control`, `Surrogate-Key`, `Vary`, `ETag`.
+
+### 4.1 HTML (personalized/consent-sensitive)
+
+| Header | Expected | Verdict if wrong |
+|--------|----------|-----------------|
+| `Cache-Control` | Contains `private` AND (`no-store` OR `no-cache`) | FAIL: "HTML served without private/no-store risks sharing personalized content across users" |
```
aram356: 🔧 **wrench** — This rule (and the RTB rule at line 106) FAILs the strongest correct configuration. Requiring `private` AND (`no-store` OR `no-cache`) means an origin sending exactly `Cache-Control: no-store` — which per RFC 9111 already forbids storage in both shared and private caches — fails the audit.

**Fix**: pass on `no-store` alone, OR `private` combined with `no-cache`/`no-store`. Apply the same change to Section 4.5.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
...
+|------|-------------|---------|
+| `--config <path>` | Path to `trusted-server.toml` | `./trusted-server.toml` |
+| `--origin <url>` | Override origin URL (skips config lookup) | from config `publisher.origin_url` |
+| `--json` | Machine-readable JSON output | human table |
+
+### Exit codes
+
+| Code | Meaning |
+|------|---------|
+| 0 | All groups pass |
```
aram356: 🤔 **thinking** — Exit code 2 collides with clap's usage-error exit code: `ts dev audit headers --bad-flag` also exits 2, so a CI gate can't distinguish "warnings only" from CLI misuse. Consider warn-only → 0 with a `--strict` flag promoting warn to fail (which also answers Open Question 2, line 425), or a code ≥ 3.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
...
+| 1 | At least one group has a FAIL verdict |
+| 2 | Warnings only (no fails) |
```
aram356: 🤔 **thinking** — No output/exit contract is defined for operational failure: origin unreachable, timeouts, or all probes 404. Task 3 says "handle gracefully", but the report shape and exit code for a failed audit run are unspecified. Worth defining a distinct exit code so CI can tell "audit ran and failed" from "audit couldn't run".

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
...
+### 4.5 RTB/JSON (real-time, never cached)
+
+| Header | Expected | Verdict if wrong |
+|--------|----------|-----------------|
+| `Cache-Control` | Contains `private` AND `no-store` | FAIL: "RTB responses cached = stale bids served to users" |
+| `Surrogate-Control` | If present, `no-store` | FAIL: "CDN caching RTB responses" |
+
+### 4.6 ETag handling (ref: #428)
+
+When evaluating ETag presence, handle multi-value `If-None-Match` correctly:
```
aram356, line 119: 🤔 **thinking** — This conflates request-side `If-None-Match` with response auditing. Issue #428 is about the *server comparing* the `If-None-Match` request header; this tool only inspects origin *response* headers, so multi-value `If-None-Match` never enters the picture.

**Fix**: reframe as "accept both strong and weak ETag response formats" and drop the `If-None-Match` clause here and from Task 2's acceptance criteria (line 314).

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
...
+
+Responses are classified into groups by their `Content-Type` response header:
+
+| Group | Matching Content-Types |
+|-------|----------------------|
+| `Html` | `text/html` |
+| `JavaScript` | `application/javascript`, `text/javascript`, `application/x-javascript` |
+| `Image` | `image/*` |
+| `StaticAsset` | `text/css`, `font/*`, `application/font-*`, `application/woff*` |
+| `RtbJson` | `application/json` |
```
aram356: 🤔 **thinking** — Classifying all `application/json` as `RtbJson` over-flags: any origin JSON (search endpoints, manifests, feeds) would get "RTB responses cached = stale bids served to users" as a FAIL. Consider renaming the group `Json` with posture-neutral messaging, or requiring a path hint before applying RTB semantics.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
...
+
+Each content-type group has an expected caching posture. Rules are evaluated against: `Cache-Control`, `Surrogate-Control`, `Surrogate-Key`, `Vary`, `ETag`.
+
+### 4.1 HTML (personalized/consent-sensitive)
+
+| Header | Expected | Verdict if wrong |
+|--------|----------|-----------------|
+| `Cache-Control` | Contains `private` AND (`no-store` OR `no-cache`) | FAIL: "HTML served without private/no-store risks sharing personalized content across users" |
+| `Vary` | Should NOT contain `*` | WARN: "Vary: * disables all caching including CDN" |
+| `Surrogate-Control` | If present, `no-store` or `private` | FAIL: "CDN may cache personalized HTML" |
```
aram356: 🤔 **thinking** — `Surrogate-Control: private` is not a Fastly-honored directive. Fastly's documented Surrogate-Control directives are `max-age`, `no-store`, `stale-while-revalidate`, and `stale-if-error`; recommending `private` gives publishers a directive their CDN ignores. Suggest checking for `no-store` only (verify against the Fastly reference before finalizing the rule).

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
...
+  fetch.rs        -- AuditHeadersArgs, origin fetching, classification
+  analyze.rs      -- AuditReport, GroupReport, HeaderVerdict, run_analysis()
+  output.rs       -- human table + JSON rendering
+```
+
+### 7.2 Key types
+
+```rust
+pub enum ContentTypeGroup { Html, JavaScript, Image, StaticAsset, RtbJson, Other }
+pub enum Verdict { Pass, Warn(String), Fail(String) }
```
aram356: ♻️ **refactor** — `Verdict::Warn(String)` / `Fail(String)` embed a message that duplicates `HeaderVerdict.recommendation`, and the JSON example (line 172) shows `"verdict": "Fail"` as a bare string — which payload-carrying variants can't serialize to without a custom serde impl.

**Fix**: make `Verdict` a unit enum (`Pass, Warn, Fail`) and keep messages on `HeaderVerdict`.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
...
+| `Image` | `image/*` |
+| `StaticAsset` | `text/css`, `font/*`, `application/font-*`, `application/woff*` |
+| `RtbJson` | `application/json` |
+| `Other` | Everything else (not evaluated, reported as info) |
+
+---
+
+## 4. Cacheability Rules
+
+Each content-type group has an expected caching posture. Rules are evaluated against: `Cache-Control`, `Surrogate-Control`, `Surrogate-Key`, `Vary`, `ETag`.
```
aram356: 🌱 **seedling** — `s-maxage` is ignored throughout. It's the shared-cache TTL Fastly honors between `Surrogate-Control` and `max-age`, so the CDN-vs-browser comparison in 4.3 (and the RTB/HTML shared-cache rules) should account for it.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
...
+|------|---------|
+| 0 | All groups pass |
+| 1 | At least one group has a FAIL verdict |
+| 2 | Warnings only (no fails) |
+
+---
+
+## 3. Content-Type Taxonomy
+
+Responses are classified into groups by their `Content-Type` response header:
```
aram356: 🌱 **seedling** — Two gaps worth noting: (1) `Vary` is listed among the five audited headers but only evaluated for HTML — `Vary: User-Agent` or `Vary: Cookie` on JS/CSS/images are classic hit-ratio killers; (2) classification must strip Content-Type parameters (`text/html; charset=utf-8`) before matching, which the spec doesn't state.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
...
+
+| Group | Matching Content-Types |
+|-------|----------------------|
+| `Html` | `text/html` |
+| `JavaScript` | `application/javascript`, `text/javascript`, `application/x-javascript` |
+| `Image` | `image/*` |
+| `StaticAsset` | `text/css`, `font/*`, `application/font-*`, `application/woff*` |
```
aram356: ⛏ **nitpick** — `application/woff*` isn't a real MIME pattern: fonts are `font/woff2` (covered by `font/*`) or legacy `application/font-woff` (covered by `application/font-*`). The entry is redundant.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,426 @@
+# Technical Specification: Origin Cache-Header Audit (`ts dev audit headers`)
+
+**Status:** Draft
+**Author:** @vasujai
```
aram356: ⛏ **nitpick** — `@vasujai` doesn't match the author's GitHub login (`vasujain00`).

aram356 (changes requested):
## Summary

The spec is a solid starting shape for `ts dev audit headers` (clear taxonomy, verdict rollup, task breakdown), but several load-bearing claims about the codebase are stale or incorrect — `ts dev` is already a subcommand tree, there is no `CliError` enum, `reqwest` is not a CLI dependency, and the discovery section probes edge-only `/_ts/` routes against the publisher origin. Two cacheability rules also fail the strictest correct configuration, and the doc itself is what's breaking the `format-docs` CI check.

Details are in the inline comments (8 blocking 🔧, 5 🤔, 1 ♻️, 2 🌱, 2 ⛏). Cross-cutting items below.

## Blocking

### 🔧 wrench

- **format-docs CI failure**: prettier rejects this file's tables — `cd docs && npm run format:write` (line 30)
- **Dependency graph must be mermaid**: docs use vitepress-plugin-mermaid; ASCII art won't render (line 391)
- **Stale `ts dev` premise**: `DevCommand` subcommand tree already exists on main; no `Serve` to default to (lines 254, 289)
- **`CliError` doesn't exist**: CLI errors are `CliResult<T> = Result<T, String>` (line 271)
- **`reqwest` is not a CLI dep + target scoping unaddressed**: workspace default target is `wasm32-wasip1`; spec must say where deps are scoped and that audit is cross-platform, unlike the macOS-only proxy (line 323)
- **Discovery probes edge routes against origin**: `/_ts/*` paths 404 on `publisher.origin_url` (line 131)
- **Wrong issue link**: #293 text links to /issues/834 (line 7)
- **HTML/RTB rules fail `no-store` alone**: RFC 9111's strongest posture would FAIL the audit (lines 75, 106)

## Non-blocking

### 📌 out of scope

- **Plan document and epic breakdown still pending**: issue #835 asks for a spec **and** a plan under `docs/superpowers/plans/` plus a task breakdown in epic #834; this PR delivers the spec only (Section 10 is a partial plan). Also, the PR body retains the unfilled template (empty Summary/Changes tables, dangling `Closes #`).

## CI Status

- format-docs: FAIL (caused by this file)
- All other checks (fmt, clippy targets, Rust/JS tests, integration, parity, CodeQL): PASS

ChristianPavilonis (changes requested):
I found blocking specification inaccuracies that need correction before this can guide implementation:

- Section 8.1 describes `ts dev` as a flat `fastly compute serve` wrapper and replaces it with only `Serve` and `Audit`. The checked-out CLI already has `DevCommand` as a subcommand enum whose macOS `Proxy` variant backs the documented `ts dev proxy`; the proposed enum drops that supported command and invents a bare-`ts dev` serve compatibility behavior that does not exist. Rebase the design and Task 1 on the current command tree, retaining the cfg-gated proxy and explicitly defining audit availability on supported non-Wasm hosts.
- The origin-discovery design directs requests for `/_ts/` and `/_ts/api/v1/identify` to `publisher.origin_url`. Those are Trusted Server edge routes, not publisher-origin routes, so they will generally 404 and make the origin audit report false failures. Remove these probes or define a separate edge-surface audit with distinct endpoint and policy semantics.
- The HTML and RTB rules fail a response with `Cache-Control: no-store` unless it also contains `private`. `no-store` alone already prohibits storage by both private and shared caches, so the audit would flag the strongest valid no-store response as unsafe. Accept `no-store` independently (and apply the same correction to both groups).

There are also factual/documentation corrections to make, including the #293 link pointing to #834, the request-side `If-None-Match` requirement in a response-header audit, and the claim that `reqwest` is already a CLI dependency.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
vasujain00, line 1: Thanks @aram356 -- thorough review. All 8 blocking items and non-blocking feedback addressed in e268913. Rebased on latest main.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
vasujain00: Fixed

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
vasujain00: Fixed: [#293] now links to /issues/293.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
vasujain00: Fixed. Ran `cd docs && npx prettier --write` -- tables are now padded. CI should pass.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
vasujain00: Replaced ASCII art with mermaid flowchart using the same vitepress-plugin-mermaid syntax as the ts-dev-proxy spec.

aram356:
@vasujain00 Thank you for addressing Please implement it per spec. You can use this pr for the implementation as well.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+
+### 4.1 HTML (personalized/consent-sensitive)
+
+| Header              | Expected                                                        | Verdict if wrong                                                                            |
+| ------------------- | --------------------------------------------------------------- | ------------------------------------------------------------------------------------------- |
+| `Cache-Control`     | Contains `no-store`, OR (`private` AND `no-cache`)              | FAIL: "HTML served without no-store or private+no-cache risks sharing personalized content" |
+| `Vary`              | If present, should NOT contain `*`                              | WARN: "Vary: \* disables all caching including CDN"                                         |
+| `Vary`              | Should NOT contain `User-Agent` or `Cookie` (hit-ratio killers) | WARN: "Vary on User-Agent/Cookie destroys CDN hit ratio for HTML"                           |
+| `Surrogate-Control` | If present, contains `no-store` or `private`                    | FAIL: "CDN may cache personalized HTML"                                                     |
```
aram356: 🔧 **wrench** — The HTML rule fails correct configurations and, with the template cache on, tells publishers to turn it off.

What happens to origin HTML on main:

- When the ad stack runs, TS fetches the origin with the Fastly cache bypassed (`publisher.rs:4407`, `:4708`) and rewrites the response to `no-store, private`, stripping validators and edge headers (`response_privacy.rs:96`, called at `publisher.rs:4874`). On those pages the leak this row guards against is already closed at the edge.
- With `creative_opportunities.assembly_mode = "esi"` (opt-in, Fastly-only, documented as an experimental spike for #1009, and merged the day after this spec was last updated), TS keeps origin HTML in a shared template cache. It refuses HTML whose `Cache-Control` or `Surrogate-Control` contains `private`, `no-store` or `no-cache` (`publisher.rs:6029`, `:5963`). Every response that passes this row is refused there, and the §6.1 advice (`no-store` or `private, no-cache`) is exactly the setting that turns it off.
- The rule also fails `private, max-age=0`, which TS itself emits for cookie-bearing responses (`response_privacy.rs:138`), and Rails' default `private, max-age=0, must-revalidate`. `private` alone already forbids shared storage (RFC 9111 §5.2.2.7), and TS's own privacy check is `private` or `no-store` (`cache_policy.rs:338`).

The `Vary` rows conflict with the rest of the spec. They grade CDN hit ratio for a group this table requires to be uncacheable and §4.7 calls non-cacheable, while Task 2 (line 345) calls HTML cacheable. They also tell publishers to remove `Vary: User-Agent` and `Vary: Cookie`, which protect dynamic serving and cookie-dependent pages from cross-serving. The template cache refuses `Vary: Cookie` and `Vary: *` on purpose (`publisher.rs:6163`, `:6166`).

§1 (line 14) still says TS "mostly passes origin cache directives through untouched". That sentence came from the epic, which predates TS's HTML privacy enforcement and #860.

Suggested direction (apply manually, since §1, §4.1, §4.7, §6 and Task 2 all change):

- Key the privacy check on `private` or `no-store`, as a WARN that names the real risk. Shared-cacheable HTML goes through Fastly's read-through cache whenever the ad stack doesn't run, and bots, prefetches and consent-denied requests get the origin's headers. A header-only audit can't tell whether a page varies by cookie, login or consent state, so a FAIL for the whole content type is too strong.
- When `assembly_mode = "esi"`, report template-cache eligibility: shareable freshness, no `private`/`no-store`/`no-cache`, no `Set-Cookie`, no `Vary: *` or `Cookie`, and every `Vary` name listed in `template_cache_vary` (`publisher.rs:6118`-`6175`). Label these as necessary conditions only, since eligibility also depends on the request (`Authorization`, cookies).
- Drop the HTML `Vary` rows or make them informational, and never recommend removing `Vary`.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+| `RtbJson`     | `application/json`                                                      |
+| `Other`       | Everything else (not evaluated, reported as info)                       |
+
+---
+
+## 4. Cacheability Rules
+
+Each content-type group has an expected caching posture. Rules are evaluated against: `Cache-Control`, `Surrogate-Control`, `s-maxage`, `Surrogate-Key`, `Vary`, `ETag`.
+
+Note on `s-maxage`: Fastly honors `Surrogate-Control` first, then `s-maxage` (the shared-cache TTL from RFC 9111), then `max-age`. Rules below that reference CDN TTL behavior account for all three.
```
aram356: 🔧 **wrench** — The rules grade `Cache-Control` as if it were the edge policy, and on TS's runtimes it often isn't.

- Fastly Compute reads `Surrogate-Control` first: "Compute targets `Surrogate-Control` before `Cache-Control` and honors the first targeted header that parses successfully. This may cause values of `Surrogate-Control` to take priority over more strict values of `Cache-Control` on a shared-cache response." ([Fastly, HTTP caching semantics](https://www.fastly.com/documentation/guides/concepts/edge-state/cache/cache-freshness/))
- `CDN-Cache-Control`, which TS's Cloudflare adapter emits (`cache_policy.rs:24`-`29`), makes a cache "ignore the Cache-Control and Expires header fields" (RFC 9213 §2.2). The spec never reads it.
- Fastly subtracts `Age` from `max-age` and falls back to `Expires`. The spec reads neither.

Concrete misses:

- JS with `public, max-age=31536000, immutable` plus `Surrogate-Control: max-age=60` passes, but the edge keeps it for 60 seconds.
- Browser `no-cache` plus a long `Surrogate-Control`, the usual way to cache at the edge while browsers revalidate, gets flagged in every group.
- `Surrogate-Control: max-age=0` keeps a response out of Fastly's cache, yet it fails the RTB row at line 113, while `Surrogate-Control: private, max-age=86400` passes the HTML row at line 80. Fastly documents only `max-age`, `stale-while-revalidate` and `stale-if-error` for `Surrogate-Control` (the repo says the same at `publisher.rs:5927`). This was a 🤔 last round.
- The checks look for tokens only. `public, max-age=31536000, immutable, no-store` passes the JS row even though nothing gets stored.

Suggested fix: compute two policies per URL and grade them separately. The shared-cache policy follows the runtime's precedence: on Fastly `Surrogate-Control`, then `s-maxage`, then `max-age`, then `Expires`, minus `Age`; the targeted field wherever one is present. The browser policy is `Cache-Control` alone. Run privacy rules on the shared policy and TTL rules on both. `origin_shared_ttl_at` (`publisher.rs:6012`) already implements the Fastly precedence, and `cache_control_value_has_directive` (`cache_policy.rs:291`) is a quoted-string-aware directive parser the CLI can reuse. If targeted fields are out of scope for v1, report them as "not evaluated" rather than PASS.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+| `Vary`              | If present, should NOT contain `*`                              | WARN: "Vary: \* disables all caching including CDN"                                         |
+| `Vary`              | Should NOT contain `User-Agent` or `Cookie` (hit-ratio killers) | WARN: "Vary on User-Agent/Cookie destroys CDN hit ratio for HTML"                           |
+| `Surrogate-Control` | If present, contains `no-store` or `private`                    | FAIL: "CDN may cache personalized HTML"                                                     |
+
+### 4.2 JavaScript bundles (hashed filenames, immutable)
+
+| Header          | Expected                                                  | Verdict if wrong                                                                               |
+| --------------- | --------------------------------------------------------- | ---------------------------------------------------------------------------------------------- |
+| `Cache-Control` | Contains `public` AND `max-age>=31536000` AND `immutable` | WARN: "Hashed JS bundles should use long-lived immutable caching"                              |
```
aram356: 🔧 **wrench** — This row and the static-asset row at line 104 ask for `public, max-age=31536000, immutable` on every script, stylesheet and font that discovery finds, fingerprinted or not. Discovery collects every `<script src>` and stylesheet on the page, including stable URLs like `/js/app.js` or `/wp-includes/js/jquery/jquery.min.js?ver=3.7.1`. Following the advice on those pins stale code in browsers for a year, and no purge reaches a browser cache.

The repo has already settled this:

- The #860 cache-control design keeps "non-versioned or config-dependent stable URLs out of long-lived immutable browser caches" and leaves arbitrary publisher-origin CSS/JS/images origin-controlled (`docs/superpowers/specs/2026-07-06-cache-control-header-design.md:17`, `:75`).
- Config validation refuses `immutable` without a `fingerprint_style` or content-addressed preset (`settings.rs:2243`).
- For paths an enabled `cache.asset_rules` rule matches, TS replaces the origin's headers (`publisher.rs:1570`), so grading the origin value there reports something browsers never see.

Requiring `public` also warns on a correct `max-age=31536000, immutable`. RFC 9111 §5.2.2.9 calls `public` unnecessary on a response that is already cacheable, and Fastly's docs say the same. The image row at line 95 has this problem too.

Suggested rule:

- The path matches an enabled `cache.asset_rules` rule: report "policy set by TS rule `<id>`" and skip the origin check. `Settings::asset_cache_policy_for_path` (`settings.rs:3164`) is public, so the CLI can ask TS directly.
- The URL carries a recognizable fingerprint: expect a long `max-age` plus `immutable`.
- Anything else: expect a validator and a moderate TTL, and WARN on `immutable`, since that's the harmful direction.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+| `--origin <url>`  | Override origin URL (skips config lookup) | from config `publisher.origin_url` |
+| `--json`          | Machine-readable JSON output              | human table                        |
+
+### Exit codes
+
+| Code | Meaning                               |
+| ---- | ------------------------------------- |
+| 0    | All groups pass                       |
+| 1    | At least one group has a FAIL verdict |
+| 2    | Warnings only (no fails)              |
```
aram356: 🔧 **wrench** — Exit code 2 already means "the tool failed" in this CLI. `main.rs:18` exits 2 on any `Err`, which covers every `cli_error()` this spec plans for unreachable origins, and clap exits 2 on a bad flag. `docs/guide/cli.md:136`-`139` documents the codes for the CI-oriented commands: 0 passed, 1 ran and found drift (audit verification uses `--strict` for this), 2 argument, configuration or tool failure. With this table a CI gate can't tell "warnings only" from "origin unreachable".

This was a 🤔 last round. That documentation landed with #823 on 2026-09-21, after the last round, so it's a direct conflict now. `ts audit ad-templates verify` has the matching shape (`--strict`, `--json`, `RunOutcome`). Adopting it also answers Open Question 2 (line 456), which can go, and Task 5's criterion at line 403 needs the same change.

`dev::run` returns `Result<(), String>` (`commands/dev/mod.rs:31`) and `run.rs:162` maps success to exit 0, so Task 1 also needs `dev::run` to return `Result<RunOutcome, String>` the way `run_audit` does. Otherwise exit 1 takes an ad-hoc `process::exit`.

```suggestion
| `--json`          | Machine-readable JSON output              | human table                        |
| `--strict`        | Treat WARN verdicts as failures (exit 1)  | off                                |

### Exit codes

Same exit codes as the other `ts` assertion commands (see `docs/guide/cli.md` and `RunOutcome` in `crates/trusted-server-cli/src/run.rs`):

| Code | Meaning                                                                        |
| ---- | ------------------------------------------------------------------------------ |
| 0    | Audit ran; no FAIL verdicts (WARN verdicts allowed unless `--strict`)          |
| 1    | Audit ran and found a FAIL verdict, or a WARN verdict with `--strict`          |
| 2    | Tool failure: bad arguments, config load error, or origin could not be fetched |
```

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+1. Read `publisher.origin_url` from `trusted-server.toml`
+2. Fetch the root HTML page from origin (`/`)
+3. Parse the HTML response to discover asset URLs per content type:
+   - HTML: the `/` response itself
+   - JS: extract `<script src="...">` URLs from the HTML response
+   - Images: extract `<img src="...">` and `<link rel="icon">` URLs; fall back to `/favicon.ico`
+   - Static: extract `<link rel="stylesheet" href="...">` URLs from the HTML response
+   - RTB/JSON: not discoverable from HTML; requires explicit `--url` or presence of `/_ts/api/v1/identify` in config routes (see note below)
+4. Resolve relative URLs against `publisher.origin_url`
+5. Fetch each discovered URL directly from origin and classify by response `Content-Type`
```
aram356: 🔧 **wrench** — The audit's requests don't match what TS sends to origin, so it can grade responses TS never sees.

- Only relative URLs resolve cleanly against `publisher.origin_url`. Absolute URLs on the public `publisher.domain` go to the TS edge, so the audit would record TS-rewritten headers and label them origin. Many CMSs, WordPress among them, write absolute public-host URLs everywhere. Absolute origin-host URLs are fine: TS rewrites those to the public host in HTML (`host_rewrite.rs`), and they can be fetched directly.
- Third-party hosts (GPT, Prebid, analytics, tag managers) would be graded as if the publisher controlled them.
- `[[proxy.asset_routes]]` sends matching path prefixes to a different backend, so `origin_url` is the wrong host for those paths.
- When `publisher.origin_host_header_override` is set, TS sends it as the Host header (`settings.rs:47`, `publisher.rs:4144`). A direct fetch without it can reach a different vhost, and `--origin` "skips config lookup" (line 39), so it drops the override too.
- The request itself is unspecified. reqwest sends no `User-Agent` and, with the workspace features, no `Accept-Encoding`, while TS forwards the browser's headers. WAF and bot-protection challenge pages would get graded, and `Vary`/`ETag` results can differ by encoding.
- Relative URLs should resolve against the final document URL and any `<base href>`, not against `origin_url`.

Suggested fix: spell out the mapping in §5. Map same-site absolute URLs onto the backend TS would use (`origin_url` or the matching `asset_routes` entry), send the Host override, report third-party URLs as skipped instead of grading them, and define a browser-like default request profile with a repeatable `--header`. Open Question 1 matters here too. Many origins accept only CDN traffic (IP allowlists, a shared-secret header, or S3 SigV4 through `proxy.asset_routes.auth`), so an origin-direct audit from a laptop or CI runner gets a 403 there. It's worth answering before implementation, for example with a mode that audits through a TS staging host.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+
+**Parallel tracks:** Tasks 1 and 2 can start immediately and independently.
+
+---
+
+## 12. Open Questions
+
+1. Should `ts dev audit headers` require a running local server, or always hit the remote origin directly?
+2. Should we add a `--strict` flag that treats warnings as failures (for CI gates)?
+3. Should the URL discovery crawl follow redirects from the origin?
```
aram356: 🔧 **wrench** — This should be settled as "no", and the answer changes the fetcher. No TS adapter follows origin redirects. Fastly's backend API has no follow mode and "an original 3xx remains visible to core" (`trusted-server-adapter-fastly/src/platform.rs:464`-`467`), the Axum adapter uses `redirect::Policy::none()` (`platform.rs:296`), and Cloudflare uses `RequestRedirect::Manual` (`platform.rs:278`). reqwest follows up to 10 redirects by default, across hosts. So when the origin redirects `/` (http to https, apex to www, a locale or consent redirect), the audit grades the target instead of what TS receives. If `Location` points at the public host, it grades TS itself.

The 3xx deserves a verdict of its own. Fastly caches 301 and 302 responses by default, so a geo or consent redirect without `private` gets replayed to other visitors.

Suggested fix: fetch with `redirect::Policy::none()`, report each 3xx as its own row (status, `Location`, cacheability), warn when a 301 or 302 is shared-cacheable, and never follow to another host.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+}
+
+pub struct GroupReport {
+    pub content_type: ContentTypeGroup,
+    pub urls_sampled: Vec<String>,
+    /// Group-level (per-type) verdict: worst-of rollup over `verdicts`.
+    /// Any Fail -> Fail; else any Warn -> Warn; else Pass.
+    pub verdict: Verdict,
+    pub verdicts: Vec<HeaderVerdict>,
+}
```
aram356: 🔧 **wrench** — The report can't tell a publisher which URL to fix, and it has no model for status codes or fetch errors.

- `HeaderVerdict` has one `actual` value and no URL. On a page with twelve scripts where two send `no-cache`, the JavaScript group gets a single `Cache-Control` verdict that can show only one observed value.
- Task 3 routes fetch failures through `cli_error()` (line 368), which ends the whole run with exit 2 and no report. Grading whatever comes back instead is worse: a 404 or bot-challenge page is usually `text/html`, so it lands in the HTML group, and the `/favicon.ico` fallback's 404 can decide the HTML verdict.
- Classification looks only at the response `Content-Type`. A bundle served as `text/plain` or `application/octet-stream`, or with no `Content-Type`, drops to `Other` and the JavaScript group disappears. The discovery context is a better classifier (a `<script src>` URL should be JavaScript), with a WARN on mismatch.
- `Other` has no verdict, so under the rollup at lines 239-241 an unevaluated group counts as a pass. Groups with no sampled URL, like StaticAsset in §6.1, vanish without a trace.
- `recommendation` is the corrective value in §6.1 and the problem statement in §6.2, and the JSON's `"expected": "no-store"` contradicts the §4.1 rule, which also accepts `private, no-cache`.

`ts audit ad-templates verify --json` (`ad_templates/output.rs:113`-`156`) is a ready model: per-URL entries in input order, a per-URL `error`, a stable `code` next to a human `message`, snake_case enum values, and `ok`/`strict` at the top. The PascalCase `"Fail"` and `"Html"` in §6.2 and Task 5 don't match it.

Sketch (apply manually, since §6, §7.2 and Tasks 3 to 5 change together):

```rust
pub(crate) struct UrlResult {
    pub url: url::Url,
    pub expected_group: ContentTypeGroup,
    pub status: Option<u16>,
    pub content_type: Option<String>,
    pub error: Option<String>,
    pub checks: Vec<HeaderCheck>,
}

pub(crate) struct HeaderCheck {
    pub header: AuditedHeader,
    pub verdict: Verdict,
    pub code: &'static str,
    pub actual: Option<String>,
    pub message: String,
    pub recommendation: Option<String>,
}

pub(crate) enum Verdict { Pass, Warn, Fail, NotAudited }
```

Apply content rules only to final 2xx responses, treat only a failed root fetch as fatal, and roll group verdicts up from the `UrlResult`s.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+| `Html`        | `text/html`                                                             |
+| `JavaScript`  | `application/javascript`, `text/javascript`, `application/x-javascript` |
+| `Image`       | `image/*`                                                               |
+| `StaticAsset` | `text/css`, `font/*`, `application/font-*`                              |
+| `RtbJson`     | `application/json`                                                      |
+| `Other`       | Everything else (not evaluated, reported as info)                       |
+
+---
+
+## 4. Cacheability Rules
```
aram356: 🔧 **wrench** — `Set-Cookie` isn't audited, and on main it overrides every other cache header. Every adapter's finalize step runs `enforce_set_cookie_cache_privacy` (on Fastly at `middleware.rs:230`, and again at `main.rs:390`), which rewrites any non-private `Cache-Control` on a cookie-bearing response to `private, max-age=0` and strips the edge headers (`response_privacy.rs:121`-`144`). TS doesn't remove origin cookies on the publisher path. Fastly's read-through cache also skips responses with `Set-Cookie` by default, and the template cache refuses them (`publisher.rs:6137`).

So an origin asset sent as `public, max-age=31536000, immutable` with a stray session or load-balancer cookie reaches browsers as `private, max-age=0` and never lands in the edge cache. It passes every rule in §4, and it's the under-caching case the epic describes. Suggest adding `Set-Cookie` to the audited headers, with a WARN on any shared-cacheable response that carries it.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+3. Parse the HTML response to discover asset URLs per content type:
+   - HTML: the `/` response itself
+   - JS: extract `<script src="...">` URLs from the HTML response
+   - Images: extract `<img src="...">` and `<link rel="icon">` URLs; fall back to `/favicon.ico`
+   - Static: extract `<link rel="stylesheet" href="...">` URLs from the HTML response
+   - RTB/JSON: not discoverable from HTML; requires explicit `--url` or presence of `/_ts/api/v1/identify` in config routes (see note below)
+4. Resolve relative URLs against `publisher.origin_url`
+5. Fetch each discovered URL directly from origin and classify by response `Content-Type`
+
+**Note on `/_ts/` paths:** Routes under `/_ts/` are served by the Trusted Server edge application, NOT by the publisher origin. The audit fetches directly from the origin server, so these paths would 404. RTB/JSON endpoints must be provided explicitly via `URLS...` arguments or discovered from the origin's HTML (e.g. inline fetch targets).
```
aram356: 🔧 **wrench** — Leftover from last round's `/_ts/` fix. This bullet still points RTB discovery at `/_ts/api/v1/identify`, an edge route (`trusted-server-adapter-fastly/src/app.rs:29`) that serves identity, not bids. The note at line 145 and Task 3 (line 370) both rule it out, and `trusted-server.toml` has no route list to read it from. The bullet also names a `--url` flag the command doesn't have (URLs are positional), and the note says RTB endpoints can be "discovered from the origin's HTML" while this bullet says they can't.

```suggestion
   - RTB/JSON: not discoverable from HTML; audited only when passed explicitly as a `URLS...` argument (see note below)
4. Resolve relative URLs against `publisher.origin_url`
5. Fetch each discovered URL directly from origin and classify by response `Content-Type`

**Note on `/_ts/` paths:** Routes under `/_ts/` are served by the Trusted Server edge application, NOT by the publisher origin. The audit fetches directly from the origin server, so these paths would 404; discovery never generates them. RTB/JSON endpoints are audited only when provided explicitly via `URLS...` arguments.
```

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+| `Surrogate-Control` | If present, contains `no-store` | FAIL: "CDN caching RTB responses"                         |
+
+Note: `no-store` alone is sufficient per RFC 9111 Section 5.2.2.5. The audit accepts any directive set that includes `no-store` (e.g. `private, no-store`, or bare `no-store`).
+
+### 4.6 ETag handling (ref: #428)
+
+When evaluating ETag presence, handle multi-value `If-None-Match` correctly:
+
+- Accept both strong (`"abc"`) and weak (`W/"abc"`) ETags
+- Multiple comma-separated values in `If-None-Match` are valid per RFC 7232
```
aram356: 🔧 **wrench** — Still open from both reviews last round. #428 is about the server comparing the `If-None-Match` request header in `serve_static_with_etag` (`http_util.rs:315`). This tool reads response headers and never sends a conditional request, and an `ETag` response header carries a single entity tag, so there's nothing multi-valued to handle. RFC 9110 obsoletes the cited RFC 7232.

The section also misses `Last-Modified`. It's a validator in its own right, so the ETag row at line 87 warns "no conditional revalidation" on assets that revalidate fine through `If-Modified-Since`. Along with this suggestion, change that row to "Validator (`ETag` or `Last-Modified`)" and Task 2's criterion at line 352 to "Accepts strong and weak `ETag` values and `Last-Modified`".

```suggestion
### 4.6 Validator handling

The audit inspects response validators only. It never sends conditional requests, so request-side `If-None-Match` parsing (#428) does not apply.

- Accept both strong (`"abc"`) and weak (`W/"abc"`) `ETag` values (RFC 9110 §8.8.3)
- Accept `Last-Modified` (RFC 9110 §8.8.2) as a validator: a response with `Last-Modified` and no `ETag` still supports conditional revalidation
```

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+
+Unlike `ts dev proxy` (macOS-only due to keychain/TLS dependencies), `ts dev audit` is cross-platform. It does not require `tokio`, `ring`, or `aws-lc-sys`.
+
+### 8.2 Error handling
+
+The CLI uses `CliResult<T> = Result<T, String>` with `cli_error()` / `report_error()` helpers (see `crates/trusted-server-cli/src/error.rs`). The audit command follows this pattern — returning descriptive `String` errors via `cli_error()` for failures like unreachable origins or invalid config paths.
+
+### 8.3 HTTP client dependency scoping
+
+`reqwest` is defined at workspace level but not currently consumed by `trusted-server-cli`. It must be added to the CLI crate's `[target.'cfg(not(target_arch = "wasm32"))'.dependencies]` section (same scope as `chromiumoxide`, `tokio`, etc.) so it does not break the workspace default `wasm32-wasip1` target:
```
aram356: 🔧 **wrench** — Following this section literally breaks `ts dev proxy`. `reqwest` 0.13 already reaches the CLI through `chromiumoxide` and `edgezero-cli`, with rustls on the `aws-lc-rs` provider. The workspace pin is 0.12, whose `rustls-tls` feature turns on rustls's `ring` provider as well. With both enabled, rustls can't pick a process default, and the proxy's `ServerConfig::builder()` (`dev/proxy/ca.rs:198`) and `ClientConfig::builder()` (`upstream/connect.rs:288`) panic.

I checked this against the PR head on macOS. Adding only `reqwest = { workspace = true }` to the CLI turns `cargo test -p trusted-server-cli --lib commands::dev::proxy::ca::tests` from 5 passed into 3 failed, each with:

```text
Could not automatically determine the process-level CryptoProvider from Rustls crate features.
```

The CLI test job in CI runs on macOS, so an implementation that follows this section would go red with that error.

```suggestion
`reqwest` is not a direct dependency of `trusted-server-cli`, but `reqwest` 0.13 already reaches the CLI's dependency tree through `chromiumoxide` and `edgezero-cli`, with rustls on the `aws-lc-rs` provider. The workspace pin is still `reqwest` 0.12, whose `rustls-tls` feature also enables rustls's `ring` provider. With both providers enabled, rustls can't choose a process default, and the `ServerConfig::builder()` and `ClientConfig::builder()` calls in `ts dev proxy` panic. First move the workspace pin to `reqwest` 0.13 with its `rustls` feature (0.13 renamed `rustls-tls`; the bump also covers `trusted-server-adapter-axum` and `trusted-server-integration-tests`). Then add the dependency to the CLI crate's `[target.'cfg(not(target_arch = "wasm32"))'.dependencies]` section (same scope as `chromiumoxide`, `tokio`, etc.) so it does not break the workspace default `wasm32-wasip1` target:
```

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+
+Responses are classified into groups by their `Content-Type` response header. The MIME type is normalized before matching: parameters are stripped (e.g. `text/html; charset=utf-8` becomes `text/html`) and the value is lowercased.
+
+| Group         | Matching Content-Types                                                  |
+| ------------- | ----------------------------------------------------------------------- |
+| `Html`        | `text/html`                                                             |
+| `JavaScript`  | `application/javascript`, `text/javascript`, `application/x-javascript` |
+| `Image`       | `image/*`                                                               |
+| `StaticAsset` | `text/css`, `font/*`, `application/font-*`                              |
+| `RtbJson`     | `application/json`                                                      |
```
aram356: 🤔 **thinking** — Raised last round and still open. RTB JSON comes from the TS edge (`POST /auction`, `GET /_ts/page-bids`), not the publisher origin. §5 can't discover JSON, and `/auction` only accepts POST while the audit sends GET, so this group only ever sees URLs an operator passes by hand. Any cacheable JSON passed that way (a search API, a feed, a config file) then fails with "stale bids served to users". Renaming the group `Json` with neutral wording, or requiring an explicit RTB flag before the `no-store` rule applies, avoids the false alarm.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+| Cacheability rules     | Hardcoded in v1                                               | YAGNI; configurable rules add complexity before we know the right defaults                                                            |
+| CDN assumption         | Fastly-aware (check `Surrogate-Control`) but don't require it | TS is Fastly-first but may run elsewhere                                                                                              |
+| URL discovery          | HTML-parse from origin root page                              | Config-free DX for common case; explicit URLs for CI and edge cases                                                                   |
+| `no-store` sufficiency | `no-store` alone passes for HTML and RTB                      | RFC 9111 §5.2.2.5: `no-store` is the strongest single directive; requiring `private` AND `no-store` would fail correct configurations |
+| CLI placement          | `ts dev audit headers`                                        | Groups under existing `dev` subcommand tree; `audit` is the sub-group for future audit types                                          |
```
aram356: 🤔 **thinking** — `ts audit` has been a top-level command since #800 (2026-07-01), and #823 (2026-09-21) turned it into a group with `page`, `ad-templates` and `generate`. Putting this under `ts dev` creates a second audit tree (`ts audit …` and `ts dev audit headers`) and a second `commands/**/audit` module. `ts audit headers` would sit next to the commands whose conventions it should share (`--json`, `--strict`, `AppConfigArgs`, `parse_http_url`, `RunOutcome`), and `ts audit generate` already builds a JavaScript asset audit from a live page that discovery could reuse. The catch is that `ts audit` describes itself as browser-backed. Worth settling against the epic before Task 1.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+
+| Argument | Description                       |
+| -------- | --------------------------------- |
+| `URLS…`  | Explicit URLs to audit (optional) |
+
+### Options
+
+| Flag              | Description                               | Default                            |
+| ----------------- | ----------------------------------------- | ---------------------------------- |
+| `--config <path>` | Path to `trusted-server.toml`             | `./trusted-server.toml`            |
```
aram356: ♻️ **refactor** — The CLI already has shared config flags for this. `AppConfigArgs` (`app_config.rs:18`) provides `--app-config` (defaulting to `<app.name>.toml` beside `edgezero.toml`), `--manifest` and `--no-env`, and `load_settings` applies the EdgeZero environment overlay. `ts config ad-templates` and `ts audit ad-templates` both flatten it. A separate `--config` defaulting to `./trusted-server.toml` would miss env-only settings such as `TRUSTED_SERVER__PUBLISHER__ORIGIN_URL` (`docs/guide/configuration.md:41`), and `--config` already means "draft config output path" on `ts audit generate`. Loading the full `Settings` also gives the audit what several other comments need: `origin_host_header_override`, `proxy.asset_routes`, `cache.asset_rules`, and `creative_opportunities`.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+| `Vary`          | Should NOT contain `User-Agent` or `Cookie`               | WARN: "Vary on User-Agent/Cookie destroys CDN hit ratio for static assets" |
+
+### 4.5 RTB/JSON (real-time, never cached)
+
+| Header              | Expected                        | Verdict if wrong                                          |
+| ------------------- | ------------------------------- | --------------------------------------------------------- |
+| `Cache-Control`     | Contains `no-store`             | FAIL: "RTB responses cached = stale bids served to users" |
+| `Surrogate-Control` | If present, contains `no-store` | FAIL: "CDN caching RTB responses"                         |
+
+Note: `no-store` alone is sufficient per RFC 9111 Section 5.2.2.5. The audit accepts any directive set that includes `no-store` (e.g. `private, no-store`, or bare `no-store`).
```
aram356: 🤔 **thinking** — Bare `no-store` should keep passing, as last round asked, but the citation doesn't support "sufficient". RFC 9111 §5.2.2.5 says the directive "is not a reliable or sufficient mechanism for ensuring privacy". Fastly's full-site delivery docs also list `no-store` among the `Cache-Control` directives Fastly ignores, with `private` as the one that means "Don't store". A publisher whose `origin_url` is itself a Fastly VCL service would have bare-`no-store` HTML or JSON cached there. TS itself always emits `no-store, private` (`cache_policy.rs:45`). A non-blocking advisory that recommends adding `private`, plus a corrected citation here and in the §9 row at line 313, would cover it.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+| `Surrogate-Control` | If present, contains `no-store` or `private`                    | FAIL: "CDN may cache personalized HTML"                                                     |
+
+### 4.2 JavaScript bundles (hashed filenames, immutable)
+
+| Header          | Expected                                                  | Verdict if wrong                                                                               |
+| --------------- | --------------------------------------------------------- | ---------------------------------------------------------------------------------------------- |
+| `Cache-Control` | Contains `public` AND `max-age>=31536000` AND `immutable` | WARN: "Hashed JS bundles should use long-lived immutable caching"                              |
+| `ETag`          | Present (strong preferred)                                | WARN: "Missing ETag means no conditional revalidation"                                         |
+| `Surrogate-Key` | Present                                                   | WARN: "Without Surrogate-Key, selective CDN purge is impossible -- full-cache purges required" |
```
aram356: 🤔 **thinking** — §9 says the audit is Fastly-aware "but don't require it", yet a missing `Surrogate-Key` is a WARN on every cacheable group, and the worst-of rollup turns it into a group WARN. Every Cloudflare, Spin and Axum deployment would get a WARN on JavaScript, Image and StaticAsset. The message is also inaccurate: Fastly can purge a single URL, so purging without `Surrogate-Key` is still selective, just per URL instead of per tag. Suggest making it advisory, gated on a Fastly target, and left out of the rollup.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+
+When no explicit URLs are provided:
+
+1. Read `publisher.origin_url` from `trusted-server.toml`
+2. Fetch the root HTML page from origin (`/`)
+3. Parse the HTML response to discover asset URLs per content type:
+   - HTML: the `/` response itself
+   - JS: extract `<script src="...">` URLs from the HTML response
+   - Images: extract `<img src="...">` and `<link rel="icon">` URLs; fall back to `/favicon.ico`
+   - Static: extract `<link rel="stylesheet" href="...">` URLs from the HTML response
```
aram356: 🤔 **thinking** — Discovery can't reach parts of the taxonomy. Fonts load from CSS `@font-face` or `<link rel="preload" as="font">`, and neither is parsed, so StaticAsset only ever means CSS. Responsive images (`srcset`, `<picture>`), lazy-loaded `data-src`, and `rel="shortcut icon"` are missed. HTML is sampled from `/` only, and many sites serve articles from a different template. Creatives never appear in origin HTML (they go through TS's `/first-party/proxy`), so the "Images / Creatives" heading at line 91 promises more than this can audit. Parsing preload hints, `srcset`, and one level of CSS `url()`, plus accepting several HTML sample URLs, would close most of it. The CLI already depends on `scraper` for the parsing.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+- Unit tests cover all classification + evaluation logic
+
+---
+
+### Task 3: Origin request engine
+
+**Type:** Net-new (`commands/dev/audit/headers/fetch.rs`)
+**Dependencies:** Task 2 (uses `ContentTypeGroup`)
+
+Fetch responses from origin using `reqwest` (scoped to `cfg(not(target_arch = "wasm32"))`). Support explicit URL list and HTML-parsed discovery from origin root page.
```
aram356: 🤔 **thinking** — The fetcher follows URLs taken from untrusted HTML, on operator laptops and CI runners, with no limits. A discovered URL or redirect target can point at loopback, RFC 1918 addresses, or a cloud metadata endpoint. There's no cap on URL count, concurrency, or bytes, and a GET downloads every asset in full when only headers matter. Origin-controlled strings (header values, URLs) reach the terminal, and the crate already has `escape_terminal_text` (`ad_templates/output.rs:28`) for exactly that. If third-party URLs stay in scope, fetching ad pixels from the operator's IP also pollutes the publisher's analytics and invalid-traffic signals. Suggest allowing only the hosts from the URL-mapping comment, rejecting private and link-local targets unless the origin itself resolves there, capping URL count and total time, dropping bodies after headers, and routing human output through `escape_terminal_text`.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+```rust
+/// The `ts dev …` command group.
+#[derive(Debug, clap::Subcommand)]
+pub enum DevCommand {
+    /// Run the local production-hostname dev proxy (macOS only).
+    #[cfg(target_os = "macos")]
+    Proxy(proxy::ProxyArgs),
+
+    /// Audit origin cache headers for correctness.
+    Audit(DevAuditCommand),
```
aram356: ⛏ **nitpick** — This snippet doesn't compile. A variant that wraps a `Subcommand` enum needs `#[command(subcommand)]`, as `run.rs` does for `Config` and `Dev`. With the workspace's clap 4.6.1, the snippet as written fails with `the trait bound 'DevAuditCommand: clap::Args' is not satisfied`; with the attribute, `ts dev audit headers --json` parses.

```suggestion
    /// Audit origin cache headers for correctness.
    #[command(subcommand)]
    Audit(DevAuditCommand),
```

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+}
+
+#[derive(Debug, clap::Subcommand)]
+pub enum DevAuditCommand {
+    /// Audit cache-related response headers per content type.
+    Headers(AuditHeadersArgs),
+}
+```
+
+Unlike `ts dev proxy` (macOS-only due to keychain/TLS dependencies), `ts dev audit` is cross-platform. It does not require `tokio`, `ring`, or `aws-lc-sys`.
```
aram356: ⛏ **nitpick** — The conclusion holds, but not the reason. `reqwest` depends on `tokio` on every non-wasm target, and `aws-lc-sys` is already in the CLI's Linux tree through `reqwest` 0.13. What keeps `ts dev proxy` macOS-only is the keychain and `networksetup` integration, and the `cfg(not(target_arch = "wasm32"))` scope is what keeps these dependencies off `wasm32-wasip1`. See the comment on line 296 for the `ring` part.

```suggestion
Unlike `ts dev proxy` (macOS-only: CA trust via the login keychain and Safari proxy automation via `networksetup`), `ts dev audit` is cross-platform. Its HTTP client stack (`reqwest`, `tokio`, and a rustls crypto provider) builds on every host target, so scoping it to `cfg(not(target_arch = "wasm32"))` is enough to keep it off the workspace default `wasm32-wasip1` target.
```

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+| `Image`       | `image/*`                                                               |
+| `StaticAsset` | `text/css`, `font/*`, `application/font-*`                              |
+| `RtbJson`     | `application/json`                                                      |
+| `Other`       | Everything else (not evaluated, reported as info)                       |
+
+---
+
+## 4. Cacheability Rules
+
+Each content-type group has an expected caching posture. Rules are evaluated against: `Cache-Control`, `Surrogate-Control`, `s-maxage`, `Surrogate-Key`, `Vary`, `ETag`.
```
aram356: ⛏ **nitpick** — `s-maxage` is a `Cache-Control` directive, not a header.

```suggestion
Each content-type group has an expected caching posture. Rules are evaluated against: `Cache-Control` (including its `s-maxage` directive), `Surrogate-Control`, `Surrogate-Key`, `Vary`, `ETag`.
```

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+When explicit URLs are provided, skip discovery and classify each by its response `Content-Type`.
+
+---
+
+## 6. Output Format
+
+### 6.1 Human-readable (default)
+
+```
+Origin: https://origin.publisher.com
```
aram356: ⛏ **nitpick** — `publisher.com` is a registered domain, and AGENTS.md asks for example domains in docs. The JSON example needs the same fix at lines 175 and 179 (separate suggestion there).

```suggestion
Origin: https://origin.example.com
```

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+
+### 6.2 JSON (`--json`)
+
+```json
+{
+  "origin": "https://origin.publisher.com",
+  "groups": [
+    {
+      "content_type": "Html",
+      "urls_sampled": ["https://origin.publisher.com/"],
```
aram356: ⛏ **nitpick** — Same example-domain fix as line 156.

```suggestion
  "origin": "https://origin.example.com",
  "groups": [
    {
      "content_type": "Html",
      "urls_sampled": ["https://origin.example.com/"],
```

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+
+New modules in `crates/trusted-server-cli/src/commands/dev/`:
+
+```
+audit/
+  mod.rs          -- pub entry point: DevAuditCommand enum + dispatch
+  headers/
+    mod.rs        -- run_audit_headers()
+    rules.rs      -- ContentTypeGroup, CachePolicy, evaluate()
+    fetch.rs      -- AuditHeadersArgs, origin fetching, classification
```
aram356: ⛏ **nitpick** — A few wiring details for the tasks:

- `AuditHeadersArgs` lives in `fetch.rs`, which Task 3 creates, but Task 1 needs it for `Headers(AuditHeadersArgs)` while the graph gives Task 1 no dependencies. Defining the args in `headers/mod.rs` as part of Task 1 fixes the ordering.
- Task 1 should also drop the `cfg_attr(not(target_os = "macos"), allow(clippy::needless_pass_by_value))` and the "no subcommands off macOS" comments in `commands/dev/mod.rs`, plus the macOS-only wording in the `Dev` help text (`run.rs:47`), since `DevCommand` won't be empty off macOS anymore.
- Printing with `println!` in Task 5 would trip the workspace `print_stdout`/`print_stderr` lints (`Cargo.toml:145`), which CI runs with `-D warnings`, and the crate's `output` helper module only compiles on macOS (`lib.rs:24`). Write through an injected `io::Write` the way `ts audit ad-templates verify` and `prebid_bundle.rs` do, and keep stdout to JSON only in `--json` mode.

docs/superpowers/specs/2026-07-14-cache-header-audit-design.md
```diff
@@ -0,0 +1,457 @@
...
+- Exit codes: 0=pass, 1=fail, 2=warn-only
+
+---
+
+### Task 6: End-to-end wiring + integration tests
+
+**Type:** Integration
+**Dependencies:** Tasks 1-5
+
+Wire `run_audit_headers()` into dispatch. Integration test with mock HTTP server.
```
aram356: 🌱 **seedling** — `crates/trusted-server-cli/tests/audit_config_diagnostics.rs` already has a plain `TcpListener` fixture server that isn't macOS-gated, so Task 6 can reuse it without a new mock-server dependency. Worth adding to the acceptance criteria: exit codes 0, 1 and 2 including a tool failure, a `--json` golden file, the URL-mapping cases (relative, same-site absolute, third-party), a redirect, and the Host override. For Task 7, the new guide page also needs a sidebar entry in `docs/.vitepress/config.mts` next to CLI and Dev Proxy (lines 109-112), or nothing links to it.

aram356 (changes requested):
## Summary

The round-1 blockers on the mermaid graph, the `DevCommand` premise, `CliResult`, bare `no-store`, and the #293 link are resolved. Since that round, main has changed under the spec: TS now sets HTML cache privacy itself, has an opt-in template cache and operator asset rules, and documents the CLI's exit codes and `--json` conventions. Several rules now contradict shipped behavior, and three design gaps (redirects, edge-vs-browser policy, per-URL results) would have the tool grade responses TS never serves. Most of the fixes land in §4, §5 and §7.2.

> 9 of the inline comments below carry a one-click GitHub `suggestion`. Use **Commit suggestion** (or **Add suggestion to batch** for several at once) to apply them as commits on the PR branch. The remaining comments describe the fix in prose because it spans several sections of the spec.

## Blocking

### 🔧 wrench

- **HTML rule fails correct configs and disables the template cache**: see inline at `2026-07-14-cache-header-audit-design.md:77-80`
- **Rules grade Cache-Control as the edge policy (Surrogate-Control, CDN-Cache-Control, Age, Expires ignored)**: see inline at `2026-07-14-cache-header-audit-design.md:71`
- **`immutable` recommended for unfingerprinted JS/CSS/fonts**: see inline at `2026-07-14-cache-header-audit-design.md:86`
- **Exit code 2 collides with the CLI's tool-failure code**: see inline at `2026-07-14-cache-header-audit-design.md:40-48`
- **Audit requests don't match what TS sends to origin**: see inline at `2026-07-14-cache-header-audit-design.md:142-143`
- **Following redirects grades a different response than TS receives**: see inline at `2026-07-14-cache-header-audit-design.md:457`
- **No per-URL record: verdicts can't name the URL; errors and status codes unmodeled**: see inline at `2026-07-14-cache-header-audit-design.md:228-243`
- **`Set-Cookie` isn't audited, and it overrides every other cache header**: see inline at `2026-07-14-cache-header-audit-design.md:67`
- **RTB bullet still routes discovery to an edge-only route**: see inline at `2026-07-14-cache-header-audit-design.md:141-145`
- **§4.6 still conflates request-side `If-None-Match` (#428); `Last-Modified` missing**: see inline at `2026-07-14-cache-header-audit-design.md:117-122`
- **§8.3's `reqwest = { workspace = true }` makes `ts dev proxy` panic**: see inline at `2026-07-14-cache-header-audit-design.md:296`

## Non-blocking

- 🤔 **All `application/json` graded as RTB**: see inline at `2026-07-14-cache-header-audit-design.md:62`
- 🤔 **Placement: `ts audit` is now the audit group**: see inline at `2026-07-14-cache-header-audit-design.md:314`
- ♻️ **Reuse `AppConfigArgs` instead of a new `--config`**: see inline at `2026-07-14-cache-header-audit-design.md:38`
- 🤔 **Bare `no-store`: citation doesn't say "sufficient"; advise adding `private`**: see inline at `2026-07-14-cache-header-audit-design.md:115`
- 🤔 **Missing `Surrogate-Key` WARNs every non-Fastly deployment; message is inaccurate**: see inline at `2026-07-14-cache-header-audit-design.md:88`
- 🤔 **Discovery can't reach parts of the taxonomy**: see inline at `2026-07-14-cache-header-audit-design.md:138-140`
- 🤔 **Fetcher has no host allowlist, limits, or output escaping**: see inline at `2026-07-14-cache-header-audit-design.md:363`
- ⛏ **§8.1 snippet needs `#[command(subcommand)]`**: see inline at `2026-07-14-cache-header-audit-design.md:277-278`
- ⛏ **Cross-platform rationale misstates the dependency facts**: see inline at `2026-07-14-cache-header-audit-design.md:288`
- ⛏ **`s-maxage` is a directive, not a header**: see inline at `2026-07-14-cache-header-audit-design.md:69`
- ⛏ **Real domain in examples (human output)**: see inline at `2026-07-14-cache-header-audit-design.md:156`
- ⛏ **Real domain in examples (JSON)**: see inline at `2026-07-14-cache-header-audit-design.md:175-179`
- ⛏ **Task wiring details (args location, cfg cleanup, output writer)**: see inline at `2026-07-14-cache-header-audit-design.md:215`
- 🌱 **Test fixture reuse, acceptance cases, docs sidebar**: see inline at `2026-07-14-cache-header-audit-design.md:412`

- 📝 **PR description is still the unfilled template**: the Summary and Changes table are empty, and there's a dangling `Closes #` below the real `Closes #835`. Worth filling in, especially if this PR will also carry the implementation.

## CI Status

- Analyze (actions): PASS
- Analyze (rust): PASS
- cargo fmt: PASS (required)
- cargo test: PASS (required)
- CodeQL: PASS
- integration tests: PASS
- vitest: PASS

PR #933, opened 2026-07-18
Preserve integration configuration string types

### LINKED ISSUE #932: Preserve string types in Prebid bidder parameter overrides
## Problem

`IntegrationSettings::normalize_env_value` recursively parses every string as JSON. Numeric-looking string values inside Prebid bidder parameter overrides are therefore converted to JSON numbers during app-config loading and serialization.

For example, a TOML value such as:

```toml
[integrations.prebid.bid_param_overrides.example]
publisherId = "12345"
```

can become:

```json
{"publisherId": 12345}
```

Prebid Server adapters whose schemas require string identifiers then reject the OpenRTB request.

## Expected behavior

Bidder parameter override payloads are opaque JSON supplied by the operator. Their nested scalar types should survive TOML loading, config-store serialization, and runtime JSON loading unchanged. JSON environment overrides for the complete override container should still be decoded once.

## Acceptance criteria

- Preserve nested scalar types under `bid_param_overrides`, `bid_param_zone_overrides`, and `bid_param_override_rules`.
- Continue accepting JSON-encoded environment overrides for those fields.
- Add coverage for numeric-looking strings across the TOML → JSON → runtime round trip.

ChristianPavilonis:
## Summary

- preserve scalar types for every integration configuration on production TOML and config-blob loading paths; this fixes Prebid bidder parameters that look numeric
- remove the obsolete legacy test-only integration environment coercion and its stale coverage
- regress TOML → runtime JSON config loading through typed Prebid rule compilation and emitted OpenRTB bidder parameters
- document the breaking migration for quoted integration booleans and numbers

Closes #932

Follow-up: #972 tracks outdated integration environment-override documentation that predates this PR.

## Testing

- `cargo test_details -p trusted-server-core`
- `cargo clippy-fastly && cargo clippy-axum && cargo clippy-cloudflare && cargo clippy-cloudflare-wasm && cargo clippy-spin-native && cargo clippy-spin-wasm`
- `cargo test-fastly && cargo test-axum && cargo test-cloudflare && cargo test-spin`
- `cd crates/trusted-server-js/lib && npx vitest run && npm run format`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - **Breaking** — `bid_param_zone_overrides` inner values must now be JSON objects; previously non-object or empty values (`"header" = "x"`, `"header" = {}`) were accepted and silently produced a dead rule at runtime. They now fail at startup with a configuration error. Operators upgrading should audit their `bid_param_zone_overrides` config for non-object zone entries.
+- **Breaking** — Integration configuration strings are no longer globally reinterpreted as JSON scalars. Operators upgrading should audit `[integrations.*]` settings and use native TOML/typed-config booleans and numbers (for example, `enabled = true`, not `enabled = "true"`); quoted numeric and boolean scalars now fail validation instead of silently converting.
 - **Breaking** — Sourcepoint browser module inclusion now requires explicit `[integrations.sourcepoint].enabled = true`; operators relying on the previous unconditional Sourcepoint module should enable the integration before upgrading.
 
 ### Security
--- a/crates/trusted-server-core/src/integrations/prebid.rs
+++ b/crates/trusted-server-core/src/integrations/prebid.rs
@@ -289,11 +289,6 @@ pub struct PrebidIntegrationConfig {
     /// param1 = 12345
     /// param2 = "value"
     /// ```
-    ///
-    /// Example via environment variable:
-    /// ```text
-    /// TRUSTED_SERVER__INTEGRATIONS__PREBID__BID_PARAM_OVERRIDES='{"bidder-name":{"param1":12345,"param2":"value"}}'
-    /// ```
     #[serde(default)]
     pub bid_param_overrides: HashMap<String, serde_json::Map<String, Json>>,
     /// Canonical ordered bidder-param override rules.
@@ -310,11 +305,6 @@ pub struct PrebidIntegrationConfig {
     /// when.bidder = "kargo"
     /// when.zone = "header"
     /// set = { placementId = "_abc" }
-    /// ```
-    ///
-    /// Example via environment variable:
-    /// ```text
-    /// TRUSTED_SERVER__INTEGRATIONS__PREBID__BID_PARAM_OVERRIDE_RULES='[{"when":{"bidder":"kargo","zone":"header"},"set":{"placementId":"_abc"}}]'
     /// ```
     #[serde(default)]
     pub bid_param_override_rules: Vec<BidParamOverrideRule>,
--- a/crates/trusted-server-core/src/integrations/prebid.rs
+++ b/crates/trusted-server-core/src/integrations/prebid.rs
@@ -5719,6 +5719,91 @@ set = { keywords = { genre = "news" } }
     }
 
     #[test]
+    fn bid_param_override_numeric_strings_survive_runtime_config_roundtrip() {
+        let toml_str = format!(
+            r#"{}
+
+[integrations.prebid]
+enabled = true
+server_url = "https://prebid.example"
+bidders = ["pubmatic"]
+
+[integrations.prebid.bid_param_overrides.pubmatic]
+publisherId = "12345"
+adSlot = "67890"
+
+[integrations.prebid.bid_param_zone_overrides.pubmatic]
+header = {{ placementId = "24680" }}
+
+[[integrations.prebid.bid_param_override_rules]]
+when = {{ bidder = "pubmatic", zone = "in_content" }}
+set = {{ placementId = "13579" }}
+"#,
+            TOML_BASE
+        );
+        let settings = Settings::from_toml(&toml_str).expect("should parse TOML settings");
+
+        // This mirrors the typed data transition beneath `ts config push`:
+        // `BlobEnvelope` data is deserialized by `settings_from_config_blob`.
+        let serialized = serde_json::to_value(&settings).expect("should serialize settings");
+        let runtime_settings =
+            Settings::from_json_value(serialized).expect("should parse runtime JSON settings");
+        let config = runtime_settings
+            .integration_config::<PrebidIntegrationConfig>(PREBID_INTEGRATION_ID)
+            .expect("should parse Prebid config")
+            .expect("should enable Prebid config");
+
+        let static_request = make_auction_request(vec![make_ts_slot(
+            "ad-static-0",
+            &json!({ "pubmatic": { "keep": "client" } }),
+            None,
+        )]);
+        let static_ortb = call_to_openrtb(config.clone(), &static_request);
+        let static_params = bidder_params(&static_ortb);
+        assert_eq!(
+            static_params["pubmatic"]["publisherId"],
+            json!("12345"),
+            "static override publisherId should remain a string"
+        );
+        assert_eq!(
+            static_params["pubmatic"]["adSlot"],
+            json!("67890"),
+            "static override adSlot should remain a string"
+        );
+        assert_eq!(
+            static_params["pubmatic"]["keep"],
+            json!("client"),
+            "static override should preserve client parameters"
+        );
+
+        let header_request = make_auction_request(vec![make_ts_slot(
+            "ad-header-0",
+            &json!({ "pubmatic": {} }),
+            Some("header"),
+        )]);
+        let header_ortb = call_to_openrtb(config.clone(), &header_request);
+        let header_params = bidder_params(&header_ortb);
+        assert_eq!(
+            header_params["pubmatic"]["placementId"],
+            json!("24680"),
+            "zone override placementId should remain a string"
+        );
+
+        let in_content_request = make_auction_request(vec![make_ts_slot(
+            "ad-in-content-0",
+            &json!({ "pubmatic": {} }),
+            Some("in_content"),
+        )]);
+        let in_content_ortb = call_to_openrtb(config, &in_content_request);
+        let in_content_params = bidder_params(&in_content_ortb);
+        assert_eq!(
+            in_content_params["pubmatic"]["placementId"],
+            json!("13579"),
+            "canonical rule placementId should remain a string"
+        );
+    }
+
+    #[test]
     fn zone_override_replaces_placement_id() {
         let mut config = base_config();
         config.bidders = vec!["kargo".to_string()];
--- a/crates/trusted-server-core/src/settings.rs
+++ b/crates/trusted-server-core/src/settings.rs
@@ -223,7 +223,7 @@ impl IntegrationSettings {
     fn normalize_opaque_json_value(value: JsonValue) -> JsonValue {
         match value {
             JsonValue::String(raw) => {
-                serde_json::from_str::<JsonValue>(&raw).unwrap_or_else(|_| JsonValue::String(raw))
+                serde_json::from_str::<JsonValue>(&raw).unwrap_or(JsonValue::String(raw))
             }
             other => other,
         }
--- a/crates/trusted-server-core/src/settings.rs
+++ b/crates/trusted-server-core/src/settings.rs
@@ -191,7 +191,19 @@ impl IntegrationSettings {
         match value {
             JsonValue::Object(map) => JsonValue::Object(
                 map.into_iter()
-                    .map(|(key, val)| (key, Self::normalize_env_value(val)))
+                    .map(|(key, value)| {
+                        let value = if matches!(
+                            key.as_str(),
+                            "bid_param_overrides"
+                                | "bid_param_zone_overrides"
+                                | "bid_param_override_rules"
+                        ) {
+                            Self::normalize_opaque_json_value(value)
+                        } else {
+                            Self::normalize_env_value(value)
+                        };
+                        (key, value)
+                    })
                     .collect(),
             ),
             JsonValue::Array(items) => {
@@ -203,6 +215,15 @@ impl IntegrationSettings {
                 } else {
                     JsonValue::String(raw)
                 }
+            }
+            other => other,
+        }
+    }
+
+    fn normalize_opaque_json_value(value: JsonValue) -> JsonValue {
+        match value {
+            JsonValue::String(raw) => {
+                serde_json::from_str::<JsonValue>(&raw).unwrap_or_else(|_| JsonValue::String(raw))
             }
             other => other,
         }
@@ -3199,6 +3220,38 @@ origin_host_header_overide = "www.example.com""#,
     }
 
     #[test]
+    fn prebid_numeric_string_bid_param_overrides_survive_config_roundtrip() {
+        let toml_str = format!(
+            r#"{}
+
+[integrations.prebid.bid_param_overrides.pubmatic]
+publisherId = "12345"
+adSlot = "67890"
+"#,
+            crate_test_settings_str()
+        );
+        let settings = Settings::from_toml(&toml_str).expect("should parse TOML settings");
+        let serialized = serde_json::to_value(&settings).expect("should serialize settings");
+        let runtime_settings =
+            Settings::from_json_value(serialized).expect("should parse runtime JSON settings");
+        let raw = runtime_settings
+            .integrations
+            .get("prebid")
+            .expect("should contain Prebid settings");
+
+        assert_eq!(
+            raw["bid_param_overrides"]["pubmatic"]["publisherId"],
+            json!("12345"),
+            "should preserve numeric-looking publisherId as a string"
+        );
+        assert_eq!(
+            raw["bid_param_overrides"]["pubmatic"]["adSlot"],
+            json!("67890"),
+            "should preserve numeric-looking adSlot as a string"
+        );
+    }
+
+    #[test]
     fn test_prebid_bid_param_overrides_override_with_json_env() {
         let toml_str = crate_test_settings_str();
         let env_key = format!(
@@ -3221,7 +3274,7 @@ origin_host_header_overide = "www.example.com""#,
             || {
                 temp_env::with_var(
                     env_key,
-                    Some(r#"{"criteo":{"networkId":99999,"pubid":"server-pub"}}"#),
+                    Some(r#"{"criteo":{"networkId":99999,"pubid":"24680"}}"#),
                     || {
                         let settings = Settings::from_toml_and_env(&toml_str)
                             .expect("Settings should parse with bidder param override env");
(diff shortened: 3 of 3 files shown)


crates/trusted-server-core/src/settings.rs
```diff
@@ -2025,12 +2023,13 @@ impl Settings {
...
                 message: "Failed to build configuration".to_string(),
             })?;
-        let settings: Self =
+        let mut settings: Self =
             config
                 .try_deserialize()
                 .change_context(TrustedServerError::Configuration {
                     message: "Failed to deserialize configuration".to_string(),
                 })?;
+        settings.integrations.normalize_legacy_env();
```
aram356: ❓ **question** — Why keep the four legacy-env tests that still depend on this coercion?

This PR deletes the two Prebid JSON-env tests, but leaves four tests that are now the *only* consumers of `normalize_env_value` / `normalize_legacy_env`:

- `test_env_var_roundtrip_normalizes_integration_types` (settings.rs:4222)
- `test_integration_settings_from_env` (settings.rs:3995)
- `test_datadome_protection_scope_overrides_with_json_env` (settings.rs:3249)
- `test_datadome_protection_scope_overrides_with_indexed_env` (settings.rs:3340)

I verified this by deleting the call on this line and running `cargo test -p trusted-server-core`: **exactly those 4 fail, the other 1643 pass.**

Two reasons they look stale rather than load-bearing:

1. `test_env_var_roundtrip_normalizes_integration_types` is documented as testing "the full build.rs round-trip … env vars are baked into `Settings` at build time via `from_toml_and_env`, serialized to TOML, then parsed back at runtime". There is no build.rs settings pipeline anymore — `ts config push` goes through EdgeZero's typed app-config loader.
2. `test_datadome_protection_scope_overrides_with_json_env` asserts a *table* can be replaced wholesale by a JSON env var. EdgeZero's `apply_env_overlay` coerces each env string into the existing TOML value's type and explicitly rejects array/table targets (`env var … cannot override array / table values — env overlay supports scalar leaves only`), which is exactly why removing the Prebid env doc examples in this PR is correct.

If those four are migrated to `from_toml` / `from_json_value`, then `normalize_env_value`, `normalize_legacy_env`, and `from_toml_and_env` can all be deleted outright — finishing what this PR started rather than leaving a `#[cfg(test)]` shim that keeps semantics alive which no production path has.

crates/trusted-server-core/src/settings.rs
```diff
@@ -208,11 +209,8 @@ impl IntegrationSettings {
         }
     }
 
-    /// Normalizes all entries in place, converting JSON-encoded strings from
-    /// environment variables into their proper typed representations.
-    /// Called eagerly after deserialization so that TOML serialization in
-    /// build.rs preserves correct types.
-    pub fn normalize(&mut self) {
+    #[cfg(test)]
+    fn normalize_legacy_env(&mut self) {
```
aram356: ⛏ **nitpick** — The doc comment that `normalize()` carried was dropped in the rename. Since this is now a test-only shim, a one-liner explaining *why* it still exists would save the next reader a `git blame`:

```rust
/// Legacy test-only shim: expands JSON-encoded strings produced by the
/// `config` crate's `Environment` source. Production config never carries
/// untyped strings — `EdgeZero`'s env overlay coerces against the existing
/// TOML value's type — so this must not run on the runtime paths.
#[cfg(test)]
fn normalize_legacy_env(&mut self) {
```

Also, the `val` → `value` rename in `normalize_env_value` above adds a diff line without changing anything.

crates/trusted-server-core/src/settings.rs
```diff
@@ -3199,109 +3197,51 @@ origin_host_header_overide = "www.example.com""#,
...
-    fn test_prebid_bid_param_overrides_override_with_json_env() {
-        let toml_str = crate_test_settings_str();
-        let env_key = format!(
-            "{}{}INTEGRATIONS{}PREBID{}BID_PARAM_OVERRIDES",
-            ENVIRONMENT_VARIABLE_PREFIX,
-            ENVIRONMENT_VARIABLE_SEPARATOR,
-            ENVIRONMENT_VARIABLE_SEPARATOR,
-            ENVIRONMENT_VARIABLE_SEPARATOR
-        );
+    fn prebid_numeric_string_bid_param_overrides_survive_config_roundtrip() {
```
aram356: ♻️ **refactor** — Good regression test (I confirmed it fails on `main` with `left: Number(12345), right: String("12345")`), but it stops one layer short of the reported defect.

It asserts on the raw `integrations.get("prebid")` JSON map. The bug in #932 is about the params actually sent to bidders, which are produced by the compiled rule engine — `CompiledBidParamOverrideRule::from_bidder_override` / `from_zone_override` in `integrations/prebid.rs`. A type regression introduced during rule compilation or the shallow merge would not be caught here.

Suggest extending the same test to also assert through the typed path:

```rust
let cfg = runtime_settings
    .integration_config::<PrebidIntegrationConfig>("prebid")
    .expect("Prebid config query should succeed")
    .expect("Prebid config should exist");
assert_eq!(
    cfg.bid_param_overrides["pubmatic"]["publisherId"],
    json!("12345"),
    "should preserve numeric-looking publisherId through the typed config"
);
```

and ideally one assertion on the merged params emitted by the compiled rules.

One more thing worth a comment in the test: the `TOML -> Settings -> serde_json::to_value -> from_json_value` chain it exercises *is* the production `ts config push` -> `BlobEnvelope` -> `settings_from_config_blob` path (`config_payload.rs:39`). Saying so makes it obvious this isn't a synthetic round-trip.

aram356 (changes requested):
## Summary

Correct fix for a real and broader-than-Prebid defect: `IntegrationSettings::normalize()` ran from `finalize_deserialized()`, so it applied to `Settings::from_toml` (adapters) **and** `Settings::from_json_value` — the runtime config-store blob path in `config_payload.rs:39`. Every string in every integration's config was re-parsed as JSON, turning `publisherId = "12345"` into `12345` before it reached the auction request. Scoping the coercion to `#[cfg(test)]` and calling it only from the already-`#[cfg(test)]` `from_toml_and_env` helper is the right layering, since `deserialize_with = "from_value_or_str"` is the sanctioned per-field string-tolerance mechanism.

I verified the removal is safe rather than assuming it: EdgeZero v0.0.4's `apply_env_overlay` (`edgezero-core/src/app_config.rs:404-480`) coerces each env string into the **existing TOML value's type** and explicitly rejects array/table targets. So no untyped strings reach production config, and the env-var doc examples removed from `prebid.rs` documented a capability production never had.

Requesting changes on two points: a missing operator-facing migration note, and four leftover tests that now assert behavior no production path has.

## Blocking

### 🔧 wrench

- **Operator-visible behavior change has no CHANGELOG entry** (`CHANGELOG.md`): This is not only a Prebid fix — it changes type handling for all 15 integration configs. None of their `enabled: bool` fields use `deserialize_with = "from_value_or_str"` (checked every file in `crates/trusted-server-core/src/integrations/`), so a config carrying `enabled = "true"` or `timeout_ms = "1000"` previously loaded fine and now fails at startup with `invalid type: string "true", expected a boolean`. Relatedly, `is_explicitly_disabled` can no longer see `enabled = "false"`, so that config becomes a parse error instead of a clean disable — loud failure is arguably the better outcome, but it should be an intentional, documented one. `Unreleased → Changed` already documents an equivalent `bid_param_zone_overrides` breaking change; please add a sibling entry telling operators to audit quoted scalars in `[integrations.*]`.

### ❓ question

- **Four legacy-env tests still assert behavior production does not have** (`settings.rs:2032` — details inline).

## Non-blocking

### ♻️ refactor

- **Regression test stops one layer short of the defect** (`settings.rs:3200` — details inline).

### 🤔 thinking

- **PR description understates the scope**: the title and body read as Prebid-only, but the behavior change spans every integration's config and every non-test load path. Worth stating so reviewers and operators reading the merge commit understand the blast radius.

### 📌 out of scope

- **Docs still advertise integration env-var overrides, including forms EdgeZero rejects**: `docs/guide/integrations-overview.md:293-308`, `docs/guide/error-reference.md:117-123`, and `trusted-server.example.toml:163` (`TRUSTED_SERVER__CREATIVE_OPPORTUNITIES__SLOT='[{…}]'` — an array override the overlay refuses). These were already inaccurate before this PR, but removing the equivalent Prebid examples here makes the inconsistency conspicuous. Worth a follow-up issue.

### ⛏ nitpick

- **`normalize_legacy_env` lost its doc comment** and the `val` → `value` rename is diff noise (`settings.rs:213` — details inline).

### 👍 praise

- **Fixed at the source instead of special-casing Prebid** — deleting the blanket coercion rather than adding a string-preserving exception for `bid_param_overrides` removes a whole class of silent retyping.
- **The new test is a genuine regression test** (`settings.rs:3200`) — cherry-picked onto `main` it fails with `left: Number(12345), right: String("12345")`, and it covers `bid_param_zone_overrides` and `bid_param_override_rules` in addition to the two shapes named in the PR body.

## CI Status

All 19 GitHub checks pass (fmt, clippy, cargo test across fastly/axum/cloudflare/spin/CLI, cross-adapter parity, integration tests, vitest, CodeQL).

Independently reproduced locally against a worktree of the head commit:

- `cargo fmt --all -- --check`: PASS
- `cargo clippy-axum`: PASS
- `cargo test -p trusted-server-core`: PASS (1647 passed, 0 failed; `main` baseline 1648)
- `cargo test-axum`: PASS

crates/trusted-server-core/src/settings.rs
ChristianPavilonis: Addressed in e62ff903. Removed the four legacy environment-override tests and deleted both recursive normalizers. `from_toml_and_env` remains only for unrelated test-only parser and validation coverage; it no longer performs integration-value coercion.

crates/trusted-server-core/src/settings.rs
ChristianPavilonis: Resolved in e62ff903: both `normalize_env_value` and `normalize_legacy_env` were removed, so the requested test-only-shim documentation and rename are no longer applicable.

crates/trusted-server-core/src/settings.rs
ChristianPavilonis: Addressed in e62ff903. The regression now lives with the Prebid override-engine tests and exercises TOML → `Settings` → runtime JSON → typed `PrebidIntegrationConfig` → compiled-rule OpenRTB emission. It asserts numeric-looking strings after static, zone, and canonical rule merging, and documents the `ts config push` / `BlobEnvelope` / `settings_from_config_blob` production relevance.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -5728,6 +5718,91 @@ set = { keywords = { genre = "news" } }
         )
     }
 
+    #[test]
+    fn bid_param_override_numeric_strings_survive_runtime_config_roundtrip() {
```
prk-Jr, line 5722: 👍 **praise** — This is the right shape for the regression test. Going through `serde_json::to_value` → `Settings::from_json_value` mirrors the real `BlobEnvelope` / `settings_from_config_blob` transition instead of only exercising `from_toml`, and it covers all three override surfaces (static `bid_param_overrides`, `bid_param_zone_overrides`, and a canonical `bid_param_override_rules` entry) in one pass. This is the version of the test that would have caught #932.

crates/trusted-server-core/src/settings.rs
```diff
@@ -2050,7 +2019,6 @@ impl Settings {
         mut settings: Self,
         validation_label: &str,
     ) -> Result<Self, Report<TrustedServerError>> {
-        settings.integrations.normalize();
         settings.proxy.normalize();
```
prk-Jr, line 2022: 📝 **note** — Confirmed this is safe to drop rather than move: `from_toml_and_env` (the only source that produced JSON-encoded strings for integration entries) is already `#[cfg(test)]`-gated with no callers outside this file's tests, and EdgeZero's app-config env overlay coerces each env value to the existing TOML leaf's type (`edgezero-core` `coerce_env_value`, v0.0.4) rather than injecting raw strings. So after this change no production path can hand `IntegrationSettings` an untyped scalar.

One follow-on worth a CHANGELOG sentence: the already-pushed `app_config` blob was serialized by the *old* CLI and still has the coerced values baked in (`publisherId` stored as a number). Deploying the new WASM without re-running `ts config push` leaves #932 unfixed and silent.

prk-Jr (approved):
## Summary

Deletes `IntegrationSettings::normalize` / `normalize_env_value` and its call in `finalize_deserialized`, so integration config values are no longer recursively re-parsed as JSON. This is the right fix for #932: the recursive re-parse ran on *every* load path (`from_toml` and `from_json_value`, via `finalize_deserialized`), so a TOML `publisherId = "12345"` was silently rewritten to the number `12345` before it ever reached `PrebidIntegrationConfig`.

Verified the removal is safe for production paths:

- `Settings::from_toml_and_env` — the only source that ever produced JSON-encoded strings for integration values — is already `#[cfg(test)]`-gated and has no callers outside `settings.rs` tests.
- The live path is `ts config push` → `BlobEnvelope` → `settings_from_config_blob` → `from_json_value`. EdgeZero's app-config env overlay (`edgezero-core` `coerce_env_value`, v0.0.4) coerces each env value to the *existing* TOML leaf's type and errors otherwise, so no untyped strings reach `IntegrationSettings` from the overlay.
- Quoted scalars now surface loudly rather than silently: `validate_settings_for_deploy` → `validate_enabled_integrations` parses every integration at `ts config push` time, so `enabled = "true"` fails there instead of coercing.

Local verification (native host target): `cargo test -p trusted-server-core --lib integrations::prebid` → 143 passed; `--lib settings` → 118 passed; `--lib config` → 99 passed.

Approving — no blocking findings.

## Non-blocking

### 📌 out of scope

- **Env-override docs still advertise the removed behavior**: `docs/guide/configuration.md:142` (`BIDDERS='["kargo","rubicon"]'`) and `docs/guide/configuration.md:1071-1073` (`BID_PARAM_OVERRIDES`, `BID_PARAM_ZONE_OVERRIDES`, `BID_PARAM_OVERRIDE_RULES` as JSON strings) document exactly the JSON-string decoding this PR deletes. The PR removes the identical examples from the rustdoc on `PrebidIntegrationConfig`, so the two now disagree. Acknowledged and tracked in #972 — noting it only so the follow-up isn't lost, since an operator copying those lines today gets a config that no longer applies.

### ♻️ refactor

- **No test locks in the new failure behavior**: the CHANGELOG entry claims "quoted numeric and boolean scalars now fail validation instead of silently converting", but nothing asserts it. The deleted `test_env_var_roundtrip_normalizes_integration_types` was the only test that pinned the old direction, and it went away without a replacement pinning the new one. A small test in `settings.rs` — TOML with `[integrations.testlight] enabled = "true"`, asserting `integration_config::<TestlightConfig>` (or `validate_settings_for_deploy`) returns a configuration error — would keep a future refactor from quietly reintroducing coercion.

### 📝 note

- **Upgrade ordering deserves a CHANGELOG line**: the breaking-change entry tells operators to audit `[integrations.*]`, but the already-pushed `app_config` blob was produced by the *old* CLI and still has the coerced values baked in (`publisherId` stored as a number). Deploying the new WASM without re-running `ts config push` leaves #932 unfixed with no error anywhere. Worth one sentence: re-push config after upgrading.

### 🌱 seedling

- **`from_toml_and_env` is now a test-only fixture with no production analogue** (`crates/trusted-server-core/src/settings.rs:1995`): with `normalize` gone it exercises `config`-crate `Environment` semantics that no runtime path uses, while the real overlay lives in `edgezero-core`. The remaining tests built on it (`test_handlers_override_with_env`, the ec/partner env tests) are now testing a legacy helper rather than shipped behavior. Not for this PR, but the helper looks ready to retire.

### 👍 praise

- **Regression test targets the real path** (`crates/trusted-server-core/src/integrations/prebid.rs:5722`): asserting through `serde_json::to_value` → `from_json_value` mirrors the actual `BlobEnvelope` transition rather than only `from_toml`, and it covers all three override surfaces (static, zone, canonical rule) in one pass. That is the version of this test that would have caught #932.

## CI Status

- fmt: PASS
- clippy: PASS (covered by the per-adapter check/build jobs)
- rust tests: PASS (`cargo test`, axum native, cloudflare native + wasm32-unknown-unknown, spin native + wasm32-wasip1, ts CLI native, cross-adapter parity)
- js tests: PASS (vitest, format-typescript, format-docs)
- `prepare integration artifacts`: still pending at review time — worth a glance before merge.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - **Breaking** — `bid_param_zone_overrides` inner values must now be JSON objects; previously non-object or empty values (`"header" = "x"`, `"header" = {}`) were accepted and silently produced a dead rule at runtime. They now fail at startup with a configuration error. Operators upgrading should audit their `bid_param_zone_overrides` config for non-object zone entries.
+- **Breaking** — Integration configuration strings are no longer globally reinterpreted as JSON scalars. Operators upgrading should audit `[integrations.*]` settings and use native TOML/typed-config booleans and numbers (for example, `enabled = true`, not `enabled = "true"`); quoted numeric and boolean scalars now fail validation instead of silently converting.
```
aram356, line 13: 📝 **note** — Accurate for typed fields, but the sentence "quoted numeric and boolean scalars now fail validation" could be misread as applying to the free-form bidder-param maps too. Those are the opposite: `bid_param_overrides` / `bid_param_zone_overrides` / rule `set` values are `serde_json::Map<String, Json>`, so `publisherId = "12345"` is now **preserved** as a string rather than failing — which is the whole fix. Consider a trailing clause, e.g. "… free-form bidder-param override values are preserved verbatim (a quoted `"12345"` stays a string)." Non-blocking.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -289,11 +289,6 @@ pub struct PrebidIntegrationConfig {
     /// param1 = 12345
     /// param2 = "value"
     /// ```
-    ///
-    /// Example via environment variable:
-    /// ```text
-    /// TRUSTED_SERVER__INTEGRATIONS__PREBID__BID_PARAM_OVERRIDES='{"bidder-name":{"param1":12345,"param2":"value"}}'
-    /// ```
     #[serde(default)]
     pub bid_param_overrides: HashMap<String, serde_json::Map<String, Json>>,
```
aram356, line 293: 📌 **out of scope** — Removing these `TRUSTED_SERVER__INTEGRATIONS__PREBID__BID_PARAM_OVERRIDES='{...}'` doc examples is correct, but equivalent env-var override examples still live in the guide docs and would now mislead operators: `docs/guide/integrations-overview.md:293-308`, `docs/guide/error-reference.md:117-123`, and `trusted-server.example.toml:163` (`TRUSTED_SERVER__CREATIVE_OPPORTUNITIES__SLOT='[{…}]'`, an array override EdgeZero's overlay rejects). Worth a follow-up issue rather than expanding this PR.

aram356 (approved):
## Summary

Re-review of the reworked PR. All findings from the prior `CHANGES_REQUESTED` review are resolved:

- The global JSON re-interpretation (`normalize` / `normalize_env_value`) is deleted, not shimmed; no production path called it.
- The regression test now runs the full production shape — TOML → `Settings` → runtime JSON → typed `PrebidIntegrationConfig` → compiled-rule OpenRTB emission — and asserts numeric-looking strings survive static, zone, and canonical-rule merges. Ported onto `main` it fails with `Number(12345)` vs `String("12345")`, so it is a genuine regression test.
- CHANGELOG documents the breaking config change.
- Deleting the two datadome env tests is not a coverage loss — `protection_scope.rs` retains 8 dedicated tests.

Approving. Two non-blocking notes inline.

## CI Status

Locally reproduced against a worktree of the head commit:

- `cargo fmt --all -- --check`: PASS
- `cargo clippy-axum`: PASS
- `cargo test -p trusted-server-core`: PASS (1729 passed, 0 failed)

GitHub checks green except two still-running at review time (`cargo test (axum native)`, `prepare integration artifacts`).

PR #934, opened 2026-07-20
Fix ext.trusted_server.request_host to use the publisher domain

### LINKED ISSUE #888: Fix ext request from prebid server to send correct request host
### Description

```
    "trusted_server": {
      "kid": "ts-2026-04-30",
      "request_host": "ts.example.com",
      "request_scheme": "https",
      "signature": "...",
      "ts": 1783705878272,
      "version": "1.1"
    }
```

prk-Jr:
## Summary

- `ext.trusted_server.request_host` on outbound Prebid Server auction requests was set from the raw incoming `Host` header, which on the SSAT proxy path is the trusted-server edge domain (e.g. `ts.example.com`), not the publisher's own domain.
- This is now sourced from `AuctionRequest.publisher.domain` (scheme fixed to `https`), matching how `site.domain`/`publisher.domain` are already built, and matching what PBS's `trusted_server` verification module expects ("Host header value expected by the receiving service").

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/integrations/prebid.rs` | `PrebidAuctionProvider::request_bids` now builds `request_info` from `request.publisher.domain` instead of `RequestInfo::from_request` (raw edge `Host` header). Added a regression test asserting `ext.trusted_server.request_host` matches the publisher domain, not the edge host. |
| `crates/trusted-server-core/src/platform/test_support.rs` | `StubHttpClient::send_async` now captures the outgoing request body (mirroring `send()`), so tests exercising the async fan-out path can assert on outbound bodies via `recorded_request_bodies()`. |

## Closes

Closes #888

## Test plan

- [x] `cargo test-fastly && cargo test-axum && cargo test-cloudflare`
- [x] WASM build: `cargo check-fastly` (wasm32-wasip1)
- [ ] Other:

## Checklist

- [x] Uses `tracing`/`log` macros (not `println!`)
- [x] New code has tests

### DIFF
--- a/crates/trusted-server-core/src/integrations/prebid.rs
+++ b/crates/trusted-server-core/src/integrations/prebid.rs
@@ -2229,8 +2229,17 @@ impl AuctionProvider for PrebidAuctionProvider {
     ) -> Result<PlatformPendingRequest, Report<TrustedServerError>> {
         log::info!("Prebid: requesting bids for {} slots", request.slots.len());
 
-        let request_info =
-            RequestInfo::from_request(context.request, context.services.client_info());
+        // `ext.trusted_server.request_host` must be the publisher's own domain
+        // (matching `site.domain`/`publisher.domain` below), not the raw edge
+        // `Host` header of the incoming `/auction` request. On the SSAT proxy
+        // path the browser calls `/auction` against the trusted-server edge
+        // domain (e.g. `ts.example.com`), which is not what PBS's
+        // `trusted_server` verification module expects — see the "Host header
+        // value expected by the receiving service" doc on `SigningParams`.
+        let request_info = RequestInfo {
+            host: request.publisher.domain.clone(),
+            scheme: "https".to_owned(),
+        };
 
         // Create signer and compute signature if request signing is enabled
         let signer_with_signature =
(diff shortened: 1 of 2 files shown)


crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -2229,8 +2229,17 @@ impl AuctionProvider for PrebidAuctionProvider {
...
+        // `ext.trusted_server.request_host` must be the publisher's own domain
+        // (matching `site.domain`/`publisher.domain` below), not the raw edge
+        // `Host` header of the incoming `/auction` request. On the SSAT proxy
+        // path the browser calls `/auction` against the trusted-server edge
+        // domain (e.g. `ts.example.com`), which is not what PBS's
+        // `trusted_server` verification module expects — see the "Host header
+        // value expected by the receiving service" doc on `SigningParams`.
+        let request_info = RequestInfo {
+            host: request.publisher.domain.clone(),
+            scheme: "https".to_owned(),
```
aram356, line 2246: 📝 **note** — The scheme is now hardcoded to `"https"`, replacing the previous `RequestInfo::from_request` TLS detection. This also feeds `SigningParams` and `ext.trusted_server.request_scheme`. Defensible (the publisher origin is https in production), but the comment above justifies only the *host* — consider a one-line note that the scheme is intentionally canonical-https, not request-derived. Confirm PBS's `trusted_server` module verifies https unconditionally (matters for local `fastly compute serve` over http). Non-blocking.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -2229,8 +2229,17 @@ impl AuctionProvider for PrebidAuctionProvider {
...
-            RequestInfo::from_request(context.request, context.services.client_info());
+        // `ext.trusted_server.request_host` must be the publisher's own domain
+        // (matching `site.domain`/`publisher.domain` below), not the raw edge
+        // `Host` header of the incoming `/auction` request. On the SSAT proxy
+        // path the browser calls `/auction` against the trusted-server edge
+        // domain (e.g. `ts.example.com`), which is not what PBS's
+        // `trusted_server` verification module expects — see the "Host header
+        // value expected by the receiving service" doc on `SigningParams`.
+        let request_info = RequestInfo {
+            host: request.publisher.domain.clone(),
```
aram356, line 2245: 📝 **note** — This is correct wherever `request.publisher.domain` is the publisher's own domain — true on the `/auction` path (`convert_tsjs_to_auction_request` uses validated `settings.publisher.domain`). On the SSAT navigation path, `build_auction_request` (`publisher.rs`) still sets `publisher.domain` from the edge `Host`, so `request_host` would remain the edge host there. That's fixed at the source by #937 (which also corrects `site.domain`/`site.page`/`page_url`). The two compose cleanly — no action needed here.

aram356 (approved):
## Summary

Correct, well-scoped change. `ext.trusted_server.request_host` (and the signed host in `SigningParams`) now comes from the validated `publisher.domain` instead of the raw incoming `Host` header, which on the SSAT proxy path is the trusted-server edge host — not what PBS's `trusted_server` verification module expects. CI is fully green and the regression test is meaningful. Approving.

## Notes

### 📝 Scope: correct on the `/auction` path; navigation path is covered by #937

The fix is correct wherever `request.publisher.domain` is the publisher's own domain — which holds on the `/auction` endpoint path, where `convert_tsjs_to_auction_request` sources it from the validated `settings.publisher.domain`.

On the SSAT navigation path, `build_auction_request` (`crates/trusted-server-core/src/publisher.rs`) still populates `publisher.domain`/`site.domain` from `request_info.host` (the edge host), so on that path `request_host` would remain the edge host. That builder is fixed at the source by **#937**, which also corrects `site.domain`/`site.page`/`page_url` (untouched here). The two compose cleanly: because this PR reads `publisher.domain` at runtime, #937 making that field correct on the navigation path makes this PR's `request_host` correct there too. No ordering dependency, no conflict.

### 📝 Scheme is now canonical `https`, not request-derived

`request_scheme` (and the signed scheme) is now hardcoded to `"https"` rather than detected via TLS. Defensible — the publisher origin is https in production — but it's a behavior change the code comment doesn't call out (the comment justifies only the host). Worth a one-line note that the scheme is intentionally canonical, and confirming PBS verifies https unconditionally (relevant for local `fastly compute serve` over http). Non-blocking.

### 👍 Praise

- `StubHttpClient::send_async` body capture mirrors `send()` and propagates capture failures via `?` instead of silently recording an empty body — a capture failure can't masquerade as an intentionally empty body.
- The regression test uses a distinct edge host (`ts.pub.example`) vs publisher domain (`pub.example`), so the assertion actually catches the leak rather than passing vacuously; `.example` domains throughout, per project guidelines.

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests (fastly / axum / cloudflare / spin): PASS
- cross-adapter parity: PASS
- js (vitest) + format: PASS
- integration / browser: PASS

PR #937, opened 2026-07-20
Use configured publisher domain for navigation-path auction requests

### LINKED ISSUE #936: Injected creatives + IAS brand-safety pixel carry the TS edge host instead of the publisher domain
## Summary

On the SSAT inline-creative injection path, the server-side auction builds its OpenRTB `Site` from the **raw incoming `Host` header** (the Trusted Server edge domain, e.g. `ts.example.com`) instead of the configured publisher domain (`www.example.com`). As a result, every injected creative and the IAS brand-safety pixel carry the staging edge host, not the real publisher host.

The `/auction` endpoint path is unaffected — it already sources the host from config — which is why an `/auction` request capture looks correct while injected creatives do not.

## Root cause

There are two `AuctionRequest` builders that populate `publisher.domain` / `site.domain` from opposite sources:

- **Correct** — `convert_tsjs_to_auction_request` (`crates/trusted-server-core/src/auction/formats.rs:235` and `:245`) uses `settings.publisher.domain` (from `[publisher] domain` config). Used by the `/auction` endpoint.
- **Wrong** — `build_auction_request` (`crates/trusted-server-core/src/publisher.rs:3037` and `:3051`) uses `request_info.host`, the raw incoming `Host` header. On the SSAT proxy model that is the TS edge domain.

The wrong builder is called on both navigation paths:

- `AuctionSource::InitialNavigation` — `crates/trusted-server-core/src/publisher.rs:2626`
- `AuctionSource::SpaNavigation` — `crates/trusted-server-core/src/publisher.rs:3509`

Downstream, `PrebidAuctionProvider::request_bids` builds the OpenRTB `Site` directly from these fields (`crates/trusted-server-core/src/integrations/prebid.rs:1812-1818`), so PBS and any IAS brand-safety integration keyed on `site.domain` / `site.page` receive the edge host and reflect it into creative markup and the pixel URL.

## Relationship to PR #934

PR #934 ("Fix ext.trusted_server.request_host to use the publisher domain", commit `5d7f696c2`) only changed `ext.trusted_server.request_host` inside the Prebid provider (`crates/trusted-server-core/src/integrations/prebid.rs:2274`) to read from `request.publisher.domain`. That does **not** fix this path, because:

- On the navigation path, `request.publisher.domain` is itself the poisoned edge host (set by `build_auction_request`).
- #934 never touched `site.domain` / `site.publisher.domain` / `site.page`.

So the leak into injected creatives + the IAS pixel remains.

Note: PR #916 (optional creative rewriting) is unrelated — that rewriter only makes asset URLs relative and is not the source of the host.

## Suggested fix

Make `build_auction_request` (`crates/trusted-server-core/src/publisher.rs:3012`) source `publisher.domain`, `site.domain`, and `page_url` from `settings.publisher.domain` — the way `convert_tsjs_to_auction_request` already does — rather than from `request_info.host`.

## Affected branch

Found on the `rc/july` release-candidate branch (PR #919).

aram356:
## Summary

On the navigation / SSAT proxy path, the server-side auction advertised the **incoming request `Host` header** — which on that path is the trusted-server edge host (e.g. the staging domain) — as the publisher domain. That value flowed into the outbound OpenRTB bid request and, through it, into injected creatives and the IAS brand-safety pixel, so both carried the edge/staging host instead of the real publisher domain.

`build_auction_request` set `publisher.domain`, `site.domain`, and the `page_url` host from `request_info.host`. This change sources those fields from `settings.publisher.domain` instead — matching what `convert_tsjs_to_auction_request` already does on the `/auction` endpoint path.

Closes #936.

## Root cause

Two `AuctionRequest` builders derived the publisher host from opposite sources:

- `convert_tsjs_to_auction_request` (`auction/formats.rs`) — used config (`settings.publisher.domain`). ✅
- `build_auction_request` (`publisher.rs`) — used the raw incoming `Host` header. ❌ (navigation path: `AuctionSource::InitialNavigation` and `SpaNavigation`)

PR #934 only corrected `ext.trusted_server.request_host` inside the Prebid provider; it did not touch `site.domain` / `site.publisher.domain` / `site.page`, and on the navigation path `request.publisher.domain` was itself the poisoned edge host. This fixes the builder at the source.

## Changes

- Add a `publisher_domain: &str` parameter to `build_auction_request` and use it for `publisher.domain`, `site.domain`, and the `page_url` host (scheme and path are unchanged).
- Both call sites pass `&settings.publisher.domain`.
- New regression test `auction_request_uses_configured_publisher_domain_not_edge_host`: with a divergent edge host (`ts.example.com`) and configured domain (`www.example.com`), the bid request advertises the configured domain for `publisher.domain`, `site.domain`, `page_url`, and `site.page`. Written test-first (verified it fails on the old behavior before the fix).

## Test plan

- [x] `cargo test-fastly` (wasm — core + fastly, incl. new regression test)
- [x] `cargo test-axum` (native — shared core)
- [ ] `cargo clippy-*` — could not run locally: `clippy-fastly` fails in the `trusted-server-js` build script (`Unrecognized option: 'p'`), which reproduces on unmodified `main`, so it is a pre-existing local-environment issue unrelated to this change. Deferred to CI.

## Notes

Targets `main` (the buggy code exists on `main`, not just the release branch). Can be cherry-picked / merged into `rc/july` as needed.

### DIFF
--- a/crates/trusted-server-core/src/publisher.rs
+++ b/crates/trusted-server-core/src/publisher.rs
@@ -1627,6 +1627,7 @@ pub async fn handle_publisher_request(
                 ec_id,
                 &consent_context,
                 &request_info,
+                &settings.publisher.domain,
                 req.headers()
                     .get("user-agent")
                     .and_then(|v| v.to_str().ok()),
@@ -2007,6 +2008,7 @@ pub(crate) fn build_auction_request(
     ec_id: Option<&str>,
     consent_context: &crate::consent::ConsentContext,
     request_info: &crate::http_util::RequestInfo,
+    publisher_domain: &str,
     user_agent: Option<&str>,
 ) -> AuctionRequest {
     let slots = slots_ctx
@@ -2014,9 +2016,13 @@ pub(crate) fn build_auction_request(
         .iter()
         .map(crate::creative_opportunities::CreativeOpportunitySlot::to_ad_slot)
         .collect();
+    // Advertise the configured publisher domain (not the incoming edge `Host`)
+    // so SSPs, injected creatives, and brand-safety pixels see the publisher's
+    // own origin. On the SSAT proxy path `request_info.host` is the trusted
+    // server edge host, which must not leak into the bid request.
     let page_url = format!(
         "{}://{}{}",
-        request_info.scheme, request_info.host, slots_ctx.request_path
+        request_info.scheme, publisher_domain, slots_ctx.request_path
     );
     let ec_id = ec_id.filter(|id| !id.is_empty());
     let request_id = ec_id.map_or_else(
@@ -2027,7 +2033,7 @@ pub(crate) fn build_auction_request(
         id: request_id,
         slots,
         publisher: PublisherInfo {
-            domain: request_info.host.clone(),
+            domain: publisher_domain.to_owned(),
             page_url: Some(page_url.clone()),
         },
         user: UserInfo {
@@ -2041,7 +2047,7 @@ pub(crate) fn build_auction_request(
             geo: None,
         }),
         site: Some(SiteInfo {
-            domain: request_info.host.clone(),
+            domain: publisher_domain.to_owned(),
             page: page_url,
         }),
         context: std::collections::HashMap::new(),
@@ -2446,6 +2452,7 @@ pub async fn handle_page_bids(
                 ec_id,
                 consent_context,
                 &request_info,
+                &settings.publisher.domain,
                 req.headers()
                     .get("user-agent")
                     .and_then(|v| v.to_str().ok()),
@@ -4713,6 +4720,7 @@ mod tests {
                 None,
                 &ConsentContext::default(),
                 &request_info,
+                "publisher.example.com",
                 Some("Mozilla/5.0"),
             );
 
@@ -4721,6 +4729,52 @@ mod tests {
                 request.id.starts_with("ts-req-"),
                 "should use a non-EC request id, got {}",
                 request.id
+            );
+        }
+
+        #[test]
+        fn auction_request_uses_configured_publisher_domain_not_edge_host() {
+            // On the SSAT proxy path the browser addresses the trusted-server
+            // edge host, but the auction must advertise the configured
+            // publisher domain to SSPs — otherwise injected creatives and the
+            // brand-safety pixel leak the edge/staging host.
+            let slot = make_slot();
+            let slots = [slot];
+            let slots_ctx = MatchedSlotsContext {
+                matched_slots: &slots,
+                request_path: "/2024/01/my-article/?edition=fictional",
+            };
+            let request_info = RequestInfo {
+                host: "ts.example.com".to_string(),
+                scheme: "https".to_string(),
+            };
+
+            let request = build_auction_request(
+                &slots_ctx,
+                None,
+                &ConsentContext::default(),
+                &request_info,
+                "www.example.com",
+                Some("Mozilla/5.0"),
+            );
+
+            assert_eq!(
+                request.publisher.domain, "www.example.com",
+                "publisher.domain should be the configured publisher domain, not the edge host"
+            );
+            let site = request.site.expect("should populate site metadata");
+            assert_eq!(
+                site.domain, "www.example.com",
+                "site.domain should be the configured publisher domain, not the edge host"
+            );
+            assert_eq!(
+                request.publisher.page_url.as_deref(),
+                Some("https://www.example.com/2024/01/my-article/?edition=fictional"),
+                "page_url host should be the configured publisher domain, not the edge host"
+            );
+            assert_eq!(
+                site.page, "https://www.example.com/2024/01/my-article/?edition=fictional",
+                "site.page host should be the configured publisher domain, not the edge host"
             );
         }
 
@@ -4742,6 +4796,7 @@ mod tests {
                 Some("ec-abc"),
                 &ConsentContext::default(),
                 &request_info,
+                "publisher.example.com",
                 Some("Mozilla/5.0"),
             );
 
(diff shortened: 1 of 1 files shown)


prk-Jr (approved):
## Summary

Fixes the navigation-path (`SSAT`/SPA) `AuctionRequest` builder so `publisher.domain`, `site.domain`, and the page URL host come from the configured `settings.publisher.domain` instead of the incoming edge `Host` header, matching the pattern already used by `convert_tsjs_to_auction_request` on the `/auction` path. Minimal, well-targeted diff (`publisher.rs` only) with a solid test-first regression test.

## Non-blocking

### 🌱 seedling

- **Scheme source differs between the two `AuctionRequest` builders**: `build_auction_request` (`publisher.rs:2021`) still derives `page_url`'s scheme from `request_info.scheme` (edge-detected), while `convert_tsjs_to_auction_request` (`auction/formats.rs:202`) hardcodes `"https"` for the same field. Not a bug today (the edge always terminates TLS as `https` in production), but worth aligning in a follow-up so both builders source scheme the same way.

### 👍 praise

- **Test-first regression coverage**: `auction_request_uses_configured_publisher_domain_not_edge_host` uses a deliberately divergent edge host (`ts.example.com`) vs. configured domain (`www.example.com`) and asserts all four leak points (`publisher.domain`, `site.domain`, `page_url`, `site.page`). Confirmed to fail on the pre-fix code per the PR description.

## CI Status

- fmt: PASS
- clippy: not reported as a distinct check (cargo check jobs pass)
- rust tests: PASS (fastly/axum/cloudflare/spin, cross-adapter parity)
- js tests (vitest): PASS
- integration tests: PASS

crates/trusted-server-core/src/publisher.rs
```diff
@@ -1627,6 +1627,7 @@ pub async fn handle_publisher_request(
                 ec_id,
                 &consent_context,
                 &request_info,
+                &settings.publisher.domain,
```
ChristianPavilonis, line 1635: 🔧 **P1 / High — Navigation telemetry still records the edge host as the publisher domain**

The auction request now correctly receives `settings.publisher.domain` here, but the `AuctionObservationContext::from_parts` call immediately above still receives `request_host`. The SPA path has the same mismatch at lines 2438–2440, where it passes `request_info.host`. On the affected routing model those values are the edge/staging host this PR is removing from the auction request.

`AuctionObservationContext` copies the value into `publisher_domain`, which Tinybird uses for sorting, rollups, dashboard filters, and ingestion-freshness reporting. Navigation metrics will therefore remain attributed to the edge host while `/auction` metrics and outbound bids use the configured publisher domain.

**Suggested fix:** Pass `&settings.publisher.domain` to both observation constructors and add coverage with divergent request/configured hosts that asserts the emitted event rows use the configured publisher domain.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -4724,6 +4732,52 @@ mod tests {
             );
         }
 
+        #[test]
+        fn auction_request_uses_configured_publisher_domain_not_edge_host() {
```
ChristianPavilonis, line 4743: 🌱 **P2 / Medium — The regression test does not exercise either production call path**

This test invokes `build_auction_request` directly and supplies `"www.example.com"` itself. It would continue passing if either newly changed handler reverted to passing `request_info.host`, because both sources have the same `&str` type.

The original defect was a source-selection error between builders and request paths, so helper-only coverage proves field assignment but not that initial and SPA navigation choose the correct source.

**Suggested fix:** Add handler-level tests for initial navigation and `/__ts/page-bids` with divergent edge/configured hosts, capturing the dispatched auction or OpenRTB request. Those tests can also cover the telemetry inconsistency noted above.

ChristianPavilonis (approved):
## Summary

The core OpenRTB/APS domain correction is narrow and correct. Approving with two follow-up findings noted inline: navigation telemetry still uses the edge host as `publisher_domain`, and the new regression test covers the helper rather than the production handlers.

crates/trusted-server-core/src/publisher.rs
aram356: Confirmed and fixed in 4a020e2f1. Both `AuctionObservationContext::from_parts` call sites (initial navigation and `/__ts/page-bids`) now pass `&settings.publisher.domain`, so navigation telemetry is attributed to the same publisher identity as the outbound bid request and as the `/auction` rows built via `from_auction_request`.

Covered by the new handler-level tests below, which assert every emitted telemetry row carries the configured domain. Verified they fail with `publisher_domain: "ts.example.com"` when the observation argument is reverted.

crates/trusted-server-core/src/publisher.rs
aram356: Right — the helper test pins field assignment, not source selection. Added `navigation_publisher_domain_tests` in 4a020e2f1 with two handler-level tests:

- `initial_navigation_advertises_configured_publisher_domain` — drives `handle_publisher_request`
- `page_bids_advertises_configured_publisher_domain` — drives `handle_page_bids`

Both use a divergent edge host (`ts.example.com`) against the configured `test-publisher.com`, capture the dispatched `AuctionRequest` via a recording `AuctionProvider` (same pattern as `EidCapturingProvider` in `auction/endpoints.rs`), and record telemetry through a `RecordingTelemetrySink`. They assert `publisher.domain`, `site.domain`, `page_url` host, `site.page`, and every emitted telemetry row.

Verified both fail if either call site reverts:
- builder argument reverted to `&request_info.host` → `publisher.domain` assertion fails with `ts.example.com`
- observation argument reverted → telemetry `publisher_domain` assertion fails with `ts.example.com`

One note: `page_url` is asserted on the host component only, not the full URL, because the two builders still disagree on scheme (edge-detected here vs. hardcoded `https` in `convert_tsjs_to_auction_request`) — the seedling prk-Jr raised. Left as the separate follow-up they suggested rather than folded in here.

PR #316, opened 2026-02-16
Remove immutable cache headers from integrations

ChristianPavilonis:
Removes immutable cache headers from integrations
- Lockr
- Permutive
- Prebid

Closes #309, #310

PR #940, opened 2026-07-20
Upgrade EdgeZero to v0.0.7 and complete staging runtime integration

### LINKED ISSUE #939: Upgrade edgezero dependencies to the deploy-actions branch (PR #316)
### Summary

The six edgezero workspace dependencies are pinned to `tag = "v0.0.4"` in the root `Cargo.toml`. Upgrade them to the [`feature/edgezero-deploy-actions`](https://github.com/stackpop/edgezero/pull/316) branch to pick up the layered deploy actions and Fastly staging lifecycle, and adapt Trusted Server to the accompanying API changes.

Successor to #861 (which pinned the deps to v0.0.4).

### Affected dependencies

- `edgezero-adapter-axum`
- `edgezero-adapter-cloudflare`
- `edgezero-adapter-fastly`
- `edgezero-adapter-spin`
- `edgezero-cli`
- `edgezero-core`

### Changes

- Repoint the six edgezero deps from `tag = "v0.0.4"` to `branch = "feature/edgezero-deploy-actions"` and refresh `Cargo.lock`.
- Wire the new `ts` CLI subcommands surfaced by `edgezero-cli`: `active-version`, `healthcheck`, `rollback`, plus `deploy --stage` and a top-level `--version` flag.
- Migrate `TrustedServerAppConfig` to the `AppConfigMeta::secret_fields()` method that replaces the removed `SECRET_FIELDS` associated constant.

### Verification

- `cargo clippy` across all adapter targets (fastly, axum, cloudflare native/wasm, spin native/wasm)
- Adapter + core + CLI test suites, and integration `parity`
- JS build/test/format

### Follow-ups

- **Blocked on upstream:** stackpop/edgezero#316 is not yet merged. The `branch =` ref is a moving target; before this can merge, re-pin to a tagged release (or `rev`) once #316 lands.
- Adopting the new deploy GitHub Actions (`deploy-fastly` / `healthcheck-fastly` / `rollback-fastly`) in this repo's workflows is separate, unstarted work.

### LINKED ISSUE #1082: Staged deploys load production config: resolve config selectors from edgezero_runtime_env
## Problem

Found in #940 review: the Fastly staging lifecycle relinks the version-linked `edgezero_runtime_env` Config Store and redirects `EDGEZERO__STORES__CONFIG__TRUSTED_SERVER_CONFIG__KEY` to `trusted_server_config_staging`. Trusted Server's Fastly entry point bypasses `edgezero_adapter_fastly::run_app` and resolves its config selectors via `EnvConfig::from_env()` (`crates/trusted-server-core/src/settings_data.rs:33-41`), which is empty on Compute.

A staged version therefore silently loads the **production** config blob: `ts config push --staging` writes a key nothing reads, and `ts healthcheck --staging` exercises the new binary against production config, which can false-pass an incompatible staged code/config pairing. Documented as a known limitation in `docs/guide/cli.md`.

## Fix

Blocked on stackpop/edgezero#349, which exposes edgezero's runtime-env loader as public API (the store name, key spellings, and derivation rules are edgezero internals; replicating them here would drift).

Once it lands:

1. Build the `EnvConfig` from the exposed API in the Fastly adapter before constructing `TrustedServerApp`.
2. Resolve both the physical config-store name and the blob key through it (parameterize `settings_data`'s selector helpers on `EnvConfig`).
3. Add an adapter test proving staging selects `trusted_server_config_staging` while production selects `trusted_server_config`.
4. Remove the known-limitation callout from `docs/guide/cli.md`.

aram356:
## Summary

Upgrades all six EdgeZero workspace dependencies from `v0.0.4` to `v0.0.7` and completes the Trusted Server integration for the full release delta.

- Pin `edgezero-adapter-{axum,cloudflare,fastly,spin}`, `edgezero-cli`, and `edgezero-core` to `tag = "v0.0.7"` (`5c9886e5`). The lockfile update is scoped to the EdgeZero tree; the required non-EdgeZero additions are `syn 3` and `toml_edit 0.25`.
- Declare Trusted Server's `trusted_server_config` store metadata and add `EnvConfig`-aware store-name/key selectors while retaining the existing native default helpers.
- Load `runtime_env_config(TrustedServerApp::stores())` once in the Fastly entry point and reuse it for the JA4 settings read, physical store open, application construction, and both fallback settings reloads.
- Production therefore reads `trusted_server_config` from the `trusted_server_config` physical store, while staged versions can select `trusted_server_config_staging` through the version-linked `edgezero_runtime_env` store. Physical store-name overrides remain independent of key selection.
- Migrate `TrustedServerAppConfig` to the v0.0.7 `AppConfigMeta::secret_fields()` API.
- Remove the obsolete staging limitation from the docs and document capture-before-deploy rollback guidance.

## v0.0.4 to v0.0.7 audit

The upstream functional changes were audited through the v0.0.7 tag:

- App-config metadata/secret-field API changes are adopted. Trusted Server deliberately keeps its existing inline app-config blob model; moving application secrets into platform secret stores remains separate operator-migration work.
- Config-store chunk GC is exposed as `ts config gc`, including preview/destructive gates, age filtering, `--store`, and `--no-env` targeting behavior.
- Fastly staged deploy, healthcheck, rollback, and active-version lifecycle commands are exposed, with staging config push/diff flags and fail-closed deploy passthrough parsing.
- The v0.0.7 runtime selector API is wired into Trusted Server's custom Fastly entry point, closing the prior staging-config gap.
- Rollback documentation no longer implies that the previously active production version can be reconstructed after deployment.
- Dependency/tooling-only upstream changes were covered by the complete Rust, adapter, JavaScript, documentation, and integration matrix below.

## CLI coverage

Every downstream operational EdgeZero handler is exposed by `ts`:

| `ts` surface      | EdgeZero handler                                      |
| ----------------- | ----------------------------------------------------- |
| `active-version`  | `run_active_version`                                  |
| `auth`            | `run_auth`                                            |
| `build`           | `run_build`                                           |
| `config diff`     | `run_config_diff_typed::<TrustedServerAppConfig>`     |
| `config gc`       | `run_config_gc`                                       |
| `config push`     | `run_config_push_typed::<TrustedServerAppConfig>`     |
| `config validate` | `run_config_validate_typed::<TrustedServerAppConfig>` |
| `deploy`          | `run_deploy`                                          |
| `healthcheck`     | `run_healthcheck`                                     |
| `provision`       | `run_provision`                                       |
| `rollback`        | `run_rollback`                                        |
| `serve`           | `run_serve`                                           |

Trusted Server's own `audit`, `config init`, `prebid`, and `dev` commands remain intact. `ts new` is intentionally not included. EdgeZero's feature-gated contributor-only demo server is not a downstream CLI operation and is not enabled.

Argument coverage includes the remaining proof gaps from the v0.0.4→v0.0.7 audit: `healthcheck --path`, `config gc --store`, and `config gc --no-env`, in addition to retry/staging/rollback/destructive-GC behavior.

## Verification

Self-review and an independent code review found no remaining critical, important, or minor issues. The final branch passes:

- Locked dependency metadata, narrow lockfile audit, `cargo fmt --all -- --check`, and `git diff --check`.
- Clippy for Fastly, Axum, Cloudflare native/wasm, Spin native/wasm, CLI, codegen, and the integration manifest.
- Rust tests for Fastly/core, Axum, Cloudflare, Spin, CLI, cross-adapter parity, codegen, and integration crates. The final `cargo test-fastly` run passed 166 adapter tests, 2,257 core tests (6 expected ignores), 2 JS bridge tests, 21 OpenRTB tests, and 3 doctests (4 expected ignores).
- Adapter build/check gates for Fastly release wasm, Axum, Cloudflare native/wasm, and Spin native/release wasm, plus benchmark smoke coverage.
- JavaScript build, 893 Vitest tests, type checking, lint, and formatting; documentation lint, formatting, and build.
- Template-cache local harnesses: ESI 21/21 and inline 8/8. The harness's auction/provider budgets were raised above its deliberate three-second bid delay to remove an invalid boundary race exposed by this verification run.
- All GitHub checks on the final commit, including Rust/JS/Actions CodeQL, formatting, all adapter/CLI/parity tests, and core/browser/Fastly-EC integration suites.

## Explicit exclusions

- No `ts new` command.
- No deploy-action workflow adoption or other `.github/workflows` changes.
- No new pull request; this continues PR #940.

Closes #939
Closes #1082


aram356:
### Status update — P0/P1 tracking

**Head:** \`96dd1f72\` (main merged in twice since the original commit; edgezero dep unchanged, no conflicts).

#### P0.2 — CI + output-line contract
- **Local verification at head is green:** fmt, clippy (fastly), core+fastly tests (1648+99), vitest (411). The earlier red checks were a systemic/environmental run on \`96dd1f72\` (every job failed, incl. JS/fmt that pass locally) — fresh reruns of all four workflows are in progress.
- **Machine-readable output lines the deployer parses are verified end-to-end.** All three are emitted by edgezero-cli at \`log::info!\` (Info→stdout), and \`ts\`'s \`main.rs\` calls \`edgezero_cli::init_cli_logger()\`, so they reach stdout rather than being swallowed:
  | Line | Source (edgezero @ 145f1699) |
  |------|------|
  | \`version=<N>\` | \`edgezero-cli/src/lib.rs:253\` |
  | \`pushed-key=<key>\` | \`edgezero-cli/src/config.rs:386\` |
  | \`rolled-back-to=<N>\` | \`edgezero-adapter-fastly/src/cli.rs:2807\` |
- Dispatcher/arg-parse tests for \`active-version\`, \`healthcheck\`, \`rollback\` are in \`crates/trusted-server-cli/src/run.rs\`.

#### P1.3 — edgezero library ↔ action lockstep (recorded pair)
The \`ts\` CLI is built from this repo's \`Cargo.toml\` (edgezero **library** rev), and the deployer separately pins \`stackpop/edgezero/.github/actions/deploy-fastly@<ref>\` (edgezero **action** ref). These MUST be the same edgezero commit so library and action can't diverge.

- **Current pair:** both = \`stackpop/edgezero@145f1699\` (branch \`feature/edgezero-deploy-actions\`).
- **On merge:** when edgezero #316 lands and is tagged, move BOTH the six \`Cargo.toml\` deps and the deployer's \`deploy-fastly@<ref>\` to that same tag/SHA in lockstep.

#### P0.1 — re-pin off the moving branch (blocker, still open)
edgezero #316 is **not yet merged** and no new tag exists. Cannot re-pin until it lands; PR stays draft until then.

Cargo.toml
```diff
@@ -54,12 +54,12 @@ criterion = { version = "0.5", default-features = false, features = ["cargo_benc
...
-edgezero-adapter-spin = { git = "https://github.com/stackpop/edgezero", tag = "v0.0.4", default-features = false }
-edgezero-cli = { git = "https://github.com/stackpop/edgezero", tag = "v0.0.4" }
-edgezero-core = { git = "https://github.com/stackpop/edgezero", tag = "v0.0.4", default-features = false }
+edgezero-adapter-axum = { git = "https://github.com/stackpop/edgezero", branch = "feature/edgezero-deploy-actions", default-features = false }
+edgezero-adapter-cloudflare = { git = "https://github.com/stackpop/edgezero", branch = "feature/edgezero-deploy-actions", default-features = false }
+edgezero-adapter-fastly = { git = "https://github.com/stackpop/edgezero", branch = "feature/edgezero-deploy-actions", default-features = false }
+edgezero-adapter-spin = { git = "https://github.com/stackpop/edgezero", branch = "feature/edgezero-deploy-actions", default-features = false }
+edgezero-cli = { git = "https://github.com/stackpop/edgezero", branch = "feature/edgezero-deploy-actions" }
+edgezero-core = { git = "https://github.com/stackpop/edgezero", branch = "feature/edgezero-deploy-actions", default-features = false }
```
prk-Jr: 🔧 **wrench** — Six deps now point at a live branch of an unmerged upstream PR, and this PR is no longer a draft.

Verified against the upstream API: `stackpop/edgezero#316` is `state: open`, `merged: false`. This PR reports `isDraft: false` and `mergeable: MERGEABLE`, so nothing mechanically prevents merging a `branch =` ref into `main`. `Cargo.lock` currently pins `5f3d648c`, but that pin is only as durable as the lockfile — any `cargo update`, or a `cargo install`/fresh resolve that does not pass `--locked`, silently retargets whatever the branch tip is at that moment. A force-push or branch deletion upstream breaks the fetch outright.

The PR body already flags this, but the PR state doesn't match the caveat.

**Fix** — either flip back to draft until #316 merges and a tag is cut, or pin immutably in the meantime:

```toml
edgezero-core = { git = "https://github.com/stackpop/edgezero", rev = "5f3d648c3c6c38fc6e6b22b5c65c66177363aad8", default-features = false }
```

A `rev` gives the same code as today's lock while making the reference immutable, so the repin-before-merge step becomes `rev` → `tag` rather than "remember not to merge this."

Cargo.lock
```diff
@@ -767,7 +767,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.48.0",
```
prk-Jr: 🤔 **thinking** — The lock refresh carries downgrades unrelated to the edgezero repin:

- `windows-sys` `0.61.2` → `0.48.0` (2 sites)
- `itertools` `0.13.0` → `0.10.5` (3 `prost-*` crates)

Neither is edgezero-adjacent. This pattern usually means the lockfile was re-resolved wholesale rather than repinned in place — possibly under a different toolchain or resolver setting. Both crates are host-build-only (Windows shims, protobuf codegen), so there is no `wasm32-wasip1` runtime exposure and CI is green, but it widens the diff's blast radius beyond "upgrade EdgeZero" and makes the lock harder to audit.

**Fix** — scope the update to the intended crates and drop the incidental deltas:

```bash
cargo update -p edgezero-core -p edgezero-cli -p edgezero-adapter-fastly \
             -p edgezero-adapter-axum -p edgezero-adapter-cloudflare -p edgezero-adapter-spin
```

If the downgrades turn out to be forced by the new edgezero tree (e.g. via `edgezero-macros`' move to `syn 3`), a note in the PR body would settle it.

crates/trusted-server-core/src/config.rs
```diff
@@ -112,7 +112,9 @@ impl edgezero_core::app_config::AppConfigMeta for TrustedServerAppConfig {
     // app-config blob. Migrating app-level secrets to `EdgeZero` secret-store
     // references needs nested/array extraction support and operator migration
     // work tracked separately.
-    const SECRET_FIELDS: &'static [edgezero_core::app_config::SecretField] = &[];
+    fn secret_fields() -> Vec<edgezero_core::app_config::SecretField> {
```
prk-Jr, line 116: 📝 **note** — Straight port of the removed `SECRET_FIELDS` associated const to the new `secret_fields()` method; `Vec::new()` doesn't allocate, so the const → owned-`Vec` signature change costs nothing here.

Worth restating for anyone reading this later: the comment above still holds — the returned set is empty, so app-level secrets continue to live in plaintext inside the pushed config blob. Nothing in this PR changes that, and the upstream method signature doesn't unblock it either (the nested/array extraction gap is the actual blocker).

crates/trusted-server-cli/src/run.rs
```diff
@@ -13,14 +13,16 @@ use crate::commands::config::init::{ConfigInitArgs, run_config_init};
 use crate::prebid_bundle::{NpmPrebidBundleGenerator, PrebidBundleArgs, run_bundle};
 
 #[derive(Debug, Parser)]
-#[command(name = "ts", about = "Trusted Server CLI")]
+#[command(name = "ts", version, about = "Trusted Server CLI")]
```
prk-Jr, line 16: 🌱 **seedling** — `version` here resolves to the workspace `CARGO_PKG_VERSION`, so `ts --version` prints `ts 0.1.0`.

That's a fine default, but now that `deploy`, `rollback`, `healthcheck`, and `active-version` all delegate to an edgezero pinned at a moving branch ref, the operational question during an incident is "which edgezero is baked into this binary?" — and the version string can't answer it. Not for this PR, but embedding the resolved edgezero rev (build script + `long_version`) would make rollback triage a lot cheaper.

crates/trusted-server-cli/src/run.rs
```diff
@@ -131,6 +143,198 @@ mod tests {
...
+        ]);
+        let Command::Deploy(deploy) = args.command else {
+            panic!("expected deploy command");
+        };
+        assert_eq!(deploy.service_id, Some("service-123".to_owned()));
+        assert!(deploy.staging);
+    }
+
+    #[test]
+    fn deploy_rejects_renamed_stage_flag_before_separator() {
```
prk-Jr, line 321: 👍 **praise** — This test pins exactly the dangerous interaction: `--stage` was renamed to `--staging` upstream, and deploy passthrough is `last = true`, so without a parse-time rejection a stale `--stage` in someone's shell history or CI job would be swallowed as passthrough, leave `staging` false, and ship a staging-intended deploy to production.

The inline comment explains the failure mode rather than the mechanics, which is the right thing to leave behind. Same for `deploy_captures_adapter_passthrough_after_separator` covering the other half of the boundary, and the `config gc` trio covering the preview default, the destructive sweep, and the `--dry-run`/`--yes` conflict.

prk-Jr (changes requested):
## Summary

Tight, well-tested upgrade: six edgezero deps repointed at the deploy-actions branch, `SECRET_FIELDS` migrated to `secret_fields()`, and the new `active-version` / `healthcheck` / `rollback` / `config gc` subcommands wired through with argument-parsing coverage for each. Dispatch is a clean passthrough and every new command has a doc comment. Two things block: the dependency reference is mutable while upstream #316 is still open, and the operator docs weren't updated for the new (partly destructive) surface.

## Blocking

### 🔧 wrench

- **Mutable `branch =` dep on an unmerged upstream PR** (`Cargo.toml:57-62`): upstream `stackpop/edgezero#316` is confirmed `open` / `merged: false`, but this PR is `isDraft: false` and `MERGEABLE`. See the inline comment for the `rev` pin suggestion.

- **`docs/guide/cli.md` not updated for the new CLI surface**: four subcommands ship undocumented — `active-version`, `healthcheck`, `rollback`, and `config gc`. `config gc` is the one that matters most: with `--yes` and `--older-than` it destructively sweeps config-store entries, and the only place its semantics are written down right now is a one-line clap doc string and three parse tests. The "Lifecycle commands" block (`docs/guide/cli.md:79-90`) and "Configuration commands" block (`:43-77`) both enumerate the available commands, so they read as complete and are now silently wrong.

  The same file also carries a version claim that this PR invalidates (`docs/guide/cli.md:67-70`):

  > EdgeZero **v0.0.4** only overrides leaves already present in the TOML; add newly introduced fields to existing configs before relying on their overrides.

  The dependency is no longer v0.0.4. Either confirm the behavior still holds on the new tree and drop the version qualifier, or update it to whatever gets pinned at merge time.

## Non-blocking

### 🤔 thinking

- **Unrelated dependency downgrades in the lock refresh** (`Cargo.lock`): `windows-sys 0.61.2 → 0.48.0` and `itertools 0.13.0 → 0.10.5`. Details and a scoped `cargo update` inline.

### 🌱 seedling

- **`ts --version` can't identify the embedded edgezero** (`crates/trusted-server-cli/src/run.rs:16`): follow-up, not this PR.

### 📝 note

- **`secret_fields()` port** (`crates/trusted-server-core/src/config.rs:115`): mechanical and correct; the empty-set caveat above it is unchanged.

### 👍 praise

- **`deploy_rejects_renamed_stage_flag_before_separator`** (`crates/trusted-server-cli/src/run.rs:306`): catches a silent staging→production routing hazard created by the upstream flag rename plus `last = true` passthrough.

## CI Status

All 19 checks green on `fe5767e` — reported from GitHub, not re-run locally:

- fmt: PASS
- clippy (fastly / axum / cloudflare native + wasm / spin native + wasm): PASS
- rust tests (core, axum, cloudflare, spin, ts CLI, cross-adapter parity): PASS
- integration + browser integration + Fastly EC lifecycle: PASS
- js tests (vitest) / format-typescript / format-docs: PASS

Cargo.toml
aram356: Pinned in f90f63846: the six deps now use `rev = "5f3d648c3c6c38fc6e6b22b5c65c66177363aad8"`, the exact commit the lock already resolved, so the reference is immutable and the pre-merge step becomes rev → tag. Checked the alternative: upstream cut a v0.0.5 tag since this review, but it has diverged from `feature/edgezero-deploy-actions` (not a superset), so the rev pin stands until #316 merges and a tag containing it exists.

Cargo.lock
aram356: Rebuilt the lock in f90f63846: restored main's Cargo.lock, then ran exactly the scoped update suggested here. That reproduced the same switches, so they are not from a wholesale re-resolve: re-resolving the edgezero subtree makes the resolver unify wide version ranges onto co-present lower majors (prost-build/prost-derive's `itertools >=0.10,<0.14` onto criterion's 0.10.5; colored/winapi-util onto dirs-sys's windows-sys 0.48.0). Since main's lock proves the higher-version edges are valid, I restored those five edges to main's picks and verified with `cargo metadata --locked`, `cargo check-fastly --locked`, and the host CLI suite. The only non-edgezero lock changes left are syn 3.0.3 and toml_edit 0.25.12, both required by the new edgezero tree (edgezero-macros moved to syn 3, as suspected).

crates/trusted-server-cli/src/run.rs
aram356: Agreed this stays a follow-up. With the switch to a `rev =` pin the embedded edgezero commit is now at least readable from Cargo.toml/Cargo.lock for a given checkout; embedding it in `ts --version` via a build script + `long_version` remains the durable fix for triaging a deployed binary.

crates/trusted-server-core/src/config.rs
aram356: Confirming the reading: the empty set is intentional and unchanged in this PR. The comment above the impl stays as the marker — inline settings remain the model, so app-level secrets continue to ship in plaintext inside the pushed blob. Checked the upstream tree at the pinned rev (5f3d648c): `AppConfigMeta::secret_fields()` only changes the carrier from an associated const to an owned `Vec` (and `Vec::new()` doesn't allocate); `SecretField` paths do model `ArrayEach`/nested segments, but the derive-side extraction plus the operator migration are the real work, tracked separately from this upgrade.

Cargo.toml
```diff
@@ -54,12 +54,12 @@ criterion = { version = "0.5", default-features = false, features = ["cargo_benc
 derive_more = { version = "2.0", features = ["display", "error"] }
 directories = "5"
 ed25519-dalek = { version = "2.2", features = ["rand_core"] }
-edgezero-adapter-axum = { git = "https://github.com/stackpop/edgezero", tag = "v0.0.4", default-features = false }
-edgezero-adapter-cloudflare = { git = "https://github.com/stackpop/edgezero", tag = "v0.0.4", default-features = false }
-edgezero-adapter-fastly = { git = "https://github.com/stackpop/edgezero", tag = "v0.0.4", default-features = false }
-edgezero-adapter-spin = { git = "https://github.com/stackpop/edgezero", tag = "v0.0.4", default-features = false }
-edgezero-cli = { git = "https://github.com/stackpop/edgezero", tag = "v0.0.4" }
-edgezero-core = { git = "https://github.com/stackpop/edgezero", tag = "v0.0.4", default-features = false }
+edgezero-adapter-axum = { git = "https://github.com/stackpop/edgezero", rev = "5f3d648c3c6c38fc6e6b22b5c65c66177363aad8", default-features = false }
```
prk-Jr, line 57: 🔧 **wrench** — `stackpop/edgezero#316` is still `state: open`, `merged: false`, and this PR is non-draft and `MERGEABLE`.

Credit where due: the switch from `branch =` to `rev = 5f3d648c…` closes the silent-retarget hole from the last pass, and I verified the pin is real — `5f3d648c` is an ancestor of the current branch tip `b3e02111` (`behind_by: 0`, 21 commits behind the tip), and `v0.0.5` (`a2d0c538`) is genuinely *diverged* from it (117 behind / 134 ahead), so "no existing tag contains this commit" is correct.

Two residuals the rev pin does not cover, both worth writing down before this merges:

1. **Reachability is branch-scoped.** `5f3d648c` is 116 commits *ahead* of edgezero `main` (`5f3d648c...main` → `ahead_by: 0, behind_by: 116`), so the only ref keeping it alive is `feature/edgezero-deploy-actions`. If that branch is deleted after merge — the GitHub default — or force-pushed to a rewritten history, the commit becomes unreachable and `cargo fetch` fails for every checkout and CI runner without a warm git cache. A rev pin is immutable, but immutable is not the same as fetchable.

2. **A squash merge breaks the "rev → tag" plan.** Because the pin sits 116 commits ahead of `main`, squash-merging #316 produces a single commit that is *not* an ancestor of `5f3d648c`. Re-pinning to a tag containing that squash is therefore a re-pin to **different bytes**, not a mechanical reference swap, and needs the full gate re-run rather than a lock refresh. If upstream merges with a real merge commit (or fast-forwards), `5f3d648c` stays an ancestor and the swap really is mechanical. Worth confirming which one #316 will get, since it decides how much work the pre-merge step actually is.

No code change requested here — this stays blocking purely as the merge gate the PR body already describes. Apply manually / resolve upstream.

docs/guide/cli.md
```diff
@@ -88,6 +111,42 @@ ts deploy --adapter fastly
...
+`ts deploy` accepts `--staging` (Fastly only) to build and upload a staged
+draft version cloned from the active one instead of activating a production
+deploy. Adapter passthrough arguments must follow a `--` separator; unknown
+flags before `--` (including the renamed-away `--stage`) are rejected at parse
+time rather than forwarded:
+
+```bash
+ts deploy --adapter fastly --service-id <service-id> --staging
+ts deploy --adapter fastly -- --comment "release"
+```
```
prk-Jr, line 123: 🔧 **wrench** — The new staging docs describe `deploy --staging` → `healthcheck --staging` with no config leg, and following that sequence literally produces a staged version with no config.

I traced this at the pinned rev. A staged Fastly deploy calls `relink_runtime_env_for_staging`, which does `ensure_staging_selector_store` + `mirror_production_to_staging`. That mirror copies production's **runtime-override selectors** with the config selectors redirected to `<logical>_staging` — it does *not* copy the config blob. Upstream's own comment is explicit that the twin holds "only the derived staging selectors".

Meanwhile `ConfigPushArgs.staging` and `ConfigDiffArgs.staging` both exist at this rev (each `conflicts_with = "key"`), writing and diffing the `<logical-store-id>_staging` key in the same store. Both flags arrive automatically on `ts config push` / `ts config diff` because this crate already threads those arg structs through `ConfigCommand` — so this PR ships two new operator-visible flags with neither documentation nor parse coverage, and they happen to be the missing half of the lifecycle the rest of this section documents.

Net effect for someone following the guide as written: `ts deploy --adapter fastly --service-id … --staging` stages a version whose `EDGEZERO__STORES__CONFIG__APP_CONFIG__KEY` resolves to `app_config_staging`, that key was never written, and the `healthcheck --staging` on the next line is probing a version with no config at all.

Suggested docs fix below (verified with the pinned `docs/node_modules` prettier 3.8.1, `proseWrap: preserve`):

````suggestion
```bash
ts deploy --adapter fastly --service-id <service-id> --staging
ts deploy --adapter fastly -- --comment "release"
```

A staged deploy only redirects the staged version's config selector at the
`<logical-store-id>_staging` key — it does not copy the production config blob
there. Push the staged config before probing the staged version, or it comes up
with no config at all:

```bash
ts config push --adapter fastly --staging
ts config diff --adapter fastly --staging
```

`--staging` on `config push` / `config diff` writes and compares the
`<logical-store-id>_staging` key in the same store. It is mutually exclusive
with `--key`: the staging key is derived from the store's logical id, so an
explicit key would be written where nothing reads it.
````

File: docs/guide/cli.md [hunk 8]
```diff
@@ -88,6 +111,42 @@ ts deploy --adapter fastly
...
+ts healthcheck --adapter fastly --service-id <service-id> \
+  --version <version> --domain edge.example
+
+# Re-activate a previously active version
+ts rollback --adapter fastly --service-id <service-id> \
+  --version <bad-version> --rollback-to <previous-version>
+```
+
+`healthcheck` probes `/` by default (`--path` overrides) and retries 3 times
+with a 5 second delay and 10 second timeout (`--retry`, `--retry-delay`,
```
prk-Jr: ♻️ **refactor** — `--retry 3` is 3 *total attempts*, not 3 retries after a first try.

Upstream implements it as `let attempts = retry.max(1)` with `for attempt in 0..attempts`, and only sleeps when `attempt.saturating_add(1) < attempts`. The upstream test `probe_with_retries_treats_zero_retry_as_one_attempt` pins the clamp. So the default budget is 3 attempts and 2 delays — worst case 3×10s + 2×5s = 40s — where "retries 3 times" reads as 4 attempts and 3 delays, or 55s. That is the number an operator uses to size a deploy job's timeout, so the off-by-one is worth correcting.

Same correction applies to the assertion message in `run.rs` — see the companion comment there.

```suggestion
`healthcheck` probes `/` by default (`--path` overrides) and makes 3 total
attempts — not 3 retries after a first try — with a 5 second delay between
attempts and a 10 second per-attempt timeout (`--retry`, `--retry-delay`,
```

crates/trusted-server-cli/src/run.rs
```diff
@@ -131,6 +143,198 @@ mod tests {
...
+            "7",
+            "--domain",
+            "edge.example",
+        ]);
+        let Command::Healthcheck(healthcheck) = args.command else {
+            panic!("expected healthcheck command");
+        };
+        assert_eq!(healthcheck.domain, "edge.example");
+        assert_eq!(healthcheck.version, "7");
+        assert_eq!(healthcheck.retry, 3, "should default to 3 retries");
```
prk-Jr: ♻️ **refactor** — Same off-by-one as the docs (see the companion comment on `docs/guide/cli.md:140`): upstream computes `attempts = retry.max(1)` and loops `0..attempts`, so `retry: 3` is three *total* attempts, not three retries following an initial try.

The suggestion also pins `path`, which is currently unasserted even though the docs in this same PR promise `/` as the default — so nothing in the suite would catch an upstream default change from `/` to something else.

```suggestion
        assert_eq!(healthcheck.path, "/", "should default to probing `/`");
        assert_eq!(
            healthcheck.retry, 3,
            "should default to 3 total attempts, not 3 retries after a first try"
        );
```

crates/trusted-server-cli/src/run.rs
```diff
@@ -270,6 +474,54 @@ mod tests {
...
+        let Command::Config(ConfigCommand::Gc(gc)) = args.command else {
+            panic!("expected config gc command");
+        };
+        assert_eq!(gc.adapter, "fastly");
+        assert_eq!(
+            gc.older_than, None,
+            "should not require an older-than window to preview"
+        );
+        assert!(!gc.dry_run);
+        assert!(!gc.no_env);
```
prk-Jr, line 549: ♻️ **refactor** — This test is named `config_gc_previews_by_default` but never asserts the field that actually decides preview-versus-delete.

It covers `adapter`, `older_than`, `dry_run`, and `no_env` — yet `yes` is the sole destructive gate (`gc` deletes config-store entries if and only if `--yes` is passed; `--dry-run` is documented as merely stating the default intent explicitly). As written, an upstream change flipping `yes` to `default_value_t = true` would leave this test green while `ts config gc --adapter fastly` silently became a deleting command.

```suggestion
        assert!(!gc.no_env);
        assert!(
            !gc.yes,
            "should not delete without an explicit --yes; --yes is the only destructive gate"
        );
```

crates/trusted-server-core/src/config.rs
prk-Jr, line 116: 🤔 **thinking** — The retained comment above now names a blocker that this very upgrade removes.

It says migration "needs nested/array extraction support and operator migration work tracked separately". At the pinned rev, the nested/array extraction support exists: `SecretField.path` is a `Vec<SecretPathSegment>` with `Field` and `ArrayEach` variants, `dotted_path()` renders `[*]`, the derive emits `Field` + `ArrayEach` for `Vec<T>`/slice fields and flattens leaves reached through `#[app_config(nested)]` children (`edgezero-macros/src/app_config.rs:51,236,246`, with a `SecretFieldsRecursionGuard` for cyclic graphs), and `edgezero-core/src/extractor.rs` consumes it.

So the remaining blocker is narrower than the comment claims: annotating Trusted Server's own secret leaves plus the operator migration. Worth noting `TrustedServerAppConfig` hand-implements `AppConfigMeta` rather than deriving `AppConfig`, so it would not inherit those nested paths for free — the paths would have to be built by hand, or the type moved onto the derive.

Not asking for a behaviour change; the empty set is clearly intentional and I agree it stays empty here. The concern is that the next reader hits "needs nested/array extraction support", takes it at face value, and defers the migration for a reason that stopped being true at this rev. Trimming the stale half — keeping "operator migration work tracked separately", dropping the extraction-support clause — would keep the marker honest.

Apply manually — the comment lines sit above the changed range, so this can't be expressed as a single-hunk suggestion on the added lines.

crates/trusted-server-cli/src/run.rs
prk-Jr, line 16: ⛏ **nitpick** — `version` here is new behaviour with no test behind it.

The PR body lists "a top-level `--version` flag, with argument-parsing coverage", but every `--version` in the added tests is a *subcommand argument* (`healthcheck --version`, `rollback --version`). Nothing asserts that `ts --version` itself parses, so dropping this `version` token would leave the whole suite green while the flag disappeared.

One cheap test pins it — clap surfaces `--version` as a parse *error* carrying `ErrorKind::DisplayVersion`:

```rust
#[test]
fn top_level_version_flag_is_available() {
    let err = Args::try_parse_from(["ts", "--version"])
        .expect_err("--version short-circuits parsing");
    assert_eq!(
        err.kind(),
        clap::error::ErrorKind::DisplayVersion,
        "should print the version rather than fail to parse"
    );
}
```

Apply manually — this adds a new test function rather than replacing lines in an existing hunk, so it can't be a one-click suggestion. Also note the earlier 🌱 still stands: this prints `ts 0.1.0`, which can't answer "which edgezero is in this binary?" during an incident.

prk-Jr (changes requested):
## Summary

Re-review of the three commits added since the last pass (`f90f6384`, `bf47680d`, `40dd9b37`). Both prior findings are resolved: the six deps moved from a mutable `branch =` ref to `rev = 5f3d648c…`, and the unrelated lock downgrades are gone from the diff. Verified independently — `5f3d648c` is an ancestor of the branch tip `b3e02111` (`behind_by: 0`), and the `v0.0.5` tag is genuinely *diverged* from the pin (117 behind / 134 ahead), so the "no tag contains it" reasoning holds.

What remains blocking is the upstream merge gate plus one gap in the new staging docs: the documented `deploy --staging` → `healthcheck --staging` sequence never mentions the config leg, so following it literally stages a version pointed at a config key nothing wrote.

> 4 of the inline comments below carry a one-click GitHub `suggestion` — use **Commit suggestion** (or **Add suggestion to batch**) to apply them as commits on the PR branch. The remaining comments describe the fix in prose because the change is "wait for upstream", needs a new test function, or touches files outside this diff.

## Blocking

### 🔧 wrench

- **Upstream #316 still open; branch-only reachability and squash-merge consequences** — see inline at `Cargo.toml:57`
- **Staged-deploy docs omit the config leg** — see inline at `docs/guide/cli.md:120-123`

## Non-blocking

### ♻️ refactor

- **`--retry 3` is 3 total attempts, not 3 retries** — see inline at `docs/guide/cli.md:140-141` and `crates/trusted-server-cli/src/run.rs:182`
- **`config_gc_previews_by_default` never asserts `yes`** — see inline at `crates/trusted-server-cli/src/run.rs:489`

### 🤔 thinking

- **`secret_fields()` comment names a blocker this upgrade removes** — see inline at `crates/trusted-server-core/src/config.rs:115`

### ⛏ nitpick

- **No test for the new top-level `--version`** — see inline at `crates/trusted-server-cli/src/run.rs:16`

- ♻️ **`40dd9b37` missed two live `v0.0.4` qualifiers.** The commit dropped the stale version qualifier across `docs/guide/*.md`, but two non-archival sites still restate exactly the same claim:
  - `trusted-server.example.toml:124` — the operator-facing template: *"Keep this leaf present when using the EdgeZero v0.0.4 environment override."*
  - `crates/trusted-server-cli/tests/config_env_overlay.rs:47` — *"EdgeZero v0.0.4 environment overlays cannot create missing TOML leaves, …"*

  Neither file is in this diff, so neither can carry a `suggestion` — apply manually, dropping `v0.0.4` the same way the docs did. (`docs/superpowers/plans/2026-07-15-gam-ts-cohort-attribution.md:835` also mentions it, but that is a dated archived plan describing a past experiment; leaving it is defensible.)

  Separately, the genericization itself checks out: at the pinned rev `apply_env_overlay` still only overrides keys already present in the parsed tree, and `env_overlay_only_overrides_existing_keys` asserts an env var for an absent key is silently ignored. So this is a by-design property, not a v0.0.4 bug, and dropping the version qualifier is the right call.

- 📝 **Two `syn` majors and two `toml_edit` majors now compile.** The lock carries `syn 2.0.118` alongside the new `syn 3.0.3` (pulled by `edgezero-macros`), and `toml_edit 0.23.10+spec-1.0.0` alongside `0.25.12+spec-1.1.0`. Both duplicates are host-build-only — proc-macro expansion and CLI-side TOML editing — with no `wasm32-wasip1` runtime exposure, so this is compile-time cost rather than shipped weight. Nothing to change in this PR; it collapses when the rest of the tree catches up.

- 👍 **Two things worth keeping.** First, the rev pin was done as a *scoped* lock rebuild — restore main's lock, run the narrow update, then restore the five edges the resolver had needlessly walked back — and verified with `cargo metadata --locked` rather than asserted. That is exactly the right way to answer "is this lock diff minimal?". Second, the `config gc` documentation catches a real footgun rather than paraphrasing `--help`: it spells out that `--older-than` is a safety assertion about the *whole physical store*, and that `gc`'s `--no-env` means something different from every other `config` subcommand's `--no-env` (it reselects the target store instead of skipping the app-config overlay). On a destructive command, that distinction is the one an operator would otherwise learn the hard way.

## CI Status

All 19 reported checks PASS at `40dd9b37`:

- cargo fmt: PASS (required)
- cargo test: PASS (required)
- vitest: PASS
- integration tests: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS

## Verification performed for this review

Every suggestion below was applied in an isolated worktree at `40dd9b37`, verified alone and then batched, with a byte-exact pre/post patch snapshot each time:

| | `cargo fmt --all -- --check` | prettier 3.8.1 (`docs/node_modules`) | clippy `-D warnings` | tests |
| --- | --- | --- | --- | --- |
| each suggestion alone | PASS | PASS | PASS | 165 pass |
| all four batched | PASS | PASS | PASS | 165 pass |

No drift between the approved patch and the post-verification tree on any run. Scope note: clippy and tests were run as `--package trusted-server-cli --target aarch64-apple-darwin` rather than the full six-alias chain, because both Rust suggestions are assertion lines inside that crate's own `#[cfg(test)]` module and cannot reach an adapter target.

Cargo.toml
aram356: Recorded both residuals in the PR body's blocked section: the branch-scoped reachability (making the rev → tag repin a hard pre-merge requirement rather than housekeeping) and the squash-merge consequence (different bytes, full gate re-run). Merge method on stackpop/edgezero#316 is not settled on our side; flagging it there before the repin is planned.

docs/guide/cli.md
aram356: Applied your docs text verbatim in 87546d36d, and added the missing parse coverage in 01d7f611b: config_push_parses_staging_and_rejects_explicit_key and config_diff_parses_staging_and_rejects_explicit_key cover both new flags plus the conflicts_with = "key" boundary. Confirmed the trace at the pinned rev before applying: mirror_production_to_staging copies only the derived staging selectors (edgezero-adapter-fastly/src/cli.rs:3508), so the staged config blob really does need its own push.

docs/guide/cli.md
aram356: Applied in 87546d36d after confirming attempts = retry.max(1) with the sleep gated on attempt + 1 < attempts at the pinned rev (edgezero-adapter-fastly/src/cli.rs:4377-4380), so the worst-case budget arithmetic in your comment is right: 3 attempts, 2 delays.

crates/trusted-server-cli/src/run.rs
aram356: Applied in 01d7f611b (rustfmt re-wrapped the multi-line assert; content unchanged). The path default is now pinned alongside the reworded retry assertion.

crates/trusted-server-cli/src/run.rs
aram356: Applied in 01d7f611b. The test now asserts all four defaults plus yes, so an upstream default_value_t flip on the destructive gate fails the suite.

crates/trusted-server-core/src/config.rs
aram356: Reworded in 4aee05272. The comment now states the narrower remaining blocker: the secret-field paths must be spelled out in this hand-written impl (it does not inherit the derive's nested/array paths), plus the operator migration. This corrects my earlier reply on the other thread too — the derive-side extraction does exist at the pinned rev (Field/ArrayEach emission in edgezero-macros/src/app_config.rs), so the extraction-support clause was the stale half.

crates/trusted-server-cli/src/run.rs
aram356: Added top_level_version_flag_is_available in 01d7f611b, asserting ErrorKind::DisplayVersion as suggested. The 🌱 stands as a follow-up and is listed in the PR body's out-of-scope section.

Cargo.toml
prk-Jr: 🔧 **wrench** — Upstream `stackpop/edgezero#316` is still open, so this pin is immutable but branch-scoped.

Re-verified against the upstream API at review time, not carried over from the last round:

- `repos/stackpop/edgezero/pulls/316` -> `state: open`, `merged: false`, `merged_at: null`, head `b3e02111`, `mergeable_state: clean`.
- `compare/5f3d648c...main` -> `status: behind, ahead_by: 0, behind_by: 116` — the pinned commit is 116 commits ahead of edgezero `main`, so `feature/edgezero-deploy-actions` is the only ref keeping it fetchable.
- Tag list unchanged: `v0.0.5 = a2d0c538`, `v0.0.4 = 9e661ae5`. No tag contains `5f3d648c`.

**Why this still blocks even though the body records it:** this PR is non-draft and `MERGEABLE`, so nothing mechanical stops it landing before the repin. Two consequences follow.

1. Deleting the feature branch after #316 merges — the GitHub default — makes `cargo fetch` fail on every cold checkout and CI runner. The rev pin is immutable, not unconditionally fetchable.
2. A **squash** merge of #316 produces a commit that is not an ancestor of `5f3d648c`, so "rev -> tag" becomes a repin to *different bytes* plus a full gate re-run, not a reference swap. A merge commit or fast-forward keeps `5f3d648c` an ancestor and keeps the swap mechanical.

No code change requested here — this comment is the merge gate, and it is the only reason the verdict is `REQUEST_CHANGES` rather than `COMMENT`. Worth confirming the merge method on #316 before planning the repin, and moving the deployer's `deploy-fastly@<ref>` in lockstep.

docs/guide/cli.md
```diff
@@ -88,6 +111,58 @@ ts deploy --adapter fastly
...
+ts active-version --adapter fastly --service-id <service-id>
+
+# Probe a deployed version until it reports healthy
+ts healthcheck --adapter fastly --service-id <service-id> \
+  --version <version> --domain edge.example
+
+# Re-activate a previously active version
+ts rollback --adapter fastly --service-id <service-id> \
+  --version <bad-version> --rollback-to <previous-version>
+```
```
prk-Jr: 🔧 **wrench** — The newly documented rollback lifecycle gives an operator no way to obtain `--rollback-to`.

Lines 140-153 present `active-version` -> `healthcheck` -> `rollback` as one "inspect and verify" group, and the `active-version` comment reads *"Print the currently active deployment version"*. An operator who deploys, sees breakage, and then reaches for rollback runs `ts active-version` — and gets the **bad** version back. Traced at the pinned rev:

- `edgezero_cli::run_active_version` resolves the *currently active* version (`Action::EmitVersion`), and its own doc says the value must be "captured BEFORE a deploy ... Fastly's version list has no field to infer a previously-live version afterward."
- `run_rollback` hard-errors without `--rollback-to`: *"a production rollback requires --rollback-to ... it must be captured before the deploy that superseded it -- use `deploy`'s `previous-version` output."* That upstream hint is itself stale: `grep -rn 'previous.version'` across the pinned tree finds **no** such emission. Production deploy emits only the *activated* version, via `log::info!("version={version}")`.
- Upstream's own GitHub action encodes the ordering the docs omit: "A production deploy runs `active-version` to capture the rollback target and fails fast (before touching the provider) if it is missing."

Two supporting corrections are folded into the same contiguous range, since they sit inside it: the `<version>` placeholder for `healthcheck` has no stated source (it comes from the deploy's `version=<N>` line, which `run_deploy` only emits when `args.service_id.is_some() && adapter == "fastly"`), and all three commands are Fastly-only — `edgezero-adapter-{axum,cloudflare,spin}/src/cli.rs` each return `Err` for `EmitVersion | Healthcheck | Rollback` ("The Fastly staging lifecycle is Fastly-only") — yet only `deploy --staging` is labelled "(Fastly only)".

Same class as the config-leg gap fixed last round: following the guide literally leaves you stuck, here at the worst possible moment.

````suggestion
Inspect and verify deployments with the deploy lifecycle commands. All three are
Fastly-only — the axum, cloudflare, and spin adapters reject them:

```bash
# Capture the production rollback target BEFORE deploying: after a deploy this
# prints the NEW version, and Fastly keeps no record of which version was live
# before it, so the target is then unrecoverable.
ts active-version --adapter fastly --service-id <service-id>

# Probe a deployed version until it reports healthy. `<version>` is the version
# the deploy activated; pass `--service-id` to `ts deploy` and it emits that as
# a machine-readable `version=<N>` line.
ts healthcheck --adapter fastly --service-id <service-id> \
  --version <version> --domain edge.example

# Re-activate the version captured before the deploy
ts rollback --adapter fastly --service-id <service-id> \
  --version <bad-version> --rollback-to <previous-version>
```
````

File: docs/guide/cli.md [hunk 12]
```diff
@@ -76,6 +76,29 @@ Trusted Server settings JSON. This blob model is intentional because full
...
+ts config gc --adapter fastly --yes --older-than 7d
+```
+
+`config gc` sweeps every root in the selected physical store, so `--older-than`
+is a safety assertion about the whole store: nothing in it changed within the
+window and no writer is targeting it. Unlike the other `config` subcommands,
+`gc` never loads the typed app config; its `--no-env` flag instead ignores
+`EDGEZERO__STORES__CONFIG__<ID>__NAME` when resolving which physical store to
+sweep. On a destructive run, check the store id `gc` reports before passing
+`--yes`.
```
prk-Jr: ♻️ **refactor** — `config gc`'s destructive-target warning enumerates `--no-env` but omits `--store`.

This paragraph deliberately spells out that `gc`'s `--no-env` retargets the *physical store* rather than skipping an overlay — that is the real footgun and it is good that it is called out. But `ConfigGcArgs` also carries `#[arg(long)] pub store: Option<String>` ("Override the config-store id (defaults to the manifest's)"), which retargets more directly and is both undocumented and untested. An operator who reads this paragraph as the complete list of ways `--yes` can hit the wrong store is wrong.

```suggestion
`config gc` sweeps every root in the selected physical store, so `--older-than`
is a safety assertion about the whole store: nothing in it changed within the
window and no writer is targeting it. Unlike the other `config` subcommands,
`gc` never loads the typed app config; its `--no-env` flag instead ignores
`EDGEZERO__STORES__CONFIG__<ID>__NAME` when resolving which physical store to
sweep, and `--store <id>` overrides the manifest's config-store id outright.
Both change which store gets swept, so on a destructive run check the store id
`gc` reports before passing `--yes`.
```

docs/guide/cli.md
```diff
@@ -88,6 +111,58 @@ ts deploy --adapter fastly
 ts serve --adapter fastly
 ```
 
+`ts deploy` accepts `--staging` (Fastly only) to build and upload a staged
+draft version cloned from the active one instead of activating a production
+deploy. Adapter passthrough arguments must follow a `--` separator; unknown
+flags before `--` (including the renamed-away `--stage`) are rejected at parse
+time rather than forwarded:
```
prk-Jr: ♻️ **refactor** — The `deploy` passthrough boundary is a breaking change, documented as if it were always the rule.

Upstream changed `DeployArgs.adapter_args` from `#[arg(trailing_var_arg = true, allow_hyphen_values = true)]` to `#[arg(last = true)]` — confirmed by diffing `edgezero-cli/src/args.rs` between `9e661ae` and `5f3d648`. So `ts deploy --adapter fastly --comment "release"` used to work and now fails at parse time.

It fails *loudly*, which is the right design, and the docs state the new rule correctly. The gap is that they read as a new safety property rather than as a migration. Nothing in this repo invokes `ts deploy` with passthrough args (only `docs/guide/cli.md` mentions it; no workflow or script does), so the blast radius is operator runbooks living outside the repo — exactly the audience for this page.

```suggestion
`ts deploy` accepts `--staging` (Fastly only) to build and upload a staged
draft version cloned from the active one instead of activating a production
deploy. Adapter passthrough arguments must now follow a `--` separator; unknown
flags before `--` (including the renamed-away `--stage`) are rejected at parse
time rather than forwarded. This is a change: passthrough args previously
worked without the separator, so existing runbooks and CI jobs that pass
adapter flags directly need the `--` added:
```

crates/trusted-server-core/src/config.rs
```diff
@@ -110,9 +110,12 @@ impl edgezero_core::app_config::AppConfigMeta for TrustedServerAppConfig {
     // Phase 1 intentionally preserves the existing inline-settings model:
     // `ts config push` publishes the validated Trusted Server config as one
     // app-config blob. Migrating app-level secrets to `EdgeZero` secret-store
-    // references needs nested/array extraction support and operator migration
-    // work tracked separately.
-    const SECRET_FIELDS: &'static [edgezero_core::app_config::SecretField] = &[];
+    // references needs the secret-field paths spelled out here (this
+    // hand-written impl does not inherit the derive's nested/array paths)
+    // plus operator migration work tracked separately.
+    fn secret_fields() -> Vec<edgezero_core::app_config::SecretField> {
```
prk-Jr, line 116: 👍 **praise** — The reworded `secret_fields()` comment is now accurate.

Last round flagged that the retained comment named a blocker the upgrade had removed. `4aee0527` fixed it correctly rather than cosmetically: the derive **does** emit `Field`/`ArrayEach` paths at the pinned rev, and `TrustedServerAppConfig` hand-implements `AppConfigMeta` so it genuinely does not inherit them — which is exactly what the new wording says.

Also verified the migration itself is behavior-preserving: the trait has exactly one item in both revs (`const SECRET_FIELDS: &'static [SecretField]` -> `fn secret_fields() -> Vec<SecretField>`), `Vec::new()` does not allocate, and the only consumer is deploy-time `validate_excluding_secrets` (`edgezero-core/src/app_config.rs:304`), so an empty set remains a no-op. Secrets still ship plaintext in the pushed blob — unchanged by this PR and out of scope.

prk-Jr (changes requested):
## Summary

Repoints the six edgezero workspace deps from `tag = "v0.0.4"` to an immutable `rev = "5f3d648c"` pin on the unmerged `feature/edgezero-deploy-actions` branch, wires the new `ts` lifecycle subcommands, and documents them. The Rust side is in good shape: the lockfile claims in the PR body check out, the `SECRET_FIELDS` -> `secret_fields()` migration is behavior-preserving, and no production wire format moved across the 39k-line upstream jump. The remaining blockers are the upstream merge gate and one operator-docs gap that leaves a rollback unreachable.

> 3 of the inline comments below carry a one-click GitHub `suggestion` — use **Commit suggestion** (or **Add suggestion to batch** for several at once) to apply them as commits on the PR branch. All three are `docs/guide/cli.md` prose and were verified against the pinned prettier in `docs/node_modules` (3.8.1), individually and batched, with no drift.

## Blocking

### 🔧 wrench

- **Upstream `stackpop/edgezero#316` is still open; the pin is immutable but branch-scoped** — see inline at `Cargo.toml:57`
- **The documented rollback lifecycle gives an operator no way to obtain `--rollback-to`** — see inline at `docs/guide/cli.md:140-153`

## Non-blocking

### ♻️ refactor

- **`config gc`'s destructive-target warning enumerates `--no-env` but omits `--store`** — see inline at `docs/guide/cli.md:94-100`
- **`deploy` passthrough is a breaking change, documented as if it were always the rule** — see inline at `docs/guide/cli.md:114-118`

### 👍 praise

- **The reworded `secret_fields()` comment is now accurate** — see inline at `crates/trusted-server-core/src/config.rs:113-116`

- 📝 **Lockfile claims audited and confirmed; the "second lockfile" CI risk does not exist on this branch.** All six `Cargo.toml` deps carry the same `rev = "5f3d648c3c6c38fc6e6b22b5c65c66177363aad8"`. `Cargo.lock` has 8 matching source lines (the six plus transitive `edgezero-adapter` and `edgezero-macros`), all identical, and zero remaining `tag=` / `v0.0.4` references. `cargo metadata --locked --offline` exits 0, so the lock is consistent with the manifest and needs no re-resolve. The body's "only `syn 3.0.3` and `toml_edit 0.25.12` are non-edgezero changes" is **true** — the rest of the lock diff is source-line repointing plus mechanical `toml_edit` -> `toml_edit 0.23.10+spec-1.0.0` / `0.25.12+spec-1.1.0` disambiguation, and every new transitive (`winnow 1.0.3`, `toml_datetime 1.1.1`, `toml_writer`, `toml_parser 1.1.2`, `indexmap 2.14.0`) already existed in the lock. The earlier-round `windows-sys 0.61.2->0.48.0` and `itertools 0.13.0->0.10.5` regressions are gone. On the separate-lockfile concern raised in an earlier round: `crates/trusted-server-integration-tests/Cargo.lock` does not exist on this branch — the crate is a workspace member using `{ workspace = true }` deps, so it shares the root lock. The only other tracked lockfile is `crates/trusted-server-openrtb-codegen/Cargo.lock`, which has no edgezero edges, is untouched here, and has no `--locked` CI gate. No lockfile-mismatch gate is at risk.

- 📝 **Production wire formats and runtime behavior are unchanged across the upstream jump.** Diffed `9e661ae..5f3d648` directly, since neither the PR body nor CI answers this. Config blob compatibility holds: `BlobEnvelope` keeps `ENVELOPE_VERSION_V1` and the same four fields (only error `Display`/`Debug` gained hash redaction), and `POINTER_KIND = "fastly_config_chunks"`, `CHUNK_KEY_INFIX = ".__edgezero_chunks."`, `FASTLY_CONFIG_ENTRY_LIMIT = 8_000`, `CHUNK_PAYLOAD_TARGET = 7_000` are all unchanged — which matters because `crates/trusted-server-core/src/settings_data.rs` duplicates those constants in this repo's own resolver. `edgezero-core/src/manifest.rs` changed only in style and `env_config.rs` is not in the diff at all, so `EnvConfig::store_key("config", ...)` still resolves the staging selector a staged deploy redirects. `EdgeError::status()` maps identically in both revs, so no status-code drift. New surface is additive with no-op defaults (`Router::with_state` plus a per-request `extensions.extend(state_extensions.clone())`, empty here; the Fastly `dispatch_with_handles` `extend` closure is `|_req, _extensions| {}` on the default path). One thing deliberately **not** raised as a finding: edgezero's `ConfigOutOfDate` error now emits a redacted `field_path` and a literal `<app-cli>` placeholder, which would be an operator-visible 503-body regression — except this repo never uses edgezero's `AppConfig<C>` extractor, loading config through its own `config_payload.rs` -> `BlobEnvelope` -> `Settings::from_json_value` path with `TrustedServerError::Configuration`. Not reachable.

- 📝 **What was and was not verified locally.** Run in an isolated worktree at this exact head: `cargo fmt --all -- --check` clean; pinned prettier 3.8.1 from `docs/node_modules` (not `npx`, which resolves a different version and reports false failures) `--check .` clean at baseline, for each suggestion alone, and for all three batched; `cargo metadata --locked --offline` exit 0; `./scripts/test-cli.sh` exit 0; `cargo test -p trusted-server-cli --target aarch64-apple-darwin --lib` **168 passed, 0 failed**, with all 14 new `run::tests::*` present and green. Byte-exact pre/post patch snapshots showed no drift on every verification run. Relied on green CI at this head for the six-alias clippy chain, `cargo test-fastly/-axum/-cloudflare/-spin`, the parity suite, and vitest — no Rust suggestion is proposed, so none was required for scratch verification. **Not verifiable from here at all:** the runtime behavior of every lifecycle command (`deploy --staging`, `healthcheck`, `rollback`, `active-version`, `config gc`), which needs a live Fastly service and API token. All coverage in this PR and in this review is clap parse-level only.

## CI Status

- cargo fmt: PASS (required)
- cargo test: PASS (required)
- vitest: PASS
- integration tests: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS

docs/guide/cli.md
```diff
@@ -88,6 +111,58 @@ ts deploy --adapter fastly
...
+
+```bash
+ts deploy --adapter fastly --service-id <service-id> --staging
+ts deploy --adapter fastly -- --comment "release"
+```
+
+A staged deploy only redirects the staged version's config selector at the
```
ChristianPavilonis, line 128: 🔧 **Staged deployments continue loading the production config**

When an operator runs `config push --staging` followed by `deploy --staging`, EdgeZero relinks the staged version's `edgezero_runtime_env` resource and sets `EDGEZERO__STORES__CONFIG__TRUSTED_SERVER_CONFIG__KEY=trusted_server_config_staging`. Trusted Server's Fastly entry point bypasses `edgezero_adapter_fastly::run_app`, however: `crates/trusted-server-adapter-fastly/src/main.rs:125` builds the app directly, and `crates/trusted-server-core/src/settings_data.rs:39` resolves the key through `EnvConfig::from_env()` without ever opening the linked `edgezero_runtime_env` store. It therefore falls back to `trusted_server_config`.

The staged blob is never read, so the staging healthcheck exercises the new binary against production config. That can produce a false pass and miss an incompatible staged code/config pairing before activation. It also means the statement here that the staged version has no config is inaccurate; it silently gets production config. The added tests only assert Clap's `staging` boolean, so the full CLI suite passes without covering this path.

Load the version-linked `edgezero_runtime_env` dictionary before building `TrustedServerApp`, then use that `EnvConfig` to resolve both the physical config-store name and key. Add an adapter test proving staging selects `trusted_server_config_staging` while production selects `trusted_server_config`.

ChristianPavilonis (approved):
## Summary

Reviewed `dd8bcdde7e7c5773ebda09411e37525d877ca2f6` against `d97bda69f89b15bc61f30f28ad12cc2767924897`. The dependency migration and CLI wiring are otherwise consistent, and the relevant local checks and CI pass. I left one inline P1 finding on the staged-config runtime path.

Cargo.toml
aram356: Unblocked in 6462eb84d: stackpop/edgezero#316 merged and v0.0.6 (c8d6e414) contains it, so the six deps now use tag = "v0.0.6". Your squash prediction was right — the merge commit has a single parent, making this different bytes rather than a mechanical swap — so it was treated accordingly: args.rs is byte-identical between 5f3d648c and v0.0.6, the three documented behaviors (retry.max(1) clamp, selector-only staging mirror, leaves-only overlay) re-verified at the tag, and the full gate list re-run locally (fmt, six clippy targets, all adapter suites, core 2181, parity 13, host CLI, JS 871, docs prettier). Both residuals are now moot: the tag keeps the commit fetchable independent of branch lifecycle, and the repin question is settled.

docs/guide/cli.md
aram356: Confirmed the bug and traced the same path: the entry point bypasses run_app, EnvConfig::from_env() is empty on Compute, so a staged version silently loads the production blob. Not fixed in this PR, deliberately: the store name, key spellings, and derivation rules for edgezero_runtime_env are edgezero internals whose loader (env_config_from_runtime_dictionary, lib.rs:177) is private, and replicating them here would be a drift-prone copy. Filed stackpop/edgezero#349 to expose that loader as public API, and #1082 here for the wiring plus the staging-vs-production selector adapter test once it lands. Meanwhile 89c171e9a corrects the docs: the staged-deploy section now carries a known-limitation callout stating a staged version loads the production config (replacing the inaccurate 'no config at all') and warns against relying on --staging to validate config changes.

docs/guide/cli.md
aram356: Applied your suggested block verbatim in 89c171e9a after re-verifying the claims at v0.0.6 (the repin landed between rounds): no previous-version emission exists in the tree, deploy emits version=<N> at edgezero-adapter-fastly/src/cli.rs:4811 only on the Fastly path with a service id, and the axum/spin adapters reject the lifecycle trio. The stale hint in run_rollback's error message is an upstream bug, filed as stackpop/edgezero#350 with the deploy-emits-previous-version option noted as the better fix.

docs/guide/cli.md
aram356: Applied in 89c171e9a, and added parse coverage for --store (config_gc_parses_store_override, plus a store-defaults-to-None assertion in the preview test) so both retargeting paths named in the warning are now documented and tested.

docs/guide/cli.md
aram356: Applied in 89c171e9a. Confirmed the framing: the trailing_var_arg to last = true change is upstream's, and nothing in this repo passes adapter flags without the separator, so the migration note targets exactly the external-runbook audience.

Cargo.toml
aram356: Resolved since this round was written: #316 merged (as a squash, as you predicted) and v0.0.6 contains it, so 6462eb84d moved the six deps to tag = "v0.0.6" with the full gate list re-run. Details on the sibling thread.

aram356:
Need to merge https://github.com/stackpop/edgezero/pull/351 and https://github.com/stackpop/edgezero/pull/352

prk-Jr (approved):

jwrosewell:
Approved, and no objection to the upgrade itself. One sequencing note before it lands.

This PR changes `config.rs`, `trusted-server.example.toml`, `configuration.md` and `cli.md`, which the provider stack in #1043–#1047 and #1094 also touches. To avoid a conflict that one side then has to unpick, could #940 merge behind that stack rather than ahead of it?

If the stack can't be merged quickly and #940 needs to go first, the alternative we suggested to Rowena, Jason and Shailley is a release branch model: keep the current live branch and the next release branch in sync, so both sets of changes land without either blocking the other.

Either order works for us. Flagging it now so whoever merges first knows the other is coming.

aram356:
> Approved, and no objection to the upgrade itself. One sequencing note before it lands.
> 
> This PR changes `config.rs`, `trusted-server.example.toml`, `configuration.md` and `cli.md`, which the provider stack in #1043–#1047 and #1094 also touches. To avoid a conflict that one side then has to unpick, could #940 merge behind that stack rather than ahead of it?
> 
> If the stack can't be merged quickly and #940 needs to go first, the alternative we suggested to Rowena, Jason and Shailley is a release branch model: keep the current live branch and the next release branch in sync, so both sets of changes land without either blocking the other.
> 
> Either order works for us. Flagging it now so whoever merges first knows the other is coming.

This PR Iis currently blocking deployment workflows for active prospects, so holding it behind the provider stack would have a direct customer impact.
The #1043–#1047/#1094 stack can then rebase onto #940 and resolve the overlapping config and documentation changes against that baseline. I don’t think we need the added release-branch overhead solely for merge sequencing.

jwrosewell:
For the record on sequencing. This merge landed while our request of
31 August on this PR to land the provider stack first, or say here that
this PR goes first, was open without an answer. The cost landed on the
stack: five of the seven PRs (#1044 to #1047, #1094) went into conflict
on `crates/trusted-server-adapter-fastly/src/main.rs`, and we have
rebased all seven branches again, re-run the full suite, and they are
green on required CI as of this morning.

We would rather not spend review-ready branches on repeated rebases, so
the standing request stays the same and is now urgent: a maintainer
decision on the order these land. The stack has been open since
19 August, carrying work under review since 2 July as #838, and every
review comment on it has an answer. Tell us what lands next and we will
sequence behind it once, rather than after each merge.

PR #943, opened 2026-07-21
Make the SSAT debug comment configurable

### LINKED ISSUE #935: For SSAT, make debug comment configurable
### Description

For SSAT, make debug comment configurable

prk-Jr:
## Summary

- Adds `[debug.auction_html_comment_options]` for controlling SSAT `<!-- ts-debug: ... -->` auction comments: provider/mediator/bid section toggles, a fail-closed metadata subset, sensitivity modes, and compact or pretty outer JSON formatting.
- Provides three explicit sensitivity levels: `redacted` (default, schema-validated server-owned diagnostics), `upstream` (six named provider-controlled diagnostics), and `full` (raw response metadata and untruncated creatives for local investigation).
- Preserves unconditional safeguards in every mode and format: HTML-comment terminator neutralization and a 256 KiB total dump cap.

## Configuration

```toml
[debug]
auction_html_comment = true

[debug.auction_html_comment_options]
include_provider_responses = true
include_mediator_response = true
include_bids = true
metadata_keys = ["error_type", "http_status", "message"]
verbosity = "redacted" # redacted | upstream | full
format = "compact"     # compact | pretty
```

`format = "pretty"` indents the outer auction dump for easier navigation. Nested JSON-looking values such as PBS `requestbody` and `responsebody` remain strings exactly as captured.

`upstream` and `full` can expose request or identity data and must not be enabled in production.

## Security model

- In `redacted`, `metadata_keys` can only narrow the fixed Rust allowlist; it cannot expose arbitrary keys such as PBS `debug`.
- `error_type` and `http_status` are schema-validated, and `message` is generated from fixed server-owned wording rather than copied from a provider.
- Provider-controlled errors, warnings, timings, bid statuses, and upstream messages require `upstream` or `full`.
- Raw PBS `debug.httpcalls` and `resolvedrequest` require `full`.
- Pretty formatting changes whitespace only; it does not change the dump schema or sensitivity boundary.

## Main changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/settings.rs` | Adds configurable section toggles, metadata subset, `AuctionDebugCommentVerbosity`, and `AuctionDebugCommentFormat` with backward-compatible defaults |
| `crates/trusted-server-core/src/publisher.rs` | Applies mode-specific metadata/creative handling and compact/pretty serialization while retaining the unconditional cap and terminator neutralization |
| `trusted-server.example.toml` | Documents all options, accepted modes/formats, and sensitive-mode warnings |
| `docs/guide/auction-orchestration.md` | Explains how to enable, push, inspect, and safely use the SSAT HTML debug comment |
| `docs/superpowers/specs/` | Includes reviewed designs for configuration/security and output formatting |
| `docs/superpowers/plans/` | Includes reviewed test-driven implementation plans |

## Closes

Closes #935

## Test plan

- [x] `cargo test-fastly`
- [x] `cargo test-axum`
- [x] `cargo test-cloudflare`
- [x] `cargo test-spin`
- [x] `cargo clippy-fastly`
- [x] `cargo clippy-axum`
- [x] Focused red/green tests cover config defaults and invalid values, compact/pretty semantic equivalence, nested-string fidelity, the total cap, and comment-terminator safety
- [x] Manual local validation with Full output confirmed provider HTTP calls, resolved requests, bids, prices, and currency are visible according to section toggles
- [ ] GitHub Actions (running after the latest push)

## Checklist

- [x] Changes follow `CLAUDE.md` conventions
- [x] Compact remains the default for backward compatibility
- [x] Unknown verbosity and format values fail configuration loading
- [x] New behavior has focused regression tests


crates/trusted-server-core/src/settings.rs
```diff
@@ -1878,16 +1878,18 @@ pub struct DebugConfig {
...
+    /// per-provider auction diagnostics. The default validates response-level
+    /// metadata, but bid fields and bounded creative previews remain visible;
+    /// this is not a fully anonymized dump. Never enable in production.
     #[serde(default)]
     pub auction_html_comment: bool,
 
+    /// Content and verbosity of the `auction_html_comment` dump. Ignored
+    /// when `auction_html_comment` is false.
+    #[serde(default)]
+    pub auction_html_comment_options: AuctionDebugCommentOptions,
```
ChristianPavilonis, line 2401: 🔧 **P1 — Preserve rollback compatibility for default config blobs.**

`ts config push` serializes the complete `Settings` blob. Because this field has no `skip_serializing_if`, an omitted/default options table is emitted in every blob. The previous binary's `DebugConfig` uses `deny_unknown_fields`, so it rejects that blob during a mixed-version deployment or emergency rollback. This is the compatibility contract already enforced for default `AuctionConfig` additions.

Add a default predicate (or `PartialEq`-based helper) and `skip_serializing_if` so default options are omitted while non-default options still serialize. Please add a `config_payload` regression test that deserializes the default `debug` payload through a legacy `DebugConfig` schema with `deny_unknown_fields`.

ChristianPavilonis (changes requested):
## Summary

🔧 Requesting changes because the new default configuration serialization breaks rollback and mixed-version deployment compatibility. One P1 inline finding is included.

crates/trusted-server-core/src/settings.rs
```diff
@@ -1903,6 +1905,135 @@ pub struct DebugConfig {
...
+            include_bids: true,
+            metadata_keys: default_auction_debug_metadata_keys(),
+            verbosity: AuctionDebugCommentVerbosity::Redacted,
+            format: AuctionDebugCommentFormat::Compact,
+        }
+    }
+}
+
+impl AuctionDebugCommentOptions {
+    pub(crate) fn normalize(&mut self) {
```
aram356, line 2528: 🔧 **wrench** — Unknown `metadata_keys` entries are silently dead config. `normalize()` trims and drops empty strings but never checks membership in `AUCTION_DEBUG_METADATA_ALLOWLIST`. The struct's own doc comment justifies `deny_unknown_fields` with "an operator typo must fail config load loudly, not be silently ignored" — yet `metadata_keys = ["http_staus"]` (typo) or `["errors"]` (a plausible expectation, since `errors` was in the pre-PR default allowlist) silently renders `metadata: {}` with no signal anywhere. Since upstream keys are unlocked by `verbosity`, not by `metadata_keys`, and `full` ignores the field, its meaningful domain is exactly the three allowlist keys — any other entry is dead config.

**Fix**: reject keys outside the allowlist during config load (e.g. in `prepare_runtime`, which already returns `Result`), keeping the render-time intersection as defense-in-depth. This doesn't contradict the spec's silent-drop security invariant — that boundary stays; this adds the loud-failure UX the config layer already promises. Please add a test alongside, and one for `deny_unknown_fields` itself (e.g. `metadata_key = [...]`), which is currently untested.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -1906,19 +1891,114 @@ fn truncate_with_marker(value: &str, max: usize) -> String {
     format!("{}…(truncated {} bytes)", &value[..end], value.len() - end)
 }
 
-/// Build a redacted JSON view of a single provider response for the `ts-debug`
-/// dump: only [`DEBUG_DUMP_METADATA_ALLOWLIST`] metadata keys survive, and each
-/// bid's creative is previewed to [`MAX_BID_CREATIVE_DUMP_BYTES`].
+/// Return a recognized server-owned provider error classification.
+fn validated_error_type(
```
aram356, line 1965: ♻️ **refactor** — `validated_error_type` and `safe_error_message` hardcode `"parse_response" | "launch_failed" | "transport" | "timeout" | "http_status"`, duplicating the `ERROR_TYPE_*` constants at `auction/orchestrator.rs:112-120` (mostly private today). A future error classification added in the orchestrator silently vanishes from redacted dumps via the `_ => None` arm, and no test would catch the drift.

**Fix**: make the constants `pub(crate)` and use them as match patterns (`&'static str` consts are valid patterns), or centralize the classification→message mapping next to the constants.

crates/trusted-server-core/src/settings.rs
```diff
@@ -1903,6 +1905,135 @@ pub struct DebugConfig {
...
+    /// Include each provider's `bids` array (vs. status/metadata only).
+    #[serde(default = "default_true")]
+    pub include_bids: bool,
+
+    /// Subset of [`AUCTION_DEBUG_METADATA_ALLOWLIST`] to surface in
+    /// [`AuctionDebugCommentVerbosity::Redacted`] mode. Keys outside the
+    /// fixed allowlist are always dropped, config or not. This selector cannot
+    /// unlock provider diagnostics. Ignored when `verbosity` is `Full`.
+    #[serde(default = "default_auction_debug_metadata_keys")]
+    pub metadata_keys: Vec<String>,
```
aram356, line 2493: ⛏ **nitpick** — The docstring says the subset applies in `Redacted` mode and is "Ignored when `verbosity` is `Full`", but in `Upstream` mode it still gates the three base keys (Upstream builds on `redacted_metadata_for_dump` in `publisher.rs`). The guide table in `docs/guide/auction-orchestration.md` has the same gap. One sentence in each would make the Upstream behavior explicit.

docs/guide/auction-orchestration.md
```diff
@@ -823,3 +823,72 @@ Every auction response includes structured metadata in `ext.orchestrator`:
...
+
+```toml
+[debug]
+auction_html_comment = true
+
+[debug.auction_html_comment_options]
+include_provider_responses = true
+include_mediator_response = false
+include_bids = false
+metadata_keys = ["error_type", "http_status", "message"]
```
aram356: ⛏ **nitpick** — This copy-paste recipe sets `metadata_keys = [...]` together with `verbosity = "full"`, where the field has no effect. Operators tweaking the recipe may expect it to filter. Drop the line from this example or annotate it.

aram356 (changes requested):
## Summary

Adds `[debug.auction_html_comment_options]` — section toggles, a `metadata_keys` subset selector, a `redacted`/`upstream`/`full` verbosity ladder, and `compact`/`pretty` formatting — replacing the old single hardcoded allowlist. The security design is sound: the redacted-mode schema validation matches exactly what `orchestrator.rs` and `prebid.rs` actually write, all six upstream keys have real writers, the dropped `"status"` allowlist key has no production writer, and terminator neutralization plus the 256 KiB cap are exercised across every verbosity × format combination. One blocking config-validation gap and three smaller items, all inline.

## Blocking

### 🔧 wrench

- **Unknown `metadata_keys` entries are silently dead config**: validate membership in `AUCTION_DEBUG_METADATA_ALLOWLIST` at config load (`crates/trusted-server-core/src/settings.rs:2007`)

## Non-blocking

### ♻️ refactor

- **`error_type` literals duplicated from `orchestrator.rs`**: reuse the `ERROR_TYPE_*` constants to prevent drift (`crates/trusted-server-core/src/publisher.rs:1895`)

### ⛏ nitpick

- **`metadata_keys` docs omit its Upstream-mode behavior**: it still gates the three base keys in `upstream` mode (`crates/trusted-server-core/src/settings.rs:1966`, `docs/guide/auction-orchestration.md:863`)
- **Guide recipe pairs `metadata_keys` with `verbosity = "full"`**, where the field has no effect (`docs/guide/auction-orchestration.md:842`)

## CI Status

- GitHub checks: pending at review time (fresh push)
- fmt (local): PASS
- targeted rust tests, new settings + publisher tests via `cargo test-fastly` (local): PASS (11/11)

crates/trusted-server-core/src/settings.rs
prk-Jr: Fixed in b3ae41fa.

`auction_html_comment_options` now carries `skip_serializing_if = "is_default_auction_debug_comment_options"` (with `Eq, PartialEq` on the options struct), matching the `is_default_rewrite_creatives` predicate in `auction_config_types.rs`. Defaults serialize to nothing; a non-default table still serializes and requires a compatible blob before rollback, which the field docstring now states.

Regression test `default_auction_debug_comment_options_stay_out_of_serialized_config` in `settings.rs` declares a local legacy `DebugConfig` schema with `deny_unknown_fields` (the three pre-PR fields), asserts the default payload omits the table and deserializes through that schema, and asserts a non-default table still serializes.

crates/trusted-server-core/src/settings.rs
prk-Jr: Fixed in b3ae41fa.

`AuctionDebugCommentOptions::validate_metadata_keys()` rejects any entry outside `AUCTION_DEBUG_METADATA_ALLOWLIST` and names every offender plus the supported set. It is called from `Settings::prepare_runtime`, which every load path reaches through `finalize_deserialized` — including the runtime blob path at `config.rs:96` — and runs after `normalize()`, so trimming happens before validation. The render-time intersection stays untouched as defense-in-depth.

Three tests added in `settings.rs`:
- `auction_debug_comment_options_unknown_metadata_key_fails_config_load` — `["http_staus", "errors"]` fails load, error names both.
- `auction_debug_comment_options_allowlisted_metadata_keys_load` — `[" message "]` loads and normalizes to `["message"]`.
- `auction_debug_comment_options_unknown_field_fails_config_load` — the previously untested `deny_unknown_fields` case (`metadata_key = [...]`).

crates/trusted-server-core/src/settings.rs
prk-Jr: Fixed in b3ae41fa.

Confirmed: `Upstream` builds on `redacted_metadata_for_dump`, so `metadata_keys` gates the three base keys there while `verbosity` alone unlocks the six upstream diagnostics.

The field docstring now says exactly that and that `Full` ignores the field. In the guide, the table row reads "gates them in `redacted` and `upstream`, ignored in `full`", the `upstream` ladder bullet spells out the split, and the `full` bullet states the field is ignored.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Fixed in b3ae41fa.

The four private constants are now `pub(crate)` and both functions in `publisher.rs` use them: `safe_error_message` matches on the constants, and `validated_error_type` checks membership in a new `ERROR_TYPE_ALL` list next to them, so it auto-follows a new classification instead of carrying a literal copy.

`pub(crate)` alone would not close the drift hole you named — the `_ => None` arm would still swallow a new classification. New test `redacted_metadata_covers_every_orchestrator_error_type` iterates `ERROR_TYPE_ALL` and asserts each entry both validates and maps to safe wording, so adding a constant without wording fails the suite.

docs/guide/auction-orchestration.md
prk-Jr: Fixed in b3ae41fa — the `metadata_keys` line is dropped from that recipe, since `verbosity = "full"` ignores it.

The guide also gained a sentence after the options table stating that `metadata_keys` is a subset selector against the fixed allowlist and that any other entry now fails config load.

ChristianPavilonis (commented):
**Automated Review:**

## Summary

Reviewed the configurable SSAT auction HTML debug-comment implementation at `b3ae41facba97cdeab6a38161b486abc4b6ff645`. I found no additional high-confidence merge-blocking defects.

## Findings by Priority

No new findings.

## CI / Existing Reviews

All visible GitHub checks are passing. I verified the prior rollback-compatibility, metadata-key validation, error-type drift, and documentation concerns were addressed in the current head commit, so they are not duplicated here.

aram356 (approved):

ChristianPavilonis (approved):

Issue #418, opened 2026-03-04
JS bundle hash recomputed on every HTML response instead of cached


aram356:
## Summary

`concatenate_modules()` builds a full `String` of all JS modules (potentially hundreds of KB) solely to hash it, then drops it. This runs on every HTML response via `tsjs_script_src()`. Since module content is `&'static str` (compile-time `include_str!`), the hash is constant for a given set of module IDs and should be computed once. Additionally, `serve_static_with_etag` rehashes the same content for the ETag.

## Refs

- `crates/js/src/bundle.rs` lines 50-55 — `concatenated_hash` allocates full bundle to hash
- `crates/common/src/tsjs.rs` line 7 — called on every HTML response
- `crates/common/src/http_util.rs` lines 182-186 — `serve_static_with_etag` rehashes

## Recommendation

Hash modules incrementally without concatenation (stream each part into the hasher), and cache the result in a `OnceLock`.

## Context

Production readiness audit — see #396

PR #945, opened 2026-07-21
Defer initial adInit until after React hydration on Next.js App Router

### LINKED ISSUE #938: TS HTML injection breaks React hydration on Next.js App Router (React #418)
## Summary

On a Next.js **App Router** publisher, Trusted Server's injection is applied to the server-rendered HTML *after* SSR but *before* React hydrates. React compares the mutated DOM against its RSC/flight payload, they disagree, and it throws **React #418** (`Hydration failed because the server rendered HTML didn't match the client`), regenerating the affected subtree — visible as flashes/reflow. Core functionality still works because React recovers by client-rendering, but the hydration break is real.

## Confirmed root cause: `adInit` mutates ad-slot subtrees during hydration

The publisher SSR-renders each ad slot as `<div id="ad-…-container">` wrapping an inner `<div id="ad-…">` (present in both the SSR HTML and the RSC flight). TS's `adInit` calls `googletag.defineSlot(…, "${divId}-container")` on the wrapper; GPT then renders into the container and removes/replaces the publisher's inner slot div. The `</body>` bids bootstrap invoked `adInit()` **synchronously at parse time**, so that mutation landed inside React's hydration window.

## Evidence

A live A/B toggling TS on the same page via the tester cookie, plus builds isolating each candidate change:

| Build | head append | flight rewrite | defer `adInit` | React #418 |
| --- | --- | --- | --- | --- |
| Pure publisher (TS inactive) | — | — | — | **0** |
| TS active, deployed | ✗ | ✗ | ✗ | 1, 0 |
| TS active | ✓ | ✓ | ✗ | **1, 1, 2** |
| TS active | ✓ | ✓ | ✓ | 0, 0, 0, 0 |
| TS active | ✗ | ✗ | ✓ | **0, 0, 0** |

The last two rows are decisive: deferring `adInit` alone drives the count to 0, and the other two changes neither fix it nor are needed. The count tracks **whether `adInit` processes ad slots**, not the injection position.

## Fix

PR #945 — defer the initial `adInit()` until after hydration: gate on window `load`, then a double `requestAnimationFrame`. Run-once, no retry timer. The deferred callback captures the route it was scheduled for and no-ops if an SPA navigation has since committed a different route, so it cannot re-run `adInit` against the newer route's live slots/bids.

Verified end to end through the dev proxy against a live App Router publisher: **#418 = 0** across runs, with services enabled, TS container slots defined, ads rendering, and the **first** ad request carrying `ts_initial` / `hb_pb` / `hb_bidder` targeting.

## Identified but unconfirmed mechanisms

Two further mismatch sources were identified and prototyped, but neither changed the observed #418 count, and both prototypes were withdrawn as defective. They are **not** fixed. If either is ever shown to fire, a correct implementation must avoid the defects found:

1. **Head injection position** — the tsjs bundle is prepended to the React-owned `<head>`, shifting the origin's first child. Moving it to the end of `<head>` reorders origin blocking scripts ahead of TS's bootstraps, which breaks the GPT shim's first-party rewrite of dynamically loaded scripts and can install the `disableInitialLoad` detector too late. Any change here needs to separate ordering-sensitive bootstraps from hydration-sensitive trailing injections, with an execution-order test using a blocking origin GPT script.

2. **Attribute-rewrite asymmetry** — TS rewrites third-party integration hosts on rendered tags to first-party `/integrations/...` paths, but the RSC/flight rewriter is anchored to the origin host only, so the DOM and the flight disagree on those tags. A correct fix must: emit raw (non-escaped) bytes, since HTML-escaping a `<script>` payload corrupts it and desyncs recomputed T-chunk lengths; be stateful across `lol_html` text-chunk and cross-script boundaries, since a T-row split across flight scripts corrupts the stream; and target actual script `src` values / `__next_s` descriptors rather than every URL in flight data.

aram356:
## Summary

On a Next.js **App Router** publisher, `adInit()` defines GPT slots on the publisher's `-container` wrappers, mutating those ad-slot subtrees. The `</body>` bids bootstrap called it **synchronously at parse time**, landing that mutation inside React's hydration window — React threw **#418** and re-rendered the affected subtrees (visible flashes/reflow).

A live A/B, toggling TS on the same page via the tester cookie, isolated the trigger:

| Condition | React #418 |
| --- | --- |
| Pure publisher (TS inactive) | **0** |
| TS active | **1–2** |

The count tracks **whether `adInit` processes ad slots** — not the injection position. `adInit` is now deferred to after hydration: gated on window `load`, then a double `requestAnimationFrame`. Run-once, no retry timer.

The deferral lifecycle lives in the GPT bundle module as `tsjs.scheduleInitialAdInit` (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts`); the server-emitted `</body>` bids script assigns `tsjs.bids` and delegates to it. The GPT module ships in the synchronous head bundle, so it has always executed before the inline script runs.

## Scope

Deliberately reduced to the single change with a **measured** effect on #418. Two other hydration-safety changes were considered and **dropped from this PR**, because neither moved the observed #418 count and review surfaced real defects in both:

- **Appending the tsjs head bundle** instead of prepending — reordered origin blocking scripts ahead of TS's bootstraps, which breaks the GPT shim's first-party rewrite of dynamically loaded scripts and can install the `disableInitialLoad` detector too late.
- **Rewriting integration hosts in the RSC flight** — as written it HTML-escaped the serialized payload (corrupting bytes and desyncing recomputed T-chunk lengths), was not safe across `lol_html` text-chunk or cross-script boundaries, and rewrote *every* matching URL in flight data rather than only script `src` values.

A correct flight rewrite needs the existing RSC stream state machine, raw (non-escaping) insertion, and `src`-targeted parsing — that belongs in its own PR.

The gate's **timing** (window `load` is dominated by page weight — ~52s on heavy publishers) is addressed in the agreed stacked follow-up `958-adinit-hydration-chunk-gate` (spec: `docs/superpowers/specs/2026-07-24-adinit-hydration-gate-design.md`), which waits on the async `/_next/static/chunks/` hydration scripts and keeps `load` as the can't-hang fallback. Per that spec's PR strategy, this PR lands the conservative gate; the timing change gets its own PR and its own #418 A/B.

## Review findings addressed

- **Stale deferred callback vs. SPA navigation** — the deferred callback now captures `tsjs.navGeneration`, a monotonic counter the SPA auction hook increments synchronously the moment it accepts a pathname navigation, and no-ops if it moved. Unlike the earlier URL-equality guard, this cannot diverge from the hook's pathname-only route identity: a query-only `replaceState` before `load` no longer cancels the initial `adInit`, and an `/a → /b → /a` round trip (URL compares equal again) now correctly stands the stale callback down.
- **Executable lifecycle coverage** — the bootstrap moved out of the Rust format string into the bundle, so its lifecycle is executed (not string-matched) under Vitest: `test/integrations/gpt/schedule_initial_ad_init.test.ts` covers load + two-frame ordering, already-complete documents, exactly-once invocation, query-only history changes, `/a → /b → /a` cancellation, and a publisher-displayed slot receiving `setTargeting` before the refresh that delivers it. The Rust tests pin the inline script's delegation and XSS safety.
- **Browser globals shadowing** — the scheduler window-qualifies `requestAnimationFrame` / `addEventListener`.
- **Possible loss of the first targeted impression** — measured rather than assumed. Instrumenting real GPT ad requests, the **first** request (t≈3.7s) carries `ts_initial=1`, `hb_pb`, and `hb_bidder`, so server-side targeting is applied to the first impression on the tested publisher; the #958 wire investigation confirmed the TS slot's first fetch is adindex 0 with `ts_initial=1` and that publisher-owned slots are distinct. Residual risk remains in principle for a publisher whose GPT activation precedes window `load`; the follow-up chunk gate narrows that window to seconds.

### Second round (c3e8df3fd)

- **Stale SSR payload adopted after a streaming-window navigation** — the `</body>` script no longer assigns `tsjs.bids`; it hands the payload to `scheduleInitialAdInit`, which applies it and runs `adInit()` only while the page is still on navigation generation 0 (the SSR document). A navigation that commits before — or after — scheduling drops the payload and cancels the initial run, so it can neither clobber the live route's bids nor re-run it.
- **Queued GPT mutation outliving its generation check** — `adInit()` now rechecks its captured generation as the first act of the queued `googletag.cmd` callback, so a navigation that commits before GPT drains the queue (e.g. consent-gated GPT) cancels the stale slot work. Applied in the bundle and the head bootstrap.
- **Bundle-failure degradation path restored** — `gpt_bootstrap.js` installs a minimal fallback `scheduleInitialAdInit` (same generation-0 + load + double-rAF semantics), so a failed TSJS bundle load still initializes initial ads via the bootstrap's fallback `adInit`. The shipped bootstrap file is now evaluated verbatim under Vitest (`gpt_bootstrap.test.ts`).

### Third round (756cfaffe)

- **Hidden-tab behavior documented as intended** — rAF is unserviced while a document is hidden, so a background-tab load holds the initial `adInit()` until first view, spending the impression on a viewed tab; docblocks in the bundle scheduler and `gpt_bootstrap.js` now say so, and both executable suites cover the hidden case.
- **Uniform-deferral tradeoff stated in-code** — the deferral is deliberately unconditional (one code path, no framework-detection/config surface); the chunk-gate follow-up recovers the non-React latency.
- Doc fix in `types.ts` (generation-0 pinning, not counter capture) and the whole-blob `!setTimeout` assertion in `gpt.rs` dropped in favor of the executable no-retry coverage.

## Verification

- Executable Vitest lifecycle suite `schedule_initial_ad_init.test.ts` (6 tests) plus a `navGeneration` accounting test in `spa_hook.test.ts`; full JS suite **432 passed, 0 failed**, format clean, bundles build.
- Rust: `cargo test-fastly` **1734 passed, 0 failed** (+ 21 adapter tests), `cargo test-axum` green, parity suite green, `cargo fmt` clean, clippy clean on `trusted-server-core`.
- End to end through the dev proxy against a live App Router publisher: **#418 = 0**, `servicesEnabled: true`, TS container slot defined, ads rendering, and the first ad request fully targeted.

Refs #938


crates/trusted-server-core/src/publisher.rs
```diff
@@ -2169,8 +2169,31 @@ pub(crate) fn build_bids_script(bid_map: &serde_json::Map<String, serde_json::Va
...
+    // (and run its own `adInit` via the SPA auction hook) before this callback
+    // fires. The callback therefore captures the route it was scheduled for and
+    // no-ops when the route has since changed; without that guard it would run
+    // `adInit` a second time against the newer route's live slots/bids, which
+    // destroys and redefines that route's TS slots and refreshes it twice.
     format!(
-        "<script>(window.tsjs=window.tsjs||{{}}).bids=JSON.parse(\"{}\");(function(){{var f=window.tsjs.adInit;if(typeof f===\"function\")f();}})();</script>",
+        "<script>(window.tsjs=window.tsjs||{{}}).bids=JSON.parse(\"{}\");\
+(function(){{\
+var p=location.pathname+location.search;\
```
ChristianPavilonis: 🔧 **P1 — URL equality is not a safe navigation identity**

The guard compares `pathname + search`, while the SPA auction hook intentionally identifies routes using only `pathname` (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts:715-724`). This creates two failures:

- A query-only `pushState`/`replaceState` before `load` makes this callback abort, but the SPA hook also declines to auction because the pathname is unchanged. Initial `adInit()` is therefore never called.
- An `/a → /b → /a` transition before this callback defeats the stale-route check because the URL equals `/a` again. The SPA hook auctions the new `/a` route and calls `adInit()`, after which this initial callback can call it again, destroying/redefining slots and issuing duplicate requests.

This can leave all initial ads uninitialized after query normalization or double-refresh inventory after a rapid round trip. Please use a shared monotonic navigation generation maintained synchronously by the SPA hook, capture that generation here, and run only if it is unchanged. Keep query-only changes ignored unless `/__ts/page-bids` and the SPA hook are deliberately changed to treat them as auction-relevant.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -2169,8 +2169,31 @@ pub(crate) fn build_bids_script(bid_map: &serde_json::Map<String, serde_json::Va
...
+var f=function(){{\
+if(location.pathname+location.search!==p)return;\
+var a=window.tsjs.adInit;if(typeof a===\"function\")a();}};\
+var d=function(){{window.requestAnimationFrame(function(){{window.requestAnimationFrame(f);}});}};\
```
ChristianPavilonis: 🔧 **P1 — The global `load` gate moves TS targeting behind normal publisher GPT requests**

This generic bootstrap now waits for every load-blocking resource and then two animation frames. A publisher that defines/displays GPT before or during `window.load` sends its first request before TS targeting is applied. `adInit()` subsequently finds the publisher-owned slot (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts:512-518`) and refreshes it (`index.ts:621-633`), turning the TS-targeted request into a second impression. This behavior applies to every creative-opportunity page, not only the tested Next.js App Router publisher.

That can lose the server-auction bid on the first impression, create a duplicate request, delay TS-owned slots behind unrelated resources, and age the auction result. Testing one publisher whose first request happened after this callback does not establish compatibility with other supported initialization patterns.

Please preserve pre-request targeting rather than delaying the entire operation: defer only the hydration-unsafe TS-owned slot definition/display work while intercepting or targeting existing publisher slots before their first `display()`. If that cannot be done safely here, make delayed initialization an explicit App Router/publisher opt-in and retain synchronous behavior elsewhere.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -4695,6 +4718,58 @@ mod tests {
             );
         }
 
+        #[test]
+        fn bids_script_defers_ad_init_until_after_hydration() {
```
ChristianPavilonis, line 8129: 🔧 **P2 — This test does not execute the lifecycle being changed**

The assertions only check for substrings such as `requestAnimationFrame`, `"load"`, and `location.pathname`. The test would still pass if `adInit()` ran synchronously, if only one frame were used, or with both navigation bugs noted above. The central guarantees—post-load ordering, exactly-once invocation, stale-navigation cancellation, and interaction with the SPA hook—remain unprotected.

Please add an executable browser/JavaScript test covering loading and already-complete documents, two-frame ordering, query-only history changes, `/a → /b → /a`, and a publisher GPT `display()` during `load`. Assert both the `adInit()` count and first-request targeting.

ChristianPavilonis (changes requested):
## Review summary

🔧 Requesting changes. The diff is narrow, but it changes initial ad initialization for every matched-slot publisher and introduces high-risk navigation and GPT request-ordering regressions. CI is green, but the current automated coverage does not exercise these lifecycle paths.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -2180,8 +2180,31 @@ pub(crate) fn build_bids_script(bid_map: &serde_json::Map<String, serde_json::Va
...
+    // (and run its own `adInit` via the SPA auction hook) before this callback
+    // fires. The callback therefore captures the route it was scheduled for and
+    // no-ops when the route has since changed; without that guard it would run
+    // `adInit` a second time against the newer route's live slots/bids, which
+    // destroys and redefines that route's TS slots and refreshes it twice.
     format!(
-        "<script>(window.tsjs=window.tsjs||{{}}).bids=JSON.parse(\"{}\");(function(){{var f=window.tsjs.adInit;if(typeof f===\"function\")f();}})();</script>",
+        "<script>(window.tsjs=window.tsjs||{{}}).bids=JSON.parse(\"{}\");\
+(function(){{\
+var p=location.pathname+location.search;\
```
prk-Jr: 🔧 **wrench — URL equality is not a safe navigation identity**

This guard uses `pathname + search`, while the SPA auction hook intentionally identifies routes using only `pathname` (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts:715-724`). That produces two failures:

- A query-only `pushState`/`replaceState` before `load` makes this callback abort, but the SPA hook also declines to auction because the pathname is unchanged. Initial `adInit()` is never called.
- An `/a → /b → /a` transition before this callback defeats the stale-route check because the URL equals `/a` again. After the SPA hook applies the new `/a` state and calls `adInit()`, this callback can call it again, destroying/redefining slots and issuing duplicate requests.

Please use a shared monotonic navigation generation maintained synchronously by the SPA hook, capture it here, and run only if it remains unchanged. Query-only changes should remain ignored unless the SPA auction path is deliberately changed to treat them as auction-relevant.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -2180,8 +2180,31 @@ pub(crate) fn build_bids_script(bid_map: &serde_json::Map<String, serde_json::Va
...
+var f=function(){{\
+if(location.pathname+location.search!==p)return;\
+var a=window.tsjs.adInit;if(typeof a===\"function\")a();}};\
+var d=function(){{window.requestAnimationFrame(function(){{window.requestAnimationFrame(f);}});}};\
```
prk-Jr: 🔧 **wrench — The global `load` gate can move targeting behind the first GPT request**

This generic bootstrap now waits for every load-blocking resource and two animation frames. A publisher that defines/displays GPT before or during `window.load` can send its first request before TS targeting is applied. The later `adInit()` finds the publisher-owned slot (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts:512-518`) and refreshes it (`index.ts:621-633`), making the TS-targeted request a second impression.

That can lose the server-auction bid on the first impression, create a duplicate request, delay TS-owned slots behind unrelated resources, and age the auction result. Please preserve pre-request targeting—e.g. defer only hydration-unsafe TS-owned slot definition/display work—or make delayed initialization an explicit App Router/publisher opt-in while retaining synchronous behavior elsewhere.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -4709,6 +4732,58 @@ mod tests {
             );
         }
 
+        #[test]
+        fn bids_script_defers_ad_init_until_after_hydration() {
```
prk-Jr, line 8129: 🔧 **wrench — This test does not execute the lifecycle being changed**

These assertions only check for substrings such as `requestAnimationFrame`, `"load"`, and `location.pathname`. They would still pass if `adInit()` ran synchronously, if only one frame were used, or with both navigation failures above. The core guarantees—post-load ordering, exactly-once invocation, stale-navigation cancellation, and interaction with publisher GPT—remain unprotected.

Please add executable browser/JavaScript coverage for loading and already-complete documents, two-frame ordering, query-only history changes, `/a → /b → /a`, and a publisher GPT `display()` during `load`. Assert both the `adInit()` count and first-request targeting.

prk-Jr (changes requested):
## Summary

The hydration mitigation is narrow, but it changes initial ad initialization for every matched-slot publisher and leaves navigation and GPT request-ordering regressions. CI is green; the current automated coverage does not execute these lifecycle paths.

## Blocking

- 🔧 Use a real navigation generation instead of URL equality.
- 🔧 Preserve first-request GPT targeting instead of globally delaying all initialization.
- 🔧 Add executable lifecycle coverage for the deferred bootstrap.

## CI Status

- fmt: PASS
- clippy/checks: PASS
- rust tests: PASS
- JavaScript tests: PASS

crates/trusted-server-core/src/publisher.rs
aram356: Implemented in 6f9dbff78. The SPA auction hook now maintains `tsjs.navGeneration` — a monotonic counter incremented synchronously the moment a pathname navigation is accepted (before the page-bids fetch) — and the deferred initial-adInit callback captures it at schedule time and stands down if it has moved.

Query-only history changes remain ignored by both the hook and the guard, so the initial `adInit()` still runs after a query-only `replaceState`; an `/a → /b → /a` round trip advances the counter twice, so the stale callback no-ops. Both scenarios have executable coverage in `schedule_initial_ad_init.test.ts`, plus a counter-accounting test in `spa_hook.test.ts`.

crates/trusted-server-core/src/publisher.rs
aram356: Implemented in 6f9dbff78. The SPA auction hook now maintains `tsjs.navGeneration` — a monotonic counter incremented synchronously the moment a pathname navigation is accepted (before the page-bids fetch) — and the deferred initial-adInit callback captures it at schedule time and stands down if it has moved.

Query-only history changes remain ignored by both the hook and the guard, so the initial `adInit()` still runs after a query-only `replaceState`; an `/a → /b → /a` round trip advances the counter twice, so the stale callback no-ops. Both scenarios have executable coverage in `schedule_initial_ad_init.test.ts`, plus a counter-accounting test in `spa_hook.test.ts`.

crates/trusted-server-core/src/publisher.rs
aram356: Two separable points here.

Refreshing a publisher-owned slot to deliver TS targeting is pre-existing behavior, not introduced by this PR — before this change `adInit()` ran synchronously at body-parse time and refreshed publisher-owned slots the same way whenever the publisher's initial load had already fired. This PR changes only how long that window stays open. On the live App Router publisher this was measured against, the TS slot's first GPT request (adindex 0) carries `ts_initial=1` + `hb_*`, and the wire investigation for #958 confirmed the publisher's own slots are distinct from TS's slots (the early pre-SSAT fetch belonged to the publisher).

The gate timing is addressed in the agreed stacked follow-up (`958-adinit-hydration-chunk-gate`, spec at `docs/superpowers/specs/2026-07-24-adinit-hydration-gate-design.md`): wait for the async `/_next/static/chunks/` hydration scripts (a few seconds) instead of `window.load` (~52s on heavy pages), with `load` kept as the can't-hang fallback — effectively the App Router-specific fast path this comment asks for, with no new config surface. That spec's PR strategy is to land #945 with the conservative gate and make the timing change in its own PR with its own #418 A/B evidence, so the gate here is unchanged. A sync-targeting/deferred-DOM split of `adInit()` is the spec's explicit non-goal (the "hydration-safe adInit rewrite" — a larger change, separate effort).

What did land here for this finding's test ask: the deferral now lives in the GPT bundle (`tsjs.scheduleInitialAdInit`, commit 6f9dbff78), and an executable test pins that a publisher-displayed slot receives `setTargeting` before the refresh that delivers it.

crates/trusted-server-core/src/publisher.rs
aram356: Two separable points here.

Refreshing a publisher-owned slot to deliver TS targeting is pre-existing behavior, not introduced by this PR — before this change `adInit()` ran synchronously at body-parse time and refreshed publisher-owned slots the same way whenever the publisher's initial load had already fired. This PR changes only how long that window stays open. On the live App Router publisher this was measured against, the TS slot's first GPT request (adindex 0) carries `ts_initial=1` + `hb_*`, and the wire investigation for #958 confirmed the publisher's own slots are distinct from TS's slots (the early pre-SSAT fetch belonged to the publisher).

The gate timing is addressed in the agreed stacked follow-up (`958-adinit-hydration-chunk-gate`, spec at `docs/superpowers/specs/2026-07-24-adinit-hydration-gate-design.md`): wait for the async `/_next/static/chunks/` hydration scripts (a few seconds) instead of `window.load` (~52s on heavy pages), with `load` kept as the can't-hang fallback — effectively the App Router-specific fast path this comment asks for, with no new config surface. That spec's PR strategy is to land #945 with the conservative gate and make the timing change in its own PR with its own #418 A/B evidence, so the gate here is unchanged. A sync-targeting/deferred-DOM split of `adInit()` is the spec's explicit non-goal (the "hydration-safe adInit rewrite" — a larger change, separate effort).

What did land here for this finding's test ask: the deferral now lives in the GPT bundle (`tsjs.scheduleInitialAdInit`, commit 6f9dbff78), and an executable test pins that a publisher-displayed slot receives `setTargeting` before the refresh that delivers it.

crates/trusted-server-core/src/publisher.rs
aram356: Addressed in 6f9dbff78. The bootstrap moved out of the Rust-emitted inline string into the GPT bundle (`scheduleInitialAdInit` in `gpt/index.ts`); the `</body>` bids script now only assigns `tsjs.bids` and delegates to it.

The lifecycle now runs for real under Vitest/jsdom in `test/integrations/gpt/schedule_initial_ad_init.test.ts` — the real scheduler plus, where relevant, the real `adInit()` and SPA auction hook: load + two-frame ordering, already-complete documents, exactly-once invocation across duplicate load events, query-only history changes, `/a → /b → /a` cancellation, and a publisher-displayed slot receiving `setTargeting` before the refresh that delivers it. The Rust tests now pin only the inline script's delegation and XSS safety.

Real first-request GPT targeting stays verified by the live A/B in the PR description — an actual GPT ad request is not reachable from jsdom.

crates/trusted-server-core/src/publisher.rs
aram356: Addressed in 6f9dbff78. The bootstrap moved out of the Rust-emitted inline string into the GPT bundle (`scheduleInitialAdInit` in `gpt/index.ts`); the `</body>` bids script now only assigns `tsjs.bids` and delegates to it.

The lifecycle now runs for real under Vitest/jsdom in `test/integrations/gpt/schedule_initial_ad_init.test.ts` — the real scheduler plus, where relevant, the real `adInit()` and SPA auction hook: load + two-frame ordering, already-complete documents, exactly-once invocation across duplicate load events, query-only history changes, `/a → /b → /a` cancellation, and a publisher-displayed slot receiving `setTargeting` before the refresh that delivers it. The Rust tests now pin only the inline script's delegation and XSS safety.

Real first-request GPT targeting stays verified by the live A/B in the PR description — an actual GPT ad request is not reachable from jsdom.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -449,9 +449,54 @@ function installInitialLoadDetector(ts: TsjsApi): void {
...
+ * and no-ops when a navigation has committed since — running anyway would
+ * re-run `adInit()` against the newer route's live slots/bids, destroying and
+ * redefining that route's TS slots and double-refreshing it. The generation
+ * counter (not a URL comparison) keeps this guard aligned with the SPA
+ * auction hook's own navigation identity: a query-only history change the
+ * hook ignores must not cancel the initial call, while an `/a → /b → /a`
+ * round trip — where the URL compares equal again — must.
+ */
+function installScheduleInitialAdInit(ts: TsjsApi): void {
+  ts.scheduleInitialAdInit = function () {
```
ChristianPavilonis: 🔧 **P1 — A navigation before the body-end bootstrap is incorrectly adopted as the initial generation**

`scheduleInitialAdInit()` captures `navGeneration` only when the closing-body script calls it. The SPA hook is already installed by the synchronous head bundle, so a pathname navigation can advance the generation while the HTML is still streaming. The scheduler then captures that advanced value and considers the stale initial bootstrap current. In addition, `build_bids_script()` unconditionally overwrites `tsjs.bids` with the original SSR bids before scheduling; if the SPA response already applied, this corrupts the new route's live bid state before `adInit()` runs again. This can mix new-route slots with old-route bids, break creative/beacon lookup, destroy new TS-owned slots, and issue an additional refresh.

Please tie the initial payload to the original document generation rather than capturing the generation at body end. Prefer passing the parsed initial bids into the scheduler and assigning them only if the original generation is still current. Add coverage where navigation—and separately the page-bids response—occurs before `scheduleInitialAdInit()` is called.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -449,9 +449,54 @@ function installInitialLoadDetector(ts: TsjsApi): void {
...
+    const generation = ts.navGeneration ?? 0;
+    const runUnlessNavigated = (): void => {
+      if ((ts.navGeneration ?? 0) !== generation) return;
```
ChristianPavilonis: 🔧 **P1 — Generation cancellation ends before the queued GPT mutation executes**

This check protects the call to `adInit()`, but `adInit()` only queues its destructive slot work through `googletag.cmd`. If GPT is loaded lazily—for example after consent—a navigation can advance `navGeneration` after this check but before GPT drains the queue. The stale callback then applies its captured slots and bids before the newer route callback, potentially targeting the new DOM with old data and issuing duplicate ad requests. The lifecycle test uses an immediately executing `cmd.push` mock, so it cannot expose this ordering.

Capture the generation in each `adInit()` invocation and recheck it as the first operation inside the `g.cmd.push` callback. Add a deferred-command-queue test that enqueues initialization, navigates, drains the queue, and verifies that the stale callback performs no GPT mutations or refresh.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -3314,8 +3314,25 @@ pub(crate) fn build_bids_script(bid_map: &serde_json::Map<String, serde_json::Va
...
+    // Vitest (schedule_initial_ad_init.test.ts) and the navigation-generation
+    // guard is shared with the SPA auction hook. The GPT module ships in the
+    // synchronous head bundle, so it has always executed by the time this
+    // inline script runs; when the GPT integration is disabled the scheduler is
+    // absent — but so is `adInit`, so there is nothing to schedule.
     format!(
-        "<script>(window.tsjs=window.tsjs||{{}}).bids=JSON.parse(\"{}\");(function(){{var f=window.tsjs.adInit;if(typeof f===\"function\")f();}})();</script>",
+        "<script>(window.tsjs=window.tsjs||{{}}).bids=JSON.parse(\"{}\");\
+(function(){{\
+var s=window.tsjs.scheduleInitialAdInit;\
```
ChristianPavilonis: 🔧 **P2 — A TSJS bundle load failure now bypasses the existing initial-ad fallback**

When `scheduleInitialAdInit` is absent, this body script silently does nothing. The earlier head bootstrap still defines a functional `tsjs.adInit`, but it does not define the new scheduler. Previously the body script called that fallback directly. Therefore a failed main TSJS bundle request, or an exception before the GPT module evaluates, now prevents initial server-side ads from initializing at all. This defeats the purpose of the standalone head bootstrap's degradation path.

Please provide the hydration-safe scheduler in the head bootstrap as well, or persist a one-shot initialization request that the bundle consumes while retaining a safe fallback if the bundle never arrives. Add coverage for a missing or failed main bundle.

ChristianPavilonis (approved):
## Summary

Reviewed the hydration deferral, navigation-generation lifecycle, GPT command-queue behavior, head/body injection ordering, fallback bootstrap, and executable coverage. CI is green and the earlier URL-equality and string-only-test concerns were addressed. I found two high-priority lifecycle race windows and one medium-priority fallback regression, detailed inline.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
aram356: Implemented in c3e8df3fd, as suggested. The `</body>` script no longer assigns `tsjs.bids` — it hands the parsed payload to `scheduleInitialAdInit`, and the scheduler applies it (and later runs `adInit()`) only while the page is still on navigation generation 0, which is the SSR document's generation by definition. If a navigation committed while the document was streaming, the payload is dropped entirely: the live route's bids are not overwritten and the initial callback never fires.

Executable coverage added for both interleavings: navigation-before-scheduling (payload dropped, `adInit` never runs) and page-bids-response-applied-before-scheduling (applied slots/bids stay intact, `adInit` count stays at the SPA hook's single call).

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
aram356: Implemented in c3e8df3fd. `adInit()` captures its generation at invocation and rechecks it as the first statement of the queued `googletag.cmd` callback, so a navigation that commits between the call-site check and GPT's queue drain cancels the stale mutation — no `defineSlot`/`destroySlots`/`refresh`/`enableSingleRequest` runs. Applied to both the bundle `adInit` and the head-bootstrap fallback.

The new test uses a real non-executing command-queue array: enqueue via `adInit()`, navigate (generation advances), drain the queue, and assert the stale callback performs no GPT mutations or refresh.

crates/trusted-server-core/src/publisher.rs
aram356: Restored in c3e8df3fd. `gpt_bootstrap.js` now installs a minimal fallback `scheduleInitialAdInit` mirroring the bundle's semantics (generation-0 payload adoption, window `load` + double rAF, generation recheck), so a failed bundle load still initializes initial server-side ads through the bootstrap's fallback `adInit`. The bootstrap's existing convention is kept: it stands down entirely when the bundle already installed `adInit`, and the bundle overwrites both functions when it loads later.

Coverage for the degradation path is executable: `gpt_bootstrap.test.ts` evaluates the shipped bootstrap file verbatim under jsdom — fallback install, no-override when the bundle is present, payload + deferral, generation stand-down, define/target/display through the command queue, and queue-drain cancellation.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -449,19 +449,80 @@ function installInitialLoadDetector(ts: TsjsApi): void {
...
+  ts.scheduleInitialAdInit = function (initialBids?: Record<string, AuctionBidData>) {
+    if ((ts.navGeneration ?? 0) !== 0) return;
+    if (initialBids) ts.bids = initialBids;
+    const runUnlessNavigated = (): void => {
+      if ((ts.navGeneration ?? 0) !== 0) return;
+      ts.adInit?.();
+    };
+    const afterHydrationFrames = (): void => {
+      window.requestAnimationFrame(() => {
+        window.requestAnimationFrame(runUnlessNavigated);
```
prk-Jr, line 546: ❓ **question** — Hidden-tab consequence: was it intended?

`load` fires in background tabs, but `requestAnimationFrame` callbacks are not serviced while the document is hidden. They survive and run on first visibility, so a Cmd+click / open-in-new-tab load now issues **zero** initial ad requests until the tab is first viewed — previously the request went out at body-parse time. Served impressions for tabs that are never viewed drop to zero.

That may well be the better outcome for viewability, but it is not mentioned in the docblock and neither `schedule_initial_ad_init.test.ts` nor `gpt_bootstrap.test.ts` covers `document.hidden`.

If intended: worth a sentence in the docblock so the next reader does not treat it as a bug. If not, the hidden case needs its own path, e.g.

```ts
const afterHydrationFrames = (): void => {
  // rAF is not serviced while the document is hidden, so a background-tab
  // load would not reach adInit until first view. React schedules hydration
  // off MessageChannel/microtasks, not rAF, so frames buy nothing here.
  if (document.hidden) {
    setTimeout(runUnlessNavigated, 0);
    return;
  }
  window.requestAnimationFrame(() => {
    window.requestAnimationFrame(runUnlessNavigated);
  });
};
```

Note that collides with the `!contains("setTimeout")` assertions in `publisher.rs` and `gpt.rs:1234`, which currently forbid any timer rather than a *retry* timer.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -42,12 +42,43 @@
...
+  ts.scheduleInitialAdInit = function (initialBids) {
+    if ((ts.navGeneration || 0) !== 0) return;
+    if (initialBids) ts.bids = initialBids;
+    var fire = function () {
+      if ((ts.navGeneration || 0) !== 0) return;
+      if (typeof ts.adInit === "function") ts.adInit();
+    };
+    var afterFrames = function () {
+      window.requestAnimationFrame(function () {
+        window.requestAnimationFrame(fire);
```
prk-Jr, line 106: ❓ **question** — Same hidden-document gap as the bundle scheduler (see the comment on `gpt/index.ts`): while the document is hidden these frames are not serviced, so a background-tab load defers the initial `adInit()` until first view. Whatever resolution the bundle takes, this mirror needs the same one — the fallback is exactly the path that runs when the bundle failed to load, so it must not be the weaker of the two.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -3314,8 +3314,34 @@ pub(crate) fn build_bids_script(bid_map: &serde_json::Map<String, serde_json::Va
...
+    // assignment would clobber the live route's bids with the stale SSR
+    // payload. Only when no scheduler exists at all (GPT integration active
+    // without its head bootstrap — not an expected deployment) does the script
+    // fall back to a plain assignment, where no SPA hook exists to race with.
     format!(
-        "<script>(window.tsjs=window.tsjs||{{}}).bids=JSON.parse(\"{}\");(function(){{var f=window.tsjs.adInit;if(typeof f===\"function\")f();}})();</script>",
+        "<script>(function(){{\
+var t=window.tsjs=window.tsjs||{{}};\
+var b=JSON.parse(\"{}\");\
+var s=t.scheduleInitialAdInit;\
```
prk-Jr, line 3351: ❓ **question** — The deferral is unconditional; the bug it fixes is not.

This moves the initial ad request from body-parse to window `load` for **every** publisher on **every** page, including non-React publishers with no React #418 exposure. `load` waits on the slowest subresource, which on an ad-heavy page is seconds, not frames.

Was gating considered — the `nextjs` integration being active, a hydration-sensitive framework signal, or a config toggle — so publishers who never had the hydration mismatch keep the parse-time request? If the global deferral is deliberate (uniform behavior, easier to reason about), that tradeoff is worth stating in this comment, since the comment currently justifies the deferral purely in terms of the Next.js App Router failure.

crates/trusted-server-js/lib/test/integrations/gpt/schedule_initial_ad_init.test.ts
```diff
@@ -0,0 +1,366 @@
...
+    await flushAsync();
+    expect(ts.navGeneration).toBe(2);
+
+    window.dispatchEvent(new Event('load'));
+    flushFrame();
+    flushFrame();
+    expect(adInit).not.toHaveBeenCalled();
+  });
+
+  it('applies server targeting to a publisher-displayed slot before its refresh', async () => {
```
prk-Jr, line 199: ❓ **question** — This test encodes "publisher already defined and `display()`ed its slot before window `load`" as the expected state, which is right — but it makes the widened window explicit.

GPT is ready long before `load`, so the gap between the publisher's initial *untargeted* ad request and TS's targeted `refresh()` grows from *GPT-ready* to *window load*. More slots will therefore render untargeted first, and the first render per slot is the only one that can claim `ssat` — later ones are `gam-refresh`.

Do the deployed render numbers show the SSAT-claim rate holding after this change? If the initial render flips to the publisher's untargeted one for a meaningful share of slots, the deferral trades a hydration warning for attribution.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -824,6 +892,7 @@ export function installSpaAuctionHook(): void {
   async function onNavigate(path: string): Promise<void> {
     if (path === currentPath) return;
     currentPath = path;
+    ts.navGeneration = (ts.navGeneration ?? 0) + 1;
```
prk-Jr, line 949: 🤔 **thinking** — Any pathname-changing history call before window `load` — including one from a third-party script during hydration, not just a real user navigation — bumps the generation and permanently cancels the initial pass. The no-rollback decision is documented and correct on its own terms, but recovery then rests entirely on that navigation's page-bids succeeding: the `if (!data)` branch below returns without calling `adInit()`, so the route gets no TS ads at all and the cancelled SSR pass cannot be revived.

Pre-change the SSR pass had already rendered by then, so this failure mode is new. Not blocking — the window is small and the trigger is unusual — but if you have a way to distinguish "framework committed a route change" from "someone called `pushState`", it would tighten the guard.

crates/trusted-server-js/lib/src/core/types.ts
```diff
@@ -139,4 +139,28 @@ export interface TsjsApi {
   gptInitialLoadDisabled?: boolean;
   /** Guards SPA pushState hook installation. */
   spaHookInstalled?: boolean;
+  /**
+   * Monotonic count of committed SPA navigations, incremented synchronously by
+   * the SPA auction hook the moment it accepts a route change. The deferred
+   * initial-adInit bootstrap ([`scheduleInitialAdInit`]) captures this counter
```
prk-Jr: ⛏ **nitpick** — Doc/impl mismatch: the scheduler does not *capture* the counter, it pins to generation 0 unconditionally (`if ((ts.navGeneration ?? 0) !== 0) return;`). The distinction is the whole point of the design — capturing an advanced value would adopt the stale SSR bootstrap as current — and the `installScheduleInitialAdInit` docblock states it correctly. Suggest matching it here:

```
   * initial-adInit bootstrap ([`scheduleInitialAdInit`]) is pinned to
   * generation 0 (the SSR document) and no-ops when a navigation has
   * committed — before it was called, or while it was pending.
```

crates/trusted-server-core/src/integrations/gpt.rs
```diff
@@ -1196,6 +1196,46 @@ mod tests {
...
+        assert!(
+            combined.contains("requestAnimationFrame"),
+            "fallback scheduler should defer past hydration frames"
+        );
+        assert!(
+            combined.contains("\"load\""),
+            "fallback scheduler should gate on window load"
+        );
+        assert!(
+            !combined.contains("setTimeout"),
```
prk-Jr: ⛏ **nitpick** — This negative assertion covers the whole joined head-insert output, not just the fallback scheduler. Any future unrelated `setTimeout` anywhere in `gpt_bootstrap.js` (or another GPT head insert) fails this test with "fallback scheduler must not retry on a timer", which will send the next reader to the wrong place.

The surrounding asserts are string-matching a JS blob, which the comment already concedes — real coverage is `gpt_bootstrap.test.ts`, which now *executes* the fallback. Either scope this to the scheduler substring or drop it and let the Vitest suite own the no-retry property.

crates/trusted-server-js/lib/test/integrations/gpt/gpt_bootstrap.test.ts
```diff
@@ -0,0 +1,209 @@
...
+type AnyRecord = Record<string, any>;
+
+type TestWindow = Window & {
+  googletag?: AnyRecord;
+  tsjs?: AnyRecord;
+};
+
+function runBootstrap(): void {
+  // Evaluate in the jsdom global scope, exactly as an inline <script> would.
+  new Function(BOOTSTRAP_SOURCE)();
```
prk-Jr, line 33: 👍 **praise** — Evaluating the shipped `gpt_bootstrap.js` verbatim in jsdom is the right call. The head-injected degradation path is the one that runs precisely when everything else has failed, and until now it was only ever string-matched from Rust. Now it is executed — including generation cancellation across the `googletag.cmd` gap.

Same praise for choosing a monotonic counter over URL comparison, with both the query-only change and the `/a → /b → /a` round trip covered as tests rather than as prose.

prk-Jr (approved):
## Summary

The `</body>` bids script stops calling `adInit()` synchronously and hands the SSR payload to `tsjs.scheduleInitialAdInit`, which applies bids and runs `adInit()` after window `load` plus a double `requestAnimationFrame`, pinned to navigation generation 0. The new `tsjs.navGeneration` counter is bumped synchronously in the SPA hook and rechecked inside `adInit`'s `googletag.cmd` callback, so work queued before a navigation stands down instead of mutating the newer route's DOM.

The design is sound and the reasoning is unusually well documented at every layer — including why a counter beats URL comparison, why the pass pins to generation 0 rather than capturing, and why a failed navigation deliberately does not roll the counter back. Approving: nothing here blocks merge. The questions below are about consequences I could not verify from the diff, not defects.

## Blocking

_None._

## Non-blocking

### ❓ question

- **Hidden-tab rAF suspension is an unstated behavior change**: `load` fires in background tabs but `requestAnimationFrame` does not run while hidden, so an open-in-new-tab load now makes zero initial ad requests until first view (`gpt/index.ts:493`, mirrored in `gpt_bootstrap.js:63`). Intended, or does the hidden case need its own path?
- **The deferral is global; the bug is Next.js App Router-specific**: every publisher's initial ad request moves from body-parse to window `load`, including those with no React #418 exposure (`publisher.rs:3341`). Was gating considered?
- **Widened untargeted-request window vs. SSAT single-shot attribution**: the gap between the publisher's initial untargeted request and TS's targeted `refresh()` grows from GPT-ready to window `load`, and only the first render per slot can claim `ssat` (`schedule_initial_ad_init.test.ts:198`). Do the deployed numbers hold?

### 🤔 thinking

- **A third-party `pushState` before `load` permanently cancels the initial pass** (`gpt/index.ts:895`): recovery then depends entirely on that navigation's page-bids succeeding, and the `!data` branch returns without `adInit()`. Pre-change the SSR pass had already rendered, so this failure mode is new.

### ⛏ nitpick

- **Doc/impl mismatch** (`core/types.ts:145`): says the scheduler "captures this counter"; it pins to generation 0. The `installScheduleInitialAdInit` docblock has it right.
- **Over-broad negative assertion** (`integrations/gpt.rs:1234`): `!contains("setTimeout")` spans all GPT head inserts, so an unrelated future timer fails it with a misleading message.

### 👍 praise

- **`gpt_bootstrap.test.ts` evaluates the shipped bootstrap verbatim** — the head-injected degradation path is now executed rather than string-matched from Rust, generation cancellation across the `googletag.cmd` gap included.
- **Generation counter over URL comparison**, with the query-only change and the `/a → /b → /a` round trip both covered as executable tests.

## CI Status

- fmt: PASS
- clippy / cargo check (fastly, axum, cloudflare native + wasm, spin native + wasm): PASS
- rust tests (fastly, axum, cloudflare, spin, ts CLI, cross-adapter parity): PASS
- js tests (vitest) + format-typescript + format-docs: PASS
- integration tests (browser, Fastly EC lifecycle): pending at time of review

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
aram356: Intended — and now documented and tested in 756cfaffe. The docblock states the behavior and the rationale: a background-tab load holds the initial `adInit()` until first view, so the impression is spent on a tab someone is actually looking at rather than firing unviewable at parse time in a tab that may never be foregrounded. Staying on rAF also keeps a single code path whose post-hydration-commit guarantee holds whenever the request is actually issued — a `document.hidden` `setTimeout(0)` branch would trade that for a weaker "after load, hopefully after hydration" guarantee precisely in the throttled case, and would reintroduce a timer path the tests forbid.

Both suites now cover the hidden case executably: frames queue but stay unserviced while hidden (adInit waits), then fire exactly once when frames are first serviced.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
aram356: Resolved identically to the bundle scheduler in 756cfaffe — documented as intended in the bootstrap comment, with the same executable hidden-document test in `gpt_bootstrap.test.ts`, so the fallback and the bundle stay behavior-identical on this axis.

crates/trusted-server-core/src/publisher.rs
aram356: Gating was considered and deliberately not added here; 756cfaffe states the tradeoff in the comment itself. Uniform deferral keeps one code path and avoids a framework-detection or per-publisher config surface that has to be kept truthful; the cost — non-React pages moving the initial request from parse time to `load` — is real and now stated. The agreed follow-up (`958-adinit-hydration-chunk-gate`, spec in `docs/superpowers/specs/2026-07-24-adinit-hydration-gate-design.md`) narrows the gate to the Next.js hydration chunks with `load` as the can't-hang fallback, which recovers most of that latency with no new config surface — that's where the conditionality lands.

crates/trusted-server-js/lib/test/integrations/gpt/schedule_initial_ad_init.test.ts
aram356: No deployed post-change claim-rate numbers exist yet — this branch hasn't shipped beyond the dev-proxy A/B, where the TS slot's first GPT request (adindex 0) still carried `ts_initial=1`. The widened publisher-request-to-refresh window is real, and SSAT-claim rate is exactly what the canary rollout should watch (the follow-up spec's acceptance criteria already pin "first targeted request still carries ts_initial=1" as blocking). If the canary shows initial renders flipping to the publisher's untargeted request for a meaningful share of slots, the chunk gate — which moves the refresh from ~load to a few seconds — is the mitigation already in flight.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
aram356: Agreed on the mechanics, and it's a deliberate accept: any pathname-changing history call is indistinguishable from a framework navigation at the History API level (Navigation API would distinguish traversal sources, but isn't available cross-browser), so the guard treats them alike. Two things bound the damage: the hook rolls `currentPath` back to the last applied route on a failed page-bids, so the next real navigation — including back to the affected route — retries and re-auctions; and the trigger requires a third-party script to rewrite the pathname (not just query/hash, which are ignored) inside the streaming-to-load window. If this shows up in practice, a page-bids retry on the committed route would be the tightening I'd reach for, rather than trying to classify pushState callers.

crates/trusted-server-js/lib/src/core/types.ts
aram356: Fixed in 756cfaffe with the suggested wording — the bootstrap is pinned to generation 0, it does not capture the counter.

crates/trusted-server-core/src/integrations/gpt.rs
aram356: Dropped in 756cfaffe. The no-retry property is owned by the executable suite (`gpt_bootstrap.test.ts` asserts adInit runs exactly once, including in the new hidden-document test), and a comment in the Rust test now says so, so a future unrelated timer in the head inserts can't be misattributed to the scheduler.

PR #948, opened 2026-07-22
Detect GPT initial load configured through setConfig

### LINKED ISSUE #946: Detect GPT initial load configured through setConfig
## Problem

Publishers can disable GPT initial loading with the modern `googletag.setConfig({ disableInitialLoad: true })` API. Trusted Server currently detects only the legacy `googletag.pubads().disableInitialLoad()` method.

When the modern API is used, `window.tsjs.gptInitialLoadDisabled` remains unset. A TS-owned slot is registered with `display()`, but it is omitted from the follow-up `refresh()`, so GPT never emits the GAM request and reports a fetch count of zero.

## Acceptance criteria

- Detect initial-load disabling through both GPT configuration APIs.
- Refresh a TS-owned slot after `display()` when initial loading is disabled.
- Avoid adding a second request when initial loading remains enabled.
- Cover the modern `setConfig()` path with regression tests.

ChristianPavilonis:
## Summary

- Detect GPT initial-load disabling through the modern `googletag.setConfig({ disableInitialLoad: true })` API as well as the legacy pubads method.
- Ensure TS-owned slots receive the required `refresh()` after `display()`, preventing GPT slots from stopping at fetch count zero.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/integrations/gpt_bootstrap.js` | Wrap both GPT initial-load configuration APIs in the early page bootstrap. |
| `crates/trusted-server-js/lib/src/integrations/gpt/index.ts` | Add equivalent detection to the bundled GPT integration and preserve the existing refresh behavior. |
| `crates/trusted-server-js/lib/src/core/types.ts` | Document both supported initial-load configuration paths. |
| `crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts` | Add regression coverage for `googletag.setConfig()`. |
| `crates/trusted-server-core/src/integrations/gpt.rs` | Assert that the injected bootstrap includes both detectors. |

## Closes

Closes #946

## Test plan

- [x] Other: Cloudflare and Spin tests/clippy passed; Playwright confirmed the current publisher `setConfig()` call now sets `gptInitialLoadDisabled`.

## Checklist

- [x] New code has tests


crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -430,16 +436,32 @@ function installInitialLoadDetector(ts: TsjsApi): void {
...
-    const pubads = win.googletag?.pubads?.();
+    const gpt = win.googletag as
+      | (Partial<GoogleTag> & { __tsInitialLoadConfigHooked?: boolean })
+      | undefined;
+    if (!gpt) return;
+
+    if (typeof gpt.setConfig === 'function' && !gpt.__tsInitialLoadConfigHooked) {
+      const originalSetConfig = gpt.setConfig.bind(gpt);
+      gpt.setConfig = function (config: GoogleTagConfig) {
+        if (config?.disableInitialLoad === true) {
```
aram356: 🔧 **wrench** — `setConfig` can only ever *set* the flag, never clear it, which produces a duplicate ad request.

`setConfig` is a settings *merge* API: `{ disableInitialLoad: false }` re-enables initial load and `{ disableInitialLoad: null }` resets it to the default. Matching only `=== true` means that once a publisher disables and later re-enables initial load (a real pattern — disable while waiting on the CMP, re-enable after consent), `ts.gptInitialLoadDisabled` is stuck `true`. `adInit()` then takes the `slotsToRefresh.concat(newSlots)` branch and issues `display()` **and** `refresh()` on a TS-owned slot — exactly the double-request the comment below at line 641 says to avoid.

The asymmetry is new to this PR: the legacy path is safe only because GPT has no `enableInitialLoad()`.

Verified against this branch with a scratch vitest, not theoretical:

```
setConfig({ disableInitialLoad: true });
setConfig({ disableInitialLoad: false });
→ gptInitialLoadDisabled === true
→ adInit(): display('div-atf-sidebar') AND pubads.refresh([slot])
```

**Fix** — key-presence, which matches GPT's merge semantics (an absent key means "unchanged"):

```ts
gpt.setConfig = function (config: GoogleTagConfig) {
  if (config && 'disableInitialLoad' in config) {
    ts.gptInitialLoadDisabled = config.disableInitialLoad === true;
  }
  return originalSetConfig(config);
};
```

The same change is needed in `gpt_bootstrap.js`. Please add the re-enable case as a regression test alongside the new one.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -20,24 +20,40 @@
...
   (window.googletag = window.googletag || { cmd: [] }).cmd.push(function () {
-    var pubads = googletag.pubads && googletag.pubads();
+    var gpt = window.googletag;
+    if (
+      typeof gpt.setConfig === "function" &&
+      !gpt.__tsInitialLoadConfigHooked
+    ) {
+      var originalSetConfig = gpt.setConfig.bind(gpt);
+      gpt.setConfig = function (config) {
+        if (config && config.disableInitialLoad === true) {
```
aram356: 🔧 **wrench** — Same defect as the bundle (see the comment on `gpt/index.ts:447`): `config.disableInitialLoad === true` never clears the flag, so `setConfig({ disableInitialLoad: false })` (or `null`) leaves `ts.gptInitialLoadDisabled` stuck `true` and `adInit()` double-requests TS-owned slots via `display()` + `refresh()`.

**Fix**:

```js
gpt.setConfig = function (config) {
  if (config && "disableInitialLoad" in config) {
    ts.gptInitialLoadDisabled = config.disableInitialLoad === true;
  }
  return originalSetConfig(config);
};
```

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -430,16 +436,32 @@ function installInitialLoadDetector(ts: TsjsApi): void {
...
   cmd.push(() => {
-    const pubads = win.googletag?.pubads?.();
+    const gpt = win.googletag as
+      | (Partial<GoogleTag> & { __tsInitialLoadConfigHooked?: boolean })
+      | undefined;
+    if (!gpt) return;
+
+    if (typeof gpt.setConfig === 'function' && !gpt.__tsInitialLoadConfigHooked) {
+      const originalSetConfig = gpt.setConfig.bind(gpt);
+      gpt.setConfig = function (config: GoogleTagConfig) {
```
aram356: ♻️ **refactor** — The wrapper is signature-locked to today's single-argument API and silently drops anything else GPT may pass. Forwarding all arguments is future-proof at no cost:

```ts
gpt.setConfig = function (...args: [GoogleTagConfig]) {
  const config = args[0];
  // ...
  return originalSetConfig(...args);
};
```

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
```diff
@@ -243,6 +243,62 @@ describe('installTsAdInit', () => {
     expect(mockPubads.refresh).toHaveBeenCalledWith([mockSlot]);
   });
 
+  it('refreshes TS-defined slots when setConfig disables GPT initial load', async () => {
```
aram356: ♻️ **refactor** — Missing negative coverage for the new hook. This test only exercises the positive path; two cheap additions would have caught the `wrench` finding above and would pin down the wrapper's other contracts:

1. `setConfig({ singleRequest: true })` must leave `gptInitialLoadDisabled` unset and must not add TS-owned slots to `refresh()`.
2. Calling `installTsAdInit()` twice must not double-wrap `setConfig` — `__tsInitialLoadConfigHooked` is currently untested, and a double wrap would invoke the original `setConfig` twice.

Also 🏕 — this case is a ~40-line verbatim clone of the preceding legacy test, differing only in the `setConfig` mock. A shared `setupGptMocks()` helper would keep the pair from drifting.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -120,6 +124,7 @@ interface GoogleTag {
   destroySlots(slots?: GoogleTagSlot[]): boolean;
   enableServices(): void;
   display(elementId: string): void;
+  setConfig(config: GoogleTagConfig): void;
```
aram356: ⛏ **nitpick** — Every sibling GPT API the shim guards at runtime is declared optional (`disableInitialLoad?()`, `_loaded_?`), and the detector does check `typeof gpt.setConfig === 'function'`. `setConfig?(config: GoogleTagConfig): void` would be more honest about the surface actually being probed.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -109,6 +109,10 @@ interface GoogleTagPubAdsService {
   disableInitialLoad?(): void;
 }
 
+interface GoogleTagConfig extends Record<string, unknown> {
```
aram356, line 112: ⛏ **nitpick** — Extending `Record<string, unknown>` suppresses excess-property checking entirely, so the type documents `disableInitialLoad` without enforcing anything else about the config. Reasonable as an escape hatch for a partially-modelled third-party API — just noting the trade-off.

crates/trusted-server-core/src/integrations/gpt.rs
```diff
@@ -1261,8 +1262,12 @@ mod tests {
         };
         let combined = integration.head_inserts(&ctx).join("");
         assert!(
-            combined.contains("disableInitialLoad"),
-            "bootstrap should wrap disableInitialLoad() to detect the disabled state"
+            combined.contains("gpt.setConfig"),
```
aram356, line 1265: 📌 **out of scope** — `gpt_bootstrap.js` correctness is still asserted only by substring matching from Rust. This PR duplicates a second piece of non-trivial hooking logic across the bootstrap and the bundle, so the two can now silently diverge — and per the `wrench` finding, they need the same fix applied twice.

Worth a follow-up issue for a jsdom harness that `eval`s `GPT_BOOTSTRAP_JS` and drives it the way `ad_init.test.ts` drives the bundle.

aram356 (changes requested):
## Summary

Extends initial-load detection to `googletag.setConfig({ disableInitialLoad: true })` alongside the legacy `pubads().disableInitialLoad()`, in both the edge bootstrap and the bundle. The approach is right and the idempotency markers are correctly scoped per object, but the new `setConfig` hook is write-only: it can set `gptInitialLoadDisabled` and never clear it, which turns a publisher re-enabling initial load into a duplicate ad request on TS-owned slots.

## Blocking

### 🔧 wrench

- **`setConfig` never clears `gptInitialLoadDisabled` → duplicate ad request**: `setConfig` is a settings *merge* API, so `{ disableInitialLoad: false }` re-enables and `{ disableInitialLoad: null }` resets to default. Matching only `=== true` leaves the flag stuck on, and `adInit()` then does `display()` **and** `refresh()` on a TS-owned slot — the exact double-request the code comment warns against. Verified with a scratch vitest against this branch. Needs the same key-presence fix in both copies (`gpt/index.ts:447`, `gpt_bootstrap.js:37`).

## Non-blocking

### 🤔 thinking

- **Detection remains ordering-dependent**: both hooks install from the `googletag.cmd` queue, so a publisher that configures GPT ahead of the TS injection point still lands in the old blank-slot failure mode. `setConfig` widens coverage but does not remove the race. Longer term it may be more robust for TS to own the fetch for its own slots outright rather than inferring publisher state through wrappers.

### ♻️ refactor

- **Missing negative coverage for the new hook** (`ad_init.test.ts:246`) — no test that a `setConfig` call without `disableInitialLoad` leaves the flag unset, and none that the `__tsInitialLoadConfigHooked` guard prevents double-wrapping.
- **Wrapper drops extra arguments** (`gpt/index.ts:446`) — forward with rest/spread instead of a fixed single parameter.

### 🏕 camp site / 📌 out of scope

- The new test case is a ~40-line verbatim clone of the preceding legacy test; a shared `setupGptMocks()` helper would keep them in sync.
- `gpt_bootstrap.js` still has no behavioral tests — correctness rests on Rust substring assertions while the duplicated hooking logic grows. Follow-up issue suggested (`gpt.rs:1265`).

### ⛏ nitpick

- `setConfig` is declared required on `GoogleTag` while every sibling runtime-guarded API is optional (`gpt/index.ts:127`).
- `GoogleTagConfig extends Record<string, unknown>` suppresses excess-property checking entirely (`gpt/index.ts:112`).

## CI Status

GitHub, at time of review:

- vitest: PASS
- CodeQL (javascript-typescript, actions): PASS
- cargo fmt / clippy / test (fastly, axum, cloudflare, spin, parity): PENDING

Verified locally in a worktree at the PR head:

- `npx vitest run test/integrations/gpt/` — 74 passed
- `cargo test -p trusted-server-core --lib integrations::gpt::tests` — 30 passed
- `tsc --noEmit` — no new errors in the changed files (pre-existing errors only, on untouched lines)

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
prk-Jr: 🔧 **wrench** — Track GPT's effective initial-load configuration, not just attempted `setConfig` calls.

This branch only ever assigns `gptInitialLoadDisabled = true`, and does so before calling the original setter. After:

```js
setConfig({ disableInitialLoad: true });
setConfig({ disableInitialLoad: false });
```

GPT is enabled again while TS still records `true`. `adInit()` then calls both `display()` and `refresh()` for TS-owned slots, producing the duplicate request the code below is designed to avoid. A late or rejected configuration can similarly leave the tracked value different from GPT's effective state.

GPT now exposes the authoritative setting through [`googletag.getConfig("disableInitialLoad")`](https://developers.google.com/publisher-tag/guides/config-migration#initial_load). Prefer reading that value at the point where `slotsNeedingRefresh` is computed, or synchronize from it after the original setter and when the detector installs. Apply the same correction to `gpt_bootstrap.js:37`, and add regression coverage for `true → false` plus configuration already effective before detector installation.

prk-Jr (changes requested):
## Summary

The PR adds detection for the modern `googletag.setConfig({ disableInitialLoad: true })` path, but the new hook tracks attempted setter calls instead of GPT's effective configuration. That state can become stale and cause duplicate requests for TS-owned slots.

## Blocking

### 🔧 wrench

- **Initial-load state can diverge from GPT's effective configuration**: the flag is only set to `true`, never cleared, and it is updated before GPT processes the configuration. Both the bundle and inline bootstrap need to read the authoritative GPT state and cover re-enabling in regression tests.

## CI Status

- All GitHub checks currently report PASS, including formatting, Rust and JavaScript analysis, adapter tests, browser integration tests, and Vitest.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Addressed in 3b02916a. The wrapper now reads GPT’s effective getConfig("disableInitialLoad") value after the original setter, on detector installation, and before deciding refreshes; it retains key-presence fallback when no getter is available. Regression coverage includes true → false and null reset, preventing the duplicate refresh.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
ChristianPavilonis: Addressed in 3b02916a. The edge bootstrap now mirrors the effective-state synchronization, forwards its arguments, and falls back safely when getConfig is unavailable. Added a focused bootstrap behavioral regression as well.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Addressed in 3b02916a. The TypeScript wrapper uses rest/spread and the bootstrap delegates with apply, so all received arguments reach GPT.

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
ChristianPavilonis: Addressed in 3b02916a. Tests now cover unrelated singleRequest configuration without a TS-owned refresh, double install with a single original setter call, false/null re-enable behavior, and configuration effective before installation. The state is false rather than unset when getConfig is available, because that is GPT’s authoritative effective default.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Addressed in 3b02916a: setConfig is now optional in the local GPT surface.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Acknowledged. Retained the Record escape hatch deliberately because this is a partial local model of GPT configuration and callers must pass other supported GPT settings such as singleRequest.

crates/trusted-server-core/src/integrations/gpt.rs
ChristianPavilonis: Addressed in 3b02916a with a focused jsdom behavioral regression that evaluates the edge bootstrap and verifies legacy disableInitialLoad state is preserved when getConfig has no value. The existing Rust emission assertion also now checks for getConfig.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Addressed in 3b02916a. The detector synchronizes from getConfig after setConfig and at installation, and adInit synchronizes again immediately before computing slotsNeedingRefresh. Tests cover true → false/null and state effective before detector installation; the bootstrap follows the same approach.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -413,37 +419,71 @@ function queueWinBillingBeacon(url: string): boolean {
...
     if (typeof service.disableInitialLoad !== 'function' || service.__tsInitialLoadHooked) {
       return;
     }
-    const original = service.disableInitialLoad.bind(service);
+    const originalDisableInitialLoad = service.disableInitialLoad.bind(service);
     service.disableInitialLoad = function () {
-      ts.gptInitialLoadDisabled = true;
-      return original();
+      const result = originalDisableInitialLoad();
+      if (!syncInitialLoadDisabled(gpt, ts)) {
```
prk-Jr, line 487: 🔧 **wrench** — `getConfig()` is now allowed to override the legacy `disableInitialLoad()` signal, but GPT never reports the legacy API through `getConfig()`. A page that mixes both APIs loses the disabled state and TS-owned slots render blank again.

GPT's reference says `getConfig()` "gets a frozen copy of the general configuration options for the page **set by `setConfig`**" — `pubads().disableInitialLoad()` does not populate it. The comment above this function says the same thing ("may not report state set through the legacy API"), yet `syncInitialLoadDisabled()` returning `true` short-circuits the fallback, so any *defined* value from `getConfig()` — including a stale `false` written earlier by `setConfig()` — wins over a legacy call that actually disabled initial load.

Reproduced on this branch with a scratch vitest (`getConfig` backed only by `setConfig`, matching the documented behavior):

```
setConfig({ disableInitialLoad: false });   // publisher re-enables via modern API
pubads().disableInitialLoad();             // ...then disables via legacy API
→ ts.gptInitialLoadDisabled === false      // expected true
→ adInit(): display('div-atf-sidebar'), pubads.refresh NOT called
```

The TS-owned first-impression slot is registered but never requested — the blank-slot regression this detector exists to prevent. `main` is not affected: the legacy wrapper there sets `true` unconditionally.

The `syncInitialLoadDisabled(g, ts)` call added before `slotsNeedingRefresh` widens the window: it re-clobbers legacy state at `adInit()` time even when the legacy call came last.

**Fix** — keep the legacy signal sticky and let `getConfig()` be authoritative only for `setConfig()`-driven state:

```ts
service.disableInitialLoad = function () {
  const result = originalDisableInitialLoad();
  // The legacy API is not reflected by getConfig(), and GPT offers no way to
  // undo it, so record it separately and never let a sync downgrade it.
  ts.gptLegacyInitialLoadDisabled = true;
  ts.gptInitialLoadDisabled = true;
  return result;
};
```

and in `syncInitialLoadDisabled`:

```ts
if (ts.gptLegacyInitialLoadDisabled && value !== true) return false;
```

Please add the mixed-API ordering (`setConfig(false)` → legacy `disableInitialLoad()` → `adInit()`) as regression coverage; none of the new tests exercise both APIs on the same page.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -413,37 +419,71 @@ function queueWinBillingBeacon(url: string): boolean {
...
  * import that does not activate the GPT integration must not touch
  * `window.googletag`. When the GPT shim is active it creates the stub before
  * `installTsAdInit` runs, so the detector is still queued ahead of the
  * publisher's GPT setup.
  */
+function syncInitialLoadDisabled(gpt: Partial<GoogleTag>, ts: TsjsApi): boolean {
+  if (typeof gpt.getConfig !== 'function') return false;
+
+  const config = gpt.getConfig('disableInitialLoad');
+  if (!config || config.disableInitialLoad === undefined) return false;
```
prk-Jr, line 447: 🔧 **wrench** — The unset guard checks only `undefined`, so a `null` from `getConfig()` is treated as an authoritative `false`.

`GoogleTagConfig.disableInitialLoad` is typed `boolean | null` right here in this PR, and `null` is GPT's documented "reset to default" value — the new test even asserts `setConfig({ disableInitialLoad: null })` behavior. But when `getConfig()` returns `{ disableInitialLoad: null }`, this function returns `true` after writing `false`, which suppresses the legacy fallback below.

Reproduced on this branch:

```
getConfig('disableInitialLoad') → { disableInitialLoad: null }
pubads().disableInitialLoad();
→ ts.gptInitialLoadDisabled === false   // expected true
```

**Fix** — treat anything that is not a boolean as unset:

```ts
const value = config?.disableInitialLoad;
if (typeof value !== 'boolean') return false;

ts.gptInitialLoadDisabled = value;
return true;
```

Same change is needed in `gpt_bootstrap.js:33`.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -19,25 +19,61 @@
...
     ) {
       return;
     }
-    var original = pubads.disableInitialLoad.bind(pubads);
+    var originalDisableInitialLoad = pubads.disableInitialLoad.bind(pubads);
     pubads.disableInitialLoad = function () {
-      ts.gptInitialLoadDisabled = true;
-      return original();
+      var result = originalDisableInitialLoad.apply(pubads, arguments);
+      if (!syncInitialLoadDisabled(gpt)) {
```
prk-Jr, line 74: 🔧 **wrench** — Same two defects as the bundle (see the comments on `gpt/index.ts:483` and `:443`): a defined-but-stale `getConfig()` value — or a `null` reset, which line 33 accepts as a real value — suppresses the `ts.gptInitialLoadDisabled = true` fallback, and GPT never reports legacy `disableInitialLoad()` state through `getConfig()`. `syncInitialLoadDisabled(window.googletag)` at line 189 re-applies the clobber at `adInit()` time.

**Fix**:

```js
function syncInitialLoadDisabled(gpt) {
  if (typeof gpt.getConfig !== "function") return false;
  var config = gpt.getConfig("disableInitialLoad");
  var value = config && config.disableInitialLoad;
  if (typeof value !== "boolean") return false;
  // The legacy API is invisible to getConfig() and cannot be undone.
  if (ts.gptLegacyInitialLoadDisabled && value !== true) return false;
  ts.gptInitialLoadDisabled = value;
  return true;
}

pubads.disableInitialLoad = function () {
  var result = originalDisableInitialLoad.apply(pubads, arguments);
  ts.gptLegacyInitialLoadDisabled = true;
  ts.gptInitialLoadDisabled = true;
  return result;
};
```

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
```diff
@@ -243,6 +249,202 @@ describe('installTsAdInit', () => {
     expect(mockPubads.refresh).toHaveBeenCalledWith([mockSlot]);
   });
 
+  it('keeps the legacy disabled state in the edge bootstrap when getConfig is unavailable', async () => {
```
prk-Jr: ♻️ **refactor** — The new jsdom harness only drives the bootstrap's *legacy* path, so the bootstrap `setConfig` wrapper — the feature this PR adds — still has no behavioral coverage, only the Rust substring assertions. Now that the harness exists, the `setConfig({ disableInitialLoad: true })` → `adInit()` case and the `true → false` re-enable case are a few lines each, and they would pin down the bootstrap/bundle pair that the earlier 📌 finding flagged as prone to divergence.

⛏ Also, the title says "when getConfig is unavailable" but the fixture *does* provide `getConfig`; it just returns `undefined`. "when getConfig does not report legacy state" matches what is asserted.

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
```diff
@@ -243,6 +249,202 @@ describe('installTsAdInit', () => {
...
+          formats: [[300, 250]],
+          targeting: {},
+        },
+      ],
+      bids: {},
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } as any;
+
+    const bootstrap = await readFile(
+      path.resolve(process.cwd(), '../../trusted-server-core/src/integrations/gpt_bootstrap.js'),
```
prk-Jr: ⛏ **nitpick** — Resolving the bootstrap relative to `process.cwd()` ties the test to being launched from `crates/trusted-server-js/lib`. `new URL(...)` against the module location is invocation-independent:

```ts
const bootstrapPath = new URL(
  '../../../../trusted-server-core/src/integrations/gpt_bootstrap.js',
  import.meta.url
);
const bootstrap = await readFile(bootstrapPath, 'utf8');
```

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -120,6 +124,8 @@ interface GoogleTag {
   destroySlots(slots?: GoogleTagSlot[]): boolean;
   enableServices(): void;
   display(elementId: string): void;
+  setConfig?(config: GoogleTagConfig): void;
+  getConfig?(key: 'disableInitialLoad'): GoogleTagConfig | undefined;
```
prk-Jr: ⛏ **nitpick** — GPT's actual signature is variadic (`getConfig(...keys)`) and returns a frozen `PageSettingsConfig` whose values may be `null`. `getConfig?(...keys: string[]): GoogleTagConfig | undefined` models the surface more honestly, and — given the `null` finding above — makes the missing `null` case visible at the type level.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -619,12 +659,13 @@ export function installTsAdInit(): void {
...
-      // display() above — but when the publisher called
-      // pubads().disableInitialLoad(), display() only registers the slot and the
-      // ad request must come from refresh(). Without this, a TS-owned
+      // display() above — but when the publisher disabled initial load through
+      // setConfig() or the legacy pubads() method, display() only registers the
+      // slot and the ad request must come from refresh(). Without this, a TS-owned
       // first-impression slot renders blank on initial-load-disabled pages. Only
       // add them in that case; otherwise display() + refresh() would
       // double-request the impression.
+      syncInitialLoadDisabled(g, ts);
```
prk-Jr, line 672: 👍 **praise** — Re-reading the effective state immediately before computing `slotsNeedingRefresh` is the right place for it: it closes the gap where configuration lands after the detector installs but before `adInit()` runs. Sharing `__tsInitialLoadConfigHooked` with the edge bootstrap so exactly one wrapper is installed across the two scripts, and forwarding all arguments through both wrappers, are equally solid.

prk-Jr (changes requested):
## Summary

The move from "record what was attempted" to "read GPT's effective state" is the right direction, and the idempotency guards, argument forwarding, and jsdom bootstrap harness all land well. One blocking defect remains: `getConfig()` is now treated as authoritative for the *legacy* `pubads().disableInitialLoad()` path too, and per GPT's own documentation it never reflects that API. When a page mixes both APIs, the legacy detection this PR is meant to preserve is silently overwritten with `false` and TS-owned slots go back to rendering blank — the exact failure #946 fixes.

## Blocking

### 🔧 wrench

- **`getConfig()` overrides legacy `disableInitialLoad()` state**: `syncInitialLoadDisabled()` wins over the legacy fallback, but GPT's `getConfig()` only reports `setConfig()` state, so a mixed-API page loses the disabled flag (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts:483`, `crates/trusted-server-core/src/integrations/gpt_bootstrap.js:73`).
- **`null` is treated as an authoritative `false`**: the unset guard checks only `undefined`, while `GoogleTagConfig.disableInitialLoad` is typed `boolean | null` and `null` is GPT's documented "reset to default" (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts:443`, `crates/trusted-server-core/src/integrations/gpt_bootstrap.js:33`).

Both were reproduced against this branch with a scratch vitest run; details and suggested fixes are inline.

## Non-blocking

### ♻️ refactor

- **Bootstrap coverage stops at the legacy path**: the new jsdom harness never drives the bootstrap's `setConfig` wrapper, which is the feature this PR adds (`crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts:252`).

### ⛏ nitpick

- **Test resolves the bootstrap through `process.cwd()`**: breaks if vitest is invoked from the repo root (`crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts:291`).
- **`getConfig` type is narrower than the real API**: GPT's signature is variadic and values may be `null` (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts:128`).

### 👍 praise

- **Effective-state re-sync immediately before `slotsNeedingRefresh`**, plus `__tsInitialLoadConfigHooked` sharing between bootstrap and bundle so the wrapper is installed exactly once, and full argument forwarding on both wrappers (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts:668`).

## CI Status

- fmt: PASS
- clippy / CodeQL analyze: PASS
- rust tests (fastly, cloudflare, CLI, cross-adapter parity): PASS
- js tests (vitest) + TS/docs format: PASS
- pending at review time: axum native, spin native + wasm32-wasip1, integration artifacts

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -413,37 +423,71 @@ function queueWinBillingBeacon(url: string): boolean {
...
-    const pubads = win.googletag?.pubads?.();
+    const gpt = win.googletag as
+      | (Partial<GoogleTag> & { __tsInitialLoadConfigHooked?: boolean })
+      | undefined;
+    if (!gpt) return;
+
+    syncInitialLoadDisabled(gpt, ts);
+
+    if (typeof gpt.setConfig === 'function' && !gpt.__tsInitialLoadConfigHooked) {
+      const originalSetConfig = gpt.setConfig.bind(gpt);
```
aram356, line 466: ♻️ **refactor** — This fallback branch (`!syncInitialLoadDisabled(gpt, ts) && config && 'disableInitialLoad' in config`) is never exercised by the tests. Every new setConfig test injects a `getConfig` mock, so `syncInitialLoadDisabled` always returns `true` and this line is skipped.

The branch exists specifically for older GPT that exposes `setConfig` but not `getConfig`. I confirmed it works with a scratch probe (`setConfig({disableInitialLoad:true})` → `true`, then `{disableInitialLoad:false}` → `false`) with `getConfig` absent — but nothing guards it against regression. Suggest a test where `googletag` has `setConfig` and no `getConfig`.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -19,25 +19,62 @@
...
+    var gpt = window.googletag;
+    syncInitialLoadDisabled(gpt);
+    if (
+      typeof gpt.setConfig === "function" &&
+      !gpt.__tsInitialLoadConfigHooked
+    ) {
+      var originalSetConfig = gpt.setConfig.bind(gpt);
+      gpt.setConfig = function (config) {
+        var result = originalSetConfig.apply(gpt, arguments);
+        if (
```
aram356, line 51: ♻️ **refactor** — Same untested fallback as in the bundle (`gpt/index.ts:466`): with a `getConfig` mock always present in the setConfig tests, the `!syncInitialLoadDisabled(gpt) && ... 'disableInitialLoad' in config` path here is never hit. Worth an eval-harness case with `getConfig` absent, mirroring the existing `keeps the legacy disabled state ... when getConfig is unavailable` bootstrap test.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -619,12 +663,13 @@ export function installTsAdInit(): void {
       // Slots needing an explicit ad request via refresh(). Reused
       // publisher-owned slots always need one to pick up the just-applied
       // server-side targeting. TS-defined slots are normally fetched by the
-      // display() above — but when the publisher called
-      // pubads().disableInitialLoad(), display() only registers the slot and the
-      // ad request must come from refresh(). Without this, a TS-owned
+      // display() above — but when the publisher disabled initial load through
+      // setConfig() or the legacy pubads() method, display() only registers the
+      // slot and the ad request must come from refresh(). Without this, a TS-owned
```
aram356, line 668: 🤔 **thinking** — Re-reading `getConfig` here at decision time is a nice robustness gain: it picks up setConfig state regardless of whether the publisher configured GPT before or after the detector installed. The only remaining blind spot is a publisher calling the *legacy* `pubads().disableInitialLoad()` before the detector wraps it, on a GPT build where `getConfig` doesn't reflect legacy state — narrow enough that it's probably not worth more code. Just flagging the residual edge.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -120,6 +128,8 @@ interface GoogleTag {
   destroySlots(slots?: GoogleTagSlot[]): boolean;
```
aram356, line 128: ⛏ **nitpick** — `getConfig?(key: 'disableInitialLoad')` restricts the key to that single literal; real GPT accepts `string | string[]`. Fine for a minimal internal stub — just noting the divergence in case another key is ever read.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -19,25 +19,62 @@
...
-  // queue so it runs before the publisher's own disableInitialLoad() call.
+  // Track whether the publisher disabled GPT initial load. Read the effective
+  // googletag.getConfig() value when available, and wrap googletag.setConfig()
+  // and the legacy pubads().disableInitialLoad() method so changes are
+  // synchronized immediately and still detected when getConfig() is
+  // unavailable. With initial load disabled, display() only registers a slot
+  // and the ad request must come from a later refresh(); adInit() reads this to
+  // refresh its own freshly defined
+  // slots so they are not left blank. Pushed onto the command queue so it runs
+  // before the publisher's own GPT configuration.
```
aram356, line 30: ⛏ **nitpick** — `syncInitialLoadDisabled` is duplicated near-identically between this bootstrap and the bundle's `gpt/index.ts`. Inherent to the two-implementation design, and the new bootstrap eval-test helps, but the two copies can still drift — keep them in lockstep on any future change.

aram356 (approved):
## Summary

Re-review after the update responding to the earlier `REQUEST_CHANGES`. The prior blocking issue — the `setConfig` hook could only ever set `gptInitialLoadDisabled`, never clear it, producing a duplicate ad request on re-enable — is resolved: the code now reads GPT's authoritative `getConfig('disableInitialLoad')` at detection *and* at `adInit()` decision time, and falls back to key-presence (`'disableInitialLoad' in config`) with `=== true`. I traced the true → false → null sequence and confirmed the flag now clears and `adInit()` no longer double-requests. Verified against Google's GPT reference that `getConfig('disableInitialLoad')` returns a frozen object `{ disableInitialLoad: ... }` and that `disableInitialLoad` is a real page-level `setConfig`/`getConfig` key, so the read is sound.

Approving. The remaining notes are non-blocking.

## Non-blocking

### ♻️ refactor / test-coverage

- **The `getConfig`-absent fallback is untested** (`gpt/index.ts:466`, `gpt_bootstrap.js:51`) — every new setConfig test injects a `getConfig` mock, so `syncInitialLoadDisabled` always returns `true` and the fallback branch never runs. That branch exists precisely for older GPT that has `setConfig` but not `getConfig`; a regression test with `getConfig` absent would lock it in. Verified working via a scratch probe.

### 🤔 thinking

- **Residual ordering dependency** (`gpt/index.ts:668`) — the authoritative `getConfig` re-read at `adInit()` now catches setConfig state regardless of call order (a real gain). The only remaining blind spot is a publisher using the *legacy* `pubads().disableInitialLoad()` before the detector wraps it, on a GPT build where `getConfig` does not reflect legacy state. Narrow; noting for completeness.

### ⛏ nitpick

- `getConfig?(key: 'disableInitialLoad')` types the key as only that literal; real GPT accepts `string | string[]` (`gpt/index.ts:128`).
- `syncInitialLoadDisabled` is duplicated near-identically between the bootstrap (JS) and the bundle (TS) — inherent to the two-implementation design and now partly guarded by the new bootstrap eval-test, but the divergence risk remains (`gpt_bootstrap.js:30`).

## CI Status

All checks green on GitHub: cargo fmt, clippy (all adapters), cargo test (fastly / axum / cloudflare / spin / parity), vitest, browser + Fastly EC integration, CodeQL, format-typescript, format-docs.

Verified locally at the PR head: `integrations::gpt::tests` 30 passed; GPT vitest 90 passed; `tsc --noEmit` clean on the changed files.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Addressed in 370403f7 with mixed-API ordering regressions, but I deliberately did not make the legacy signal permanently sticky.

I verified the current GPT runtime (`m202607230101`) in headless Chrome. `pubads().disableInitialLoad()` changes both `getConfig("disableInitialLoad").disableInitialLoad` and `isInitialLoadDisabled()` to `true`; a subsequent `setConfig({ disableInitialLoad: false })` changes both back to `false`, and `null` resets both to `false`. This also aligns with Google's migration guide, which lists `getConfig("disableInitialLoad")` as the replacement for the legacy getter. A sticky TS flag would therefore create the duplicate `display()` + `refresh()` request when GPT has actually been re-enabled.

The new regression models GPT's shared effective state and covers both directions: `setConfig(false)` → legacy disable → `adInit()` refreshes, while legacy disable → `setConfig(false/null)` → `adInit()` does not refresh the TS-owned slot.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Addressed in 370403f7 by separating setter input from getter output and adding reset coverage. `setConfig` accepts `boolean | null`, but the documented `getConfig` return property is boolean and the current GPT runtime normalizes a `null` reset to `false` rather than returning `null`. The regression now models that effective output and verifies that `false` and `null` both prevent the duplicate TS-owned refresh.

The scratch fixture's `null` getter result and setConfig-only state did not match the current GPT runtime; retaining that model would require the sticky behavior that breaks a real legacy-disable → modern-re-enable sequence.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
ChristianPavilonis: Addressed in 370403f7 without adding the proposed sticky flag. The bootstrap now has behavioral coverage using GPT's shared effective state: modern disable refreshes the TS-owned slot, while `false` and `null` re-enable/reset it without a duplicate refresh. The current GPT runtime also reports a legacy disable through `getConfig` and permits a later modern false/null reset, so permanently preserving legacy state would diverge from GPT and double-request.

The bootstrap and bundle comments now describe `getConfig` as the effective source, with wrappers retained as the fallback when the getter is unavailable.

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
ChristianPavilonis: Addressed in 370403f7. The jsdom harness now executes the bootstrap's `setConfig` wrapper and covers `true` → refresh, `true → false` → no duplicate refresh, and `true → null` → no duplicate refresh. The legacy test was renamed to say that the fixture's getter does not report the state, which matches what it actually asserts.

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
ChristianPavilonis: Addressed in 370403f7. Bootstrap loading is now anchored to `import.meta.url` rather than a fixed relationship to `process.cwd()`. The helper also normalizes Vitest's `/@fs/` and root-relative HTTP module URLs before passing the path to `readFile`. I verified the focused suite both from `crates/trusted-server-js/lib` and from the repository root; both runs pass 42/42.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Addressed in 370403f7 using the current documented GPT surface. The official signature takes one `keys: string | string[]` argument (rather than variadic arguments), so the local type now models that form. Setter input and effective getter output are separate types: the setter accepts `boolean | null`, while the getter exposes the normalized boolean effective value observed in GPT (`null` resets to `false`).

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Thanks. The additional regressions in 370403f7 now lock this down for both the bundle and edge bootstrap, including modern re-enable/reset behavior and mixed modern/legacy API ordering.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Addressed in 0f173128. The bundle test now installs the detector with `setConfig` present and `getConfig` absent, then verifies the direct-value fallback transitions `true → false`. It subsequently enables the getter and continues through the authoritative effective-state and mixed-API cases, so both branches are covered in one fixture.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
ChristianPavilonis: Addressed in 0f173128. The bootstrap eval harness now begins with `setConfig` available and `getConfig` absent, verifies fallback tracking for `true → false`, then adds the getter and exercises the effective-state `true`, `false`, and `null` paths.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Acknowledged. The current GPT runtime reports legacy state through `getConfig`, so the blind spot is limited to a hypothetical older/non-conforming runtime where the legacy call also occurred before either wrapper was installed. The detector is queued ahead of publisher GPT setup in the supported injection order, and I agree extra state machinery is not justified for that residual case.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Addressed in 370403f7. The local signature now accepts the documented `string | string[]` keys argument, and setter input is separated from the getter's effective output type.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
ChristianPavilonis: Acknowledged. The two copies remain intentionally in lockstep, and the expanded eval harness now exercises the bootstrap's getter-present and getter-absent setConfig paths alongside the equivalent bundle coverage. That should make future drift visible in Vitest rather than only through Rust substring assertions.

prk-Jr (approved):

PR #951, opened 2026-07-22
Handle oversized HEAD response metadata

### LINKED ISSUE #950: Oversized S3 assets fail Image Optimizer HEAD preflight
A signed S3 HEAD preflight for an Image Optimizer asset fails when the object Content-Length exceeds the adapter’s 10 MiB buffered-response limit.

HEAD responses contain no body; their Content-Length describes the corresponding GET representation. The Fastly adapter currently treats that metadata as a body it will buffer and rejects the preflight before the final streamed GET can run.

The adapter should preserve HEAD response headers while skipping body materialization and body-size enforcement. Buffered non-HEAD responses must retain the existing 10 MiB safety limit.

ChristianPavilonis:
## Summary

- Allow HEAD responses to preserve oversized `Content-Length` metadata without triggering the buffered-body safety limit.
- Keep the existing 10 MiB limit intact for buffered responses that can allocate origin body bytes in WASM memory.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-adapter-fastly/src/platform.rs` | Track whether the request expects a response body, return an empty body for HEAD responses, and add oversized HEAD/GET regression coverage. |

## Closes

Closes #950

## Test plan

- [x] Other: `cargo test-fastly`; `cargo clippy-fastly`

## Checklist

- [x] New code has tests

### DIFF
--- a/crates/trusted-server-adapter-fastly/src/platform.rs
+++ b/crates/trusted-server-adapter-fastly/src/platform.rs
@@ -360,11 +360,14 @@ fn fastly_response_to_platform(
     mut resp: fastly::Response,
     backend_name: impl Into<String>,
     stream_response: bool,
+    response_body_expected: bool,
 ) -> Result<PlatformResponse, Report<PlatformError>> {
     // Pre-flight: reject oversized responses before copying bytes into WASM heap.
     // Content-Length is advisory but covers most origin responses; chunked
     // responses without it fall through to the post-materialization check below.
-    if !stream_response
+    // HEAD responses report the corresponding GET size but contain no body.
+    if response_body_expected
+        && !stream_response
         && let Some(claimed_len) = resp
             .get_header("content-length")
             .and_then(|v| v.to_str().ok())
@@ -382,7 +385,9 @@ fn fastly_response_to_platform(
     for (name, value) in resp.get_headers() {
         builder = builder.header(name.as_str(), value.as_bytes());
     }
-    let body = if stream_response {
+    let body = if !response_body_expected {
+        edgezero_core::body::Body::empty()
+    } else if stream_response {
         fastly_body_to_edge_stream(resp.take_body())
     } else {
         let body_bytes = resp.take_body_bytes();
@@ -431,6 +436,7 @@ impl PlatformHttpClient for FastlyPlatformHttpClient {
         let backend_name = request.backend_name.clone();
         let image_optimizer = request.image_optimizer;
         let stream_response = request.stream_response;
+        let response_body_expected = request.request.method() != edgezero_core::http::Method::HEAD;
         let mut fastly_req = edge_request_to_fastly(request.request)?;
         if let Some(options) = image_optimizer {
             apply_fastly_image_optimizer(&mut fastly_req, options)?;
@@ -438,7 +444,12 @@ impl PlatformHttpClient for FastlyPlatformHttpClient {
         let fastly_resp = fastly_req
             .send(&backend_name)
             .change_context(PlatformError::HttpClient)?;
-        fastly_response_to_platform(fastly_resp, backend_name, stream_response)
+        fastly_response_to_platform(
+            fastly_resp,
+            backend_name,
+            stream_response,
+            response_body_expected,
+        )
     }
 
     async fn send_async(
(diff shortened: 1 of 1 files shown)


crates/trusted-server-adapter-fastly/src/platform.rs
```diff
@@ -501,7 +512,7 @@ impl PlatformHttpClient for FastlyPlatformHttpClient {
                         .attach("select: response has no backend name; correlation impossible"));
                 };
                 (
-                    fastly_response_to_platform(fastly_resp, backend_name, false),
+                    fastly_response_to_platform(fastly_resp, backend_name, false, true),
```
aram356: 🤔 **thinking** — This `select()` path hardcodes `response_body_expected = true`, so a HEAD request sent via `send_async` + `select` (or the trait's `wait()` helper) bypasses the fix: an oversized `Content-Length` would still be falsely rejected. No current caller sends HEAD on this path (auction fan-out, `pull_sync`, `aps`, `adserver_mock`, `prebid` are all GET/POST), so this is latent rather than live — but the asymmetry is invisible at the call site.

**Suggestion**: carry the flag on `PlatformPendingRequest` the same way `backend_name` is already carried for correlation:

```rust
let response_body_expected = request.request.method() != edgezero_core::http::Method::HEAD;
// ...
Ok(PlatformPendingRequest::new(pending)
    .with_backend_name(backend_name)
    .with_response_body_expected(response_body_expected))
```

or at minimum leave a comment here stating the HEAD-not-supported assumption.

crates/trusted-server-adapter-fastly/src/platform.rs
```diff
@@ -360,11 +360,14 @@ fn fastly_response_to_platform(
...
     backend_name: impl Into<String>,
     stream_response: bool,
+    response_body_expected: bool,
 ) -> Result<PlatformResponse, Report<PlatformError>> {
     // Pre-flight: reject oversized responses before copying bytes into WASM heap.
     // Content-Length is advisory but covers most origin responses; chunked
     // responses without it fall through to the post-materialization check below.
-    if !stream_response
+    // HEAD responses report the corresponding GET size but contain no body.
+    if response_body_expected
```
aram356, line 482: 🌱 **seedling** — Gating on the request method alone misses the other bodiless responses: RFC 9110 also defines 204, 304, and 1xx as having no content, and a 304 may legitimately carry the representation's `Content-Length`. A buffered conditional GET for a >10 MiB asset answering 304 would still be falsely rejected here. Not reachable via the S3 preflight (a 304 short-circuits to the streamed GET), so future-focused: consider gating on response semantics (`method == HEAD || status ∈ {204, 304, 1xx}`) — core already has `response_carries_body` in `publisher.rs` with exactly this shape.

📌 **out of scope** — The Cloudflare adapter has the identical gap: `crates/trusted-server-adapter-cloudflare/src/platform.rs` (~line 335) mirrors this same two-stage cap with no HEAD exemption (its comment even says "Mirrors the Fastly adapter's two-stage cap"). Lower urgency since the Image Optimizer preflight is Fastly-gated today, but any HEAD proxied through Workers to an oversized origin object fails the same way — worth a follow-up issue to mirror this fix there.

crates/trusted-server-adapter-fastly/src/platform.rs
```diff
@@ -736,6 +747,55 @@ mod tests {
...
+                .and_then(|value| value.to_str().ok()),
+            Some("10485761"),
+            "should preserve the origin Content-Length"
+        );
+        assert!(
+            platform_response
+                .response
+                .into_body()
+                .into_bytes()
+                .unwrap_or_default()
```
aram356: ⛏ **nitpick** — `into_bytes()` returns `Option<Bytes>` (`None` for stream bodies), so `unwrap_or_default()` maps a streaming body to empty bytes: if the HEAD branch ever regressed into the streaming path, this "empty body" assertion would still pass vacuously. Prefer:

```rust
.into_bytes()
.expect("should return a buffered body")
.is_empty()
```

Relatedly, the hardcoded `Some("10485761")` here and `"exceeds 10485760-byte"` in the rejection test could derive from `MAX_PLATFORM_RESPONSE_BODY_BYTES` so the tests survive a limit change.

aram356 (approved):

PR #957, opened 2026-07-23
Support per-section GAM ad unit paths via gam_unit_path templates

### LINKED ISSUE #954: creative_opportunities: express per-section gam_unit_path without one rule per (slot × section)
## Summary

`creative_opportunities.slot.gam_unit_path` is a static string, so a publisher whose ad unit varies by site section cannot express that in one rule. The only way to model it is **one slot rule per (slot × section)**, which multiplies out fast: a site with 3 slots and 10 sections needs **30 near-identical rules** to express a single pattern.

## Why it happens

`resolved_gam_unit_path` is literal-or-default — there is no substitution step:

```rust
pub fn resolved_gam_unit_path(&self, gam_network_id: &str) -> String {
    self.gam_unit_path
        .clone()
        .unwrap_or_else(|| format!("/{}/{}", gam_network_id, self.id))
}
```

`crates/trusted-server-core/src/creative_opportunities.rs`

So `gam_unit_path` is either a fixed path or `/<network_id>/<slot_id>`. Neither can vary by request path, even though `page_patterns` already matches on exactly that.

## Impact

Each duplicated rule repeats `id`, `div_id`, `formats`, and the `[creative_opportunities.slot.providers.prebid]` sub-table. Only `gam_unit_path` and `page_patterns` differ. That means:

- **Config bloat** — 30 rules where 3 would do; ~270 lines to express one mapping.
- **Silent drift** — adding a format or changing `div_id` requires editing every section variant. Miss one and that section quietly diverges.
- **Rot on growth** — every new site section requires N new rules (one per slot), and forgetting them means those pages fall through to whatever rule matches, bidding on the wrong inventory.

The failure mode is not loud: a rule with the wrong `gam_unit_path` still returns bids, just against inventory that does not correspond to the page. That is a discrepancy/reporting problem that is easy to ship and hard to notice.

## Proposed direction

Some way to express "unit varies by matched section" in a single rule. Options, roughly in order of blast radius:

1. **Placeholder substitution** in `gam_unit_path`, resolved at request time from the matched path — e.g. `gam_unit_path = "/12345678/publisher/{section}"`.
2. **Explicit mapping table** on the slot — pattern → unit — keeping resolution data-driven rather than string-templated:
   ```toml
   [[creative_opportunities.slot]]
   id = "ad-header"
   [creative_opportunities.slot.unit_by_pattern]
   "/" = "/12345678/publisher/homepage"
   "/news/*" = "/12345678/publisher/news"
   ```
3. **Derivation rule** with an explicit default (e.g. first path segment, with a configurable value for `/`).

## Design constraint worth calling out

Whatever the mechanism, the *derivation policy* is publisher-specific. "First path segment, with `/` mapping to `homepage`" is one site's URL convention, not a universal one — it should not be hardcoded in core. Option 2 keeps the policy in config; option 1 needs a well-defined, documented placeholder set plus a decision on what `{section}` means for `/`, multi-segment paths, and unmatched routes.

## Implementation note

There is exactly one runtime call site:

`crates/trusted-server-core/src/publisher.rs` — `build_slot_json(slot, co_config)`

It does **not** currently have the request path in scope. It is shared by `build_ad_slots_script` (initial render) and `handle_page_bids` (SPA navigation), both of which do know the path, so threading it through is mechanical. `CreativeOpportunitySlot::validate` also calls `resolved_gam_unit_path` without a path and would need a path-independent validation story.

## Acceptance criteria

- [ ] A publisher with N slots across M sections can express per-section ad units without N×M rules
- [ ] Resolution is covered by tests, including `/`, multi-segment paths, and paths matching no rule
- [ ] Existing static `gam_unit_path` configs keep working unchanged
- [ ] `validate()` still catches an empty/unresolvable unit path
- [ ] Documented in the configuration guide, including what happens on an unmatched route

prk-Jr:
## Summary

- `creative_opportunities.slot.gam_unit_path` is now a template with `{network_id}`, `{section}`, and `{slot_id}` placeholders, so a publisher whose GAM ad unit varies by site section expresses that in **one** slot rule instead of one rule per (slot × section).
- `{section}` is derived from the request path (first segment, sanitized), with a required `section_root` for `/` — the URL→section convention lives in config, not core.
- Server-only change: `gam_unit_path` reaches `googletag.defineSlot` client-side but is not in the OpenRTB request, so the wire shape is unchanged and there is no JS change.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/creative_opportunities.rs` | Add `UnitTemplatePart`, `parse_unit_template`, `sanitize_section`, `derive_section`; slot `compiled_unit` cache + `compile_unit_template`/`render_gam_unit_path`/`template_uses_section`; config `section_root` + `compile_unit_templates`; `validate_runtime` now enforces `section_root` when `{section}` is used and drops the render-time emptiness check |
| `crates/trusted-server-core/src/publisher.rs` | `build_slot_json`/`build_ad_slots_script` take `request_path`; render the template on both the initial-render and `handle_page_bids` (SPA) paths |
| `crates/trusted-server-core/src/settings.rs` | `prepare_runtime` parses templates via `compile_unit_templates` before validation, surfacing parse errors as `Configuration` errors |
| `docs/guide/configuration.md` | New "Creative Opportunities Configuration" section documenting templating, `{section}` derivation, the no-decode rule, and unmatched-route behavior |
| `trusted-server.example.toml` | Document `section_root` and a templated-slot example |
| `docs/superpowers/specs/`, `docs/superpowers/plans/` | Design spec and implementation plan |

## Closes

Closes #954

## Test plan

- [x] `cargo test-fastly && cargo test-axum` (also `test-cloudflare`, `test-spin` — all pass)
- [x] `cargo clippy-fastly` (remaining clippy targets left to CI)
- [ ] JS tests / JS format — no JS change in this PR; left to CI
- [ ] `clippy-axum` / `clippy-cloudflare*` / `clippy-spin*` — left to CI

New tests: template parsing (unknown placeholder, unmatched/nested brace, empty), section derivation (`/`, single/multi-segment, unsafe/undecoded segment `new%20s` → `new_20s`), `render_gam_unit_path` (template, default, verbatim), startup validation (missing/invalid `section_root`, parse error), and initial-render/SPA equivalence for the same path.

## Hardening note

Template parsing and `section_root` validation are config-derived and run at startup in `Settings::prepare_runtime`. Invalid enabled config fails startup via `Report<TrustedServerError::Configuration>` — no `panic!`, `unwrap()`, or `expect()` on the config path. `compile_unit_templates` rejects malformed templates (unknown placeholder, unmatched/nested brace, empty); `validate_runtime` rejects a `{section}` template with a missing or non-`[A-Za-z0-9_-]` `section_root`. Regression coverage: `compile_unit_templates_surfaces_parse_error`, `validate_runtime_requires_section_root_when_template_uses_section`, `validate_runtime_rejects_invalid_section_root`, and `settings_rejects_creative_opportunity_slot_with_empty_gam_unit_path`. Existing static `gam_unit_path` configs are unaffected (verbatim when no placeholders; `/{network_id}/{slot_id}` when absent).

## Checklist

- [x] New code has tests


crates/trusted-server-core/src/creative_opportunities.rs
```diff
@@ -14,6 +14,105 @@ use crate::auction::types::{AdFormat, AdSlot, MediaType};
...
+
+/// Derives the `{section}` value from a request path.
+///
+/// Uses the first non-empty path segment, sanitized to `[A-Za-z0-9_-]`. Falls
+/// back to `section_root` when the path has no segment (`/`, repeated slashes).
+///
+/// The path is used **raw** (not percent-decoded) so this stays consistent with
+/// how [`page_patterns`](CreativeOpportunitySlot::page_patterns) glob-match the
+/// same path — e.g. `/new%20s` yields `new_20s`, never the decoded `new_s`.
+pub(crate) fn derive_section(path: &str, section_root: &str) -> String {
```
ChristianPavilonis: 🔧 **The section derivation policy remains hardcoded despite the issue requiring it to be publisher-configurable**

`derive_section` always selects the first non-empty path segment; only the root fallback is configurable. This contradicts issue #954's explicit constraint that the URL-to-section convention is publisher-specific, as well as the PR's claim that the convention lives in config. A publisher using `/en/news/article` will silently render the `en` GAM unit rather than `news`, potentially serving and reporting against the wrong inventory.

**Suggested fix:** Make segment selection configurable (for example, `section_segment = 0`) or provide a pattern-to-section mapping. At minimum, resolve the requirement conflict explicitly before merging and remove claims that the complete convention is config-driven.

crates/trusted-server-core/src/creative_opportunities.rs
```diff
@@ -40,6 +139,14 @@ pub struct CreativeOpportunitiesConfig {
     /// Price granularity for header-bidding price bucketing. Defaults to `Dense`.
     #[serde(default)]
     pub price_granularity: PriceGranularity,
+    /// Value substituted for `{section}` when the request path has no first
+    /// segment (e.g. `/`).
+    ///
+    /// Required when any slot's [`gam_unit_path`](CreativeOpportunitySlot::gam_unit_path)
+    /// template contains `{section}`. No default — a home-section name is
+    /// publisher-specific, so the URL→section convention stays in config, not core.
+    #[serde(default)]
```
ChristianPavilonis: 🔧 **New config blobs can make rollback to the previous server version fail at startup**

`section_root: Option<String>` uses `#[serde(default)]` without `skip_serializing_if`. Every config push containing `[creative_opportunities]` therefore serializes `"section_root": null` even when the feature is unused. Previous binaries deserialize this struct with `deny_unknown_fields` and do not know this key, so a normal binary rollback can fail to read the persisted app-config blob. The active `section_root` in `trusted-server.example.toml` creates the same incompatibility for newly initialized configs.

**Suggested fix:** Add `skip_serializing_if = "Option::is_none"`, leave `section_root` commented out in the default example unless the templated slot is enabled, and add an old-schema compatibility test asserting that an unused key is not serialized.

crates/trusted-server-core/src/creative_opportunities.rs
```diff
@@ -254,15 +410,50 @@ impl CreativeOpportunitySlot {
...
+        };
+        Ok(())
+    }
+
+    /// Renders the resolved GAM unit path for a given network id and section.
+    ///
+    /// Substitutes `{network_id}`, `{section}`, and `{slot_id}` in the parsed
+    /// template. Falls back to `/<network_id>/<id>` when the slot has no template.
+    #[must_use]
+    pub(crate) fn render_gam_unit_path(&self, gam_network_id: &str, section: &str) -> String {
```
ChristianPavilonis: 🔧 **The cache conflates 'no template' with 'not compiled,' silently routing slots to the default GAM path**

`compiled_unit: None` represents both an absent `gam_unit_path` and an explicit path that was deserialized or constructed without calling `compile_unit_templates`. This branch interprets both as absent and returns `/<network_id>/<slot_id>`. `validate_runtime` similarly misses `{section}` requirements when the cache is uninitialized. Public `AuctionDispatch` accepts arbitrary deserialized slots, and `matches_path` explicitly supports legacy/unprepared slots, so those callers can silently bid against the wrong inventory. The updated fixture at `publisher.rs:4291` already constructs this state but does not assert the emitted path.

**Suggested fix:** Represent `Uncompiled`, `Absent`, and `Compiled` distinctly, or parse/fall back from raw `gam_unit_path` when the cache is absent. Validation should inspect the raw template. Add direct-deserialization and static-path regression tests.

crates/trusted-server-core/src/creative_opportunities.rs
```diff
@@ -170,13 +324,15 @@ impl CreativeOpportunitySlot {
...
 
-        if self
-            .resolved_gam_unit_path(gam_network_id)
-            .trim()
-            .is_empty()
+        // A present-but-blank `gam_unit_path` renders to an empty/whitespace
+        // unit path. An empty string also fails template parsing at startup;
+        // this keeps the slot-level check self-contained (tests call
+        // `validate_runtime` without compiling templates first).
+        if let Some(raw) = &self.gam_unit_path
```
ChristianPavilonis, line 492: 🔧 **A successful startup can still produce an empty rendered GAM unit path**

This validates only whether the raw template text is blank. With `gam_network_id = ""` and `gam_unit_path = "{network_id}"`, parsing and startup validation succeed, but rendering produces an empty string. GPT then receives an invalid `defineSlot` path. This violates the acceptance criterion that startup catch empty or unresolvable paths and the implementation's 'non-empty by construction' assumption.

**Suggested fix:** Validate substitutions used by each template—particularly a referenced network ID—and reject a rendered path that is empty or whitespace. Add a startup regression test for this configuration.

crates/trusted-server-core/src/creative_opportunities.rs
```diff
@@ -115,7 +269,7 @@ impl CreativeOpportunitySlot {
     ///
     /// Returns an error string when required slot fields are empty, invalid,
     /// or semantically unusable at runtime.
-    pub fn validate_runtime(&self, gam_network_id: &str) -> Result<(), String> {
+    pub fn validate_runtime(&self) -> Result<(), String> {
```
ChristianPavilonis, line 433: 🔧 **The PR removes or changes existing public Rust APIs without a compatibility path**

`CreativeOpportunitySlot::validate_runtime(&self, gam_network_id)` changed to `validate_runtime(&self)`, while the public `resolved_gam_unit_path` method was removed and its replacement is only `pub(crate)`. Source consumers and custom adapters using these public methods will fail to compile even though the change is presented as backward-compatible.

**Suggested fix:** Preserve the old signature where practical and retain `resolved_gam_unit_path` as a deprecated compatibility wrapper. Introduce a separately named path-aware API rather than removing the existing one silently.

docs/guide/configuration.md
```diff
@@ -1229,6 +1229,81 @@ TRUSTED_SERVER__AUCTION__TIMEOUT_MS=2000
...
+gam_network_id = "123456789"
+price_granularity = "dense"
+
+# Shared placeholder value for the site root ("/") — see {section} below.
+section_root = "home"
+
+[[creative_opportunities.slot]]
+id = "ad-header"
+gam_unit_path = "/{network_id}/example/{section}"
+page_patterns = ["/", "/news/*", "/reviews/*"]
```
ChristianPavilonis: 🔧 **The documented configuration does not match its `/news` result**

With the `glob` crate, `/news/*` does not match `/news`; it only matches `/news/` and descendants. The result table later claims `/news` produces `/123456789/example/news`, but this configuration produces no slot on that path. Publishers copying it will lose slots on section landing pages, and the renderer test bypasses `match_slots`, so it cannot catch the discrepancy.

**Suggested fix:** Include both `/news` and `/news/*` (and corresponding review patterns), or change the result table. Add a match-plus-render test for the exact documented paths.

ChristianPavilonis (changes requested):
## Summary

The implementation is generally well-tested and CI is green, but the current change has important configuration-compatibility and inventory-routing risks. In particular, config blobs can break deployment rollback, the template cache can silently discard explicit GAM paths outside the finalized-settings path, and the section derivation policy remains hardcoded despite the originating requirement.

crates/trusted-server-core/src/creative_opportunities.rs
prk-Jr: Implemented. `section_segment` is now a config key (0-based index over non-empty path segments, default `0`), so both halves of the convention are publisher-controlled: `section_segment` picks which segment names the section, `section_root` names it when there is none.

`/en/news/article` with `section_segment = 1` renders `news`. `/en` has no segment at index 1, so it falls back to `section_root` rather than reusing the locale.

Also added `CreativeOpportunitiesConfig::section_for_path()` so a call site cannot apply one knob and forget the other, and marked the "locale offset" non-goal in the design doc as pulled into scope with a dated revision note. Docs, example config, and CHANGELOG updated; the claim that the *complete* convention was config-driven is gone either way.

Tests: `derive_section_uses_configured_segment_index`, `derive_section_uses_root_when_segment_index_out_of_range`, `section_for_path_applies_both_policy_knobs`, `build_slot_json_honours_configured_section_segment`.

crates/trusted-server-core/src/creative_opportunities.rs
prk-Jr: Confirmed and fixed. Verified the mechanism rather than assuming it: `ts config push` runs `edgezero_cli::run_config_push_typed::<TrustedServerAppConfig>`, which deserializes the TOML into the typed struct and re-serializes it into the blob — so `Option::None` did land in the payload as `"section_root": null`, and an older binary's `deny_unknown_fields` rejects it.

- `skip_serializing_if = "Option::is_none"` on `section_root`, and on the new `section_segment` for the same reason (chose `Option<usize>` over a plain `usize` with a serde default precisely so an unset key stays out of the blob).
- Both keys commented out in `trusted-server.example.toml`, since no shipped slot uses `{section}`.
- Test `section_root_is_omitted_from_serialized_config_when_unset` asserts absence when unset and round-trip when set; `section_segment_is_omitted_from_serialized_config_when_unset` covers the new key.

One limit worth stating plainly: this only makes rollback clean for configs that *omit* the keys. A config that actually sets `section_root`/`section_segment` is pinned to a binary that knows them — inherent to adding any key under `deny_unknown_fields`. That caveat is now in the CHANGELOG and the configuration guide.

crates/trusted-server-core/src/creative_opportunities.rs
prk-Jr: Correct, and it was already live in the tree: `publisher.rs` `make_slot()` builds a slot with `gam_unit_path: Some("/…/atf-sidebar")` and `compiled_unit: None`, which the old branch rendered as `/<network_id>/<slot_id>`.

`render_gam_unit_path` now matches on `(compiled_unit, gam_unit_path)`:

- `(Some(parts), _)` — compiled cache, hot path unchanged.
- `(None, Some(raw))` — re-parse the raw template per call and render it; on a parse error use the raw string verbatim (pre-templating behavior). A malformed template cannot reach a compiled config, since startup rejects it.
- `(None, None)` — the only case that yields the `/<network_id>/<slot_id>` default.

`template_uses_section` reads the raw template on the same fallback, so an uncompiled config can no longer skip the `section_root` requirement. The `compiled_unit` doc comment now says `None` means *not compiled*, not *absent*.

Tests: `render_gam_unit_path_honours_raw_template_without_compiled_cache` (built via `serde_json::from_value`, asserting the cache really is empty first), `render_gam_unit_path_honours_static_path_without_compiled_cache`, `validate_runtime_requires_section_root_for_uncompiled_template`.

Went with the fallback rather than a three-state enum: it matches the shape `matches_path` already uses for the glob cache, so both caches degrade the same documented way.

crates/trusted-server-core/src/creative_opportunities.rs
prk-Jr: Fixed. `CreativeOpportunitiesConfig::validate_runtime` now rejects a blank `gam_network_id`, which closes the hole exhaustively: a template is non-empty (parser rejects empty, slot validation rejects blank/whitespace), `{network_id}` is non-empty, `{section}` is non-empty (the sanitizer guarantees a non-empty result for any non-empty segment, and `section_root` is charset-validated non-empty whenever a template uses `{section}`), and `{slot_id}` is validated. So no substitution can render blank.

One deliberate narrowing: the check is gated on a non-empty slot list. An empty `slot` array disables the feature entirely — no path is ever rendered — so failing startup there would take a site down over a value it does not use. Covered both ways by `validate_runtime_rejects_blank_network_id` and `validate_runtime_allows_blank_network_id_when_no_slots_configured`.

crates/trusted-server-core/src/creative_opportunities.rs
prk-Jr: Pushing back on the deprecated wrapper, taking the underlying point.

`resolved_gam_unit_path(&self, gam_network_id)` is the path-independent resolver this issue filed as the bug — it is quoted verbatim in #954 as the thing that cannot vary by request path. A `#[deprecated]` shim would still compile for existing callers and return the untemplated path, so a source consumer would keep building and silently serve the wrong inventory. That is a worse failure mode than a build break, and it is exactly the class of silent-discrepancy problem the issue describes as "easy to ship and hard to notice."

`trusted-server-core` is `publish = false`, so there is no semver contract to honor here; the compile error is the safe signal.

The legitimate part of the finding was that the replacement was `pub(crate)`, leaving source consumers with a break and no migration target. Fixed: `render_gam_unit_path` and `derive_section` are now `pub`, plus a `pub` `CreativeOpportunitiesConfig::section_for_path()` that applies the section policy for you. `render_gam_unit_path`'s rustdoc names it as the replacement for `resolved_gam_unit_path`. Migration is `slot.resolved_gam_unit_path(id)` → `slot.render_gam_unit_path(id, &config.section_for_path(path))`.

Happy to reverse this if you still want the shim — say so and I will add it.

docs/guide/configuration.md
prk-Jr: Correct — `/news/*` needs the trailing separator, so the documented config produced no slot at all on `/news` while the table claimed a rendered path for it.

Fixed the config rather than the table, since losing slots on section landing pages is the worse outcome: `page_patterns = ["/", "/news", "/news/*", "/reviews", "/reviews/*"]` in both `docs/guide/configuration.md` and `trusted-server.example.toml`, with a comment stating why both forms are listed.

Two tests, both going through `match_slots` so they cannot bypass matching the way the renderer test did:

- `documented_page_patterns_match_and_render_their_documented_paths` — asserts each documented path matches exactly one slot *and* renders its documented unit path.
- `bare_section_pattern_does_not_match_without_trailing_separator` — pins the underlying glob behavior, so a future edit back to `/news/*` alone fails loudly.

Also added a second results table for the `section_segment = 1` locale case.

crates/trusted-server-core/src/creative_opportunities.rs
```diff
@@ -40,28 +150,128 @@ pub struct CreativeOpportunitiesConfig {
...
+    /// The URL→section convention is publisher-specific: a site that prefixes a
+    /// locale (`/en/news/article`) sets `section_segment = 1` to select `news`
+    /// instead of `en`. Paths with no segment at this index fall back to
+    /// [`section_root`](Self::section_root), so on `/en` a config with
+    /// `section_segment = 1` renders the root section.
+    ///
+    /// `Option` rather than a plain `usize` with a serde default for the same
+    /// rollback reason as [`section_root`](Self::section_root): an unset key
+    /// must not appear in the serialized config blob.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
ChristianPavilonis, line 245: 🔧 **P1 — Dynamic templates can silently change meaning after a binary rollback**

Templates using only `{network_id}` and/or `{slot_id}` do not require `section_root` or `section_segment`, and both fields are omitted during serialization. An older binary therefore accepts the persisted blob because it already understands `gam_unit_path`, but treats the placeholders as literal text. A rollback can start successfully and pass paths such as `/{network_id}/{slot_id}` directly to `googletag.defineSlot`, causing invalid slots, no-fill, or delivery against nonexistent inventory on both initial and SPA renders. This also contradicts the changelog claim that configurations omitting the new keys roll back cleanly.

**Suggested fix:** Persist an automatic compatibility marker whenever any dynamic placeholder is compiled. The smallest change is to materialize `section_segment = Some(0)` for every dynamic template before typed serialization; a dedicated `gam_unit_path_template_version` marker would be clearer. Add a push-shaped test proving the legacy schema rejects dynamic-template blobs while still accepting static and absent paths.

crates/trusted-server-core/src/creative_opportunities.rs
```diff
@@ -40,28 +150,128 @@ pub struct CreativeOpportunitiesConfig {
...
     pub fn validate_runtime(&self) -> Result<(), String> {
+        // Every rendered unit path either substitutes `{network_id}` or uses the
+        // `/<network_id>/<slot_id>` default, so a blank network id produces a
+        // path GAM cannot resolve (`""` or `//slot`). Reject at startup rather
+        // than shipping it to `googletag.defineSlot`.
+        //
+        // Gated on a non-empty slot list: an empty list disables the feature, so
+        // no path is ever rendered and the id is inert. Failing startup there
+        // would take a site down over a value it does not use.
+        if !self.slot.is_empty() && self.gam_network_id.trim().is_empty() {
```
ChristianPavilonis: 🔧 **P2 — Blank network-ID validation rejects previously valid static configurations**

This rejects a blank `gam_network_id` whenever any slot exists, although explicit static paths and templates without `{network_id}` never consume it. `main` accepted a blank ID when every slot supplied a nonempty explicit path, so an existing static configuration can now fail startup despite the compatibility promise that static `gam_unit_path` configurations remain unchanged.

**Suggested fix:** Require a nonblank network ID only if a slot has no `gam_unit_path` or its parsed/raw template contains `{network_id}`. Retain the current `{network_id}` regression test and add coverage for a blank ID with an explicit static path.

crates/trusted-server-core/src/creative_opportunities.rs
```diff
@@ -254,15 +480,83 @@ impl CreativeOpportunitySlot {
...
+    /// deserialized without [`CreativeOpportunitiesConfig::compile_unit_templates`])
+    /// re-parses its template on every call — same fallback shape as
+    /// [`matches_path`](Self::matches_path). It must never silently degrade to
+    /// the default path, which would bid against the wrong inventory.
     #[must_use]
-    pub fn resolved_gam_unit_path(&self, gam_network_id: &str) -> String {
-        self.gam_unit_path
-            .clone()
-            .unwrap_or_else(|| format!("/{}/{}", gam_network_id, self.id))
+    pub fn render_gam_unit_path(&self, gam_network_id: &str, section: &str) -> String {
```
ChristianPavilonis: 🔧 **P2 — Repeated placeholders can amplify a request path into an unbounded allocation**

Neither template compilation nor rendering bounds placeholder count or rendered length. A valid template repeating `{section}` many times multiplies a client-controlled path segment into a much larger `String`, followed by further JSON and script-escaping allocations. On memory-constrained edge/WASM runtimes, an accepted configuration plus a crafted matching path can exhaust an isolate.

**Suggested fix:** Define a maximum rendered GAM path length, calculate it with checked arithmetic before allocation, and fail or omit the slot recoverably when exceeded. Add coverage for repeated placeholders and a long section.

docs/guide/configuration.md
```diff
@@ -1266,6 +1266,111 @@ TRUSTED_SERVER__AUCTION__TIMEOUT_MS=2000
...
+`gam_unit_path` is a template. A publisher whose ad unit varies by site section
+expresses that in **one** slot rule instead of one rule per (slot × section).
+
+Supported placeholders:
+
+| Placeholder    | Resolves to                                                    |
+| -------------- | -------------------------------------------------------------- |
+| `{network_id}` | `gam_network_id`                                               |
+| `{slot_id}`    | the slot's `id`                                                |
+| `{section}`    | first path segment of the request (see derivation rules below) |
```
ChristianPavilonis: 🔧 **P3 — Documentation still contains the pre-`section_segment` description**

This table says `{section}` is always the first path segment, while the implementation and the following section use the configured segment. Related rustdoc in `creative_opportunities.rs` also describes outdated first-segment and compiled-only behavior.

```suggestion
| `{section}`    | configured non-empty path segment (see derivation rules below) |
```

Please also update the `section_root` field docs and clarify that raw-template fallback enforces the `section_root` requirement even without the cache, while compilation remains required to reject malformed templates.

ChristianPavilonis (approved):
## Summary

Reviewed the current head after the previous feedback round. The request-path threading, parser, sanitization, initial/SPA consistency, and prior fixes are generally sound, and CI is green. I found one high rollback-compatibility risk, two medium correctness/resource risks, and one low documentation inconsistency; details are inline.

docs/guide/configuration.md
aram356: 🔧 **wrench** — This table cell still documents pre-`section_segment` behavior: `{section}` is not the "first path segment of the request" whenever `section_segment` is set. The derivation rules below this table are correct; the table contradicts them.

The same staleness exists in the `section_root` rustdoc (`creative_opportunities.rs:153`, "when the request path has no first segment") — with `section_segment = 1`, `/en` *has* a first segment and still falls back to `section_root`.

**Fix** (both places):

```markdown
| `{section}`    | the non-empty path segment at `section_segment` (default: the first; see derivation rules below) |
```

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `creative_opportunities.slot.gam_unit_path` is now a template supporting `{network_id}`, `{slot_id}`, and `{section}`, so a publisher whose ad unit varies by site section expresses it in one slot rule instead of one per (slot × section). `{section}` derives from the request path: `[creative_opportunities].section_segment` selects which path segment names the section (0-based, default `0`; set `1` for locale-prefixed URLs), and `section_root` supplies the value for paths with no such segment. `section_root` is required when a template uses `{section}`. Existing static and absent `gam_unit_path` configs are unchanged. Startup now also rejects a blank `gam_network_id` when slots are configured. Note that a config setting `section_root` or `section_segment` requires a binary that knows those keys — rolling the binary back below this release while the keys are present fails config load; configs that omit them roll back cleanly.
```
aram356: 🔧 **wrench** — "configs that omit them roll back cleanly" is affirmatively misleading for templates that use only `{network_id}` and/or `{slot_id}`. Such a config omits both new keys, so an older binary **accepts** the blob at load — and then renders the braces literally into `googletag.defineSlot`, silently targeting a nonexistent ad unit. That is the worst failure mode: no startup error, wrong inventory.

**Fix** (docs-only): state here — and in the rollback paragraph of `docs/guide/configuration.md` — that any `gam_unit_path` containing placeholders must be reverted to a static path before rolling the binary back below this release; only fully static configs roll back cleanly.

crates/trusted-server-core/src/creative_opportunities.rs
```diff
@@ -40,28 +150,128 @@ pub struct CreativeOpportunitiesConfig {
...
+        // Every rendered unit path either substitutes `{network_id}` or uses the
+        // `/<network_id>/<slot_id>` default, so a blank network id produces a
+        // path GAM cannot resolve (`""` or `//slot`). Reject at startup rather
+        // than shipping it to `googletag.defineSlot`.
+        //
+        // Gated on a non-empty slot list: an empty list disables the feature, so
+        // no path is ever rendered and the id is inert. Failing startup there
+        // would take a site down over a value it does not use.
+        if !self.slot.is_empty() && self.gam_network_id.trim().is_empty() {
+            return Err("gam_network_id must not be empty".to_string());
```
aram356, line 331: 🤔 **thinking** — This rejection is broader than what actually consumes the id: a `main`-era config where every slot has an explicit static `gam_unit_path` and a blank `gam_network_id` started fine before and now fails startup. The CHANGELOG documents the break, so this is defensible as-is — but the check could be scoped to configs that consume the id:

```rust
let network_id_consumed = self.slot.iter().any(|slot| {
    slot.gam_unit_path.is_none() || slot.template_uses_network_id()
});
if network_id_consumed && self.gam_network_id.trim().is_empty() {
    return Err("gam_network_id must not be empty".to_string());
}
```

(with a `template_uses_network_id` sibling of `template_uses_section`). That keeps the fail-loud property for every config that can render the id while not breaking previously-working all-static configs.

crates/trusted-server-core/src/creative_opportunities.rs
```diff
@@ -14,6 +14,116 @@ use crate::auction::types::{AdFormat, AdSlot, MediaType};
...
+    if !literal.is_empty() {
+        parts.push(UnitTemplatePart::Literal(literal));
+    }
+    Ok(parts)
+}
+
+/// Collapses each run of characters outside `[A-Za-z0-9_-]` to a single `_`.
+///
+/// Returns a non-empty string for any non-empty input.
+fn sanitize_section(segment: &str) -> String {
```
aram356, line 140: 🤔 **thinking** — Nothing bounds the rendered path size: a template repeating `{section}` multiplies a client-controlled path segment (bounded only by the platform URL limit) into a larger `String`, followed by JSON and script-escaping allocations. The amplification factor is config-controlled, so real-world risk is low — but a cap here is one line and makes the bound structural. GAM ad unit code levels max out at 100 characters anyway, so truncating the sanitized segment loses nothing legitimate:

```rust
// GAM ad unit codes cap at 100 chars per level; bound client-controlled input.
out.truncate(100);
out
```

(`out` is ASCII-only by construction, so byte truncation is char-safe.)

docs/guide/configuration.md
```diff
@@ -1266,6 +1266,111 @@ TRUSTED_SERVER__AUCTION__TIMEOUT_MS=2000
...
+pre-templating behavior, so existing static configs are unchanged.
+
+### `{section}` derivation
+
+`{section}` is derived from the request path at request time:
+
+- It is the non-empty path segment at `section_segment` (0-based, default `0`).
+  With the default, `/news/article-123` → `news`. A site that prefixes a locale
+  sets `section_segment = 1`, so `/en/news/article` → `news` rather than `en`.
+- It is sanitized: each run of characters outside `[A-Za-z0-9_-]` becomes a
```
aram356, line 1368: 🤔 **thinking** — Worth a bullet here: the section keeps the URL's casing, and GAM ad unit paths are case-sensitive. `/News/article` renders `.../News`, which is a different (likely nonexistent) unit than `.../news`. A publisher with case-variant URLs hits this silently at request time. Either document that URLs must match the ad unit casing, or ASCII-lowercase in `sanitize_section`.

crates/trusted-server-core/src/creative_opportunities.rs
```diff
@@ -14,6 +14,116 @@ use crate::auction::types::{AdFormat, AdSlot, MediaType};
...
+///
+/// `section_segment` exists because the URL→section convention is
+/// publisher-specific: a site that prefixes a locale (`/en/news/article`) sets
+/// `section_segment = 1` to get `news` rather than `en`.
+///
+/// The path is used **raw** (not percent-decoded) so this stays consistent with
+/// how [`page_patterns`](CreativeOpportunitySlot::page_patterns) glob-match the
+/// same path — e.g. `/new%20s` yields `new_20s`, never the decoded `new_s`.
+#[must_use]
+pub fn derive_section(path: &str, section_root: &str, section_segment: usize) -> String {
```
aram356: ♻️ **refactor** — `derive_section` is `pub` in a `pub` module but is only called by `section_for_path` and same-module tests — and its own rustdoc steers callers to `section_for_path` instead. Making it private (or `pub(crate)`) enforces that guidance and keeps it off the crate's public API surface before external consumers can pin it (CLAUDE.md: avoid unnecessary `pub`).

crates/trusted-server-core/src/publisher.rs
```diff
@@ -3335,11 +3335,16 @@ pub(crate) fn build_empty_bids_script() -> String {
...
+pub(crate) fn build_slot_json(
     slot: &crate::creative_opportunities::CreativeOpportunitySlot,
     co_config: &crate::creative_opportunities::CreativeOpportunitiesConfig,
+    request_path: &str,
 ) -> serde_json::Value {
-    let gam_path = slot.resolved_gam_unit_path(&co_config.gam_network_id);
+    // `{section}` derives from the same raw path `page_patterns` matched
+    // against; `section_root` covers the no-segment case (`/`) and paths shorter
+    // than the configured `section_segment`.
+    let section = co_config.section_for_path(request_path);
```
aram356: ⛏ **nitpick** — `section_for_path` re-derives (and re-allocates) the same section once per slot for the same request path. Both callers could derive it once per request and pass `&str` into `build_slot_json`. Trivial at realistic slot counts — take it or leave it.

crates/trusted-server-core/src/creative_opportunities.rs
```diff
@@ -40,28 +150,128 @@ pub struct CreativeOpportunitiesConfig {
...
+            .slot
+            .iter()
+            .any(CreativeOpportunitySlot::template_uses_section)
+        {
+            match self.section_root.as_deref() {
+                Some(root)
+                    if !root.is_empty()
+                        && root
+                            .chars()
+                            .all(|c| c.is_ascii_alphanumeric() || c == '_' || c == '-') => {}
```
aram356: ⛏ **nitpick** — This inlines the `[A-Za-z0-9_-]` predicate that `sanitize_section` (line 90) also encodes. A shared `fn is_section_char(c: char) -> bool` keeps the two charsets from drifting if the set ever changes.

aram356 (changes requested):
## Summary

Reviewed the current head against `main`. The core design — startup-compiled templates, fail-loud validation, and the raw-path no-decode contract — is sound, and the initial-render and SPA paths cannot diverge: both share `build_slot_json`, and the SPA `path` parameter round-trips percent-encoding correctly (`encodeURIComponent` → `form_urlencoded` decode) before the same normalization. Two documentation corrections are required (both cheap): the placeholder table still describes pre-`section_segment` behavior, and the rollback note affirmatively claims a clean rollback for configs that would silently render placeholder braces literally on an older binary. Six non-blocking suggestions are inline.

## CI Status

- fmt: PASS
- clippy (all targets): PASS
- js tests (vitest): PASS
- integration + browser tests, CodeQL: PASS

crates/trusted-server-core/src/publisher.rs
```diff
@@ -3476,10 +3478,14 @@ fn build_slot_json(
...
     co_config: &crate::creative_opportunities::CreativeOpportunitiesConfig,
+    request_path: &str,
 ) -> String {
+    // `{section}` derives from the same raw path `page_patterns` matched
+    // against; derive it once for every slot on this request.
+    let section = co_config.section_for_path(request_path);
     let slots: Vec<serde_json::Value> = matched_slots
         .iter()
-        .map(|slot| build_slot_json(slot, co_config))
+        .filter_map(|slot| build_slot_json(slot, co_config, &section))
```
aram356, line 3517: ♻️ **refactor** — When `render_gam_unit_path` returns `None` at request time (a long URL segment inflating `{section}`), the slot is dropped with no log — here and at the `handle_page_bids` call site (line 3905). Startup validation cannot catch this case since it only checks configured values, so an operator debugging "this slot vanishes on some article URLs" has zero signal. Suggest a `log::warn!` naming the slot id and request path at both `filter_map` sites:

```rust
.filter_map(|slot| {
    let json = build_slot_json(slot, co_config, &section);
    if json.is_none() {
        log::warn!(
            "Omitting slot `{}`: dynamic gam_unit_path exceeds the render limit for path `{}`",
            slot.id,
            request_path
        );
    }
    json
})
```

crates/trusted-server-core/src/publisher.rs
```diff
@@ -3893,9 +3899,10 @@ pub async fn handle_page_bids(
     // Gate slots on the ad-stack kill switch / consent: when disabled, return no
     // slots so the SPA hook does not call `adInit()` / create GPT slots.
     let slots_json: Vec<serde_json::Value> = if ad_stack_enabled {
+        let section = co_config.section_for_path(&path_param);
         matched_slots
             .iter()
-            .map(|slot| build_slot_json(slot, co_config))
+            .filter_map(|slot| build_slot_json(slot, co_config, &section))
```
aram356, line 3930: 🤔 **thinking** — `matched_slots` feed `build_auction_request` before renderability is known, so a slot omitted from the response here still fired real SSP bid requests, and its wins land in the bid map with no client-side slot to render into. Same shape on the initial-render path. Only reachable on over-limit request paths, so the impact is small — but it spends partner request quota on impressions that cannot serve. Filtering `matched_slots` by renderability before the auction would close it; deferring with a note is also reasonable.

docs/guide/configuration.md
```diff
@@ -1304,6 +1304,128 @@ TRUSTED_SERVER__AUCTION__TIMEOUT_MS=2000
...
+| -------------- | ----------------------------------------------------------------------- |
+| `{network_id}` | `gam_network_id`                                                        |
+| `{slot_id}`    | the slot's `id`                                                         |
+| `{section}`    | non-empty path segment at `section_segment` (default: first; see below) |
+
+A template with **no** placeholders is used verbatim. A slot with **no**
+`gam_unit_path` falls back to `/<network_id>/<slot_id>`. Both preserve the
+pre-templating behavior, so existing static configs are unchanged.
+
+Trusted Server conservatively caps the whole rendered dynamic path at 100 UTF-8
```
aram356, line 1352: 🌱 **seedling** — Google's 100-character limit is per ad-unit *code* (per path level), not the whole path, so a legitimate multi-level dynamic path can exceed 100 bytes total and be rejected at startup or dropped per-request. The conservative whole-path cap is documented and safe to ship; worth revisiting (per-level capping, or a larger whole-path bound) if a real publisher configuration hits it.

aram356 (approved):
## Summary

Re-reviewed the head after the previous feedback round. All prior findings are addressed with regression tests: the rollback marker (`section_segment` materialized during typed/startup finalization) is verified against the actual `ts config push` path (`TrustedServerAppConfig` deserialize → `finalize_deserialized` → `prepare_runtime` → `compile_unit_templates`) and pinned by the legacy-schema tests; network-id validation is scoped to consumption; dynamic rendering is byte-bounded with startup validation of configured values and byte-boundary tests; the case-preservation guidance matches Google's documentation. Three non-blocking suggestions are inline.

## CI Status

All checks PASS on `52630418b` (fmt, clippy targets, all adapter tests, parity, vitest, integration + browser tests, CodeQL).

PR #918, opened 2026-07-16
Add first-class APS OpenRTB integration

ChristianPavilonis:
## Summary

- Replace the legacy APS contextual/TAM request path with a first-class APS OpenRTB provider using `/e/pb/bid`.
- Carry minimized, typed APS renderer state through direct auctions, navigation/page-bids, mediation, and GPT/Universal Creative rendering.
- Add default-off script-creative gating, opaque-origin rendering, nonce-bound messaging, PBS coexistence safeguards, migration docs, and regression coverage.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/integrations/aps.rs` | Builds independent APS OpenRTB requests, validates and reduces responses, creates minimized renderer envelopes, and serves the sandboxed static renderer. |
| `crates/trusted-server-core/src/auction/*` | Adds typed APS renderer metadata, preserves bid identity through floors and orchestration, and exposes APS winners without `adm`. |
| `crates/trusted-server-core/src/integrations/prebid.rs` | Prevents APS from also being sent to Prebid Server for Trusted Server APS cohorts. |
| `crates/trusted-server-core/src/{publisher.rs,creative_opportunities.rs}` | Carries APS renderer data through navigation/page-bids while retaining legacy slot configuration only for compatibility. |
| `crates/trusted-server-js/lib/src/integrations/aps/render.ts` | Validates exact renderer descriptors and envelopes, creates opaque sandboxed frames, and enforces fragment-bound nonces and readiness acknowledgements. |
| `crates/trusted-server-js/lib/src/integrations/gpt/index.ts` | Routes APS winners through the Universal Creative dynamic-renderer bridge without cache fetches, beacons, or native APS display calls. |
| `crates/trusted-server-integration-tests/browser/tests/shared/aps-renderer.spec.ts` | Verifies CSP sandboxing, nonce and replay protections, same-origin rejection, restrictive-CSP compatibility, and parent-DOM isolation. |
| `trusted-server.example.toml` and integration fixtures | Adds canonical `account_id`, the APS OpenRTB endpoint, and default-off `allow_script_creatives`. |
| `docs/**`, `CHANGELOG.md`, `TESTING.md` | Documents migration, security boundaries, orchestration behavior, rollout gates, and validation coverage. |

## Closes

Closes #764

## Test plan

- [x] `cargo test-fastly && cargo test-axum`
- [x] `cargo clippy-fastly && cargo clippy-axum`
- [x] `cargo fmt --all -- --check`
[...]

PR #963, opened 2026-07-24
Register APS renderer via bid.meta so it survives Prebid field stripping

### LINKED ISSUE #962: APS bids fail to render ("Missing ad markup or URL") — renderer descriptor lost before registration
## Summary

APS OpenRTB bids never render. Every APS bid fails with Prebid's `Missing ad markup or URL` (reason `noAd`), so the slot is served an empty creative.

## Root cause

APS is bid-by-reference: `interpretResponse` sets the renderer descriptor on the Prebid bid as the custom `trustedServerRenderer` field, and a `bidResponse` listener (`installApsBidResponseRegistry`) registers it into `window.tsjs.apsPrebidRenderers` keyed by Prebid's generated `adId`. The GAM Universal Creative later requests that `adId` and the bridge serves the APS creative.

Prebid normalizes each bid into its own internal object during `addBidResponse` and **drops unknown top-level fields**, so the custom `trustedServerRenderer` field is gone before the `bidResponse` listener runs. The listener sees `renderer === undefined`, returns without registering, and the registry stays empty. When the Universal Creative requests the descriptor, the bridge finds nothing and Prebid's default renderer throws `Missing ad markup or URL`.

## Evidence

Observed on a live deployment with a runtime-instrumented registry:
- The custom field is absent as early as the `bidAccepted` event (the earliest bid event), for every APS bid.
- The renderer registry receives **zero** writes across a full session (`registrationsCount: 0`, validated by proxying the registry object).
- The descriptor itself is valid and registers correctly when exercised in isolation — so it is delivered by `/auction` and set by `interpretResponse`, then lost in Prebid's pipeline.
- Result: all APS render failures are `bidder=aps`, `reason=noAd`.

## Impact

APS renderer bids do not render at all. This affects any deployment serving first-class APS OpenRTB bids.

## Fix direction

Make registration independent of the custom field surviving Prebid's normalization: also stash the descriptor keyed by `requestId` (a first-class field Prebid preserves) at `interpretResponse` time, and fall back to it in the `bidResponse` listener when the custom field is absent.

aram356:
Fixes #962. Stacked on #918 (targets `issue-764-aps-openrtb`).

## Problem

APS bid-by-reference renderers never register, so every APS bid renders empty and Prebid throws `Missing ad markup or URL` (reason `noAd`).

`interpretResponse` sets the renderer descriptor on the bid as the custom `trustedServerRenderer` field; the `bidResponse` listener registers it into `window.tsjs.apsPrebidRenderers` keyed by the generated `adId`. But Prebid drops unknown top-level fields when it normalizes bids during `addBidResponse`, so the custom field is gone before the listener runs (absent as early as `bidAccepted`). The listener sees `renderer === undefined`, returns, and the registry stays empty.

## Fix

Carry the descriptor in `bid.meta` as well. `meta` is a first-class Prebid bid field this adapter already populates (`meta.advertiserDomains`), and Prebid's `bidderFactory` explicitly assigns `bid.meta = bidResponse.meta` onto the normalized bid — the same preservation guarantee `requestId` has. Because `meta` is per bid, multiple APS bids on the same imp each keep their own descriptor.

- `interpretResponse` (`auctionBidsToPrebidBids`): sets `meta[trustedServerRenderer]` alongside the existing top-level field
- `bidResponse` listener: `bid[APS_RENDERER_FIELD] ?? meta?.[APS_RENDERER_FIELD]`, then scrubs **both** copies unconditionally after the registration attempt
- No module-global state: no stash `Map`, no size cap, no eviction, no take-on-read lifecycle

An earlier revision stashed descriptors in a module-global `Map` keyed by `requestId`; review flagged that the key is per-adUnit (collides for multiple APS bids on one imp) and that entries could outlive their auction. The `meta` carrier eliminates that machinery entirely.

## Verification

- Unit tests: registration via `meta` with the top-level field stripped; distinct renderers registered for two APS bids sharing one imp/`requestId`; a stripped bid carrying no descriptor registers nothing (no stale reuse); capability scrubbed from the delivered bid after registration.
- `npx vitest run` in `crates/trusted-server-js/lib` against the post-rebase tree: 481 passed, 1 pre-existing failure in `gpt/spa_hook.test.ts` that also fails on clean `issue-764-aps-openrtb` (unrelated to this PR). `npm run format` clean; `node build-all.mjs` builds.

## Branch history

Recreated from current `origin/issue-764-aps-openrtb` with the fix reapplied as a single commit, per review — the previous 43-file diff was an artifact of the base rebase. The base's APS security fixes (debug-header allowlist, streaming, inline SSAT) are all preserved.


crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -291,394 +322,753 @@ impl IntegrationConfig for ApsConfig {
...
-            return Err(());
+    fn serialize_openrtb_request(
+        request: &OpenRtbRequest,
+    ) -> Result<Vec<u8>, Report<TrustedServerError>> {
+        serde_json::to_vec(request).change_context(TrustedServerError::Auction {
+            message: "Failed to serialize APS OpenRTB request".to_string(),
+        })
+    }
+
+    fn debug_headers(headers: &HeaderMap) -> BTreeMap<String, Vec<String>> {
```
ChristianPavilonis, line 587: 🔧 **The stale head reintroduces client-visible leakage of all APS upstream response headers**

`debug_headers` copies every upstream response header into APS debug metadata, which is returned to page JavaScript. This can expose `Set-Cookie`, identity headers, internal tracing data, or authentication-related metadata whenever APS debugging is enabled. The current base intentionally fixes this with a fail-closed allowlist containing only `Content-Type`.

Please preserve that allowlist while rebasing. Any additional debug headers should be explicitly reviewed and redacted rather than copied wholesale.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -501,6 +550,43 @@ function collectAuctionEids(): AuctionEid[] | undefined {
...
+  if (prebid[APS_BID_RESPONSE_LISTENER_SENTINEL] === true) return;
+
+  pbjs.onEvent('bidResponse', (rawBid) => {
+    const bid = rawBid as unknown as Record<string, unknown>;
+    if (bid['adapterCode'] !== ADAPTER_CODE || bid['bidderCode'] !== APS_BIDDER_CODE) {
+      return;
+    }
+    // Prefer the custom field; fall back to the requestId stash when Prebid has stripped
+    // it during bid normalization (the field is often gone before this listener runs).
+    const renderer = bid[APS_RENDERER_FIELD] ?? takePendingApsRenderer(bid['requestId']);
```
ChristianPavilonis: 🔧 **Pending renderer descriptors are not consumed when the custom field survives**

Because `??` short-circuits, `takePendingApsRenderer` is never called when `trustedServerRenderer` survives normalization. The descriptor stashed earlier therefore remains until FIFO eviction. A renderer envelope may contain up to 349,528 base64 characters, so retaining 256 entries can preserve roughly 85 Mi characters on a long-lived publisher or SPA page; it also leaves stale descriptors available if a request ID is reused.

Consume the pending value before selecting the primary/fallback renderer:

```suggestion
    const pendingRenderer = takePendingApsRenderer(bid['requestId']);
    const renderer = bid[APS_RENDERER_FIELD] ?? pendingRenderer;
```

Please also add a regression test that registers through the surviving custom-field path, then sends a field-stripped event with the same request ID and verifies that no stale renderer is registered again.

ChristianPavilonis (changes requested):
## Summary

The requestId fallback is directionally correct, but this PR is not safe to merge in its current state. The stacked base was rebased: GitHub reports the PR as conflicting, the merge base is stale, and the displayed change set has expanded to 43 files rather than the intended two-file fix. The head omits current-base commits covering publisher streaming, inline SSAT rendering, APS security review fixes, and rebased CI fixes; the final-tree comparison removes approximately 1,426 lines from `streaming_processor.rs` alone.

Please recreate or rebase this branch from the current `origin/issue-764-aps-openrtb`, then cherry-pick or reapply only `7406ed7d`. Confirm that the resulting PR contains only the intended Prebid implementation and test changes, and rerun the complete CI gates. The currently green integration checks predate the latest base commits and do not validate the conflicting merge result.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -204,23 +238,38 @@ export function auctionBidsToPrebidBids(auctionBids: AuctionBid[], bidRequests:
...
-      ttl: 300,
-      creativeId: bid.creativeId,
-      netRevenue: true,
-      currency: 'USD',
-      bidderCode: bid.seat,
-      meta: {
-        advertiserDomains: bid.adomain,
+    const requestId = origReq?.bidId ?? bid.impid;
+    // Stash by requestId so registration survives Prebid stripping the custom field.
+    if (renderer) stashPendingApsRenderer(requestId, renderer);
```
prk-Jr: 🔧 **wrench** — Stash key collides when one imp gets multiple APS bids.

`requestId` here is `origReq?.bidId ?? bid.impid` — a per-**adUnit** value, not per-bid. `parseAuctionResponse` flattens every `seatbid[].bid[]` (`src/core/auction.ts`), so two or more APS bids for the same `impid` map to the same `requestId`. The second `stashPendingApsRenderer` call overwrites the first, and `takePendingApsRenderer` deletes on read — so the first `bidResponse` consumes the single entry and the second bid registers nothing.

If the auction winner is that second bid, the slot renders empty and Prebid throws `Missing ad markup or URL` — the exact failure this PR fixes. The old custom-field path did not have this collision because each normalized bid object carried its own copy.

**Fix**: key the stash per bid rather than per request, or store a list per `requestId`:

```ts
// per-bid key
const rendererKey = `${requestId}|${bid.creativeId}|${index}`;
if (renderer) stashPendingApsRenderer(rendererKey, renderer);
```

…which requires the key to be reconstructible in the `bidResponse` listener, so a list-per-`requestId` (shift on read) is likely the smaller change.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -501,6 +550,43 @@ function collectAuctionEids(): AuctionEid[] | undefined {
...
+      bid['adId'],
+      bid['adUnitCode'],
+      renderer,
+      bid['ttl'],
+      {
+        markWinner: () => markWinner(rawBid),
+        markRendered: () => markBidAsRendered(rawBid),
+      }
+    );
+    if (registered) {
```
prk-Jr: 🔧 **wrench** — The descriptor scrub silently became conditional.

Before this PR the listener deleted `bid[APS_RENDERER_FIELD]` unconditionally, immediately after the registration attempt. Now the delete only runs on the success branch, so a descriptor that **fails** registration stays attached to the Prebid bid object and travels onward into targeting and analytics consumers.

That contradicts the comment directly below it ("Keep the executable capability only in the bounded, one-time registry") and the PR description's claim that the existing custom-field path is unchanged.

**Fix**: restore the unconditional scrub.

```ts
const registered = registerApsPrebidRenderer(/* … */);
delete bid[APS_RENDERER_FIELD];
if (!registered) {
  log.warn('[tsjs-prebid] rejected APS renderer capability that failed registration');
}
```

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -188,6 +193,35 @@ function recordUserIdModuleDiagnostics(): PrebidUserIdDiagnostics {
...
 
+// Prebid normalizes each bid into its own internal object during `addBidResponse`,
+// which can drop unknown top-level fields — so the custom `trustedServerRenderer`
+// descriptor set in `interpretResponse` may be gone by the time the `bidResponse`
+// listener runs (observed in production: the field is absent as early as `bidAccepted`).
+// To make registration independent of custom-field survival, the descriptor is also
+// stashed here keyed by `requestId` (a first-class field Prebid preserves) at
+// `interpretResponse` time, and the `bidResponse` listener falls back to it. Bounded.
+const MAX_PENDING_APS_RENDERERS = 256;
+const pendingApsRenderersByRequestId = new Map<string, unknown>();
```
prk-Jr: ♻️ **refactor** — `bid.meta` may remove the need for module-global state entirely.

`meta` is a first-class Prebid bid field that this adapter already populates a few lines down (`meta: { advertiserDomains: bid.adomain }`) and that Prebid preserves through `addBidResponse` — the same property `requestId` is being relied on for here. Carrying the descriptor as `meta.trustedServerRenderer` would eliminate this `Map`, the 256-entry cap, the FIFO eviction, the take-on-read lifecycle, and the key-collision problem flagged on the stash call below.

Worth confirming against the same live deployment used to validate this fix before committing to module-global state, since the stripping behaviour was observed empirically rather than derived from Prebid's contract.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
prk-Jr: 🤔 **thinking** — Two stash-lifetime leaks, both bounded but avoidable.

1. When the custom field *does* survive, `??` short-circuits and the stashed copy for that `requestId` is never taken — it lingers until FIFO eviction. Taking unconditionally (or deleting after a successful registration) would keep the map tight.
2. Bids Prebid rejects before delivery (below floor, size filtered, invalid) never reach this listener at all, so their descriptors are also held until eviction.

Neither is unbounded, but clearing the map on `auctionEnd` would collapse both cases and make the map's lifetime match the auction it belongs to.

prk-Jr (changes requested):
## Summary

The root cause is well diagnosed and the fix is the right shape: `requestId` is a first-class field Prebid preserves, the fallback is additive, and the new test simulates the actual stripping rather than asserting on internals. Three things block merge: the PR cannot currently be merged into its base, the descriptor scrub was quietly narrowed to the success path, and the stash key is per-adUnit rather than per-bid so multiple APS bids on one imp lose all but the first renderer.

Note on review surface: the substantive change is one commit and two files (`crates/trusted-server-js/lib/src/integrations/prebid/index.ts` and its test). The other 41 files in the GitHub diff are an artifact of the base branch having been rebased under this branch — see the first finding.

## Blocking

### 🔧 wrench

- **PR is unmergeable into its base**: GitHub reports `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`. `issue-764-aps-openrtb` was rebased after this branch merged an older copy of it, so a test merge conflicts in 8 files: `auction/orchestrator.rs`, `integrations/aps.rs`, `integrations/prebid.rs`, `publisher.rs`, `browser/tests/shared/aps-renderer.spec.ts`, `gpt/index.ts`, `prebid/index.ts`, `prebid/index.test.ts`. Both files this PR actually changes are in that list, and the base's `bidResponse` listener still has the pre-fix shape, so the conflict resolution is not mechanical. Merge the current base in and re-verify the fix still applies on top of it.
- **Descriptor scrub became conditional**: `delete bid[APS_RENDERER_FIELD]` now only runs when registration succeeds, so a failed registration leaves the descriptor on the Prebid bid object (`prebid/index.ts:579`). Inline comment has the detail.
- **Stash key collides on multiple APS bids per imp**: `requestId` is per-adUnit, and `takePendingApsRenderer` deletes on read, so only the first `bidResponse` for a given imp registers (`prebid/index.ts:254`). Inline comment has the detail.

## Non-blocking

### ♻️ refactor

- **Consider `bid.meta` over a module-global `Map`**: `meta` is first-class, already populated by this adapter, and preserved through `addBidResponse` — it would remove the stash, the cap, the eviction, and the key collision above (`prebid/index.ts:204`).

### 🤔 thinking

- **Stash lifetime**: entries are left behind both when the custom field survives (`??` short-circuits before `take`) and when Prebid rejects a bid before delivery. Bounded by the 256 cap, but clearing on `auctionEnd` would scope the map to the auction it belongs to (`prebid/index.ts:564`).

### 👍 praise

- **Root-cause writeup and test design**: the commit message pins the failure precisely (field absent as early as `bidAccepted`, not merely at `bidResponse`), and the test removes the custom field to reproduce stripping instead of asserting on the stash internals. `registerApsPrebidRenderer` re-validating the descriptor (`aps/render.ts:210`) also means the new path inherits full validation rather than trusting the stash — good layering.

## CI Status

Four checks ran and all pass: browser integration tests, integration tests, integration tests (Fastly EC lifecycle), prepare integration artifacts.

The repository's main gates — `cargo fmt`, the six `clippy-*` targets, the four `cargo test-*` targets, the parity suite, and the JS build/test/format jobs — did not run on this PR, since it targets `issue-764-aps-openrtb` rather than `main`. Given the change is JS-only, please confirm `npx vitest run` and `npm run format` in `crates/trusted-server-js/lib` were run locally against the post-merge tree.

aram356:
Rebuilt the branch and reworked the fix per review — all blocking findings addressed:

**Branch recreation (@ChristianPavilonis)** — recreated from current `origin/issue-764-aps-openrtb` and reapplied the fix as a single commit (`6ab6d59ea`). The PR diff is back to the intended 2 files and GitHub reports it mergeable. The base's debug-header allowlist in `aps.rs` and the streaming/SSAT/CI commits are untouched.

**Conditional scrub (@prk-Jr)** — resolved. `delete bid[APS_RENDERER_FIELD]` runs unconditionally after the registration attempt (the rebased base had already restored this shape), and the `meta` copy is scrubbed in the same place.

**Stash key collision (@prk-Jr)** — resolved structurally by adopting your `bid.meta` suggestion instead of patching the stash. The descriptor now rides on `meta[trustedServerRenderer]`, which is per **bid** by construction, so two APS bids on one imp each carry and register their own renderer. New regression test covers exactly that scenario (shared `requestId`, distinct renderers, both registered).

**`bid.meta` refactor + stash lifetime (@prk-Jr)** — adopted; the module-global `Map`, 256 cap, FIFO eviction, and take-on-read lifecycle are gone entirely, which also collapses both lifetime leaks. On the "confirm empirically" caveat: Prebid 10.26's `bidderFactory` assigns `bid.meta = bidResponse.meta` (same object reference) onto the normalized bid before `addBidWithCode`, so `meta` has the same first-class preservation the observed-surviving `requestId` has. I'd still like to validate against the live deployment before merge — will follow up with that check.

**Stale-renderer regression test (@ChristianPavilonis)** — added, adapted to the stash-free design: a bid registers via the surviving custom-field path, then a field-stripped bid reusing the same `requestId` with no descriptor of its own registers nothing.

**CI note** — `npx vitest run` against the post-rebase tree: 481 passed; the single failure (`gpt/spa_hook.test.ts`) also fails on clean `issue-764-aps-openrtb`, so it's pre-existing on the base and unrelated to these files. `npm run format` and `node build-all.mjs` are clean.

crates/trusted-server-core/src/integrations/aps.rs
aram356: Resolved by the branch recreation: rebuilt from current `origin/issue-764-aps-openrtb` with only the JS fix reapplied (`6ab6d59ea`), so this PR no longer touches `aps.rs` at all. The fail-closed `debug_headers` allowlist (Content-Type only) from the base is preserved verbatim.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
aram356: Resolved by removing the stash entirely (`6ab6d59ea`): the descriptor now rides per-bid on `meta[trustedServerRenderer]`, so there is no pending entry to leave unconsumed when the custom field survives — the listener scrubs both the top-level field and the `meta` copy unconditionally after the registration attempt, and nothing outlives the bid object.

Added the requested regression test, adapted to the stash-free design: a bid registers through the surviving custom-field path, then a field-stripped bid reusing the same `requestId` with no descriptor of its own arrives — nothing is registered for it (`does not register anything for a stripped bid that carries no meta descriptor`).

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
aram356: Resolved structurally by adopting the `bid.meta` suggestion from your refactor comment (`6ab6d59ea`) instead of patching the stash key: the descriptor is carried on each built bid's `meta`, which is per-bid by construction, so two APS bids sharing an imp/`requestId` each keep and register their own renderer — no shared key, no take-on-read, no ordering dependency.

Regression test added: `registers a distinct renderer for each of multiple APS bids on one imp` builds two APS bids with the same `impid`/`bidId` and distinct renderers, delivers both field-stripped, and asserts both register under their own `adId`.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
aram356: Resolved (`6ab6d59ea`): the scrub is unconditional again — `delete bid[APS_RENDERER_FIELD]` runs immediately after the registration attempt regardless of outcome (the rebased base had already restored this shape, and the recreated branch keeps it), with the `log.warn` on failure after it. The new `meta` copy is scrubbed in the same place, so a failed registration leaves no descriptor on the bid object via either carrier. The existing malformed-renderer test still asserts the field is gone after a failed registration.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
aram356: Adopted (`6ab6d59ea`): the module-global `Map`, 256 cap, FIFO eviction, and take-on-read lifecycle are all gone; the descriptor rides on `meta[trustedServerRenderer]` next to the `advertiserDomains` this adapter already sets, and the listener falls back to it before scrubbing both copies.

On the empirical caveat: I checked Prebid 10.26's source — `bidderFactory`'s `onBid` assigns `bid.meta = bidResponse.meta` (same object reference) onto the normalized bid before `addBidWithCode`, so `meta` has the same first-class preservation guarantee as `requestId`, unlike unknown top-level fields. I still intend to validate against the same live deployment used for the original fix before merge and will report back on this thread.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
aram356: Both leaks are gone with the stash itself (`6ab6d59ea`): the descriptor now lives only on the bid object (`meta`), so its lifetime is exactly the bid's — no module-global copy to linger when the custom field survives, and a bid Prebid rejects before delivery takes its descriptor with it when the bid is garbage-collected. No `auctionEnd` hook needed.

aram356:
The failing checks were inherited: the base branch itself was red — its rebase had orphaned #867's streaming machinery in `streaming_processor.rs` (dead-code errors under `-D warnings` killed every workflow at compile). Fixed on the base in `bbaf5735a` (see the writeup on #918) and rebased this branch onto it (`ce2b7124a`, same two-file JS diff). CI is re-running.

aram356:
Rebased onto the recreated base (`488f7b108` — the new rebase restores the publisher streaming integration and its "Harden APS auction delivery and rendering" commit closes the foreign-frame Universal Creative leak that failed the browser suite). Same two-file JS diff on top (`6cc9983cc`); the `bidResponse` listener on the new base is still the pre-fix custom-field-only shape, so this PR's meta fallback remains necessary. Full JS suite (524 tests), format, and build pass against the new base.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -196,6 +196,15 @@ function recordUserIdModuleDiagnostics(): PrebidUserIdDiagnostics {
 /** Resolved endpoint — set by installPrebidNpm, read by the adapter. */
 let auctionEndpoint = '/auction';
 
+// Prebid normalizes each bid into its own internal object during `addBidResponse`,
+// which can drop unknown top-level fields — so the custom `trustedServerRenderer`
+// descriptor set in `interpretResponse` may be gone by the time the `bidResponse`
+// listener runs (observed in production: the field is absent as early as `bidAccepted`).
+// `meta` is a first-class field Prebid preserves through that normalization (its
+// bidderFactory assigns `bid.meta = bidResponse.meta` onto the normalized bid), so the
```
prk-Jr: ❓ **question** — I could not reproduce the stated stripping mechanism against the vendored Prebid, so I want to make sure this fixes the production cause rather than a symptom.

Against `prebid.js@10.26.0` (this repo's dependency), nothing in the bid path allowlist-copies a bid response:

- `bidderFactory.onBid` builds the normalized bid with `Object.assign(createBid(bidRequest), bid, pick(bidRequest, Object.keys(TIDS)))` (`src/adapters/bidderFactory.ts:355`) — that copies **all** own enumerable properties, unknown custom ones included.
- `addCommonResponseProperties` and `getPreparedBidForAuction` mutate the same object in place (`src/auction.ts:748`, `775-817`); no new object is constructed.
- `events.getEvents()` shallow-copies the event record only; `args` stay by reference (`src/events.ts:187`).

Empirical corroboration: the base branch's real-Prebid browser test `crates/trusted-server-integration-tests/browser/tests/shared/aps-renderer.spec.ts` drives `pbjs.requestBids` → `interpretResponse` → registry → Universal Creative and asserts the descriptor is delivered — and it passes on #918 with the top-level field as the *only* carrier.

So: which Prebid build showed the field missing in production (version and module set)? And is the deployed `trusted-prebid.js` — built separately via `build-prebid-external.mjs`, outside the main bundle path — actually rebuilt and shipped from this source? If production runs a stale or different Prebid, `meta` may not be what unblocks #962.

To be clear: `meta` is a strictly more robust carrier either way, so this is a confirm-the-root-cause question, not a defect in the diff.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -826,12 +839,15 @@ function installApsBidResponseRegistry(): void {
...
-      renderer === undefined
-    ) {
+    if (bid['adapterCode'] !== ADAPTER_CODE || bid['bidderCode'] !== APS_BIDDER_CODE) {
+      return;
+    }
+    // Prefer the custom top-level field; fall back to the per-bid copy in `meta` when
+    // Prebid has stripped the top-level one during bid normalization (the field is
+    // often gone before this listener runs).
+    const meta = bid['meta'] as Record<string, unknown> | undefined;
+    const renderer = bid[APS_RENDERER_FIELD] ?? meta?.[APS_RENDERER_FIELD];
```
prk-Jr, line 1024: 🤔 **thinking** — the `meta` copy is observable to third parties for a window before this listener scrubs it, in a way the unknown top-level field was not.

`bidAccepted` is emitted at `src/auction.ts:533`, and `bidResponse` only later at `:616` (with `bidAdjustment` earlier still, inside `getPreparedBidForAuction`). Any analytics adapter or publisher listener registered on those earlier events sees `meta.trustedServerRenderer`, which carries the full base64 `aaxResponse` — the validator ceiling is ~350 KB base64 (`MAX_RENDER_ENVELOPE_BASE64_BYTES` in `aps/render.ts`). `meta` is also precisely the field generic analytics adapters serialize wholesale, whereas an unknown top-level key usually is not.

Two options if that window matters:

1. Register (and scrub) on `bidAccepted` as well — `adId` is already assigned by `createBid`, so the entry can be built there and the `bidResponse` pass becomes a no-op fallback.
2. Carry a short handle in `meta` instead of the envelope, and keep the descriptor itself off the bid object.

Non-blocking; option 1 shrinks the exposure to just `bidAdjustment` for a few lines of change.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -238,6 +248,9 @@ export function auctionBidsToPrebidBids(auctionBids: AuctionBid[], bidRequests:
         bidderCode: bid.seat,
         meta: {
           advertiserDomains: bid.adomain,
+          // Carry the descriptor in `meta` too so registration survives Prebid
```
prk-Jr: ♻️ **refactor** — with `meta` in place, there are now two carriers and two scrub paths, and the top-level one is the copy this PR shows does not survive.

Grep says nothing outside this file and its tests reads top-level `trustedServerRenderer`, so it could be dropped entirely: one carrier, one `delete`, one exposure surface, and the `?? meta?.[APS_RENDERER_FIELD]` fallback collapses to a single read.

If the duplicate is deliberate belt-and-braces against a future Prebid that filters `meta` sub-keys, that intent is worth stating in the comment above — right now it reads as transitional.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -826,12 +839,15 @@ function installApsBidResponseRegistry(): void {
...
-      bid['bidderCode'] !== APS_BIDDER_CODE ||
-      renderer === undefined
-    ) {
+    if (bid['adapterCode'] !== ADAPTER_CODE || bid['bidderCode'] !== APS_BIDDER_CODE) {
+      return;
+    }
+    // Prefer the custom top-level field; fall back to the per-bid copy in `meta` when
+    // Prebid has stripped the top-level one during bid normalization (the field is
+    // often gone before this listener runs).
+    const meta = bid['meta'] as Record<string, unknown> | undefined;
```
prk-Jr: ⛏ **nitpick** — the cast is unchecked, so a truthy non-object `meta` (a module overwriting it with a string, say) passes the `if (meta)` guard at line 867 and the optional read at 849. A narrow `isRecord`-style check would make both sites safe without changing the happy path.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -196,6 +196,15 @@ function recordUserIdModuleDiagnostics(): PrebidUserIdDiagnostics {
 /** Resolved endpoint — set by installPrebidNpm, read by the adapter. */
 let auctionEndpoint = '/auction';
 
+// Prebid normalizes each bid into its own internal object during `addBidResponse`,
```
prk-Jr: ⛏ **nitpick** — this nine-line block floats between the `auctionEndpoint` declaration and `auctionBidsToPrebidBids`'s own doc comment, so it reads as documentation for the wrong thing. The same rationale then appears twice more (lines 251-253 and 843-845). Consider attaching the long version to `APS_RENDERER_FIELD` (line 41), where both the producer and the consumer can point at it, and keeping the two call sites to one line each.

crates/trusted-server-js/lib/test/integrations/prebid/index.test.ts
```diff
@@ -387,6 +391,136 @@ describe('prebid/installPrebidNpm', () => {
     );
   });
 
+  it('registers APS renderer via meta when Prebid strips the custom top-level field', () => {
```
prk-Jr, line 510: 🌱 **seedling** — these three tests hand-construct the bid Prebid "delivered", which means they assert the assumption under test (that Prebid preserves `meta` sub-keys) rather than verify it. That is the same blind spot that let the original `trustedServerRenderer` regression ship: the mocked path was always green.

The real-Prebid browser spec (`crates/trusted-server-integration-tests/browser/tests/shared/aps-renderer.spec.ts`) is where this can be proven — e.g. delete the top-level field from a `bidAdjustment` hook, then assert the registry still populates and that `meta.trustedServerRenderer` is gone from `pbjs.getBidResponsesForAdUnitCode(...)` after the auction. Worth a follow-up rather than a blocker.

crates/trusted-server-js/lib/test/integrations/prebid/index.test.ts
```diff
@@ -387,6 +391,136 @@ describe('prebid/installPrebidNpm', () => {
...
+
+    const entry = (window as any).tsjs.apsPrebidRenderers['stripped-field-ad-id'];
+    expect(entry).toEqual(
+      expect.objectContaining({ adUnitCode: 'div-aps', renderer, markWinner: expect.any(Function) })
+    );
+    // The capability is scrubbed from the delivered bid after registration.
+    expect(delivered.meta).not.toHaveProperty('trustedServerRenderer');
+  });
+
+  it('registers a distinct renderer for each of multiple APS bids on one imp', () => {
```
prk-Jr, line 557: 👍 **praise** — this is the test that makes the `meta` carrier convincing: two APS bids sharing one imp and therefore one `requestId`, each keeping its own descriptor. Together with the no-stale-reuse case below it, it encodes exactly the failure modes the previous `requestId`-keyed stash revision was flagged for, so a regression back to a per-adUnit key would now fail loudly.

prk-Jr (approved):
## Summary

Carrying the APS renderer descriptor in `bid.meta` alongside the custom top-level field, with both copies scrubbed after the registration attempt, is a sound and well-scoped change: `meta` is a first-class Prebid bid field with an open index signature (`BidMeta` in `src/bidfactory.ts`), it is per bid so multiple APS bids on one imp keep their own descriptor, and the whole module-global stash/eviction machinery from the earlier revision is gone. Approving — no defect found in the diff.

One question stays open on the root cause: I could not reproduce the stated stripping mechanism against the vendored Prebid, and the base branch's real-Prebid browser test passes with the top-level field as the only carrier. Details inline. It does not block merge, because `meta` is strictly more robust than what it replaces, but it does bear on whether this actually closes #962 in production.

## Blocking

### ❓ question

- **Stated stripping mechanism does not reproduce against `prebid.js@10.26.0`**: `bidderFactory.onBid` copies all own enumerable properties, `auction.ts` mutates the bid in place, and `events.getEvents()` shares `args` by reference; #918's browser test confirms the top-level carrier works end to end. Which Prebid build dropped the field in production, and is the separately built `trusted-prebid.js` rebuilt from this source? (`crates/trusted-server-js/lib/src/integrations/prebid/index.ts:204`)

## Non-blocking

### 🤔 thinking

- **`meta` is exposed to earlier listeners than the scrub**: `bidAccepted` (`auction.ts:533`) precedes `bidResponse` (`:616`), so analytics adapters and publisher listeners can observe `meta.trustedServerRenderer` with the full base64 `aaxResponse` (~350 KB ceiling) before this listener removes it. Registering on `bidAccepted` too would close most of the window. (`index.ts:849`)

### ♻️ refactor

- **Two carriers, two scrub paths**: nothing outside this file and its tests reads the top-level field anymore; dropping it collapses the fallback read and the second `delete`. If the duplicate is deliberate insurance, say so in the comment. (`index.ts:251`)

### 🌱 seedling

- **New tests assert the assumption, not the behavior**: all three hand-construct the delivered bid, so they cannot prove Prebid preserves `meta` sub-keys — the same blind spot that let the original regression ship. The real-Prebid browser spec can prove it. (`test/integrations/prebid/index.test.ts:394`)

### ⛏ nitpick

- **Orphan comment block**: nine lines sitting between `auctionEndpoint` and `auctionBidsToPrebidBids`'s own doc comment, with the rationale repeated twice more. (`index.ts:199`)
- **Unchecked `meta` cast**: a truthy non-object `meta` passes both the optional read and the `if (meta)` delete guard. (`index.ts:848`)

### 👍 praise

- **Multi-bid-per-imp and no-stale-reuse tests**: they encode precisely the failure modes the earlier `requestId`-keyed stash was flagged for, so a regression back to a per-adUnit key fails loudly. (`test/integrations/prebid/index.test.ts:440`)
- **No module-global state**: the `meta` carrier removes the stash `Map`, its size cap, and its take-on-read lifecycle outright.

## CI Status

Only four check runs exist for head SHA `6cc9983`: browser integration tests, integration tests, integration tests (Fastly EC lifecycle), prepare integration artifacts — all PASS. The `vitest`, `format-typescript`, `cargo fmt` and CodeQL checks that ran on #918 did **not** run on this commit, so the JS unit-test and formatting gates are unverified by CI here. Verified locally against this branch instead:

- vitest: PASS — 31 files, 524 tests passed (`npx vitest run` in `crates/trusted-server-js/lib`)
- prettier: PASS — `npx prettier --check "src/**/*.ts" "test/**/*.ts"` reports all files formatted
- browser integration tests: PASS (CI)
- fmt / clippy: not applicable — the diff is TypeScript only

ChristianPavilonis (approved):
## Summary

Reviewed the current two-file diff at `6cc9983c` against the PR's actual `issue-764-aps-openrtb` base. The per-bid `meta` carrier, fallback registration, validation, unconditional scrubbing, and regression coverage are coherent and narrowly scoped. No actionable defects found.

All four checks attached to the head SHA pass. Independent validation of the exact committed tree also passed the full Vitest suite (524 tests), focused Prebid suite (104 tests), changed-file Prettier check, standard JS build, and external Prebid bundle build.

👍 Approved.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
aram356: Adopted option 1 (`d3c850d6e`): the registration/scrub handler is now shared and installed on both `bidAccepted` and `bidResponse` — the descriptor is registered and both carriers scrubbed at `bidAccepted`, the first event after Prebid assigns `adId`, so `bidResponse` and later-event analytics consumers never observe `meta.trustedServerRenderer`. The `bidResponse` pass remains as a fallback that no-ops once the `bidAccepted` pass has scrubbed (verified against vendored 10.26.0 that both events emit the same in-place-mutated object). Exposure is now limited to `bidAdjustment`, as you projected.

Regression test added: `registers and scrubs on bidAccepted before later events can observe the descriptor` — registers via the `bidAccepted` listener, asserts both carriers are gone, then replays the same object through `bidResponse` and asserts a clean no-op.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
aram356: Kept both carriers, and documented the intent (`d3c850d6e`): the duplication is deliberate belt-and-braces while your root-cause question on the other thread is open — the vendored 10.26.0 build preserves the top-level field (your reading, which I verified), while the production observation behind #962 says a deployed build dropped it. Until live validation settles which build production runs, the top-level copy is the carrier the browser spec proves works and `meta` is the carrier that survives stripping, so dropping either now risks the environment the other one covers.

The rationale now lives on `APS_RENDERER_FIELD` per your placement comment, stating explicitly that the top-level copy is belt-and-braces against a future build filtering `meta` sub-keys. If live validation confirms `meta` survives the production build, collapsing to a single carrier is a safe follow-up.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
aram356: Fixed (`d3c850d6e`): the listener now narrows `bid['meta']` with a `typeof === 'object' && !== null` check before either the fallback read or the scrub, so a truthy non-object `meta` is treated as absent. Test added: `tolerates a non-object meta value on the bid` covers both a string `meta` with no top-level field (nothing registers, nothing throws) and a string `meta` alongside a surviving top-level field (registration unaffected).

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
aram356: Fixed (`d3c850d6e`): the floating block is gone — the full rationale (including why there are two carriers) now sits on `APS_RENDERER_FIELD`, and both the producer in `auctionBidsToPrebidBids` and the listener reference it in a line each.

crates/trusted-server-js/lib/test/integrations/prebid/index.test.ts
aram356: Filed as #983: extend `aps-renderer.spec.ts` to strip the top-level field from a `bidAdjustment` hook and assert the registry populates via `meta`, plus assert `meta.trustedServerRenderer` is absent from `pbjs.getBidResponsesForAdUnitCode(...)` after the auction (with the scrub now on `bidAccepted`, no later-event consumer should see it). Keeping it a follow-up per your framing since the browser suite exercises the vendored build, which the other thread shows preserves the top-level field anyway — the spec extension proves the `meta` path against real Prebid regardless.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
aram356: Your source reading checks out against the vendored build — I re-verified on `prebid.js@10.26.0`: `onBid`'s `Object.assign(createBid(bidRequest), bid, ...)` copies unknown own-enumerable fields, `addCommonResponseProperties`/`getPreparedBidForAuction` mutate in place, and `bidAccepted`/`bidResponse` emit the same object. So on this repo's build the top-level field survives, consistent with the browser spec passing on #918 with it as the only carrier.

What the repo can and cannot answer about production: `build-prebid-external.mjs` builds the external bundle from this tree's `prebid.js` dependency and stamps `prebidVersion` + sha256/SRI into `manifest.json`, and the server config pins `external_bundle_url` + SRI — so production executes whichever bundle+config pair was last pushed, not automatically this source. I can't reconstruct from the repo which pair was live when the #962 stripping was observed, and scripted fetches of the deployment are DataDome-gated (401), so the live check needs the tester-cookie session. That's the validation I committed to on the earlier thread; when I run it I'll report the deployed manifest's `prebidVersion`/sha and whether the top-level field is stripped there.

Why the fix doesn't hinge on the answer: `bid.meta = bidResponse.meta` predates 10.x by many majors, so `meta` is a safe carrier for any plausible deployed build, and the top-level copy is retained as belt-and-braces (now documented on `APS_RENDERER_FIELD`). If the live check shows production preserving the top-level field — i.e. #962's cause was elsewhere, e.g. a stale bundle since replaced — the meta carrier is harmless, and the registration move to `bidAccepted` in `d3c850d6e` plus #983's real-Prebid spec extension give us the observability to catch whatever does strip.

PR #965, opened 2026-07-24
Exclude GAM paths from Prebid refresh auctions

### LINKED ISSUE #964: Exclude GAM paths from Prebid refresh auctions
## Description

Trusted Server currently creates a Prebid refresh auction for every GPT slot. Some tracking-only GAM slots must still refresh in GAM but should not enter that auction.

## Expected behavior

- Operators can configure exact GAM ad-unit-path suffixes to exclude from Trusted Server refresh auctions.
- Matching slots retain their GPT refresh and stale Trusted Server/Prebid targeting is cleared.
- Normal slots continue to auction, including mixed global refreshes.
- Missing or failing GAM path metadata fails open.

## Acceptance criteria

- Configuration validates suffixes and injects them into the managed Prebid browser config.
- Explicit, global all-excluded, and mixed refreshes have the intended GAM and Prebid behavior.
- Initial-load refresh pass-through remains unchanged.
- Operator documentation covers matching and external bundle rollout.

ChristianPavilonis:
## Summary

- Add an operator-configured GAM ad-unit-path suffix opt-out for Trusted Server Prebid refresh auctions.
- Preserve GAM refreshes and normal-slot auctions, including mixed global refreshes.
- Document matching, validation, and external-bundle rollout requirements.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/integrations/prebid.rs` | Add validated, deduplicated suffix config and inject it into the managed Prebid browser config. |
| `crates/trusted-server-js/lib/src/integrations/prebid/index.ts` | Exclude matching GPT paths from synthetic refresh ad units while retaining their GPT refresh. |
| `crates/trusted-server-js/lib/test/integrations/prebid/index.test.ts` | Cover explicit, all-excluded, mixed, fail-open, and literal matching behavior. |
| `docs/guide/integrations/prebid.md` | Document configuration semantics and deployment requirements. |
| `trusted-server.example.toml` | Add a commented configuration example. |
| `docs/superpowers/{specs,plans}/2026-07-24-prebid-refresh-gam-path-opt-out*` | Add the approved design and implementation plan. |

## Closes

Closes #964

## Test plan

- [x] Other: Cloudflare and Spin tests; all configured clippy targets; integration parity tests; supported external Prebid bundle build to `/tmp/trusted-server-prebid-refresh-path-opt-out`.

## Checklist

- [x] No new logging was required; no `println!` was added
- [x] New code has tests


prk-Jr (changes requested):
## Summary

The refresh-path exclusion behavior, validation, injection, and test coverage look sound. One blocking CI failure must be resolved before merge.

## Blocking

### 🔧 wrench

- **TypeScript lint failure**: The new refresh-exclusion tests introduce 12 explicit `any` casts rejected by `@typescript-eslint/no-explicit-any` in `crates/trusted-server-js/lib/test/integrations/prebid/index.test.ts` (beginning at line 1534 and continuing through line 1695). Replace them with the existing typed test-window/PBJS helpers or narrow test interfaces so the blocking lint job passes.

## CI Status

- TypeScript lint/format job: **FAIL** — 12 `no-explicit-any` errors
- Rust tests and adapter checks: **PASS**
- Browser and integration tests: **PASS**
- Vitest: **PASS**
- Rust and docs formatting: **PASS**
- CodeQL and analysis: **PASS**

crates/trusted-server-js/lib/test/integrations/prebid/index.test.ts
```diff
@@ -1516,6 +1518,192 @@ describe('prebid/installRefreshHandler', () => {
...
+      getSlotElementId: vi.fn(() => 'div-ad-tracking'),
+      getAdUnitPath: vi.fn(() => '/123/trackingonly'),
+      getTargeting: vi.fn(() => []),
+      clearTargeting,
+    };
+    const pubads = {
+      refresh: originalRefresh,
+      getSlots: vi.fn(() => [gptSlot]),
+    };
+    (window as any).googletag = {
```
aram356: 🔧 **wrench** — The `format-typescript` CI check fails with 10 `@typescript-eslint/no-explicit-any` errors, all from the `(window as any)` / `(mockPbjs as any)` casts in these new tests (lines 1534-1667).

The file already defines `TestWindow` at line 190, and the new `beforeEach`/`afterEach` cleanup already casts through it. Extend its `__tsjs_prebid` shape and reuse it:

```ts
type TestWindow = Omit<Window, 'tsjs'> & {
  __tsjs_prebid?: {
    // existing fields...
    excludedGamAdUnitPathSuffixes?: string[];
  };
  // ...
};
```

then replace `(window as any)` with `(window as TestWindow)` throughout, and give `setTargetingForGPTAsync` a typed home on the mock instead of `(mockPbjs as any)`.

docs/guide/integrations/prebid.md
```diff
@@ -322,6 +327,41 @@ set = { placementId = "_s2sHeaderPlacement", keep = "server" }
...
+duplicates are injected once. Matching is literal: paths are not case-normalized or
+slash-normalized. Use a specific terminal GAM path segment, not a broad size rule or
+div ID.
+
+If GPT does not expose `getAdUnitPath()` for a slot or the getter fails, Trusted
+Server fails open and runs the normal refresh auction. The option affects only this
+Trusted Server GPT-refresh wrapper; it does not block direct publisher Prebid,
+APS, or other auction flows.
+
+The external Prebid bundle and the injected Trusted Server configuration must be
```
aram356: 🔧 **wrench** — This rollout requirement is no longer correct after #967 (merged to `main` after this branch's base). The refresh wrapper, including this filter, now ships in the server-served `tsjs-prebid` shim; the external bundle is pure Prebid.js. The filter and the injected `excludedGamAdUnitPathSuffixes` config therefore deploy together with the server, and no bundle regeneration is required for this feature.

The only remaining bundle caveat is the pre-split lockstep note this guide already carries in the External Bundle section (a bundle generated before the shim split still bakes in an old shim copy). Please rewrite this paragraph after rebasing, and correct the same claim in spec §8 and plan Task 5.3.

docs/superpowers/specs/2026-07-24-prebid-refresh-gam-path-opt-out-design.md
```diff
@@ -0,0 +1,331 @@
...
+`/trackingonly`; its element ID is unstable and cannot be used as configuration.
+
+Operators need a configuration-only way to identify that inventory by its supported
+GAM ad-unit path and prevent **this refresh wrapper** from running a Trusted Server
+or native-Prebid refresh auction for it. GPT must still refresh the slot so GAM can
+record the tracking impression. Publishers must not change component source.
+
+The successful 2026-07-23 browser capture established that GPT exposes
+`slot.getAdUnitPath()` and returned `/88059007/autoblog/trackingonly` for the
+tracking slot. The capture also showed that the current wrapper included that slot
```
aram356: 🔧 **wrench** — This commits a real production GAM network code and publisher name. CLAUDE.md forbids real domains, customer names, and similar real-world identifiers in docs, and this plan's own constraint ("Use only fictional paths and hostnames in checked-in tests and documentation") says the same.

Replace with a fictional path, e.g. `/123456/example-news/trackingonly`, and drop or genericize the production-host references in this section.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -1182,7 +1217,11 @@ export function installRefreshHandler(timeoutMs = 1500): void {
             log.error('[tsjs-prebid] refresh targeting failed', error);
           }
         }
-        originalRefresh(slots, opts);
+        // A bare refresh that filters slots must pass the resolved complete list
+        // to GPT; otherwise the excluded slots would be refreshed implicitly but
+        // would not be represented by the wrapper's concrete target set. Keep
+        // the existing bare-refresh delivery behavior when no filtering occurs.
+        originalRefresh(slots === undefined && hasExcludedSlots ? targetSlots : slots, opts);
```
aram356: 🤔 **thinking** — This conditional looks redundant, and the comment above it does not match GPT behavior. A bare `originalRefresh(undefined)` makes GPT refresh every registered slot, excluded ones included, so passing `undefined` here would already deliver the excluded slots' GAM refresh; they are "represented" by the bare call.

The only real effect of substituting `targetSlots` is snapshot pinning: slots defined between the wrap-time `getSlots()` snapshot and auction completion get skipped, and only when exclusions exist. That is an inconsistency between the two bare-refresh paths rather than a needed fix.

Suggest either simplifying to the pre-existing behavior:

```ts
originalRefresh(slots, opts);
```

or, if snapshot pinning is actually intended, applying it to all delayed bare refreshes and saying so in the comment.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -1120,9 +1142,22 @@ export function installRefreshHandler(timeoutMs = 1500): void {
         return originalRefresh(slots, opts);
       }
 
+      // Clear stale Trusted Server/Prebid targeting from independent slots before
+      // filtering so excluded slots still receive a clean GAM refresh.
       independentSlots.forEach(clearRefreshTargeting);
 
-      const adUnits = independentSlots.map((slot) => {
+      const excludedGamAdUnitPathSuffixes = new Set(
```
aram356: ♻️ **refactor** — The `Set` buys nothing here: matching in `isExcludedFromRefreshAuction` is `[...excludedGamAdUnitPathSuffixes].some(...)`, which re-materializes the Set into a fresh array for every slot and never uses membership lookup. Keep the plain injected array:

```ts
const excludedGamAdUnitPathSuffixes = getInjectedConfig()?.excludedGamAdUnitPathSuffixes ?? [];
```

and have the predicate take `readonly string[]` and call `.some()` on it directly.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -338,6 +345,58 @@ impl IntegrationConfig for PrebidIntegrationConfig {
...
+
+fn validate_excluded_gam_ad_unit_path_suffixes(values: &[String]) -> Result<(), ValidationError> {
+    for value in values {
+        validate_excluded_gam_ad_unit_path_suffix(value)?;
+    }
+
+    Ok(())
+}
+
+fn canonicalize_excluded_gam_ad_unit_path_suffixes(config: &mut PrebidIntegrationConfig) {
```
aram356, line 390: ♻️ **refactor** — Canonicalization has to be remembered at every production call site: today both `build()` and `validate_config_for_startup()` call this after `integration_config::<PrebidIntegrationConfig>()`, but a future third consumer would silently get the non-deduplicated list. Consider folding the dedup into one shared load-and-normalize helper (or into deserialization) so the canonical form is the only form that can escape.

aram356 (changes requested):
## Summary

The feature itself is in good shape: strict fail-closed config validation on the server, a fail-open matcher in the browser, and a thorough JS test matrix (explicit, all-excluded, mixed, fail-open, literal matching). Requesting changes because the PR is currently unmergeable and stale against `main` (#967 landed after this branch's base and invalidates the rollout documentation), CI fails on TypeScript lint, and the committed spec contains real production identifiers.

## Blocking

### 🔧 wrench

- **CI failure: `format-typescript`**: 10 `@typescript-eslint/no-explicit-any` errors in the new tests; use the existing `TestWindow` type instead of `as any` (crates/trusted-server-js/lib/test/integrations/prebid/index.test.ts:1534-1667, inline comment).
- **Merge conflict and stale base vs #967**: GitHub reports the PR as CONFLICTING (`index.test.ts` conflicts on a test merge of `main`). The base predates #967, which decoupled the prebid tsjs shim from the bundled Prebid.js and rewrote `index.ts`. Please rebase onto current `main`, then re-run the full gate list; the unchecked test-plan items (WASM build, `fastly compute serve`) should also be completed since the shim architecture underneath this change moved.
- **Rollout documentation wrong after #967**: the new guide section says the external bundle must be regenerated for the filter to apply; post-#967 the refresh wrapper ships in the server-served shim, so the filter and injected config deploy together with the server (docs/guide/integrations/prebid.md:359-364, inline comment; same claim in spec §8 and plan Task 5.3).
- **Real production identifiers in the committed spec**: real GAM network code and publisher name in the capture path; replace with fictional values per CLAUDE.md and the plan's own constraint (docs/superpowers/specs/2026-07-24-prebid-refresh-gam-path-opt-out-design.md:22, inline comment).

## Non-blocking

### 🤔 thinking

- **Redundant `targetSlots` substitution on bare refresh**: `originalRefresh(undefined)` already refreshes all slots, excluded ones included; the conditional only introduces inconsistent snapshot pinning and its comment misstates GPT behavior (crates/trusted-server-js/lib/src/integrations/prebid/index.ts:1224, inline comment).

### ♻️ refactor

- **Set adds no value in the suffix matcher**: keep the plain injected array instead of spreading a Set per slot (crates/trusted-server-js/lib/src/integrations/prebid/index.ts:1149, inline comment).
- **Canonicalization drift risk**: dedup is invoked separately in `build()` and `validate_config_for_startup()`; fold it into one shared load/normalize path so a future consumer cannot read the non-canonical list (crates/trusted-server-core/src/integrations/prebid.rs:390, inline comment).

### 🌱 seedling

- **Browser-side guard against a degenerate empty suffix**: `adUnitPath.endsWith('')` is always true, so an empty string in the injected list would exclude every slot. Server validation prevents it today; a one-line filter (`suffix.startsWith('/') && suffix.length > 1`) in the browser would preserve the fail-open guarantee if the injected config is ever produced by another path.

## CI Status

- fmt: PASS
- clippy: PASS (all six configured targets)
- rust tests: PASS (fastly, axum, cloudflare, spin, CLI, parity)
- js tests: PASS (vitest)
- js lint/format: FAIL (`format-typescript`, see blocking finding)
- docs format: PASS

crates/trusted-server-js/lib/test/integrations/prebid/index.test.ts
ChristianPavilonis: Resolved in 22c14deb4. After rebasing onto #967, the refresh tests use its typed test window and optional mock PBJS targeting method; the new test cases contain no explicit any casts. Added fail-open coverage for malformed injected suffix configuration. npm run lint and the full Vitest suite pass.

docs/guide/integrations/prebid.md
ChristianPavilonis: Resolved in 22c14deb4. The guide, design §8, and implementation plan now state that this filter ships in the server-served tsjs-prebid shim with the injected server configuration, so this option does not require a new external bundle. The pre-split bundle migration caveat remains.

docs/superpowers/specs/2026-07-24-prebid-refresh-gam-path-opt-out-design.md
ChristianPavilonis: Resolved in 22c14deb4. Replaced the captured production path/publisher reference with the fictional /123456/example-news/trackingonly example and generalized the external-page wording.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
ChristianPavilonis: Resolved in 22c14deb4. Removed the targetSlots substitution and now call originalRefresh(slots, opts) for both the no-auction fast path and post-auction completion. Bare-refresh tests now assert that GPT receives the original undefined slots argument.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
ChristianPavilonis: Resolved in 22c14deb4. The matcher now receives a readonly plain array and calls some() directly; the Set allocation/spread was removed. The injected value is additionally checked as an array and filtered to usable non-root slash-prefixed strings so malformed values fail open.

crates/trusted-server-core/src/integrations/prebid.rs
ChristianPavilonis: Resolved in 22c14deb4. Added one load_config helper that parses and canonicalizes the enabled Prebid config, used by both startup validation and runtime build. The canonicalization test now exercises both paths and asserts the runtime injected payload.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -690,6 +693,33 @@ function publisherZoneForRefresh(candidateCodes: Array<string | undefined>): str
...
 
+function isUsableRefreshAuctionExclusionSuffix(suffix: unknown): suffix is string {
+  return typeof suffix === 'string' && suffix.startsWith('/') && suffix.length > 1;
+}
+
+function refreshAuctionExclusionSuffixes(value: unknown): string[] {
+  return Array.isArray(value) ? value.filter(isUsableRefreshAuctionExclusionSuffix) : [];
+}
+
+function isExcludedFromRefreshAuction(
```
prk-Jr, line 704: 🌱 **seedling** — No signal is emitted when a slot is excluded, so a too-broad suffix is invisible in the field. Because `clearRefreshTargeting()` already ran and no replacement auction follows, a display slot that matches by accident loses Trusted Server demand on every refresh after the first impression, and the only symptom is missing `hb_*` targeting in GAM.

A single debug line (or a counter surfaced through the GPT diagnostics overlay from #974) would make that misconfiguration diagnosable without a code change:

```ts
const excluded =
  typeof adUnitPath === 'string' &&
  excludedGamAdUnitPathSuffixes.some((suffix) => adUnitPath.endsWith(suffix));
if (excluded) {
  log.debug(`[tsjs-prebid] refresh auction excluded ${adUnitPath}`);
}
return excluded;
```

Not for this PR if you'd rather keep the predicate allocation-free — noting it as the follow-up that makes the opt-out observable.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -1291,6 +1333,9 @@ export function installRefreshHandler(timeoutMs = 1500): void {
             log.error('[tsjs-prebid] refresh targeting failed', error);
           }
         }
+        // Preserve the publisher's original refresh form. In particular, a bare
```
prk-Jr, line 1336: 🤔 **thinking** — Correct for SRA, but worth naming the remaining cost: in a mixed global refresh the excluded slot's GAM refresh is still withheld until the auction completes (or the `timeoutMs` watchdog fires), because the whole original list goes out in one `originalRefresh(slots, opts)` call. So an excluded tracking slot avoids `/auction` but not the added latency — up to ~1.5s — unless every slot in that refresh is excluded, which is the only path that returns immediately.

That is the right tradeoff (splitting the refresh would break Single Request Architecture and change the publisher's request shape), and the design doc's §5.3 table does describe the ordering. The guide's caveat list is where an operator would look, though: it currently says matching slots "still refresh through GAM" without noting the mixed-refresh delay. One sentence there would set the expectation for anyone opting out a slot for timing reasons rather than for demand reasons.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -338,6 +345,70 @@ impl IntegrationConfig for PrebidIntegrationConfig {
     }
 }
 
+fn excluded_gam_ad_unit_path_suffix_validation_error(message: &'static str) -> ValidationError {
+    let mut error = ValidationError::new("invalid_gam_ad_unit_path_suffix");
+    error.message = Some(message.into());
+    error
+}
+
+fn validate_excluded_gam_ad_unit_path_suffix(value: &str) -> Result<(), ValidationError> {
```
prk-Jr, line 354: 🌱 **seedling** — The validator rejects every input that cannot be a path suffix except one that can never match: a trailing slash. `"/trackingonly/"` passes all four checks, injects fine, and then silently matches nothing, since GAM ad-unit paths do not end in `/` — the PR's own `'/123/trackingonly/'` test case documents that direction of the asymmetry.

Given the design deliberately does no slash normalization (a good call — literal matching is auditable), rejecting the dead form at startup is more consistent than accepting it:

```rust
if value.ends_with('/') {
    return Err(excluded_gam_ad_unit_path_suffix_validation_error(
        "excluded_gam_ad_unit_path_suffixes entries must not end with '/'",
    ));
}
```

This also subsumes the `value == "/"` case. Fine as a follow-up — the current behavior is fail-open, not wrong.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -338,6 +345,70 @@ impl IntegrationConfig for PrebidIntegrationConfig {
...
+    let mut canonical = Vec::with_capacity(config.excluded_gam_ad_unit_path_suffixes.len());
+    for suffix in std::mem::take(&mut config.excluded_gam_ad_unit_path_suffixes) {
+        if !canonical.contains(&suffix) {
+            canonical.push(suffix);
+        }
+    }
+    config.excluded_gam_ad_unit_path_suffixes = canonical;
+}
+
+fn load_config(
```
prk-Jr, line 400: 📝 **note** — This is the right resolution of the earlier canonicalization concern: `build()` and `validate_config_for_startup()` can no longer disagree, and the test now asserts both paths plus the injected payload.

For the record on what remains: canonicalization is still a property of this helper rather than of deserialization, so a future caller that reaches for `settings.integration_config::<PrebidIntegrationConfig>(PREBID_INTEGRATION_ID)` directly would get the raw list. Today that is only reachable from tests, and because matching is `some(endsWith)`, a duplicate suffix changes nothing observable beyond a slightly larger injected payload — so no action needed here.

prk-Jr (approved):
## Summary

Approving. Every finding from the previous round is resolved in `22c14deb`, and the re-review found no blocking issues. Reviewed at `22c14deb` in a clean worktree: all seven changed files read in full, plus the surrounding `installRefreshHandler()` flow, `Settings::get_typed()`, and the shim's browser bootstrap.

The design holds up under the checks that matter here:

- **Fail-closed server, fail-open browser.** Invalid suffixes are rejected in `get_typed()`, so the config-store deploy path is covered and not just TOML parsing. In the browser, a missing/non-string/throwing `getAdUnitPath()` and a malformed injected list all keep the slot auction-eligible, which is the safe direction — a bad match suppresses demand, a missed match only costs the optimization.
- **The leading-slash requirement is doing real work.** It turns `endsWith()` into an implicit segment-boundary match: `/123/nottrackingonly` does not match `/trackingonly`. That is what makes literal, un-normalized matching safe to expose to operators.
- **Excluded slots keep their GAM refresh.** They stay in `originalRefresh(slots, opts)`, `setTargetingForGPTAsync()` is scoped to the auction codes only, and `clearRefreshTargeting()` runs before filtering so a matching slot cannot serve on a stale header-bid winner.
- **The `adInitRefreshInProgress` bypass is untouched**, so the `disableInitialLoad()` handoff still delivers server-applied targeting to GAM.

### Resolved from the previous round

| Finding | Status |
| ------- | ------ |
| 12 `@typescript-eslint/no-explicit-any` errors failing `format-typescript` | Fixed — tests use the typed test window; `npm run lint` is clean at `--max-warnings=0` |
| Rollout doc invalidated by #967 | Fixed — guide, design §8, and plan all state the filter ships in the server-served `tsjs-prebid` shim, no bundle regeneration |
| Real GAM network code / publisher name in design doc | Fixed — now `/123456/example-news/trackingonly` |
| Redundant `targetSlots` substitution on the no-auction path | Fixed — both paths call `originalRefresh(slots, opts)`, with bare-refresh assertions |
| `Set` re-materialized per slot | Fixed — plain readonly array, `some()` called directly |
| Canonicalization repeated at each call site | Fixed — one `load_config()` used by build and startup validation |

## Non-blocking

### 🤔 thinking / 🌱 seedling / 📝 note

Left inline: mixed-refresh latency for excluded slots (`index.ts:1336`), no signal on exclusion (`index.ts:704`), trailing-slash suffix is dead-but-valid config (`prebid.rs:354`), and canonicalization scope (`prebid.rs:400`). None block merge.

### ⛏ nitpick

Design-record drift, docs only — the implementation is right and these are the checked-in artifacts describing it:

- `docs/superpowers/specs/2026-07-24-prebid-refresh-gam-path-opt-out-design.md:160` still says "derive a `Set`" after the array refactor.
- Same file at line 194, and the plan's acceptance criteria at line 196, say stale targeting is cleared from "every target slot". The implementation clears only `independentSlots` — publisher delivery slots intentionally keep their already-applied targeting, and the pseudocode omits that split entirely.

### 👍 praise

- The validation split is the part worth copying: strict and field-specific on the server where a bad value is an operator error, permissive in the browser where a bad value would cost impressions. The four rejection cases each carry their own message, so a startup failure names the problem.
- The browser test matrix goes past the happy path — `it.each` over missing, non-string, and throwing `getAdUnitPath()`, plus empty-string and non-array injected suffix lists, plus case-mismatch and trailing-slash cases that pin the literal-matching contract rather than just asserting it in prose.
- Rust coverage asserts the negative too: `excludedGamAdUnitPathSuffixes` is absent from the injected script when the list is empty, which is what keeps rollback and older-bundle pages behaviorally identical.

## CI Status

All 19 checks pass at `22c14deb`, including the three integration jobs that were still pending during the previous review.

- `cargo fmt`: PASS
- clippy (fastly, axum, cloudflare native + wasm, spin native + wasm): PASS
- rust tests (fastly, axum, cloudflare, spin, cross-adapter parity, ts CLI): PASS
- vitest: PASS
- `format-typescript` / `format-docs`: PASS
- integration tests, browser integration tests, Fastly EC lifecycle: PASS
- CodeQL / Analyze (rust, javascript-typescript, actions): PASS

Verified locally at `22c14deb`: `npx vitest run test/integrations/prebid/index.test.ts` (114 passed), `npm run lint` (clean), `cargo test -p trusted-server-core integrations::prebid` (147 passed).

docs/superpowers/specs/2026-07-24-prebid-refresh-gam-path-opt-out-design.md
```diff
@@ -0,0 +1,329 @@
...
+```html
+<script>
+  window.__tsjs_prebid={"excludedGamAdUnitPathSuffixes":["/trackingonly","/measurement-only"],...};
+</script>
+```
+
+For `[]`, omit the property with `skip_serializing_if`, matching the existing
+`clientSideBidders` convention. A browser that receives no property treats it as an
+empty list. This makes upgrade and rollback backwards compatible: old configuration
+has no behavior change; an older external bundle safely ignores the extra injected
```
aram356: ⛏ **nitpick** — Post-#967 the consumer of the injected property is the server-served shim, not the external bundle; §8 of this document already states that correctly. For consistency, this compatibility sentence could read: "an older shim safely ignores the extra injected property; and a newer shim with old configuration has no exclusions." Not blocking.

aram356 (approved):

PR #966, opened 2026-07-24
Prevent duplicate GPT slot requests

### LINKED ISSUE #944: Duplicate GPT slots per ad unit (TS container slot competes with publisher inner-div slot)
## Summary

TS defines a slot on a `...-container` div alongside the publisher's own inner-div slot on the same GAM unit path (`/88059007/prospect-a/homepage`), yielding ~9 registered slots for ~3-4 real placements. Risk of competing `defineSlot` / double ad requests.

## Evidence

`googletag.pubads().getSlots()` shows both `ad-header-0-_R_...-container` (TS) and `ad-header-0-_r_1_` (publisher) for the same unit; same pattern for `fixed_bottom` and `in_content`. Observed on a live Next.js App Router publisher with TS enabled.

## Verification — confirmed production impact

A fresh headless Chrome capture of the live Autoblog homepage on 2026-07-24 confirmed this is not just duplicate `getSlots()` bookkeeping: both slot sets issued successful GAM requests.

1. At `16:52:49.929Z`, GAM received a successful SRA request for **three TS-owned container slots**, all carrying `ts_initial=1`:
   - `ad-header-0-_R_…-container`
   - `ad-fixed_bottom-0-_R_…-container`
   - `ad-in_content-…-in_content-0-container`
2. At `16:52:53.419Z` (~3.5 seconds later), GAM received another successful SRA request for publisher-owned inner slots, including:
   - `ad-header-0-_r_1_`
   - `ad-fixed_bottom-0-_r_f_`
   - publisher `in_content` slots and the interstitial
3. GPT emitted `slotRequested` for both groups. The initial TS `in_content` slot and the later publisher header/fixed-bottom slots returned non-empty renders.

SRA combines slots *within* each request, but these are separate GAM requests. Header and fixed-bottom are clearly requested twice for the same semantic placement. The content element IDs change during React hydration, so the exact one-to-one content mapping needs a DOM-lifecycle trace; it does not change the confirmed header/footer duplication.

## Root cause

The current TS implementation deliberately takes this path when the publisher has not yet defined the inner slot:

```ts
// Use outer container div for TS's slot when publisher hasn't defined
// theirs yet — keeps both slots on separate divs so publisher's
// later defineSlot on the inner div doesn't conflict.
const containerEl = document.getElementById(`${actualDivId}-container`);
const slotDivId = containerEl?.id ?? actualDivId;
const defined = g.defineSlot?.(slot.gam_unit_path, slot.formats, slotDivId);
```

[Current source, lines 503–556](https://github.com/IABTechLab/trusted-server/blob/5efa34c95c6daca988489de79c94f1d76c88c097/crates/trusted-server-js/lib/src/integrations/gpt/index.ts#L503-L556) then calls `display()` for the TS-defined slot, whereas reused publisher slots are refreshed. This intentionally prevents a GPT conflict but creates two independently requestable slots when the publisher later defines the inner div.

## Impact

- Extra GAM requests and slot lifecycle work for the same user-visible placement.
- Potential duplicate auction/creative rendering and viewability/revenue-reporting distortion.
- Extra page/network cost. A duplicate GAM ad-unit path is valid GPT usage when it represents genuinely distinct placements; here the header and fixed-bottom pairs are not distinct placements.

## Acceptance

- One registered/requested GPT slot per placement, **or**
- an explicitly documented dual-ownership design that guarantees only one of the container/inner slots is displayed or refreshed and therefore only one GAM request is sent per placement.

## Suggested regression coverage

Model the observed ordering: TS `adInit()` executes before the publisher’s `defineSlot(innerDiv)`. Assert that the later publisher definition cannot result in a second displayed/refreshed GAM slot for the same configured placement, and capture the corresponding `slotRequested` events (not only `getSlots()`).

ChristianPavilonis:
## Summary

- Prevent competing GPT requests when TS initializes a placement before the publisher defines its inner div.
- Hand the TS-created inner-div slot to the late publisher definition and preserve SPA ownership cleanup.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-js/lib/src/integrations/gpt/index.ts` | Define fallback slots on the inner div and add scoped late-publisher definition/display/refresh handoff wrappers. |
| `crates/trusted-server-js/lib/src/core/types.ts` | Add shared `window.tsjs` handoff lifecycle state. |
| `crates/trusted-server-core/src/integrations/gpt_bootstrap.js` | Mirror the initial-load handoff behavior in the head bootstrap. |
| `crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts` | Add request-level tests for late definitions and `disableInitialLoad()`. |
| `crates/trusted-server-core/src/integrations/gpt.rs` | Assert the injected bootstrap uses inner-div handoff behavior. |
| `docs/superpowers/{specs,plans}/` | Add the design specification and implementation plan. |

## Closes

Closes #944

## Test plan

- [x] Other: `cargo test-axum`; target-matched `trusted-server-core` GPT tests; `cargo clippy-axum`

## Checklist

- [x] New code has tests


crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -449,9 +455,190 @@ function installInitialLoadDetector(ts: TsjsApi): void {
...
+  ts: TsjsApi,
+  pubads: GoogleTagPubAdsService,
+  adUnitPath: string,
+  formats: Array<number | number[]>,
+  elementId: string
+): GptSlotHandoff | undefined {
+  const exact = ts.gptSlotHandoffs?.[elementId];
+  if (exact) return exact;
+
+  const candidates = new Set(Object.values(ts.gptSlotHandoffs ?? {})).values();
```
aram356, line 754: 🔧 **wrench** — Publisher `defineSlot()` without a div ID throws here, breaking unrelated slots.

GPT's API makes the third parameter optional — `defineSlot(adUnitPath, size, div?)` is legal, with the slot displayed later via `display(slot)`. Once TS has created any fallback slot (the exact scenario this PR targets), every publisher `defineSlot` call flows through `matchingHandoff`, and a div-less call hits `elementId.startsWith(handoff.divIdPrefix)` with `elementId === undefined`:

```
TypeError: Cannot read properties of undefined (reading 'startsWith')
```

The publisher's call throws where native GPT would succeed — including for slots completely unrelated to TS placements. Reproduced with a Vitest against this module.

**Fix**: bail out of matching when the element ID is not a string, and short-circuit the internal guard before doing any matching work (currently `matchingHandoff` — `JSON.stringify` over all handoffs plus a `getSlots()` scan — runs even for TS's own internal calls):

```ts
const patchedDefineSlot = (
  adUnitPath: string,
  formats: Array<number | number[]>,
  elementId?: string
): GoogleTagSlot | null => {
  if (!ts.gptSlotHandoffInternal && typeof elementId === 'string') {
    const handoff = matchingHandoff(ts, pubads, adUnitPath, formats, elementId);
    ...
  }
  return originalDefineSlot(adUnitPath, formats, elementId as string);
};
```

Same bug exists in the bootstrap mirror (see companion comment). Please add regression tests for the div-less form to both the bundle and bootstrap-eval suites.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -42,6 +42,164 @@
...
+  function matchingHandoff(pubads, adUnitPath, formats, elementId) {
+    var exact = ts.gptSlotHandoffs && ts.gptSlotHandoffs[elementId];
+    if (exact) return exact;
+
+    var candidates = Object.values(ts.gptSlotHandoffs || {}).filter(
+      function (handoff, index, allHandoffs) {
+        return (
+          allHandoffs.indexOf(handoff) === index &&
+          !handoff.publisherClaimed &&
+          elementId.startsWith(handoff.divIdPrefix) &&
```
aram356, line 148: 🔧 **wrench** — Same crash as the bundle's `matchingHandoff`: a publisher `defineSlot(adUnitPath, sizes)` call with the optional div ID omitted reaches `elementId.startsWith(handoff.divIdPrefix)` with `elementId === undefined` and throws `TypeError` whenever an unclaimed handoff exists, breaking the publisher's ad setup. Guard `typeof elementId === "string"` (and check `ts.gptSlotHandoffInternal` before matching) here as well so both implementations stay in sync.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -449,9 +455,190 @@ function installInitialLoadDetector(ts: TsjsApi): void {
...
+  pubads: GoogleTagPubAdsService,
+  elementId: string
+): GoogleTagSlot | undefined {
+  return pubads.getSlots?.().find((slot) => slot.getSlotElementId() === elementId);
+}
+
+function handoffForSlot(ts: TsjsApi, slot: GoogleTagSlot): GptSlotHandoff | undefined {
+  return ts.gptSlotHandoffs?.[slot.getSlotElementId()];
+}
+
```
aram356, line 725: ♻️ **refactor** — This assumes a non-string target is either a same-realm `Element` or a slot. Anything failing `instanceof Element` (e.g. a cross-realm element from a friendly iframe) falls through to `target.getSlotElementId()` and throws `TypeError: target.getSlotElementId is not a function` (reproduced) — native GPT `display()` handles such targets. The bootstrap version duck-types safely; align the bundle with it since both implement one protocol:

```ts
function displayTargetElementId(target: GoogleTagDisplayTarget): string | undefined {
  if (typeof target === 'string') return target;
  if (typeof (target as GoogleTagSlot).getSlotElementId === 'function') {
    return (target as GoogleTagSlot).getSlotElementId();
  }
  return (target as Element).id || undefined;
}
```

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -449,9 +455,190 @@ function installInitialLoadDetector(ts: TsjsApi): void {
...
+                (ownedSlot) => ownedSlot !== existingSlot
+              );
+              if (
+                handoff.gamUnitPath !== adUnitPath ||
+                JSON.stringify(handoff.formats) !== JSON.stringify(formats)
+              ) {
+                log.warn('GPT slot handoff: publisher definition differs from TS configuration', {
+                  elementId,
+                  tsGamUnitPath: handoff.gamUnitPath,
+                  publisherGamUnitPath: adUnitPath,
```
aram356, line 827: 🤔 **thinking** — After a claim, a *second* publisher `defineSlot` for the same live div lands here and returns the existing slot with no suppression re-arm and no warning. Native GPT returns `null` and logs "slot already exists", so publisher code relying on that null contract to detect double definitions now behaves differently. Falling through to native `defineSlot` when `handoff.publisherClaimed` is already true would preserve native semantics — was returning the slot for repeat definitions intentional?

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -449,9 +455,190 @@ function installInitialLoadDetector(ts: TsjsApi): void {
...
+      const originalDefineSlot = defineSlot.bind(g);
+      const patchedDefineSlot = (
+        adUnitPath: string,
+        formats: Array<number | number[]>,
+        elementId: string
+      ): GoogleTagSlot | null => {
+        const handoff = matchingHandoff(ts, pubads, adUnitPath, formats, elementId);
+        if (!ts.gptSlotHandoffInternal && handoff) {
+          const existingSlot = findGptSlotByElementId(pubads, handoff.slotElementId);
+          if (existingSlot) {
```
aram356: 🤔 **thinking** — This one-shot suppression never expires. If the claiming publisher never calls `display()` (some disabled-initial-load flows only `refresh()`), the pending suppression lingers and swallows a much later legitimate `display()` for this div. Impact is low today (TS already displayed/registered the slot), but worth a code comment documenting the assumed `defineSlot → display` sequence, matching the spec's risk table.

aram356 (changes requested):
## Summary

The inner-div fallback plus narrowly scoped `defineSlot`/`display`/`refresh` wrappers with a shared `window.tsjs` registry is a clean solution to the competing container-slot problem, and the test suite is thorough (options preservation, global-refresh filtering, ambiguity rejection, bootstrap-eval parity). However, the wrappers introduce two reproducible crash paths — one of which breaks publisher `defineSlot` calls for slots unrelated to TS placements — so this needs another pass before merge. Details inline.

## Non-blocking

### 🌱 seedling

- **Handoff registry is never pruned**: `gptSlotHandoffs` entries and hydration aliases accumulate for the page lifetime across SPA navigations; correctness is protected only by the live-slot lookup (`findGptSlotByElementId`) on every claim. Consider deleting entries when `adInit()` destroys `prevGptSlots`, as future-proofing against a matching path that forgets the liveness check.

## CI Status

No GitHub checks have run on this branch. Verified locally:

- JS tests (`npx vitest run`): PASS (418/418)
- Core GPT Rust tests (`cargo test -p trusted-server-core --target aarch64-apple-darwin integrations::gpt`): PASS (31/31)
- Prettier on gated `lib/` files: PASS

Both inline crash findings were reproduced with Vitest against the built module.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
ChristianPavilonis: Fixed in 0fdd13e7. The bootstrap now mirrors the bundle: it skips matching for internal calls and only matches a string div ID, forwarding the native two-argument form unchanged. Added bootstrap-eval coverage.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Fixed in 0fdd13e7. The bundle now duck-types slot targets before falling back to an element-like id, matching the bootstrap protocol. Added a friendly-iframe/cross-realm element regression that verifies one-shot display suppression without throwing.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Fixed in 0fdd13e7. Claimed exact handoffs now fall through to native defineSlot rather than returning the shared slot again. Bundle and bootstrap lifecycle tests now verify the first definition reuses the TS slot and the second receives the fake native GPT null duplicate result.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Addressed in 0fdd13e7 with an explicit comment in both implementations: suppression intentionally waits for the supported defineSlot → addService → display lifecycle and does not use a time heuristic. I also added SPA cleanup that prunes destroyed TS-owned handoffs and aliases.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Fixed in 0fdd13e7. The bundle now bypasses handoff matching for TS-internal and div-less defineSlot calls, preserving the native two-argument form. Added bundle and bootstrap-eval regressions for an unrelated div-less publisher definition with an unclaimed handoff.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -449,9 +455,196 @@ function installInitialLoadDetector(ts: TsjsApi): void {
...
+  elementId: string
+): GptSlotHandoff | undefined {
+  const exact = ts.gptSlotHandoffs?.[elementId];
+  if (exact) return exact.publisherClaimed ? undefined : exact;
+
+  const candidates = new Set(Object.values(ts.gptSlotHandoffs ?? {})).values();
+  const matching = Array.from(candidates).filter(
+    (handoff) =>
+      !handoff.publisherClaimed &&
+      elementId.startsWith(handoff.divIdPrefix) &&
```
aram356, line 759: 🔧 **wrench** — The prefix path can hijack a sibling placement's slot. It only checks `startsWith(divIdPrefix)` + same GAM path + same formats + live slot, and two sibling placements sharing a unit path and format are common (e.g. two 300×250 MPUs at `div-mpu` and `div-mpu-2`). With TS configured for `div-mpu` only, a publisher late-defining `div-mpu-2` gets handed the TS slot bound to `div-mpu`, and their `display('div-mpu-2')` is suppressed — the second placement silently renders blank. If they then define `div-mpu` itself, the exact path sees a claimed handoff, falls through to native `defineSlot` on an occupied div, gets `null`, and their `.addService(...)` throws.

The legitimate aliasing case is a hydration *rename*, in which the original element ID no longer exists in the DOM. Require that as evidence before prefix-aliasing:

```ts
(handoff) =>
  !handoff.publisherClaimed &&
  !document.getElementById(handoff.slotElementId) &&
  elementId.startsWith(handoff.divIdPrefix) &&
  ...
```

This is compatible with the new hydration tests (they rename the SSR div before the publisher defines, so the old ID is absent). Needs the same guard in `gpt_bootstrap.js`.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -42,6 +42,168 @@
...
+  function matchingHandoff(pubads, adUnitPath, formats, elementId) {
+    var exact = ts.gptSlotHandoffs && ts.gptSlotHandoffs[elementId];
+    if (exact) return exact.publisherClaimed ? null : exact;
+
+    var candidates = Object.values(ts.gptSlotHandoffs || {}).filter(
+      function (handoff, index, allHandoffs) {
+        return (
+          allHandoffs.indexOf(handoff) === index &&
+          !handoff.publisherClaimed &&
+          elementId.startsWith(handoff.divIdPrefix) &&
```
aram356, line 148: 🔧 **wrench** — Same sibling-hijack issue as the bundle's `matchingHandoff` (see comment there): add `!document.getElementById(handoff.slotElementId) &&` before the prefix check so only a hydration-renamed div (original ID gone from the DOM) can be aliased.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -449,9 +455,196 @@ function installInitialLoadDetector(ts: TsjsApi): void {
...
+          const handoff = matchingHandoff(ts, pubads, adUnitPath, formats, elementId);
+          if (handoff) {
+            const existingSlot = findGptSlotByElementId(pubads, handoff.slotElementId);
+            if (existingSlot) {
+              registerHandoffAlias(ts, elementId, handoff);
+              handoff.publisherClaimed = true;
+              // The supported publisher lifecycle is defineSlot → addService → display.
+              // Intentionally wait for that display instead of applying a time heuristic.
+              handoff.suppressPublisherDisplay = true;
+              handoff.suppressPublisherRefresh = ts.gptInitialLoadDisabled === true;
```
aram356, line 819: 🤔 **thinking** — The one-shot suppression can swallow a genuinely-later publisher refresh: under `disableInitialLoad`, the claimed slot's first publisher refresh is dropped even if it arrives minutes later with real refresh intent (viewability-triggered, updated targeting, new correlator). The spec accepts this tradeoff — noting it so the limitation stays a conscious choice. No action requested.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -449,9 +455,196 @@ function installInitialLoadDetector(ts: TsjsApi): void {
...
+  const exact = ts.gptSlotHandoffs?.[elementId];
+  if (exact) return exact.publisherClaimed ? undefined : exact;
+
+  const candidates = new Set(Object.values(ts.gptSlotHandoffs ?? {})).values();
+  const matching = Array.from(candidates).filter(
+    (handoff) =>
+      !handoff.publisherClaimed &&
+      elementId.startsWith(handoff.divIdPrefix) &&
+      handoff.gamUnitPath === adUnitPath &&
+      JSON.stringify(handoff.formats) === JSON.stringify(formats) &&
```
aram356: 🤔 **thinking** — GPT accepts a single size as `[300, 250]` (un-nested) and named sizes like `'fluid'`. A publisher using the shorthand against TS's `[[300, 250]]` fails this strict JSON equality, skips the handoff, and reintroduces the duplicate slot this PR aims to prevent. Consider normalizing (wrap a single `[w, h]` into a nested array) before comparing, in both implementations.

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
```diff
@@ -185,7 +190,684 @@ describe('installTsAdInit', () => {
...
+    installTsAdInit();
+
+    expect(() =>
+      (googletag.display as unknown as (target: Element) => void)(crossRealmElement)
+    ).not.toThrow();
+    expect(nativeDisplay).not.toHaveBeenCalled();
+    iframe.remove();
+  });
+
+  it('runs the embedded bootstrap handoff for a hydrated publisher ID', async () => {
```
aram356, line 932: ♻️ **refactor** — The spec calls for the same black-box cases on both implementations, but the eval'd-bootstrap suite covers only the hydrated handoff and the div-less define. The disabled-load global-refresh filter, refresh-options preservation, and ambiguity cases run only against the bundle. Since the bootstrap is a hand-maintained duplicate, extending the bootstrap suite with those three cases would catch future divergence.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -42,6 +42,168 @@
...
+    window.googletag.cmd.push(function () {
+      var tag = window.googletag;
+      var pubads = tag.pubads && tag.pubads();
+      if (!tag.defineSlot || !tag.display || !pubads) return;
+
+      if (!tag.defineSlot.__tsSlotHandoffPatched) {
+        var originalDefineSlot = tag.defineSlot.bind(tag);
+        var patchedDefineSlot = function (adUnitPath, formats, elementId) {
+          if (!ts.gptSlotHandoffInternal && typeof elementId === "string") {
+            var handoff = matchingHandoff(pubads, adUnitPath, formats, elementId);
```
aram356: ⛏ **nitpick** — `main`'s copy of this file passes `prettier --check`; this version has two over-width lines (here and ~112). The file sits outside the CI format glob (which covers only `lib/**`), so nothing fails, but the next `prettier --write` on it will generate noise. Trivial fix: run prettier on this file.

aram356 (changes requested):
## Summary

Solid, conservative design for the late-publisher slot handoff: single-candidate ambiguity guard, internal-call guard, idempotence markers shared between bootstrap and bundle, and ownership transfer out of `destroySlots()`. One blocking correctness issue in the prefix-matching path, plus five non-blocking items.

## Blocking

### 🔧 wrench

- **Prefix handoff can hijack a sibling placement's slot**: the prefix path can alias a publisher `defineSlot()` for a *different* sibling div to the TS slot and suppress its `display()`, leaving that placement blank (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts:495`, `crates/trusted-server-core/src/integrations/gpt_bootstrap.js:63`). See inline comments — a DOM-presence guard on `handoff.slotElementId` fixes it without breaking the new hydration tests.

## Non-blocking

### 🤔 thinking

- **One-shot `suppressPublisherRefresh` can swallow a genuinely-later refresh** (`index.ts:554`): documented spec tradeoff, no action requested.
- **`JSON.stringify` size comparison defeats valid GPT shorthand** (`index.ts:497`): a publisher passing `[300, 250]` un-nested skips the handoff and reintroduces the duplicate slot; consider normalizing before comparing.

### ♻️ refactor

- **Bootstrap parity gap in tests** (`ad_init.test.ts:589`): the eval'd-bootstrap suite covers only the hydrated handoff and div-less define; the disabled-load global-refresh filter, refresh-options preservation, and ambiguity cases run only against the bundle.

### 📌 out of scope

- **`gptInitialLoadDisabled` recorded even when the call was too late to take effect** (`index.ts:451`, unchanged by this PR): GPT ignores `disableInitialLoad()` called after `enableServices()`, but the detector still records `true`. On exactly this PR's target pages (hydration-deferred GPT setup), TS's `adInit()` enables services first, so the publisher's later `disableInitialLoad()` is a GPT no-op — yet every subsequent SPA-navigation `adInit()` both `display()`s (which really requests) and `refresh()`es TS-owned slots: a double request, the bug class this PR fixes. Pre-existing detector behavior — worth a follow-up issue (record only when `!ts.servicesEnabled` at call time).

### ⛏ nitpick

- **`gpt_bootstrap.js` is no longer prettier-clean** (`gpt_bootstrap.js:105`): two over-width lines; outside the CI format glob but `main`'s copy is clean.

## CI Status

GitHub checks: only CodeQL Analyze ran — PASS; fmt/clippy/test workflows have not run on this PR.

Local verification:

- rust tests: PASS (`cargo test-fastly gpt`, 31/31 including the new bootstrap assertion test)
- js tests: PASS (`npx vitest run`, 477/477)
- clippy: not run (only Rust change is a test)

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Fixed in b200be53. Prefix aliases now require the original TS fallback div to be absent from the DOM, so a live sibling such as `div-mpu-2` delegates to native GPT rather than claiming/suppressing the TS slot. Added the bundle sibling-slot regression; the existing hydration-rename case continues to prove the intended alias path.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
ChristianPavilonis: Fixed in b200be53. The bootstrap now applies the same original-div DOM-absence guard as the bundle, with bootstrap-eval coverage for the sibling-slot case.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Acknowledged. This remains the documented, accepted one-shot suppression tradeoff; no behavioral change was made for this non-actionable note.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Fixed in b200be53. Both implementations normalize a numeric singleton size (`[300, 250]`) before handoff comparison, and bundle/bootstrap-eval regressions verify reuse and display suppression. TS handoff configuration is numeric-banner-only, so named sizes such as `fluid` have no equivalent TS fallback format to match.

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
ChristianPavilonis: Fixed in b200be53. The bootstrap-eval suite now covers global refresh filtering, refresh-options preservation, and ambiguous prefix rejection, in addition to the existing hydrated and div-less handoff cases.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
ChristianPavilonis: Fixed in b200be53. Reformatted `gpt_bootstrap.js`; the bootstrap file now passes Prettier formatting checks.

prk-Jr (approved):

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -449,9 +455,205 @@ function installInitialLoadDetector(ts: TsjsApi): void {
...
+  elementId: string
+): GptSlotHandoff | undefined {
+  const exact = ts.gptSlotHandoffs?.[elementId];
+  if (exact) return exact.publisherClaimed ? undefined : exact;
+
+  const candidates = new Set(Object.values(ts.gptSlotHandoffs ?? {})).values();
+  const matching = Array.from(candidates).filter(
+    (handoff) =>
+      !handoff.publisherClaimed &&
+      !document.getElementById(handoff.slotElementId) &&
```
aram356, line 758: 🔧 **wrench** — This guard treats "original div gone" as proof of a hydration rename, but SPA navigation also removes divs, and that lets a cross-route publisher definition claim a dead slot (reproduced with a Vitest against this module):

1. Route A: TS creates its fallback slot on the resolved div.
2. Navigation to route B removes that div. The route-B `adInit()` — which destroys `prevGptSlots` and prunes handoffs — only runs after the page-bids fetch resolves (a network round trip).
3. In that window the publisher's framework defines route B's placement (`div-atf-sidebar-2`, or the next generated `ad-header-0-_r_N_`) with the same GAM path/formats. Every prefix-path condition passes: unclaimed, original div absent, prefix match, path/formats match, slot still live.
4. The publisher receives route A's slot bound to a removed div, its first `display()` is suppressed, and the claim filters the slot out of `prevGptSlots` — so route B's `adInit()` never destroys it. The publisher's placement is permanently bound to a nonexistent element and the impression is silently dropped.

This hits the primary use case — prefix-configured placements with framework-generated suffixes — on any SPA navigation where the publisher defines before TS's fetch returns, which is the normal ordering (frameworks define on mount; TS waits on the network).

**Fix**: invalidate unclaimed handoffs at navigation start — in `onNavigate()`, synchronously before the fetch, delete every handoff with `publisherClaimed === false`. The registry lives on `window.tsjs`, so the bootstrap wrappers observe the same cleanup. The hydration-rename case is unaffected (renames happen without navigation), and route B's TS placement gets a fresh entry from its own `adInit()`. Please add a regression test for the cross-route sequence:

```ts
// route A adInit creates fallback on div-atf-sidebar; then:
routeADiv.remove();
const routeBDiv = document.createElement('div');
routeBDiv.id = 'div-atf-sidebar-2';
document.body.appendChild(routeBDiv);
const publisherSlot = googletag.defineSlot('/123/sidebar', [[300, 250]], 'div-atf-sidebar-2');
googletag.display('div-atf-sidebar-2');
// must be a fresh native slot, and the display must go through:
expect(publisherSlot.getSlotElementId()).toBe('div-atf-sidebar-2');
```

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -42,6 +42,194 @@
...
+  function matchingHandoff(pubads, adUnitPath, formats, elementId) {
+    var exact = ts.gptSlotHandoffs && ts.gptSlotHandoffs[elementId];
+    if (exact) return exact.publisherClaimed ? null : exact;
+
+    var candidates = Object.values(ts.gptSlotHandoffs || {}).filter(
+      function (handoff, index, allHandoffs) {
+        return (
+          allHandoffs.indexOf(handoff) === index &&
+          !handoff.publisherClaimed &&
+          !document.getElementById(handoff.slotElementId) &&
```
aram356, line 147: 🔧 **wrench** — Same cross-route mis-claim as the bundle's `matchingHandoff` (see the companion comment on `index.ts`): after an SPA navigation removes the original div, a same-prefix/path/formats publisher definition for the next route claims the not-yet-destroyed route-A slot. The registry cleanup proposed for the bundle's `onNavigate()` covers this wrapper too, since both read the shared `window.tsjs.gptSlotHandoffs` — no bootstrap-side change should be needed beyond the shared-state fix, but please extend the bootstrap-eval suite with the cross-route case to prove it.

aram356 (changes requested):
## Summary

Re-review at `b200be53c`. All findings from the previous rounds are verified fixed with regression coverage (div-less `defineSlot` guard with native two-argument passthrough, duck-typed display targets, native null contract for repeat definitions of a claimed div, SPA handoff pruning, sibling-slot DOM-absence guard, shorthand size normalization, bootstrap-eval parity coverage, Prettier-clean bootstrap). One new blocking issue remains: the DOM-absence guard added for the sibling case does not distinguish a hydration rename from an SPA navigation, which lets a cross-route publisher definition claim a dead slot — details inline, reproduced with a Vitest against the module.

## CI Status

Only CodeQL ran on GitHub (pass). Verified locally at this head:

- JS tests (`npx vitest run`): PASS (484/484)
- Core GPT Rust tests (`cargo test -p trusted-server-core --target aarch64-apple-darwin integrations::gpt`): PASS (31/31)
- Prettier on `gpt_bootstrap.js` and the changed `lib/` files: PASS

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Implemented in 72ff8cbf2. Accepted SPA navigations now synchronously remove every unclaimed GPT handoff entry, including hydration aliases, before the page-bids fetch begins; claimed publisher handoffs are preserved. Added a regression covering the cross-route sequence with the bundle implementation.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
ChristianPavilonis: Implemented in 72ff8cbf2. The bootstrap wrapper reads the shared window.tsjs.gptSlotHandoffs registry, so it observes the same synchronous navigation cleanup without a separate production-code change. Added the matching cross-route regression against the evald bootstrap implementation.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -614,7 +949,7 @@ export function installTsAdInit(): void {
       // called without a matching display call") and misses its impression.
       // Must run after enableServices(); on SPA navigation services are already
       // enabled, so this runs unconditionally for any newly-defined slots.
-      slotsToDisplay.forEach((divId) => g.display?.(divId));
+      slotsToDisplay.forEach((divId) => withGptSlotHandoffInternal(ts, () => g.display?.(divId)));
```
aram356, line 1139: 🔧 **wrench** — Just below this line, the self-merge dropped main's `syncInitialLoadDisabled(g, ts);` re-sync that sat immediately above the `slotsNeedingRefresh` computation (main's `index.ts:742`, now `index.ts:963`). The helper and its other four call sites survived the merge — only this call was lost, here and in the bootstrap mirror (companion comment).

Impact: on pages using the modern `googletag.setConfig({ disableInitialLoad: true })` API, if no queue-drain has re-synced the flag before `adInit()` reaches this decision, `ts.gptInitialLoadDisabled` is stale-false, TS-owned slots are only `display()`ed (registration, no request) and never refreshed — blank placements. This is exactly what the four failing `ad_init.test.ts` setConfig tests assert (`tracks the effective initial-load state from setConfig`, `reads initial-load configuration effective before detector installation`, `preserves legacy state in the edge bootstrap when getConfig does not report it`, `tracks setConfig state and re-enabling in the edge bootstrap`).

**Fix**: restore the one-line call directly above the `slotsNeedingRefresh` computation:

```ts
syncInitialLoadDisabled(g, ts);
const slotsNeedingRefresh = ts.gptInitialLoadDisabled
  ? slotsToRefresh.concat(newSlots)
  : slotsToRefresh;
```

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -143,12 +416,14 @@
...
+          googletag.display(divId);
+        });
       });
       // Reused publisher-owned slots always need a refresh to pick up the
       // server-side targeting. TS-defined slots are fetched by display() above
       // unless the publisher disabled initial load, in which case display() only
-      // registers them and refresh() must request the ad — otherwise they render
+      // registers the slot and refresh() must request the ad — otherwise they render
       // blank. Only add them in that case to avoid double-requesting.
       var slotsNeedingRefresh = ts.gptInitialLoadDisabled
```
aram356, line 534: 🔧 **wrench** — Same merge loss as the bundle (see companion comment on `index.ts`): main's `syncInitialLoadDisabled(window.googletag);` call directly above this `slotsNeedingRefresh` computation (main's `gpt_bootstrap.js:227`) was dropped in the conflict resolution while the helper and its other call sites survived. Without it, a `setConfig({ disableInitialLoad: true })` page renders TS-owned bootstrap slots blank. Restore:

```js
syncInitialLoadDisabled(window.googletag);
var slotsNeedingRefresh = ts.gptInitialLoadDisabled
  ? slotsToRefresh.concat(newSlots)
  : slotsToRefresh;
```

aram356 (changes requested):
## Summary

Round-3 review at `72ff8cbf2`. The cross-route mis-claim from the previous round is properly fixed: `onNavigate()` now synchronously deletes unclaimed handoffs (aliases included) before the page-bids fetch, with bundle and bootstrap regression coverage — I re-ran the previous round's reproduction against this head and it passes. However, the branch is red: vitest fails 6 tests in CI and locally. The head is a self-merge (`Merge origin/fix/duplicate-gpt-slots into fix/duplicate-gpt-slots`) of the old branch line into a rebased duplicate, and the conflict resolution dropped newer main-side GPT work — one inline finding per implementation, plus the test-reconciliation finding below (that file is outside this PR's diff, so it cannot carry an inline comment).

Content-wise the diff against main is still exactly the 8 PR files (GitHub's live compare confirms; the PR page's "106 files changed" stat is stale from before main advanced and should settle).

## Blocking

### 🔧 wrench

- **main's new `gpt_bootstrap.test.ts` contract tests are not reconciled with the handoff wrappers** (2 of the 6 failures; `crates/trusted-server-js/lib/test/integrations/gpt/gpt_bootstrap.test.ts:197` and `:244`). Two distinct problems: (1) the tests assert on `googletag.defineSlot` / `mockPubads.refresh` after the bootstrap has replaced them with `patchedDefineSlot`/`patchedRefresh`, so vitest throws `[Function patchedDefineSlot] is not a spy` — capture the native spy references before evaluating the bootstrap, or assert on the local spy variables; (2) the first test still encodes the pre-PR container behavior, expecting `defineSlot('/123/atf', [[300, 250]], 'div-atf-sidebar-container')` — this PR deliberately defines the TS fallback on the inner div, so the expected element ID must become `'div-atf-sidebar'`. That changed expectation is part of this PR's behavioral contract and belongs in this diff.

## Non-blocking

### 📝 note

- **Branch topology**: the head merges a rebased duplicate of this branch with its old self — 16 commits, two parallel lines with identical commit messages (`3637ebb0e..41ea00232` vs `76c56263f..b200be53c`). The net diff is correct, but consider collapsing to a single line before merge (or plan on a squash-merge) so the history doesn't permanently carry both copies of every commit.

## CI Status

- vitest (GitHub CI): FAIL — 6 tests, reproduced locally (570/576 pass)
- Core GPT Rust tests (`cargo test -p trusted-server-core --target aarch64-apple-darwin integrations::gpt`): PASS locally (39/39)
- Prettier on the five touched JS/TS files: PASS locally

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Fixed in 44f0d101. Restored syncInitialLoadDisabled(g, ts) immediately before the slotsNeedingRefresh decision, and updated the bundle tests to assert against the native refresh spy. Targeted GPT tests pass (67/67), the full Vitest suite passes (576/576), and Fastly GPT Rust tests pass (40/40).

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -867,16 +1175,30 @@ function waitForSlotElements(slots: AuctionSlot[], signal: AbortSignal): Promise
 
   return new Promise<void>((resolve) => {
```
aram356, line 1286: 🔧 **wrench** — `allPresent` (just above, line 1171) now runs through the tiered resolver, which returns `null` for a unique prefix match that is `display:none`. A breakpoint-hidden placement whose div exists in the DOM (a mobile-only slot config while on desktop, a consent-gated section) can never "appear", so **every SPA navigation to such a route waits out the full 2s `SPA_SLOT_WAIT_MS` before applying bids to every slot on the route** — visible ones included — and adInit then skips the hidden slot anyway. Reproduced with a Vitest against this module: with one `display:none` sibling placement, `adInit` fires 0 times before the timeout and once after ~2.1s. Pre-#978, the first prefix match resolved immediately and there was no stall — this is a systematic 2-second render delay for affected routes.

**Fix**: for the wait (not for adInit), presence and eligibility are different questions — count a slot as present when its prefix matches any element:

```ts
const allPresent = (): boolean =>
  slots.every((slot) => {
    const resolution = resolveSlotElementByDivId(slot.div_id);
    return resolution.element !== null || resolution.prefixMatchCount > 0;
  });
```

adInit still applies the strict tiers when it runs; the wait just stops stalling on placements that exist but are ineligible. Please add a regression test: route with one visible and one `display:none` prefix-configured placement must apply bids without waiting for the timeout.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -619,11 +907,23 @@ export function installTsAdInit(): void {
...
-        // the suffix is dynamically generated by the framework at render time.
-        const el = findSlotElementByDivId(slot.div_id);
-        if (!el) return;
+        // Resolve actual div ID: exact match first, then the visibility and
+        // geometry tiers for prefix matches. div_id in config may be a stable
+        // prefix (e.g. "ad-header-0-") when the suffix is dynamically
+        // generated by the framework at render time.
+        const resolution = resolveSlotElementByDivId(slot.div_id);
+        const el = resolution.element;
+        if (!el) {
```
aram356, line 990: 🤔 **thinking** — This warning only fires when `prefixMatchCount > 1`. The single-hidden-match case (`prefixMatchCount === 1, activeMatchCount === 0` — the common breakpoint-hidden config) is skipped with no log at all, and the exact-id tier above skips visibility checks entirely, so a hidden exact-id match still gets a slot while a hidden prefix match silently does not. Given how much production time goes into diagnosing blank placements, consider logging the `prefixMatchCount >= 1 && activeMatchCount === 0` case too (debug level is fine), and a short comment documenting the exact-vs-prefix visibility asymmetry as intentional.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -133,25 +401,28 @@
...
-            ) {
-              el = candidate;
-              break;
+          if (
+            resolution.prefixMatchCount > 1 &&
+            !warnedResolutionFailures[slot.div_id]
+          ) {
+            warnedResolutionFailures[slot.div_id] = true;
+            if (ts.log && typeof ts.log.warn === "function") {
+              ts.log.warn("GPT slot prefix did not resolve to one active element", {
```
aram356: ⛏ **nitpick** — This `ts.log.warn(...)` call isn't Prettier-formatted (Prettier wants the argument list broken). The file isn't covered by the CI format gate, but it was made Prettier-clean earlier in this PR — one `npx prettier --write` away from staying that way.

aram356 (changes requested):
## Summary

Round-4 review at `bda19d304`. Both round-3 findings are verified fixed: `syncInitialLoadDisabled` is restored at the refresh decision in both implementations, and the bootstrap contract tests now assert on native spy references with the inner-div expectations this PR owns. CI is fully green (vitest included); locally the JS suite passes 611/611 and the core GPT Rust tests pass 39/39.

This round's findings are all in the newly stacked-in #978 responsive-slot resolution, which is now part of this PR's diff. The tiered visible/layout resolution design is sound and well-tested, but it introduces one reproducible SPA latency regression plus two smaller items — all inline.

## CI Status

- All GitHub checks: PASS (vitest, browser integration, parity, adapters, format, CodeQL)
- JS tests locally: PASS (611/611)
- Core GPT Rust tests locally: PASS (39/39)
- Prettier: one deviation in `gpt_bootstrap.js` (inline nitpick; file is not CI-gated)

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
aram356: Fixed in 8bfb44275. `allPresent` now counts a slot as present when its prefix matches any element (`resolution.element !== null || resolution.prefixMatchCount > 0`), so a breakpoint-hidden placement no longer stalls the route for the full timeout; adInit still applies the strict tiers when it runs. Added the requested regression test (one visible + one `display:none` prefix placement applies bids without waiting) and verified it fails against the previous `allPresent`.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
aram356: Added in 8bfb44275. The `prefixMatchCount === 1 && activeMatchCount === 0` case now logs at debug level in both the bundle and the bootstrap, and the resolver documents the exact-vs-prefix visibility asymmetry as intentional at the exact-id tier.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
aram356: Fixed in 8bfb44275; the block is Prettier-formatted again.

aram356 (approved):

PR #969, opened 2026-07-27
Gate initial adInit on ad-slot container hydration

### LINKED ISSUE #958: Investigate refresh too quickly after SSAT with client side auction (timing issue with GAM)
### Description

Investigate refresh too quickly after SSAT with client side auction (timing issue with GAM)

prk-Jr:
## Summary

Follow-up to #945, which defers the initial `adInit()` past React hydration to fix a
#418 hydration mismatch on Next.js App Router, gating on `window.load`. `window.load`
waits for every image and tracker, so on a heavy publisher it fires **~52s** after
navigation and the Trusted Server slots sit empty until then.

An earlier revision of this branch replaced that gate with the Next.js App Router
runtime signal (`window.__next_f.push` being patched). **Measurement has since shown
that gate is unsafe, and it has been replaced.** This PR now gates on **React having
hydrated the ad-slot containers `adInit` is about to mutate**, with `window.load`
retained as the unconditional fallback.

## Why the runtime signal was dropped

A controlled capture against a live App Router article page, varying only what drives
`adInit` and holding the page, proxy and build identical:

| arm | #418 | SSAT ad destroyed | slots rendered |
| --- | --- | --- | --- |
| Trusted Server off (publisher alone) | 0/8 | — | — |
| TS on, `adInit` suppressed | 0/8 | — | — |
| TS on, `window.load` gate (#945) | 0/6 | 0/6 | 5/6 |
| TS on, `__next_f` gate (previous revision here) | **8/8** | **3/6** | 6/6 |
| TS on, container-hydration gate (**this PR**) | **0/8** | **0/8** | **8/8** |

Three things this establishes that the earlier evidence on this PR did not:

1. **`adInit` is the sole cause of the #418.** The publisher throws none on its own,
   and TS's `<head>` / `</body>` script injection throws none — only the `adInit` DOM
   mutation does.
2. **`__next_f` patched means hydration *started*.** It lands a median 16ms after
   React's *first* commit and on some runs *before* it, while hydration continues for
   seconds (830–21,876 commits observed per run). The double `requestAnimationFrame`
   is a ~32ms margin against a multi-second tail.
3. **The #418 costs impressions.** React regenerates the mismatched subtree, so the
   creative that just rendered is destroyed — on 3/6 runs the `in_content` iframe
   appeared and was then removed ~80ms after the error.

The earlier "#418 ≈5%, not impression-costing" reading came from pages whose ad slots
are not React-owned, and had no control arm. On an article page, whose containers are
React-rendered with `useId` ids (`ad-in_content-<hash>-in_content-0-container`), the
`__next_f` gate reproduces at 100%.

## Change

- `crates/trusted-server-js/lib/src/integrations/gpt/index.ts` — the gate polls for
  every currently-matched slot container reporting React-hydrated, via the
  `__reactFiber$` / `__reactProps$` own property React attaches to host nodes it owns.
  New `isReactHydrated` helper. `window.load` listener, `fired` guard + `clearInterval`
  exactly-once semantics, synchronous pre-check, `navGeneration` cancellation and the
  double `requestAnimationFrame` are all unchanged. The pass stays a single `adInit`
  call so GPT still issues one batched ad request.
- `crates/trusted-server-js/lib/test/integrations/gpt/schedule_initial_ad_init.test.ts`
  — the five signal-specific tests moved onto the container signal, plus a new case
  where one container hydrates and another never does (the batch must wait for the
  laggard, then fall back to `load`).
- `crates/trusted-server-core/src/publisher.rs`, `crates/trusted-server-js/lib/src/core/types.ts`
  — **comments only**, describing the gate that is actually shipped.

## Why containers and not a page-global signal

`adInit` defines GPT slots on the publisher's ad-slot subtrees, so what matters is
whether React has hydrated *those elements* — not whether the framework runtime booted
or every subresource finished. This introduces no timing constant: each page gates on
its own DOM. If React renames the internals, nothing reports hydrated and the gate
degrades to the `load` path, so the failure mode stays slow rather than broken. Non-React
publishers take the `load` path for the same reason, exactly as before.

## Latency — measured and unmeasured

Measured locally through `ts dev proxy`, the container gate fires at ~11.85s versus
~11.80s for `window.load` — **indistinguishable**, because through the proxy `load`
already fires at roughly the moment the containers hydrate, so there is no gap to win.
SSAT still beat the publisher's own client-side GAM request by ~2.2s.

**The latency win this branch exists for is therefore unverified.** It only appears
where `load` is dominated by images and trackers — ~52s on the deployed publisher
against ~9s to hydrate. The container gate should fire at hydration time regardless of
what `load` is waiting for, but that is reasoning, not measurement, and reasoning is
what put the previous revision here at 8/8 #418.

## Note on the previous acceptance bar

The prior test plan asked to "confirm #418 = 0". That zero was measured on the *light*
publisher #945 tested; the `window.load` gate's #418 rate on a heavy publisher had never
been measured. In the capture above it is 0/6 for `load` and 0/8 for this gate, so the
bar is now both measured and met on this page.

## Test plan

- [x] `cd crates/trusted-server-js/lib && npx vitest run` — 562/562; scheduler suite 16/16
- [x] `cd crates/trusted-server-js/lib && npm run format` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy-fastly` — clean
- [x] Live capture, 8 loads of a live App Router article page through `ts dev proxy`:
      0/8 #418, 0/8 destroyed creatives, 8/8 slots rendered, `adInit` a median 1096ms
      after the hydration commit (min 983ms, never before)
- [ ] **Deploy A/B for the latency claim** — confirm the gate fires at hydration
      (~seconds) rather than at `window.load` (~52s) on the deployed heavy publisher,
      with `ts_initial=1` on the first GPT request and #418 still 0
- [ ] Remaining CI gates: `test-axum` / `test-cloudflare` / `test-spin`, adapter clippy
      variants, parity suite

## Follow-ups (out of scope here)

- TS fires SSAT requests at off-breakpoint duplicate `in_content` containers
  (`display:none`) — 19 present on the tested article page. Wasted GAM traffic,
  independent of the gate.

Refs #958, #938. Builds on #945.


crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -487,9 +488,32 @@ function installScheduleInitialAdInit(ts: TsjsApi): void {
...
+    let fired = false;
+    let poll: ReturnType<typeof setInterval> | undefined;
+    const fire = (): void => {
+      if (fired) return;
+      fired = true;
+      if (poll !== undefined) clearInterval(poll);
+      afterHydrationFrames();
+    };
+    window.addEventListener('load', fire, { once: true });
+    const nextRuntimeReady = (): boolean => {
```
ChristianPavilonis: 🔧 **`__next_f.push` is replaced before hydration commits—and can be replaced before hydration even starts**

`nextRuntimeReady()` treats `window.__next_f.push !== Array.prototype.push` as readiness, then waits two animation frames. Next.js replaces `__next_f.push` while initializing its Flight-data consumer, not after React commits. Current Next.js source patches `push` during module initialization, while `hydrate()` can subsequently await the initial RSC payload before calling `hydrateRoot`. Even after `hydrateRoot` starts, concurrent/selective hydration may span additional frames; double `requestAnimationFrame` has no synchronization relationship with React's commit.

This matters because `adInit()` mutates React-owned ad-slot subtrees. On a slow CPU, delayed RSC stream, or Suspense/selective-hydration path, it can run while hydration is pending and reintroduce React #418, duplicated/replaced DOM, or missing ad slots. The tests prove that the heuristic fires, not that it fires after commit.

Please avoid using `__next_f.push` as a generic post-hydration signal. Retain the safe `load` path unless the publisher provides an explicit post-commit event, or make `adInit()` safe to run without mutating React-owned DOM. If this heuristic remains experimental, it should at least be publisher-opt-in and covered by a browser test that delays hydration and proves `adInit` occurs after a commit/effect marker.

docs/superpowers/plans/2026-07-25-adinit-hydration-gate.md
```diff
@@ -0,0 +1,318 @@
+# adInit Hydration-Chunk Gate Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make the initial `window.tsjs.adInit()` fire as soon as the Next.js App Router hydration chunks have executed (a few seconds) instead of on `window.load` (~52s on heavy publishers), without reintroducing the React #418 hydration mismatch.
```
ChristianPavilonis, line 5: 🔧 **The committed design describes the superseded implementation and wrong validation path**

The plan's goal, architecture, file list, tasks, and validation still describe a Rust-emitted hydration-chunk gate and state that no JS build is needed. The actual change is in the TypeScript GPT scheduler and requires Vitest/JS validation. The specification likewise retains the old chunk implementation as its primary Approach and Testing sections, while `TsjsApi.scheduleInitialAdInit` still documents only the `window.load` behavior.

This directs reviewers and future maintainers toward the wrong code, tests, safety argument, and rollout behavior. Please rewrite the active design around `installScheduleInitialAdInit` and the `__next_f` heuristic, move the rejected chunk experiment into a concise history/rejected-alternative section, and update the API comment, test commands, file list, PR title, and PR body to match the final diff.

ChristianPavilonis (approved):
## Summary

Reviewed the scheduler change against `fix/react418-hydration-safety`. The integration checks reported on this stacked PR pass, but I found one high-risk hydration-safety concern and one documentation/validation discrepancy; both are called out inline and should be addressed before merge. The blocking live React #418 A/B and the full JS test/format gates are still outstanding.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -487,9 +488,32 @@ function installScheduleInitialAdInit(ts: TsjsApi): void {
     };
     if (document.readyState === 'complete') {
       afterHydrationFrames();
-    } else {
-      window.addEventListener('load', afterHydrationFrames, { once: true });
+      return;
     }
+    // Fire on whichever hydration signal arrives first, exactly once:
```
aram356: 🤔 **thinking** — The spec's own "Why this gate" section rejects exactly this gate: `__next_f.push` being patched means the RSC runtime *started* hydrating, not that hydration *committed*, and the only margin here is two rAFs (~32 ms) — while React 18 hydrates concurrently and time-slices a heavy tree across many frames. So a mid-hydration fire is plausible on precisely the heavy publishers this PR targets.

The spec rejected `requestIdleCallback` alone because it fires before runtime boot (~3s). But the combination answers that objection: **signal → `requestIdleCallback` (timeout-bounded) → double rAF**. Idle-after-runtime-boot means the hydration work has drained, which is a much stronger post-commit proxy than two frames after boot.

The #418 A/B stays the arbiter either way, and the `window.load` fallback keeps the failure mode safe-but-slow — but worth considering before the A/B run rather than after a failed one.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -487,9 +488,32 @@ function installScheduleInitialAdInit(ts: TsjsApi): void {
...
+      fired = true;
+      if (poll !== undefined) clearInterval(poll);
+      afterHydrationFrames();
+    };
+    window.addEventListener('load', fire, { once: true });
+    const nextRuntimeReady = (): boolean => {
+      const flight = (window as unknown as { __next_f?: { push?: unknown } }).__next_f;
+      return !!flight && flight.push !== Array.prototype.push;
+    };
+    poll = setInterval(() => {
```
aram356: ♻️ **refactor** — This script runs at `</body>` on a *streamed* App Router page, so the async chunks can have executed while the body was still streaming — meaning the runtime may already be patched at schedule time. Today that path wastes an interval and up to 50 ms. Check once synchronously before installing the poll (guarded so the interval is never created after `fire()` has run):

```ts
if (nextRuntimeReady()) {
  fire();
  return;
}
poll = setInterval(() => {
  if (nextRuntimeReady()) fire();
}, 50);
```

(The `load` listener registered above is harmless to leave behind — `once: true` plus the `fired` guard.)

crates/trusted-server-js/lib/test/integrations/gpt/schedule_initial_ad_init.test.ts
```diff
@@ -244,4 +244,90 @@ describe('scheduleInitialAdInit', () => {
...
+      window.dispatchEvent(new Event('load'));
+      flushFrame();
+      flushFrame();
+      expect(adInit).toHaveBeenCalledTimes(1);
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
+  it('fires exactly once when the runtime signal and a later load event both occur', async () => {
```
aram356: ⛏ **nitpick** — This covers poll-fires-then-load. The reverse order — `load` fires first (non-Next-yet page), then `__next_f` gets patched while the poll is still conceptually alive — is logically covered by the `fired` guard and `clearInterval`, but untested. One small test (dispatch `load`, then advance timers with a patched `__next_f`, assert exactly one `adInit`) closes the signal-order matrix.

aram356 (changes requested):
## Summary

Replaces #945's `window.load` gate for the initial `adInit()` with a first-signal-wins hybrid: poll for the Next.js App Router runtime patching `window.__next_f.push`, with `window.load` retained as the can't-hang fallback. The gate logic itself is sound — exactly-once semantics (`fired` guard + `clearInterval` in both signal orders), `navGeneration` cancellation, and the `readyState === 'complete'` short-circuit all check out, and the new Vitest lifecycle tests are executable rather than string-matched. The blocking issues are descriptive, not functional: the PR title/body/docs describe two superseded implementations, and the #418 A/B this PR self-gates on is still outstanding.

## Blocking

### 🔧 wrench

- **PR title, body, and docs describe the superseded implementation.** The title says "hydration chunks"; the body says the change is `crates/trusted-server-core/src/publisher.rs::build_bids_script` and cites `bids_script_is_xss_safe` / `bids_script_defers_ad_init_until_after_hydration` — none of which is in this diff. The actual change is the `__next_f.push` poll in `scheduleInitialAdInit` (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts`) plus its Vitest suite; there are no Rust changes at all. The docs' UPDATE banners cover the chunk-await → `__next_f` pivot but not the `publisher.rs` → `gpt/index.ts` move, so both docs still point future readers at the wrong file (the plan calls `publisher.rs` "the only production change"). Please retitle (e.g. "…on the Next.js runtime signal…"), rewrite the Change/Test-plan sections, and add one line to each doc's banner noting the final implementation location.

### ❓ question

- **Is the #418 A/B still blocked, and will its evidence land on this PR?** The spec says the blocking A/B was stuck behind the rc/july `elementId.startsWith` crash (fix = adopt #966's revert in the deploy). Has that deploy happened, and will the A/B result (confirming #418 = 0 and the ~9s vs ~52s timing) be attached here before merge?

## Non-blocking

### 🤔 thinking

- **Runtime-boot signal can still fire mid-hydration** — see inline comment on `gpt/index.ts`; consider `requestIdleCallback` after the signal before the A/B run.

### ♻️ refactor

- **Check `nextRuntimeReady()` synchronously before installing the poll** — see inline comment on `gpt/index.ts`.

### 🌱 seedling

- **`__next_f` is a Next.js internal.** Worth documenting which Next versions the signal is verified against (App Router 13+); Pages Router publishers and any future Next version that renames the flight global silently take the `window.load` path. The failure mode is safe (slow, not broken), but a canary check per major Next upgrade of a gated publisher is warranted.

### 🏕 camp site

- **Stale gate description in `publisher.rs`.** The base's `build_bids_script` doc comment (around `crates/trusted-server-core/src/publisher.rs:3320`) still says the deferral is "gate on window `load`, then a double `requestAnimationFrame`". This PR changes that behavior; since the file isn't in this diff the comment can't take an inline note, but updating that cross-reference here would avoid shipping a wrong comment at merge.

### ⛏ nitpick

- **Untested signal order** — load first, runtime signal later; see inline comment on the test file.

## CI Status

- GitHub checks (browser integration, integration, EC lifecycle, artifacts): PASS
- JS tests (`npx vitest run`, full suite at PR head): PASS (435/435; new scheduler suite 9/9)
- JS + docs format (prettier): PASS
- fmt / clippy / Rust tests: not re-run — the net diff vs base contains no Rust changes.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
prk-Jr: Done in a7fa1d024 — the synchronous pre-check is in, in the guarded form you wrote:

```ts
if (nextRuntimeReady()) {
  fire();
  return;
}
poll = setInterval(() => {
  if (nextRuntimeReady()) fire();
}, 50);
```

`fire()` before `poll` is assigned is safe — it guards with `if (poll !== undefined) clearInterval(poll)` — and the `load` listener is left registered as you noted (`once: true` + the `fired` guard).

Covered by a new test, `fires without waiting a poll tick when the runtime is already patched at schedule time`, which asserts `vi.getTimerCount() === 0` after the schedule call, so the pre-check is observable rather than just present. Two existing tests that had patched `__next_f` *before* calling `scheduleInitialAdInit` now patch it after, so they still exercise the interval path instead of silently moving onto the new synchronous one.

crates/trusted-server-js/lib/test/integrations/gpt/schedule_initial_ad_init.test.ts
prk-Jr: Done in a7fa1d024 — `fires exactly once when a load event precedes the runtime signal`: dispatch `load`, flush both frames, assert one `adInit`, then patch `__next_f` and advance 500 ms of interval time, assert still one. That closes the matrix against the existing `fires exactly once when the runtime signal and a later load event both occur`.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
prk-Jr: Agreed on the mechanism: `__next_f.push` being patched means the RSC runtime *started*, and idle-after-runtime-boot is a strictly stronger post-commit proxy than two frames. The `requestIdleCallback`-alone objection in the spec (fires ~3s, before the runtime) genuinely does not apply to signal → rIC → double-rAF.

Not adopting it in this PR, for three reasons, and it is now recorded in the spec's Update section as considered-and-deferred rather than dropped:

1. **The timeout would be an invented constant.** Nothing measured here says when the tree finishes draining, so any value is a guess that delays *every* gated publisher, including the ones where two frames were already enough. This repo's standing rule is not to introduce magic numbers to satisfy review — ground them in a measurement or track them.
2. **The risk being bought down is measured small.** Live capture on the deployed build: #418 on 2/38 runs (~5%), and **0/38** correlation with slot loss — every #418 run still rendered header and fixed_bottom; the one header miss was a non-#418 run. So mid-hydration fire is real but currently not impression-costing, which does not justify unmeasured added latency ahead of the A/B.
3. **It widens the gate.** `requestIdleCallback` needs its own fallback branch (Safari only from 16.4), and the gate's whole value is that every failure mode degrades to "slow, not broken".

Sequencing, which I think matches your "before a failed A/B" point without pre-committing a constant: run the A/B on the current gate. If it shows #418 **with** slot loss, add the idle step and derive its timeout from the hydration tail that same A/B measures. The `window.load` fallback keeps the downside bounded either way.

docs/superpowers/plans/2026-07-25-adinit-hydration-gate.md
prk-Jr: Addressed in a7fa1d024 and 01e7621e1:

- **PR title and body** rewritten around `installScheduleInitialAdInit` and the `__next_f` signal. The old body cited `publisher.rs::build_bids_script` and Rust test names that are not in this PR's diff at all (they come from the base branch, #945). The Change section now states plainly that the behavior here is TypeScript-only, and the test plan lists the vitest / prettier / cargo commands actually run.
- **`TsjsApi.scheduleInitialAdInit`** doc comment (`src/core/types.ts`) now describes the first-signal-wins gate instead of `window.load`-only.
- **`build_bids_script`** doc comment in `publisher.rs` likewise — comment only, the emitted script is unchanged.
- **Both docs** carry an added banner paragraph recording the second pivot: the gate moved out of `publisher.rs` into `crates/trusted-server-js/lib/src/integrations/gpt/index.ts`, with coverage in `test/integrations/gpt/schedule_initial_ad_init.test.ts`, and the instruction to edit the TS module rather than the Rust `format!`. The plan's banner names the sections it supersedes ("Architecture", "File Structure") explicitly, since those still call `publisher.rs` the only production change.

One deliberate deviation: I did not restructure the two documents so the `__next_f` gate becomes the primary Approach with chunk-await demoted to an appendix. They are dated design/plan records (`2026-07-24`, `2026-07-25`) whose value is partly the trail — chunk-await was chosen for stated reasons, deployed, and disproven by live measurement, and rewriting them into a clean present-tense design erases the reason the next person should not retry it. Instead every superseded claim is fenced by a banner at the top that names its replacement and the file to edit. If you would rather they read as current design docs, say so and I will restructure — but I would want the rejected-alternative reasoning kept in full, not compressed.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
prk-Jr: Not disputing the mechanism: `__next_f.push` is patched during Flight-consumer initialization, `hydrateRoot` can follow later, concurrent/selective hydration can span frames, and a double `requestAnimationFrame` has no synchronization relationship with React's commit. The spec says the same in its Update section — the gate is a heuristic with a two-frame margin, not a commit proof — and that is why rollout is gated on the #418 A/B rather than on the argument.

What the measured position is, from the live capture on the deployed build:

- #418 occurred on **2/38 runs (~5%)**, both on one article page, 0 on homepage/mobile/news.
- **#418 → slot-loss correlation was 0/38**: every #418 run still rendered `header` and `fixed_bottom`; the single header miss was a non-#418 run.

So the failure you describe is real but, on this build, not impression-costing — which is the reason it is being shipped behind an A/B rather than blocked on a stronger signal that Next/React do not expose.

On the three specific asks:

1. **Retain the safe `load` path** — retained, unconditionally. It is still registered on every page, is the only signal on non-Next publishers (`__next_f` is never patched there, so the poll never fires), and wins the race whenever it arrives first. There is no path where the runtime signal replaces it; it only ever fires *earlier*.
2. **Publisher opt-in** — not adding a config flag in this PR. It would be a new knob with no operator-facing decision behind it, and the gate is already publisher-scoped by construction: non-App-Router publishers are unaffected because the signal does not exist for them. If the A/B shows harm, the correct response is reverting to the `load`-only gate from #945, which is one branch away, not carrying a permanent flag.
3. **Browser test that delays hydration and proves `adInit` runs after a commit/effect marker** — this is the right shape of proof and I cannot fake it in jsdom; the Vitest suite deliberately claims only lifecycle correctness (exactly-once across both signal orders, two-frame deferral, `navGeneration` cancellation), not post-commit ordering. The live #418 A/B is the substitute, and it is listed as blocking acceptance in the PR body. If the A/B comes back ambiguous, a real-browser hydration-delay test is the next step rather than a re-argued heuristic.

Also considered and deferred on the sibling thread: inserting a timeout-bounded `requestIdleCallback` between the signal and the frames, which is a stronger post-commit proxy. Deferred because its timeout would be an unmeasured constant; recorded in the spec with the condition that would adopt it.

prk-Jr:
Review items resolved in a7fa1d024 and 01e7621e1. Summary, including one item where I'm answering rather than changing code.

### Blocking 1 — title/body/docs described the superseded implementation

Fixed. Title is now "Gate initial adInit on the Next.js runtime signal instead of window load"; the body's Change and Test-plan sections are rewritten around `installScheduleInitialAdInit` and state plainly that the behavior here is TypeScript-only (the `publisher.rs::build_bids_script` work and the Rust test names the old body cited come from the base branch, #945, not this PR). Both docs' UPDATE banners now record the second pivot — `publisher.rs` → `crates/trusted-server-js/lib/src/integrations/gpt/index.ts` — naming the sections they supersede, since those still call `publisher.rs` the only production change. Details and one deliberate deviation (I kept the chunk-await design trail intact rather than restructuring the docs) in the thread.

### Blocking 2 — is the #418 A/B still blocked?

The A/B evidence exists; it just does not read as "#418 = 0", so stating it precisely.

The `elementId.startsWith` crash was fixed on the deployed branch by c1e963bf3 ("Guard configuredSlotForElementId against non-string display() args"), and the build deployed for the capture carried this same `__next_f` gate — at that point still emitted from `publisher.rs`, before #945 moved it into the GPT module. The gate logic under test was the same first-signal-wins poll; nothing since changes *when* it fires (the synchronous pre-check added today can only make it fire earlier or at the same time).

Live headless correlation capture on the real App Router publisher (tester cookie + DataDome, GAM SSAT reservation line `6843566157`):

| Result | Measurement |
| --- | --- |
| SSAT render+stay, `header` | 100% (90px, `ts_initial=1`) |
| SSAT render+stay, `fixed_bottom` | 100% (90px) |
| #418 occurrences | 2/38 runs (~5%), both on one article page; 0 on homepage/mobile/news |
| #418 → slot loss | **0/38** — every #418 run still rendered `header` and `fixed_bottom`; the single header miss was a non-#418 run |
| Local runs | 0/25 #418 |

So: **#418 is not zero, it is ~5% and costs no impressions on this build.** The timing side of the claim holds — the runtime signal lands ~9s versus ~40–52s for `window.load` on the same publisher.

I've left the A/B unchecked in the test plan and would rather it be signed off explicitly than have me mark it green on this, since the acceptance line as originally written asked for #418 = 0 and this is "#418 present but non-costing". If ~5% non-impression-costing #418 is acceptable, the blocking item is met and I'll update the checkbox; if the bar is literally zero, this gate does not clear it and the `requestIdleCallback` step in the next section becomes required rather than deferred.

Two non-hydration findings from the same capture, both publisher-agnostic and out of scope here: `in_content` 0px collapses are autoblog's per-breakpoint duplicate slots (`not-size-a/b/c/d`, `display:none` off-breakpoint) plus below-fold GPT lazy-fetch, not the gate — TS currently fires SSAT requests at hidden duplicate containers, which is worth a follow-up.

### Non-blocking

- **`requestIdleCallback` after the signal** — deferred, not dropped, with the reasoning and the adoption condition recorded in the spec's Update section. Discussion in the thread.
- **Synchronous pre-check before installing the poll** — done, in the guarded form suggested, with a test asserting `vi.getTimerCount() === 0` so the pre-check is observable. Two existing tests were adjusted to keep exercising the interval path rather than silently moving onto the new synchronous one.
- **`__next_f` is a Next.js internal** — documented on `installScheduleInitialAdInit`: App Router only (Next 13+), Pages Router and any future rename of the flight global take the `load` path, failure mode is slow rather than broken, re-check on a gated publisher's major Next upgrade.
- **Stale `build_bids_script` comment** — updated (comment only; emitted script unchanged). Same for `TsjsApi.scheduleInitialAdInit` in `src/core/types.ts`, which also still described the `load`-only gate.
- **Reverse signal order** — new test `fires exactly once when a load event precedes the runtime signal`.

### Verification

- `npx vitest run` — 437/437 (scheduler suite 11/11)
- `npm run format` (js) and `npm run format` (docs) — clean
- `cargo fmt --all -- --check`, `cargo clippy-fastly` — clean
- `bids_script_is_xss_safe`, `bids_script_defers_ad_init_until_after_hydration`, `bids_script_schedules_ad_init_without_retry_timer` — green

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -524,14 +539,28 @@ function installInitialLoadDetector(ts: TsjsApi): void {
  * initial call, while an `/a → /b → /a` round trip — where the URL compares
  * equal again — must.
  *
+ * `window.__next_f` is a Next.js **internal**: the App Router RSC flight-data
```
aram356: 🔧 **wrench** — This doc block still documents the abandoned `__next_f` gate the PR body reports measured at 8/8 #418:

- Lines 518–522: "Deferral: gate on the first hydration signal to arrive — the Next.js App Router runtime patching `window.__next_f`…"
- Lines 542–552: the entire `window.__next_f` internals paragraph ("the name we poll" — nothing polls it anymore)
- Lines 561–563: "The `__next_f` poll fires independently of rAF … unchanged by the runtime signal"

The function body 20 lines below implements (and correctly comments) the container-hydration gate, so the doc block directly contradicts the code it documents. The 542–552 paragraph has a natural replacement: `__reactFiber$`/`__reactProps$` are React internals, a rename means the containers never report hydrated and the publisher takes the `load` path — which the `isReactHydrated` helper comment already states; this block should say the same for the shipped signal.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -94,6 +94,11 @@
   // and deliberately identical to the bundle scheduler — the impression is
   // spent on a viewed tab, and the post-hydration guarantee holds whenever
   // the request is actually issued.
+  //
+  // This fallback stays on `load` only — it does not poll for the Next.js
+  // runtime signal the bundle scheduler uses. It exists for the
```
aram356: 🔧 **wrench** — "it does not poll for the Next.js runtime signal the bundle scheduler uses" — the bundle scheduler no longer uses the Next.js runtime signal; it polls the ad-slot containers for React hydration. Suggest: "it does not poll for container hydration the way the bundle scheduler does."

crates/trusted-server-js/lib/src/core/types.ts
```diff
@@ -250,16 +250,20 @@ export interface TsjsApi {
...
+   * server-injected `</body>` bids script with the SSR bids payload. The whole
+   * initial pass is pinned to navigation generation 0 (the SSR document): if an
+   * SPA navigation has already committed — or commits while the deferred
+   * callback is pending — the payload is dropped and `adInit()` is not run, so a
+   * stale SSR bootstrap can neither clobber the live route's bids nor re-run it.
    * Lives in the bundle so the lifecycle is executable under test and shares
    * [`navGeneration`] with the SPA auction hook; `gpt_bootstrap.js` installs
-   * a minimal fallback for pages where the bundle fails to load.
+   * a minimal fallback for pages where the bundle fails to load (that fallback
+   * gates on `load` only — the runtime signal is a bundle-side optimization).
```
aram356: 🔧 **wrench** — "the runtime signal is a bundle-side optimization" — stale term from the `__next_f` revision; the bundle-side signal is now container hydration. Suggest: "the container-hydration signal is a bundle-side optimization."

docs/superpowers/specs/2026-07-24-adinit-hydration-gate-design.md
```diff
@@ -0,0 +1,375 @@
...
+
+1. A subsequent client-side refresh clears `ts_initial` and re-auctions the slot. This is the
+   intended **single-shot SSAT** design, not a defect.
+2. SSAT itself is delivered **very late** (gated on `window.load`), so it lands close to — or on top
+   of — the publisher's ongoing refresh cadence.
+
+The root cause worth fixing is (2): SSAT is late. That is what this spec addresses. The refresh
+behavior in (1) is left as-is.
+
+## Update (post-deployment live test): pivot from chunk-await to `__next_f`
```
aram356, line 26: 🔧 **wrench** — The design trail ends here at the `__next_f` revision, presenting as current a gate the PR body reports was measured unsafe: this section says the `__next_f` gate "is reliable and early", and the `requestIdleCallback` deferral below leans on the "#418 at 2/38 (~5%) … not impression-costing" reading that the PR body explicitly retracts (controlled capture: 8/8 #418, 3/6 destroyed creatives on React-owned article-page slots). Neither this spec nor the plan mentions the container-hydration gate at all.

Keeping the historical trail intact is fine — but both docs need a third banner paragraph recording: (a) the `__next_f` gate measured unsafe, with the capture numbers; (b) the pivot to gating on the ad-slot containers reporting React-hydrated (`isReactHydrated`, `gpt/index.ts`); (c) a correction note where the ~5% claim is used as justification. Without it, a future reader following these docs restores a gate that reproduces #418 at 100% on React-owned slots. (Same one-paragraph addition to the plan's banner at line 13.)

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -548,9 +577,58 @@ function installScheduleInitialAdInit(ts: TsjsApi): void {
...
-      window.addEventListener('load', afterHydrationFrames, { once: true });
+      return;
+    }
+    // Fire on whichever signal arrives first, exactly once:
+    //  - the slot containers adInit is about to mutate having been hydrated by
+    //    React, checked directly on those elements. This is the early signal.
+    //  - window.load: unconditional fallback, and the only signal where the
+    //    containers never report hydrated (non-React publishers, or a React
+    //    internal rename), so the failure mode stays "slow, not broken".
+    // The double requestAnimationFrame after the trigger still lands the call
```
aram356: 🤔 **thinking** — "The double requestAnimationFrame after the trigger still lands the call after React commits" asserts more than the mechanism guarantees. React attaches `__reactFiber$` when the hydration **render pass** claims the host node; the enclosing commit lands later, and a time-sliced hydration pass can span multiple frames after the containers are claimed. Your measurement (adInit median ~1.1s after the hydration commit, never before) shows this holds in practice — likely because the last containers are claimed near the end of the pass — but that's an empirical property of the measured page, not a construction. Suggest restating this as the empirical claim it is, and keeping #418 in the deploy A/B checklist (it's already there). No code change requested.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -548,9 +577,58 @@ function installScheduleInitialAdInit(ts: TsjsApi): void {
...
+    };
+    window.addEventListener('load', fire, { once: true });
+    // The whole initial pass is one adInit call, so GPT issues a single batched
+    // ad request; requiring every currently-present container to be hydrated
+    // keeps that batch intact. Containers that appear later are not part of the
+    // initial pass today either.
+    const slotContainersHydrated = (): boolean => {
+      const slots = ts.adSlots ?? [];
+      if (slots.length === 0) return false;
+      const roots = slots.flatMap((slot) => candidateSlotRoots(slot.div_id));
```
aram356: 🤔 **thinking** — A configured slot whose element is **not in the DOM** contributes nothing to `roots` (`candidateSlotRoots` returns `[]`), so the gate can fire while that container is still missing — and `adInit` then skips it silently with no retry. That's inconsistent with the laggard rule (present-but-unhydrated holds the batch; absent doesn't hold anything), and the comment above — "Containers that appear later are not part of the initial pass today either" — isn't quite true: under the `load` gate, a client-rendered container that mounts post-hydration but pre-`load` *was* part of the initial pass.

Strict form, which also short-circuits per slot instead of resolving every root up front:

```ts
const slotContainersHydrated = (): boolean => {
  const slots = ts.adSlots ?? [];
  if (slots.length === 0) return false;
  return slots.every((slot) => {
    const roots = candidateSlotRoots(slot.div_id);
    return roots.length > 0 && roots.every(isReactHydrated);
  });
};
```

Trade-off to weigh: a config-drift slot that never materializes on a route would pin that route to the `load` path (safe-slow, same failure mode as the laggard case) — but your own follow-up note about the 19 off-breakpoint duplicate containers shows config↔DOM looseness is real. Either tighten it or state the chosen trade-off in the comment.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -548,9 +577,58 @@ function installScheduleInitialAdInit(ts: TsjsApi): void {
...
+    // already have hydrated the containers by then, so check once synchronously
+    // before installing the interval. The `load` listener above stays
+    // registered; `once: true` plus the `fired` guard make it inert.
+    if (slotContainersHydrated()) {
+      fire();
+      return;
     }
+    // The poll stops itself on the first tick after either signal wins, so a
+    // later hydration cannot re-run adInit (the `fired` guard also holds).
+    const poll: ReturnType<typeof setInterval> = setInterval(() => {
```
aram356: ⛏ **nitpick** — Cost note: each 50ms tick re-resolves every slot, and a prefix `div_id` that misses `getElementById` runs `document.querySelectorAll('[id]')` over the whole document per slot per tick (~180 full scans across the measured ~9s hydration window). Per-tick re-resolution is the right call — React can discard and recreate SSR nodes — so no caching suggested; the per-slot short-circuit in the `slotContainersHydrated` restructure above would cut most of it if profiles ever show cost.

aram356 (changes requested):
## Summary

Re-review of the container-hydration revision (head bf831b815, base `main`). The pivot is well-founded: the controlled capture (8/8 #418, 3/6 destroyed creatives under the `__next_f` gate, with control arms isolating `adInit` as the sole cause) converts the previous revision's open risk into measurement, and the new gate — poll the ad-slot containers `adInit` will mutate for React's `__reactFiber$`/`__reactProps$` own-properties, all containers required, `load` fallback, sync pre-check, double rAF — is the right shape: per-element, no invented timing constant, failure mode slow-not-broken. The gate logic checks out (exactly-once across both signal orders, poll self-cleanup, per-tick re-resolution handling React node replacement), and the 16-test scheduler suite covers every lifecycle path raised in the previous review. What blocks is documentation lagging one revision behind the code — the same class as the previous review's blocking finding — in the shipped doc comments and in both committed design docs, which still present the measured-unsafe `__next_f` gate as current.

## Blocking

### 🔧 wrench

- **Shipped code comments still document the abandoned `__next_f` gate.** The `installScheduleInitialAdInit` doc block (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts:518-522`, `542-552`, `561-563`) describes polling `window.__next_f` — contradicting the implementation and the correct block comment inside the function — and the same stale term survives in `crates/trusted-server-core/src/integrations/gpt_bootstrap.js:98-99` and `crates/trusted-server-js/lib/src/core/types.ts:266`. See inline comments.

- **Both committed docs end the design trail at the `__next_f` revision, presenting a measured-unsafe gate as current.** The spec's Update section calls the `__next_f` gate "reliable and early" and justifies deferring `requestIdleCallback` with the "~5%, not impression-costing" reading the PR body explicitly retracts; the plan's banner likewise stops at "The gate now polls `window.__next_f.push`". Neither doc mentions the container gate. Both banners need a third paragraph recording the unsafe measurement, the pivot, and a correction where the ~5% claim is load-bearing. See inline comment on the spec.

### ❓ question

- **What is the merge bar now?** The #418 safety bar is met and measured (0/8 with control arms — a stronger result than the original "confirm 0" ask, since it now has a baseline). The latency win — the reason this branch exists — is explicitly unverified: through the local proxy the container gate and `load` fire ~50ms apart, and the ~seconds-vs-~52s claim only materializes on the deployed heavy publisher. Is the deploy A/B a merge blocker, or do you merge on safety evidence and verify latency post-deploy? Relatedly: CodeQL on this head failed on runner acquisition ("job was not acquired by Runner of type hosted") — an infrastructure flake that needs a re-run — and the remaining CI gates have not reported on this head yet.

## Non-blocking

### 🤔 thinking

- **`__reactFiber$` appears during hydration's render pass, not at commit** — the comment's "still lands the call after React commits" is an empirical claim, not a construction; see inline comment (`gpt/index.ts:588`).
- **Configured slots with no matching element don't hold the gate**, contradicting the batching rationale and the "not part of the initial pass today either" claim; strict per-slot form suggested with its trade-off; see inline comment (`gpt/index.ts:613`).

### ⛏ nitpick

- **Poll cost on prefix `div_id`s** — full-document `[id]` scans per slot per 50ms tick; see inline comment (`gpt/index.ts:627`).

## CI Status

- GitHub checks (this head): CodeQL FAIL — runner-acquisition infrastructure error, not a code alert; re-run needed. Other gates not yet reported.
- JS tests (`npx vitest run` at head, local): PASS — 562/562, scheduler suite 16/16
- JS + docs format (prettier, local): PASS
- Rust: diff is comment-only (`publisher.rs`, `gpt_bootstrap.js` verified line-by-line); fmt/clippy claimed clean in the PR body and enforced by CI on re-run.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
prk-Jr: Fixed in b431a225.

The doc block is rewritten around the container gate, and all three stale passages you listed are gone:

- **518–522** now describes polling the ad-slot containers for React hydration, with `load` starting a bounded grace period rather than firing the call.
- **542–552** — replaced with exactly the parallel you suggested: `__reactFiber$` / `__reactProps$` are React internals with a per-instance randomized suffix; a rename means the containers never report hydrated and the publisher takes the `load` path, so the failure mode stays slow-not-broken. Re-check on a gated publisher's major **React** upgrade.
- **561–563** — the hidden-tab paragraph now says "the container poll fires independently of rAF".

One deliberate addition: a short paragraph recording that the `__next_f` gate was measured unsafe (8/8 #418, creative destroyed 3/6) and why it was replaced. Without it the next reader has no record of why a page-global signal was tried and rejected, and the shape looks arbitrary.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
prk-Jr: Fixed in b431a225, using your wording verbatim:

```js
// This fallback stays on `load` only — it does not poll for container
// hydration the way the bundle scheduler does. It exists for the
// bundle-failed-to-load path, where staying small matters more than being
// early.
```

crates/trusted-server-js/lib/src/core/types.ts
prk-Jr: Fixed in b431a225:

```ts
* a minimal fallback for pages where the bundle fails to load (that fallback
* gates on `load` only — the container-hydration signal is a bundle-side
* optimization).
```

docs/superpowers/specs/2026-07-24-adinit-hydration-gate-design.md
prk-Jr: Fixed in b431a225 — both docs, structured as you asked (trail intact, third banner paragraph on top).

**Spec** gets an "Update 2" section carrying (a) the measurement, (b) the pivot, (c) the correction:

| arm | #418 | SSAT creative destroyed |
| --- | --- | --- |
| Trusted Server off (publisher alone) | 0/8 | — |
| TS on, `adInit` suppressed | 0/8 | — |
| TS on, `window.load` gate (#945) | 0/6 | 0/6 |
| TS on, **`__next_f` gate** | 8/8 | **3/6** |
| TS on, **container-hydration gate** | 0/8 | 0/8 |

**Plan** gets the same as a third banner paragraph at the top.

On the `~5%` correction specifically — you were right that it is load-bearing, and it does not hold. That figure came from pages whose ad slots are not React-owned. On an article page whose containers are React-rendered with `useId` ids, the same gate reproduces #418 at **100%**, and it does cost impressions: React regenerates the mismatched subtree and destroys the creative that just rendered. Both docs now say so at the point where the deferral argument uses it.

Both banners end with the warning you asked for — restoring the `__next_f` gate from the sections below reproduces #418 at 100% on React-owned slots.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
prk-Jr: You're right, and it's restated as an empirical claim in b431a225:

```ts
// Note what this does and does not guarantee. React attaches the fiber
// properties when the hydration *render* pass claims a host node; the
// enclosing commit lands later, and a time-sliced pass can span frames after
// the containers are claimed. So "after React commits" is an empirical
// result, not a construction — measured on the gated publisher, adInit lands
// a median 1.1s after the hydration commit and never before it, on either
// page type. #418 stays on the deploy A/B checklist for that reason.
```

Your read of *why* it holds in practice matches what the data shows: the gate waits for the **last** configured container, which is claimed near the end of the pass, so the remaining commit work is short relative to the two frames plus poll granularity. That is a property of the measured pages, not a guarantee — hence keeping #418 in the A/B.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
prk-Jr: Adopted your strict form verbatim in b431a225:

```ts
const slotContainersHydrated = (): boolean => {
  const slots = ts.adSlots ?? [];
  if (slots.length === 0) return false;
  return slots.every((slot) => {
    const roots = candidateSlotRoots(slot.div_id);
    return roots.length > 0 && roots.every(isReactHydrated);
  });
};
```

You were also right that "containers that appear later are not part of the initial pass today either" was wrong — that comment is gone. Measured on the gated publisher's **homepage**, whose `ad-header-0` / `ad-fixed_bottom-0` containers are client-rendered and absent from the SSR HTML entirely: under the `load` gate, **4 of 5 loads applied no server-side targeting at all** — `adInit` fired ~100ms before React created those containers, matched nothing, and every slot rendered without `ts_initial`. With the strict form, 8/8 loads target all three slots. So this is a coverage bug independent of the #418 timing, and the larger of the two effects.

Tightening it did surface the trade-off you flagged, in a sharper form than expected. Making the predicate strict makes it **unsatisfiable before `load`** on both page types (the article page has no `ad-header-0` until ~13.5s), so everything fell through to `load` — and measuring that revealed `load` has the same defect the predicate was fixing: it fires **65ms (article) to ~1000ms (homepage) before the containers are ready**, and on some runs before React had created its root at all. That measured 7/8 #418 on both pages.

So `load` is no longer a trigger. It starts a bounded grace period; the container check still ends the wait:

```ts
if (!loadSeen) return;
// No configured slots, or a static page whose slots are all present and
// React-owns nothing — fire, exactly as before.
if (slots.length === 0) { fire(); return; }
const everySlotResolves = slots.every((slot) => candidateSlotRoots(slot.div_id).length > 0);
if (everySlotResolves && !documentUsesReact()) { fire(); return; }
ticksSinceLoad += 1;
if (ticksSinceLoad >= GRACE_TICKS) fire();
```

Two notes on that:

- The config-drift case you raised is handled by the grace period rather than pinning to `load` forever: a slot that never materializes costs a bounded wait, then fires degraded. It never elapsed in testing — containers arrived a median 250ms after `load`, against a 3s cap.
- `documentUsesReact()` checks `document` for `__reactContainer$` first, not `document.body`. `hydrateRoot(document, …)` marks `document` itself, and that marker appears ~700ms before the per-element fibers — checking body alone misreads a React page as static, because on this publisher `load` can fire before React has created its root.

Result on both pages, 8 loads each: 0/8 #418, 0/8 destroyed creatives, and all three configured slots targeted on 8/8 homepage loads.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
prk-Jr: The restructure landed, so most of it is cut as you predicted — `slots.every(...)` short-circuits on the first slot that is missing or unhydrated, which on a cold page is the first slot checked. Per-tick re-resolution is retained for exactly the reason you gave: React discards and recreates the SSR nodes, and on the homepage the containers `adInit` ultimately targets do not exist at all until React creates them, so a cached lookup would pin to nodes that get thrown away.

The full-document `[id]` scan is still there for a prefix `div_id` that misses `getElementById`. No caching added. Worth noting the window is now bounded on both ends: the poll stops at the first satisfied tick, and `load` starts a 3s grace cap, so the unbounded-until-hydration case that motivated the ~180-scan estimate no longer exists.

aram356:
Closing in place of #1013

Issue #970, opened 2026-07-27
Remove the deprecated /__ts/page-bids transition alias


prk-Jr:
## Summary

Follow-up to #942, which renamed the SPA re-auction endpoint to `/_ts/page-bids` and kept `/__ts/page-bids` registered as a deprecated transition alias. This issue tracks removing that alias.

## Why the alias exists

A browser runs whatever `tsjs` bundle the server already served it. Pages loaded before the rename — and bundles still sitting in browser/CDN caches — keep requesting `/__ts/page-bids`, and on a SPA that path is what delivers ads for in-session navigations. Removing it in the same change would have dropped ads for those clients until their bundle refreshed, so both paths route to the same handler during the transition.

## Precondition

Confirm from access logs that `/__ts/page-bids` traffic has fallen to zero (or to bot-only noise) before removing. Removing it early reintroduces exactly the ad-serving gap the alias was added to prevent.

## Scope

Remove the alias and its coverage:

- `crates/trusted-server-core/src/publisher.rs` — delete the `PAGE_BIDS_LEGACY_PATH` const and its doc comment; drop the alias mention from the `page_bids_preflight_denied` doc.
- `crates/trusted-server-adapter-fastly/src/app.rs` — delete the alias `NamedRoute` entry and the `PAGE_BIDS_LEGACY_PATH` imports (module-level and in `mod tests`).
- `crates/trusted-server-adapter-axum/src/app.rs` — delete the alias `NamedRoute` entry; `named_routes()` return type shrinks `[NamedRoute; 13]` -> `[NamedRoute; 12]`.
- `crates/trusted-server-adapter-cloudflare/src/app.rs` — delete the alias `.get(...)` and `.route(..., Method::OPTIONS, ...)` registrations.
- `crates/trusted-server-adapter-spin/src/app.rs` — delete the alias `.get(...)` / `.route(...)` registrations and the `named_fallback_paths()` entry; return type shrinks `[(&'static str, &'static [Method]); 13]` -> `[...; 12]`. The `page_bids_handler.clone()` at the canonical registration can drop the `.clone()`.
- `crates/trusted-server-adapter-fastly/src/app.rs` (tests) — `page_bids_serves_canonical_path_and_deprecated_alias`: drop the alias arm and rename to reflect canonical-only coverage.
- `crates/trusted-server-integration-tests/tests/parity.rs` — `page_bids_options_preflight_denied_parity`: drop `/__ts/page-bids` from the loop.

## Verification

[...]

PR #971, opened 2026-07-27
Rename /__ts/page-bids to /_ts/page-bids

### LINKED ISSUE #942: Rename /__ts/page-bids to /_ts/page-bids for internal-route naming consistency
## Summary

`/__ts/page-bids` is the only internal Trusted Server URL path that uses a **double-underscore (`__`) prefix**. Every other internal route lives under the single-underscore `/_ts/` namespace. Rename the route to **`/_ts/page-bids`** so the internal-path convention is consistent.

## Current state

Non-compliant (double underscore):
- `/__ts/page-bids`

All other internal routes use `/_ts/`:
- `/_ts/set-tester`, `/_ts/clear-tester`
- `/_ts/admin/keys/rotate`, `/_ts/admin/keys/deactivate`
- `/_ts/api/v1/identify`, `/_ts/api/v1/batch-sync`
- `/_ts/debug/ja4`

A survey of URL string literals confirms `/__ts/page-bids` is the **only** path using `__`. (`__ts_rsc_payload_` and the `__tsjs_*` window globals are internal JS identifiers, not URL paths, so they are out of scope.)

There is also an existing internal inconsistency that suggests the `__` was unintentional: the unit test `make_page_bids_request` already builds its request with the single-underscore form `/_ts/page-bids` (`crates/trusted-server-core/src/publisher.rs` ~L4945), even though the route is registered as `/__ts/page-bids`. The test passes only because the same-origin gate inspects headers/Fetch-Metadata, not the path.

## Scope — references to update in lockstep (~42 occurrences)

The `tsjs` client and the server route must agree on the path, so all of these change together:

Server route registration (all four adapters):
- `crates/trusted-server-adapter-fastly/src/app.rs:1086`
- `crates/trusted-server-adapter-axum/src/app.rs:331`
- `crates/trusted-server-adapter-cloudflare/src/app.rs:487, 494`
- `crates/trusted-server-adapter-spin/src/app.rs:153, 734, 736`

Client (must match the served bundle):
- `crates/trusted-server-js/lib/src/integrations/gpt/index.ts:731` (the `fetch`), plus the doc at L699

Tests + docs:
- `crates/trusted-server-integration-tests/tests/parity.rs:705-707`
- `crates/trusted-server-core/src/publisher.rs:2337` (doc comment), and align the test at ~L4945 (already uses `/_ts/`)
- any references under `docs/`

## Compatibility / rollout note

The browser runs whatever `tsjs` bundle the server served it, so a single atomic deploy updates client + server together. However, **already-cached/older bundles in the wild will keep requesting `/__ts/page-bids`**, and on a Next.js SPA that path drives ad delivery for in-session navigations — a hard rename would drop ads for those clients until their bundle refreshes.

Recommended rollout:
1. Register **both** `/__ts/page-bids` (deprecated alias) and `/_ts/page-bids`, routing to the same handler.
2. Switch the client `fetch` to `/_ts/page-bids`.
3. After old bundles have aged out of caches, remove the `/__ts/page-bids` alias.

## Acceptance criteria

- The route is served at `/_ts/page-bids` across all four adapters.
- The `tsjs` client fetches `/_ts/page-bids`.
- Parity tests, unit tests, and docs reference the single-underscore path.
- A transition alias for `/__ts/page-bids` exists (and a follow-up to remove it), so no ad-serving gap during rollout.
- No other internal URL paths use a `__` prefix (confirmed: `/__ts/page-bids` is the only one).

prk-Jr:
## Summary

- `/__ts/page-bids` was the only internal Trusted Server route using a double-underscore prefix; every other internal path lives under `/_ts/`. This moves the SPA re-auction endpoint to `/_ts/page-bids` and switches the tsjs client fetch to match.
- A browser runs whichever tsjs bundle it was already served, so pages loaded before the rename — and cached bundles — keep requesting the old path. On a SPA that path is what delivers ads for in-session navigations, so `/__ts/page-bids` stays registered to the same handler as a transition alias. Removal is tracked by #970.
- Route coverage was missing for the page-bids `GET` registrations on Cloudflare and Spin. Mutation testing confirmed a broken registration previously passed every suite; that gap is now closed.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/publisher.rs` | Add `PAGE_BIDS_PATH` / `PAGE_BIDS_LEGACY_PATH` as the single source of truth; log an `info` line when the alias serves a gate-passed request; add `deprecated_alias_response_matches_canonical_path`; split `make_page_bids_request_on` out of `make_page_bids_request`; update doc comments |
| `crates/trusted-server-adapter-fastly/src/app.rs` | Register both paths via `NAMED_ROUTES`; add `page_bids_serves_canonical_path_and_deprecated_alias`, pinning the consts to their literal values |
| `crates/trusted-server-adapter-axum/src/app.rs` | Register both paths; `named_routes()` widens to `[NamedRoute; 13]` |
| `crates/trusted-server-adapter-cloudflare/src/app.rs` | Collapse four near-identical registrations into a loop over both paths, matching the existing `legacy_admin_deny` pattern |
| `crates/trusted-server-adapter-spin/src/app.rs` | Register both paths; `named_fallback_paths()` widens to 13 so non-primary methods still reach the publisher fallback |
| `crates/trusted-server-adapter-axum/tests/routes.rs` | Pin `GET`/`OPTIONS` on both paths in `all_explicit_routes_are_registered` |
| `crates/trusted-server-adapter-cloudflare/tests/routes.rs` | Same route-table assertions |
| `crates/trusted-server-adapter-spin/tests/routes.rs` | Add `page_bids_get_is_routed_on_canonical_path_and_alias` (Spin has no route-list test) |
| `crates/trusted-server-integration-tests/tests/parity.rs` | Preflight-denial parity now loops both paths |
| `crates/trusted-server-js/lib/src/integrations/gpt/index.ts` | SPA hook fetches `/_ts/page-bids` |
| `crates/trusted-server-js/lib/test/integrations/gpt/spa_hook.test.ts` | Update three fetch assertions |
| `crates/trusted-server-core/src/auction/{endpoints,orchestrator,telemetry}.rs`, `integrations/gpt.rs` | Doc-comment path updates |

## Closes

Closes #942

## Test plan

- [x] `cargo test-fastly && cargo test-axum` — also `cargo test-cloudflare && cargo test-spin`
- [x] `cargo clippy-fastly && cargo clippy-axum` — also `clippy-cloudflare`, `clippy-cloudflare-wasm`, `clippy-spin-native`, `clippy-spin-wasm`
- [x] JS tests: `cd crates/trusted-server-js/lib && npx vitest run` (411 passed)
- [x] Other: parity suite (`--test parity`, 13 passed); mutation testing of all 8 registrations

### Mutation testing

Because the alias exists precisely to avoid a silent ad-serving gap, each registration was deliberately broken to confirm a test fails. All 8 are caught (Fastly / Axum / Cloudflare / Spin × canonical / alias). Before this PR's added coverage, breaking the Cloudflare `GET` registration — canonical *or* alias — passed every suite.

Verified the built bundle rather than assuming: `dist/` is gitignored and regenerated by `build.rs`; the rebuilt `tsjs-gpt.js` contains `/_ts/page-bids` once and `/__ts/` zero times. The separately-shipped external Prebid bundle takes only `prebid/index.ts` and does not embed this path, so there is no half-ship risk.

Also confirmed the namespace move does not change behaviour: every handler regex in the repo is `^/_ts/admin`, which does not match `/_ts/page-bids`, so the route is not pulled behind basic auth. No `starts_with("/_ts")` special-casing exists; integration path filters are all under `/integrations/`; cache and privacy logic key on headers, never path.

## Notes

The `docs/superpowers/` plan and spec files still reference `/__ts/page-bids`. They are dated point-in-time design records that also describe code shapes since changed, so they were left as history rather than rewritten. Happy to sweep them if reviewers prefer.

Not applicable: no config-derived regex or pattern compilation is touched.

## Checklist

- [x] New code has tests


crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -735,7 +735,7 @@ export function installSpaAuctionHook(): void {
     inflight = controller;
 
     try {
-      const res = await fetch(`/__ts/page-bids?path=${encodeURIComponent(path)}`, {
+      const res = await fetch(`/_ts/page-bids?path=${encodeURIComponent(path)}`, {
```
ChristianPavilonis: 🔧 **The renamed route can silently put SPA auctions behind operator Basic Auth**

`[[handlers]].path` is an operator-controlled regex evaluated before route dispatch. An existing valid configuration such as `path = "^/_ts"` did not match `/__ts/page-bids`, but now matches `/_ts/page-bids` and returns `401`. The new bundle always requests the canonical path and does not fall back to the retained legacy alias.

This can stop anonymous users on affected deployments from receiving ads after SPA navigations. Repository searches for `^/_ts/admin` cannot rule this out because production `trusted-server.toml` files are operator-owned and gitignored.

Make the rollout safe for broad handler patterns: either explicitly exempt both page-bids paths from Basic Auth, or retain a client fallback to `/__ts/page-bids` while documenting and migrating affected operator configurations. Add an adapter-level regression test using `path = "^/_ts"`.

ChristianPavilonis (approved):
## Summary

Reviewed the route rename and transition alias across core, all four adapters, the TSJS client, rollout behavior, and test coverage. The routing and security-gate implementation is otherwise consistent, and all CI checks pass. Approving with one high-severity compatibility finding that should be addressed for deployments using broad operator-defined path authentication rules.

aram356 (approved):

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
aram356: 🤔 **thinking** — No reverse alias on rollback. The server-side alias protects forward skew (old bundle → new server), but a server rollback after this release inverts it: browsers holding this new bundle fetch `/_ts/page-bids`, which a pre-rename server doesn't register — the request falls through to the publisher-origin proxy and SPA navigations silently lose ads, the exact failure mode the alias exists to prevent. Exposure is bounded (the tsjs bundle is served by this same server with `max-age=300`, plus already-open SPA tabs), but rollback is lossier than roll-forward. Worth a one-line note in #970 or the release notes: prefer roll-forward, or accept a bounded ad gap on rollback.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -3409,7 +3409,27 @@ fn is_supported_content_encoding(encoding: &str) -> bool {
     matches!(encoding, "" | "identity" | "gzip" | "deflate" | "br")
 }
 
-/// Same-origin gate for `/__ts/page-bids`.
+/// Canonical URL path of the SPA re-auction endpoint.
+///
+/// Lives in the internal `/_ts/` namespace shared by every other Trusted
+/// Server route. Adapters register this path; the tsjs SPA hook fetches it.
+pub const PAGE_BIDS_PATH: &str = "/_ts/page-bids";
```
aram356, line 3416: 🤔 **thinking** — Operator-owned `[[handlers]]` regexes could capture this path. `enforce_basic_auth` matches operator-configured path regexes from the gitignored `trusted-server.toml`. The PR verified all in-repo regexes are `^/_ts/admin`, but an operator using a broader pattern like `^/_ts` would now put the SPA ad endpoint behind basic auth. Browser-facing `/_ts/api/v1/identify` already makes blanket patterns unlikely, so this is low-probability — but a deployment-check line in #970 or the release notes would close it.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -3519,6 +3540,22 @@ pub async fn handle_page_bids(
...
+    // Deprecation signal for the transition alias. Logged only after the
+    // cross-site gate passes, so the count reflects genuine SPA clients still
+    // running a pre-rename tsjs bundle rather than anything a third-party page
+    // can inflate. This is the only in-app signal that
+    // `PAGE_BIDS_LEGACY_PATH` is still in use — the removal precondition in
+    // IABTechLab/trusted-server#970 is "no remaining traffic on the legacy
+    // path", which is otherwise only answerable from edge access logs. The line
+    // is self-limiting: it goes silent as old bundles age out, which is exactly
+    // the condition being waited on.
+    if req.uri().path() == PAGE_BIDS_LEGACY_PATH {
```
aram356: 🌱 **seedling** — This log line is the only in-app usage signal, and #970's removal precondition is otherwise answerable only from edge access logs. Adding a `Deprecation` header (RFC 9745) and `Link: <https://github.com/IABTechLab/trusted-server/issues/970>; rel="deprecation"` to responses served via the legacy path would let CDN/edge log pipelines measure remaining traffic without app-log access, from any vantage point.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -3519,6 +3540,22 @@ pub async fn handle_page_bids(
         return Ok(page_bids_preflight_denied());
     }
 
+    // Deprecation signal for the transition alias. Logged only after the
```
aram356: 📝 **note** — The `Creative opportunities not configured` 404 early-returns before this alias log, so legacy-path traffic on deployments without creative opportunities is never counted. Harmless (those deployments serve no SPA ads anyway) — noting only so a silent log isn't over-read as "zero traffic" on such configs when evaluating #970.

aram356 (approved):

PR #961, opened 2026-07-23
Add privacy-safe auction-to-creative tracing

ChristianPavilonis:
## Summary

- Add an opt-in, privacy-safe trace console that links Trusted Server auction outcomes to independently observed browser delivery evidence without exposing upstream request IDs, GAM identifiers, targeting, creative content, URLs, consent, or identity data.
- Preserve factual boundaries between Trusted Server selection, Prebid evidence, GAM render classification, iframe load, selected-creative acknowledgement, visibility, and GPT viewability.
- Cover publisher-owned and lazy GPT slots with bounded per-request generations and privacy-safe overlays even when no Trusted Server candidate exists.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/auction/` | Generate fresh auction and per-winning-bid trace IDs, retain terminal outcomes, and attach privacy-safe telemetry linkage. |
| `crates/trusted-server-core/src/integrations/ad_trace.rs` and adapter middleware | Implement exact console activation, private session handling, response sanitization, and outer-boundary finalization across supported adapters. |
| `crates/trusted-server-js/lib/src/core/ad_trace.ts` | Store bounded, monotonic, generation-scoped trace facts without retaining private publisher or ad-platform identifiers. |
| `crates/trusted-server-js/lib/src/integrations/ad_trace/` | Render factual console and overlay evidence while hiding unobserved stages and preserving independent evidence semantics. |
| `crates/trusted-server-js/lib/src/integrations/{gpt,prebid}/` | Correlate exact slot generations, adopt existing Trusted Server candidates, cover generic GPT lifecycles, bind safe top-document anchors, and fail closed on stale or ambiguous refresh events. |
| `crates/trusted-server-integration-tests/` | Add deterministic fixtures and browser coverage for TS winners, publisher-only pages, lazy slots, refreshes, backfill, empty renders, acknowledgement, visibility, viewability, and privacy boundaries. |
| Configuration, Tinybird, and documentation files | Add opt-in configuration and nullable privacy-safe trace linkage. |

## Closes

Closes #920

## Test plan

- [x] `cargo test-fastly && cargo test-axum`
- [x] `cargo clippy-fastly && cargo clippy-axum`
- [x] `cargo fmt --all -- --check`
- [x] JS tests: `cd crates/trusted-server-js/lib && npx vitest run` (503 passed)
- [x] JS format: `cd crates/trusted-server-js/lib && npm run format`
- [ ] Docs format: `cd docs && npm run format`
[...]

PR #974, opened 2026-07-28
Add GPT runtime diagnostics overlay

### LINKED ISSUE #973: Add opt-in GPT runtime diagnostics overlay
## User story

As a publisher operator, I want an opt-in, tab-local GPT runtime diagnostics overlay so that I can inspect GPT-observed slot lifecycles and exact DOM bindings without modifying publisher or auction behavior.

## Scope

- Deployment-controlled and query-activated diagnostics runtime.
- Bounded GPT lifecycle capture with conservative correlation.
- Read-only API, local export, closed-shadow panel, and viewport badges.
- Automated browser, runtime, and privacy/non-interference coverage.

ChristianPavilonis:
## Summary

- Add an opt-in, tab-local GPT runtime diagnostics console for inspecting GPT-observed slot lifecycles and exact DOM bindings.
- Keep diagnostics memory-only, bounded, local, and inactive by default without changing GPT, publisher, browser, auction, or network behavior.
- Add deterministic unit and browser coverage plus operator documentation for activation, interpretation, export, privacy, and limitations.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/integrations/gpt_diagnostics*` | Register the deployment-controlled integration and inject the early `ts_console` activation bootstrap. |
| `crates/trusted-server-core/src/{config,html_processor,migration_guards}.rs` | Wire configuration, injection ordering, validation, and migration guards. |
| `crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/` | Implement bounded lifecycle storage, documented GPT observation, conservative correlation, exact DOM binding, the read-only API/export, closed-shadow panel, and viewport badges. |
| `crates/trusted-server-js/lib/src/core/` | Publish the versioned diagnostics API and allowlisted export types. |
| `crates/trusted-server-js/lib/test/integrations/gpt_diagnostics/` | Cover activation, retention, matching, invalid order, binding ambiguity, API isolation, presentation lifecycle, badges, and runtime composition. |
| `crates/trusted-server-integration-tests/` | Add a deterministic GPT stub, Next.js fixture, and Playwright coverage for active/inactive behavior, lifecycle capture, ambiguity, remounting, export, privacy, and non-interference. |
| `docs/guide/integrations/gpt-diagnostics.md` | Document deployment, tab activation, observed facts, export, privacy boundaries, and limitations. |
| `docs/superpowers/{specs,plans}/2026-07-28-gpt-runtime-diagnostics-overlay*` | Record the finalized design, implementation plan, decisions, and verification status. |
| `trusted-server.example.toml` | Add the disabled-by-default integration configuration example. |

## Closes

Closes #973

## Test plan

- [x] Other:
  - `cargo test-cloudflare && cargo test-spin`
  - `cargo clippy-cloudflare && cargo clippy-cloudflare-wasm && cargo clippy-spin-native && cargo clippy-spin-wasm`
  - `./scripts/test-cli.sh`
  - `CI=1 ./scripts/integration-tests.sh --nocapture`
  - `./scripts/integration-tests-browser.sh`
  - `cd crates/trusted-server-js/lib && node build-all.mjs`
  - `cd docs && npm run build`

Live-publisher acceptance remains pending because no source-controlled publisher URL or credentials are available. The deterministic browser fixture covers the same activation, lifecycle, binding, export, privacy, and non-interference boundaries locally.

## Checklist

- [x] Uses project logging macros rather than `println!`
- [x] New code has tests


ChristianPavilonis:
Fixed the browser integration failure in 79d08323. The test used an unscoped `Home` link locator, which matched both the site navigation and the GPT fixture navigation in CI. The locator is now scoped to the `Fixture navigation` landmark.

Local verification: `./scripts/integration-tests-browser.sh` — Next.js 15 passed / 1 skipped; WordPress 5 passed / 11 skipped.

crates/trusted-server-core/src/integrations/gpt_diagnostics.rs
```diff
@@ -0,0 +1,233 @@
...
+    let Some(_config) =
+        settings.integration_config::<GptDiagnosticsConfig>(GPT_DIAGNOSTICS_INTEGRATION_ID)?
+    else {
+        return Ok(None);
+    };
+
+    let integration = Arc::new(GptDiagnosticsIntegration);
+    Ok(Some(
+        IntegrationRegistration::builder(GPT_DIAGNOSTICS_INTEGRATION_ID)
+            .with_head_injector(integration)
```
prk-Jr: 🤔 **thinking** — Worth weighing: enabling diagnostics ships 30 KB of tester-only code to every visitor.

Built from a clean worktree of `79d0832`, `tsjs-gpt_diagnostics.js` is **30,343 bytes** minified — the largest module in the bundle and 3× `tsjs-core.js` (10,031 B). With no `.with_deferred_js()`, `js_module_ids_immediate()` includes it in the synchronous first-party bundle injected at `<head>` start, so once an operator sets `enabled = true` every visitor downloads and parses it even though every tab stays inactive until `?ts_console=1`.

`enabled_registry_includes_diagnostics_in_immediate_modules` shows this is deliberate, and the reasoning holds: deferring would delay listener installation past the first `slotRequested`/`slotRenderEnded` and lose the initial impression's cycle, which is the one operators most want. So this is not a suggestion to defer.

The tension comes from `sessionStorage`-only activation leaving the server no signal, which forces an all-or-nothing payload decision. A server-set session cookie on the activation directive would resolve both at once — the module is concatenated only for an activated tester, and for that tester it still loads immediately with listeners installed exactly as early as today. That is what #961 did with `__Host-ts-console`, and this test could keep asserting immediate loading, just scoped to activated sessions.

If shipping to all visitors is an accepted cost for now, it may be worth recording the rationale in `gpt-diagnostics.md` beside the deployment configuration so operators know what enabling it costs their page weight.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts
```diff
@@ -0,0 +1,203 @@
...
+    this.mutationObserver = new Observer((records) => {
+      const slotElementIds = new Set(
+        this.store
+          .snapshot()
+          .slots.map((slot) => slot.slotElementId)
+          .filter((slotElementId): slotElementId is string => Boolean(slotElementId))
+      );
+      const relevant = records.some(
+        (record) =>
+          record.type === 'childList' ||
```
prk-Jr: 🤔 **thinking** — Any DOM insertion anywhere on the page triggers a full snapshot and a forced layout.

Attribute records are correctly filtered to known slot element IDs, but `record.type === 'childList'` matches every childList mutation in the document. On an ad-heavy page that fires more or less continuously, and each resulting `update()` deep-clones the whole store and calls `getBoundingClientRect()` for every bound slot.

It compounds because four consumers snapshot independently, each cloning the entire store per frame: `binding.ts:109`, `badges.ts:119`, `overlay.ts:353`, `api.ts:78`.

Filtering childList records the way the attribute records already are — added or removed nodes intersecting a known slot element ID — and/or sharing one snapshot per frame would remove most of the churn.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
```diff
@@ -0,0 +1,473 @@
...
+  }
+
+  recordSlotOnload(slot: GptDiagnosticsSlotLike): void {
+    const timestampMs = this.timestamp();
+    this.matchCycle(
+      'slotOnload',
+      slot,
+      timestampMs,
+      (cycle) =>
+        cycle.renderAtMs !== undefined && cycle.isEmpty === false && cycle.loadAtMs === undefined,
```
prk-Jr: 🤔 **thinking** — A load or viewable callback is dropped when GPT omits `isEmpty`.

This matcher and the `impressionViewable` one at line 264 both require `cycle.isEmpty === false`. The observer normalizes a non-boolean `isEmpty` to `undefined`, and the overlay already treats that state as "Pending render" rather than "not filled" — so a legitimate `slotOnload` on such a cycle becomes `unmatched` plus a `no_compatible_request_cycle` issue instead of being recorded.

`isEmpty !== true` keeps the conservative intent — never attach a load to a known-empty cycle — without discarding the ambiguous one:

```suggestion
        cycle.renderAtMs !== undefined && cycle.isEmpty !== true && cycle.loadAtMs === undefined,
```

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/binding.ts
```diff
@@ -0,0 +1,232 @@
...
+    if (!candidate || !(candidate instanceof this.window.HTMLElement) || !candidate.isConnected) {
+      return {
+        binding: { status: 'unbound', reason: 'missing_element' },
+        visible: false,
+      };
+    }
+
+    const element = candidate as HTMLElement;
+    const escape = this.window.CSS?.escape;
+    if (typeof escape !== 'function') {
```
prk-Jr, line 203: ♻️ **refactor** — A missing `CSS.escape` is reported to the operator as a duplicate DOM ID.

When `CSS.escape` is unavailable this returns `ambiguous / duplicate_dom_id`, and `GptDiagnosticsBindingReason` is part of the exported JSON, so an operator debugging an unbound slot is told they have an ID collision that does not exist.

A distinct reason in `core/types.ts` would keep the export honest:

```ts
export type GptDiagnosticsBindingReason =
  | 'missing_slot_element_id'
  | 'missing_element'
  | 'duplicate_dom_id'
  | 'duplicate_check_unavailable'
  | 'duplicate_gpt_slot_id';
```

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts
```diff
@@ -0,0 +1,203 @@
...
+
+      const rectangle = element.getBoundingClientRect();
+      if (!intersectsViewport(rectangle, this.window)) continue;
+      observedElements.push(element);
+
+      const badge = this.document.createElement('div');
+      badge.className = 'tsgd-badge';
+      badge.dataset.runtimeSlot = String(slot.runtimeSlotNumber);
+      badge.textContent = badgeText(cycle);
+      badge.style.left = `${Math.max(4, Math.min(rectangle.left, this.window.innerWidth - 264))}px`;
```
prk-Jr: ⛏ **nitpick** — `264` duplicates the stylesheet's `max-width: 260px` plus margin. The two will drift the next time the badge width changes; a shared constant would keep them together.

prk-Jr (commented):
## Summary

A well-built, self-contained diagnostics module: bounded memory, exact-ID binding that refuses to guess, ambiguity kept as a first-class outcome, closed shadow DOM, and no mutation of publisher slot elements. No correctness or security defects found, so everything below is a suggestion rather than a blocker.

## Suggestions

### 🤔 thinking

- **Enabling diagnostics ships 30 KB of tester-only code to every visitor**: `tsjs-gpt_diagnostics.js` is 30,343 bytes minified, the largest module in the bundle and 3× core, concatenated synchronously for 100% of traffic while every tab stays inactive until `?ts_console=1`. Immediate loading is the right call and is test-locked; the tension is that `sessionStorage`-only activation gives the server no way to scope the payload to activated sessions. (`crates/trusted-server-core/src/integrations/gpt_diagnostics.rs:56`)
- **Unfiltered childList mutations drive full snapshots and forced layout** on every DOM insertion, amplified by four independent per-frame snapshot consumers. (`badges.ts:183`)
- **`slotOnload` / `impressionViewable` are discarded when GPT omits `isEmpty`**, because both matchers require `isEmpty === false` while the overlay treats `undefined` as pending. (`store.ts:245`, `store.ts:264`)

### ♻️ refactor

- **A missing `CSS.escape` surfaces as `duplicate_dom_id`** in the exported binding reason, reporting a collision that does not exist. (`binding.ts:180`)

### ⛏ nitpick

- **Badge clamp hardcodes the stylesheet width** (`264` vs `max-width: 260px`). (`badges.ts:134`)

### 🌱 seedling

- **Confirmed auction-to-render traceability is the natural next layer.** `slotRenderEnded` already carries `lineItemId`, `campaignId`, `advertiserId` and the `sourceAgnostic*` pair on the same event object the observer reads, and the render bridge already knows when the rendered creative requests markup from Trusted Server. Together those answer whether the line item Trusted Server won is the one that rendered, or whether a house ad, direct-sold line item, or Ad Manager default took the slot — on observed evidence, never inferred from targeting or price. Implemented and opening stacked on this PR; it keeps the export clear of the privacy regex this PR's spec asserts.

### 👍 praise

- The `observed = matched + unmatched + ambiguous` invariant asserted directly in both the unit and browser tests.
- Binding refuses prefix and container-ID guesses, and keeps ambiguity visible instead of picking a winner.
- `referencesUnchanged()` pins the non-invasiveness claim to an assertion — `cmd.push`, `display`, `defineSlot`, `refresh`, `fetch`, `XHR.open`, and the history methods all verified unpatched — and the export is checked against a privacy regex.
- Remount handling for hydration-driven removal, with an explicit host-collision guard.

## CI Status

All 19 checks pass on `79d0832`.

- fmt: PASS
- clippy / cargo check (fastly, axum, cloudflare, spin): PASS
- rust tests (incl. cross-adapter parity, ts CLI): PASS
- js tests (vitest) and format-typescript: PASS
- format-docs: PASS

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts
```diff
@@ -0,0 +1,203 @@
...
+    this.scheduleFrame(() => {
+      this.scheduled = false;
+      this.update();
+    });
+  };
+
+  private installMutationObserver(): void {
+    const Observer = this.window.MutationObserver;
+    if (typeof Observer !== 'function' || !this.document.documentElement) return;
+    this.mutationObserver = new Observer((records) => {
```
aram356, line 228: ♻️ **refactor** — This MutationObserver callback calls `this.store.snapshot()` — a deep copy of up to 64 slots × 10 cycles including derived durations — on every mutation batch matching `childList` or `id/style/class` attribute changes anywhere in the document. It runs even when `this.layer` is undefined (panel hidden), where `update()` then immediately returns, so the snapshot cost is paid for nothing. `BindingManager.refresh()` (binding.ts:109) similarly takes a full snapshot per scroll/resize frame when it only needs `runtimeSlotNumber` + `slotElementId`.

Opt-in tab only, so non-blocking — but on mutation-heavy SPA pages this is real overhead.

**Suggestion**: cache the slot-element-ID set on store notifications (or add a lightweight `slotBindingInputs()` accessor returning just `{ runtimeSlotNumber, slotElementId }`), and early-return in this callback when no layer is set.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
```diff
@@ -0,0 +1,473 @@
...
+    const existingNumber = this.slotNumbers.get(slot);
+    if (existingNumber !== undefined) {
+      const existingRecord = this.slots.get(existingNumber);
+      if (existingRecord) {
+        this.refreshSlotMetadata(existingRecord, slot);
+        return existingRecord;
+      }
+
+      this.incrementDisposition(kind, 'unmatched');
+      this.addIssue(
```
aram356: 🤔 **thinking** — Once a slot record is evicted, the `slotNumbers` WeakMap still maps that Slot object to its old number, so every later callback — including a fresh `slotRequested` that starts a brand-new lifecycle — lands in this branch and is dropped as an `evicted_slot` issue forever. Combined with FIFO-by-creation eviction (`slotOrder.shift()` below evicts the oldest-created slot, not the least recently active one), on a >64-slot page the most active slot can be evicted and never re-enter tracking.

Everything is truthfully counted, so this is not a correctness bug — but it limits usefulness on exactly the large pages diagnostics target. Consider re-registering a record on `slotRequested` for a previously evicted slot, or evicting by least recent activity.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/binding.ts
aram356, line 203: ♻️ **refactor** — When `CSS.escape` is missing (and in the `querySelectorAll` catch below), the binding is reported as `ambiguous` / `duplicate_dom_id` even though no duplicate was actually observed. For a feature whose contract is “report only observed facts,” a distinct `GptDiagnosticsBindingReason` value would keep the export truthful:

```ts
export type GptDiagnosticsBindingReason =
  | 'missing_slot_element_id'
  | 'missing_element'
  | 'duplicate_dom_id'
  | 'duplicate_gpt_slot_id'
  | 'uniqueness_unverifiable';
```

and return `{ status: 'ambiguous', reason: 'uniqueness_unverifiable' }` from these two paths.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts
```diff
@@ -0,0 +1,495 @@
...
+        if (!this.dismissed) this.mount();
+      });
+    });
+  }
+
+  private mount(): void {
+    if (this.destroyed || this.dismissed || !this.visualReady || this.host?.isConnected) return;
+
+    const existing = this.document.getElementById(GPT_DIAGNOSTICS_HOST_ID);
+    if (existing) {
```
aram356, line 277: 🤔 **thinking** — If the publisher page already has an element with id `trusted-server-gpt-diagnostics`, the panel silently never mounts (capture continues, correctly). A `log.warn('gpt diagnostics: host element ID collision, panel not mounted')` here would save a confusing debugging session, and the docs’ troubleshooting section (“The API or panel is absent”) doesn’t cover this case — worth adding a bullet there.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts
```diff
@@ -0,0 +1,495 @@
...
+      this.renderScheduled = false;
+      this.render();
+    });
+  }
+
+  private render(): void {
+    if (!this.panel || !this.host?.isConnected) return;
+    const snapshot = this.store.snapshot();
+    const panel = this.panel;
+    panel.replaceChildren();
```
aram356, line 365: ⛏ **nitpick** — `render()` rebuilds the whole panel on every coalesced update, so an open “Previous requests” `<details>` and the content scroll position reset while events stream in. Fine for v1; preserving open/scroll state (or updating incrementally) would improve usability on busy pages.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts
```diff
@@ -0,0 +1,495 @@
...
+  slot: GptDiagnosticsStoreSlotSnapshot
+): GptDiagnosticsRequestCycle | undefined {
+  return slot.requests[slot.requests.length - 1];
+}
+
+function primaryState(cycle: GptDiagnosticsRequestCycle | undefined): string {
+  if (!cycle) return 'Waiting for request';
+  if (cycle.isEmpty === true) return 'Empty';
+  if (cycle.isEmpty === false) return 'Filled';
+  if (cycle.renderAtMs !== undefined) return 'Pending render';
```
aram356: ⛏ **nitpick** — When `renderAtMs` is set but GPT didn’t expose a boolean `isEmpty`, this label reads “Pending render” although the render already ended. “Rendered (fill unknown)” or similar would be more accurate. Rare path (non-standard `slotRenderEnded` payload).

docs/guide/integrations/gpt-diagnostics.md
```diff
@@ -0,0 +1,191 @@
...
+## Deployment Configuration
+
+The module is unavailable unless explicitly enabled for the deployment:
+
+```toml
+[integrations.gpt_diagnostics]
+enabled = true
+```
+
+Deployment configuration only makes the module available. Every browser tab remains inactive until explicitly activated.
```
aram356: 📝 **note** — The module is deliberately part of the synchronous immediate bundle (necessary for early listener install — the registration test documents this), so enabling `gpt_diagnostics` adds bundle weight for **all** visitors even though virtually no tab ever activates. The performance table in the integrations overview says “Low when active,” but the bundle cost is unconditional. Worth a sentence here so operators treat this as a debug-window toggle rather than an always-on setting.

aram356 (approved):
## Summary

Adds an opt-in, tab-local GPT runtime diagnostics console with a deployment-gated Rust integration, an early activation bootstrap, and a bounded client-side observer/store/binding/overlay stack. The design is disciplined about reporting only GPT-observed facts (conservative unmatched/ambiguous handling instead of guessing), storage is bounded and memory-only, panel rendering uses `textContent` throughout, nothing patches GPT/network/history (asserted by tests), and coverage across Rust, Vitest, and Playwright is thorough. No blocking findings; the inline comments are all non-blocking suggestions.

## Non-blocking

### ♻️ refactor

- **Full store snapshots on every DOM mutation / scroll frame**: `badges.ts` MutationObserver and `BindingManager.refresh()` deep-copy the entire store far more often than needed (crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts:174, binding.ts:109)
- **Misleading binding reason when uniqueness is unverifiable**: reports `duplicate_dom_id` with no observed duplicate (crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/binding.ts:180)

### 🤔 thinking

- **Evicted slots are permanently untrackable**: a fresh `slotRequested` after eviction can never re-enter tracking; FIFO eviction can evict the most active slot (crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts:355)
- **Silent no-mount on host ID collision**: no `log.warn` and no troubleshooting-doc entry (crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts:277)

### ⛏ nitpick

- **Full panel re-render resets details/scroll state** (crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts:355)
- **`primaryState` labels a completed render “Pending render”** (crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts:119)

### 📝 note

- **Enabled deployments ship the diagnostics module to every visitor** in the synchronous bundle; worth an operator-facing sentence in the docs (docs/guide/integrations/gpt-diagnostics.md:25)

## CI Status

- fmt: PASS
- clippy (fastly/axum/cloudflare/cloudflare-wasm/spin-native/spin-wasm): PASS
- rust tests (fastly, axum, cloudflare, spin, parity, CLI): PASS
- js tests (vitest): PASS
- js/docs format: PASS
- integration + browser integration tests: PASS
- CodeQL: PASS

crates/trusted-server-core/src/integrations/gpt_diagnostics.rs
ChristianPavilonis: Implemented in acfad932. Diagnostics is now excluded from the normal unified bundle and is delivered as a standalone synchronous module only for an active browser session. Exact ts_console directives establish or clear the host-only HttpOnly __Host-ts-console session cookie; active/directive HTML is private/no-store, while the standalone asset remains cookie-independent and publicly cacheable. Browser integration coverage verifies inactive delivery, activation, persistence, and deactivation.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts
ChristianPavilonis: Implemented in acfad932. Binding and badge mutation observers now cache lightweight slot binding inputs, ignore unrelated child-list mutations, handle relevant ID/subtree replacements, and skip badge work while hidden. The focused binding/badge tests cover irrelevant mutations, replacements, ID changes, and coalescing.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
ChristianPavilonis: Implemented in acfad932. slotOnload and impressionViewable now match rendered cycles when isEmpty is not true, while known-empty cycles remain ineligible. The store retains the observed timestamps without inferring Filled, and the lifecycle tests cover this distinction.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/binding.ts
ChristianPavilonis: Implemented in acfad932. Added the exported dom_uniqueness_unverifiable reason for missing/throwing CSS.escape or selector evaluation; duplicate_dom_id is now reserved for a duplicate actually observed by the selector check. Focused binding tests cover both failure paths.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts
ChristianPavilonis: Implemented in acfad932. Badge width and edge-gutter values are now named constants, with the generated badge max-width and positioning clamp derived from the same values.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts
ChristianPavilonis: Implemented in acfad932. The mutation paths now use cached lightweight binding inputs instead of deep snapshots, filter mutations to relevant slot IDs/subtrees, and return before doing work when the badge layer is hidden. Existing detached public snapshot behavior remains unchanged.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
ChristianPavilonis: Implemented in acfad932. Retention now evicts the least-recently-active slot, and an evicted Slot can re-enter only on a later slotRequested. Non-request callbacks remain evicted_slot unmatched, and weak per-slot request counters preserve monotonic request numbering. Store tests cover boundedness, re-entry, and coverage accounting.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/binding.ts
ChristianPavilonis: Implemented in acfad932 with the same truthful binding distinction: selector verification failures return dom_uniqueness_unverifiable, while proven duplicate IDs still return duplicate_dom_id. Tests cover CSS.escape and querySelectorAll failures.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts
ChristianPavilonis: Implemented in acfad932. Host-ID collisions now emit a one-time warning per collision episode, and the troubleshooting guide explains that capture/API continue while the visual panel remains unmounted until the reserved ID is removed.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts
ChristianPavilonis: Implemented in acfad932. Full panel rebuilds now preserve open Previous requests disclosures keyed by runtime slot and restore content scrollTop. Overlay tests cover state preservation during live updates.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts
ChristianPavilonis: Implemented in acfad932. A completed render with no boolean isEmpty now displays Rendered (fill unknown), and badges/docs use the same factual wording.

docs/guide/integrations/gpt-diagnostics.md
ChristianPavilonis: Implemented in acfad932. The documentation now describes conditional standalone delivery, browser-session activation, strict active/directive HTML cache policy, and the fact that inactive visitors do not receive diagnostics JavaScript.

PR #978, opened 2026-07-29
Select active GPT responsive slots

### LINKED ISSUE #977: Fix GPT display gating for responsive slots
## Problem

The GPT initial-request wrapper assumes `googletag.display()` receives a string element ID. GPT also accepts DOM elements and Slot objects. If a publisher passes a Slot before the initial request gate is released, the wrapper calls string methods on the object and aborts the display request.

Stable configured slot prefixes also resolve to the first DOM match. Responsive pages can expose multiple mutually exclusive siblings for one prefix, causing initial targeting to attach to a hidden sibling.

## Expected behavior

- Preserve GPT's `display(string | Element | Slot)` contract.
- Use a normalized element ID only for internal gate and handoff bookkeeping.
- Resolve a stable prefix to the unique active, laid-out element.
- Fail closed when a prefix is empty or ambiguous.
- Keep the injected bootstrap and runtime bundle behavior aligned.

## Acceptance criteria

- Slot and Element display targets do not throw before gate release.
- Held displays replay once after initial targeting is applied.
- Released displays preserve their original argument.
- Responsive prefix resolution does not select a hidden sibling.
- Runtime and bootstrap implementations have regression coverage.

ChristianPavilonis:
## Summary

- Resolve stable GPT slot prefixes to the unique active responsive sibling instead of the first matching DOM element.
- Preserve exact-ID and unique lazy-slot behavior while failing closed when multiple prefix matches have no unique visible, laid-out candidate.
- Keep the edge bootstrap and runtime implementation aligned with shared regression coverage.

This PR is stacked on #966, which already preserves GPT's `display(string | Element | Slot)` contract in the permanent slot-handoff implementation. It does not depend on #974.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/integrations/gpt_bootstrap.js` | Select a unique active responsive prefix match before defining or reusing a GPT slot. |
| `crates/trusted-server-js/lib/src/integrations/gpt/index.ts` | Mirror responsive slot resolution in the runtime integration and fail closed on ambiguity. |
| `crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts` | Cover runtime/bootstrap parity for element and container layout, hidden siblings, ambiguity, empty IDs, and publisher-owned slots. |
| `crates/trusted-server-core/src/integrations/gpt.rs` | Update the injected-bootstrap source assertion for the shared resolver. |

## Closes

Closes #977

## Test plan

- [x] `cargo clippy-cloudflare-wasm && cargo clippy-spin-native && cargo clippy-spin-wasm`
- [x] JS tests: `cd crates/trusted-server-js/lib && npx vitest run` (430 passed)
- [x] JS lint and format: `npm run lint && npm run format`
- [x] JS production build: `npm run build`
- [x] Fastly WASM release build: `cargo build --package trusted-server-adapter-fastly --release --target wasm32-wasip1`

## Checklist

- [x] New code has tests


crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -78,6 +78,63 @@
...
+        !candidate.id.endsWith("-container")
+      ) {
+        prefixMatches.push(candidate);
+      }
+    }
+    // A unique prefix match may be a lazy slot that has not been sized yet.
+    // Geometry is only needed to disambiguate multiple responsive siblings.
+    if (prefixMatches.length === 1) return prefixMatches[0];
+
+    var activeMatches = prefixMatches.filter(slotElementHasLayout);
```
prk-Jr: 🔧 **wrench** — Geometry is required before any sibling can be picked, so the single *visible* responsive sibling is skipped whenever it reserves no space.

`slotElementHasLayout` needs `width > 0 && height > 0` on the element or on `<id>-container`. An empty ad div awaiting its creative commonly computes to height 0, and its empty wrapper does too. With several prefix matches, the active sibling then fails the check and `activeMatches.length === 0` → this function returns `null` → `adInit` does `if (!el) return`, so nothing is defined and no ad request is made. Before this PR the first prefix match was defined and requested, so this trades a mis-targeted impression for no impression at all.

Reproduced against both implementations with a throwaway probe — three `display:none` dupes plus one visible sibling, default 0×0 rects:

```
GPT slot prefix did not resolve to one active element { divId: 'ad-probe-', prefixMatchCount: 4, activeMatchCount: 0 }
→ defineSlot not called, divToSlotId {}
```

That is exactly the `not-size-a/b/c/d` shape this PR targets: inactive siblings are hidden with `display:none`, so visibility alone already disambiguates them and geometry is only needed when two or more siblings are visible.

**Fix** — add a visibility tier before the geometry tier:

```js
    var visibleMatches = prefixMatches.filter(isElementVisible);
    if (visibleMatches.length === 1) return visibleMatches[0];

    var activeMatches = visibleMatches.filter(slotElementHasLayout);
    if (activeMatches.length === 1) return activeMatches[0];
```

Filtering the geometry pass over `visibleMatches` (not `prefixMatches`) also keeps the hidden-sibling case unchanged, since `slotElementHasLayout` already starts with `isElementVisible`.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -52,15 +52,47 @@ interface SlotRenderEndedEvent {
...
-      (el) => el.id.startsWith(divId) && !el.id.endsWith('-container')
-    ) ?? null
+  const prefixMatches = Array.from(document.querySelectorAll<HTMLElement>('[id]')).filter(
+    (element) => element.id.startsWith(divId) && !element.id.endsWith('-container')
   );
+  // A unique prefix match may be a lazy slot that has not been sized yet.
+  // Geometry is only needed to disambiguate multiple responsive siblings.
+  if (prefixMatches.length === 1) return prefixMatches[0] ?? null;
+
+  const activeMatches = prefixMatches.filter(slotElementHasLayout);
```
prk-Jr: 🔧 **wrench** — Same geometry-gate issue as `gpt_bootstrap.js:121`; the runtime needs the matching change so the two stay in parity.

```ts
  const visibleMatches = prefixMatches.filter(isElementVisible);
  if (visibleMatches.length === 1) return visibleMatches[0] ?? null;

  const activeMatches = visibleMatches.filter(slotElementHasLayout);
  if (activeMatches.length === 1) return activeMatches[0] ?? null;
```

Worth extending the `it.each` table with a `hiddenElementIndexes: [0, 1, 3]` + `activeIndexes: []` case so the intended contract ("one visible sibling wins even with no reserved height") is pinned in both implementations.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -78,6 +78,63 @@
...
+    if (activeMatches.length === 1) return activeMatches[0];
+
+    if (
+      prefixMatches.length > 1 &&
+      ts.log &&
+      typeof ts.log.warn === "function"
+    ) {
+      ts.log.warn("GPT slot prefix did not resolve to one active element", {
```
prk-Jr: 🤔 **thinking** — This warn can never fire: `window.tsjs.log` is declared optional on `TsjsApi` (`core/types.ts:97`) but is never assigned anywhere in the repo, so `ts.log && typeof ts.log.warn === "function"` is always false. The probe above confirms it — the bundle path logged, the bootstrap path was silent.

That matters more here than for the pre-existing handoff warn on line 180: the bootstrap is the first-load path, and it is the only place that can report why a placement resolved to nothing. As written, an ambiguous prefix produces a blank slot with no signal anywhere.

Either fall back to `console.warn`, or record a counter the bundle can surface later, e.g.:

```js
      ts.gptSlotResolveFailures = (ts.gptSlotResolveFailures || []).concat({
        divId: divId,
        prefixMatchCount: prefixMatches.length,
        activeMatchCount: activeMatches.length,
      });
```

crates/trusted-server-core/src/integrations/gpt.rs
```diff
@@ -1213,7 +1213,7 @@ mod tests {
             "bootstrap should scan ID-bearing elements instead of interpolating div_id into CSS"
         );
         assert!(
-            combined.contains(".startsWith(slot.div_id)"),
+            combined.contains("candidate.id.startsWith(divId)"),
```
prk-Jr, line 1256: 🌱 **seedling** — Not for this PR: the resolver now exists twice (`gpt_bootstrap.js` and `gpt/index.ts`) and the only thing tying them together is this substring assertion plus the shared `it.each` table. The assertion pins one identifier, so a divergence in the tiering logic itself (the exact thing this PR changes) would pass. Worth a follow-up to generate the bootstrap from the TS source, or at least to assert on the tier names (`activeMatches` / `prefixMatches`) as well.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -52,15 +52,47 @@ interface SlotRenderEndedEvent {
...
+  if (elementRect.width > 0 && elementRect.height > 0) return true;
+
+  const container = document.getElementById(`${element.id}-container`);
+  if (!container || !isElementVisible(container)) return false;
+  const containerRect = container.getBoundingClientRect();
+  return containerRect.width > 0 && containerRect.height > 0;
+}
+
 function findSlotElementByDivId(divId: string): HTMLElement | null {
+  if (!divId) return null;
```
prk-Jr: 👍 **praise** — This fixes a real wildcard bug, not just an edge case. Previously an empty configured `div_id` fell through `getElementById('')` to `el.id.startsWith('')`, which matches every element with an id, so the slot bound to whatever `[id]` element happened to come first in the document. Good catch, and good that the table covers it for both implementations.

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
```diff
@@ -1257,6 +1295,113 @@ describe('installTsAdInit', () => {
     expect(mockPubads.refresh).toHaveBeenCalled();
   });
 
+  it.each([
```
prk-Jr, line 2209: 👍 **praise** — Driving both the bundle and the edge bootstrap from one case list (with the bootstrap read straight off `gpt_bootstrap.js`) is the right way to test two hand-maintained copies of the same resolver. It is what made the geometry-gate finding above verifiable in both paths from a single fixture.

prk-Jr (changes requested):
## Summary

Replaces first-match-in-document-order prefix resolution with "unique active prefix match", failing closed on ambiguity, and mirrors it across the edge bootstrap and the bundle with a shared test table. The direction is right and the parity coverage is good. One blocking issue: the resolver requires geometry before it will pick a sibling, so the single visible responsive sibling is still skipped when it reserves no space — and unlike before this PR, nothing is defined or requested at all in that case.

## Blocking

### 🔧 wrench

- **Geometry gate drops the one visible responsive sibling when it reserves no space**: `slotElementHasLayout` requires `width > 0 && height > 0` on the element or its `-container`. An empty ad div awaiting its creative commonly has height 0, and its empty wrapper does too, so the active sibling fails the check and `activeMatches.length === 0` → `null` → the slot is skipped entirely (no `defineSlot`, no ad request). Before this PR the first prefix match was defined and requested. Verified against both implementations with a throwaway probe (three `display:none` dupes plus one visible sibling, default 0×0 rects):

  ```
  GPT slot prefix did not resolve to one active element { divId: 'ad-probe-', prefixMatchCount: 4, activeMatchCount: 0 }
  → defineSlot not called, divToSlotId {}
  ```

  Adding a visibility-only tier before the geometry tier resolves the `not-size-a/b/c/d` pattern this PR targets without requiring reserved height. Details inline on `gpt_bootstrap.js:121` and `gpt/index.ts:85`.

## Non-blocking

### 🤔 thinking

- **Forced synchronous layout inside the SPA MutationObserver** (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts:897`): `waitForSlotElements` calls `allPresent()` on every mutation batch, and `findSlotElementByDivId` now runs `getComputedStyle` + `getBoundingClientRect` per prefix candidate whenever a prefix has more than one match. During framework hydration that is a forced reflow per batch. The presence check only needs existence, not activeness — consider a geometry-free probe there (or a resolver flag that stops at the prefix-match tier).
- **Ambiguity warn never fires from the edge bootstrap** — see inline on `gpt_bootstrap.js:129`.

### 🌱 seedling

- **Resolver logic now exists twice** (edge bootstrap JS and bundle TS), kept aligned by a substring assertion in `gpt.rs`. Worth a follow-up to generate the bootstrap from one source — see inline on `gpt.rs:1216`.

### ⛏ nitpick

- Test table covers ambiguous, none-active, hidden-sibling and empty-`div_id` cases, but not "exactly one visible sibling, nothing laid out" — the case that motivates the blocking finding — and nothing asserts the warn payload (`prefixMatchCount` / `activeMatchCount`), which is the only diagnostic for the fail-closed path.

### 👍 praise

- Empty-`div_id` guard and the runtime/bootstrap parity table — see inline on `gpt/index.ts:74` and `ad_init.test.ts:1298`.

## CI Status

Only the Integration Tests workflow ran on `340d1efb4` (4 checks, all pass). `format.yml` and `test.yml` trigger on `pull_request: branches: [main, "feature/**"]`, so with this PR based on `fix/duplicate-gpt-slots` the fmt / clippy / Rust test / JS test gates did not run on CI. Verified locally on the head commit instead:

- fmt: NOT RUN on CI
- clippy: NOT RUN on CI
- rust tests: PASS locally (`cargo test -p trusted-server-core --target aarch64-apple-darwin gpt` — 31 passed)
- js tests: PASS locally (`npx vitest run` — 30 files, 430 passed)
- integration tests: PASS on CI

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
ChristianPavilonis: Implemented in ca678fe6. Both resolvers now choose a unique visible prefix match before consulting geometry, so a visible zero-size responsive slot is defined and displayed. The geometry tier remains for multiple visible siblings.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Implemented in ca678fe6 in runtime parity with the edge bootstrap. The shared table now covers three hidden siblings plus one visible sibling whose element and container have no layout, for both implementations.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
ChristianPavilonis: I verified the production ordering before changing this. `html_processor.rs` injects the synchronous unified bundle after the bootstrap and before the end-of-body bid script invokes `ts.adInit`; `core/index.ts` assigns `api.log = log`. Therefore the bootstrap resolver has the logger when it runs in production, so no fallback was added.

crates/trusted-server-core/src/integrations/gpt.rs
ChristianPavilonis: Agreed this is a maintainability risk. I kept generation/stronger structural parity checks out of this focused bug fix, as suggested; the shared behavioral table remains the regression guard for this PR.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Thank you.

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
ChristianPavilonis: Thank you.

prk-Jr (approved):
The previously blocking geometry-gate issue is resolved in ca678fe6 with the visibility-first selection tier and runtime/bootstrap parity coverage. No remaining blocking or non-blocking findings.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -52,15 +52,50 @@ interface SlotRenderEndedEvent {
...
+
+function slotElementHasLayout(element: HTMLElement): boolean {
+  if (!isElementVisible(element)) return false;
+  const elementRect = element.getBoundingClientRect();
+  if (elementRect.width > 0 && elementRect.height > 0) return true;
+
+  const container = document.getElementById(`${element.id}-container`);
+  if (!container || !isElementVisible(container)) return false;
+  const containerRect = container.getBoundingClientRect();
+  return containerRect.width > 0 && containerRect.height > 0;
```
aram356: ❓ **question** — Fail-closed drops the impression entirely when no candidate has reserved height.

`slotElementHasLayout` requires `width > 0 && height > 0` on the element or its `-container`. An empty ad div has height 0 until an ad renders, unless the publisher reserves space with `min-height`. In the container-based responsive-hiding pattern this PR targets (inactive containers `display:none`), every inner element still passes the element-visibility check (computed `display` does not inherit from ancestors), so resolution always falls through to geometry — and with no reserved height, `activeMatches` is empty and the slot is skipped entirely, where the old code served the first match.

Is total no-fill acceptable for publishers that don't reserve slot heights? A width-only container check would disambiguate without requiring reserved height, since a `display:none` container has a 0×0 rect while the active one keeps its block width:

```ts
const containerRect = container.getBoundingClientRect();
return containerRect.width > 0;
```

Same applies to the mirrored `slotElementHasLayout` in `crates/trusted-server-core/src/integrations/gpt_bootstrap.js`.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -52,15 +52,50 @@ interface SlotRenderEndedEvent {
...
-    ) ?? null
+  const prefixMatches = Array.from(document.querySelectorAll<HTMLElement>('[id]')).filter(
+    (element) => element.id.startsWith(divId) && !element.id.endsWith('-container')
   );
+  // A unique prefix match may be a lazy slot that has not been sized yet.
+  // Geometry is only needed to disambiguate multiple responsive siblings.
+  if (prefixMatches.length === 1) return prefixMatches[0] ?? null;
+
+  const visibleMatches = prefixMatches.filter(isElementVisible);
+  if (visibleMatches.length === 1) return visibleMatches[0] ?? null;
```
aram356: 🤔 **thinking** — Element-level visibility can't see container-hidden siblings.

`isElementVisible` reads the element's own computed style, and `display` doesn't inherit — an element inside a `display:none` container still counts as visible. When no sibling is actually active and exactly one is hidden only at container level, this `visibleMatches.length === 1` shortcut returns that hidden element instead of failing closed, defining a GPT slot in an invisible subtree (unviewable impression).

`element.checkVisibility?.()` checks ancestor display for exactly this case and could be layered in with the current heuristic as fallback:

```ts
function isElementVisible(element: HTMLElement): boolean {
  if (typeof element.checkVisibility === 'function') return element.checkVisibility();
  const style = window.getComputedStyle(element);
  // existing checks…
}
```

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -52,15 +52,50 @@ interface SlotRenderEndedEvent {
...
+
+  const activeMatches = visibleMatches.filter(slotElementHasLayout);
+  if (activeMatches.length === 1) return activeMatches[0] ?? null;
+
+  if (prefixMatches.length > 1) {
+    log.warn('GPT slot prefix did not resolve to one active element', {
```
aram356: ♻️ **refactor** — Ambiguity warning fires on every animation frame during the SPA slot wait.

`waitForSlotElements`' `allPresent()` calls `findSlotElementByDivId` per slot per mutation-batched frame for up to 2 s. An ambiguous prefix emits this `log.warn` on every call — potentially ~120 duplicate console warns per navigation. Move the warn to the `adInit` resolution site, or add a warn-once guard keyed on `divId`.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -52,15 +52,50 @@ interface SlotRenderEndedEvent {
...
+      divId,
+      prefixMatchCount: prefixMatches.length,
+      activeMatchCount: activeMatches.length,
+    });
+  }
+  return null;
 }
 
 function candidateSlotRoots(divId: string): HTMLElement[] {
```
aram356: 🤔 **thinking** — Render bridge re-resolves the prefix at message time and can disagree with adInit.

`candidateSlotRoots` re-runs the geometry-sensitive resolver when the Prebid Request message arrives. If layout shifted since `adInit` (hydration, rotation), or the prefix has become ambiguous by then, resolution returns a different element or null, `slotIdForMessageSource` misses the iframe, and the TS creative and win/billing beacons are silently dropped.

`adInit` already records the resolved ID in `ts.divToSlotId` — deriving candidate roots from that map would keep both sites consistent with the element adInit actually chose.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -869,16 +904,26 @@ function waitForSlotElements(slots: AuctionSlot[], signal: AbortSignal): Promise
...
       resolve();
     };
     const observer = new MutationObserver(() => {
-      if (allPresent()) finish();
+      if (animationFrame !== undefined) return;
+      if (typeof requestAnimationFrame === 'undefined') {
+        if (allPresent()) finish();
+        return;
+      }
+      animationFrame = requestAnimationFrame(() => {
```
aram356, line 1198: 🤔 **thinking** — rAF gating always costs the full 2 s in background tabs.

`requestAnimationFrame` doesn't fire in hidden tabs, so a background-tab SPA navigation never runs the observer check and applies bids only at the `SPA_SLOT_WAIT_MS` timeout. Minor, but a `document.visibilityState === 'hidden'` branch that checks `allPresent()` directly would restore prompt application.

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
```diff
@@ -53,6 +53,44 @@ type TestWindow = Window & {
...
+    ({
+      width: elementHasLayout ? 300 : 0,
+      height: elementHasLayout ? 250 : 0,
+    }) as DOMRect;
+  container.appendChild(element);
+  document.body.appendChild(container);
+  return element;
+}
+
+function runGptBootstrap(): void {
```
aram356, line 150: 🏕 **camp site** — Reuse this helper in the older bootstrap test.

The pre-existing handoff test ('runs the embedded bootstrap handoff for a hydrated publisher ID', ~line 516) still inlines the identical `readFileSync` + `window.eval` that `runGptBootstrap()` wraps.

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
```diff
@@ -53,6 +53,44 @@ type TestWindow = Window & {
   tsjs?: any;
 };
 
+function appendResponsiveSlotElement(
```
aram356, line 117: ⛏ **nitpick** — `appendResponsiveSlotElement` takes five positional booleans.

Call sites like `(id, true, false, true, false)` are hard to read; an options object (`{ containerHasLayout, elementHidden, … }`) would make the parameterized cases self-describing. Test-only.

aram356 (changes requested):
## Summary

Replaces first-prefix-match GPT slot resolution with a visibility/geometry ladder (exact ID → unique prefix → unique visible → unique laid-out → fail closed), mirrored in the edge bootstrap and runtime bundle with a shared parity test suite. The direction is right and the empty-`div_id` guard fixes a real pre-existing bug; the open question is whether the geometry gate's reserved-height requirement can drop impressions for publishers that don't reserve slot heights.

## Blocking

### ❓ question

- **Fail-closed drops the impression when no candidate has reserved height**: `slotElementHasLayout` requires height > 0 on the element or its `-container`; empty ad divs without `min-height` never qualify, so container-hidden responsive setups without reserved space go from "first match" to no fill at all (crates/trusted-server-js/lib/src/integrations/gpt/index.ts:70, mirrored in crates/trusted-server-core/src/integrations/gpt_bootstrap.js:98). See inline comment for a width-only alternative.

## Non-blocking

### 🤔 thinking

- **Container-hidden siblings count as visible**: computed `display` doesn't inherit, so the `visibleMatches.length === 1` shortcut can return an element inside a `display:none` container (index.ts:86). `Element.checkVisibility()` would cover this.
- **Render bridge re-resolves the prefix at message time**: `candidateSlotRoots` can disagree with the element `adInit` chose (or return null on later ambiguity), silently dropping the TS creative and beacons; consider deriving roots from `ts.divToSlotId` (index.ts:101).
- **rAF gating in background tabs**: hidden tabs never run the observer check, so background SPA navigations always pay the full 2 s `SPA_SLOT_WAIT_MS` (index.ts:923).

### ♻️ refactor

- **Ambiguity warn fires per animation frame during the SPA slot wait**: up to ~120 duplicate warns per navigation; warn once per divId or only from adInit (index.ts:92).

### 🏕 camp site

- **Reuse `runGptBootstrap()`** in the pre-existing embedded-bootstrap handoff test that still inlines readFileSync + eval (ad_init.test.ts:516).

### ⛏ nitpick

- **`appendResponsiveSlotElement` five positional booleans**: an options object would make the it.each cases self-describing (ad_init.test.ts:56).

## CI Status

- GitHub checks on head commit: integration tests, browser integration tests, Fastly EC lifecycle, prepare artifacts — all PASS. The fmt/clippy/unit-test/JS workflows did not trigger for this stacked base.
- Verified locally: JS tests 432/432 PASS; eslint PASS; `trusted-server-core` gpt tests 31/31 PASS on wasm32-wasip1 via Viceroy. (Prettier warn on `src/core/templates/iframe.html` is pre-existing on the base branch, untouched here.)

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Implemented in a5cba2b5. A visible responsive container with width but no reserved height now disambiguates the slot; the runtime and edge bootstrap use the same width-only fallback. Added regression coverage for zero-height responsive containers in both implementations.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Implemented in a5cba2b5. Visibility now checks ancestors explicitly and uses checkVisibility() when available, so a unique element inside a display:none container is rejected. The resolver tests cover this hidden-container case for both runtime paths.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Implemented in a5cba2b5. Ambiguity logging was removed from the reusable resolver and moved to adInit, with one warning per divId per initialization. The regression test verifies duplicate slot entries produce only one warning.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Implemented in a5cba2b5. The render bridge now derives roots from ts.divToSlotId, preserving the element selected by adInit even if the responsive prefix later becomes ambiguous. Added a regression test covering that layout change.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Implemented in a5cba2b5. Hidden documents now run the slot-presence check directly instead of waiting for requestAnimationFrame, with a regression test covering unavailable/throttled rAF behavior.

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
ChristianPavilonis: Implemented in a5cba2b5. The existing embedded-bootstrap handoff test now reuses runGptBootstrap().

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
ChristianPavilonis: Implemented in a5cba2b5. appendResponsiveSlotElement now accepts a named options object, including independent container/element visibility and geometry fields; all responsive test cases use the clearer form.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -688,11 +760,23 @@ export function installTsAdInit(): void {
...
-        const el = findSlotElementByDivId(slot.div_id);
-        if (!el) return;
+        // Resolve actual div ID: exact match first, then the visibility and
+        // geometry tiers for prefix matches. div_id in config may be a stable
+        // prefix (e.g. "ad-header-0-") when the suffix is dynamically
+        // generated by the framework at render time.
+        const resolution = resolveSlotElementByDivId(slot.div_id);
+        const el = resolution.element;
+        if (!el) {
+          if (resolution.prefixMatchCount > 1 && !warnedResolutionFailures.has(slot.div_id)) {
```
aram356, line 917: ♻️ **refactor** — A unique-but-hidden prefix match is skipped silently, below the warn gate.

The warn fires only for `prefixMatchCount > 1`. The tightened unique-match tier means exactly-one-match-but-hidden (e.g. the `#slot:empty { display: none }` CLS pattern, or consent-gated CSS hiding the slot at adInit time) now goes from served (pre-PR first-match) to no-fill with zero diagnostics.

Warning on `prefixMatchCount > 0` would keep the genuinely-absent case (count 0, the normal SPA pre-commit skip) silent while making the hidden-unique case debuggable in the field. Same gate in `gpt_bootstrap.js` (line 308).

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -52,29 +52,100 @@ interface SlotRenderEndedEvent {
...
+    checkVisibility?: () => boolean;
+  };
+  if (
+    typeof elementWithVisibilityCheck.checkVisibility === 'function' &&
+    !elementWithVisibilityCheck.checkVisibility()
+  ) {
+    return false;
+  }
+
+  for (let current: HTMLElement | null = element; current; current = current.parentElement) {
```
aram356, line 75: ♻️ **refactor** — The manual ancestor walk can override `checkVisibility()`'s correct answer.

When the API exists and returns true, the walk still runs — and for an element that re-enables `visibility: visible` under a `visibility: hidden` ancestor (visible in real browsers), the per-ancestor own-computed-style check wrongly reports hidden, so modern browsers get a worse answer than their native API gives.

Consider making the walk fallback-only:

```ts
if (typeof elementWithVisibilityCheck.checkVisibility === 'function') {
  return elementWithVisibilityCheck.checkVisibility({
    checkVisibilityCSS: true,
    visibilityProperty: true,
  });
}
// getComputedStyle ancestor walk only when the API is absent…
```

(both option spellings for older Chromium; unknown dictionary members are ignored). Also trims the O(ancestors) `getComputedStyle` cost per candidate per frame during the SPA slot wait. Mirror in `gpt_bootstrap.js`.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -869,16 +953,26 @@ function waitForSlotElements(slots: AuctionSlot[], signal: AbortSignal): Promise
...
       settled = true;
+      if (animationFrame !== undefined) cancelAnimationFrame(animationFrame);
       observer.disconnect();
       clearTimeout(timer);
       signal.removeEventListener('abort', finish);
       resolve();
     };
     const observer = new MutationObserver(() => {
-      if (allPresent()) finish();
+      if (animationFrame !== undefined) return;
```
aram356: ⛏ **nitpick** — A pending rAF pins the hidden-tab path to the 2 s timer.

This guard runs before the visibility check, so a frame scheduled just before the tab hides never fires, and every later mutation early-returns — the direct hidden-document check below becomes unreachable until the timeout. Checking visibility first and cancelling the pending frame closes the gap:

```ts
if (document.visibilityState === 'hidden' || typeof requestAnimationFrame === 'undefined') {
  if (animationFrame !== undefined) {
    cancelAnimationFrame(animationFrame);
    animationFrame = undefined;
  }
  if (allPresent()) finish();
  return;
}
if (animationFrame !== undefined) return;
```

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
```diff
@@ -1257,6 +1298,273 @@ describe('installTsAdInit', () => {
     expect(mockPubads.refresh).toHaveBeenCalled();
   });
 
+  it.each([
```
aram356, line 2209: ⛏ **nitpick** — The `checkVisibility` branch is untested.

jsdom lacks the API, so every case here exercises only the fallback ancestor walk. One parity case stubbing `checkVisibility` (e.g. returning false on an otherwise-visible candidate) would cover the modern-browser path in both implementations.

aram356 (approved):
## Summary

The hardening commit resolves all findings from the previous review: width-only container geometry, ancestor-aware visibility with `checkVisibility()`, warn-once diagnostics at the adInit call site, render bridge keyed off `ts.divToSlotId`, and a hidden-document fallback in the SPA slot wait — with parity coverage for both implementations. Remaining comments are non-blocking refinements.

## CI Status

- GitHub checks had not run on this head at review time.
- Verified locally: JS tests 440/440 PASS; eslint PASS; `trusted-server-core` gpt tests 31/31 PASS on wasm32-wasip1 via Viceroy.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Thanks. I considered warning for any nonzero prefix match, but left the threshold unchanged intentionally: a unique hidden element can be an expected responsive or consent-gated no-fill, so warning in that case could be noisy. Ambiguous multi-match failures remain warn-once diagnostics. I am deferring broader hidden-slot observability and resolving this as intentionally deferred.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Implemented in 6b1dd6fa. The hidden-document check now runs before the pending-frame guard, cancels a stale rAF, and checks slot presence immediately. Added a regression test for a frame scheduled before the tab becomes hidden.

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
ChristianPavilonis: Implemented in 6b1dd6fa. Added a runtime/bootstrap parity case that stubs checkVisibility in ad_init.test.ts and verifies native visibility is honored.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Implemented in 6b1dd6fa. Both resolver copies now treat checkVisibility({ checkVisibilityCSS: true, visibilityProperty: true }) as authoritative and use the manual ancestor walk only when the API is unavailable. Added a runtime/bootstrap parity regression covering the native API path.

PR #984, opened 2026-07-31
Lint the full JS package in CI, not just src

### LINKED ISSUE #985: CI lint gate misses JS test files and Node build scripts
The JS lint gate in CI (`npm run lint` in `format.yml`) only covers `src/**/*.{ts,tsx}` because of the lint script's glob in `crates/trusted-server-js/lib/package.json`. ESLint errors in `test/**`, the Node build scripts (`build-all.mjs`, `build-prebid-external.mjs`), and config files are invisible to CI — `npx eslint .` on `main` reports 280 accumulated errors:

- 249 × `@typescript-eslint/no-explicit-any` in test files (the established `(window as any)` mock convention)
- 16 × `no-undef` in the `.mjs` Node scripts (`process`/`console`/`Buffer` — no Node globals declared for them)
- 10 × `import/order`, 4 × `@typescript-eslint/no-unused-vars` (intentionally unused `_`-prefixed params), 1 × `prefer-const`

New lint errors in these files land unnoticed.

**Fix**: widen the lint script to `eslint .` and make the widened surface clean — disable `no-explicit-any` for tests (matching the existing convention), declare Node globals for the `.mjs` scripts, honor the `^_` unused-arg convention, and fix the mechanical remainder. No workflow change needed since `format.yml` already runs `npm run lint`.

aram356:
Fixes #985.

## Problem

The CI lint gate (`npm run lint` in format.yml) only covered `src/**/*.{ts,tsx}`, so eslint errors in test files, the Node build scripts, and config files were invisible: 280 of them had accumulated (`npx eslint .` on main), and new ones land unnoticed.

## Fix

Widen the lint script to `eslint . --max-warnings=0` and make the widened surface clean, with no rule exceptions:

- **Tests**: all 249 `no-explicit-any` violations are fixed with real types rather than suppressed. Window mocks are typed as `Omit<Window, 'tsjs'>` intersections with `Partial<TsjsApi>` from `src/core/types`; the prebid hoisted mock gets structural `TestPbjs`/`TestAdUnit`/`TestBid` types; many casts were simply unnecessary and are deleted (untyped `vi.fn()` already assigns to `typeof fetch`). One documented `as unknown as` double-cast remains, for a test that deliberately feeds malformed input. Side effect: repo-wide `tsc --noEmit` errors drop from 50 to 12, all remaining ones pre-existing in files this PR does not touch.
- **Node scripts**: declare ESM-safe Node globals (`globals.nodeBuiltin`) for `*.mjs` build scripts and Node-run test harnesses, so `process`/`console`/`Buffer` resolve but CommonJS-only names (`__dirname`, `require`) still fail `no-undef`.
- **Unused args**: honor the `^_` underscore convention in `@typescript-eslint/no-unused-vars` via a top-level rules block covering every linted file, not just `*.ts(x)`.
- **Warnings**: `--max-warnings=0` so rules landing at `warn` level cannot pass CI silently; this also fails on unused eslint-disable directives, keeping the remaining `src/` directives honest.
- **Mechanical fixes**: `import/order` in five files, one `prefer-const` (`mockPbjs` declaration merged with its single assignment).

No workflow changes: format.yml already runs `npm run lint`; the script now covers the whole package.

## Verification

- `npx eslint . --max-warnings=0` exits 0
- `npm run format` clean
- `npx vitest run`: 488 passed, 0 failed (32 files)
- `npx tsc --noEmit`: no new errors vs main (50 -> 12)

### DIFF
--- a/crates/trusted-server-js/lib/eslint.config.js
+++ b/crates/trusted-server-js/lib/eslint.config.js
@@ -35,17 +35,26 @@ export default [
       'unicorn/prevent-abbreviations': 'off',
       'unicorn/filename-case': 'off',
       'import/order': ['error', { 'newlines-between': 'always' }],
+    },
+  },
+  // Honor the `_`-prefix convention for intentionally unused bindings in every
+  // linted file — tseslint recommended enables the rule globally, so scoping
+  // this to *.ts(x) would leave .mjs at the pattern-less defaults
+  {
+    rules: {
       '@typescript-eslint/no-unused-vars': [
         'error',
         { argsIgnorePattern: '^_', varsIgnorePattern: '^_', caughtErrorsIgnorePattern: '^_' },
       ],
     },
   },
-  // Node build scripts and Node-run test harnesses
+  // Node build scripts and Node-run test harnesses — nodeBuiltin, not node,
+  // so CommonJS-only names (__dirname, require, module) still fail no-undef
+  // in these ES modules
   {
     files: ['*.mjs', 'test/**/*.mjs'],
     languageOptions: {
-      globals: globals.node,
+      globals: globals.nodeBuiltin,
     },
   },
   // Tests routinely poke at private state and mock boundaries via `any`
--- a/crates/trusted-server-js/lib/eslint.config.js
+++ b/crates/trusted-server-js/lib/eslint.config.js
@@ -57,11 +57,4 @@ export default [
       globals: globals.nodeBuiltin,
     },
   },
-  // Tests routinely poke at private state and mock boundaries via `any`
-  {
-    files: ['test/**/*.ts', 'test/**/*.tsx'],
-    rules: {
-      '@typescript-eslint/no-explicit-any': 'off',
-    },
-  },
 ];
--- a/crates/trusted-server-js/lib/eslint.config.js
+++ b/crates/trusted-server-js/lib/eslint.config.js
@@ -38,7 +38,7 @@ export default [
     },
   },
   // Honor the `_`-prefix convention for intentionally unused bindings in every
-  // linted file — tseslint recommended enables the rule globally, so scoping
+  // linted file: tseslint recommended enables the rule globally, so scoping
   // this to *.ts(x) would leave .mjs at the pattern-less defaults
   {
     rules: {
@@ -48,7 +48,7 @@ export default [
       ],
     },
   },
-  // Node build scripts and Node-run test harnesses — nodeBuiltin, not node,
+  // Node build scripts and Node-run test harnesses use nodeBuiltin, not node,
   // so CommonJS-only names (__dirname, require, module) still fail no-undef
   // in these ES modules
   {
--- a/crates/trusted-server-js/lib/eslint.config.js
+++ b/crates/trusted-server-js/lib/eslint.config.js
@@ -1,5 +1,6 @@
 // ESLint v9 flat config
 import js from '@eslint/js';
+import globals from 'globals';
 import tseslint from 'typescript-eslint';
 import importPlugin from 'eslint-plugin-import';
 import jsdoc from 'eslint-plugin-jsdoc';
@@ -34,6 +35,24 @@ export default [
       'unicorn/prevent-abbreviations': 'off',
       'unicorn/filename-case': 'off',
       'import/order': ['error', { 'newlines-between': 'always' }],
+      '@typescript-eslint/no-unused-vars': [
+        'error',
+        { argsIgnorePattern: '^_', varsIgnorePattern: '^_', caughtErrorsIgnorePattern: '^_' },
+      ],
+    },
+  },
+  // Node build scripts and Node-run test harnesses
+  {
+    files: ['*.mjs', 'test/**/*.mjs'],
+    languageOptions: {
+      globals: globals.node,
+    },
+  },
+  // Tests routinely poke at private state and mock boundaries via `any`
+  {
+    files: ['test/**/*.ts', 'test/**/*.tsx'],
+    rules: {
+      '@typescript-eslint/no-explicit-any': 'off',
     },
   },
 ];
--- a/crates/trusted-server-js/lib/package.json
+++ b/crates/trusted-server-js/lib/package.json
@@ -10,8 +10,8 @@
     "dev": "vite build --watch",
     "test": "vitest run",
     "test:watch": "vitest",
-    "lint": "eslint .",
-    "lint:fix": "eslint --fix .",
+    "lint": "eslint . --max-warnings=0",
+    "lint:fix": "eslint --fix . --max-warnings=0",
     "format": "prettier --check \"**/*.{ts,tsx,js,json,css,md}\"",
     "format:write": "prettier --write \"**/*.{ts,tsx,js,json,css,md}\""
   },
--- a/crates/trusted-server-js/lib/package.json
+++ b/crates/trusted-server-js/lib/package.json
@@ -10,8 +10,8 @@
     "dev": "vite build --watch",
     "test": "vitest run",
     "test:watch": "vitest",
-    "lint": "eslint \"src/**/*.{ts,tsx}\"",
-    "lint:fix": "eslint --fix \"src/**/*.{ts,tsx}\"",
+    "lint": "eslint .",
+    "lint:fix": "eslint --fix .",
     "format": "prettier --check \"**/*.{ts,tsx,js,json,css,md}\"",
     "format:write": "prettier --write \"**/*.{ts,tsx,js,json,css,md}\""
   },
@@ -29,6 +29,7 @@
     "eslint-plugin-import": "^2.29.1",
     "eslint-plugin-jsdoc": "^62.5.4",
     "eslint-plugin-unicorn": "^62.0.0",
+    "globals": "^16.0.0",
     "jsdom": "^28.0.0",
     "prettier": "^3.2.5",
     "typescript": "^5.5.4",
--- a/crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
+++ b/crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
@@ -49,7 +49,6 @@ type TestWindow = Window & {
   // Typed as `any` to avoid the TypeScript intersection with the global
   // Window.tsjs declaration (TsjsApi from core/types.ts), which would require
   // every test fixture to satisfy the full TsjsApi shape.
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   tsjs?: any;
 };
 
@@ -142,7 +141,6 @@ describe('installTsAdInit', () => {
           burl: 'https://ssp/bill',
         },
       },
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
     } as any;
 
     const fetchSpy = vi.spyOn(global, 'fetch');
@@ -198,7 +196,6 @@ describe('installTsAdInit', () => {
         },
       ],
       bids: {},
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
     } as any;
 
     const { installTsAdInit } = await import('../../../src/integrations/gpt/index');
@@ -253,7 +250,6 @@ describe('installTsAdInit', () => {
         },
       ],
       bids: {},
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
     } as any;
 
     const { installTsAdInit } = await import('../../../src/integrations/gpt/index');
@@ -307,7 +303,6 @@ describe('installTsAdInit', () => {
         },
       ],
       bids: {},
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
     } as any;
 
     await runGptBootstrap(googletag);
@@ -364,7 +359,6 @@ describe('installTsAdInit', () => {
         },
       ],
       bids: {},
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
     } as any;
 
     await runGptBootstrap(googletag);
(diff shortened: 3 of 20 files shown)


crates/trusted-server-js/lib/eslint.config.js
```diff
@@ -34,6 +35,24 @@ export default [
       'unicorn/prevent-abbreviations': 'off',
       'unicorn/filename-case': 'off',
       'import/order': ['error', { 'newlines-between': 'always' }],
+      '@typescript-eslint/no-unused-vars': [
```
prk-Jr, line 45: ♻️ **refactor** — This options object only applies to `**/*.ts`/`**/*.tsx`, because it lives in the block scoped to those files. But `@typescript-eslint/no-unused-vars` is *also* active on `.mjs` with **default** options: `tseslint.configs.recommended` enables it globally and turns the base `no-unused-vars` off there.

So the underscore convention this PR adds silently does not hold in the build scripts this PR just brought under lint. Confirmed with a probe file in the package root:

```js
export function probe(_unusedArg) { return 1; }
```

```
1:23  error  '_unusedArg' is defined but never used  @typescript-eslint/no-unused-vars
```

**Fix**: repeat the rule in the `*.mjs` block, or lift it into a top-level rules config so it covers every linted file:

```js
{
  rules: {
    '@typescript-eslint/no-unused-vars': [
      'error',
      { argsIgnorePattern: '^_', varsIgnorePattern: '^_', caughtErrorsIgnorePattern: '^_' },
    ],
  },
},
```

crates/trusted-server-js/lib/eslint.config.js
```diff
@@ -34,6 +35,24 @@ export default [
...
+        'error',
+        { argsIgnorePattern: '^_', varsIgnorePattern: '^_', caughtErrorsIgnorePattern: '^_' },
+      ],
+    },
+  },
+  // Node build scripts and Node-run test harnesses
+  {
+    files: ['*.mjs', 'test/**/*.mjs'],
```
prk-Jr, line 55: 🤔 **thinking** — Worth stating explicitly so nobody over-reads the PR title: this block gives `.mjs` Node globals, but the project's own rules stay scoped to `**/*.ts(x)`, so the build scripts get a thinner gate than `src/` does.

```
$ npx eslint --print-config build-all.mjs
import/order: undefined
plugins: (no import / unicorn / jsdoc)
```

Net effect is still a clear improvement over zero coverage — just don't assume `import/order` now guards `build-all.mjs`. Adding `'*.mjs'` to the TS block's `files` (with the import plugin registered) would close it, if that's wanted here rather than in a follow-up.

crates/trusted-server-js/lib/eslint.config.js
```diff
@@ -34,6 +35,24 @@ export default [
...
+    languageOptions: {
+      globals: globals.node,
+    },
+  },
+  // Tests routinely poke at private state and mock boundaries via `any`
+  {
+    files: ['test/**/*.ts', 'test/**/*.tsx'],
```
prk-Jr: 🤔 **thinking** — No change requested; recording the tradeoff. Turning the rule fully off across `test/**` also covers shared helpers and fixtures where a real type is cheap, and it means new `any` in tests lands with no signal at all — not even a warning.

The narrower alternative would have been a file-level `/* eslint-disable @typescript-eslint/no-explicit-any */` in the handful of GPT/prebid mock files that account for most of the 249. Given the count and that `any` is already the file convention here, the blanket approach seems like the right call for this PR.

crates/trusted-server-js/lib/package.json
```diff
@@ -10,8 +10,8 @@
     "dev": "vite build --watch",
     "test": "vitest run",
     "test:watch": "vitest",
-    "lint": "eslint \"src/**/*.{ts,tsx}\"",
-    "lint:fix": "eslint --fix \"src/**/*.{ts,tsx}\"",
+    "lint": "eslint .",
```
prk-Jr: 🌱 **seedling** — Not for this PR necessarily, but it's the same class of hole this PR closes: `eslint .` exits 0 on warnings. There are 0 warnings today, so no live gap — but any rule that later lands at `'warn'` (several plugin presets default that way) would pass CI silently, and the widened surface makes that more likely, not less.

```json
"lint": "eslint . --max-warnings=0",
"lint:fix": "eslint --fix . --max-warnings=0",
```

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
```diff
@@ -1354,7 +1341,7 @@ describe('installTsRenderBridge', () => {
       .mockImplementation(
         (type: string, handler: EventListenerOrEventListenerObject, opts?: unknown) => {
           if (type === 'message') bridgeListener = handler as (e: MessageEvent) => unknown;
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+
```
prk-Jr: ⛏ **nitpick** — Removing the disable directive left a blank line splitting a two-line arrow body:

```ts
(type: string, handler: EventListenerOrEventListenerObject, opts?: unknown) => {
  if (type === 'message') bridgeListener = handler as (e: MessageEvent) => unknown;

  origAdd(type, handler as EventListener, opts as any);
}
```

Same shape at 1374 and 1731. Prettier won't collapse them. The other seven removals (before `const parsed = JSON.parse(...)`) read fine as paragraph breaks — only these three look accidental.

crates/trusted-server-js/lib/test/integrations/prebid/index.test.ts
```diff
@@ -22,7 +22,16 @@ const {
     () => [] as Array<{ source: string; uids?: Array<{ id: string; atype?: number }> }>
   );
   const mockGetConfig = vi.fn();
-  let mockPbjs: {
+  // Only called after mockPbjs is initialized below, so the forward reference is safe.
```
prk-Jr, line 29: 👍 **praise** — Good comment to leave here. The `let`→`const` move puts a reference to `mockPbjs` above its declaration, which reads like a TDZ hazard at a glance; this line says precisely why it isn't. Verified the claim holds: `mockRemoveAdUnit`'s implementation is never invoked inside the `vi.hoisted` factory, only later through `mockPbjs.removeAdUnit` at test time.

prk-Jr (approved):
## Summary

Widens the JS lint gate from `src/**` to the whole package and cleans the newly-covered surface. Behavior-neutral: no production `src/` file is touched — every code edit is a comment removal, an import blank line, or the `let`→`const` restructure in the prebid mock. Verified locally: `npx eslint .` exits 0 (1.7s), and the `dist` ignore holds (dropped a deliberately-broken probe file into `dist/`, still exit 0). Approving; the findings below are all non-blocking.

## Non-blocking

### 🤔 thinking

- **Blanket `no-explicit-any` off for `test/**`**: defensible at 249/280 and it matches the existing file convention, but it also covers shared helpers/fixtures where types are cheap, and new `any` in tests now lands with no signal. (eslint.config.js:53)
- **`.mjs` gets a weaker gate than "full package" implies**: project rules are scoped to `**/*.ts(x)`, so the newly-covered build scripts only get `js.configs.recommended` + tseslint defaults. (eslint.config.js:46)

### ♻️ refactor

- **The `^_` convention does not reach `.mjs`**: the `no-unused-vars` options live in the `**/*.ts` block, but the rule is active with *default* options on `.mjs`. (eslint.config.js:38)

### 🌱 seedling

- **`lint` has no `--max-warnings=0`**: 0 warnings today, so no live gap — but any future `'warn'`-level rule passes CI silently, which is the same class of hole this PR exists to close. (package.json:13)

### 📌 out of scope

- **A second TypeScript package is still unlinted**: `crates/trusted-server-integration-tests/browser/` (Playwright specs, `helpers/`, `global-setup.ts`, `playwright.config.ts`) has no `lint` script and no format.yml job. If the intent of #985 is "no unlinted TS in CI", that is the remaining gap — worth a follow-up issue rather than scope creep here.

### ⛏ nitpick

- **Stray blank lines where disable directives were removed**: ten spots in `test/integrations/gpt/ad_init.test.ts`. Seven (1408, 1491, 1523, 1559, 1759, 1820, 1895) read fine as paragraph breaks; three inside two-line arrow bodies (1344, 1374, 1731) read oddly. Prettier will not collapse either.

### 👍 praise

- **Forward-reference comment in the prebid mock**: the `let`→`const` restructure could read as a TDZ hazard; the comment states exactly why it is not. (test/integrations/prebid/index.test.ts:25)
- **Behavior-neutral diff**: keeping the gate widening free of any production change makes "widen the gate, then clean the surface" easy to trust in review.

## CI Status

All 19 checks pass on 959f4b9.

- cargo fmt / clippy: PASS
- rust tests (fastly, axum, cloudflare, spin, parity, CLI): PASS
- vitest: PASS
- format-typescript (eslint + prettier): PASS
- format-docs: PASS
- integration + browser integration tests: PASS

crates/trusted-server-js/lib/eslint.config.js
```diff
@@ -34,6 +35,24 @@ export default [
...
+        'error',
+        { argsIgnorePattern: '^_', varsIgnorePattern: '^_', caughtErrorsIgnorePattern: '^_' },
+      ],
+    },
+  },
+  // Node build scripts and Node-run test harnesses
+  {
+    files: ['*.mjs', 'test/**/*.mjs'],
+    languageOptions: {
+      globals: globals.node,
```
ChristianPavilonis: 🔧 **P2 / Medium — Use ESM-only Node globals for `.mjs` files**

Every file matched by this block is an ES module, but `globals.node` combines universal Node globals with CommonJS-only names such as `module`, `exports`, `__dirname`, and `__filename`. ESLint therefore accepts identifiers that are undefined at runtime in `.mjs` files. A probe using `__dirname`, for example, passes `no-undef` with this configuration but fails under Node with `ReferenceError`.

This PR specifically strengthens CI detection for build-script errors, but the declared environment creates false negatives for runtime-breaking CommonJS references. Both build scripts currently derive `__dirname` explicitly, so accidentally removing that declaration would pass lint and break the build. If `.cjs` files are introduced later, configure those separately with `globals.node`.

```suggestion
globals: globals.nodeBuiltin,
```

ChristianPavilonis (approved):
## Summary

Reviewed the lint-surface expansion against `main`. The change is low risk and behavior-neutral overall; all CI checks and clean-checkout JS validation pass. I found one medium-severity lint-environment gap, noted inline.

crates/trusted-server-js/lib/eslint.config.js
aram356: Fixed in 95317a46c — lifted the rule into a top-level rules block so the ignore patterns apply to every linted file. Re-ran the probe: `_unusedArg` in a package-root `.mjs` no longer flags.

crates/trusted-server-js/lib/eslint.config.js
aram356: Fixed in 95317a46c — switched to `globals.nodeBuiltin`. Verified with a probe file: `__dirname` in a `.mjs` now fails `no-undef`, and the 16 original `process`/`console`/`Buffer` fixes still hold (`npm run lint` exits 0).

crates/trusted-server-js/lib/package.json
aram356: Done in 95317a46c — `--max-warnings=0` on both `lint` and `lint:fix`. Zero warnings today, so no behavior change until a preset lands a `warn`-level rule.

crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
aram356: Fixed in 95317a46c — collapsed the blank line in all three arrow bodies (1344, 1374, 1731). The paragraph-break removals before `JSON.parse` are left as-is.

crates/trusted-server-js/lib/eslint.config.js
aram356: Agreed on the read — leaving the `.mjs` rule surface (import/order etc.) for a follow-up rather than widening this PR. The `globals.nodeBuiltin` change in 95317a46c does tighten the environment side of this block.

crates/trusted-server-js/lib/eslint.config.js
aram356: Went with the narrower alternative in 7aef9a509. Measured first: 255 occurrences, 246 of them in six mock-heavy files — those get a file-level disable with a reason. The tail got real types instead: fetch stubs are `as unknown as typeof fetch`, registry uses `as AdUnit` via a type-only import, and config's `'info' as any` cast was dropped entirely ('info' is already in the `logLevel` union). One scoped `disable-next-line` remains for gpt_bootstrap's `AnyRecord` alias. New test files now get full signal, and since ESLint 9 reports unused disable directives as warnings, `--max-warnings=0` fails CI if any directive goes stale. Lint/format clean, 488 tests pass, no new `tsc --noEmit` errors vs baseline.

aram356: Superseded by 23ae86e59: the per-file disables are gone too. All 247 test-tree `any`s are now really typed. The main mechanisms: `Omit<Window, 'tsjs'>` intersections with `Partial<TsjsApi>` for the window mocks (which was the actual fix for the intersection problem the old ad_init comment cited as the reason for `any`), structural `TestPbjs`/`TestAdUnit` types over the prebid hoisted mock, and deleting casts that turned out unnecessary (untyped `vi.fn()` already assigns to `typeof fetch`). One documented `as unknown as` double-cast survives, for a test that deliberately feeds malformed `bids` input. `no-explicit-any` now holds across the whole package with zero exceptions in `test/`. Verified: eslint/prettier clean, 488 tests pass, repo `tsc --noEmit` errors drop 50 -> 12 (all remaining pre-existing in untouched files).

PR #986, opened 2026-07-31
Add design specs for the provider architecture and permission model epic

aram356:
## What this is

Five design specs for the provider/permission epic (#777–#782), written after review of PR #838 so the next implementation pass has a normative behavioral contract instead of discovering design decisions inside a 67-file diff. They live in `docs/superpowers/specs/` per the existing convention. Where a spec contradicts PR #838 — or one of the source issues (each spec ends with an explicit divergence table) — the spec is authoritative.

| Spec | Covers |
| --- | --- |
| `pluggable-providers-design` | EC/device/geo provider architecture: full identity lifecycle contract (mint / parse-canonicalize with provider-declared equivalence / canonical graph key / optional cluster prefix / tombstone), declarative namespace descriptors core proves disjoint, version-rotation schema, active-writer + legacy-reader switching with confirmation-by-presentation rewrite, graph-store startup requirement and no-active-until-commit, runtime failure matrix, adapter-capability matrix, four-adapter parity |
| `permission-model-design` | Policy as `[permissions]` in `trusted-server.toml` (not build-time YAML), enforced-vocabulary rule, three-class signal taxonomy with **regime-scoped grant evidence**, normative normalization matrix preserving current conflict-mode/expiry/proxy semantics, exhaustive withdrawal triggers with a family revocation record, concrete raw-EC egress inventory (path → permission table), stored-provenance S2S authority with legacy fail-closed, regime-based auction dispatch matrix including the raw-TCF arm, fail-closed jurisdiction resolution with an acknowledged static-jurisdiction escape hatch |
| `provider-migration-rollout-design` | Behavior-preservation matrix (16 rows incl. US grant-signal cases and declared hardenings; silent changes are defects), identity stability via deterministic HMAC-prefix vectors, dual-read config release with strict reader-first ordering, graph-schema expand-contract rollout, per-adapter committed preserving fixtures run through the decision matrix in CI, telemetry with retirement thresholds |
| `client-cycle-ec-resolve-design` | Threat model and bar for the browser-resolve EC provider type #838 shipped undeclared: exact-Origin/CSRF authorization, session-bound replay handling with stated reservation states, identity-graph invariant, existing-identity semantics, bounded inputs. **On hold** until its open questions get a dedicated issue |
| `integration-response-header-hook-design` | #782 sharpened: structured mutation operations core validates and attributes, three-stage ordering ending in an inviolable cache/privacy invariant pass, reserved surface incl. framing/hop-by-hop headers and cookie-name granularity, response-eligibility matrix, panic-is-fatal on wasm, lands only with a real consumer |

## Key decisions encoded

- **Policy lives in `trusted-server.toml`**, validated at `ts config push` and startup; absent policy falls back to a compiled-in everything-`requires_signal`, `regime = "gdpr"` posture.
- **Opt-out signals always override a consenting TCF string** and are honored globally; **grant evidence is regime-scoped** so US non-opt-out values cannot authorize identity under GDPR rules.
- **EC providers own their identifiers end to end** (parse, keys, tombstoning); permission gating covers minting and identity use only — parse and revocation always run. Geo is ungated because it is circular (it feeds resolution); device is ungated by recorded decision (security classification authorized by operator selection, with fingerprint-derived buyer-facing row fields removed pending a vocabulary extension).
- **Every raw-EC egress is inventoried and gated** (`user.id`, derived IDs, page bids, forwarding, identify, sync), not only EIDs; batch sync is authorized by stored provenance recomputed against current policy, and legacy rows fail closed until backfilled.
- **Withdrawal writes a family revocation record first** — durable intent, sibling discovery, and fail-closed read marker in one key; the migration story is expand-contract for the graph schema and dual-read for config, with strict reader-first ordering.
- PR #838's `host-signals` and `client-fixed` providers are deliberately not carried over; configs selecting them fail loudly.

The specs have been through three review rounds (an adversarial fresh-eyes pass plus two maintainer reviews); the commit history records each round's findings and fixes.


aram356:
Closing for now and waiting revised specs

PR #992, opened 2026-08-03
Align DataDome tag injection with IP exclusions and add a staging bypass

### LINKED ISSUE #994: DataDome: suppress client-side tag for IP-excluded requests
## Problem

When a request matches a Trusted Server DataDome IP exclusion, server-side Protection API validation is skipped, but the HTML head injector still emits the DataDome browser tag. The browser can therefore still run client-side bot protection even though the request was explicitly whitelisted for server-side protection.

## Proposed behavior

Automatically omit the Trusted Server-injected DataDome client-side tag when the authoritative client IP matches an IP-based protection exclusion. This should cover:

- `protection_excluded_ip_cidrs`
- Config Store-backed IP CIDR sources
- Structured `ip_cidr` and `ip_cidr_source` exclusion rules

This should not require a separate opt-in setting. Non-IP exclusions (path, method, ASN, and query rules) should not automatically suppress the client-side tag. Publisher-originated DataDome tags are out of scope.

## Implementation considerations

- Reuse the existing protection-scope decision rather than independently re-evaluating the client IP during HTML injection.
- Carry a request-scoped internal marker from the request filter into the HTML processor/head injector.
- Continue using Fastly's authoritative client IP; do not trust a caller-supplied header.
- Add diagnostic logging showing that the client tag was omitted and why.

## Acceptance criteria

- An inline or Config Store IP exclusion skips server-side validation and produces HTML without the Trusted Server DataDome tag.
- A non-excluded IP still receives the tag when configured.
- Non-IP protection exclusions retain current tag-injection behavior.
- Unit tests cover inline CIDRs, dynamic CIDR sources, and the non-IP cases.
- The existing `/integrations/datadome/tags.js` route remains functional when requested directly.

ChristianPavilonis:
## Summary

- Align the two DataDome protection layers: when Fastly's authoritative client IP matches an IP-based exclusion, Trusted Server skips the Protection API and omits only its automatically injected DataDome browser tag from processed HTML.
- Add a temporary, secret-backed staging bypass for access-controlled browser tests. A matching fixed `x-ts-datadome-bypass` header skips server-side protection and suppresses the automatic tag. Whenever the enabled Fastly DataDome request filter runs, all values of that header are stripped before the request can reach DataDome or the publisher origin.
- Add structured, privacy-safe protection decision logs and cache safeguards for tag-suppressed HTML.

The issue was that an IP allowlist skipped server-side DataDome validation but still allowed the automatically injected browser tag to run. The new flow keeps those layers consistent while providing a controlled staging-only test path without logging client IPs or exposing bypass credentials.

## Staging safety

`FASTLY_IS_STAGING` is supplied by the Fastly runtime (`1` in staging and `0` in production); it is not a repository-provisioned or build-time value. The code reads it with `std::env::var` at request time, so promoting the same Wasm artifact cannot carry a staging value into production. This is the environment variable equivalent of Fastly SDK 0.12.1's `compute_runtime::is_staging()`. PR #326 also records deployed verification that staging emits `X-TS-ENV: staging` and production does not.

The bypass additionally requires enabled DataDome protection, enabled bypass configuration, exactly one fixed bypass header, and a matching Secret Store credential containing at least 32 bytes. Missing, short, invalid, or duplicate credentials fail closed and proceed through normal protection.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/benches/html_processor_bench.rs` | Updates the HTML processor benchmark fixture for the suppression setting. |
| `crates/trusted-server-core/src/config.rs` | Validates DataDome configuration through full integration construction during deployment validation. |
| `crates/trusted-server-core/src/html_processor.rs` | Carries request-scoped suppression into HTML integration processing and supports multiple state types per integration. |
| `crates/trusted-server-core/src/integrations/datadome.rs` | Adds staging bypass configuration, the private suppression marker, startup validation, and the injector guard. |
| `crates/trusted-server-core/src/integrations/datadome/protection.rs` | Handles and hardens the secret-backed bypass, consumes typed IP-suppression metadata, logs complete protection outcomes, and adds regression coverage. |
| `crates/trusted-server-core/src/integrations/datadome/protection_scope.rs` | Uses typed skip reasons and detects overlapping IP exclusions independently of the primary first-match diagnostic reason. |
| `crates/trusted-server-core/src/integrations/registry.rs` | Allows request filters to attach private typed annotations and keys document state by integration ID and type. |
| `crates/trusted-server-core/src/platform/test_support.rs` | Adds test service builders with configurable client IP and supporting stores. |
| `crates/trusted-server-core/src/publisher.rs` | Propagates suppression through buffered and streaming HTML paths, preserves subresource validators/ranges, and protects synthesized document responses. |
| `crates/trusted-server-core/src/response_privacy.rs` | Centralizes the synthesized-HTML `private, no-store` policy, validator removal, and CDN cache-header stripping. |
| `docs/guide/integrations/datadome.md` | Documents bypass safety, overlap behavior, cache limitations, and diagnostic logging. |
| `docs/superpowers/plans/2026-08-03-datadome-ip-excluded-client-tag.md` | Records implementation and acceptance criteria. |
| `docs/superpowers/specs/2026-08-03-datadome-ip-excluded-client-tag-design.md` | Records design boundaries and privacy requirements. |

## Scope and compatibility

This PR remains limited to the existing DataDome request-filter, HTML-processing, publisher, and privacy paths, plus tests and operator documentation. It does not remove or alter DataDome tags already present in publisher HTML.

The bypass request filter is currently dispatched by the Fastly adapter. Axum, Cloudflare, Spin, and configurations with DataDome protection disabled do not provide the same header-stripping guarantee.

`RequestFilterInput::request` changes from `&Request<EdgeBody>` to `&mut Request<EdgeBody>` so filters can attach private request extensions. In-workspace implementations are updated; downstream source-based forks that construct or destructure this public input may need corresponding source changes.

Tag-suppressed processed HTML is returned as `private, no-store` with validators and CDN-targeted cache directives removed. This response-time policy cannot invalidate tag-bearing HTML already stored in a shared cache ahead of Trusted Server; deployments requiring guaranteed suppression must bypass or purge that cache, or avoid shared caching processed HTML ahead of Trusted Server.

## Closes

Closes #994

## Test plan

- [x] `cargo test-fastly`
- [x] `cargo test-axum`
- [x] `cargo test-cloudflare`
- [x] `cargo test-spin`
- [x] `./scripts/test-cli.sh`
- [x] `cargo clippy-fastly`
- [x] `cargo clippy-axum`
- [x] `cd crates/trusted-server-js/lib && npx vitest run`
- [x] TypeScript and docs Prettier checks
- [x] `git diff --check`
- [ ] Live staging/browser verification after deployment: confirm staging emits `X-TS-ENV: staging`, activation logs report the bypass active, production omits the staging header, and a matching staging request skips protection and suppresses auto-injection.

## Checklist

- [x] New behavior has regression coverage


crates/trusted-server-core/src/response_privacy.rs
```diff
@@ -22,6 +22,7 @@ pub const CDN_CACHE_HEADERS: &[&str] = &[
     "fastly-surrogate-control",
     "cdn-cache-control",
     "cloudflare-cdn-cache-control",
+    "cdn-cache-control",
```
aram356: ♻️ **refactor** - Duplicate entry: `"cdn-cache-control"` is already in this list at line 23, so this addition is a no-op and misleadingly implies new coverage in the diff. The new `cdn-cache-control` assertion in the publisher test passes either way. Please remove the duplicate.

```rust
pub const CDN_CACHE_HEADERS: &[&str] = &[
    "surrogate-control",
    "fastly-surrogate-control",
    "cdn-cache-control",
    "cloudflare-cdn-cache-control",
];
```

crates/trusted-server-core/src/integrations/datadome/protection.rs
```diff
@@ -126,14 +145,63 @@ impl DataDomeIntegration {
...
+    ) -> bool {
+        let Some(bypass) = self
+            .config
+            .protection_test_bypass
+            .as_ref()
+            .filter(|bypass| bypass.enabled)
+        else {
+            return false;
+        };
+        let Some(value) = req.headers_mut().remove(super::HEADER_DATADOME_TEST_BYPASS) else {
```
aram356: 🤔 **thinking** - The header is only stripped when `protection_test_bypass` is present and enabled. After an operator removes the section (the documented teardown step) while test infrastructure still sends `x-ts-datadome-bypass`, the credential is forwarded to the publisher origin. The guide states the header is "removed before the request can reach DataDome or the publisher origin", which only holds while the bypass is enabled.

Consider unconditionally removing the header whenever the DataDome filter runs, and only then consulting the bypass config:

```rust
let value = req.headers_mut().remove(super::HEADER_DATADOME_TEST_BYPASS);
let Some(bypass) = self
    .config
    .protection_test_bypass
    .as_ref()
    .filter(|bypass| bypass.enabled)
else {
    return false;
};
let Some(value) = value else {
    return false;
};
```

crates/trusted-server-core/src/publisher.rs
```diff
@@ -2859,6 +2899,15 @@ pub async fn handle_publisher_request(
     // sets the flag unconditionally and tolerates buffered fallback): adapters
     // without streaming support may reject the flag outright rather than
     // silently buffering, which would fail every publisher fetch.
+    let request_method = req.method().clone();
+    let suppress_datadome_client_side_tag = req
+        .extensions()
+        .get::<crate::integrations::datadome::DataDomeClientTagSuppressed>()
+        .is_some();
+    if suppress_datadome_client_side_tag {
+        req.headers_mut().remove(header::IF_NONE_MATCH);
```
aram356: 🤔 **thinking** - Two aspects of this strip:

1. **Scope**: the marker is attached to every request from an excluded IP, so this removes `If-None-Match`/`If-Modified-Since` on all proxied fetches, including non-HTML assets. A permanently excluded office CIDR loses conditional revalidation site-wide, turning every asset fetch into a full origin 200. The suppression decision only matters for processed HTML, but content type is unknowable pre-fetch; if the origin-load cost matters, `Accept`/`Sec-Fetch-Dest` sniffing could narrow it, otherwise it may be worth a comment documenting the trade-off.
2. **Range parity**: the `should_run_ad_stack` strip above also removes `Range`/`If-Range`; this one does not, so a ranged HTML request from an excluded IP can still put a 206 partial body through the suppressed processing pipeline. Consider extracting a shared helper so both sites strip the same header set:

```rust
fn strip_conditional_validators(req: &mut Request<EdgeBody>) {
    req.headers_mut().remove(header::IF_NONE_MATCH);
    req.headers_mut().remove(header::IF_MODIFIED_SINCE);
    req.headers_mut().remove(header::RANGE);
    req.headers_mut().remove(header::IF_RANGE);
}
```

crates/trusted-server-core/src/html_processor.rs
```diff
@@ -235,6 +246,9 @@ impl HtmlProcessorConfig {
 pub fn create_html_processor(config: HtmlProcessorConfig) -> impl StreamProcessor {
     let post_processors = config.integrations.html_post_processors();
     let document_state = IntegrationDocumentState::default();
+    if config.suppress_datadome_client_side_tag {
+        document_state.get_or_insert_with(DATADOME_INTEGRATION_ID, || DataDomeClientTagSuppressed);
```
aram356, line 250: 🤔 **thinking** - `IntegrationDocumentState` keys by integration ID only, and `get_or_insert_with` replaces the stored value when the requested type differs. DataDome stores nothing else today, so this works, but the first future DataDome rewriter or post-processor that stores document state under `"datadome"` will silently clobber `DataDomeClientTagSuppressed` (or be clobbered by it), re-enabling tag injection with no failure at the collision site. The design doc originally specified carrying this through `IntegrationHtmlContext` as request-scoped context; a dedicated context field (or keying the map by `(id, TypeId)`) would remove the one-slot hazard.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -765,7 +842,15 @@ impl IntegrationHeadInjector for DataDomeIntegration {
         DATADOME_INTEGRATION_ID
     }
 
-    fn head_inserts(&self, _ctx: &IntegrationHtmlContext<'_>) -> Vec<String> {
+    fn head_inserts(&self, ctx: &IntegrationHtmlContext<'_>) -> Vec<String> {
+        if ctx
```
aram356, line 865: ♻️ **refactor** - Missing regression test: the plan (Task 4, Step 4) and the guide both promise that publisher-originated DataDome tags survive suppression and are still rewritten by `rewrite_sdk`, but no test pins it. The guard structurally cannot touch origin markup since it only gates `head_inserts`, so this is coverage rather than a bug, but a test would keep the promise honest: process origin HTML containing a `tags.js` script element with suppression enabled, and assert the element remains and is rewritten exactly as without suppression.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -112,6 +118,28 @@ static DATADOME_URL_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
...
 
+/// Temporary static-header bypass for server-side `DataDome` protection.
+///
+/// This is intended only for an access-controlled staging environment. A
+/// matching `x-ts-datadome-bypass` header bypasses the server-side Protection
+/// API and is removed before the publisher origin receives the request. The
+/// credential itself is loaded from the Secret Store at runtime.
+#[derive(Debug, Default, Clone, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct ProtectionTestBypassConfig {
```
aram356, line 131: 🌱 **seedling** - The bypass relies on operators remembering to remove this section; "Do not enable it in production" is docs-only. The registration log helps, but a forgotten staging bypass stays silently active in production config. Consider a required expiry timestamp (validation rejects or the filter denies after the date) or a deploy-validation warning, so the credential cannot outlive its testing window.

aram356 (changes requested):
## Summary

Aligns the two DataDome protection layers (IP-based exclusions now skip the Protection API and omit only Trusted Server's auto-injected browser tag) and adds a secret-backed staging bypass header. The design is sound: single scope evaluation, typed request extension instead of a synthetic header, constant-time credential comparison, cache privacy on tag-suppressed HTML, and no client IPs in logs. One CI gate is failing and there are a handful of non-blocking concerns, detailed inline.

## Blocking

### 🔧 wrench

- **format-docs CI failure**: `docs/superpowers/plans/2026-08-03-datadome-ip-excluded-client-tag.md` fails the pinned Prettier check (stray blank lines after indented code blocks inside list items, at lines 139, 163, 223, 244, 252, 335, and 346). Fix with `cd docs && npx prettier --write superpowers/plans/2026-08-03-datadome-ip-excluded-client-tag.md`.

## Non-blocking

### 🤔 thinking

- **Bypass header reaches the origin when the bypass is disabled** (`protection.rs:176`, inline)
- **Validator stripping is over-broad and misses Range parity with the ad-stack strip** (`publisher.rs:2908`, inline)
- **Suppression marker rides a one-slot-per-integration document-state map** (`html_processor.rs:250`, inline)

### ♻️ refactor

- **Duplicate `cdn-cache-control` entry in `CDN_CACHE_HEADERS`** (`response_privacy.rs:25`, inline)
- **Missing regression test for publisher-originated tags under suppression** (`datadome.rs:846`, inline)

### 🌱 seedling

- **No expiry or environment guard on the staging bypass** (`datadome.rs:129`, inline)

## CI Status

- fmt: PASS
- rust tests (fastly/axum/cloudflare/spin + cross-adapter parity): PASS
- cloudflare/spin checks: PASS
- vitest: PASS
- format-typescript: PASS
- format-docs: FAIL
- integration tests (browser, Fastly EC lifecycle): pending at review time

crates/trusted-server-core/src/response_privacy.rs
ChristianPavilonis: Fixed in d0fffb9d: removed the duplicate `cdn-cache-control` entry from `CDN_CACHE_HEADERS`.

crates/trusted-server-core/src/integrations/datadome/protection.rs
ChristianPavilonis: Fixed in d0fffb9d: `x-ts-datadome-bypass` is now removed before checking bypass configuration or staging activation. Added coverage for unconfigured, disabled, inactive, valid, and invalid credential paths.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Fixed in d0fffb9d: extracted shared conditional/range stripping for both ad-stack and tag-suppressed requests (`If-None-Match`, `If-Modified-Since`, `Range`, and `If-Range`). The scope remains intentionally broad because content type is unknown before fetch; the comment documents that trade-off. Suppressed HTML now also drops response validators.

crates/trusted-server-core/src/html_processor.rs
ChristianPavilonis: Fixed in d0fffb9d: `IntegrationDocumentState` now keys entries by `(integration ID, TypeId)`, and a regression test verifies two state types coexist for the same integration.

crates/trusted-server-core/src/integrations/datadome.rs
ChristianPavilonis: Fixed in d0fffb9d: extended the suppressed HTML pipeline test to preserve and rewrite a uniquely identified publisher-originated `tags.js` element while confirming auto-injection remains absent.

crates/trusted-server-core/src/integrations/datadome.rs
ChristianPavilonis: Fixed in d0fffb9d: the bypass is now runtime-active only when `FASTLY_IS_STAGING=1`; non-staging configuration logs a warning and continues normal protection after stripping the header. The guide and tests cover the guard.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -2846,11 +2896,15 @@ pub async fn handle_publisher_request(
...
-    if should_run_ad_stack {
-        req.headers_mut().remove(header::IF_NONE_MATCH);
-        req.headers_mut().remove(header::IF_MODIFIED_SINCE);
-        req.headers_mut().remove(header::RANGE);
-        req.headers_mut().remove(header::IF_RANGE);
+    let suppress_datadome_client_side_tag = req
+        .extensions()
+        .get::<crate::integrations::datadome::DataDomeClientTagSuppressed>()
+        .is_some();
+    if should_run_ad_stack || suppress_datadome_client_side_tag {
```
prk-Jr: 🔧 **wrench** — This strip now runs for **every** request from an excluded IP, not just navigations.

`should_run_ad_stack` is itself gated on `is_navigation` (`publisher.rs:2718-2726`), so before this PR the strip only ever ran on GET document navigations. The new disjunct has no such gate, and the suppression marker is set for every request from an excluded IP: `ProtectionScope::evaluate` checks `client_ip` / `client_ip_source` (`protection_scope.rs:220-238`) *before* the structured rules, so the default static-asset `path_regex` exclusion never gets a chance to match.

Result: for any client whose IP is in `protection_excluded_ip_cidrs` (typically QA, monitoring, and partner IPs), every image, font, media, and XHR request that falls through to the publisher proxy loses `If-None-Match`/`If-Modified-Since` (no 304 revalidation — full re-download of every subresource on every page) and `Range`/`If-Range` (byte-range seeking on video/audio degrades to a full 200).

The comment addresses *content-type* narrowing, which is indeed unavailable pre-fetch — but the request-side signal that actually matters here is already computed in this same function and is not spoofable from JS (`is_navigation` at `publisher.rs:2634`, from `Sec-Fetch-Dest`). The tag only lands in documents, so gating on it costs nothing the feature needs.

**Fix**:

```rust
    if should_run_ad_stack || (suppress_datadome_client_side_tag && is_navigation) {
        // Conditional and range requests must not short-circuit the fetch for a
        // document whose HTML needs the DataDome tag omitted: a 304 would let the
        // browser reuse a cached tag-bearing document. Subresource requests carry
        // no injected tag, so they keep their validators and range support.
        strip_conditional_and_range_headers(&mut req);
    }
```

If the broad strip is intentional, it needs to be an explicit, documented decision in `docs/guide/integrations/datadome.md` (excluded IPs lose conditional revalidation and range support site-wide) — right now neither the guide nor the design doc mentions it.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -1432,6 +1447,39 @@ fn response_carries_body(method: &Method, status: StatusCode) -> bool {
         && status != StatusCode::NOT_MODIFIED
 }
 
+/// Prevent shared caches from replaying tag-suppressed HTML to other clients.
+fn apply_datadome_client_tag_cache_privacy(
+    response: &mut Response<EdgeBody>,
+    method: &Method,
+    suppress_datadome_client_side_tag: bool,
+    content_type: &str,
```
prk-Jr, line 1471: 🤔 **thinking** — The inverse cache direction stays unguarded.

This makes the *suppressed* variant private, but the default tag-bearing HTML keeps whatever `public, max-age=...` the origin sent. A shared cache in front of Trusted Server can therefore serve a stored tag-bearing document to an excluded IP, silently defeating suppression.

The design doc's goal list only covers the one direction (`2026-08-03-datadome-ip-excluded-client-tag-design.md:38`), so this reads as known-by-omission rather than an oversight — but it makes the feature best-effort behind any shared cache, and that belongs in the operator docs next to the existing cache-privacy paragraph.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -1432,6 +1447,39 @@ fn response_carries_body(method: &Method, status: StatusCode) -> bool {
...
+        return;
+    }
+
+    let already_uncacheable = response
+        .headers()
+        .get(header::CACHE_CONTROL)
+        .and_then(|value| value.to_str().ok())
+        .map(str::to_ascii_lowercase)
+        .is_some_and(|value| value.contains("private") || value.contains("no-store"));
+    if !already_uncacheable {
```
prk-Jr: 🤔 **thinking** — `private, max-age=0` here vs `private, no-store` on the ad-stack path (`publisher.rs:3017-3021`).

Both block shared caches, but `max-age=0` still permits browser storage. With `ETag`/`Last-Modified` stripped just below, the browser cannot revalidate, so it may reuse a tag-less document after the client leaves the excluded IP range (VPN off, DHCP lease change). `no-store` would match the sibling policy and remove the question entirely.

crates/trusted-server-core/src/integrations/datadome/protection.rs
```diff
@@ -126,14 +145,59 @@ impl DataDomeIntegration {
         match self.protection_scope.evaluate(&facts, input.services) {
             ProtectionScopeDecision::Protect => {}
             ProtectionScopeDecision::Skip { rule_id, reason } => {
-                log::debug!("[datadome] Skipping Protection API for rule {rule_id} ({reason})");
+                let client_tag_omitted = is_ip_exclusion_reason(reason);
+                if client_tag_omitted {
+                    input
```
prk-Jr, line 165: 🤔 **thinking** — Deriving suppression from the winning skip reason means non-IP rules shadow it.

`ProtectionScope::evaluate` checks `excluded_methods` *before* the IP lists and the structured rules *after* them (`protection_scope.rs:215-256`), and only the first match is reported. An operator who excludes a method (or whose ASN exclusion fires) for a client that **also** matches an IP exclusion gets `reason: "method"` / `"asn"` → no marker → the exact server-side/client-side inconsistency this PR fixes still ships for that config.

Worth having `evaluate` report IP-exclusion membership independently of which rule won the race, so the two layers stay aligned regardless of rule ordering.

crates/trusted-server-core/src/integrations/datadome/protection.rs
```diff
@@ -126,14 +145,59 @@ impl DataDomeIntegration {
...
                 return false;
             }
         }
 
         true
     }
 
+    fn take_protection_test_bypass_header(
+        &self,
+        req: &mut Request<EdgeBody>,
```
prk-Jr, line 180: ♻️ **refactor** — The env lookup runs on every request, including the vast majority that carry no bypass header.

`active_protection_test_bypass()` does an `std::env::var` plus a `String` allocation, and it is called before checking whether the header actually had a value. Swapping the two `let Some(...) else` blocks is behaviour-identical — the header is still removed unconditionally on the first line — and skips the env read on every real request:

```rust
        let value = req.headers_mut().remove(super::HEADER_DATADOME_TEST_BYPASS);
        let Some(value) = value else {
            return false;
        };
        let Some(bypass) = self.active_protection_test_bypass() else {
            return false;
        };
```

crates/trusted-server-core/src/integrations/datadome/protection.rs
```diff
@@ -413,6 +477,65 @@ impl DataDomeIntegration {
     }
 }
 
+fn is_ip_exclusion_reason(reason: &str) -> bool {
+    matches!(
+        reason,
+        "client_ip" | "client_ip_source" | "ip_cidr" | "ip_cidr_source"
```
prk-Jr: 🌱 **seedling** — String-matching the reason literals makes suppression silently breakable.

All four values are correct today, but they are `&'static str` produced in two separate places (`protection_scope.rs:224/232` and `ProtectionMatcher::reason()` at `protection_scope.rs:492-503`). Renaming one turns suppression off with no compile error and no test failure beyond the four `*_mark_requests_*` tests. A typed reason enum — or a `ProtectionMatcher::is_ip_based()` that the matcher owns — would let the compiler enforce the coupling.

crates/trusted-server-core/src/integrations/datadome/protection.rs
```diff
@@ -413,6 +477,65 @@ impl DataDomeIntegration {
...
+    );
+}
+
+fn log_protection_skip(input: &RequestFilterInput<'_>, rule_id: &str, reason: &str) {
+    if is_ip_exclusion_reason(reason) {
+        log::info!(
+            "[datadome] protection decision=skipped rule={} reason={} client_tag=omitted method={}",
+            rule_id,
+            reason,
+            input.request.method(),
```
prk-Jr: 🤔 **thinking** — This moved from `debug` to `info`.

Combined with the marker firing for every request from an excluded IP (see the `publisher.rs:2903` comment), a monitoring IP crawling the site now emits one info line per subresource. Consider keeping `info` for navigations and leaving subresources at `debug`.

crates/trusted-server-core/src/integrations/datadome/protection.rs
```diff
@@ -413,6 +477,65 @@ impl DataDomeIntegration {
...
+
+fn log_protection_result(
+    input: &RequestFilterInput<'_>,
+    status: StatusCode,
+    datadome_status: Option<u16>,
+    decision: &RequestFilterDecision,
+) {
+    let method = input.request.method();
+
+    match decision {
```
prk-Jr: 🤔 **thinking** — The `allowed` arm is silent for a subset of allowed requests.

It requires `status == 200 && datadome_status == Some(200)`, so a Protection API 200 with a missing or non-numeric `x-datadomeresponse` logs nothing at all, and the final `Continue(_)` arm is a deliberate no-op. For a decision log meant as an audit trail, "no line" and "allowed" become indistinguishable.

Separately, `status=` is the transport status of the Protection API call, not DataDome's verdict — two different meanings under one key. Worth naming them distinctly (`api_status=` / `datadome_status=`).

crates/trusted-server-core/src/integrations/datadome/protection.rs
```diff
@@ -664,6 +794,539 @@ mod tests {
...
+        assert!(
+            has_client_tag_suppression_marker(&structured_source_request),
+            "structured Config Store IP exclusions should mark the request"
+        );
+    }
+
+    #[test]
+    fn non_ip_exclusions_do_not_mark_requests_for_client_tag_suppression() {
+        let ip = IpAddr::V4(Ipv4Addr::new(192, 0, 2, 10));
+        let cases = [DataDomeConfig {
```
prk-Jr: ⛏ **nitpick** — Single-element `cases` array.

The loop iterates one `DataDomeConfig` with one `PathExact` rule. Either add the other non-IP matchers (`path_prefix`, `path_regex`, `query_param_non_empty`) so the loop earns its keep — which would also cover more of the shadowing case noted at `protection.rs:150` — or drop the loop.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -417,6 +464,48 @@ impl DataDomeIntegration {
...
+            .filter(|bypass| bypass.enabled)
+    }
+
+    fn validate_protection_test_bypass(
+        config: &DataDomeConfig,
+    ) -> Result<(), Report<TrustedServerError>> {
+        let Some(bypass) = config
+            .protection_test_bypass
+            .as_ref()
+            .filter(|bypass| bypass.enabled)
```
prk-Jr, line 497: 🌱 **seedling** — Validation covers the store and secret *names* but not the credential's strength.

`protection.rs:172-195` accepts any non-empty secret value, so a 4-character credential would be accepted for a header that skips bot protection. A minimum-length check at load time — where the value is already in hand — would cost one line.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -417,6 +464,48 @@ impl DataDomeIntegration {
         Ok(())
     }
 
+    pub(crate) fn validate_config_for_startup(
```
prk-Jr, line 474: ⛏ **nitpick** — Missing doc comments on two new `pub(crate)` items: this `validate_config_for_startup` (fallible, so it also wants an `# Errors` section) and `HEADER_DATADOME_TEST_BYPASS` at `datadome.rs:93`. `DataDomeClientTagSuppressed` right above is documented, so this reads as an oversight rather than a choice.

crates/trusted-server-core/src/integrations/registry.rs
```diff
@@ -327,7 +327,7 @@ pub trait IntegrationProxy: Send + Sync {
 pub struct RequestFilterInput<'a> {
     pub settings: &'a Settings,
     pub services: &'a RuntimeServices,
-    pub request: &'a Request<EdgeBody>,
+    pub request: &'a mut Request<EdgeBody>,
```
prk-Jr, line 330: 📝 **note** — `&'a Request<EdgeBody>` → `&'a mut Request<EdgeBody>` is a breaking change for any out-of-tree `IntegrationRequestFilter` implementor. Fine in-workspace (all call sites updated), but worth a line in the PR body so downstream forks aren't surprised.

prk-Jr (changes requested):
## Summary

Aligns DataDome's two protection layers: an IP-based server-side exclusion (or the new secret-backed staging bypass header) now also suppresses the auto-injected DataDome browser tag, carried request-scoped through a typed request extension into the HTML pipeline, with cache-privacy hardening on the suppressed HTML. The security-sensitive parts — constant-time credential comparison, unconditional header strip, fail-closed on empty/missing secret, no credential or client IP in logs — are done correctly and are the strongest part of this PR. One functional regression and one unanswered production-safety question block; everything else is non-blocking.

## Blocking

### 🔧 wrench

- **Conditional/range header stripping now applies to every request from an excluded IP, not just navigations** — inline comment on `crates/trusted-server-core/src/publisher.rs:2903`. Excluded-IP clients lose 304 revalidation and byte-range support on every subresource.

### ❓ question

- **How is `FASTLY_IS_STAGING` provisioned, and what prevents a staging-built artifact from carrying it into production?** (`crates/trusted-server-core/src/integrations/datadome.rs:473-481`)

  ```rust
  fn active_protection_test_bypass(&self) -> Option<&ProtectionTestBypassConfig> {
      if std::env::var(ENV_FASTLY_IS_STAGING).as_deref() != Ok("1") {
          return None;
      }
  ```

  This env var is the entire production safety gate for a header that skips bot protection, and it is set nowhere in this repository — not in `fastly.toml`, `.env.dev`, CI, or any script. The only other reader is `middleware.rs:216` (for the `X-TS-ENV: staging` response header), and `docs/superpowers/specs/2026-03-19-edgezero-migration-design.md:341` already flags it as a Fastly-specific wart to be replaced by adapter config. Two outcomes need ruling out before merge:

  1. If the value is baked into the Wasm at build time, promoting a staging-built artifact to the production service carries `FASTLY_IS_STAGING=1` with it, and the bypass is live in production for anyone holding the credential.
  2. If Fastly Compute never populates non-`FASTLY_*` runtime env vars, `std::env::var` always returns `Err` there, the bypass is dead code that cannot be exercised on the real staging service either — and the deferred "Live production/browser verification" item can never pass.

  Could you state how the flag is set on staging and what keeps it off production? A defence-in-depth option that holds regardless of the answer: additionally require the request host to match a configured staging host, so a mis-set env var alone is insufficient to enable the bypass.

## Non-blocking

### ♻️ refactor

- **Three copies of the same cache-privacy policy** — `response_privacy.rs:36-61` (`enforce_set_cookie_cache_privacy`), `publisher.rs:3017-3038` (ad-stack inline block), and the new `publisher.rs:1451-1481`. All three do "force private unless already private/no-store → drop `ETag`/`Last-Modified` → strip `CDN_CACHE_HEADERS`", with the `already_uncacheable` lowercase-contains check duplicated verbatim between the first and third. This PR shares the CDN header *list*; worth going one step further and moving the whole policy into `response_privacy.rs` so the three cannot drift.

### 🌱 seedling

- **Missing test cases** — the negative coverage is good (wrong credential, unconfigured, `enabled = false`, non-staging, and both invalid-config startup paths), but four gaps remain:
  - Duplicate `x-ts-datadome-bypass` headers. `HeaderMap::remove` removes all values and returns the first, so this correctly fails closed — but nothing pins that behaviour.
  - An empty secret present in the store (`protection.rs:177-182`).
  - The secret-store error path (`protection.rs:183-188`).
  - An end-to-end assertion that the marker set by the request filter survives into `head_inserts`. `suppressed_publisher_request_removes_conditional_validators` (`publisher.rs:5388`) inserts the extension by hand, so the filter → `handle_publisher_request` hop is only covered by the registry-level `req.extensions().get::<RequestAnnotation>()` assertion in `registry.rs:1524`.

### 📝 note

- **PR body claim not backed by the diff (`response_privacy.rs`)** — the body says the file "Shares the CDN cache-header list with the new suppression privacy policy and covers `cdn-cache-control`". `cdn-cache-control` is already in `CDN_CACHE_HEADERS` on `main` (`response_privacy.rs:23`); the only change to that file here is a one-word comment edit at line 40. The sharing happens by `publisher.rs` importing the existing constant.
- **Title says "configurable bypass header"** — the header name is a fixed constant, `HEADER_DATADOME_TEST_BYPASS = "x-ts-datadome-bypass"` (`datadome.rs:93`); only the credential's Secret Store location is configurable. `docs/guide/integrations/datadome.md:197` gets this right ("the fixed `x-ts-datadome-bypass` header"); the title and commit `1ce7892b` ("include configured header name when enabled") read as stale.
- **"removed before the request reaches DataDome or the publisher origin" is conditional** — the strip lives in the DataDome request filter, which is registered only when `enable_protection` is true (`datadome.rs:986-989`) and dispatched only by the Fastly adapter (`filter_request` has exactly one non-test call site, `crates/trusted-server-adapter-fastly/src/app.rs:474`). With protection disabled, or on Cloudflare/Axum/Spin, a client-supplied `x-ts-datadome-bypass` is forwarded to the publisher origin. The docs hedge this correctly ("Whenever the enabled DataDome request filter runs"); the PR body states it unconditionally.

### 👍 praise

- **The credential comparison is done right** (`protection.rs:163-198`) — the header is removed on the very first line, before any config or credential check, so it cannot leak on the disabled/inactive/invalid/absent paths; SHA-256 + `subtle::ConstantTimeEq` avoids both early-exit and length leaks; an empty secret and a store error both fail closed with a warn; and no log line in this PR carries the credential or a client IP. The commit sequence shows the client-IP log from `302d161e` was correctly walked back in `7e6f365a`.
- **`IntegrationDocumentState` re-keyed on `(integration_id, TypeId)`** (`registry.rs:101/141/164`) — this fixes a real latent bug on `main`, where `get_or_insert_with` would overwrite another type's state stored under the same integration ID when the downcast failed. `document_state_keeps_multiple_types_for_one_integration` (`registry.rs:1409`) pins it.
- **DataDome deploy validation upgraded from "does it deserialize" to full construction** (`config.rs:158-160`) — running `try_new` means invalid exclusion regexes, bad protection-API origins, and unsafe tag URLs now fail at `ts config push` instead of at runtime. Strictly stronger than the `validate_integration::<DataDomeConfig>` it replaces.
- **Suppression is wired at the single production HTML-processor construction site** — `create_html_processor` has exactly one non-test caller (`publisher.rs:960`, via `create_html_stream_processor`), which both `PublisherBodyProcessor::new` (`publisher.rs:385`) and `process_response_streaming` (`publisher.rs:463`) route through, so the buffered and streaming paths are covered by construction rather than by two parallel edits. Likewise `head_inserts` has one emission point (`datadome.rs:885`), so the guard cannot be flanked.

## CI Status

All 19 checks pass at `9468dc01`; no local runs were needed.

- fmt: PASS
- clippy / check (fastly, axum, cloudflare native + wasm, spin native + wasm): PASS
- rust tests (fastly + core via Viceroy, axum, cloudflare, spin, cross-adapter parity, ts CLI): PASS
- js tests (vitest): PASS
- integration tests / browser integration tests / Fastly EC lifecycle: PASS
- CodeQL (rust, javascript-typescript, actions): PASS

No new dependencies — `sha2`, `subtle`, and `temp-env` are already workspace deps (`Cargo.toml`/`Cargo.lock` untouched), all pure-Rust and `wasm32-wasip1`-clean. No `std::net`/`std::thread`/tokio added to core, and no Fastly APIs leaked into core.

crates/trusted-server-core/src/integrations/datadome/protection.rs
ChristianPavilonis: Fixed in 73b700bf. Skip reasons are now a typed `ProtectionSkipReason` enum with exhaustive `as_str()` and `is_ip_based()` methods; the string-matching helper was removed.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Fixed in 73b700bf. Conditional/range stripping now applies only to suppressed HTML document contexts (`document`, `iframe`, `frame`, `fencedframe`, `embed`, and `object`), while non-document subresources retain validators and range support. Added navigation, iframe, and video-subresource regressions; `cargo test-fastly` passes all 1,927 core tests.

crates/trusted-server-core/src/integrations/datadome/protection.rs
ChristianPavilonis: Fixed in 73b700bf. The fixed header is removed first and an absent header returns before `active_protection_test_bypass()`, so ordinary requests no longer perform the environment lookup. Duplicate values are also stripped and fail closed.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Documented in 73b700bf. The guide now states that the response-time policy cannot invalidate tag-bearing HTML already stored by a shared cache ahead of Trusted Server, and that guaranteed suppression requires bypassing/purging that cache or avoiding shared caching there. I intentionally did not make all normal tag-bearing HTML uncacheable.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Fixed in 73b700bf. Tag-suppressed HTML now uses the same exact synthesized-document policy as the ad-stack path: `Cache-Control: private, no-store`, no origin validators, and no CDN-targeted cache directives. The policy is centralized in `response_privacy.rs` and covered with cache regression tests.

crates/trusted-server-core/src/integrations/datadome/protection.rs
ChristianPavilonis: Fixed in 73b700bf. `ProtectionScopeDecision::Skip` now carries independent `suppress_client_tag` metadata. Evaluation preserves the original first-match rule/reason for diagnostics while continuing only as needed to establish an overlapping applicable IP exclusion. Tests cover method+IP, ASN+IP, path+IP, and method-inapplicable IP rules.

crates/trusted-server-core/src/integrations/datadome/protection.rs
ChristianPavilonis: Fixed in 73b700bf. IP-exclusion suppression skips are `info` only for HTML navigations and `debug` for subresources, using the existing navigation signal. Matching staging test-bypass events intentionally remain `info` as security audit events.

crates/trusted-server-core/src/integrations/datadome/protection.rs
ChristianPavilonis: Fixed in 73b700bf. The table now contains genuinely matching `path_exact`, `path_prefix`, `path_regex`, and `query_param_non_empty` cases, each asserting that a standalone non-IP exclusion does not attach the suppression marker.

crates/trusted-server-core/src/integrations/datadome/protection.rs
ChristianPavilonis: Fixed in 73b700bf. Result logs now classify every response as `allowed`, `blocked`, or `failed_open`; missing/non-numeric verdicts and status mismatches no longer have a silent `Continue` path. Fields are now distinct `api_status=` and `datadome_status=`, with pure classifier coverage.

crates/trusted-server-core/src/integrations/datadome.rs
ChristianPavilonis: Fixed in 73b700bf. Added documentation for `HEADER_DATADOME_TEST_BYPASS` and a summary plus `# Errors` section for `validate_config_for_startup`.

crates/trusted-server-core/src/integrations/registry.rs
ChristianPavilonis: Documented in the updated PR body. `RequestFilterInput::request` now provides mutable access so filters can attach private typed extensions; downstream source-based forks that construct or destructure the public input may require source updates.

crates/trusted-server-core/src/integrations/datadome.rs
ChristianPavilonis: Fixed in 73b700bf. The loaded Secret Store credential must contain at least 32 bytes before hashing/comparison. Short, empty, and unavailable secrets fail closed without logging their values or lengths. Tests cover 31-byte rejection, the 32-byte boundary, empty, and unavailable secrets.

ChristianPavilonis:
Addressed the latest requested changes in 73b700bf and updated the PR title/body.\n\nFor the `FASTLY_IS_STAGING` safety question: this is a Fastly-provided runtime signal (`1` in staging, `0` in production), equivalent to Fastly SDK 0.12.1 `compute_runtime::is_staging()`. The bypass reads it at request time with `std::env::var`, so it is not baked into or promoted with the Wasm artifact. PR #326 records deployed staging/production verification of the same signal through `X-TS-ENV`. The guide and PR body now document this boundary and the live verification checklist.\n\nAdditional coverage now includes duplicate headers, 31/32-byte credential boundaries, empty/unavailable secrets, overlapping non-IP/IP exclusions, complete fail-open result classification, and a real filter-to-publisher suppression path. All local test, clippy, parity, JS, and formatting gates listed in the PR body pass.

crates/trusted-server-core/src/html_processor.rs
```diff
@@ -950,6 +965,62 @@ mod tests {
         assert_eq!(config.request_scheme, "https");
     }
 
+    #[test]
+    fn suppressed_datadome_tag_preserves_and_rewrites_publisher_tag() {
```
prk-Jr, line 969: ❓ **question** — Does the publisher this targets ship its own DataDome tag?

Suppression omits only Trusted Server's *injected* tag. A publisher-authored `<script src="https://js.datadome.co/tags.js">` is still rewritten to `/integrations/datadome/tags.js` and still executes for an IP-excluded client — this test asserts exactly that (`should preserve the publisher-originated DataDome tag`).

If the target publisher ships its own tag, the two protection layers remain misaligned for them and #994 is only half-closed. Worth an explicit answer before merge, since it decides whether this is the fix or the first half of it.

crates/trusted-server-core/src/integrations/datadome/protection.rs
```diff
@@ -98,11 +116,17 @@ impl DataDomeIntegration {
             .change_context(Self::error("Failed to call DataDome Protection API"))
             .map_err(ProtectionRequestError::Runtime)?;
 
-        Ok(self.classify_protection_response(platform_response.response, input.request.method()))
+        let status = platform_response.response.status();
+        let datadome_status = datadome_response_status(platform_response.response.headers());
+        let decision =
+            self.classify_protection_response(platform_response.response, input.request.method());
+        log_protection_result(&input, status, datadome_status, &decision);
```
prk-Jr, line 130: 🤔 **thinking** — Fail-open transport errors escape the structured decision log.

`log_protection_result` only runs on the success path. The two fail-open arms just above (`Protection setup failed open` at line 64, `Protection API failed open` at line 68) emit free-form text with no `decision=` field. A dashboard counting `decision=failed_open` — which the new docs advertise — will silently under-report the most operationally important failure mode: DataDome being unreachable.

**Fix**: emit the structured line from both error arms too, e.g.

```
[datadome] protection decision=failed_open api_status=none datadome_status=unreachable method={} route=continue
```

crates/trusted-server-core/src/integrations/datadome/protection_scope.rs
```diff
@@ -210,51 +250,75 @@ impl ProtectionScope {
         facts: &ProtectionRequestFacts<'_>,
         services: &RuntimeServices,
     ) -> ProtectionScopeDecision {
-        if self.excluded_methods.matches(facts.method) {
-            return ProtectionScopeDecision::Skip {
-                rule_id: "excluded-methods".to_string(),
-                reason: "method",
-            };
-        }
+        let mut primary = self
```
prk-Jr, line 253: 🤔 **thinking** — `evaluate` no longer short-circuits, adding work to the hottest path.

Replacing the early `return` with an accumulated `primary` means a request already excluded by the default static-asset `path_regex` rule now continues scanning every remaining IP-based rule, and the `excluded_ip_cidr_sources` loop below now also runs on method-excluded requests. Each `ip_cidr_source` probe takes the global `IP_CIDR_SOURCE_CACHE` mutex and **clones the whole `Vec<IpCidr>`** (`load_cidrs`, ~line 401). On an asset-heavy page that is per-subresource cost `main` did not pay.

**Fix**: precompute a `has_ip_exclusions: bool` in `ProtectionScope::compile` and keep the original first-match short-circuit when it is `false`. Most deployments configure no IP exclusions at all and would otherwise pay the new cost for nothing.

crates/trusted-server-core/src/integrations/datadome/protection_scope.rs
```diff
@@ -210,51 +250,75 @@ impl ProtectionScope {
...
-            return ProtectionScopeDecision::Skip {
-                rule_id: "excluded-asns".to_string(),
-                reason: "asn",
-            };
+            primary = Some(("excluded-asns".to_string(), ProtectionSkipReason::Asn));
         }
 
         for rule in &self.exclusion_rules {
+            let reason = rule.matcher.reason();
+            if primary.is_some() && !reason.is_ip_based() {
```
prk-Jr, line 297: 🤔 **thinking** — The skip log can report a reason unrelated to why the tag was suppressed.

When a non-IP rule wins first-match and an IP rule *also* matches, the emitted line reads `reason=asn … client_tag=omitted` (or `reason=method`, `reason=path_exact`). The operator sees a non-IP reason paired with tag omission and has no field naming the rule that actually caused the suppression. The docs describe the behavior, but the log can't be used to debug it.

**Fix**: carry the overlapping rule id through `ProtectionScopeDecision::Skip` and add an `ip_overlap_rule={}` field to the skip line.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -1417,6 +1424,30 @@ pub async fn publisher_response_into_streaming_response(
     }
 }
 
+/// Returns whether a request can render an HTML document context.
+fn is_html_document_request(req: &Request<EdgeBody>) -> bool {
```
prk-Jr, line 1428: 🤔 **thinking** — A private-cache 304 can defeat suppression for clients without Fetch Metadata.

The fallback on line 1440 delegates to `is_navigation_request`, which returns `false` for a client sending neither `Sec-Fetch-Dest` nor `Accept: text/html`. Conditional headers then survive, the origin can answer 304, and the browser replays its own cached **tag-bearing** HTML — so suppression never takes effect for that client.

The docs cover the shared-cache limitation but not this private-cache one; worth either widening the strip or documenting the gap.

⛏ Also: this helper duplicates `is_navigation_request`'s `Sec-Fetch-Dest` parse and then calls it. `http_util.rs`, next to its sibling, is the more natural home.

crates/trusted-server-core/src/integrations/datadome/protection.rs
```diff
@@ -406,13 +493,119 @@ impl DataDomeIntegration {
...
+
+fn suppression_skip_log_level(suppress_client_tag: bool, is_navigation: bool) -> log::Level {
+    if suppress_client_tag && is_navigation {
+        log::Level::Info
+    } else {
+        log::Level::Debug
+    }
+}
+
+fn log_protection_skip(
```
prk-Jr, line 526: ♻️ **refactor** — Triplicated log statement.

`log_protection_skip` computes a `log::Level`, compares it back to `Info`, then repeats three near-identical format strings. One `log::log!` collapses it:

```rust
log::log!(
    suppression_skip_log_level(suppress_client_tag, is_navigation_request(input.request)),
    "[datadome] protection decision=skipped rule={} reason={} client_tag={} method={}",
    rule_id,
    reason.as_str(),
    if suppress_client_tag { "omitted" } else { "kept" },
    input.request.method(),
);
```

crates/trusted-server-core/src/integrations/datadome/protection.rs
```diff
@@ -664,6 +864,764 @@ mod tests {
...
+            http_client.clone(),
+        );
+        let settings = Settings::default();
+        let mut request = request_for_filter();
+        request.headers_mut().insert(
+            super::super::HEADER_DATADOME_TEST_BYPASS,
+            edgezero_core::http::HeaderValue::from_static("temporary-test-credential-32-bytes!"),
+        );
+
+        let decision = temp_env::with_var(
```
prk-Jr, line 1106: 🤔 **thinking** — This test races with the `filter_with_staging` helper on the same env var.

This test sets `FASTLY_IS_STAGING` to `None` while `filter_with_staging` (line 881) sets it to `"1"`, in the same test binary. `temp_env::with_var` mutates process-global state and provides no locking, so the two are order-dependent whenever they run concurrently.

Currently masked because core tests execute single-threaded under Viceroy — it becomes a flake the moment they run natively multi-threaded.

**Fix**: fold the non-staging assertion into one test alongside the staging cases, or add `serial_test` to the env-sensitive ones.

crates/trusted-server-core/src/integrations/datadome/protection.rs
```diff
@@ -34,9 +41,20 @@ enum ProtectionRequestError {
...
+        if test_bypass_matched {
+            input
+                .request
+                .extensions_mut()
+                .insert(super::DataDomeClientTagSuppressed);
+            log_protection_test_bypass(&input);
+            return RequestFilterDecision::Continue(RequestFilterEffects::default());
+        }
+
+        if !self.config.enable_protection || !self.is_request_protected(&mut input) {
```
prk-Jr, line 57: 📝 **note** — `!self.config.enable_protection` is unreachable here.

`register` only attaches the request filter when `enable_protection` is true (`datadome.rs:995`), so this half of the condition can never be taken. Harmless as a belt-and-braces guard; noting it so a future reader doesn't infer that the filter runs with protection disabled.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -2846,11 +2894,16 @@ pub async fn handle_publisher_request(
...
-    if should_run_ad_stack {
-        req.headers_mut().remove(header::IF_NONE_MATCH);
-        req.headers_mut().remove(header::IF_MODIFIED_SINCE);
-        req.headers_mut().remove(header::RANGE);
-        req.headers_mut().remove(header::IF_RANGE);
+    let suppress_datadome_client_side_tag = req
+        .extensions()
+        .get::<crate::integrations::datadome::DataDomeClientTagSuppressed>()
+        .is_some();
+    if should_run_ad_stack || (suppress_datadome_client_side_tag && is_html_document_request(&req))
```
prk-Jr, line 2901: 👍 **praise** — Good call scoping the validator/range strip to document contexts rather than reusing the blanket `should_run_ad_stack` behavior.

Keeping validators and `Range` on subresources means suppressed clients still get 304s and byte-range media on everything that can't carry the injected tag, and `suppressed_subresource_preserves_conditional_and_range_headers` locks it in. Easy place to have over-corrected.

crates/trusted-server-core/src/integrations/registry.rs
```diff
@@ -98,17 +98,19 @@ pub struct IntegrationScriptContext<'a> {
     pub document_state: &'a IntegrationDocumentState,
 }
 
+type IntegrationDocumentStateMap = BTreeMap<(&'static str, TypeId), Arc<dyn Any + Send + Sync>>;
```
prk-Jr, line 101: 👍 **praise** — Re-keying `IntegrationDocumentState` on `(&'static str, TypeId)` is the correct fix rather than a workaround.

The previous id-only key silently collided when one integration needed more than one state type — the old `get_or_insert_with` would even overwrite the entry on a failed downcast. This removes a real footgun for every integration, not just DataDome, and `document_state_keeps_multiple_types_for_one_integration` covers it.

crates/trusted-server-core/src/integrations/datadome.rs
```diff
@@ -765,7 +861,15 @@ impl IntegrationHeadInjector for DataDomeIntegration {
         DATADOME_INTEGRATION_ID
     }
 
-    fn head_inserts(&self, _ctx: &IntegrationHtmlContext<'_>) -> Vec<String> {
+    fn head_inserts(&self, ctx: &IntegrationHtmlContext<'_>) -> Vec<String> {
```
prk-Jr, line 864: 🌱 **seedling** — The "temporary" bypass has no expiry.

Nothing forces removal beyond the `FASTLY_IS_STAGING` gate, so a forgotten `enabled = true` plus a live Secret Store entry persists indefinitely in staging. Not for this PR, but an `expires_at` in `ProtectionTestBypassConfig` that fails closed once passed would make "temporary" enforceable rather than procedural.

prk-Jr (approved):
## Summary

Aligns the client-side DataDome tag with the server-side Protection API skip for IP-based exclusions, and adds a staging-only, Secret Store-backed header bypass. I traced the `DataDomeClientTagSuppressed` marker end to end — request filter → `RequestFilterInput` (now `&mut`) → Fastly dispatch (`execute_named:548` and `dispatch_fallback:729`, both of which pass the same `req` through to `handle_publisher_request`) → `HtmlProcessorConfig` → `IntegrationDocumentState` → `head_inserts` — and the plumbing holds on every path. Every HTML route that injects also reaches `enforce_synthesized_html_cache_privacy`; the `PassThrough` and `BufferedUnmodified` routes correctly skip it because nothing is injected there.

Approving. No correctness or security defect found. The findings below are one open question plus observability, performance, and test-hygiene items — none of which need to block.

## Blocking

_None._

## Non-blocking

### ❓ question

- **Does the target publisher ship its own DataDome tag?** Suppression omits only Trusted Server's injected tag; a publisher-authored tag is still rewritten first-party and still runs for an IP-excluded client (`html_processor.rs:969`). Determines whether #994 is fully closed.

### 🤔 thinking

- **Fail-open transport errors escape the structured decision log** (`protection.rs:63-70`) — no `decision=failed_open` field on the DataDome-unreachable path.
- **`evaluate` no longer short-circuits** (`protection_scope.rs:253`) — extra global-mutex + `Vec<IpCidr>` clone per request on the static-asset path.
- **Skip log can report a reason unrelated to the suppression** (`protection_scope.rs:297`) — no field names the overlapping IP rule.
- **Private-cache 304 can defeat suppression** for clients without Fetch Metadata (`publisher.rs:1428`).
- **`temp_env` race on `FASTLY_IS_STAGING`** between `protection.rs:881` and `protection.rs:1101` — masked today by single-threaded Viceroy execution.

### ♻️ refactor

- **Triplicated log statement** in `log_protection_skip` (`protection.rs:519`) — one `log::log!` replaces three arms.

### ⛏ nitpick

- **`is_html_document_request`** (`publisher.rs:1428`) duplicates part of `is_navigation_request` and then delegates to it; `http_util.rs` is the natural home.

### 📝 note

- **`!self.config.enable_protection` is unreachable** at `protection.rs:57` — `register` only attaches the filter when protection is enabled (`datadome.rs:995`).
- **Fastly-only, confirmed by grep.** `registry.filter_request` has exactly one production call site: `crates/trusted-server-adapter-fastly/src/app.rs:474`. Axum, Cloudflare, and Spin run no request filters at all, so on those adapters `x-ts-datadome-bypass` reaches the publisher origin unstripped and no tag suppression occurs. This matches the PR description — recording it here so the adapter asymmetry is on the review record rather than only in the PR body.

### 🌱 seedling

- **The "temporary" bypass has no expiry** (`datadome.rs:864`) — consider an `expires_at` that fails closed.

### 👍 praise

- **`IntegrationDocumentState` re-keyed on `(&'static str, TypeId)`** (`registry.rs:101`) — the correct fix, not a workaround. The previous id-only key silently collided and could overwrite state on a failed downcast; this removes a footgun for every integration.
- **The bypass fail-closed ladder is well built.** The header is stripped *before* any configuration or credential check, duplicate values are rejected, a 32-byte minimum is enforced on the stored secret, comparison is SHA-256 + `subtle::ct_eq`, and the credential is never logged. Every rung has a test.
- **Scoping the validator/range strip to document contexts** (`publisher.rs:2901`) instead of reusing the blanket `should_run_ad_stack` behavior — suppressed clients keep 304s and byte ranges on everything that cannot carry the injected tag.
- **The test matrix is genuinely thorough** — 32-byte boundary table, non-staging inactivity, duplicate headers, bypass-beats-path-exclusion, and a subresource case asserting validators are *preserved*.

## CI Status

All 19 checks pass on `ed7e1927` (reported from GitHub, not re-run locally):

- fmt: PASS
- clippy / cargo check (cloudflare native + wasm32-unknown-unknown, spin native + wasm32-wasip1): PASS
- rust tests (fastly, axum native, cloudflare, spin, ts CLI native, cross-adapter parity): PASS
- js tests (vitest): PASS
- format-typescript, format-docs: PASS
- integration tests, Fastly EC lifecycle, browser integration tests: PASS

crates/trusted-server-core/src/integrations/datadome/protection.rs
```diff
@@ -125,15 +149,75 @@ impl DataDomeIntegration {
...
+            .iter()
+            .cloned()
+            .collect::<Vec<_>>();
+        req.headers_mut().remove(super::HEADER_DATADOME_TEST_BYPASS);
+        if supplied_values.is_empty() {
+            return false;
+        }
+        if supplied_values.len() != 1 {
+            log::warn!(
+                "[datadome] Multiple DataDome test bypass headers supplied; ignoring bypass"
```
aram356, line 198: ⛏ **nitpick** - This warn fires before the staging/config activation check, so in a production runtime (bypass configured-inactive or unconfigured) any client sending two junk `x-ts-datadome-bypass` headers emits warn-level log lines at will. Reordering the multi-value check after `active_protection_test_bypass()` while keeping the unconditional strip would keep production logs quiet:

```rust
req.headers_mut().remove(super::HEADER_DATADOME_TEST_BYPASS);
if supplied_values.is_empty() {
    return false;
}
let Some(bypass) = self.active_protection_test_bypass() else {
    return false;
};
if supplied_values.len() != 1 {
    log::warn!("[datadome] Multiple DataDome test bypass headers supplied; ignoring bypass");
    return false;
}
```

crates/trusted-server-core/src/publisher.rs
aram356, line 1428: ⛏ **nitpick** - This duplicates the `sec-fetch-dest` parsing idiom of `is_navigation_request` and already calls it as the fallback; `http_util.rs` beside `is_navigation_request` seems like its natural home, keeping the Fetch Metadata interpretation logic in one module.

crates/trusted-server-core/src/integrations/datadome/protection_scope.rs
aram356, line 253: 📝 **note** - Behavior/cost observation, no action required: with the `primary` continuation, method- and path-skipped requests now also evaluate the IP exclusions (including Config Store-backed CIDR source lookups, bounded by the process-local TTL cache) to detect overlap, where evaluation previously returned at first match. This is the intentional price of overlap-aware suppression; noting it here for the record since default configs include a static-asset `path_regex` rule that takes this path on every asset request.

aram356 (approved):
Approving: all earlier blocking and non-blocking findings from previous rounds are verified as addressed at ed7e192 (unconditional bypass-header strip with fail-closed duplicates, document-context-scoped validator/range stripping, `(integration ID, TypeId)` document-state keying, centralized `private, no-store` synthesized-HTML cache policy, typed `ProtectionSkipReason` with overlap-aware suppression metadata, `FASTLY_IS_STAGING` runtime guard with a 32-byte credential minimum, and the publisher-originated tag regression test). CI is fully green including format-docs and the integration suites. The remaining inline comments are minor and do not need another round.

PR #996, opened 2026-08-04
Always emit hb_adid so server-side ad template creatives render

### LINKED ISSUE #926: Pin and browser-test the GAM-served PUC render contract for inline SSAT creatives

prk-Jr:
## Summary

- Bidders that return neither a Prebid Cache UUID nor an `adid` produced **no `hb_adid`** in `window.tsjs.bids`. `adInit` only sets targeting keys that exist on the bid, so GAM never received an `hb_adid` key, the Universal Creative's `%%PATTERN:hb_adid%%` expanded to empty, and the render bridge rejected the resulting `Prebid Request` for want of an ad ID. The line item won and served its wrapper, but the creative never rendered — a silent, per-bidder loss of every SSAT impression.
- Adds `Bid::bid_id`, populated from the OpenRTB bid object's own `id`, and uses it as the last-resort `hb_adid` source. `id` is mandatory per the OpenRTB spec, so this closes the gap for every bidder.
- `cache_id` and `ad_id` keep priority, in that order. This is not cosmetic: when `hb_cache_host`/`hb_cache_path` are present the Universal Creative uses `hb_adid` as the Prebid Cache lookup key (`https://<host><path>?uuid=<hb_adid>`), so an arbitrary value there would break cache retrieval. Both priorities are locked in by test.

### Why a separate field rather than folding into `ad_id`

`ad_id` is the creative/ad identifier and is exposed raw in the debug bid. The OpenRTB `id` is unique per bid *instance*, not per creative — surfacing it as `ad_id` would mislead any consumer treating that field as a creative identifier. The existing comment in `parse_bid` says as much, so the bid ID is carried as its own field instead.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/auction/types.rs` | Add `Bid::bid_id: Option<String>` with `#[serde(default, skip_serializing_if = "Option::is_none")]`, so the serialized wire shape is unchanged when absent |
| `crates/trusted-server-core/src/integrations/prebid.rs` | Populate `bid_id` from the OpenRTB bid `id` in `parse_bid`; two tests covering it alongside and without `adid`/cache |
| `crates/trusted-server-core/src/publisher.rs` | `build_bid_map` `hb_adid` chain becomes `cache_id → ad_id → bid_id`; test for the new fallback, renamed omit-case test, and `bid_id` added to the cache/ad-id tests to prove precedence |
| `crates/trusted-server-core/src/auction/orchestrator.rs` | `bid_id: None` at `Bid` literals (tests) |
| `crates/trusted-server-core/src/auction/formats.rs` | `bid_id: None` at `Bid` literal |
| `crates/trusted-server-core/src/auction/telemetry.rs` | `bid_id: None` at `Bid` literal (test) |
| `crates/trusted-server-core/src/integrations/adserver_mock.rs` | `bid_id: None` at `Bid` literals |
| `crates/trusted-server-core/src/integrations/aps.rs` | `bid_id: None` at `Bid` literal |
| `CHANGELOG.md` | Entry under Unreleased → Fixed |

## Test plan

- [x] Other: `cargo test-cloudflare`, `cargo test-spin`, `cargo clippy-cloudflare`, `cargo clippy-cloudflare-wasm`, `cargo clippy-spin-native`, `cargo clippy-spin-wasm`, and the parity suite (`cargo test --manifest-path crates/trusted-server-integration-tests/Cargo.toml --test parity`)

The new `bid_map_falls_back_to_bid_id_when_cache_id_and_ad_id_absent` test was confirmed to fail with the fallback removed (`left: None`), so it genuinely gates the fix rather than passing vacuously.

Not covered by automated tests: the browser-side half of the loop — that GAM echoes the new `hb_adid` back and the bridge serves the creative. That path is the subject of #926.

## Checklist

- [x] New code has tests


crates/trusted-server-core/src/integrations/adserver_mock.rs
```diff
@@ -318,6 +318,7 @@ impl AdServerMockProvider {
                     }),
                     nurl: original.and_then(|b| b.nurl.clone()),
                     burl: original.and_then(|b| b.burl.clone()),
+                    bid_id: None,
```
ChristianPavilonis: :wrench: **P1 — Preserve `bid_id` through mediation**

This reconstruction restores the original bid's `nurl`, `burl`, `ad_id`, and cache fields, but discards the new `bid_id`. With `adserver_mock` mediation configured, a Prebid winner without a cache UUID or `adid` therefore still reaches `build_bid_map` without `hb_adid`, so the Universal Creative handshake cannot render it.

Please preserve the field here and extend `parse_mediation_response_restores_original_bid_render_fields` (or add an end-to-end mediation assertion) for the `bid_id` → `hb_adid` fallback.

```suggestion
                    bid_id: original.and_then(|bid| bid.bid_id.clone()),
```

crates/trusted-server-core/src/publisher.rs
```diff
@@ -3277,7 +3277,20 @@ pub(crate) fn build_bid_map(
...
+                // a stable value GAM's Universal Creative echoes back verbatim so
+                // the render bridge can find this exact winning bid — for bidders
+                // that return neither a cache UUID nor `adid`. Without it those
+                // bids carry no hb_adid at all, so no targeting key reaches GAM and
+                // the render handshake can never start.
+                let hb_adid = bid
+                    .cache_id
+                    .as_deref()
+                    .or(bid.ad_id.as_deref())
+                    .or(bid.bid_id.as_deref());
```
ChristianPavilonis: :wrench: **P2 — Skip blank preferred IDs before falling back**

`Option::or` treats `Some("")` as present. A blank `cacheId` or `adid` consequently wins over a valid OpenRTB bid ID, emits an empty `hb_adid`, and GPT skips the falsey targeting value. That recreates the missing-targeting failure this fallback is intended to close.

Please filter or reject empty identity strings before applying the cache → ad ID → bid ID priority, and cover a blank `cacheId`/`adid` with a valid `id`.

crates/trusted-server-core/src/auction/types.rs
```diff
@@ -196,6 +196,14 @@ pub struct Bid {
...
     /// Billing notification URL
     pub burl: Option<String>,
+    /// `OpenRTB` bid identifier — the `id` of the bid object itself.
+    ///
+    /// Distinct from [`ad_id`](Self::ad_id): unique per bid instance rather
+    /// than a creative identifier. Always present per the `OpenRTB` spec, so it
+    /// is the last-resort `hb_adid` source for bidders that return neither a
+    /// Prebid Cache UUID nor `adid`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub bid_id: Option<String>,
```
ChristianPavilonis, line 206: :seedling: **Future construction API**

Adding this optional field requires updating many `Bid { ... }` literals. A follow-up `Bid::new(...)` constructor or builder could centralize optional-field initialization and make future schema additions safer. I would avoid deriving `Default`: empty slot IDs, bidders, and dimensions would make invalid production bids too easy to construct.

ChristianPavilonis (approved):
## Review Summary

Reviewed PR #996 against `main`. The direct Prebid fallback is correctly wired, but the inline findings cover an incomplete mediated path and a defensive fallback edge case.

crates/trusted-server-core/src/integrations/adserver_mock.rs
aram356: 🔧 **wrench** — Mediation passthrough drops `bid_id`, re-opening the fixed bug on the mediated path.

`parse_mediation_response` restores the render/accounting fields (`nurl`, `burl`, `ad_id`, `cache_id`, `cache_host`, `cache_path`) from the original SSP bid — its doc comment explicitly promises this — but this literal hard-codes `bid_id: None`. A prebid bid whose only `hb_adid` source is `bid_id` loses it when routed through mediation, so the exact silent no-render failure this PR fixes resurfaces on the mediated path. It also leaves mediated APS bids without any `hb_adid` (APS bids have `ad_id`/`cache_id`/`bid_id` all `None` at `aps.rs:484`, so the restore has nothing to restore). The mediation response is itself OpenRTB, so its bid `id` is available right there.

**Fix**:
```rust
bid_id: bid["id"]
    .as_str()
    .map(String::from)
    .or_else(|| original.and_then(|b| b.bid_id.clone())),
```

This closes both gaps (pass-through prebid bids and mediated APS bids), and deserves a companion test mirroring the existing restore tests.

crates/trusted-server-core/src/publisher.rs
aram356: 🤔 **thinking** — GAM's 40-char targeting-value limit can silently defeat the fallback.

Ad Manager caps key-value *values* at 40 characters. Cache UUIDs (36 chars) fit, but OpenRTB bid `id`s are arbitrary-length — some bidders return long base64/composite IDs. A too-long value gets dropped or truncated by GAM, so the echoed `%%PATTERN:hb_adid%%` fails the bridge's exact-equality check (`gpt/index.ts:1172`) and the creative silently doesn't render — indistinguishable from the original bug.

Since both ends of the handshake use the server-chosen map value (nothing requires it to literally be the bid ID), a bounded derivation is safe: truncate to 40 chars, or use a short digest when `bid_id.len() > 40`. Worth handling now, or at least logging when the fallback value exceeds the limit.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -2223,6 +2223,8 @@ impl PrebidAuctionProvider {
         // not an ad ID, so it is not used as a fallback: surfacing it as `ad_id`
         // (which is exposed raw in the debug bid) would mislead any consumer that
         // treats `ad_id` as a creative identifier. Absent `adid`, `ad_id` is None.
+        // The bid ID is carried separately in `bid_id` instead.
+        let bid_id = bid_obj.get("id").and_then(|v| v.as_str()).map(String::from);
         let ad_id = bid_obj
```
aram356, line 2304: ⛏ **nitpick** — Empty-string `id` flows into the map.

`bid_obj.get("id").and_then(|v| v.as_str()).map(String::from)` accepts `""`, producing `hb_adid: ""` in the bid map. The JS shim's falsy check skips it for targeting, so impact is nil, and the spec requires non-empty — purely defensive:

```rust
let bid_id = bid_obj
    .get("id")
    .and_then(|v| v.as_str())
    .filter(|s| !s.is_empty())
    .map(String::from);
```

aram356 (changes requested):
## Summary

Well-executed, tightly scoped fix: the `cache_id → ad_id → bid_id` precedence is locked in by tests from both directions, the new fallback test was verified to fail without the fix, and the render loop stays symmetric end-to-end (`build_bid_map` → GPT targeting → `%%PATTERN:hb_adid%%` → bridge equality check). One blocking gap on the mediated path, plus a few non-blocking observations.

## Blocking

### 🔧 wrench

- **Mediation passthrough drops `bid_id`** — see inline comment (`crates/trusted-server-core/src/integrations/adserver_mock.rs:321`).

## Non-blocking

### 🤔 thinking

- **GAM 40-char targeting-value limit** — see inline comment (`crates/trusted-server-core/src/publisher.rs:3293`).
- **`hb_cache_host`/`hb_cache_path` are emitted independently of `cache_id`**: if PBS ever returns `ext.prebid.cache.bids.url` without `cacheId`, the map now carries `hb_adid = bid_id` alongside cache coordinates, and (absent inline `adm`) the Universal Creative would fetch `https://<host><path>?uuid=<bid_id>` — a guaranteed miss. Pre-existing with the `ad_id` fallback, but `bid_id` extends it to every bidder. A cheap guard is to gate the `hb_cache_host`/`hb_cache_path` inserts on `cache_id.is_some()` (`crates/trusted-server-core/src/publisher.rs:3300-3315`).

### 🌱 seedling

- **`/auction` responses could now echo the real OpenRTB bid id**: `crates/trusted-server-core/src/auction/formats.rs:281` synthesizes `id: Some(format!("{}-{}", bid.bidder, slot_id))`. Now that `Bid` carries the upstream bid's own `id`, echoing `bid.bid_id` (with the synthetic as fallback) would preserve win-notification correlation and debugging traceability for `/auction` consumers. Follow-up scope, not this PR.

### ⛏ nitpick

- **Empty-string `id` flows into the map** — see inline comment (`crates/trusted-server-core/src/integrations/prebid.rs:2228`).
- **Stale comment in the JS bridge**: `crates/trusted-server-js/lib/src/integrations/gpt/index.ts:1162-1164` explains `hb_adid` non-uniqueness solely via the creative-id fallback; there is now a third tier (`bid_id`, unique per bid instance). The slot-scoped lookup rationale still holds, but the comment is incomplete. File isn't in this diff, hence body-level.

## CI Status

All checks pass on GitHub: fmt, clippy (all six target-matched variants), Rust tests (fastly/axum/cloudflare/spin/CLI), cross-adapter parity, vitest, browser and Fastly EC integration tests, docs/TS format, CodeQL.

crates/trusted-server-core/src/integrations/adserver_mock.rs
prk-Jr: Fixed in 80c1b1e6.

Took the superset of your suggestion and aram356's: the mediation response is itself OpenRTB, so the mediated bid's own `id` is preferred, with `original.bid_id` as the fallback you proposed.

```rust
bid_id: bid["id"]
    .as_str()
    .filter(|id| !id.is_empty())
    .map(String::from)
    .or_else(|| original.and_then(|b| b.bid_id.clone())),
```

Restoring only from `original` would have left mediated APS bids with no `hb_adid` at all — APS carries `ad_id`/`cache_id`/`bid_id` all `None`, so the restore has nothing to recover. Preferring the mediated `id` covers that case and still satisfies the pass-through Prebid case through the fallback. `parse_mediation_response`'s doc comment now records `bid_id` as the one field that is not restore-first.

Two tests: `parse_mediation_response_restores_original_bid_render_fields` gained a `bid_id` assertion, and `parse_mediation_response_falls_back_to_original_bid_id` covers a mediator that omits the per-bid `id`. Verified the latter fails without the change.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Fixed in 80c1b1e6. Correct that `Option::or` treats `Some("")` as present.

Added a `non_empty` helper above `build_bid_map` and switched to `or_else` so every tier is filtered, not just the fallback:

```rust
let hb_adid = non_empty(bid.cache_id.as_deref())
    .or_else(|| non_empty(bid.ad_id.as_deref()))
    .or_else(|| non_empty(bid.bid_id.as_deref()));
```

`bid_map_skips_blank_cache_id_and_ad_id_for_hb_adid` covers a blank `cacheId` and `adid` with a valid bid id; verified it fails without the filter. `prebid.rs` `parse_bid` now also rejects an empty OpenRTB `id` at the source, so a blank never enters `Bid` in the first place.

crates/trusted-server-core/src/auction/types.rs
prk-Jr: Agreed on both points, and agreed it is follow-up scope rather than this PR — a `Bid::new(...)`/builder refactor would touch every construction site across the auction providers and their tests, which would bury a one-line render fix.

Also agreed on not deriving `Default`: an empty `slot_id`, empty `bidder`, and 0×0 dimensions are all invalid for a production bid, and `build_bid_map` already has to defend against zero dimensions.

crates/trusted-server-core/src/integrations/adserver_mock.rs
prk-Jr: Fixed in 80c1b1e6, using your suggested expression (plus an empty-string filter for consistency with the `parse_bid` nitpick below):

```rust
bid_id: bid["id"]
    .as_str()
    .filter(|id| !id.is_empty())
    .map(String::from)
    .or_else(|| original.and_then(|b| b.bid_id.clone())),
```

Your read on mediated APS bids was the deciding factor over ChristianPavilonis's restore-only variant — with `ad_id`/`cache_id`/`bid_id` all `None` at `aps.rs:484` there is nothing for the restore to recover, so preferring the mediated `id` is what actually closes that gap.

Companion tests as asked: `parse_mediation_response_restores_original_bid_render_fields` gained a `bid_id` assertion (mediated `id` wins), and `parse_mediation_response_falls_back_to_original_bid_id` covers a mediator that omits the per-bid `id`. The doc comment now records `bid_id` as the exception to restore-from-original.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Handled with the logging option rather than a bounded derivation, deliberately.

Truncating to 40 chars is not safe here: `hb_adid` has to be unique per bid instance, and a 40-char prefix is not. Two bids sharing a prefix collide, and a non-unique `hb_adid` is precisely what makes one slot's render claim another slot's creative — the duplicate-render/blank-slot failure the slot-scoped bridge lookup at `gpt/index.ts:1172` exists to contain. A short digest avoids the collision but means inventing a hash and a width to pass review, which I would rather not do inside a render-path fix.

So `build_bid_map` now warns when the chosen value is over the limit, with the reasoning recorded in the code:

```rust
const GAM_TARGETING_VALUE_MAX_LEN: usize = 40;
// ...
if id.len() > GAM_TARGETING_VALUE_MAX_LEN {
    log::warn!(
        "hb_adid for slot '{slot_id}' is {} characters, over GAM's \
         {GAM_TARGETING_VALUE_MAX_LEN}-character targeting value limit — \
         GAM may drop the key and the creative will not render",
        id.len()
    );
}
```

The guard covers the whole `hb_adid`, not just the `bid_id` tier, since `adid` is arbitrary-length too. If the warning ever fires in practice, that is the signal to pick a specific bounded derivation with real ids to size it against.

---

On your second non-blocking point, the `hb_cache_host`/`hb_cache_path` gap: took the cheap guard. Both inserts are now behind `bid.cache_id.is_some()`, so a bid with cache coordinates but no UUID falls through to the inline `adm` instead of fetching `?uuid=<bid id>`. Covered by `bid_map_omits_cache_coordinates_without_a_cache_id`, verified failing without the gate, and called out in the changelog since it is a behavior change beyond the original fix.

On the third: agreed that `/auction` echoing the real OpenRTB bid id instead of the synthesized `{bidder}-{slot_id}` is worth doing now that `Bid` carries it, and agreed it is follow-up scope.

crates/trusted-server-core/src/integrations/prebid.rs
prk-Jr: Fixed in 80c1b1e6, exactly as suggested:

```rust
let bid_id = bid_obj
    .get("id")
    .and_then(|v| v.as_str())
    .filter(|id| !id.is_empty())
    .map(String::from);
```

Covered by `parse_bid_treats_blank_bid_id_as_absent`, verified failing without the filter. The same empty-string rejection now applies at the two other points a blank could enter: the `build_bid_map` precedence chain (ChristianPavilonis's P2) and the `adserver_mock` mediation passthrough.

Also updated the stale bridge comment you flagged at `gpt/index.ts:1162-1164` — it now names the third tier (`bid_id`, unique per bid instance) while keeping the slot-scoped-lookup rationale.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -3288,17 +3328,26 @@ pub(crate) fn build_bid_map(
...
-                        serde_json::Value::String(path.clone()),
-                    );
+                //
+                // Gated on `cache_id`: PBS reports the cache `url` and `cacheId`
+                // independently, and hb_adid falls back to a non-cache identifier
+                // (`adid`, then the bid id). Emitting the coordinates without a
+                // cache UUID would point the Universal Creative at
+                // `?uuid=<non-cache-id>` — a guaranteed cache miss — instead of
+                // letting it fall through to the inline `adm`.
+                if bid.cache_id.is_some() {
```
aram356: ⛏ **nitpick** — The cache-coordinate gate and the `hb_adid` chain disagree on a blank `cache_id`.

The gate uses `bid.cache_id.is_some()` while the precedence chain uses `non_empty(...)`. A PBS bid carrying `cacheId: ""` alongside a parseable cache `url` passes the gate but fails the chain, so the map would emit cache coordinates against an `adid`/`bid_id`-sourced `hb_adid` — the `?uuid=<non-cache-id>` miss this gate was built to prevent. One-token fix:

```suggestion
                if non_empty(bid.cache_id.as_deref()).is_some() {
```

Deeply speculative shape (PBS emitting a blank `cacheId` next to a real cache URL), hence nitpick rather than blocking.

aram356 (approved):

PR #997, opened 2026-08-04
Attribute GPT renders to Trusted Server on observed evidence

### LINKED ISSUE #973: Add opt-in GPT runtime diagnostics overlay

prk-Jr:
## Summary

This PR extends the opt-in GPT diagnostics so an operator can answer two separate questions without inferring from targeting, price, or timing:

1. What did Ad Manager report as rendered, and did the Trusted Server creative bridge run?
2. Which observed request path produced each GPT request, and did a later request replace an already-rendered ad?

The implementation remains observational and requires zero publisher-code changes. Diagnostics must not suppress, delay, reorder, add, or remove GPT requests; change targeting; alter auctions or timeouts; or affect creative rendering.

## Evidence reported

For every retained GPT request cycle, diagnostics can now report:

- GPT request, response, render, load, and viewability observations
- Ad Manager line item, order, advertiser, creative, backfill, and yield-group/company identifiers when GPT exposes them
- Ad Manager response class: `empty`, `backfill`, `reservation`, or `unclassified_non_empty`
- Trusted Server delivery evidence from the creative bridge, without treating a filled slot as proof that Trusted Server rendered
- Request source: `trusted_server_direct`, `prebid_refresh`, `publisher_refresh`, `competing`, or `unattributed`
- An opaque Trusted Server auction ID and opportunity-to-request latency when available
- Whether a filled render replaced the most recent earlier filled render, including elapsed time and observed creative transition
- Normal GPT load-before-render ordering through `loadObservedBeforeRender`, without fabricating a negative duration or marking the cycle incomplete

Calls that bypass every installed observation boundary remain honestly `unattributed`. Overlapping request cycles remain `ambiguous`; diagnostics does not guess to improve coverage counters.

## Refresh-source attribution

Request triggers are represented as short-lived, per-slot intents. Trusted Server, Prebid, and publisher evidence expire independently and the next matching `slotRequested` consumes the complete intent.

The diagnostics integration installs a standalone observer around the current `pubads.refresh` before the deferred Prebid wrapper is installed. The observer:

- delegates exactly once
- preserves receiver, argument count and identity, return value, and synchronous throws
- records explicit slot lists or defensively resolves bare refreshes with `getSlots()`
- remains fail-open when GPT, slot resolution, diagnostics, or shared context access throws
- skips publisher attribution during Trusted Server's internal `adInit` refresh and during a Prebid-managed delegation

Prebid exposes only a synchronous diagnostic context around the delegated refresh. It is restored after normal completion and throws, including hostile setter/proxy behavior, and cannot influence targeting, auction timing, callbacks, or whether refresh runs.

Pending intent expiry is coalesced to one callback per slot/source. `WeakRef` avoids retaining slots where supported; older runtimes use a bounded five-second strong-reference fallback.

## Auction correlation

The server copies the existing `AuctionRequest.id` into optional `hb_auction_id` winning-bid metadata for both initial-document and page-bids responses. The value is used only as opaque diagnostic correlation data; it does not affect bid selection or GAM targeting, and no auction payload is retained.

## Creative and replacement attribution

The existing delivery ladder remains evidence-based:

| Delivery | Meaning |
| --- | --- |
| `trusted_server_response_sent` | Trusted Server selected; markup response sent to PUC |
| `trusted_server_selected` | Trusted Server selected; no markup response confirmed |
| `candidate_unconfirmed` | Trusted Server candidate unconfirmed — another GAM result or a creative/bridge failure is possible |
| `no_candidate` | adInit observed no direct Trusted Server candidate for this request |
| `unknown` | Delivery status unknown — required GPT or direct-candidate evidence was not observed |
| `pending` | Waiting for Trusted Server creative evidence |
| `not_applicable` | No delivery conclusion is displayed before render or for an explicitly empty result. |

Replacement comparison uses GPT's primary creative ID with its source-agnostic creative ID as the fallback. Missing or one-sided evidence is omitted rather than described as a change.

## Main changes

| Area | Change |
| --- | --- |
| GPT diagnostics store | Per-source request intents, bounded expiry, auction correlation, replacement derivation, early-load handling, and existing creative-delivery evidence. |
| GPT observer | Ad Manager metadata capture plus a transparent publisher-refresh observer. |
| Prebid integration | Prebid refresh evidence and exception-safe synchronous delegation context. |
| GPT integration | Trusted Server opportunity/auction forwarding and creative bridge evidence. |
| Publisher response generation | Optional auction ID metadata for initial and page-bids winning bids. |
| Overlay/export | Request source, intent, latency, auction, replacement, creative-transition, delivery, and load-order facts. |
| Documentation | Evidence semantics, limitations, privacy boundaries, and operator verification guidance. |

This branch deliberately does not import the `rc/july` GPT slot-handoff implementation or APS renderer behavior.

## Verification

- [x] Full JS suite: 41 files, 672 tests
- [x] JS ESLint and Prettier checks
- [x] JS build: 13 modules
- [x] `cargo test-fastly` — Fastly adapter, core, OpenRTB, and doctests
- [x] `cargo test-axum`
- [x] `cargo test-cloudflare`
- [x] `cargo test-spin`
- [x] `cargo clippy-fastly`
- [x] `cargo clippy-axum`
- [x] Documentation lint and formatting
- [x] Independent specification-compliance review
- [x] Independent code-quality review
- [ ] Remaining target-specific CI checks run in GitHub Actions
- [ ] Manual verification with `?ts_console=true` on representative GAM delivery paths

## Safety and privacy

- Diagnostics remains opt-in and browser-local.
- No bid payload, targeting map, creative markup, cookie, user ID, price, or stack trace is exported.
- Auction IDs are bounded opaque correlation values.
- Request and creative attribution uses observed integration/GPT evidence only.
- Callback ambiguity is surfaced rather than resolved heuristically.
- Retention and pending expiry are bounded.

## Relationship to earlier PRs

This remains the attribution follow-up to #974 and supersedes the closed pre-squash copies in #976 and #990.


crates/trusted-server-core/src/publisher.rs
```diff
@@ -3263,6 +3294,12 @@ pub(crate) fn build_bid_map(
                     "hb_bidder".to_string(),
                     serde_json::Value::String(bid.bidder.clone()),
                 );
+                if let Some(auction_id) = auction_id {
```
ChristianPavilonis, line 3327: 🔧 **Browser-visible auction metadata discloses the HttpOnly EC identifier (P0 / Blocker)**

`auction_id` is the current `AuctionRequest.id`, but for consented users that ID is constructed as `ts-{ec_id}` at lines 3188–3194. This block serializes it into `window.tsjs.bids` and page-bids responses for every winning bid, including when diagnostics is inactive. Any publisher or third-party page script can therefore recover and exfiltrate the long-lived EC identifier, bypassing the cookie's intentional HttpOnly protection. Because the value is stable for the same EC ID, it also is not unique per auction and gives misleading correlation.

**Suggested fix:** Never expose the EC-derived `AuctionRequest.id`. Generate a fresh random per-auction diagnostics token independent of EC identity, preferably emit it only for activated diagnostics sessions, and add regression tests proving neither response path contains the EC value or a deterministic derivative.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
```diff
@@ -115,43 +195,331 @@ function derivedDurations(cycle: MutableRequestCycle): GptDiagnosticsDurations {
...
+}
+
+function responseClass(cycle: MutableRequestCycle): GptDiagnosticsResponseClass | undefined {
+  if (cycle.renderAtMs === undefined) return undefined;
+  if (cycle.isEmpty === true) return 'empty';
+  if (cycle.isEmpty !== false) return undefined;
+  if (cycle.isBackfill === true) return 'backfill';
+  return cycle.adManager?.lineItemId !== undefined ||
+    cycle.adManager?.creativeId !== undefined ||
+    cycle.adManager?.sourceAgnosticLineItemId !== undefined
```
ChristianPavilonis: 🔧 **Source-agnostic identifiers can be misreported as proof of a reservation (P2)**

When `isBackfill` is absent, `sourceAgnosticLineItemId` still makes this return `reservation`. GPT's source-agnostic identifiers apply to both reservation and line-item backfill ads, while the overlay presents the result as ‘Ad Manager reservation line item.’ A defensive or incomplete GPT event is therefore converted into an unsupported factual conclusion, contrary to the evidence-only model.

**Suggested fix:** Without an explicit `isBackfill` value, classify only reservation-specific `lineItemId`/`creativeId` as reservation. Treat a source-agnostic ID alone as `unclassified_non_empty`, and add coverage for that state.

crates/trusted-server-js/lib/src/core/types.ts
```diff
@@ -159,20 +257,46 @@ export interface GptDiagnosticsExportV1 {
...
+}
+
 export interface GptDiagnosticsApi {
   snapshot(): GptDiagnosticsExportV1;
   export(): void;
   subscribe(listener: (snapshot: GptDiagnosticsExportV1) => void): () => void;
   show(): void;
   hide(): void;
+  /** Record Trusted Server's creative opportunity for an associated GPT slot. */
+  recordTrustedServerOpportunity?(
```
ChristianPavilonis: 🔧 **The documented read-only browser API exposes evidence-mutating methods (P2)**

These recorder methods are implemented on the same object installed as `window.tsjs.gptDiagnostics`, although the operator guide documents that object as a read-only facade. Any page script can call them to fabricate request paths, opportunities, creative selection, successful responses, and failure evidence in the panel/export. This also silently expands the public API contract.

**Suggested fix:** Keep the public facade limited to `snapshot`, `export`, `subscribe`, `show`, and `hide`. Move integration writers behind a separate explicitly internal recorder channel/type, and test the public runtime API shape.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
```diff
@@ -221,11 +607,28 @@ export class GptDiagnosticsStore {
...
+        if (facts.isEmpty === true && provisionalCreativeRequest) {
+          this.addAttributionIssue('creative_request_on_empty_cycle', timestampMs, record);
+        }
+
+        if (
+          facts.isEmpty === false &&
+          (cycle.trustedServerOpportunity === 'renderable_candidate' ||
+            cycle.trustedServerOpportunity === 'unrenderable_candidate')
+        ) {
+          this.defer(() => this.notify(), TRUSTED_SERVER_ATTRIBUTION_WINDOW_MS);
```
ChristianPavilonis: 🔧 **Deferred attribution work is not bounded by retained diagnostics state (P2)**

Every candidate render schedules another callback here. Similarly, each consumed/recreated request intent leaves its source-expiry callback outstanding, and the no-`WeakRef` fallback at line 861 strongly retains every marked slot until expiry. The numeric slot/cycle limits therefore do not bound timer-queue work or fallback slot retention. A refresh burst can add event-loop and memory pressure to the page being diagnosed.

**Suggested fix:** Expire request-source evidence lazily during recording/consumption instead of scheduling per-source timers. Coalesce delivery-boundary notifications behind one store-level earliest-deadline timer, and add tests asserting bounded deferred work under rapid refreshes and without `WeakRef`.

ChristianPavilonis (changes requested):
## Summary

Reviewed PR #997 against `main`. The implementation has a confirmed blocker: the new browser-visible auction metadata exposes the EC-derived request identifier. I also found three medium-severity diagnostics correctness/API/resource-bounding issues. Details and concrete fixes are attached inline.

crates/trusted-server-core/src/publisher.rs
aram356, line 3327: 🔧 **wrench** - `hb_auction_id` exposes the HttpOnly EC ID to page JavaScript and the diagnostics export.

`AuctionRequest.id` is `ts-{ec_id}` whenever an EC ID exists (publisher.rs:3189-3192); the random `ts-req-{uuid}` is only the no-cookie fallback. This change copies that value into every winning bid's `hb_auction_id`, which lands in the page-visible `window.tsjs.bids` inline script and the page-bids JSON for **all users, unconditionally** - not just diagnostics sessions. The `ts-ec` cookie is deliberately `HttpOnly` (`ec/cookies.rs`) precisely so page JS cannot read it; this hands the identifier to any third-party script on the page, renders it in the overlay (`Trusted Server auction: ts-...`, overlay.ts:232-234), and ships it in the downloadable export operators are told contains no "cookies, user identifiers" (gpt-diagnostics.md:326). The exported JSON is designed to be attached to tickets and shared - it would carry a stable first-party identity token that is replayable as a `ts-ec` cookie and joinable against edge/SSP logs where the same value appears as the OpenRTB request ID and `user.id`.

It also fails at its stated purpose: because the ID is stable per user, the initial-document auction and every page-bids auction for the same user carry the **same** `hb_auction_id`, so it cannot distinguish auctions.

**Fix**: mint a per-auction random value for the client-visible `hb_auction_id` (keep `AuctionRequest.id` as-is if SSPs depend on it), e.g. reuse the existing `ts-req-{uuid}` form for this field. Consider also gating emission on the GPT diagnostics integration being enabled, since nothing consumes it otherwise.

crates/trusted-server-js/lib/test/integrations/gpt_diagnostics/types.test.ts
```diff
@@ -42,13 +76,79 @@ describe('GPT diagnostics public types', () => {
...
+      },
+    };
 
     expect(snapshot.version).toBe(1);
+    expect(evidenceSnapshot.slots[0]?.requests[0]).toBe(evidenceCycle);
+    expect(evidenceSnapshot.attributionIssues?.[0]).toBe(issue);
+    expect(issueReasons).toHaveLength(8);
+    expect(JSON.stringify(snapshot)).not.toContain('requestIntentId');
+    expect(JSON.stringify(snapshot)).not.toContain('trustedServerAuctionId');
+    expectTypeOf(evidenceCycle.requestPath).toEqualTypeOf<'publisher_refresh' | undefined>();
```
aram356: 🔧 **wrench** - This assertion is a genuine type error, and no gate can surface it.

`expectTypeOf(evidenceCycle.requestPath).toEqualTypeOf<'publisher_refresh' | undefined>()` fails `tsc` with TS2344 (`requestPath` is `GptDiagnosticsRequestPath | undefined`). It stays green because nothing type-checks the tests: `vitest.config.ts` has no `typecheck` block, there is no `tsc --noEmit` step, and CI runs only build + `vitest run`. A probe test containing `expectTypeOf<string>().toEqualTypeOf<number>()` passes the suite.

This matters beyond the one line: the ~16 new `expectTypeOf` assertions in this file are the entire type-level "no PII in the export" contract for the new fields, and none of them are enforced. (observer.test.ts:182 also introduces an unsurfaced TS2322.)

**Fix**: change this line to `toEqualTypeOf<GptDiagnosticsRequestPath | undefined>()`, and enable `test.typecheck` in `vitest.config.ts` scoped to `test/**/types.test.ts` (a blanket `tsc --noEmit` gate is not a drop-in - the package has pre-existing errors).

crates/trusted-server-js/lib/test/integrations/gpt_diagnostics/store.test.ts
```diff
@@ -396,6 +492,1321 @@ describe('GptDiagnosticsStore', () => {
...
+    const [previousOnlyCycle] = store.snapshot().slots[0].requests.slice(-1);
+    const [currentOnlyCycle] = store.snapshot().slots[1].requests.slice(-1);
+    expect(previousOnlyCycle).toMatchObject({ replacedRequestNumber: 1, previousCreativeId: 101 });
+    expect(previousOnlyCycle.creativeChanged).toBeUndefined();
+    expect(currentOnlyCycle).toMatchObject({ replacedRequestNumber: 1 });
+    expect(currentOnlyCycle.previousCreativeId).toBeUndefined();
+    expect(currentOnlyCycle.creativeChanged).toBeUndefined();
+  });
+
+  it('does not infer replacements once the earlier filled cycle has been evicted', () => {
```
aram356, line 875: 🔧 **wrench** - This test passes vacuously; the eviction-replacement path is never reached.

Recording 11 requests leaves all 10 retained cycles with `renderAtMs === undefined`, so the final `recordSlotRenderEnded` matches 10 compatible candidates and `matchCycle` classifies it **ambiguous** - `recordReplacement` never runs (store.ts:925 has zero coverage from the whole suite; verified with a probe: `coverage.slotRenderEnded = {observed:2, matched:1, ambiguous:1}` and `latest.renderAtMs === undefined`). The three `toBeUndefined()` assertions hold because nothing was recorded at all, not because replacement inference was suppressed. This is the only test guarding replacement-after-eviction, and it would still pass if `recordReplacement` fabricated a replacement from an evicted cycle.

**Fix**: complete each filler request (`recordSlotResponseReceived` + `recordSlotRenderEnded`) so exactly one cycle is open when the final render lands, then assert `renderAtMs` is defined **and** `replacedRequestNumber` is undefined.

docs/guide/integrations/gpt-diagnostics.md
```diff
@@ -159,32 +360,49 @@ Diagnostic records remain memory-only and use no `localStorage`, `sessionStorage
...
 
 ### Callbacks are Ambiguous
 
-Overlapping requests for the same GPT Slot object cannot be correlated safely because documented callbacks do not expose a request-cycle identifier. Avoid overlap in controlled tests, or use the issue record as evidence that correlation was not possible.
+Overlapping requests for the same GPT Slot object cannot be correlated safely because
+documented callbacks do not expose a request-cycle identifier. Avoid overlap in
+controlled tests, or use the issue record as evidence that correlation was not
+possible.
 
 ## Limits
```
aram356, line 422: 🔧 **wrench** - The guide contradicts the implementation in four places; two are the privacy statements this PR's changes falsify.

1. This Limits section says diagnostics "does not patch arbitrary publisher `display()` or `refresh()` calls" - but this PR adds `installRefreshObserver` (observer.ts:237-272), which wraps `pubads.refresh`, and lines 122-128 of this same guide describe that wrapper. Reword to: does not patch publisher `display()`, and cannot observe a `refresh` reference captured before installation.
2. The V1 export field list (~L313-327) omits every new correlation/replacement field the export now carries via the `...cycle` spread (`trustedServerAuctionId`, `requestIntentId`, `opportunityToRequestMs`, `replacedRequestNumber`, `previousRenderToRequestMs`, `previousCreativeId`, `creativeChanged`, `loadObservedBeforeRender`, `delivery`), while L326 implies no auction-level identifier is exported.
3. ~L289 calls this a "read-only operator API", but five `record*` writers are installed on the same `window.tsjs.gptDiagnostics` object. Qualify them as internal integration hooks.
4. The retention bullet (~L337) says "Direct and Prebid request-path markers: five seconds" - `REQUEST_PATH_ATTRIBUTION_WINDOW_MS` applies to all three intent sources including `publisher_refresh`.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
```diff
@@ -221,11 +607,28 @@ export class GptDiagnosticsStore {
...
         cycle.renderAtMs = timestampMs;
         cycle.isEmpty = facts.isEmpty;
         cycle.size = facts.size ? ([...facts.size] as Size) : undefined;
         cycle.isBackfill = facts.isBackfill;
         cycle.slotContentChanged = facts.slotContentChanged;
+        cycle.adManager = facts.adManager ? { ...facts.adManager } : undefined;
+
+        if (facts.isEmpty === false) this.recordReplacement(record, cycle);
+
+        if (facts.isEmpty === true && provisionalCreativeRequest) {
```
aram356, line 636: 🤔 **thinking** - An empty render leaves its creative attempt live, allowing contradictory delivery evidence.

When `facts.isEmpty === true` with a provisional creative request, this records `creative_request_on_empty_cycle` but the attempt stays `'live'`. A later `recordTrustedServerCreativeResponse(attemptId)` then sets `trustedServerCreativeResponseAtMs` on the empty cycle, and `delivery()` reports `'trusted_server_response_sent'` (its first check, ahead of the `isEmpty` branch) on a cycle whose `responseClass` is `'empty'`. The same contradiction is reachable in the other order, since `delivery()` checks the response timestamp before `isEmpty`.

Given the design goal of surfacing ambiguity rather than resolving it, consider also calling `this.evictCreativeAttempt(cycle)` here so the misattributed attempt cannot complete against an empty cycle - the attribution issue already records why.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -1226,7 +1271,8 @@ export function installRefreshHandler(timeoutMs = 1500): void {
       const deliverySlots = publisherDeliverySlots(targetSlots);
       const independentSlots = targetSlots.filter((slot) => !deliverySlots.has(slot));
       if (independentSlots.length === 0) {
-        return originalRefresh(slots, opts);
+        recordPrebidRefreshForDiagnostics(targetSlots);
```
aram356, line 1304: 🤔 **thinking** - In this path Prebid ran no auction: the publisher called `pubads.refresh` and the wrapper is passing it through because every target slot is a delivery slot. Labeling it `prebid_refresh` rather than `publisher_refresh` is defensible (the slots are under the publisher's own Prebid delivery), but it means "publisher_refresh" specifically means "publisher refresh that Prebid did not consume", which the guide does not spell out.

Related: the correctness of `prebid_refresh`-not-`competing` labeling depends on the diagnostics observer wrapping `pubads.refresh` **before** this wrapper does (otherwise the outer diagnostics wrapper records publisher intent before the flag is set, and the cycle reports `competing`). The implementation plan called for an install-ordering regression test, but it was dropped along with the `index.ts` changes it targeted, and `observer.test.ts` does not cover the both-wrappers ordering. Worth adding, since nothing structural enforces the order beyond bundle evaluation sequence.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/observer.ts
```diff
@@ -170,4 +233,41 @@ export class GptDiagnosticsObserver {
...
+    const { store, window: observerWindow } = this;
+    pubads.refresh = function (this: unknown, ...args: unknown[]): unknown {
+      try {
+        if (
+          !(
+            observerWindow.tsjs?.adInitRefreshInProgress ||
+            observerWindow.tsjs?.prebidRefreshDispatchInProgress
+          )
+        ) {
+          const rawSlots = args.length === 0 ? pubads.getSlots?.() : args[0];
```
aram356: ♻️ **refactor** - `refresh(null)` loses attribution.

GPT documents `refresh(slots?, opts?)` where an explicit `null` slots argument refreshes all slots, and `refresh(null, {changeCorrelator: ...})` is the canonical way to pass options while refreshing everything. Here only `args.length === 0` falls back to `getSlots()`; `args[0] === null` produces `[]` and the request goes `unattributed`.

```ts
const rawSlots = args.length === 0 || args[0] == null ? pubads.getSlots?.() : args[0];
```

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts
```diff
@@ -126,11 +126,152 @@ function formatMilliseconds(value: number | undefined): string | undefined {
...
+      return 'Trusted Server candidate unconfirmed — another GAM result or a creative/bridge failure is possible';
+    case 'no_candidate':
+      return 'adInit observed no direct Trusted Server candidate for this request';
+    case 'unknown':
+      return 'Delivery status unknown — required GPT or direct-candidate evidence was not observed';
+    case 'pending':
+      return 'Waiting for Trusted Server creative evidence';
+    case 'not_applicable':
+      return undefined;
+    default:
```
aram356, line 147: ♻️ **refactor** - None of the new union switches are exhaustiveness-checked; this one has a literally dead arm.

`deliveryFact` has both `case 'not_applicable': return undefined;` and `default: return undefined;` - the `default` swallows any future `GptDiagnosticsDelivery` member silently. The same pattern is in `requestPathFact` (L163-164), `trustedServerOpportunityFact`, `responseClassFact` (L221-222), and `badges.ts` `deliveryLabel` (`case 'not_applicable': default:`). Adding a union member compiles clean and renders as "Unknown (not observed)" or vanishes from the badge.

`creativeFailureFact` (L180-193) is the correct pattern in this same file: exhaustive switch with `case undefined:` and no `default`, so a new member is a compile error. Apply it to the other five. (Optionally add `@typescript-eslint/switch-exhaustiveness-check` - the eslint config currently has no type-aware rules.)

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/api.ts
```diff
@@ -94,13 +161,31 @@ export class GptDiagnosticsApiController {
           ...cycle,
```
aram356, line 192: ♻️ **refactor** - The "allowlisted export" is actually a spread.

`{ ...cycle, ... }` (and `copyCycle` in store.ts) exports whatever the store happens to hang on a cycle object, so any future field added for an unrelated diagnostic ships to the user-downloadable JSON with zero review signal. The guarding tests are blocklist regexes - and the one covering the evidence snapshot drops `auction` from its pattern, which is exactly how the EC-derived `trustedServerAuctionId` reached the export unnoticed. The test named "creates a fresh V1 allowlist snapshot" (api.test.ts:249) asserts a blocklist.

**Fix**: construct the exported cycle field-by-field, or cheaper, pin the shape in the test: `expect(Object.keys(cycle).sort()).toEqual([...])` for cycles, slots, and metadata, so any new field forces a deliberate test update.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts
```diff
@@ -76,12 +76,38 @@ function formatMilliseconds(value: number | undefined): string | undefined {
   return `${Math.round(value)} ms`;
 }
 
+function deliveryLabel(cycle: GptDiagnosticsRequestCycle): string | undefined {
+  if (cycle.trustedServerCreativeResponseAtMs !== undefined) return 'TS response sent';
```
aram356: ♻️ **refactor** - These two timestamp pre-checks re-derive `delivery()`'s first two precedence rules (store.ts:226-229). Today they agree exactly, but it is a second copy of the rule that must be kept in sync - and it is also what would surface a "TS response sent" badge on an empty cycle if the empty-render attempt-eviction gap in the store is left as is. Switch on `cycle.delivery` alone.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -369,6 +369,51 @@ type RefreshGptSlot = {
...
+  } catch {
+    // Diagnostics context must not affect refresh delegation.
+  }
+  try {
+    return refresh(slots, opts);
+  } finally {
+    if (shouldRestoreContext && tsjs) {
+      try {
+        if (hadOwnContext) {
+          tsjs.prebidRefreshDispatchInProgress = previousContext;
```
aram356, line 409: ♻️ **refactor** - This restore arm (`hadOwnContext === true`) is the only uncovered new statement in the prebid changes, and it is the true re-entrancy case: a `refresh()` fired from inside another `dispatchPrebidRefresh`. If restoration regressed here, the flag would stay `true` after the inner call returns and the diagnostics observer would silently stop attributing all subsequent publisher refreshes.

The test titled for nested refreshes (index.test.ts:1549-1614) never gives `tsjs` an own `prebidRefreshDispatchInProgress` before the wrapper runs, so only the `delete` arm executes. Add a case where `originalRefresh` re-enters `pubads.refresh(...)` and assert the flag is still `true` after the inner call returns and absent after the outer completes.

crates/trusted-server-js/lib/src/core/types.ts
```diff
@@ -159,20 +257,46 @@ export interface GptDiagnosticsExportV1 {
   };
   slots: GptDiagnosticsSlotExport[];
   callbackIssues: GptDiagnosticsCallbackIssue[];
+  attributionIssues?: GptDiagnosticsAttributionIssue[];
```
aram356: ⛏ **nitpick** - Optionality does not match the producer: `GptDiagnosticsApiController.snapshot()` always emits `attributionIssues` and `droppedAttributionIssues`, and the store snapshot declares both required. Keeping them optional makes `(store.attributionIssues ?? [])` / `?? 0` dead branches in api.ts and forces every export consumer to handle an `undefined` the producer never emits. Making them required is still V1-compatible.

Related: `GptDiagnosticsSlotHandle` is structurally identical to `GptDiagnosticsSlotLike` (store.ts:42); `ApiStore` declares the writer signatures with one while the store implements with the other, and structural typing hides future drift. Alias one to the other.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
```diff
@@ -115,43 +195,331 @@ function derivedDurations(cycle: MutableRequestCycle): GptDiagnosticsDurations {
...
+    if (existingAttemptId !== undefined) {
+      if (cycle.trustedServerCreativeResponseAtMs !== undefined) return undefined;
+
+      const existingAttempt = this.creativeAttempts.get(existingAttemptId);
+      if (!existingAttempt) {
+        this.reportAttributionIssue('creative_attempt_unknown', timestampMs, identity);
+        return undefined;
+      }
+      if (existingAttempt.status === 'live') return existingAttempt.id;
+      if (existingAttempt.status === 'completed') return undefined;
```
aram356, line 396: ⛏ **nitpick** - This guard (and the one at line 396) is unreachable: a `'completed'` attempt always set `trustedServerCreativeResponseAtMs` on the same cycle that `attemptIdsByCycle` keys on, so the check at line 377 returns first. No test can be written for them. Either drop them or add a comment marking them defensive so a future reviewer does not ask for coverage.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts
```diff
@@ -126,11 +126,152 @@ function formatMilliseconds(value: number | undefined): string | undefined {
...
+}
+
+function responseClassFact(cycle: GptDiagnosticsRequestCycle): string | undefined {
+  switch (cycle.responseClass) {
+    case 'empty':
+      return 'Ad Manager response class: empty';
+    case 'backfill':
+      return 'Ad Manager response class: backfill';
+    case 'reservation':
+      return 'Ad Manager reservation line item';
```
aram356: ⛏ **nitpick** - Sibling arms use the `Ad Manager response class: X` prefix (lines 214, 216, 221); this one returns `Ad Manager reservation line item`. Diagnostic text gets grepped - keep the pattern: `Ad Manager response class: reservation`.

aram356 (changes requested):
## Summary

The diagnostics evidence model (per-slot request intents, bounded attempt lifecycle, evidence-ladder delivery) is well designed and thoroughly tested, and the presentation layer is XSS-clean (every DOM insertion goes through `textContent`/constant `setAttribute`, closed shadow root). However, the auction-correlation feature undermines the PR's own privacy contract: `hb_auction_id` is derived from the EC ID and ships to all users, and the type-level guards meant to protect the export surface are not enforced by any gate. Requesting changes on those grounds plus a few doc/test integrity issues; the rest is non-blocking.

## Blocking

### 🔧 wrench

- **`hb_auction_id` exposes the HttpOnly EC ID to page JS and the diagnostics export** (crates/trusted-server-core/src/publisher.rs:3297, inline comment)
- **Privacy type contract unenforced; one assertion already broken** (crates/trusted-server-js/lib/test/integrations/gpt_diagnostics/types.test.ts:129, inline comment)
- **Replacement-after-eviction test passes vacuously** (crates/trusted-server-js/lib/test/integrations/gpt_diagnostics/store.test.ts:887, inline comment)
- **Operator guide contradicts the implementation** (docs/guide/integrations/gpt-diagnostics.md:399, inline comment)

## Non-blocking

### 🤔 thinking

- **Empty render leaves its creative attempt live** (store.ts:621, inline comment)
- **Pass-through publisher refreshes labeled `prebid_refresh`** (prebid/index.ts:1274, inline comment)

### ♻️ refactor

- **Snapshot recomputed per subscriber**: `scheduleNotification` calls `this.snapshot()` inside the listener loop (api.ts:233-235). Each subscriber gets a different `capturedAt`, and at the 5s `TRUSTED_SERVER_ATTRIBUTION_WINDOW_MS` boundary two subscribers in the same notification can disagree on `pending` vs `candidate_unconfirmed`. It is also O(subscribers × slots × cycles). Hoist one `const snapshot = this.snapshot();` above the loop (in its own try so a snapshot failure still cannot escape).
- **Coverage gaps in new code paths** (all verified uncovered via V8 coverage):
  - `recordPublisherRefresh` malformed inputs (store.ts:341, 343) - the writer fed directly from publisher-supplied `refresh()` arguments is the only one of the three intent writers without a malformed-input test.
  - `MAX_TRUSTED_SERVER_ASSOCIATIONS` LRU trim (store.ts:317-321) - the only bound in the file with no test; a regression silently drops the oldest `auctionSlotId` association on long-lived SPAs.
  - WeakRef-collected expiry path (store.ts:880, 884) - the WeakRef-absent fallback is tested, but not a `WeakRef` whose target was collected, which is the case the WeakRef exists for.
  - `invalid_event_order` arms for `slotRenderEnded`/`slotOnload`/`impressionViewable` (store.ts:642-644, 661-663, 681-683) and the non-finite-percentage arm of `recordSlotVisibilityChanged` (store.ts:700-710).
  - `recordTrustedServerCreativeFailure` with an unknown attempt ID (store.ts:502-503); `safelyRecordCreativeFailure` early return when the attempt was never created (gpt/index.ts:1173); `badges.ts:123` incomplete-sequence line; overlay `yieldGroupIds`/`companyIds` facts (overlay.ts:206-208).
- **Test hygiene bugs**:
  - badges.test.ts:276: `gptDiagnosticsBadgeTextForTest.toString()` returns only `badgeText`'s own body, so the `not.toMatch(/GAM winner|bidder|provenance/i)` guard cannot see `deliveryLabel`, which is where the new delivery vocabulary lives. Assert over rendered output instead.
  - overlay.test.ts:327: the store is constructed without a `defer` stub, so `recordPublisherRefresh` schedules a real 5000ms `setTimeout` that outlives the test. Add `defer: () => undefined` matching overlay.test.ts:119.

### ⛏ nitpick

- **Tautological assertions**: types.test.ts:124-128 and 135-137 assert hand-written literals against themselves (the identity `toBe` checks, the `toHaveLength(8)` on a literal array, and the blocklist regexes over objects the test itself wrote constrain nothing about production types); store.test.ts:622's `not.toContain('other_demand')` cannot fail since that string exists nowhere in the codebase; badges.test.ts:131, 144, 220-232 add `not.toMatch` guards on strings already pinned by exact `toBe`.
- **Mega-tests**: overlay.test.ts:113-322 (one `it`, ~50 assertions across 9 scenarios), observer.test.ts:131-164 (5 behaviors with interleaved arrange steps), prebid/index.test.ts:1549-1614 (4 scenarios sharing mutable state) would be cleaner and more diagnosable as `it.each`, which these same files already use well elsewhere.

### 📝 note

- **Stale PR description**: the GitHub PR body still documents the deleted 5-state delivery ladder (`trusted_server` / `other_demand`); the shipped code and guide use the 7-state ladder (`trusted_server_response_sent`, `trusted_server_selected`, `candidate_unconfirmed`, `no_candidate`, `unknown`, `pending`, `not_applicable`). Worth rewriting before merge.
- **Spec hygiene**: the 2026-08-04 design spec declares a 4-value `GptDiagnosticsRequestPath` union with no forward pointer to the 2026-08-05 extension that added `publisher_refresh`; a one-line "Extended by ..." note would prevent a reader landing there first from getting a stale union. Both spec statuses (`Proposed` / `Approved`) should become `Implemented` when this merges, matching existing repo practice.

### 🌱 seedling

- **Open Bidding classification**: a yield-group render (`yieldGroupIds` present, no line-item/creative ID) classifies as `unclassified_non_empty` despite carrying positive Ad Manager evidence; a distinct response class may be worth adding later.

## CI Status

- fmt: PASS
- clippy (all adapters): PASS
- js tests (vitest): PASS
- browser + Fastly EC integration tests: PASS
- CodeQL: PASS

crates/trusted-server-core/src/publisher.rs
prk-Jr: Fixed in a1b90464.

`hb_auction_id` no longer carries `AuctionRequest.id`. A new `diagnostics_auction_id()` mints `ts-auc-{uuid}` fresh per auction, unrelated to EC identity, and returns `None` unless the GPT diagnostics integration is enabled, so it is absent entirely for deployments that would not consume it. `AuctionRequest.id` is unchanged for SSPs.

On emitting it only for activated sessions: the per-document activation decision requires an eligible GET navigation, so it does not exist on the `/_ts/page-bids` fetch that carries the same field. The gate is therefore at the integration-config level. What removes the disclosure is that the value is now random per auction rather than derived from anything.

Regression coverage:
- `page_bids_response_includes_auction_id_only_for_winning_bids` asserts the emitted value starts `ts-auc-`, is not `auction_request.id`, and does not embed the EC ID.
- `page_bids_auction_id_is_per_auction_and_gated_on_diagnostics` asserts two auctions for the same EC context mint different tokens, and that nothing is emitted with the integration disabled.
- `diagnostics_auction_id_is_fresh_and_gated` covers the helper the stream and non-HTML collect paths share.

prk-Jr: Fixed in a1b90464 — same change as the thread above.

`diagnostics_auction_id()` mints `ts-auc-{uuid}` per auction and is emitted only when the diagnostics integration is enabled. `AuctionRequest.id` stays as-is for SSPs.

Your second point drove the shape of the fix as much as the first: a per-visitor-stable value could not distinguish auctions, so reusing the request ID failed the feature's own purpose. The token now changes every auction, which is what the diagnostics correlation actually needs.

Note on gating: the `GptDiagnosticsRequestDecision` is only computed for eligible GET navigations, so per-session gating is not available on the page-bids fetch. The config-level gate is what shipped.

Tests: `page_bids_response_includes_auction_id_only_for_winning_bids` (not the EC-derived ID, does not embed the EC ID), `page_bids_auction_id_is_per_auction_and_gated_on_diagnostics` (unique per auction, absent when disabled), `diagnostics_auction_id_is_fresh_and_gated` (the shared helper).

The guide's export section now states the token is minted per auction, is not derived from the Edge Cookie ID, and does not repeat across auctions.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
prk-Jr: Changed. `responseClass` returns `reservation` only for a reservation-specific `lineItemId`/`creativeId`, or for a source-agnostic ID alongside an explicit `isBackfill === false` — that combination is still GPT evidence rather than inference. A source-agnostic ID on its own is now `unclassified_non_empty`.

Coverage added to the response-class table test for all three states: reservation-specific ID, source-agnostic plus explicit non-backfill, and source-agnostic alone. The guide states the rule and why.

crates/trusted-server-js/lib/src/core/types.ts
prk-Jr: Split. `GptDiagnosticsApi` is now `snapshot`, `export`, `subscribe`, `show`, `hide` and nothing else. The five writers moved to a new `GptDiagnosticsRecorder` type installed separately at `window.tsjs.gptDiagnosticsRecorder`; the `gpt` and `prebid` integrations call that channel.

One caveat worth stating plainly: separate IIFE bundles can only reach each other through `window.tsjs`, so the recorder is still reachable from the page. What changes is that it is no longer part of the documented operator contract and no longer expands `GptDiagnosticsApi`. The guide now says the operator API has exactly those five methods and that the recorder is internal and unsupported.

Runtime shape is pinned in three places: `api.test.ts` and `index.test.ts` assert `Object.keys()` on both objects, and `types.test.ts` asserts `keyof GptDiagnosticsApi` and `keyof GptDiagnosticsRecorder` under the new type gate.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
prk-Jr: Both bounds addressed.

Request-intent evidence no longer owns timers at all. Sources expire lazily — on the next recording for that slot and on consumption — so the per-source `defer` chain and the reschedule loop are gone. Removing them removed the no-WeakRef fallback with them, so no marked slot is strongly retained; pending intents are keyed weakly by GPT slot object and bounded by what the page itself holds. A fully expired intent is replaced rather than revived, so intent IDs keep their old semantics.

Delivery-boundary notifications now share one store-level timer. It fires at the earliest deadline, notifies, then re-arms from the earliest remaining deadline found in retained cycles, so at most one callback is outstanding regardless of refresh rate and the work is bounded by retained state.

Tests:
- 1000 publisher-refresh observations across two slots queue zero deferred callbacks, and the boundary still expires correctly at the window edge.
- Eight candidate renders in a burst leave exactly one outstanding timer; firing it re-arms once and then stops with every cycle at `candidate_unconfirmed`.

The guide's retention and timer sections were updated to match.

crates/trusted-server-js/lib/test/integrations/gpt_diagnostics/types.test.ts
prk-Jr: Confirmed, and the probe result matches — nothing was type-checking the tests.

Fixed the assertion to `GptDiagnosticsRequestPath | undefined` and enabled `test.typecheck` in `vitest.config.ts` scoped to `test/**/types.test.ts`. It needs `ignoreSourceErrors: true`: the package has 106 pre-existing errors in other test files, which surface as unhandled source errors and fail the run otherwise. Errors inside the included files still fail.

Verified the gate is live rather than decorative: injecting a deliberate mismatch fails with `Type Errors 1 failed`.

`observer.test.ts:182` (now 190) is fixed too — the local window literal is typed as `GptObserverWindow` instead of `{ tsjs?: unknown }`.

crates/trusted-server-js/lib/test/integrations/gpt_diagnostics/store.test.ts
prk-Jr: Confirmed — eleven requests left ten open cycles, so the final render matched all of them and `matchCycle` classified it ambiguous. Nothing was recorded, which is why the three `toBeUndefined()` assertions held.

Each filler request now completes with `recordSlotResponseReceived` plus an empty `recordSlotRenderEnded`, so exactly one cycle is open when the final render lands. The test asserts the earlier filled cycle was evicted and that the final render was actually matched (`renderAtMs` defined, `creativeId` 202) before asserting `replacedRequestNumber` is undefined.

docs/guide/integrations/gpt-diagnostics.md
prk-Jr: All four corrected.

- Limits: now says the integration wraps `pubads.refresh` once and does not patch publisher `display()`, and keeps the caveat about a refresh reference captured before installation.
- V1 export list: names `trustedServerAuctionId`, `requestIntentId`, `opportunityToRequestMs`, the replacement fields, `loadObservedBeforeRender`, `delivery`, and `responseClass`. The privacy paragraph now describes the auction token explicitly rather than implying no auction-level identifier is exported.
- Browser API: states the API has exactly five methods and that the writers live on a separate internal `window.tsjs.gptDiagnosticsRecorder` channel that is not part of the contract.
- Retention: covers all three intent sources and states expiry is evaluated on the next recording or request, so retained intents hold no timer.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
prk-Jr: Taken. `evictCreativeAttempt(cycle)` now runs on the empty-render branch, alongside the attribution issue that already records why.

A later `recordTrustedServerCreativeResponse` for that attempt records `creative_attempt_evicted` and writes no response timestamp, so `trusted_server_response_sent` is unreachable on an empty cycle. Delivery stays `trusted_server_selected` — the selection genuinely was observed before GPT reported empty, and the attribution issue carries the contradiction.

Covered in the provisional-empty test: response timestamp undefined, delivery `trusted_server_selected`, latest attribution issue `creative_attempt_evicted`.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
prk-Jr: Kept the label and documented what it means. The request-path table now reads `prebid_refresh` — "including a `pubads.refresh` Prebid consumed and delegated" — and `publisher_refresh` — "without Prebid consuming it".

On the install ordering: added a regression test. `it.each` installs the two wrappers in both orders against the same fixture and pins both outcomes — diagnostics first gives `prebid_refresh`, Prebid first gives `competing` — with a comment stating that only bundle evaluation order enforces this today. The dependency is now asserted rather than implied.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/observer.ts
prk-Jr: Applied, as `args.length === 0 || args[0] == null`, so both `null` and an explicit `undefined` fall back to `getSlots()`.

Test covers `refresh(undefined)` and `refresh(null, { changeCorrelator: false })` each attributing GPT's full slot list, with the delegated arguments unchanged. The guide states the rule.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts
prk-Jr: Done, with one wrinkle worth recording.

The dead `not_applicable` arm is gone. `requestPathFact` and `trustedServerOpportunityFact` return `string`, so `case undefined:` with no default is enough on its own — a new member makes the function fall through and fail to return.

`deliveryFact`, `responseClassFact`, and `badges.deliveryLabel` return `string | undefined`, where a fall-through is silently legal, so the same pattern would not have caught anything. Their default arms now route through `unhandledCase(value: never)` in a new `exhaustive.ts`, which makes a new union member a compile error while leaving the runtime result unchanged.

Did not add `@typescript-eslint/switch-exhaustiveness-check` — it needs the type-aware lint setup the config does not have yet, which is a larger change than this PR.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/api.ts
prk-Jr: Pinned the shape in the tests rather than rewriting the builder, per your cheaper option: `Object.keys(...).sort()` assertions on the snapshot root, a slot, a cycle, and `metadata`, so any new field forces a deliberate test update before it can ship in a download.

The blocklist regexes in `types.test.ts` are gone. The remaining regex assertions are the ones that check a specific value is absent from rendered output, not name-based guesses about the schema.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts
prk-Jr: Switched to `cycle.delivery` alone — both timestamp pre-checks are gone, with a comment stating the store owns the ladder.

The test that pinned the old behaviour is rewritten to assert the badge follows the derived state, including the case you predicted: a cycle carrying both creative timestamps but `isEmpty: true` and `delivery: 'not_applicable'` now renders `Empty`, not `TS response sent`. The store-side gap that made that reachable is fixed in the empty-render thread.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
prk-Jr: Added `keeps the outer dispatch context set across a nested Prebid refresh`: `originalRefresh` re-enters `pubads.refresh` once, and the test asserts the flag is still `true` immediately after the inner call returns, that the delegated call ran twice, and that the property is absent after the outer dispatch completes.

Checked it is not vacuous — forcing the delete arm to run unconditionally makes it fail with `expected [ undefined ] to deeply equal [ true ]`, so it does cover the `hadOwnContext` branch specifically.

crates/trusted-server-js/lib/src/core/types.ts
prk-Jr: Both done.

`attributionIssues` and `metadata.droppedAttributionIssues` are required on `GptDiagnosticsExportV1`, and the `?? []` / `?? 0` dead branches in `api.ts` are gone with them.

`GptDiagnosticsSlotLike` is now a type alias for `GptDiagnosticsSlotHandle` rather than a structurally identical duplicate, so the writer signatures and the store implementation cannot drift.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
prk-Jr: Agreed they are unreachable — kept both and marked them defensive, with comments explaining why the earlier check always returns first and what each one guards against if the attempt lifecycle changes. No coverage will be requested for them now.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts
prk-Jr: Changed to `Ad Manager response class: reservation`.

crates/trusted-server-js/lib/src/core/types.ts
```diff
@@ -159,14 +257,23 @@ export interface GptDiagnosticsExportV1 {
   };
   slots: GptDiagnosticsSlotExport[];
   callbackIssues: GptDiagnosticsCallbackIssue[];
+  attributionIssues: GptDiagnosticsAttributionIssue[];
```
ChristianPavilonis: **P1 — Preserve the V1 export's source compatibility.** `attributionIssues` did not exist on `GptDiagnosticsExportV1` in `main`, so requiring it (and `metadata.droppedAttributionIssues` below) makes existing V1 object literals fail TypeScript compilation. The implementation plan explicitly calls for these additions to be optional while snapshots continue to emit them.

**Suggested fix:** Mark both fields optional and add a type test that constructs the legacy V1 shape without them.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/api.ts
```diff
@@ -145,9 +221,20 @@ export class GptDiagnosticsApiController {
...
+      // keeps the cost independent of the subscriber count.
+      let snapshot: GptDiagnosticsExportV1;
+      try {
+        snapshot = this.snapshot();
+      } catch {
+        // A snapshot failure must not escape the scheduled callback.
+        return;
+      }
+
       for (const listener of this.listeners) {
```
ChristianPavilonis, line 272: **P2 — Isolate subscribers from one another's mutations.** Every listener receives this same mutable snapshot object. A listener registered first can mutate nested evidence before a later listener runs, so the later subscriber observes data that was not captured by diagnostics.

**Suggested fix:** Capture once for a consistent `capturedAt`, then deep-clone per listener (or deep-freeze the shared value), with a two-listener regression test.

ChristianPavilonis (approved):
## Review Summary

Reviewed the current PR #997 head against `main`, including the final fixes for earlier feedback. The evidence model and safety hardening are strong, and CI is green. I found one source-compatibility regression and one subscriber-isolation issue; neither warrants blocking this otherwise solid change.

crates/trusted-server-js/lib/vitest.config.ts
```diff
@@ -24,6 +24,17 @@ export default defineConfig({
...
+    // The GPT diagnostics export contract is expressed as `expectTypeOf`
+    // assertions, which `vitest run` alone never evaluates. Scope the type
+    // gate to those files: a package-wide `tsc --noEmit` still fails on
+    // pre-existing errors elsewhere.
+    typecheck: {
+      enabled: true,
+      include: ['test/**/types.test.ts'],
+      // Errors reported from files outside `include` are pre-existing and
+      // unrelated; only the type assertions in the included files gate here.
+      ignoreSourceErrors: true,
```
aram356, line 36: 🌱 **seedling** - With `ignoreSourceErrors: true` and only `types.test.ts` in scope, the new `unhandledCase` never-guards in overlay.ts/badges.ts surface in the IDE but not in CI: a new `GptDiagnosticsDelivery`/`GptDiagnosticsRequestPath`/`GptDiagnosticsResponseClass` member compiles, builds, and passes this gate. The union memberships pinned in types.test.ts today are only `GptDiagnosticsAttributionIssueReason` and the API/recorder keys. A follow-up that pins the remaining unions the same way (`expectTypeOf<GptDiagnosticsDelivery>().toEqualTypeOf<...>()`) would make a member addition fail CI and point the author at the switches.

docs/guide/integrations/gpt-diagnostics.md
```diff
@@ -41,52 +59,225 @@ For example:
...
+4. GPT later emitted `slotOnload` for the correlated slot.
+
+Each step proves only itself. In particular, a successful response post does not prove
+that the PUC consumed the response, and `slotOnload` is a GPT slot fact. Pixel-level
+proof would require a controlled creative acknowledgement after the inner markup runs;
+that acknowledgement is outside the zero-publisher-change design.
+
+The derived `delivery` value uses these evidence-safe meanings:
+
+| Delivery state                 | Panel wording                                                                                      |
```
aram356, line 173: 📝 **note** - This table is the shipped 7-state ladder, but the GitHub PR description still documents the earlier 5-state ladder (`trusted_server` / `other_demand`). Worth rewriting the PR body before merge so the merge-commit reference material matches what shipped.

aram356 (approved):

PR #998, opened 2026-08-04
Fix APS PUC rendering and collapsed GAM shells

### LINKED ISSUE #1030: APS render bridge lost creative diagnostics and postMessage failure handling in the #998 merge
## Context

#998 landed the nonce-bound APS renderer transport (private MessageChannel for PUC sibling-frame rendering, guarded 1x1 GAM shell resizing) and, as part of aligning the render bridge with main, its final merged diff removed the creative-diagnostics instrumentation around the APS handoff in `crates/trusted-server-js/lib/src/integrations/gpt/index.ts`. This issue tracks the observability gaps that removal left behind — the rendering behavior itself is working as intended.

This is separate from #999 (nested same-origin sandbox semantics), which concerns creatives the renderer cannot support at all.

## Gap 1: APS renderer handoffs are invisible to GPT creative diagnostics

Both APS branches of `installTsRenderBridge` (the server-bid path and the Prebid-renderer path) previously wrapped the `port.postMessage` handoff with `safelyRecordCreativeRequest` / `safelyRecordCreativeResponse` / `safelyRecordCreativeFailure`, including the `missing_render_source` and `response_post_failed` failure reasons. The merged diff removed all of it, plus the tests covering it.

Consequences:

- A blank APS placement caused by a failed or rejected renderer handoff no longer produces any diagnostics signal; debugging falls back to console spelunking.
- `trustedServerOpportunity` also no longer counts a typed `renderer` as a render source, so an APS bid with neither `adm` nor cache coordinates now classifies as `unrenderable_candidate` even when it renders fine through the renderer path. Opportunity metrics undercount APS.

## Gap 2: `port.postMessage` throw burns the one-shot renderer capability silently

The removed try/catch also changed failure semantics in the Prebid path. The renderer entry is consumed before the post:

1. `consumeApsPrebidRenderer(adId, ...)` and `recordConsumedPrebidApsId(...)` claim the one-shot capability.
2. `port.postMessage(...)` runs unguarded.

If the post throws (detached/closed port), the capability is already burned: the creative is never delivered, a repeat request for the same ad ID is rejected as a replay, and nothing is logged or recorded. Previously this was caught, logged, and recorded as `response_post_failed`.

## Suggested follow-up

- Reintroduce request/response/failure recording around both APS handoff paths, or decide explicitly that APS is out of scope for the GPT creative diagnostics recorder and document that in the diagnostics docs.
- Restore the guard around `port.postMessage` (or move capability consumption after a successful post) so a throwing port does not silently consume the one-shot renderer.
- Revisit `trustedServerOpportunity` classification for renderer-only bids so APS opportunities are not undercounted.
- Restore the deleted regression tests (`response_post_failed`, renderer-typed renderable candidate) alongside whichever behavior is chosen.

## References

- #998 — merged PR (the final diff vs main is the 2-file alignment residue; the transport itself landed via the #963/#988 chain)
- #999 — separate open issue on nested same-origin creative semantics
- `crates/trusted-server-js/lib/src/integrations/gpt/index.ts` — `installTsRenderBridge` APS branches, `trustedServerOpportunity`
- `crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts` — deleted diagnostics regressions

ChristianPavilonis:
## Summary
- establish nonce-bound APS renderer communication over a private MessageChannel for PUC sibling-frame rendering while retaining the legacy direct renderer path
- resize only the authenticated, ordinary collapsed 1x1 GAM shell after APS rendering
- add PUC 1.17.2 browser coverage and focused transport/geometry regressions

## Validation
- cargo test-fastly
- cargo clippy-fastly
- cargo fmt --all -- --check
- cd crates/trusted-server-js/lib && npx vitest run
- TEST_FRAMEWORK=nextjs npx playwright test tests/shared/aps-renderer.spec.ts
- git diff --check

### DIFF
--- a/crates/trusted-server-js/lib/src/integrations/gpt/index.ts
+++ b/crates/trusted-server-js/lib/src/integrations/gpt/index.ts
@@ -60,11 +60,8 @@ function trustedServerOpportunity(bid: AuctionBidData): GptDiagnosticsTrustedSer
   const hasAdId = isNonEmptyString(bid.hb_adid);
   const hasInline = isNonEmptyString(bid.adm);
   const hasCache = isNonEmptyString(bid.hb_cache_host) && isNonEmptyString(bid.hb_cache_path);
-  const hasRenderer = bid.renderer !== undefined;
 
-  return hasAdId && (hasInline || hasCache || hasRenderer)
-    ? 'renderable_candidate'
-    : 'unrenderable_candidate';
+  return hasAdId && (hasInline || hasCache) ? 'renderable_candidate' : 'unrenderable_candidate';
 }
 
 // ------------------------------------------------------------------
@@ -1671,7 +1668,6 @@ export function installTsRenderBridge(): void {
       return;
     }
 
-    const sourceSlotId = slotIdForMessageSource(e.source);
     const prebidRendererEntry = getApsPrebidRenderer(adId);
     if (prebidRendererEntry) {
       // Fail closed for a TS-owned APS ad ID before checking its source. Native
@@ -1681,35 +1677,23 @@ export function installTsRenderBridge(): void {
       if (!messageSourceBelongsToAdUnit(e.source, prebidRendererEntry.adUnitCode)) return;
       const renderer = validateApsRenderer(prebidRendererEntry.renderer);
       const rendererUrl = apsRendererUrl();
-      if (!renderer || !rendererUrl) {
-        const attemptId = sourceSlotId ? safelyRecordCreativeRequest(sourceSlotId) : undefined;
-        safelyRecordCreativeFailure(attemptId, 'missing_render_source');
-        return;
-      }
+      if (!renderer || !rendererUrl) return;
       if (!hasConsumedPrebidApsIdCapacity(consumedPrebidApsIds, adId)) return;
       if (!consumeApsPrebidRenderer(adId, prebidRendererEntry)) return;
       recordConsumedPrebidApsId(consumedPrebidApsIds, adId, prebidRendererEntry.expiresAt);
 
-      const attemptId = sourceSlotId ? safelyRecordCreativeRequest(sourceSlotId) : undefined;
-      try {
-        port.postMessage(
-          JSON.stringify({
-            message: 'Prebid Response',
-            adId,
-            renderer: APS_UNIVERSAL_CREATIVE_RENDERER,
-            rendererVersion: APS_UNIVERSAL_CREATIVE_RENDERER_VERSION,
-            rendererUrl,
-            apsRenderer: renderer,
-            width: renderer.width,
-            height: renderer.height,
-          })
-        );
-      } catch (err) {
-        safelyRecordCreativeFailure(attemptId, 'response_post_failed');
-        log.warn(`[tsjs-gpt] APS Prebid response post failed for '${adId}'`, err);
-        return;
-      }
-      safelyRecordCreativeResponse(attemptId);
+      port.postMessage(
+        JSON.stringify({
+          message: 'Prebid Response',
+          adId,
+          renderer: APS_UNIVERSAL_CREATIVE_RENDERER,
+          rendererVersion: APS_UNIVERSAL_CREATIVE_RENDERER_VERSION,
+          rendererUrl,
+          apsRenderer: renderer,
+          width: renderer.width,
+          height: renderer.height,
+        })
+      );
 
       try {
         prebidRendererEntry.markUsed();
@@ -1719,6 +1703,7 @@ export function installTsRenderBridge(): void {
       return;
     }
 
+    const sourceSlotId = slotIdForMessageSource(e.source);
     if (!sourceSlotId) return;
 
     // Resolve the bid by the requesting slot, not by the first bid whose hb_adid
@@ -1743,33 +1728,20 @@ export function installTsRenderBridge(): void {
       if (consumedServerApsBySlot.get(slotId) === adId) return;
       const renderer = validateApsRenderer(matchedBid.renderer);
       const rendererUrl = apsRendererUrl();
-      if (!renderer || !rendererUrl) {
-        const attemptId = safelyRecordCreativeRequest(slotId);
-        safelyRecordCreativeFailure(attemptId, 'missing_render_source');
-        return;
-      }
+      if (!renderer || !rendererUrl) return;
       consumedServerApsBySlot.set(slotId, adId);
-
-      const attemptId = safelyRecordCreativeRequest(slotId);
-      try {
-        port.postMessage(
-          JSON.stringify({
-            message: 'Prebid Response',
-            adId,
-            renderer: APS_UNIVERSAL_CREATIVE_RENDERER,
-            rendererVersion: APS_UNIVERSAL_CREATIVE_RENDERER_VERSION,
-            rendererUrl,
-            apsRenderer: renderer,
-            width: renderer.width,
-            height: renderer.height,
-          })
-        );
-      } catch (err) {
-        safelyRecordCreativeFailure(attemptId, 'response_post_failed');
-        log.warn(`[tsjs-gpt] APS server response post failed for '${slotId}'`, err);
-        return;
-      }
-      safelyRecordCreativeResponse(attemptId);
+      port.postMessage(
+        JSON.stringify({
+          message: 'Prebid Response',
+          adId,
+          renderer: APS_UNIVERSAL_CREATIVE_RENDERER,
+          rendererVersion: APS_UNIVERSAL_CREATIVE_RENDERER_VERSION,
+          rendererUrl,
+          apsRenderer: renderer,
+          width: renderer.width,
+          height: renderer.height,
+        })
+      );
       return;
     }
 
--- a/crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
+++ b/crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts
@@ -262,15 +262,6 @@ describe('installTsAdInit', () => {
       'renderable_candidate',
     ],
     [
-      'typed renderer',
-      {
-        hb_pb: '1.00',
-        hb_adid: apsRenderer().bidId,
-        renderer: apsRenderer(),
-      },
-      'renderable_candidate',
-    ],
-    [
       'an ad ID without a render source',
       { hb_pb: '1.00', hb_adid: 'abc-uuid' },
       'unrenderable_candidate',
@@ -3125,14 +3116,6 @@ describe('installTsRenderBridge', () => {
       hb_cache_host: 'cache.example.com',
       hb_cache_path: '/cache',
     };
-    const recordTrustedServerCreativeRequest = vi.fn().mockReturnValue(47);
-    const recordTrustedServerCreativeResponse = vi.fn();
-    const recordTrustedServerCreativeFailure = vi.fn();
-    (window as TestWindow).tsjs.gptDiagnosticsRecorder = {
-      recordTrustedServerCreativeRequest,
-      recordTrustedServerCreativeResponse,
-      recordTrustedServerCreativeFailure,
-    } as unknown as TsjsApi['gptDiagnosticsRecorder'];
 
     const beaconSpy = vi.spyOn(navigator, 'sendBeacon').mockReturnValue(true);
     const bridgeListener = await captureBridgeListener();
@@ -3153,10 +3136,6 @@ describe('installTsRenderBridge', () => {
     expect(stopSpy).toHaveBeenCalledTimes(2);
     expect(fetchStub).not.toHaveBeenCalled();
     expect(beaconSpy).not.toHaveBeenCalled();
-    expect(recordTrustedServerCreativeRequest).toHaveBeenCalledOnce();
-    expect(recordTrustedServerCreativeRequest).toHaveBeenCalledWith('homepage_header');
-    expect(recordTrustedServerCreativeResponse).toHaveBeenCalledWith(47);
-    expect(recordTrustedServerCreativeFailure).not.toHaveBeenCalled();
     // Server-rendered APS capabilities are one-shot per slot and ad ID. A
     // repeated Universal Creative request is claimed but receives no payload.
     expect(portMessages).toHaveLength(1);
@@ -3231,14 +3210,6 @@ describe('installTsRenderBridge', () => {
         markUsed,
       },
     };
-    const recordTrustedServerCreativeRequest = vi.fn().mockReturnValue(48);
-    const recordTrustedServerCreativeResponse = vi.fn();
-    const recordTrustedServerCreativeFailure = vi.fn();
-    (window as TestWindow).tsjs.gptDiagnosticsRecorder = {
-      recordTrustedServerCreativeRequest,
-      recordTrustedServerCreativeResponse,
-      recordTrustedServerCreativeFailure,
-    } as unknown as TsjsApi['gptDiagnosticsRecorder'];
 
     const bridgeListener = await captureBridgeListener();
     const source = createTrustedSlotIframe();
@@ -3278,71 +3249,8 @@ describe('installTsRenderBridge', () => {
     expect(renderer.bidId).not.toBe(prebidAdId);
     expect((window as TestWindow).tsjs.apsPrebidRenderers[prebidAdId]).toBeUndefined();
     expect(fetchStub).not.toHaveBeenCalled();
-    expect(recordTrustedServerCreativeRequest).toHaveBeenCalledOnce();
-    expect(recordTrustedServerCreativeRequest).toHaveBeenCalledWith('homepage_header');
-    expect(recordTrustedServerCreativeResponse).toHaveBeenCalledWith(48);
-    expect(recordTrustedServerCreativeFailure).not.toHaveBeenCalled();
     foreignIframe.remove();
   });
-
-  it.each(['server', 'Prebid'] as const)(
-    'records response_post_failed when posting a %s APS renderer throws',
-    async (sourceKind) => {
-      const renderer = apsRenderer();
-      const adId = sourceKind === 'server' ? renderer.bidId : 'prebid-post-failure-ad-id';
-      const markUsed = vi.fn();
-      if (sourceKind === 'server') {
-        (window as TestWindow).tsjs.bids.homepage_header = {
-          hb_adid: adId,
-          hb_bidder: 'aps',
-          renderer,
-        };
-      } else {
-        (window as TestWindow).tsjs.apsPrebidRenderers = {
-          [adId]: {
-            adUnitCode: 'div-header',
-            renderer,
-            registeredAt: Date.now(),
-            expiresAt: Date.now() + 60_000,
-            markUsed,
-          },
-        };
-      }
-
-      const recordTrustedServerCreativeRequest = vi.fn().mockReturnValue(50);
-      const recordTrustedServerCreativeResponse = vi.fn();
-      const recordTrustedServerCreativeFailure = vi.fn();
-      (window as TestWindow).tsjs.gptDiagnosticsRecorder = {
-        recordTrustedServerCreativeRequest,
-        recordTrustedServerCreativeResponse,
-        recordTrustedServerCreativeFailure,
-      } as unknown as TsjsApi['gptDiagnosticsRecorder'];
-
-      const bridgeListener = await captureBridgeListener();
-      const source = createTrustedSlotIframe();
-      expect(() =>
-        bridgeListener(
-          Object.assign(new Event('message'), {
-            data: JSON.stringify({ message: 'Prebid Request', adId }),
-            ports: [
-              {
-                postMessage: vi.fn(() => {
-                  throw new Error('port closed');
-                }),
-              },
-            ],
-            source,
-            stopImmediatePropagation: vi.fn(),
-          }) as unknown as MessageEvent
-        )
-      ).not.toThrow();
-
-      expect(recordTrustedServerCreativeRequest).toHaveBeenCalledWith('homepage_header');
-      expect(recordTrustedServerCreativeFailure).toHaveBeenCalledWith(50, 'response_post_failed');
-      expect(recordTrustedServerCreativeResponse).not.toHaveBeenCalled();
-      expect(markUsed).not.toHaveBeenCalled();
-    }
-  );
 
   it('still serves the APS renderer when markUsed throws', async () => {
     const renderer = apsRenderer();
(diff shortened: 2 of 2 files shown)


crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -1121,25 +1203,6 @@ export function installTsRenderBridge(): void {
...
-          renderer: APS_UNIVERSAL_CREATIVE_RENDERER,
-          rendererVersion: APS_UNIVERSAL_CREATIVE_RENDERER_VERSION,
-          rendererUrl,
-          apsRenderer: renderer,
-          width: renderer.width,
-          height: renderer.height,
-        })
-      );
-      return;
-    }
```
prk-Jr, line 1749: 🔧 **wrench** — CI format gate fails here.

Moving the `matchedBid.renderer` branch below the width/height computation left a second consecutive blank line, so `prettier --check` rejects this file and CI gate 6 (`cd crates/trusted-server-js/lib && npm run format`) fails:

```
$ npx prettier --check src/integrations/gpt/index.ts
[warn] src/integrations/gpt/index.ts
[warn] Code style issues found in the above file.

$ npx prettier src/integrations/gpt/index.ts | diff - src/integrations/gpt/index.ts
1205a1206
>
```

**Fix**: delete the blank line. Note this PR's base is not `main`, so the fmt/clippy/unit-test workflows never ran on this head — only the four integration-test checks did.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -415,16 +415,20 @@ export const APS_UNIVERSAL_CREATIVE_RENDERER = String.raw`(function(){window.ren
 try{var r=d&&d.apsRenderer,u=d&&d.rendererUrl;if(!r||typeof u!=="string")throw new Error("invalid APS renderer data");
 var p=new URL(u);if((p.protocol!=="https:"&&p.protocol!=="http:")||p.username||p.password||p.pathname!=="${APS_RENDERER_PATH}"||p.search||p.hash)throw new Error("invalid APS renderer URL");
 var c=w.crypto;if(!c||typeof c.getRandomValues!=="function")throw new Error("APS renderer randomness unavailable");
+if(typeof MessageChannel!=="function")throw new Error("APS renderer channel unavailable");
 var b=new Uint8Array(16);c.getRandomValues(b);var s="";for(var i=0;i<b.length;i++)s+=String.fromCharCode(b[i]);
 var n=w.btoa(s).replace(/\+/g,"-").replace(/\//g,"_").replace(/=+$/g,"");
-var f=w.document.createElement("iframe"),done=false,t;
-function clean(){w.removeEventListener("message",receive);if(t)w.clearTimeout(t);}
+var x=w.document,b=x.body,f=x.createElement("iframe"),q=new MessageChannel(),done=false,sent=false,t;
```
prk-Jr: 🤔 **thinking** — `b` is declared twice in this renderer source.

Two lines above, `var b=new Uint8Array(16)` holds the nonce bytes; here `var b=x.body` rebinds the same name. It is correct today only because the byte buffer is dead by the time this line runs. Any future edit that moves the nonce derivation below the body capture would silently base64-encode the body element instead of the random bytes, and the resulting nonce would still match `/^[A-Za-z0-9_-]{22,128}$/` in the renderer document — so it would fail as a silent loss of entropy, not a thrown error.

**Fix**: rename one of them, e.g.

```js
var by=new Uint8Array(16);c.getRandomValues(by);var s="";for(var i=0;i<by.length;i++)s+=String.fromCharCode(by[i]);
```

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -88,32 +88,106 @@ function candidateSlotRoots(divId: string): HTMLElement[] {
...
+    !wrapper ||
+    wrapper === document.body ||
+    wrapper === document.documentElement ||
+    !frame.root.contains(wrapper) ||
+    usesFixedPositioning(wrapper)
+  ) {
+    return;
+  }
+
+  frame.iframe.style.width = `${width}px`;
```
prk-Jr: 🤔 **thinking** — the inline sizing written here is permanent and never cleared.

`frame.iframe.style.width/height` (and the wrapper's, below) are inline styles, which beat GAM's presentational `width`/`height` attributes. Two consequences on a GPT slot refresh that reuses the same iframe element:

1. GAM re-sizes the shell via attributes for the next creative — the stale inline `300px × 250px` wins and the slot renders at the previous creative's box.
2. GAM collapses the slot back to 1x1 via attributes — the computed size is still 300px, so `hasOnePixelDimension` now returns `false` and this function can never re-arm for that slot.

**Suggestion**: record what was set and clear it when the frame is observed re-collapsing (attributes back to `1`), or scope the resize to the APS renderer path where the collapsed-shell topology is the actual problem being fixed.

crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -105,9 +104,29 @@ function receive(event){
...
+}
+function receive(event){
+ var message=event.data;
+ if(keys(message,['nonce','renderer'])){
+  if(event.source!==parent||event.ports.length!==0||message.nonce!==expected)return;
+  var nonce=expected;
+  start(message.renderer,function(result){parent.postMessage({message:result,nonce:nonce},'*');});
+  return;
+ }
+ if(!keys(message,['nonce'])||event.ports.length!==1||message.nonce!==expected)return;
```
prk-Jr: 🤔 **thinking** — the MessagePort branch drops the parent-source check the direct branch keeps.

Three lines above, the direct branch guards with `event.source!==parent||event.ports.length!==0`. This port branch checks only key shape, port count and nonce. The 128-bit fragment nonce is unreadable cross-origin, so this is not exploitable — but the asymmetry reads as a deliberate relaxation, and the accompanying test now asserts only `event.ports.length!==1` for this branch.

**Fix**: keep the two branches identical in intent for one extra comparison.

```js
if(!keys(message,['nonce'])||event.source!==parent||event.ports.length!==1||message.nonce!==expected)return;
```

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -1081,32 +1155,40 @@ export function installTsRenderBridge(): void {
       // Prebid handles ad IDs globally and would otherwise answer a request from
       // an unrelated iframe when this slot-bound capability rejects it.
       e.stopImmediatePropagation();
-      if (!messageSourceBelongsToAdUnit(e.source, prebidRendererEntry.adUnitCode)) return;
+      const sourceFrame = sourceFrameForAdUnit(e.source, prebidRendererEntry.adUnitCode);
```
prk-Jr: 🤔 **thinking** — the capability is consumed before the response is known to be delivered.

In this branch the order is `stopImmediatePropagation()` → `consumeApsPrebidRenderer()` → `markWinner()` → `port.postMessage(...)`. The new `catch` below logs and returns, which leaves the slot in a terminal bad state: the one-shot renderer entry is already consumed, the bid is marked winner, Prebid's own handler was already suppressed, and `markRendered()` never runs. No retry path exists — the slot stays blank.

The same shape applies to the sibling branch further down, which also consumes before posting.

**Suggestion**: post first and consume on success, or restore the entry in the `catch` before returning. A `MessagePort.postMessage` of a string realistically only throws on a detached/closed port, but that is exactly the case where the fallback matters.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -1149,18 +1212,55 @@ export function installTsRenderBridge(): void {
...
+            adId,
+            renderer: APS_UNIVERSAL_CREATIVE_RENDERER,
+            rendererVersion: APS_UNIVERSAL_CREATIVE_RENDERER_VERSION,
+            rendererUrl,
+            apsRenderer: renderer,
+            width: renderer.width,
+            height: renderer.height,
+          })
+        );
+        resizeCollapsedCreativeFrame(e.source, sourceSlotFrame, renderer.width, renderer.height);
```
prk-Jr: ⛏ **nitpick** — inconsistent `try` scope between the two response paths.

Here `resizeCollapsedCreativeFrame` and `log.debug` sit inside the `try`, so a throw from the resize logs `pbRender bridge: APS response failed` even though the response was delivered successfully — misleading during triage. The `matchedBid.adm` branch below keeps its resize outside the `try`.

**Fix**: match the `adm` branch — keep only `port.postMessage` inside the `try`, and run the resize and the debug log after it.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -88,32 +88,106 @@ function candidateSlotRoots(divId: string): HTMLElement[] {
...
+  return sourceFrameInRoots(source, candidateSlotRoots(adUnitCode));
+}
+
+function hasOnePixelDimension(element: HTMLElement, dimension: 'width' | 'height'): boolean {
+  const computedDimension = window.getComputedStyle(element)[dimension];
+  const match = /^(\d+(?:\.\d+)?)px$/.exec(computedDimension);
+  return match !== null && Number(match[1]) <= 1;
+}
+
+function usesFixedPositioning(element: HTMLElement): boolean {
```
prk-Jr: ⛏ **nitpick** — `usesFixedPositioning` also matches `sticky`.

The name says `fixed` but the predicate covers both, which is the right behaviour for anchor/rail ads. Renaming to `usesPinnedPositioning` (or adding a one-line doc comment) keeps the call sites at lines 172 and 182 self-explanatory.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -88,32 +88,106 @@ function candidateSlotRoots(divId: string): HTMLElement[] {
...
+  const position = window.getComputedStyle(element).position;
+  return position === 'fixed' || position === 'sticky';
+}
+
+/** Resize only an authenticated, still-collapsed ordinary display-creative shell. */
+function resizeCollapsedCreativeFrame(
```
prk-Jr: 👍 **praise** — the guard set here fails closed at every step.

Requiring both the `width`/`height` attributes *and* the computed dimensions to be ≤1px, re-checking `isConnected` and `contentWindow === source` after the async hop, rejecting `fixed`/`sticky` frames, and carving out `ins[data-anchor-status]` means a hostile or merely unusual frame gets ignored rather than resized. The wrapper pass repeats the same 1x1 test before touching the parent. Good defensive design for code that mutates publisher DOM.

crates/trusted-server-integration-tests/browser/tests/shared/aps-renderer.spec.ts
```diff
@@ -192,6 +205,230 @@ const SCRIPT_CREATIVE = `(function(){
 })();`;
 
 test.describe("APS opaque renderer", () => {
+    test("renders APS through the real PUC dynamic-renderer topology and resizes only its 1x1 shell", async ({
```
prk-Jr: 👍 **praise** — this is the right shape of test for this bug class.

Driving the real PUC 1.17.2 `banner.js` through its dynamic-renderer branch (by deliberately leaving `pbjs` absent) reproduces the actual sibling-frame topology instead of a hand-rolled approximation, and the assertions cover geometry at all three frame levels plus the outer shell. The `#div-other` sibling slot that must stay 1x1 is what turns this from a happy-path test into a real regression net for the resize guard.

prk-Jr (changes requested):
## Summary

Moves APS PUC delivery onto a nonce-bound private `MessageChannel` while keeping the legacy direct-`postMessage` path in the renderer document, and adds a narrowly-guarded resize for collapsed 1x1 GAM shells. The transport rework is backward-compatible by construction and the new PUC 1.17.2 browser test is a genuine regression net. One blocking item: the JS format gate fails on a stray blank line. One question on renderer versioning.

## Blocking

### 🔧 wrench

- **JS format gate fails**: stray double blank line left by the moved APS branch — `prettier --check` rejects `src/integrations/gpt/index.ts`, failing CI gate 6 (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts:1206`, inline).

### ❓ question

- **Renderer version not bumped despite a wire-protocol change** (`crates/trusted-server-js/lib/src/integrations/aps/render.ts:7`): `APS_UNIVERSAL_CREATIVE_RENDERER_VERSION` stays `4` while `APS_UNIVERSAL_CREATIVE_RENDERER` switches from a direct `postMessage({nonce, renderer})` to a transferred `MessageChannel`. PUC keys its cached dynamic renderer on `rendererVersion`. Is holding at 4 deliberate — i.e. the renderer document retains the legacy direct branch precisely so a PUC-cached v4 source keeps working against the new document? If so, a comment on that constant saying the version is intentionally pinned would make the coupling explicit; if not, it needs a bump. (Not inline: line 7 is outside the diff.)

## Non-blocking

### 🤔 thinking

- **Inline sizing is permanent across GPT refresh** (`.../gpt/index.ts:185`, inline)
- **Port branch drops the parent-source check the direct branch keeps** (`crates/trusted-server-core/src/integrations/aps.rs:120`, inline)
- **Capability consumed before the response is known to be delivered** (`.../gpt/index.ts:1158`, inline)
- **`var b` declared twice in the renderer source** (`.../aps/render.ts:421`, inline)

### ⛏ nitpick

- **Inconsistent `try` scope between the two response paths** (`.../gpt/index.ts:1238`, inline)
- **`usesFixedPositioning` also matches `sticky`** (`.../gpt/index.ts:145`, inline)

### 📝 note

- **Resize applies to more than the APS path**: the description says "resize only the authenticated, ordinary collapsed 1x1 GAM shell **after APS rendering**", but `resizeCollapsedCreativeFrame` is also called on the inline-`adm` path and the cached-adm path. The behaviour change is wider than the summary states — worth reflecting in the PR description so the blast radius is clear to anyone bisecting a sizing regression later.
- **New browser-test dependency**: `prebid-universal-creative@1.17.2` is pinned exactly and lands as `"dev": true` (93 packages in that lockfile, all dev). Reasonable price for real PUC coverage; noting the added CI install surface (`gulp-cli`, `babel-runtime`, `core-js@2.6.12`) for awareness.

### 👍 praise

- **Conservative resize guard set** (`.../gpt/index.ts:151`, inline)
- **Real-PUC topology coverage** (`.../browser/tests/shared/aps-renderer.spec.ts:208`, inline)

## CI Status

GitHub checks on this head: 4/4 pass (browser integration tests, integration tests, integration tests (Fastly EC lifecycle), prepare integration artifacts). Because the base branch is `fix/aps-renderer-spec-decoupled-shim` rather than `main`, the fmt / clippy / unit-test workflows did not run on this PR — the results below are from running them locally at `5f29a74`.

- `cargo fmt --all -- --check`: PASS
- `cargo test-fastly aps`: PASS (60 passed)
- `npx vitest run`: PASS (528 passed, 31 files)
- `npm run lint` (eslint): PASS
- `npm run format` (prettier --check): **FAIL** — `src/integrations/gpt/index.ts`
- clippy: not run locally

(`npx tsc --noEmit` reports errors in `test/integrations/prebid/index.test.ts` and `test/shared/beacon_guard.test.ts`; those are pre-existing and unrelated to this PR.)

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -1048,6 +1203,7 @@ export function installTsRenderBridge(): void {
     // pulling slot B's creative and firing slot B's win/billing beacons.
     if (!matchedBid || matchedBid.hb_adid !== adId) return;
 
+
```
aram356, line 1749: 🔧 **wrench**: removing the old renderer branch left a double blank line here. `npx prettier --check` fails on exactly this (verified locally), so the `npm run format` CI gate will reject the PR.

**Fix**: delete one of the two blank lines.

crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -1,247 +1,207 @@
...
-    amznactt: Option<String>,
+function receive(event){
+ var message=event.data;
+ if(keys(message,['nonce','renderer'])){
+  if(event.source!==parent||event.ports.length!==0||message.nonce!==expected)return;
+  var nonce=expected;
+  start(message.renderer,function(result){parent.postMessage({message:result,nonce:nonce},'*');});
+  return;
+ }
+ if(!keys(message,['nonce'])||event.ports.length!==1||message.nonce!==expected)return;
```
aram356: 🤔 **thinking**: the port path deliberately omits the `event.source===parent` check that the legacy path above keeps. That omission is load-bearing: in PUC's sibling-frame topology the handshake sender is the hidden dynamic-renderer frame, not this frame's parent, so a source check would re-break exactly the bug this PR fixes. Nothing in the code records that, and a future hardening pass could plausibly "restore" the check.

**Suggestion**: add a short comment above this branch, e.g. `// Port path is nonce-only by design: PUC's dynamic renderer posts from a hidden sibling frame, not this frame's parent.`

crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -1,247 +1,207 @@
...
+ var port=event.ports[0];
+ if(!port)return;
+ port.onmessage=function(portEvent){
+  var descriptor=portEvent.data;
+  if(accepted||portEvent.ports.length!==0||!keys(descriptor,['renderer'])){port.close();return;}
+  var nonce=expected;
+  if(!start(descriptor.renderer,function(result){try{port.postMessage({message:result,nonce:nonce});}finally{port.close();}}))port.close();
+ };
```
aram356: ♻️ **refactor**: when the descriptor fails `validRenderer`, the frame closes the port without replying. Port closure is not observable cross-browser, so the parent waits the full 10s `RENDERER_READY_TIMEOUT_MS` before rejecting, and PUC reports `adRenderSucceeded`/`adRenderFailed` 10 seconds late.

**Suggestion**: post `trusted-server/aps/renderer-failed` on the port before closing it (and likewise return a failure reply in the legacy path), so invalid descriptors fail fast instead of timing out.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -0,0 +1,434 @@
...
+ * It reads only the validated descriptor and trusted absolute endpoint URL from data.
+ */
+export const APS_UNIVERSAL_CREATIVE_RENDERER = String.raw`(function(){window.render=function(d,_h,w){return new Promise(function(resolve,reject){
+try{var r=d&&d.apsRenderer,u=d&&d.rendererUrl;if(!r||typeof u!=="string")throw new Error("invalid APS renderer data");
+var p=new URL(u);if((p.protocol!=="https:"&&p.protocol!=="http:")||p.username||p.password||p.pathname!=="${APS_RENDERER_PATH}"||p.search||p.hash)throw new Error("invalid APS renderer URL");
+var c=w.crypto;if(!c||typeof c.getRandomValues!=="function")throw new Error("APS renderer randomness unavailable");
+if(typeof MessageChannel!=="function")throw new Error("APS renderer channel unavailable");
+var b=new Uint8Array(16);c.getRandomValues(b);var s="";for(var i=0;i<b.length;i++)s+=String.fromCharCode(b[i]);
+var n=w.btoa(s).replace(/\+/g,"-").replace(/\//g,"_").replace(/=+$/g,"");
+var x=w.document,b=x.body,f=x.createElement("iframe"),q=new MessageChannel(),done=false,sent=false,t;
```
aram356: ⛏ **nitpick**: `var b` is declared twice in this renderer source: first as the nonce byte array on line 419, then reassigned to `x.body` here. It only works because the reuse happens after the nonce is derived; a future reorder would silently corrupt the nonce.

**Suggestion**: use a different variable name for the body reference.

crates/trusted-server-integration-tests/browser/package.json
```diff
@@ -8,6 +8,7 @@
     "test:wordpress": "TEST_FRAMEWORK=wordpress npx playwright test"
   },
   "devDependencies": {
-    "@playwright/test": "^1.49.0"
+    "@playwright/test": "^1.49.0",
+    "prebid-universal-creative": "1.17.2"
```
aram356: 📝 **note**: `prebid-universal-creative@1.17.2` pulls 93 packages into the browser-test harness (including gulp-cli and core-js, which has an install script) for the single `dist/banner.js` file the spec reads. Pinned, dev-only, and integrity-hashed, so this is acceptable; if the footprint becomes a concern, consider `npm ci --ignore-scripts` in CI or vendoring just the dist file.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -1056,18 +1212,55 @@ export function installTsRenderBridge(): void {
     const width = matchedBid.w ?? fallbackWidth;
     const height = matchedBid.h ?? fallbackHeight;
 
+    if (matchedBid.renderer !== undefined) {
+      const renderer = validateApsRenderer(matchedBid.renderer);
+      const rendererUrl = apsRendererUrl();
+      if (!renderer || !rendererUrl) return;
+
+      // Ownership and the complete validated envelope are valid before this
```
aram356: ⛏ **nitpick**: this comment does not parse ("Ownership and the complete validated envelope are valid before this handler claims the message..."); please reword, e.g. "Validate ownership and the complete envelope before this handler claims the message or suppresses another legitimate handler."

Relatedly, the PBS Cache path below is now the only response path whose `port.postMessage` has no dedicated try/catch: a throw there is swallowed by the fetch `.catch` and logged as "PBS Cache fetch failed", which is misleading.

aram356 (changes requested):
## Summary

The new work here is sound: the nonce-bound MessageChannel handshake correctly fixes PUC sibling-frame rendering (where the sender is the hidden dynamic-renderer frame, so the old `event.source !== parent` check could never pass), the collapsed 1x1 GAM shell resize is tightly guarded, and the real PUC 1.17.2 end-to-end spec is strong coverage. Two blocking items: a JS format failure and the branch/base divergence below.

## Blocking

### 🔧 wrench

- **PR does not merge into its base**: `git merge-tree` against `fix/aps-renderer-spec-decoupled-shim` reports content conflicts in 15 files (aps.rs, orchestrator.rs, formats.rs, prebid.rs, publisher.rs, creative_opportunities.rs, render.ts, gpt/index.ts, types.ts, several tests/docs, plus add/add conflicts on the new files). This branch was rebuilt on a stale merge-base (`bf61fa189`) while the base chain advanced, so the two sides now hold divergent copies of the same logical commits: this branch carries the `markWinner`/`markRendered` split from the #963 feedback round while the base tip still has `markUsed`, and the base carries ~16k lines of unrelated newer work this branch lacks. As a result the PR is unmergeable as-is, and the 48-file diff attributes ~9.5k lines of already-reviewed stacked content (#918/#963/#988) to this PR. **Fix**: update the base branch to the rebased chain, or rebase this branch onto the current base tip, so the PR diff reduces to the actual new work (9 files, +1,628/−98).
- **JS format gate fails** on `crates/trusted-server-js/lib/src/integrations/gpt/index.ts` (see inline comment): CI gate 6 (`npm run format`) rejects the double blank line left where the old renderer branch was removed.

## CI Status

- GitHub checks: integration tests (4 jobs) PASS
- Local vitest: 528/528 PASS
- Local `cargo test` (wasm32-wasip1, aps-filtered): 60/60 PASS
- JS format: FAIL (`gpt/index.ts`)
- fmt/clippy: not run locally

aram356:
@ChristianPavilonis PR needs attention

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Fixed in dd58a809. The extra blank line is removed, and the JS formatting check now passes.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
ChristianPavilonis: Fixed in dd58a809. The nonce byte buffer is now named `nonceBytes`, and the document body is named `body`.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Addressed in dd58a809. Resizing is scoped to APS, and the helper tracks and restores only the inline styles it applied when the shell returns to 1x1. Added refresh coverage.

crates/trusted-server-core/src/integrations/aps.rs
ChristianPavilonis: Kept the nonce-only port check intentionally: PUC's dynamic renderer posts from a hidden sibling frame, not this document's parent. Added an explanatory comment in dd58a809.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Fixed in dd58a809. If delivery fails after consumption, the exact APS capability is restored and its tombstone is removed so the request can retry. Added regression coverage.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Fixed in dd58a809. The APS try block now covers only `port.postMessage()`; resizing and success logging happen afterward.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Fixed in dd58a809 by renaming `usesFixedPositioning` to `usesPinnedPositioning`.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Thanks. The defensive guards remain, with the APS-only refresh-safe sizing behavior added in dd58a809.

crates/trusted-server-integration-tests/browser/tests/shared/aps-renderer.spec.ts
ChristianPavilonis: Thanks. The real PUC dynamic-renderer coverage remains, and the renderer-version expectation now reflects version 5.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Fixed in dd58a809. The extra blank line is removed and the Prettier check passes.

crates/trusted-server-core/src/integrations/aps.rs
ChristianPavilonis: Added the requested comment in dd58a809 documenting why the port path intentionally relies on nonce validation instead of the parent-source check.

crates/trusted-server-core/src/integrations/aps.rs
ChristianPavilonis: Fixed in dd58a809. Invalid descriptors now send `renderer-failed` before closing the port, and the legacy path sends the same failure response. APS tests pass.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
ChristianPavilonis: Fixed in dd58a809 by using distinct names for the nonce bytes and document body.

crates/trusted-server-integration-tests/browser/package.json
ChristianPavilonis: Acknowledged. The dependency remains pinned, dev-only, and integrity-locked for real PUC coverage; the browser integration suite passes.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Fixed in dd58a809. The comment was reworded, and PBS Cache response delivery now has its own try/catch so failures are logged accurately and do not fire beacons.

ChristianPavilonis:
Addressed the review feedback in commit `dd58a809` and pushed the rebased branch.

- Fixed formatting and bumped the renderer protocol version to 5 while retaining v4 compatibility.
- Scoped and made APS sizing refresh-safe.
- Added rollback on APS response-delivery failure.
- Added fast failure replies for invalid renderer descriptors.
- Cleaned up naming, comments, and PBS Cache delivery error handling.

Validation passed: full Vitest (623 tests), browser integration suite, `cargo fmt --all -- --check`, and `cargo test-fastly aps` (62 tests). Re-requested review from `prk-Jr` and `aram356`.

prk-Jr (approved):
## Summary

Reviewed the nonce-bound APS renderer handshake, v5 renderer contract, guarded collapsed-shell sizing, and real PUC regression coverage. No findings.

## CI Status

- integration tests: PASS
- cargo fmt --all -- --check: PASS (local)
- JS format: PASS (local)
- dependency audit: PASS (0 vulnerabilities)

aram356 (approved):

PR #1010, opened 2026-08-07
Add deploying-trusted-server-to-fastly skill

jevansnyc:
## What

Adds a repo Claude Code skill, `.claude/skills/deploying-trusted-server-to-fastly/`, documenting how to deploy a TS service to Fastly Compute, and specifically how to add a second (or Nth) service to a Fastly account that already runs one without clobbering another service's config. Also extends the `.claude/*` gitignore allowlist to include `.claude/skills/` (alongside the existing `commands/` and `agents/`).

Supersedes the earlier CLAUDE.md-only version of this doc: a skill is discoverable and invocable, and keeps CLAUDE.md from growing.

## Why

Fastly config/secret/KV store names are account-level, and TS opens stores by fixed logical names (`trusted_server_config`, `ec_identity_store`, the configured secret stores). Two TS services sharing one Fastly account that both bind the default physical store names collide: `ts config push` resolves the target store by name and overwrites the other service's config. Hit this in practice standing up a second service alongside an existing one.

## The pattern (in the skill)

Decouple the physical store name from the logical name TS opens: create a service-scoped physical store (`<service>_ts_config`), link it under the logical name via `fastly resource-link create --name trusted_server_config`, and push with `EDGEZERO__STORES__CONFIG__TRUSTED_SERVER_CONFIG__NAME=<service>_ts_config ts config push`. Guardrails: `--dry-run` to confirm the target store, always `--service-id` on publish, verify the shared store's timestamp is untouched.

Verified the skill is followable by a fresh agent applying it to a clean scenario.


ChristianPavilonis (changes requested):
Please make this a skill. CLAUDE.md is loaded into context on every session adding unnecessary context to development.

ChristianPavilonis (approved):

PR #1032, opened 2026-08-13
Show requested, fill, and observed slot sizes in GPT diagnostics

### LINKED ISSUE #1031: Show requested, fill, and observed slot sizes in GPT diagnostics
## Problem

Flexible creatives, especially APS, can report `1×1` through GPT `slotRenderEnded.size` even when the publisher-visible slot is larger after rendering. Operators need to distinguish the requested slot formats, GPT’s reported fill size, and outer-slot layout geometry.

## Proposed behavior

- Preserve the full configured size list Trusted Server supplied to GPT for the correlated request.
- Preserve GPT’s `slotRenderEnded.size` as the reported fill-size fact.
- For uniquely bound, connected, filled slots, record and display the observed outer CSS slot box after render.
- Keep the observed box updated on resize when supported.
- Reject measurements for ambiguous, unbound, empty, or stale refresh cycles.
- Export all three values and document that cross-origin/SafeFrame content cannot be measured internally.

## Acceptance criteria

- `ts_console` displays the complete requested size list, GPT-reported fill size, and an optional observed outer slot box.
- A GPT-reported `1×1` and a distinct observed outer box are visible together.
- Delayed measurements cannot update a newer request cycle.
- No APS sandbox or creative-iframe access is added.

ChristianPavilonis:
## Summary
- retain each correlated Trusted Server configured slot-size list
- retain GPT `slotRenderEnded.size` as the reported fill size
- retain the measured outer slot box to explain flexible/APS `1×1` fills
- display and export all three metrics, with stale-cycle guards and SafeFrame limitations documented

Closes #1031

## Validation
- `cd crates/trusted-server-js/lib && npx vitest run` (45 files, 838 tests)
- focused GPT/diagnostics tests (283 passed)
- `cd crates/trusted-server-js/lib && npm run format`
- `git diff --check`


crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -1053,20 +1053,13 @@ export function installTsAdInit(): void {
...
-              slot.id,
-              opportunity
-            );
-          }
+          ts.gptDiagnosticsRecorder?.recordTrustedServerOpportunity(
+            gptSlot,
+            slot.id,
+            opportunity,
+            bid.hb_auction_id,
+            slot.formats
```
prk-Jr: 🔧 **wrench** — `requestedSlotSizes` records sizes GPT never received on publisher-defined slots.

`slot.formats` is passed unconditionally, but Trusted Server only supplies formats to GPT on the `defineSlot` path above (line 1023). When `existingSlot` is found — the publisher defined the div first — TS reuses that slot and never applies sizes to it; there is no `defineSizeMapping`/`setSizes` anywhere in the integration. On those slots the panel shows `Requested slot sizes 300×250, 728×90` for a list GPT never saw, and the docs assert it is "the configured `AuctionSlot.formats` list Trusted Server supplied to GPT for that request". That is the exact misdirection #1031 exists to remove.

Note `tsOwned ? slot.formats : undefined` is not the fix either: refresh cycles re-find TS's own slot through `getSlots()`, so `tsOwned` is false there and the accurate list would be dropped. Keying on the handoff record survives refreshes:

```ts
const tsDefinedFormats = ts.gptSlotHandoffs?.[actualDivId]?.formats;
ts.gptDiagnosticsRecorder?.recordTrustedServerOpportunity(
  gptSlot,
  slot.id,
  opportunity,
  bid.hb_auction_id,
  tsDefinedFormats
);
```

Alternatively keep recording it but relabel the field, badge, panel, and docs as Trusted Server's *configured* sizes rather than what GPT was given.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
```diff
@@ -0,0 +1,146 @@
...
+  private refresh(): void {
+    if (this.destroyed) return;
+    this.resizeObserver?.disconnect();
+    const observations = new Map<HTMLElement, ObservedCycle>();
+    const ResizeObserverConstructor = this.window.ResizeObserver;
+    if (typeof ResizeObserverConstructor === 'function') {
+      this.resizeObserver = new ResizeObserverConstructor((entries) => {
+        for (const entry of entries) {
+          const element = entry.target;
+          if (!(element instanceof this.window.HTMLElement)) continue;
```
prk-Jr, line 100: 🔧 **wrench** — `instanceof this.window.HTMLElement` throws for a window that satisfies the declared type.

`SlotSizeWindow` (line 17) never declares `HTMLElement`, and lib.dom's `Window` does not carry it — only `Window & typeof globalThis` does. Injecting `{ ResizeObserver }` throws `TypeError: Right-hand side of 'instanceof' is not an object`, uncaught, from inside the `ResizeObserver` callback, which contradicts the module's "diagnostics never interrupt the page" invariant. `tsc` already reports three errors here (105–107). `binding.ts` gets this right by declaring `HTMLElement: typeof HTMLElement` on its window type.

Not reachable through `installGptDiagnosticsRuntime`, which passes a real window, so downgrade this if you prefer — but the fix is one line:

```ts
type SlotSizeWindow = Window & {
  HTMLElement: typeof HTMLElement;
  ResizeObserver?: typeof ResizeObserver;
};
```

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
```diff
@@ -0,0 +1,146 @@
...
+    if (
+      !Number.isFinite(rectangle.width) ||
+      !Number.isFinite(rectangle.height) ||
+      rectangle.width < 0 ||
+      rectangle.height < 0
+    ) {
+      return;
+    }
+    this.store.recordObservedSlotSize(cycle.runtimeSlotNumber, cycle.requestNumber, [
+      rectangle.width,
```
prk-Jr: ♻️ **refactor** — round the measured box.

`getBoundingClientRect()` returns sub-pixel values and nothing rounds them, so the store retains e.g. `[728.328125, 90.5]` verbatim and the badge reads `Outer box 728.328125×90.5` next to a clean GPT `1×1`. Every sub-pixel delta also clears the equality guard in `recordObservedSlotSize`, so it notifies, re-renders badges and panel, and schedules another observer refresh. `Math.round` on both dimensions fixes the display and the churn:

```ts
this.store.recordObservedSlotSize(cycle.runtimeSlotNumber, cycle.requestNumber, [
  Math.round(rectangle.width),
  Math.round(rectangle.height),
]);
```

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
```diff
@@ -0,0 +1,146 @@
...
+  window?: SlotSizeWindow;
+  scheduleFrame?: (callback: () => void) => void;
+}
+
+interface ObservedCycle {
+  runtimeSlotNumber: number;
+  requestNumber: number;
+}
+
+function defaultScheduleFrame(callback: () => void): void {
```
prk-Jr: ⛏ **nitpick** — `defaultScheduleFrame` uses the global `requestAnimationFrame` rather than the injected window's, so an option-provided `window` is ignored for frame scheduling. Same wart already exists in `binding.ts` and `badges.ts`, so this is consistency, not a regression.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts
```diff
@@ -115,7 +119,13 @@ function badgeText(cycle: GptDiagnosticsRequestCycle): string {
   const delivery = deliveryLabel(cycle);
   if (delivery) firstLine.push(delivery);
   if (cycle.requestPath === 'competing') firstLine.push('Competing paths');
-  if (cycle.size) firstLine.push(`${cycle.size[0]}×${cycle.size[1]}`);
+  if (cycle.requestedSlotSizes) {
+    firstLine.push(`Requested ${formatSizes(cycle.requestedSlotSizes)}`);
```
prk-Jr: ♻️ **refactor** — the badge line can carry up to 16 sizes.

`MAX_REQUESTED_SLOT_SIZES` is 16, and `formatSizes` joins all of them into a badge capped at `BADGE_MAX_WIDTH_PX` (260px), so a multi-format slot wraps a long `Requested …` string over the creative. Consider truncating the badge to the first two or three plus `+N more`, and keeping the complete list in the panel, which already has room for it.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts
```diff
@@ -106,6 +106,10 @@ function deliveryLabel(cycle: GptDiagnosticsRequestCycle): string | undefined {
   }
 }
 
+function formatSizes(sizes: ReadonlyArray<readonly [number, number]>): string {
```
prk-Jr: ⛏ **nitpick** — `formatSizes` here duplicates the inline `.map(...).join(', ')` in `overlay.ts` (`cycleFacts`). One shared helper keeps the two surfaces from drifting in separator or `×` glyph.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
```diff
@@ -666,6 +698,45 @@ export class GptDiagnosticsStore {
...
+   * identify a filled render. Async DOM measurements use this guard so a prior
+   * render cannot alter a later refresh cycle.
+   */
+  recordObservedSlotSize(runtimeSlotNumber: number, requestNumber: number, size: Size): void {
+    if (
+      !Number.isSafeInteger(requestNumber) ||
+      requestNumber <= 0 ||
+      !Number.isFinite(size[0]) ||
+      !Number.isFinite(size[1]) ||
+      size[0] < 0 ||
```
prk-Jr, line 712: 🤔 **thinking** — the validator accepts `0` in both dimensions, so a breakpoint-hidden duplicate slot records `Outer box 0×0`. That may be the most useful signal available for those slots (it tells the operator the element is collapsed rather than mis-sized), so this is not a request to change it — just worth making the choice deliberate and saying so in the docs alongside the other rejection rules.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
```diff
@@ -666,6 +698,45 @@ export class GptDiagnosticsStore {
     );
   }
 
+  /**
+   * Retain an outer CSS box only when this exact slot and request cycle still
+   * identify a filled render. Async DOM measurements use this guard so a prior
+   * render cannot alter a later refresh cycle.
+   */
+  recordObservedSlotSize(runtimeSlotNumber: number, requestNumber: number, size: Size): void {
```
prk-Jr, line 706: 👍 **praise** — the stale-cycle split is the right shape: the observer only forwards the `(runtimeSlotNumber, requestNumber)` identity captured at schedule time, and the store owns rejection by re-checking latest-cycle, filled, and render-observed. Both sides are covered by tests, and the delayed-measurement case is asserted against a real refresh rather than mocked away.

prk-Jr (changes requested):
## Summary

Separating the configured request list, GPT's reported fill size, and a measured outer box is the right decomposition for diagnosing flexible/APS `1×1` fills, and the stale-cycle guard is designed well. Two things need a change before merge: on publisher-defined slots the new `requestedSlotSizes` reports a size list GPT never received, and the new observer's window type permits a call that throws inside a `ResizeObserver` callback.

## Blocking

### 🔧 wrench

- **`requestedSlotSizes` records sizes GPT never received**: `slot.formats` is passed unconditionally, but TS only supplies formats on the `defineSlot` path; the `existingSlot` (publisher-defined) path never applies sizes to the slot, and there is no `defineSizeMapping`/`setSizes` in the integration (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts:1061`).
- **`instanceof this.window.HTMLElement` throws for a type-valid window**: `SlotSizeWindow` does not declare `HTMLElement`; `tsc` reports it and an injected `{ ResizeObserver }` window throws uncaught inside the observer callback. Unreachable via `installGptDiagnosticsRuntime`, so reasonable to downgrade (`crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts:105`).

## Non-blocking

### ♻️ refactor

- **Round the measured box**: sub-pixel rects are stored and displayed verbatim, and each delta re-notifies (`slot_size_observer.ts:142`).
- **Badge line can carry 16 sizes** inside a 260px badge (`badges.ts:123`).
- **Extend the export type gate**: `test/integrations/gpt_diagnostics/types.test.ts` is the only `expectTypeOf` gate on the export allowlist and says to keep additions synchronized, but neither `requestedSlotSizes` nor `observedSlotSize` got an assertion there. `api.test.ts` covers the runtime key list only.

### 🤔 thinking

- **`0×0` observed boxes are recorded** for collapsed/hidden slots — likely desirable, worth documenting as deliberate (`store.ts:712`).

### ⛏ nitpick

- **Duplicated size formatting** between `badges.ts:109` and `overlay.ts:282`.
- **`defaultScheduleFrame` ignores the injected window** (`slot_size_observer.ts:31`), consistent with the existing pattern in `binding.ts`/`badges.ts`.

### 👍 praise

- Stale-cycle guard ownership split between observer and store, tested from both sides (`store.ts:706`).
- `normalizedRequestedSlotSizes` freezes and copies, with a test that mutates the caller's `formats` array after recording.
- Collapsing the arity-branching recorder call in `gpt/index.ts` into a single call is a clean simplification.

## CI Status

All 19 GitHub checks pass on `f5ca22e`. Re-run locally against the PR head:

- vitest (gpt + gpt_diagnostics): PASS (16 files, 359 tests)
- eslint (changed areas): PASS
- prettier: PASS
- `tsc --noEmit`: pre-existing baseline failure (170 errors on `main`), not a gate; the three new errors in `slot_size_observer.ts` are the 🔧 above.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
```diff
@@ -0,0 +1,146 @@
...
+  recordObservedSlotSize(runtimeSlotNumber: number, requestNumber: number, size: Size): void;
+  subscribe(listener: () => void): () => void;
+}
+
+interface SlotSizeBindings {
+  get: GptDiagnosticsBindingManager['get'];
+  subscribe(listener: () => void): () => void;
+}
+
+type SlotSizeWindow = Window & {
```
aram356, line 18: 🔧 **wrench** — `SlotSizeWindow` omits `HTMLElement`, so `element instanceof this.window.HTMLElement` (line 105) does not typecheck under strict mode, and the `Element → HTMLElement` narrowing at lines 106–107 fails with it. `tsc --noEmit` is not a CI gate, but these errors are new relative to `main`, and the sibling `binding.ts:10-14` already solves this exact problem by declaring the property on its window type.

**Fix**:
```ts
type SlotSizeWindow = Window & {
  HTMLElement: typeof HTMLElement;
  ResizeObserver?: typeof ResizeObserver;
};
```

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
```diff
@@ -666,6 +698,45 @@ export class GptDiagnosticsStore {
...
+      requestNumber <= 0 ||
+      !Number.isFinite(size[0]) ||
+      !Number.isFinite(size[1]) ||
+      size[0] < 0 ||
+      size[1] < 0
+    ) {
+      return;
+    }
+
+    const record = this.slots.get(runtimeSlotNumber);
```
aram356, line 718: 🔧 **wrench** — New strict-mode type error (`TS18048: 'record' is possibly 'undefined'` at line 722): `const cycle = record?.requests.find(...)` followed by `record.requests[record.requests.length - 1]` inside the same condition. TypeScript cannot correlate `cycle` being truthy with `record` being defined. Baseline `main` produces no `tsc` errors in this file; this PR introduces one.

**Fix** — narrow `record` first:
```ts
const record = this.slots.get(runtimeSlotNumber);
if (!record) return;
const cycle = record.requests.find((candidate) => candidate.requestNumber === requestNumber);
if (
  !cycle ||
  record.requests[record.requests.length - 1] !== cycle ||
  cycle.isEmpty !== false ||
  cycle.renderAtMs === undefined
) {
  return;
}
```

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
```diff
@@ -0,0 +1,146 @@
...
+    const rectangle = element.getBoundingClientRect();
+    if (
+      !Number.isFinite(rectangle.width) ||
+      !Number.isFinite(rectangle.height) ||
+      rectangle.width < 0 ||
+      rectangle.height < 0
+    ) {
+      return;
+    }
+    this.store.recordObservedSlotSize(cycle.runtimeSlotNumber, cycle.requestNumber, [
```
aram356, line 138: 🔧 **wrench** — `getBoundingClientRect()` returns fractional CSS pixels (page zoom, transforms, responsive layouts), and the value flows unrounded into the store, the badge (`Outer box 383.99667358398438×90`), the overlay, and the JSON export. The tests only exercise integer mocks, so this never surfaces. Rounding here also stabilizes the store's change-dedupe guard against sub-pixel jitter, which otherwise re-notifies and re-runs the whole refresh pipeline on every fraction-of-a-pixel resize.

**Fix**:
```ts
this.store.recordObservedSlotSize(cycle.runtimeSlotNumber, cycle.requestNumber, [
  Math.round(rectangle.width),
  Math.round(rectangle.height),
]);
```

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
```diff
@@ -0,0 +1,146 @@
...
+      this.resizeObserver?.observe(binding.element);
+      this.scheduleMeasure(binding.element, cycle);
+    }
+  }
+
+  private scheduleMeasure(element: HTMLElement, cycle: ObservedCycle): void {
+    this.scheduleFrame(() => this.measure(element, cycle));
+  }
+
+  private measure(element: HTMLElement, cycle: ObservedCycle): void {
```
aram356, line 121: ♻️ **refactor** — `measure()` has no `destroyed` guard, so pending `scheduleFrame` measure callbacks still run after `destroy()`. Today this is harmless only because `installGptDiagnosticsRuntime` destroys `bindings` in the same synchronous sequence, making `bindings.get()` return the unbound default — that is coupling to destroy order in `index.ts`. `refresh()` and `scheduleRefresh()` both carry the guard; `measure()` is the odd one out.

**Fix**: add `if (this.destroyed) return;` as the first statement.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts
```diff
@@ -115,7 +119,13 @@ function badgeText(cycle: GptDiagnosticsRequestCycle): string {
   const delivery = deliveryLabel(cycle);
   if (delivery) firstLine.push(delivery);
   if (cycle.requestPath === 'competing') firstLine.push('Competing paths');
-  if (cycle.size) firstLine.push(`${cycle.size[0]}×${cycle.size[1]}`);
+  if (cycle.requestedSlotSizes) {
```
aram356, line 112: ♻️ **refactor** — `MAX_REQUESTED_SLOT_SIZES` is 16 and the badge renders them all on the first line inside a 260px-max-width badge, so a many-format responsive slot produces a badge that wraps into a wall of text occluding the page. Suggest capping the badge display and leaving the full list to the overlay panel and export:

```ts
const shown = cycle.requestedSlotSizes.slice(0, 3);
const extra = cycle.requestedSlotSizes.length - shown.length;
firstLine.push(`Requested ${formatSizes(shown)}${extra > 0 ? ` +${extra}` : ''}`);
```

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
```diff
@@ -0,0 +1,146 @@
...
+    this.refreshScheduled = true;
+    this.scheduleFrame(() => {
+      this.refreshScheduled = false;
+      this.refresh();
+    });
+  };
+
+  private refresh(): void {
+    if (this.destroyed) return;
+    this.resizeObserver?.disconnect();
```
aram356, line 93: 🤔 **thinking** — Every store/bindings notification tears down and re-creates the `ResizeObserver` and re-observes every bound element — and per spec each `observe()` fires an initial callback, so every notification schedules a redundant measure per slot on top of the direct `scheduleMeasure` below. The no-change guard in `recordObservedSlotSize` keeps this convergent, so it is churn rather than a bug, but a single long-lived observer with set-diffing would do less layout work per notification. Fine to defer.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
```diff
@@ -0,0 +1,146 @@
...
+    this.scheduleFrame(() => this.measure(element, cycle));
+  }
+
+  private measure(element: HTMLElement, cycle: ObservedCycle): void {
+    const binding = this.bindings.get(cycle.runtimeSlotNumber);
+    if (binding.binding.status !== 'bound' || binding.element !== element || !element.isConnected) {
+      return;
+    }
+
+    const rectangle = element.getBoundingClientRect();
```
aram356, line 129: 🤔 **thinking** — A filled render whose container is `display:none` or collapsed measures 0×0 here, and both validation layers deliberately allow 0 (`< 0` checks only), so the badge shows `Outer box 0×0`. If "0×0 = collapsed slot evidence" is intended, a sentence in the docs would keep operators from reading it as a measurement bug; if not, drop zero boxes.

docs/guide/integrations/gpt-diagnostics.md
```diff
@@ -362,6 +388,7 @@ inaccessible to JavaScript.
 - Retained request cycles per slot: 10.
 - Retained callback issues: 128.
 - Retained auction-slot-to-GPT-slot associations: 64.
+- Requested slot sizes per correlated request: 16 valid positive sizes.
```
aram356: ⛏ **nitpick** — "16 valid positive sizes" overstates the cap: `normalizedRequestedSlotSizes` slices to 16 entries *before* validating, so invalid entries inside the first 16 shrink the retained count (the store test itself asserts 14). Either reword to "first 16 configured entries, invalid entries dropped" or change the implementation to filter-then-cap.

Also, lines 300–301 repeat themselves: "prevent diagnostics from inspecting iframe content. It does not inspect iframe content…" — worth merging into one sentence.

aram356 (changes requested):
## Summary

Adds three separately-labelled size facts to GPT diagnostics — configured `requestedSlotSizes`, the GPT-reported fill `size`, and a new `observedSlotSize` measured via `GptDiagnosticsSlotSizeObserver`. The design is solid: validation is thorough, the stale-cycle rejection lives at the store level (defense-in-depth), and the new paths are well tested. Requesting changes for two 🔧 items: new strict-mode type errors introduced in the new code, and unrounded sub-pixel measurements flowing into the badge/overlay/export.

## Blocking

### 🔧 wrench

- **New strict-mode type errors in new code**: `SlotSizeWindow` omits `HTMLElement` (slot_size_observer.ts:17, errors at 105–107) and `record` is possibly undefined in `recordObservedSlotSize` (store.ts:722). Both are new relative to `main`; `binding.ts` already models the window-type fix.
- **Fractional observed sizes**: `getBoundingClientRect()` values are recorded and displayed unrounded (slot_size_observer.ts:141) — round at the measurement site.

## Non-blocking

### ♻️ refactor

- **`measure()` lacks a `destroyed` guard** (slot_size_observer.ts:126) — currently safe only via destroy order in `index.ts`.
- **Badge first line can explode with up to 16 requested sizes** (badges.ts:122) — cap the badge display, keep the full list in the panel/export.

### 🤔 thinking

- **`ResizeObserver` torn down and rebuilt on every notification** (slot_size_observer.ts:98) — convergent but churny; fine to defer.
- **Hidden containers record `Outer box 0×0`** (slot_size_observer.ts:132) — document if intended, otherwise drop zero boxes.

### 🏕 camp site

- **Third copy of `defaultScheduleFrame`**: `binding.ts`, `badges.ts`, and now `slot_size_observer.ts` each carry an identical helper, and size-list formatting is now duplicated between `badges.ts` (`formatSizes`) and `overlay.ts` (inline map/join). A tiny shared helper module would leave the area cleaner.

### 🌱 seedling

- **Test gaps**: no test covers the `ResizeObserver`-unavailable path (which the docs explicitly describe: "otherwise it is the most recently sampled box"), and none covers fractional rect values — which is how the rounding issue slipped through.

### ⛏ nitpick

- **Docs overstate the requested-sizes cap** (gpt-diagnostics.md:391) — implementation slices to 16 entries before validating; also a repeated sentence at lines 300–301.

## CI Status

- fmt: PASS
- clippy: PASS
- rust tests: PASS (fastly, axum, cloudflare, spin, parity, CLI)
- js tests: PASS (vitest; focused diagnostics + ad_init suites also re-run locally: 283 passed)
- format (typescript/docs): PASS

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Fixed in 49e7094a. Diagnostics now read formats from the Trusted Server handoff keyed by the GPT slot element ID, so publisher-defined slots omit requested sizes while reused handoff slots retain them. Added coverage for both cases.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
ChristianPavilonis: Fixed in 49e7094a. `SlotSizeWindow` now requires `HTMLElement`, matching the binding manager, and the injected-window observer test covers the callback path.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
ChristianPavilonis: Fixed in 49e7094a. Measurements round both dimensions before entering the store. The observer test covers fractional initial and resize measurements.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
ChristianPavilonis: Fixed in 49e7094a. Shared frame scheduling now calls the injected window’s `requestAnimationFrame`, with a microtask fallback. The observer test verifies the injected scheduler is used.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts
ChristianPavilonis: Fixed in 49e7094a. Badges show at most three requested sizes plus `+N`; the overlay and export retain the complete list. Added badge and four-size overlay coverage.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts
ChristianPavilonis: Fixed in 49e7094a. Badge and overlay now share `formatSizes`, so separator and glyph formatting cannot drift.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
ChristianPavilonis: Documented in 49e7094a. A bound collapsed or hidden element may report `0×0` as page-layout evidence.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
ChristianPavilonis: Thanks. The stale-cycle guard and its delayed-measurement test remain unchanged.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
ChristianPavilonis: Fixed in 49e7094a. `SlotSizeWindow` now declares `HTMLElement`, removing the new strict-mode narrowing errors.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
ChristianPavilonis: Fixed in 49e7094a. `recordObservedSlotSize` now returns after narrowing a missing record before reading its request list.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
ChristianPavilonis: Fixed in 49e7094a. The observer rounds both dimensions at the measurement site, with fractional-value coverage.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
ChristianPavilonis: Fixed in 49e7094a. `measure()` now exits after `destroy()`, and a pending-frame test covers teardown.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts
ChristianPavilonis: Fixed in 49e7094a. The badge retains the first three sizes and a `+N` count; the full list remains in the overlay and export.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
ChristianPavilonis: Deferred as suggested. The current observer rebuild is convergent and not a correctness issue; this patch avoids a higher-risk lifecycle refactor.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
ChristianPavilonis: Documented in 49e7094a. `0×0` is retained as intentional collapsed or hidden layout evidence.

docs/guide/integrations/gpt-diagnostics.md
ChristianPavilonis: Fixed in 49e7094a. The retention text now says the first 16 configured entries are considered and invalid entries are dropped. The iframe-content wording is also consolidated.

crates/trusted-server-js/lib/test/integrations/gpt_diagnostics/api.test.ts
```diff
@@ -214,6 +217,10 @@ describe('GptDiagnosticsApiController', () => {
               requestNumber: 1,
               durations: {},
               incompleteSequence: false,
+              requestedSlotSizes: [
+                [300, 250],
+                [728, 90],
+              ],
```
prk-Jr: ♻️ **refactor** — This fixture literal infers as `number[][]`, not `ReadonlyArray<Size>` (`Size = readonly [number, number]`), so the mock store no longer satisfies `ApiStore` and `tsc` reports a new `TS2345` on the `new GptDiagnosticsApiController(store, …)` call at line 252.

CI does not catch it: `vitest.config.ts` scopes `typecheck.include` to `test/**/types.test.ts` with `ignoreSourceErrors: true`, and the `vitest` job runs no package-wide `tsc --noEmit`. Measured in `crates/trusted-server-js/lib`: 171 errors on the merge-base, 172 on this head. `as const` gives `readonly [readonly [300, 250], readonly [728, 90]]`, which satisfies `ReadonlyArray<Size>` and returns the count to 171.

```suggestion
              ] as const,
```

_(verified with this applied: prettier clean, `npx vitest run` 859/859 pass)_

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -1070,24 +1070,18 @@ export function installTsAdInit(): void {
           if (bid[key]) gptSlot.setTargeting(key, String(bid[key]!));
         });
         gptSlot.setTargeting(TS_INITIAL_TARGETING_KEY, '1');
+        const requestedSlotSizes = ts.gptSlotHandoffs?.[slotDivId2]?.formats;
         // Diagnostics are observational only. A missing or malformed debug
         // implementation must never interrupt slot mapping or delivery.
         try {
           const opportunity = trustedServerOpportunity(bid);
```
prk-Jr: ♻️ **refactor** — This read lands one statement above the `try` whose own comment says diagnostics "must never interrupt slot mapping or delivery". `ts.gptSlotHandoffs` hangs off the publisher-reachable `window.tsjs`, so a throwing getter or a Proxy trap there escapes into this `slots.forEach` callback and aborts slot mapping for this slot and every slot after it — precisely the failure the `try` exists to prevent.

Moving the read inside costs nothing, since `requestedSlotSizes` is only ever consumed by the recorder call below it.

```suggestion
        // Diagnostics are observational only. A missing or malformed debug
        // implementation must never interrupt slot mapping or delivery.
        try {
          const requestedSlotSizes = ts.gptSlotHandoffs?.[slotDivId2]?.formats;
          const opportunity = trustedServerOpportunity(bid);
```

_(verified with this applied: eslint and prettier clean, `npx vitest run` 859/859 pass)_

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts
```diff
@@ -115,7 +109,19 @@ function badgeText(cycle: GptDiagnosticsRequestCycle): string {
...
+    const remainingSizeCount = cycle.requestedSlotSizes.length - displayedSizes.length;
+    firstLine.push(
+      `Requested ${formatSizes(displayedSizes)}${
+        remainingSizeCount > 0 ? ` +${remainingSizeCount}` : ''
+      }`
+    );
+  }
+  if (cycle.size) firstLine.push(`GPT fill ${cycle.size[0]}×${cycle.size[1]}`);
+  if (cycle.observedSlotSize) {
+    firstLine.push(`Outer box ${cycle.observedSlotSize[0]}×${cycle.observedSlotSize[1]}`);
```
prk-Jr: 🤔 **thinking** — The badge's first line can now read `Filled · Requested 728×90, 970×250 · GPT fill 728×90 · Outer box 980×270`, about 72 characters, against `max-width: 260px` at `font: 11px/1.35` with `white-space: pre-line` (`.tsgd-badge` in `overlay.ts`). That wraps to roughly three rendered lines for the first line alone. Since `update()` positions badges with `translateY(-100%)` above the slot, a badge annotating a 90px leaderboard now covers more publisher content above the slot than the slot itself occupies.

The panel already carries all three facts in full, so the badge does not have to. Two ways to keep it compact if you agree it is worth it: keep `Requested` panel-only, or shorten the badge labels to `Req` / `Fill` / `Box`. Flagging the layout consequence rather than asking for a specific change — your call.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
```diff
@@ -0,0 +1,143 @@
...
+  requestNumber: number;
+}
+
+function latestFilledCycle(
+  slot: GptDiagnosticsStoreSnapshot['slots'][number]
+): ObservedCycle | undefined {
+  const cycle = slot.requests[slot.requests.length - 1];
+  if (!cycle || cycle.isEmpty !== false || cycle.renderAtMs === undefined) return undefined;
+  return { runtimeSlotNumber: slot.runtimeSlotNumber, requestNumber: cycle.requestNumber };
+}
```
prk-Jr, line 39: 🌱 **seedling** — `latestFilledCycle` requires `cycle.isEmpty === false`, so an outer box is never measured on an empty render. That is the self-consistent choice for this PR and the docs state it plainly, so no change requested here.

Worth noting for later, though: a slot that renders empty and collapses to `0×0` is exactly the case where an observed outer box is most diagnostic, and it is the one case this field cannot currently report on. A follow-up that also measures empty cycles, kept clearly distinct from a fill measurement, would close that gap.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
```diff
@@ -666,6 +698,47 @@ export class GptDiagnosticsStore {
     );
   }
 
+  /**
+   * Retain an outer CSS box only when this exact slot and request cycle still
+   * identify a filled render. Async DOM measurements use this guard so a prior
+   * render cannot alter a later refresh cycle.
+   */
+  recordObservedSlotSize(runtimeSlotNumber: number, requestNumber: number, size: Size): void {
```
prk-Jr, line 706: 📝 **note** — The guard correctly refuses to write a measurement onto a cycle that is no longer latest, but nothing clears `observedSlotSize` once it has been written. If the bound element later detaches or the slot becomes ambiguous, the panel keeps rendering `Observed outer slot box 970×250` next to a binding status of `unbound`.

That reads fine if the field means "the outer box at the time it was last measurable", which the docs support. Confirming that is the intent rather than an omission, since the panel shows both facts side by side and a reader could take the stale box as current.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    for (const slot of this.store.snapshot().slots) {
+      const cycle = latestFilledCycle(slot);
+      const binding = this.bindings.get(slot.runtimeSlotNumber);
+      if (!cycle || binding.binding.status !== 'bound' || !binding.element?.isConnected) continue;
+      observations.set(binding.element, cycle);
+      this.resizeObserver?.observe(binding.element);
+      this.scheduleMeasure(binding.element, cycle);
+    }
+  }
```
prk-Jr, line 115: 🤔 **thinking** — Three small costs stack up here, each paid per `refresh()`, and `refresh()` runs on both store notifications and binding notifications. The latter fires whenever a slot's viewport visibility flips, so scrolling drives it:

1. A fresh `ResizeObserver` is constructed each time, rather than reusing one and adjusting its observed set.
2. `scheduleMeasure` is called directly for each slot in addition to `observe()`, which already delivers an initial observation per spec — so each bound slot performs two `getBoundingClientRect()` layout reads per refresh.
3. `this.store.snapshot()` deep-copies every retained cycle through `copyCycle` (up to `MAX_DIAGNOSTIC_SLOTS` × `MAX_REQUEST_CYCLES_PER_SLOT` = 640), while this observer needs only each slot's `runtimeSlotNumber` plus the last cycle's `requestNumber`, `isEmpty`, and `renderAtMs`. `store.bindingInputs()` is the existing precedent for a cheap projection.

The store's `notify()` coalescing through `queueMicrotask` keeps all of this bounded, so it is not urgent — but this observer is now the third consumer calling `snapshot()` on every notification, alongside the badge manager and the overlay.

Separately, on the `if (typeof ResizeObserverConstructor === 'function')` branch: when the constructor is absent, `this.resizeObserver` keeps its previous value instead of being cleared, so a disconnected observer from an earlier refresh — whose callback closes over that refresh's `observations` map — would be re-`observe()`d below and could report against stale cycles. Unreachable today, since `ResizeObserver` availability does not change mid-page, but an `else { this.resizeObserver = undefined; }` removes the trap.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts
```diff
@@ -277,7 +270,13 @@ function cycleFacts(cycle: GptDiagnosticsRequestCycle): string[] {
   if (cycle.loadAtMs !== undefined) facts.push('GPT slot onload observed');
   if (cycle.viewableAtMs !== undefined) facts.push('GPT impressionViewable observed');
   if (cycle.incompleteSequence) facts.push('Incomplete sequence');
-  if (cycle.size) facts.push(`Rendered size ${cycle.size[0]}×${cycle.size[1]}`);
+  if (cycle.requestedSlotSizes) {
+    facts.push(`Requested slot sizes ${formatSizes(cycle.requestedSlotSizes)}`);
+  }
+  if (cycle.size) facts.push(`GPT-reported fill size ${cycle.size[0]}×${cycle.size[1]}`);
```
prk-Jr, line 276: ⛏ **nitpick** — `formatSizes` is introduced in this PR and used for the multi-size lists, but the four single-size sites still inline the template: this line, line 278, and `badges.ts:121` and `:123`. `formatSizes([cycle.size])` is arguably harder to read than the inline form for a single size, so this may well be the right call — noting it only so the helper's scope reads as a deliberate boundary rather than drift.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
```diff
@@ -205,6 +207,29 @@ function normalizedAuctionId(value: unknown): string | undefined {
   return new TextEncoder().encode(trimmed).length <= 256 ? trimmed : undefined;
 }
 
+function normalizedRequestedSlotSizes(value: unknown): ReadonlyArray<Size> | undefined {
```
prk-Jr, line 210: 👍 **praise** — Copying and freezing at the boundary here, rather than trusting the caller, is the right shape for a field arriving from publisher-reachable state. What makes it a real guarantee instead of an intention is the store test that mutates the caller's `formats` array *after* `recordTrustedServerOpportunity` returns and then asserts the retained list is unchanged. Same for the bounds-and-validation test asserting the specific rejected entries rather than just a length.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/api.ts
```diff
@@ -149,18 +152,21 @@ export class GptDiagnosticsApiController {
...
-              trustedServerAuctionId
-            );
-          }
+          this.store.recordTrustedServerOpportunity(
+            slot,
+            auctionSlotId,
+            opportunity,
+            trustedServerAuctionId,
+            requestedSlotSizes
+          );
```
prk-Jr, line 169: 👍 **praise** — Collapsing the `trustedServerAuctionId === undefined` branch here, and the matching one in `gpt/index.ts`, into a single forwarding call is a clean simplification, and it is behaviour-preserving: `normalizedAuctionId(undefined)` already returns `undefined`, and the store spreads the field in conditionally. Good to see the arity-branching pattern removed on both sides rather than a third branch grafted on for the new argument.

prk-Jr (approved):
## Summary

Adds three separately labelled slot-size facts to GPT diagnostics: the configured sizes Trusted Server supplied, GPT's reported `slotRenderEnded.size` fill, and a measured outer CSS box. The separation is held consistently through the store, the export clones, the panel, and the badge, with copy-and-freeze at the input boundary and a two-layer stale-cycle guard. Approving: nothing here blocks merge. Two findings carry one-click suggestions worth taking; the rest are optional.

> 2 of the inline comments below carry a one-click GitHub `suggestion` — use **Commit suggestion** (or **Add suggestion to batch**) to apply them as commits on this branch. The remaining comments are design calls or span several files, so they describe the change in prose instead.

## Non-blocking

### ♻️ refactor

- **New fixture literal introduces a net-new `tsc` error** — see inline at `crates/trusted-server-js/lib/test/integrations/gpt_diagnostics/api.test.ts:223`
- **Requested-sizes read sits outside the guard meant to contain it** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt/index.ts:1073-1077`

### 🤔 thinking

- **Badge first line roughly triples in length** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/badges.ts:112-123`
- **Observer rebuilds a `ResizeObserver` and a full snapshot on every notification** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts:91-115`

### 🌱 seedling

- **No outer box is measured for empty renders** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/slot_size_observer.ts:33-39`

### 📝 note

- **`observedSlotSize` is never cleared once the element unbinds** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts:706`

### ⛏ nitpick

- **New `formatSizes` helper bypassed at the single-size call sites** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts:276`

### 👍 praise

- **Copy-and-freeze at the input boundary, proven by a mutating test** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts:210`
- **Optional-argument forwarding removes two duplicated call sites** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/api.ts:155-169`

- 📌 **The docs call the export "allowlisted"; the code spreads.** `cloneExportSnapshot` (`crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/api.ts:64-80`) and `api.snapshot()` (`crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/api.ts:197-212`) both `...cycle`, so every field the store retains reaches the export by default; the allowlist is only asserted after the fact by the sorted-key test in `api.test.ts`. This is pre-existing, but the PR adds two fields to that surface and extends the docs claim in `docs/guide/integrations/gpt-diagnostics.md`. Worth a follow-up issue that makes the export an explicit key pick rather than a spread, so a future store field cannot reach the export without someone choosing it. Not a change for this PR.

## Verification notes

Both suggestions were applied in an isolated worktree at `4853ae20`, first individually and then together. With both applied: `npx eslint . --max-warnings=0` clean, `npx prettier --check` clean, `node build-all.mjs` succeeds, and `npx vitest run` passes 859/859 across 45 files. Baseline `tsc --noEmit` in `crates/trusted-server-js/lib/` is 171 errors on the merge-base and 172 on this head; the `api.test.ts` suggestion returns it to 171.

## CI Status

- integration tests: PASS
- CodeQL: PASS
- Analyze (actions): PASS
- Analyze (rust): PASS
- cargo test: PASS (required)
- cargo fmt: PASS (required)
- vitest: PASS

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -1070,24 +1070,18 @@ export function installTsAdInit(): void {
           if (bid[key]) gptSlot.setTargeting(key, String(bid[key]!));
         });
         gptSlot.setTargeting(TS_INITIAL_TARGETING_KEY, '1');
+        const requestedSlotSizes = ts.gptSlotHandoffs?.[slotDivId2]?.formats;
```
aram356: ♻️ **refactor** — The freshly-defined (`tsOwned`) slot path — the primary first-page-load path — has no test asserting that `requestedSlotSizes` flows through the handoff written at line 1050. Every `recordTrustedServerOpportunity` assertion in `ad_init.test.ts` exercises the existing-slot path (`getSlots` returns the mock slot), with the handoff either absent or manually injected. A refactor that moves this read above the handoff write would silently drop the feature on the main path while the whole suite stays green.

**Fix**: one test with `getSlots: vi.fn().mockReturnValue([])` plus a recorder, asserting the define path records the configured formats.

docs/guide/integrations/gpt-diagnostics.md
```diff
@@ -337,6 +337,32 @@ because selector support is unavailable or throws, the export reports
 `dom_uniqueness_unverifiable`. Framework replacement of an element with a new unique
 element using the same exact ID is rebound automatically.
 
+When Trusted Server associates a GPT slot with its next request, diagnostics retains
+`requestedSlotSizes`: the configured `AuctionSlot.formats` list Trusted Server supplied
```
aram356, line 341: ⛏ **nitpick** — With the handoff-based sourcing, a Trusted Server slot reused across SPA navigations reports its *defineSlot-time* formats, which can lag the current auction config (GPT genuinely still holds the original sizes, so the values are right). A clarifying phrase such as "as supplied when the slot was defined" would keep an operator from puzzling over a mismatch with their current configured formats.

aram356 (approved):

PR #1033, opened 2026-08-14
Render APS creatives in opaque data frames

### LINKED ISSUE #999: APS creatives requiring nested same-origin semantics render blank
## Description

Some APS creatives do not render even though the APS renderer and Amazon runner initialize. The failure occurs when a creative library expects a nested HTTPS iframe to retain its real origin.

A confirmed example is a Flashtalking Richload creative: its nested Richload iframe remains `visibility: hidden` with an empty document, so the placement appears blank.

This is separate from the PUC transport, GAM `1x1` shell resizing, and clipping fixes in #998.

## Root cause

The APS renderer is served from the publisher origin at:

```text
/integrations/aps/renderer
```

To keep third-party code isolated from the publisher, the outer renderer iframe and its response CSP deliberately omit `allow-same-origin` from their sandboxes.

That gives the renderer an opaque origin, but it also activates the sandboxed-origin restriction for the renderer's entire descendant browsing-context tree. A nested iframe cannot loosen an ancestor's sandbox restrictions. Consequently, adding `allow-same-origin` to the creative iframe created by Amazon's runner is ineffective: the inherited outer restriction still forces it to origin `null`.

The affected Flashtalking creative then reports errors such as:

```text
SecurityError: Blocked a frame with origin "null" from accessing a cross-origin frame.
Both frames are sandboxed and lack the "allow-same-origin" flag.

TypeError: Cannot read properties of null (reading 'open') at #setupRichload
```

When the effective sandbox permits same-origin semantics, the Richload iframe receives its expected HTTPS origin, populates its document, and becomes visible.

## Security constraint

We should not simply add `allow-same-origin` to the current publisher-origin renderer.

Combining `allow-scripts` and `allow-same-origin` on a publisher-origin document would allow code in that document to access publisher DOM and origin state and potentially remove its iframe sandbox. Any solution must preserve the boundary between APS/creative code and the publisher.

Existing protections should also remain intact:

- Bounded and canonical descriptor validation
- Nonce-bound, one-shot message transport
- Exact source-frame and MessagePort ownership
- Publisher-origin creative URL rejection
- Direct-parent compatibility
- GAM/PUC slot ownership and guarded shell resizing

### 1. Naturally opaque `data:` renderer

Instead of navigating the renderer iframe to a publisher-origin endpoint, embed the static renderer document in a `data:` URL and apply a sandbox containing `allow-scripts allow-same-origin`.

A `data:` document has a naturally unique opaque origin. `allow-same-origin` therefore does not make it publisher-origin, but it avoids imposing the sandboxed-origin restriction on its HTTPS descendants.

A controlled Chromium proof with the affected Flashtalking creative produced:

- Renderer origin: `null`
- Publisher DOM readable from renderer: no
- Renderer `frameElement` readable/removable: no
- Flashtalking frame origin: `https://cdn.flashtalking.com`
- Richload used/loaded: yes
- Richload visibility: `visible`
- Richload document populated: yes

Advantages:

- No new renderer hostname or service
- Preserves browser-enforced separation from the publisher
- Allows nested creative frames to keep their real origins
- Does not require rewriting vendor creative markup into the publisher document
- Existing nonce and MessageChannel protocols can remain

Trade-offs and open questions:

- A `data:` document cannot receive HTTP response headers. Resource restrictions can use a meta CSP, but CSP directives such as `sandbox` cannot be equivalently delivered through a meta policy. The iframe `sandbox` attribute becomes the primary boundary.
- Publisher and GAM/PUC frame policies must permit `data:` in `frame-src`/`child-src`.
- `location.origin` is `null`, so defense-in-depth validation that currently rejects creative URLs matching the publisher must receive or embed the trusted publisher origin explicitly.
- The static renderer source moves into the client/PUC program and makes the iframe URL larger and less convenient to inspect.
- Self-navigation, popup, top-navigation, and sandbox-removal attempts need explicit adversarial tests.
- Chromium, Firefox, WebKit, direct rendering, and real PUC topology all need validation.

### 2. Dedicated cross-origin renderer endpoint

Serve the renderer from a distinct browser origin and permit `allow-same-origin` there.

Advantages:

- Conventional HTTP response headers and CSP
- Straightforward origin-targeted messaging
- Clear isolation boundary when the origin has no other authority

Trade-offs:

- Requires DNS, TLS, routing, deployment, and publisher CSP changes
- The origin must be exclusively renderer-only; exposing normal Trusted Server routes under it would weaken the boundary
- Additional operational ownership for a single integration surface

This may not be operationally acceptable for deployments that do not want a dedicated renderer domain.

### 3. Render remote iframe creatives directly as siblings

Keep the opaque APS controller but ask trusted publisher-side code to create a direct cross-origin creative iframe outside the opaque controller's descendant tree.

Advantages:

- Retains a normal HTTPS origin for iframe-tag creatives
- No additional renderer hostname

Trade-offs:

- Does not generically support script-tag creatives, where Amazon's runner fetches creative HTML and renders it through `srcdoc`
- Requires reimplementing part of Amazon's rendering behavior
- Risks fragile format- or vendor-specific handling
- Increases the amount of creative orchestration performed by first-party code

### 4. Vendor-specific fallback

Detect affected creative formats/vendors and render them through a special path.

Advantages:

- Potentially narrow implementation

Trade-offs:

- Brittle and difficult to generalize
- Requires reliable vendor/format identification before execution
- Likely to regress as vendor markup changes
- Creates inconsistent security and rendering behavior

### 5. Preserve the current opaque renderer

Continue rejecting incompatible same-origin-dependent creatives.

Advantages:

- No change to the existing security model

Trade-offs:

- Affected paid placements remain blank
- Renderer success signals may not reflect whether deeply nested creative content actually painted

## Suggested investigation

The naturally opaque `data:` renderer appears to be the most promising no-domain direction, but it should not be selected without completing the security and compatibility work.

1. Build a test-only data-URL renderer using the existing static renderer program.
2. Keep descriptors out of the URL and transfer them only after nonce/source authentication.
3. Reproduce the current resource CSP with a meta policy and document the directives that cannot be preserved.
4. Add explicit publisher-origin validation rather than relying on the data document's `location.origin`.
5. Run adversarial navigation, popup, storage, DOM-access, sandbox-removal, replay, and wrong-source tests.
6. Exercise direct APS and real Prebid Universal Creative topologies in Chromium, Firefox, and WebKit.
7. Add the confirmed Flashtalking Richload creative as a controlled browser regression.
8. Preserve the current HTTPS renderer as a compatibility fallback until mixed deployments and rollback behavior are defined.

## Acceptance criteria

- Affected nested creatives render and paint at their expected size.
- Renderer and creative code cannot read or mutate publisher DOM, cookies, storage, or globals.
- Descendant creative frames retain their intended remote origins.
- Descriptor, nonce, source, port, timeout, and ownership checks remain enforced.
- Wrong-origin and publisher-origin creative URLs fail closed.
- Direct and real PUC paths behave consistently.
- The solution has cross-browser regression coverage and a documented rollback path.

ChristianPavilonis:
## Summary

- Render APS creatives through nonce-bound, nested opaque `data:` documents so HTTPS creative origins are retained without granting publisher access.
- Add exact-origin renderer containment, one-shot PUC top-page mounts, and APS-wide publisher `frame-ancestors 'self'` protection against recursive publisher framing.
- Cover direct and real PUC rendering in Chromium, Firefox, and WebKit; document the CSP and redirect compatibility trade-offs.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/integrations/aps.rs` | Serve the data-bootstrap renderer mode and static renderer sources. |
| `crates/trusted-server-core/src/response_privacy.rs` | Append an independent APS publisher `frame-ancestors 'self'` CSP policy. |
| `crates/trusted-server-js/lib/src/integrations/aps/` | Add the bootstrap/container/renderer documents and nonce- and port-bound data-frame rendering flow. |
| `crates/trusted-server-js/lib/src/integrations/gpt/index.ts` | Register and consume one-shot PUC top-page mount capabilities. |
| `crates/trusted-server-js/lib/src/core/request.ts` | Cancel pending APS work before generic slot replacement. |
| `crates/trusted-server-js/lib/test/` | Add renderer, capability refresh/replay, and replacement-cancellation coverage. |
| `crates/trusted-server-integration-tests/browser/` | Add real PUC fixture and cross-browser APS isolation coverage. |
| `docs/guide/integrations/aps.md` | Document containment architecture, CSP requirements, and compatibility trade-offs. |

## Closes

Closes #999

## Test plan

- [x] Other: `cargo test-cloudflare`, `cargo test-spin`, adapter parity tests, Cloudflare/Spin Clippy targets, `git diff --check`, Chromium browser integration tests (Next.js and WordPress), and Firefox/WebKit browser integration tests (Next.js).

## Checklist

- [x] No new logging added; existing code uses project-standard `log` macros rather than `println!`
- [x] New code has tests


crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -363,45 +398,93 @@ export function renderApsCreative({ slotId, renderer: input }: RenderApsCreative
...
+
     for (const child of Array.from(container.children)) {
-      if (child !== iframe) child.remove();
+      if (child === iframe) continue;
+      if (options.mode === 'replace' || (child as HTMLElement).dataset.tsApsRenderer === 'true') {
+        child.remove();
+      } else if (child instanceof HTMLElement) {
+        // Keep Universal Creative connected long enough to receive the success
+        // acknowledgement, but never leave two visible rendering surfaces.
+        child.style.display = 'none';
```
prk-Jr: 🔧 **wrench** — Universal-creative mount permanently forks the slot.

`commit()` in `mode: 'universal-creative'` hides the GAM/PUC iframe and never restores or removes it. The mounted APS iframe is removed only when a *newer APS* mount commits (the `dataset.tsApsRenderer` branch above). Nothing hooks slot lifecycle — the `slotRenderEnded` listener in `gpt/index.ts` only runs the debug `adm` bypass.

**Failure**: the slot renders APS through PUC, then the publisher calls `refresh()` and GAM returns a non-TS creative for the same div.

- If GPT reuses the iframe, the new creative paints inside a `display: none` iframe → blank slot on a billed impression.
- If GPT creates a fresh iframe, the stale APS creative and the new GAM creative both sit in the div.

Before this PR the renderer lived *inside* the PUC frame, so a refresh tore it down together with the GAM document. Moving the mount to the top page drops that cleanup.

Secondary: ActiveView now measures a hidden GAM iframe while the creative paints in a sibling, so viewability on these impressions reports ~0.

**Fix**: track the mounted frame per div and, on `slotRenderEnded` for that div, remove it and unhide the GAM iframe unless the new render is the TS APS render.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -363,45 +398,93 @@ export function renderApsCreative({ slotId, renderer: input }: RenderApsCreative
...
     iframe.style.display = '';
+    options.onReady?.();
   };
+  const postLegacyDescriptor = (): void => {
+    const target = iframe.contentWindow;
+    if (!target) {
+      fail();
+      return;
+    }
+    target.postMessage({ nonce, publisherOrigin, renderer }, '*');
```
prk-Jr: 🔧 **wrench** — The legacy fallback message shape can never be accepted.

After `LEGACY_RENDERER_FALLBACK_MS` without `bootstrap-ready`, this posts `{ nonce, publisherOrigin, renderer }`. The only document reachable at `?mode=data-bootstrap` that lacks the new handler is the pre-PR `APS_RENDERER_DOCUMENT` — the old Rust handler ignored the query string entirely and served the renderer for any query. That document requires exactly:

```js
keys(message, ['nonce', 'renderer'])
```

so the three-key message is dropped, `renderer-ready` never arrives, and the 10 s timeout removes the frame with no ad. The rollback path documented in `docs/guide/integrations/aps.md` does not work.

No test exercises `phase === 'legacy'` — there is no coverage in the unit specs or the browser spec.

**Fix**:

```ts
target.postMessage({ nonce, renderer }, '*');
```

The new `renderer.html` accepts both the legacy and modern shapes, so the two-key message stays compatible in both directions. Please add a test for the fallback phase.

If the intended consumer is instead a cache that collapses the query string and serves the *new* document, that is worth stating explicitly — but the pre-PR rollback case still needs handling either way.

crates/trusted-server-core/src/response_privacy.rs
```diff
@@ -134,6 +161,12 @@ pub fn apply_response_headers_with_cache_privacy(settings: &Settings, response:
     // the pre-apply pass could not see. Re-run the downgrade so the final
     // response can never pair Set-Cookie with shared cacheability.
     enforce_set_cookie_cache_privacy(response);
+
+    // APS creatives retain their HTTPS origin. A creative can therefore frame
+    // a publisher URL beneath the opaque renderer unless every publisher
+    // response rejects the cross-origin ancestor chain. Append this independent
+    // policy after operator headers so configuration cannot weaken it.
+    enforce_aps_publisher_frame_ancestors(settings, response);
```
prk-Jr, line 169: 🤔 **thinking** — This appends `frame-ancestors 'self'` to *every* Trusted Server response whenever `integrations.aps.enabled = true`, which silently changes the publisher's site-wide embedding policy: AMP caches, partner embeds, in-app webviews, and same-brand cross-origin framing all stop working the moment APS is turned on.

The containment argument for it is sound and the docs are honest about the trade-off, so this is not a request to change the behavior — but it deserves an explicit release note, since the blast radius reaches every page rather than the ad slot.

📌 **out of scope** — the docs promise "a separately reviewed ancestor-allowlist feature before enabling APS". Worth a tracked follow-up issue so that promise does not go stale.

crates/trusted-server-core/src/response_privacy.rs
```diff
@@ -24,6 +24,33 @@ pub const CDN_CACHE_HEADERS: &[&str] = &[
     "cloudflare-cdn-cache-control",
 ];
 
+const APS_INTEGRATION_ID: &str = "aps";
+const APS_PUBLISHER_FRAME_ANCESTORS_CSP: &str = "frame-ancestors 'self';";
+
+fn enforce_aps_publisher_frame_ancestors(settings: &Settings, response: &mut Response) {
+    let aps_enabled = settings
+        .integrations
+        .get(APS_INTEGRATION_ID)
```
prk-Jr, line 33: ♻️ **refactor** — This re-declares `APS_INTEGRATION_ID` and re-derives enablement from raw JSON. `Settings::integration_config::<ApsConfig>` and `IntegrationConfig::is_enabled` already exist and are the source of truth used by registration.

Using them keeps the two definitions from drifting if APS enablement ever gains another condition — otherwise a future change to `ApsConfig::is_enabled` leaves this header check silently out of sync with whether APS is actually running.

crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -1211,13 +1152,28 @@ impl IntegrationProxy for ApsRendererIntegration {
                     message: "Failed to build APS not-found response".to_string(),
                 });
         }
+        let (renderer_document, renderer_csp) = match request.uri().query() {
```
prk-Jr, line 1155: 🤔 **thinking** — Two different documents, with two different CSPs, are now served from one path and distinguished only by the query string, with no `Cache-Control` and no `Vary` on either response.

Any cache that drops or normalizes the query serves the wrong document to the wrong client: the bootstrap URL would return the legacy renderer (which the fallback then cannot talk to), or the query-free URL would return the bootstrap (which no legacy client understands).

**Fix**: set an explicit `Cache-Control` on both responses and confirm the query string is part of the Fastly cache key.

crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -2425,12 +2401,15 @@ mod tests {
     #[test]
     fn renderer_document_is_static_and_nonce_bound() {
         assert!(APS_RENDERER_DOCUMENT.contains("^#tsaps="));
-        assert!(APS_RENDERER_DOCUMENT.contains("event.source!==parent"));
-        assert!(APS_RENDERER_DOCUMENT.contains("message.nonce!==expected"));
+        assert!(APS_RENDERER_DOCUMENT.contains("event.source !== parent"));
+        assert!(APS_RENDERER_DOCUMENT.contains("message.nonce !== expected"));
+        assert!(APS_RENDERER_DOCUMENT.contains("['nonce', 'publisherOrigin', 'renderer']"));
```
prk-Jr, line 2406: ⛏ **nitpick** — This asserts on Prettier's exact spacing and wrapping of a file Prettier owns. A `printWidth` or formatting-config change reflows the array and breaks a core Rust test for a purely cosmetic reason.

Prefer markers that survive reformatting — for example asserting on `publisherOrigin` and `validPublisherOrigin` separately, rather than on the rendered array literal.

crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -47,72 +48,12 @@ const MAX_LANGUAGE_BYTES: usize = 8;
...
- script.onload=function(){parent.postMessage({message:'trusted-server/aps/renderer-ready',nonce:acceptedNonce},'*');};
- script.onerror=function(){parent.postMessage({message:'trusted-server/aps/renderer-failed',nonce:acceptedNonce},'*');};
- document.head.appendChild(script);
-}
-addEventListener('message',receive);
-})();
-</script>
-"#;
+const APS_RENDERER_DOCUMENT: &str =
+    include_str!("../../../trusted-server-js/lib/src/integrations/aps/renderer.html");
```
prk-Jr, line 54: 📝 **note** — Core now embeds a file from the JS crate's *source* tree (not `dist/`). This works and `include_str!` tracks it for rebuilds, so no change needed — but the cross-crate source coupling is easy to misread as a generated artifact. A one-line comment saying the HTML is hand-written source shared with the TS bundle would save the next reader a detour.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -410,24 +493,182 @@ export function renderApsCreative({ slotId, renderer: input }: RenderApsCreative
...
+
+/**
+ * Register a one-shot bearer capability for Universal Creative to request a top-page mount.
+ *
+ * Real PUC asks from a nested dynamic-renderer frame rather than the controller
+ * that received the authenticated bridge response, so source-window equality
+ * cannot survive the handoff. The 128-bit ID is disclosed only in that response,
+ * consumed once, and a newer registration revokes the same container's old ID.
+ */
+export function registerApsUniversalCreativeMount(
```
prk-Jr, line 564: 👍 **praise** — Good doc comment. It states plainly that real PUC asks from a nested frame rather than the controller that received the authenticated response, so source-window equality cannot survive the handoff, and then explains exactly what replaces it (128-bit ID, disclosed once, consumed once, revoked by re-registration). That is the kind of note that keeps someone from "hardening" this back into a broken source check later.

prk-Jr (changes requested):
## Summary

Nested opaque `data:` documents with per-layer nonces and a transferred `MessagePort` are a genuine improvement over the single sandboxed route: the envelope, bid ID, price, and creative URL now never appear in any URL, and the browser tests assert on real compromise markers rather than the absence of an error. The containment reasoning holds up under review.

Three things block merge: the top-page mount leaves the slot in a state no later GAM refresh can recover from, the legacy fallback posts a message shape its only possible consumer rejects, and the browser-specific behavior the whole design rests on is not exercised by CI.

## Blocking

### 🔧 wrench

- **Universal-creative mount permanently forks the slot** — `commit()` hides the GAM/PUC iframe forever and nothing removes the mounted APS frame on a later refresh, so the slot ends up either blank or double-rendered, and ActiveView measures a hidden iframe (`crates/trusted-server-js/lib/src/integrations/aps/render.ts:416`, inline).
- **Legacy fallback message shape can never be accepted** — the three-key message is rejected by the only document that path can reach, so the documented rollback never renders; the phase is also entirely untested (`crates/trusted-server-js/lib/src/integrations/aps/render.ts:428`, inline).
- **Cross-browser containment is not gated by CI** — `crates/trusted-server-integration-tests/browser/playwright.config.ts:16-21` defines only a `chromium` project, and `.github/workflows/integration-tests.yml:201` installs only chromium. The security model depends on browser-specific behavior this PR's own docs name explicitly: CSP inheritance into `data:` frames (the container `frame-src` confining the inner renderer) and WebKit opaque-origin propagation. The PR description reports Firefox and WebKit runs, but nothing in the repository executes them, so a regression in either behavior ships silently. Please add `firefox` and `webkit` projects — scoped to `aps-renderer.spec.ts` if full-suite runtime is a concern.

### ❓ question

- **No operator kill switch for the new rendering mode** — `mountApsRendererFrame` always builds the bootstrap URL, so the mode is baked into the JS bundle. The docs warn that the exact-origin `frame-src` "intentionally blocks an immediate creative-frame redirect" and ask operators to "validate real APS inventory for intermediate or redirect origins before rollout". If redirecting creatives turn up in production, is a JS rebuild and redeploy the only remedy? A config flag selecting legacy vs. data-bootstrap would make the rollout reversible at config-push speed.

## Non-blocking

### 🤔 thinking

- **`frame-ancestors 'self'` is site-wide and coupled to APS enablement** (`crates/trusted-server-core/src/response_privacy.rs:169`, inline).
- **Two documents on one path with no cache directives** (`crates/trusted-server-core/src/integrations/aps.rs:1155`, inline).

### ♻️ refactor

- **Derive APS enablement from the typed config** rather than raw JSON (`crates/trusted-server-core/src/response_privacy.rs:33`, inline).

### 📌 out of scope

- The ancestor-allowlist feature the docs promise before publishers with trusted external embedders can enable APS deserves a tracked follow-up issue.

### ⛏ nitpick

- **Rust test asserts on Prettier-formatted HTML** (`crates/trusted-server-core/src/integrations/aps.rs:2406`, inline).

### 📝 note

- **Cross-crate `include_str!` into the JS crate's source tree** (`crates/trusted-server-core/src/integrations/aps.rs:54`, inline).

### 👍 praise

- Per-layer nonces plus a transferred `MessagePort`, so the envelope, bid ID, price, and creative URL never appear in any URL (`render.ts:341-360`, `renderer-container.html:51-65`).
- The one-shot mount capability carries an honest doc comment about why source-window equality cannot survive the PUC handoff (`render.ts:520`, inline).
- `Number.isInteger` → `Number.isSafeInteger` in both the TypeScript and in-document validators.
- Browser tests assert on an actual compromise marker (`document.body.dataset.apsCompromised`) instead of on the absence of an error.

## CI Status

All 19 checks pass on `19c3709`.

- fmt: PASS
- clippy / cargo check (fastly, axum, cloudflare native + wasm, spin native + wasm): PASS
- rust tests (core, axum, cloudflare, spin, ts CLI, cross-adapter parity): PASS
- js tests (vitest) and format-typescript: PASS
- browser + integration tests: PASS (chromium only — see the blocking finding above)
- format-docs: PASS

crates/trusted-server-core/src/response_privacy.rs
```diff
@@ -24,6 +24,33 @@ pub const CDN_CACHE_HEADERS: &[&str] = &[
     "cloudflare-cdn-cache-control",
 ];
 
+const APS_INTEGRATION_ID: &str = "aps";
+const APS_PUBLISHER_FRAME_ANCESTORS_CSP: &str = "frame-ancestors 'self';";
+
+fn enforce_aps_publisher_frame_ancestors(settings: &Settings, response: &mut Response) {
```
aram356, line 30: 🌱 **seedling** — This is a security-sensitive gate, but the only new coverage (`aps_enabled_appends_an_independent_publisher_frame_policy`) exercises the enabled path. There is no negative test asserting the header is *not* appended when APS is absent or `enabled: false`, so a regression that always-appends or always-skips would still pass CI. Consider adding a disabled-case test asserting no `frame-ancestors` header is added.

crates/trusted-server-core/src/response_privacy.rs
```diff
@@ -24,6 +24,33 @@ pub const CDN_CACHE_HEADERS: &[&str] = &[
...
+    let aps_enabled = settings
```
aram356, line 31: 📝 **note** — This reads the raw `integrations.aps.enabled` bool directly (`== Some(true)`), while the rest of the codebase decides APS activation through `integration_config::<ApsConfig>()` / `IntegrationConfig::is_enabled`. They align today (the provider's `is_enabled()` is also just `config.enabled`), so behavior is correct — but it's a second source of truth for "is APS on" that could drift if APS ever gains an alternate enable path. A short comment noting the intentional coupling, or reusing the typed accessor, would guard against that.

crates/trusted-server-core/src/response_privacy.rs
aram356, line 169: ⛏ **nitpick** — The `frame-ancestors 'self'` policy is appended to every response type (JSON, images, the tsjs bundle), not just framed HTML. It's harmless (frame-ancestors is inert on non-framed resources) and the byte-exact dedup keeps it idempotent, but it's slightly broader than the stated "publisher document" intent. No change required — flagging for awareness.

aram356 (approved):

ChristianPavilonis:
Follow-up fix pushed in `0fad2cde`.

### Real-origin reproduction and root cause

I replayed captured real APS iframe descriptors through local Trusted Server WASM on `https://www.autoblog.com` via `ts dev proxy`, using the live Amazon runner and real Prebid Universal Creative 1.17.2.

Initial direct and PUC rendering both completed: one bootstrap request, one Amazon runner request, one creative request, and PUC `adRenderSucceeded`. The regression was the next GPT lifecycle: after PUC committed, TSJS left the APS frame mounted and the GAM/PUC controller at `display:none`, so a later refresh could not reclaim the slot.

The pre-fix refresh replay retained the APS frame and hidden controller. The post-fix replay removes the committed APS frame and restores the controller (`display:inline`) for the exact requested slot.

### Fix

- Track committed APS mounts and exact prior sibling `display` value/priority.
- On GPT `slotRequested`, cancel pending work, remove the committed APS frame, restore the controller, and revoke unconsumed mount capabilities for that container.
- Force-hide the controller with inline `display:none !important` and restore/remove the exact prior declaration.
- Preserve the one-shot descriptor/nonce/source/port checks and no-dual-visible-surface behavior.
- Repair old-server rollback compatibility by posting the exact legacy `{ nonce, renderer }` shape without abandoning a delayed modern `bootstrap-ready`.
- Preflight nonce/capacity failures before replacing a working committed mount.
- Add Firefox/WebKit APS-only CI projects while retaining the complete Next.js and WordPress suites on Chromium.

### Validation

- Real Autoblog direct iframe replay: passed with live runner and creative request.
- Real Autoblog PUC refresh replay: passed; APS frame removed and controller restored.
- TSJS Vitest: 838/838.
- TSJS build, ESLint, and Prettier: passed.
- Rust: `cargo test-fastly`, `cargo test-axum`, `cargo test-cloudflare`, and `cargo test-spin` passed.
- Adapter parity: 13/13.
- All six adapter Clippy targets passed.
- APS Playwright:
  - Chromium / Next.js: 7/7
  - Chromium / WordPress: 7/7
  - Firefox + WebKit / Next.js: 14/14
- Rustfmt, docs Prettier, workflow YAML, shell syntax, and `git diff --check`: passed.

Independent final review found no blocker, high, or medium findings.

aram356:
closing in place of #1042

PR #1034, opened 2026-08-17
Add GAM cohort attribution for Trusted Server traffic

### LINKED ISSUE #1027: GAM attribution for Trusted Server A/B traffic (page-delivery marker)
## Objective

Give Google Ad Manager a way to segment the Trusted Server A/B experiment. A small, cookie-sticky treatment cohort is routed through Trusted Server while control continues on the production path. We add a **page-level GPT key-value marker** on every request from a TS-delivered document so GAM reporting can compare treatment impressions/clicks against the unmarked production baseline — without changing auction eligibility, ad delivery, consent, or performance.

Provisional marker: `ts=true` (key name pending the preflight below). Production stays unmarked; the control cohort is derived by subtraction within one experiment scope.

## Design

- Spec: `docs/superpowers/specs/2026-07-15-gam-ts-cohort-attribution-design.md`
- Design PR (docs-only, `Status: Design`): #1028

Mechanism: enqueue `googletag.setConfig({ targeting: { <key>: 'true' } })` in the early GPT head bootstrap (before the `ts.adInit` guard), with a defensive fallback in the TSJS GPT bundle gated on a non-executable activation attribute. Kept distinct from the existing slot-level `ts_initial` key and excluded from slot-cleanup arrays.

## Acceptance criteria

- Every in-scope GPT PubAds request from a TS-rewritten treatment document carries the marker (initial, lazy, refreshed, publisher-owned, SPA-route).
- Production/control pages remain unmodified and unmarked.
- `ts_initial=1` keeps its current slot-level lifecycle; the marker is not cleared by Prebid refresh or SPA cleanup.
- No GAM targeting consumer changes ad eligibility/pricing/protection/routing because of the key.
- No personal data, no new network request, no blocking work.
- GAM can report treatment counts under the marker and derive control by subtraction within one scope; coverage validated against router allocation before results are interpreted.

## Work / blockers

- [ ] **Key-name preflight (blocking, do first).** Confirm the enforced GAM key-name length limit and provisioning surface (UI vs REST vs SOAP): SOAP/REST document a 10-char `CustomTargetingKey.name` limit; other Help Center material implies 20. Attempt to create `trusted_server` (14) in the target network; record the enforced key-name and value limits. Then finalize the marker contract — exactly one key, one value, no dual-write. If the limit permits, a more descriptive, less collision-prone name is preferred over the short, collision-prone `ts`. This is a product-contract decision, not an automatic fallback to `ts`.
- [ ] **Fold in design-review findings** before writing the plan: label GAM results as descriptive delivery attribution (not causal); make the production partition executable (exact eligible units, identical A/B report filters, preflight zero-counts for every excluded path, `0 ≤ Report B ≤ Report A` invariants); acknowledge monitoring cannot prove per-response completeness; specify rollback drain ordering; extend the collision audit to Prebid-generated slot targeting (`pbjs.bidderSettings`, `adserverTargeting`, `setTargetingForGPTAsync`); preflight Enhanced key-value reporting availability/billing; soften the nested-document claim (a publisher can clone the tag via `srcdoc`/`document.write`).
- [ ] **Rebase the spec branch on current `main`** and re-anchor file/line references; `main` changed the GPT/Prebid refresh surface in #912 (`bf61fa189`).
- [ ] **Write the implementation plan** (`docs/superpowers/plans/2026-07-15-gam-ts-cohort-attribution.md`) — only after the key contract is finalized and the branch is rebased.
- [ ] **Implement** per the spec's Component changes: early bootstrap marker + source-order tests; attribute-gated bundle fallback; publisher-page-only activation attribute in `tsjs.rs`/`html_processor.rs`; JSDOM bootstrap tests; slot-cleanup / operator-targeting / Prebid-collision characterization tests.
- [ ] **Browser/GAM validation** before launch (initial/lazy/refresh marker presence, control absence, CSP execution, `ts_initial` isolation, short GAM report).
- [ ] **GAM configuration** in the target network: create the reportable key + predefined value; run the cross-system collision audit as a hard launch gate.

## Non-goals

The cookie A/B router (owned externally); proving a TS server-side bid won; billing-grade or causal analysis from GAM alone; marking IMA/video, direct-tag, server-side, or non-independently-rewritten nested inventory.

prk-Jr:
## Summary

- Add a default-off `gam_attribution_enabled` GPT option that applies fixed page-level `ts=true` targeting to eligible publisher documents whose rewritten head is emitted by Trusted Server, allowing GAM to describe treatment delivery while the production/control path remains unmarked.
- Apply targeting before publisher GPT initialization through the early bootstrap, with an exact `document.currentScript` attribute-gated TSJS fallback that fails closed and isolates `setConfig` failures.
- Preserve auction behavior, consent, the slot-level `ts_initial` lifecycle, and publisher/Prebid targeting while documenting the collision audit, reporting contract, rollout, monitoring limits, and drain-safe rollback sequence.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-cli/tests/config_env_overlay.rs` | Verify the GAM attribution environment override is applied when the corresponding TOML leaf exists. |
| `crates/trusted-server-core/src/creative_opportunities.rs` | Characterize that operator-provided slot-level `ts` targeting remains forwarded unchanged. |
| `crates/trusted-server-core/src/html_processor.rs` | Add integration-provided attributes to publisher TSJS tags and cover enabled, disabled, and source-order behavior. |
| `crates/trusted-server-core/src/integrations/gpt.rs` | Add the default-off setting and drive both inline activation and fallback authorization from the parsed GPT configuration. |
| `crates/trusted-server-core/src/integrations/gpt_bootstrap.js` | Queue page-level `ts=true` before the existing `adInit` guard and isolate attribution failures from later GPT commands. |
| `crates/trusted-server-core/src/integrations/registry.rs` | Add and test the integration hook that supplies trusted static attributes for publisher bundle tags. |
| `crates/trusted-server-core/src/publisher.rs` | Cover streaming-head attribution semantics and operator-targeting preservation. |
| `crates/trusted-server-core/src/tsjs.rs` | Render trusted static attributes on publisher TSJS tags without changing generic or creative tag output. |
| `crates/trusted-server-integration-tests/browser/tests/shared/script-injection.spec.ts` | Verify attribution metadata is absent from the default-off integration fixture. |
| `crates/trusted-server-integration-tests/fixtures/configs/trusted-server.integration.toml` | Keep the new configuration leaf explicit and disabled in integration tests. |
| `crates/trusted-server-js/lib/src/integrations/gpt/index.ts` | Add the executing-script-gated bundle fallback for page-level GAM targeting. |
| `crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts` | Verify SPA cleanup does not clear the page-level `ts` key. |
| `crates/trusted-server-js/lib/test/integrations/gpt/gpt_bootstrap.test.ts` | Exercise activation ordering, disabled behavior, unavailable or throwing `setConfig`, and bootstrap continuity. |
| `crates/trusted-server-js/lib/test/integrations/gpt/index.test.ts` | Cover exact fallback authorization, fail-closed cases, duplicate-ID decoys, cloned tags, and error isolation. |
| `crates/trusted-server-js/lib/test/integrations/prebid/index.test.ts` | Characterize publisher and Prebid-generated `ts` targeting across refreshes. |
| `docs/guide/integrations/gpt.md` | Document configuration, environment-overlay requirements, attribution semantics, collision checks, reporting, rollout, and rollback. |
| `docs/superpowers/plans/2026-07-15-gam-ts-cohort-attribution.md` | Add the reviewed implementation and verification plan. |
| `docs/superpowers/specs/2026-07-15-gam-ts-cohort-attribution-design.md` | Finalize the fixed marker contract, streaming meaning, reporting scope, and operational safeguards. |
| `trusted-server.example.toml` | Add the explicit default-off setting and environment-overlay guidance. |

## Launch gates

This PR completes the repository implementation. Starting the cohort remains gated on target-network GAM provisioning, cross-system collision and privacy/CSP audits, a paired-report dry run, and router-controlled treatment validation described in the design.

GAM results are descriptive delivery attribution, not causal treatment-effect estimates.

## Closes

Closes #1027

## Test plan

- [x] JS tests: `cd crates/trusted-server-js/lib && npx vitest run` (44 files, 847 tests)
- [x] Manual testing via `fastly compute serve`: confirmed the inline flag, activation attribute, GPT page targeting, and decoded `cust_params` containing `ts=true` on a real GAM request
- [x] Other: Cloudflare and Spin adapter tests/clippy, cross-adapter parity, native CLI tests, and Next.js/WordPress browser integration suites

## Checklist

- [x] No `unwrap()` added to production code
- [x] Logging follows project conventions; no direct stdout/stderr logging was added
- [x] New code has tests

### DIFF
--- a/crates/trusted-server-core/src/integrations/gpt_bootstrap.js
+++ b/crates/trusted-server-core/src/integrations/gpt_bootstrap.js
@@ -27,10 +27,13 @@
         var gpt = window.googletag;
         if (gpt && typeof gpt.setConfig === "function") {
           // "ts" is the fixed GAM key, not the local window.tsjs alias.
-          gpt.setConfig({ targeting: { ts: 'true' } });
+          gpt.setConfig({ targeting: { ts: "true" } });
         }
-      } catch (_) {
+      } catch (error) {
         // Attribution must not interrupt the existing bootstrap queue.
+        ts.log &&
+          ts.log.warn &&
+          ts.log.warn("GAM attribution targeting failed", error);
       }
     });
   }
--- a/crates/trusted-server-core/src/integrations/gpt_bootstrap.js
+++ b/crates/trusted-server-core/src/integrations/gpt_bootstrap.js
@@ -17,6 +17,24 @@
 (function () {
   if (typeof window === "undefined") return;
   var ts = (window.tsjs = window.tsjs || {});
+  var tag;
+
+  if (window.__tsjs_gam_attribution_enabled === true) {
+    tag = window.googletag = window.googletag || { cmd: [] };
+    tag.cmd = tag.cmd || [];
+    tag.cmd.push(function () {
+      try {
+        var gpt = window.googletag;
+        if (gpt && typeof gpt.setConfig === "function") {
+          // "ts" is the fixed GAM key, not the local window.tsjs alias.
+          gpt.setConfig({ targeting: { ts: 'true' } });
+        }
+      } catch (_) {
+        // Attribution must not interrupt the existing bootstrap queue.
+      }
+    });
+  }
+
   if (ts.adInit) return;
 
   // Track whether the publisher disabled GPT initial load. Read the effective
@@ -38,7 +56,9 @@
     return true;
   }
 
-  (window.googletag = window.googletag || { cmd: [] }).cmd.push(function () {
+  tag = tag || (window.googletag = window.googletag || { cmd: [] });
+  tag.cmd = tag.cmd || [];
+  tag.cmd.push(function () {
     var gpt = window.googletag;
     syncInitialLoadDisabled(gpt);
     if (
--- a/crates/trusted-server-core/src/integrations/registry.rs
+++ b/crates/trusted-server-core/src/integrations/registry.rs
@@ -1061,11 +1061,21 @@ impl IntegrationRegistry {
     /// Collect static attributes for the publisher TSJS bundle tag.
     #[must_use]
     pub fn tsjs_script_tag_attributes(&self) -> Vec<(&'static str, &'static str)> {
-        let mut attributes = Vec::new();
+        let mut attributes: Vec<(&'static str, &'static str)> = Vec::new();
         for injector in &self.inner.head_injectors {
             for attribute in injector.tsjs_script_tag_attributes() {
-                if !attributes.iter().any(|(name, _)| *name == attribute.0) {
-                    attributes.push(attribute);
+                let existing = attributes
+                    .iter()
+                    .find(|(name, _)| *name == attribute.0)
+                    .copied();
+                match existing {
+                    None => attributes.push(attribute),
+                    Some((_, kept_value)) if kept_value != attribute.1 => log::warn!(
+                        "Integration `{}` emits conflicting value for publisher tag attribute `{}`; keeping the first",
+                        injector.integration_id(),
+                        attribute.0
+                    ),
+                    Some(_) => {}
                 }
             }
         }
--- a/crates/trusted-server-core/src/integrations/registry.rs
+++ b/crates/trusted-server-core/src/integrations/registry.rs
@@ -1061,11 +1061,15 @@ impl IntegrationRegistry {
     /// Collect static attributes for the publisher TSJS bundle tag.
     #[must_use]
     pub fn tsjs_script_tag_attributes(&self) -> Vec<(&'static str, &'static str)> {
-        self.inner
-            .head_injectors
-            .iter()
-            .flat_map(|injector| injector.tsjs_script_tag_attributes())
-            .collect()
+        let mut attributes = Vec::new();
+        for injector in &self.inner.head_injectors {
+            for attribute in injector.tsjs_script_tag_attributes() {
+                if !attributes.iter().any(|(name, _)| *name == attribute.0) {
+                    attributes.push(attribute);
+                }
+            }
+        }
+        attributes
     }
 
     /// Provide a snapshot of registered integrations and their hooks.
@@ -1367,6 +1371,25 @@ mod tests {
         }
     }
 
+    struct ConflictingMetadataHeadInjector;
+
+    impl IntegrationHeadInjector for ConflictingMetadataHeadInjector {
+        fn integration_id(&self) -> &'static str {
+            "conflicting-metadata"
+        }
+
+        fn head_inserts(&self, _ctx: &IntegrationHtmlContext<'_>) -> Vec<String> {
+            Vec::new()
+        }
+
+        fn tsjs_script_tag_attributes(&self) -> Vec<(&'static str, &'static str)> {
+            vec![
+                ("data-ts-gam-attribution", "false"),
+                ("data-third-attribute", "third"),
+            ]
+        }
+    }
+
     #[test]
     fn tsjs_script_tag_attributes_preserve_registration_order_and_default_empty() {
         let registry = IntegrationRegistry::from_rewriters_with_head_injectors(
@@ -1375,6 +1398,7 @@ mod tests {
             vec![
                 Arc::new(DefaultMetadataHeadInjector),
                 Arc::new(StaticMetadataHeadInjector),
+                Arc::new(ConflictingMetadataHeadInjector),
             ],
         );
 
@@ -1383,8 +1407,9 @@ mod tests {
             vec![
                 ("data-ts-gam-attribution", "true"),
                 ("data-test-order", "second"),
+                ("data-third-attribute", "third"),
             ],
-            "should omit default-empty metadata and preserve registered attribute order"
+            "should keep the first value for duplicate names and preserve attribute order"
         );
     }
 
--- a/crates/trusted-server-core/src/integrations/registry.rs
+++ b/crates/trusted-server-core/src/integrations/registry.rs
@@ -1058,6 +1058,16 @@ impl IntegrationRegistry {
         inserts
     }
 
+    /// Collect static attributes for the publisher TSJS bundle tag.
+    #[must_use]
+    pub fn tsjs_script_tag_attributes(&self) -> Vec<(&'static str, &'static str)> {
+        self.inner
+            .head_injectors
+            .iter()
+            .flat_map(|injector| injector.tsjs_script_tag_attributes())
+            .collect()
+    }
+
     /// Provide a snapshot of registered integrations and their hooks.
     #[must_use]
     pub fn registered_integrations(&self) -> Vec<IntegrationMetadata> {
@@ -1325,6 +1335,58 @@ mod tests {
     use crate::constants::COOKIE_TS_EC;
     use crate::platform::test_support::noop_services;
     use http::{HeaderValue, StatusCode, header};
+
+    struct DefaultMetadataHeadInjector;
+
+    impl IntegrationHeadInjector for DefaultMetadataHeadInjector {
+        fn integration_id(&self) -> &'static str {
+            "default-metadata"
+        }
+
+        fn head_inserts(&self, _ctx: &IntegrationHtmlContext<'_>) -> Vec<String> {
+            Vec::new()
+        }
+    }
+
+    struct StaticMetadataHeadInjector;
+
+    impl IntegrationHeadInjector for StaticMetadataHeadInjector {
+        fn integration_id(&self) -> &'static str {
+            "static-metadata"
+        }
+
+        fn head_inserts(&self, _ctx: &IntegrationHtmlContext<'_>) -> Vec<String> {
+            Vec::new()
+        }
+
+        fn tsjs_script_tag_attributes(&self) -> Vec<(&'static str, &'static str)> {
+            vec![
+                ("data-ts-gam-attribution", "true"),
+                ("data-test-order", "second"),
+            ]
+        }
+    }
+
+    #[test]
+    fn tsjs_script_tag_attributes_preserve_registration_order_and_default_empty() {
+        let registry = IntegrationRegistry::from_rewriters_with_head_injectors(
+            Vec::new(),
+            Vec::new(),
+            vec![
+                Arc::new(DefaultMetadataHeadInjector),
+                Arc::new(StaticMetadataHeadInjector),
+            ],
+        );
+
+        assert_eq!(
+            registry.tsjs_script_tag_attributes(),
+            vec![
+                ("data-ts-gam-attribution", "true"),
+                ("data-test-order", "second"),
+            ],
+            "should omit default-empty metadata and preserve registered attribute order"
+        );
+    }
 
     // Mock integration proxy for testing
     struct MockProxy;
--- a/crates/trusted-server-core/src/integrations/registry.rs
+++ b/crates/trusted-server-core/src/integrations/registry.rs
@@ -575,6 +575,11 @@ pub trait IntegrationHeadInjector: Send + Sync {
     fn integration_id(&self) -> &'static str;
     /// Return HTML snippets to insert at the start of `<head>`.
     fn head_inserts(&self, ctx: &IntegrationHtmlContext<'_>) -> Vec<String>;
+
+    /// Return attributes to add to the publisher TSJS bundle tag.
+    fn tsjs_script_tag_attributes(&self) -> Vec<(&'static str, &'static str)> {
+        Vec::new()
+    }
 }
 
 /// Registration payload returned by integration builders.
--- a/crates/trusted-server-core/src/tsjs.rs
+++ b/crates/trusted-server-core/src/tsjs.rs
@@ -235,16 +235,27 @@ mod tests {
     }
 
     #[test]
-    fn publisher_tsjs_script_tag_rejects_html_sensitive_attribute_values() {
-        for value in ["bad\"value", "bad&value", "bad<value", "bad>value"] {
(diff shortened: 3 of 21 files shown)


ChristianPavilonis (approved):
## Summary

Approved after a thorough review of the GAM cohort-attribution implementation. I found no actionable correctness, security, compatibility, or test-coverage findings.

Several required GitHub Actions jobs are currently red because runner setup could not download actions from GitHub codeload (429/503); rerun those checks before merge.

crates/trusted-server-core/src/integrations/registry.rs
```diff
@@ -1053,6 +1058,16 @@ impl IntegrationRegistry {
         inserts
     }
 
+    /// Collect static attributes for the publisher TSJS bundle tag.
+    #[must_use]
+    pub fn tsjs_script_tag_attributes(&self) -> Vec<(&'static str, &'static str)> {
```
aram356, line 1063: 🤔 **thinking** — `tsjs_script_tag_attributes()` flat-maps all head injectors with no duplicate-name protection. If a second integration ever emits an attribute name that GPT also emits, the publisher tag renders a duplicate HTML attribute (a parse error; browsers keep the first occurrence). Only GPT implements the hook today, so this is latent.

**Fix**: dedupe first-wins by attribute name here, or document name-uniqueness as an implementor contract on `IntegrationHeadInjector::tsjs_script_tag_attributes()`.

crates/trusted-server-core/src/tsjs.rs
```diff
@@ -11,9 +11,23 @@ pub fn tsjs_script_src(module_ids: &[&str]) -> String {
 /// `<script>` tag for injecting the tsjs bundle.
 #[must_use]
 pub fn tsjs_script_tag(module_ids: &[&str]) -> String {
+    tsjs_script_tag_with_attributes(module_ids, &[])
+}
+
+/// Publisher `<script>` tag for the tsjs bundle with trusted static attributes.
+#[must_use]
+pub fn tsjs_script_tag_with_attributes(
```
aram356, line 19: 🌱 **seedling** — `&'static str` restricts attribute names/values to compile-time strings, but nothing stops a future implementor from embedding `"` or `>` and breaking the tag. A `debug_assert!` on name/value charset (e.g. names `[a-z-]`, values without `"`, `<`, `>`, `&`) would make the trusted-static contract self-enforcing at test time without runtime cost.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -17,6 +17,24 @@
...
+    tag = window.googletag = window.googletag || { cmd: [] };
+    tag.cmd = tag.cmd || [];
+    tag.cmd.push(function () {
+      try {
+        var gpt = window.googletag;
+        if (gpt && typeof gpt.setConfig === "function") {
+          // "ts" is the fixed GAM key, not the local window.tsjs alias.
+          gpt.setConfig({ targeting: { ts: 'true' } });
+        }
+      } catch (_) {
```
aram356: 🔧 **wrench** — The bundle-side twin logs `log.warn('[tsjs-gpt] GAM attribution targeting failed', …)`, but this catch is empty. A throwing publisher-wrapped `setConfig` silently drops the marker — exactly the misclassified-as-control failure the design's coverage monitoring section calls out. The file's established guarded pattern keeps this safe inside the queue:

```js
} catch (error) {
  // Attribution must not interrupt the existing bootstrap queue.
  ts.log && ts.log.warn && ts.log.warn("GAM attribution targeting failed", error);
}
```

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
```diff
@@ -17,6 +17,24 @@
...
+
+  if (window.__tsjs_gam_attribution_enabled === true) {
+    tag = window.googletag = window.googletag || { cmd: [] };
+    tag.cmd = tag.cmd || [];
+    tag.cmd.push(function () {
+      try {
+        var gpt = window.googletag;
+        if (gpt && typeof gpt.setConfig === "function") {
+          // "ts" is the fixed GAM key, not the local window.tsjs alias.
+          gpt.setConfig({ targeting: { ts: 'true' } });
```
aram356: ⛏ **nitpick** — Single-quoted `'true'` in a file that otherwise uses double quotes. Note the ordering test in `gpt.rs` string-matches this exact literal (`gpt.setConfig({ targeting: { ts: 'true' } })`), so normalizing the quotes means updating both together — the test fails loudly if they drift, so this is consistency only.

crates/trusted-server-js/lib/test/integrations/gpt/gpt_bootstrap.test.ts
```diff
@@ -33,15 +33,34 @@ interface MockGoogleTag {
   pubads: () => unknown;
   enableServices: () => void;
   display: (divId: string) => void;
+  getConfig?: (key: string) => Record<string, unknown>;
```
aram356: ⛏ **nitpick** — `getConfig` is declared on `MockGoogleTag` but never exercised by any test in this file. Drop it, or add coverage that uses it.

aram356 (changes requested):
## Summary

Tightly scoped implementation of the default-off `gam_attribution_enabled` marker. The risky surfaces check out: attribution is queued before the `adInit` guard and any `display()`/`refresh()`, the bundle fallback fails closed on `document.currentScript` (duplicate-ID decoys and cloned tags covered), the default path is byte-for-byte unchanged, creative documents keep the unmarked unified tag, streaming emits both activation signals in the first head chunk, and no cleanup path touches the page-level `ts` key. The double `setConfig` application (bootstrap + bundle) is intentional and idempotent per the design doc.

No correctness bugs found; the requested changes below are hardening and consistency items on the new surfaces.

## CI Status

- fmt: PASS
- clippy (fastly/axum/cloudflare/cloudflare-wasm/spin-native/spin-wasm): PASS
- js tests (vitest, browser integration): PASS
- formats (typescript/docs), CodeQL: PASS

trusted-server.example.toml
```diff
@@ -104,6 +104,9 @@ rewrite_sdk = true
 
 [integrations.gpt]
 enabled = false
+# Keep this leaf present when using the corresponding EdgeZero v0.0.4
```
aram356: ⛏ **nitpick** — Pinning "EdgeZero v0.0.4" in the comment (and in `docs/guide/integrations/gpt.md`) will rot when EdgeZero moves past the leaf-creation limitation; both locations then need a coordinated update. Consider phrasing it capability-based ("the environment overlay cannot create a missing configuration leaf") so the comment stays true across versions. Same convention already exists at line 124, so this is consistency-with-a-flaw rather than a new problem.

crates/trusted-server-integration-tests/fixtures/configs/trusted-server.integration.toml
```diff
@@ -86,6 +86,7 @@ rewrite_sdk = true
 
 [integrations.gpt]
 enabled = false
+gam_attribution_enabled = false
```
aram356, line 89: 📝 **note** — This fixture line is load-bearing beyond the integration suite: `config_env_overlay.rs` in `trusted-server-cli` includes this file via `include_str!` and its new `migrated_legacy_config_applies_gam_attribution_environment_override` test only passes because this leaf exists (the EdgeZero overlay cannot create a missing leaf). Removing this line would fail a seemingly unrelated CLI test. Same coupling pattern as the existing rewrite/sanitize overlay leaves, so no action required — just flagging the dependency for future editors.

crates/trusted-server-core/src/integrations/registry.rs
prk-Jr: Implemented first-wins deduplication at the registry aggregation boundary while preserving registration order. Added regression coverage with a later conflicting value and a unique attribute.

crates/trusted-server-core/src/tsjs.rs
prk-Jr: Added debug assertions at the rendering boundary: names must be non-empty lowercase ASCII letters, digits, or hyphens, and values reject double quote, ampersand, less-than, and greater-than characters. Native panic tests cover each invalid case; release behavior remains unchanged.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
prk-Jr: Added the guarded ts.log.warn call in the attribution catch. The throwing-setConfig Vitest now verifies the warning while also confirming later publisher commands and bootstrap installation continue.

crates/trusted-server-core/src/integrations/gpt_bootstrap.js
prk-Jr: Normalized the targeting value to double quotes and updated the matching Rust source-order assertion in the same change.

crates/trusted-server-js/lib/test/integrations/gpt/gpt_bootstrap.test.ts
prk-Jr: Removed the unused getConfig member from the MockGoogleTag interface.

trusted-server.example.toml
prk-Jr: Replaced the version-specific wording with capability-based guidance in both trusted-server.example.toml and the GPT integration guide.

crates/trusted-server-integration-tests/fixtures/configs/trusted-server.integration.toml
prk-Jr: Confirmed. The fixture leaf remains unchanged by the review-resolution commits, and the CLI environment-overlay suite including the GAM attribution migration case passes.

crates/trusted-server-core/src/integrations/registry.rs
```diff
@@ -1053,6 +1058,20 @@ impl IntegrationRegistry {
...
+        let mut attributes = Vec::new();
+        for injector in &self.inner.head_injectors {
+            for attribute in injector.tsjs_script_tag_attributes() {
+                if !attributes.iter().any(|(name, _)| *name == attribute.0) {
+                    attributes.push(attribute);
+                }
+            }
+        }
+        attributes
```
aram356: ♻️ **refactor** — First-wins dedupe silently drops a *conflicting* duplicate value. The new test documents that a second injector emitting `data-ts-gam-attribution="false"` is ignored without a trace; that is the right precedence, but a dropped conflicting value indicates cross-integration misconfiguration worth surfacing, consistent with the registry's existing `log::warn!` skips. Identical duplicate values stay silent.

```suggestion
        let mut attributes: Vec<(&'static str, &'static str)> = Vec::new();
        for injector in &self.inner.head_injectors {
            for attribute in injector.tsjs_script_tag_attributes() {
                let existing = attributes
                    .iter()
                    .find(|(name, _)| *name == attribute.0)
                    .copied();
                match existing {
                    None => attributes.push(attribute),
                    Some((_, kept_value)) if kept_value != attribute.1 => log::warn!(
                        "Integration `{}` emits conflicting value for publisher tag attribute `{}`; keeping the first",
                        injector.integration_id(),
                        attribute.0
                    ),
                    Some(_) => {}
                }
            }
        }
        attributes
```

trusted-server.example.toml
```diff
@@ -104,6 +104,9 @@ rewrite_sdk = true
 
 [integrations.gpt]
 enabled = false
+# Keep this leaf present when the environment overlay cannot create a missing
+# configuration leaf. Attribution remains off until explicitly enabled.
```
aram356: ⛏ **nitpick** — The comment reads circularly: "keep this leaf present *when* the environment overlay cannot create a missing configuration leaf" — the overlay can never create one. The intent is "keep it so the override can apply".

```suggestion
# Keep this leaf present so the environment override can apply; the overlay
# cannot create a missing configuration leaf. Attribution stays off until enabled.
```

aram356 (approved):
## Summary

Re-review at bbadec37: all seven findings from the previous review are resolved — first-wins attribute dedupe with a conflict test, debug-build charset guards on publisher tag attributes with WASM-safe panic tests, guarded `ts.log.warn` in the bootstrap attribution catch with test coverage, quote normalization synchronized with the Rust ordering test, dead mock member removed, and capability-based overlay wording. Approving; the two inline comments below are optional polish.

> Both inline comments carry a one-click GitHub `suggestion` — use **Commit suggestion** (or **Add suggestion to batch**) to apply them as commits on the PR branch.

## Non-blocking

### ♻️ refactor

- **Warn when a conflicting duplicate attribute value is dropped** — see inline at `crates/trusted-server-core/src/integrations/registry.rs:1064`

### ⛏ nitpick

- **Example-TOML overlay comment reads circularly** — see inline at `trusted-server.example.toml:107`

## CI Status

All 19 checks PASS: cargo fmt (required), cargo test (required), format-typescript (required), format-docs (required), cargo test (axum native), cargo test (cross-adapter parity), cargo test (ts CLI, native), cargo check (cloudflare native + wasm32-unknown-unknown), cargo check/build/test (spin native + wasm32-wasip1), vitest, browser integration tests, integration tests, integration tests (Fastly EC lifecycle), prepare integration artifacts, Analyze (actions), Analyze (javascript-typescript) ×2, Analyze (rust), CodeQL.

PR #1042, opened 2026-08-19
Add publisher-native rendering for selected APS bids

### LINKED ISSUE #999: APS creatives requiring nested same-origin semantics render blank

ChristianPavilonis:
## Summary

- Adds an opt-in `publisher_native` mode for rendering Trusted Server-selected APS bids through APS's fixed Prebid creative runner in a publisher-origin friendly iframe.
- Keeps the existing opaque Trusted Server renderer as the default and disables its static route only when `rendering_mode = "publisher_native"` is explicitly configured.
- Keeps the change self-contained: Trusted Server injects the runner bootstrap, so publishers do not need to install custom JavaScript.

The existing opaque renderer can leave APS creatives blank when their nested HTTPS frames require their real origin. A controlled browser test through the publisher setup confirmed that the new friendly-frame path renders a real selected APS creative, while `apstag.renderImp(document, bidId)` cannot render the server-selected bid because it is absent from APS's browser-auction state.

The working path reuses the existing `prebid/creative/render` runner contract inside a friendly frame. Trusted Server does not call `apstag.fetchBids()`, call `apstag.setDisplayBids()`, mutate the publisher's APS SDK, or start another auction. The runner contract is observed vendor behavior rather than a documented external-response API, so APS account-team validation is still required before production use.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/integrations/aps.rs` | Adds the typed rendering mode, emits a CSP-safe native-mode marker, and omits the static renderer route in native mode. |
| `crates/trusted-server-integration-tests/browser/tests/shared/aps-renderer.spec.ts` | Adds a real-browser proof that native mode queues the selected response, loads the fixed runner, and creates no opaque renderer frame. |
| `crates/trusted-server-js/lib/src/core/request.ts` | Routes direct auction APS winners through the shared rendering-owner dispatcher. |
| `crates/trusted-server-js/lib/src/integrations/aps/render.ts` | Validates descriptors and implements the friendly-frame runner with slot resolution, replacement, timeout, failure, and no-fallback behavior. |
| `crates/trusted-server-js/lib/src/integrations/gpt/index.ts` | Applies the same rendering owner to server and Prebid APS paths while preserving one-shot ownership. |
| `crates/trusted-server-js/lib/test/core/request.test.ts` | Covers default opaque rendering and direct native-runner rendering. |
| `crates/trusted-server-js/lib/test/integrations/aps/render.test.ts` | Covers queue shape, runner load/failure, logical GPT slots, replacement, stale completion, timeout, and publisher-state errors. |
| `crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts` | Covers accepted and failed native-runner handoffs for server and Prebid APS capabilities. |
| `docs/guide/integrations/aps.md` | Documents configuration, implementation limits, CSP and security trade-offs, duplicate-demand risk, and rollout requirements. |
| `trusted-server.example.toml` | Shows the default rendering mode and controlled native experiment opt-in. |

## Scope

This PR is intentionally limited to configuration, bid ownership, browser rendering, focused tests, and operator documentation. It does not add infrastructure, change bid selection, call publisher `apstag` methods, or claim a publicly supported APS external-response API.

The Rust and TypeScript changes are both required: the mode must be selected and exposed server-side, then enforced consistently across direct auctions and Google Ad Manager/Prebid rendering paths. Native mode deliberately uses a friendly iframe without the default opaque-origin sandbox, so it has a larger security surface and must remain isolated during testing.

## Closes

Related to #999.

## Test plan

- [x] JS tests: `cd crates/trusted-server-js/lib && npx vitest run` — 44 files, 842 tests
- [x] Other: Cloudflare and Spin tests; all six target-specific Clippy commands; adapter parity; JS lint and bundle build
- [x] Other: APS Playwright suite — 4 tests, including the publisher-native friendly-frame proof
- [x] Other: controlled browser test through `ts dev proxy` rendered a real selected APS creative with no `/integrations/aps/renderer` request

## Checklist

- [x] Uses project logging macros (not `println!`)
- [x] New code has tests


crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -38,6 +43,36 @@ type ValidatedRendererCacheEntry = {
...
+      const mapped = findSlot(divId);
+      if (mapped) return mapped;
+    }
+
+    const configuredDivId = window.tsjs?.adSlots?.find((slot) => slot.id === slotId)?.div_id;
+    return configuredDivId ? findSlot(configuredDivId) : null;
+  } catch {
+    return null;
+  }
+}
```
prk-Jr: 🔧 **wrench** — `findApsContainer` bypasses the codebase's slot-root resolution, so native rendering mis-targets or silently declines on real GAM pages.

The Prebid path passes `prebidRendererEntry.adUnitCode` as `slotId` (`gpt/index.ts:1713`). `prebid/index.ts:553-558` documents that an ad unit code is `div_id`, `${div_id}-container`, **or** a prefix match of a dynamic div id — which is exactly why the ownership check on the line above uses `candidateSlotRootsForConfiguredDivId` (`gpt/index.ts:194-205`). `findApsContainer` does none of that: exact `findSlot(slotId)`, then a reverse `divToSlotId` lookup whose *values* are logical slot ids (so an adUnitCode never matches), then `adSlots[].id` (also a logical slot id).

Two consequences:

1. **`-container` ad unit code** — `findSlot` resolves the wrapper, the frame is appended there, and `finish()`'s sibling sweep at line 455 deletes the inner ad div.
2. **Dynamic div id** (`div-header-1` for a configured `div-header`) — the container resolves to `null` and the bid is declined. But by then `consumeApsPrebidRenderer` and `recordConsumedPrebidApsId` have burned the capability and `e.stopImmediatePropagation()` has already blocked Prebid.js, so the slot goes permanently blank with no fallback.

The ownership check passes in both cases because it *is* prefix-aware. Only container resolution fails, so the failure is silent.

**Proposed fix** (apply manually — spans `gpt/index.ts` and `render.ts` and needs a new dynamic-div-id test, so it can't be a single-file `suggestion`):

```ts
// gpt/index.ts — export the already-correct resolver
export function apsSlotRootsForCode(code: string): HTMLElement[] {
  return candidateSlotRootsForConfiguredDivId(code);
}

// aps/render.ts — prefer an inner div over a -container wrapper
function findApsContainer(slotId: string): HTMLElement | null {
  const direct = findSlot(slotId);
  if (direct) return direct;
  try {
    for (const [divId, mappedSlotId] of Object.entries(window.tsjs?.divToSlotId ?? {})) {
      if (mappedSlotId !== slotId) continue;
      const mapped = findSlot(divId);
      if (mapped) return mapped;
    }
    const configuredDivId = window.tsjs?.adSlots?.find((s) => s.id === slotId)?.div_id;
    if (configuredDivId) {
      const configured = findSlot(configuredDivId);
      if (configured) return configured;
    }
    // Last resort: the same prefix / -container aware roots the ownership check uses.
    const roots = apsSlotRootsForCode(slotId);
    return roots.find((r) => !r.id.endsWith('-container')) ?? roots[0] ?? null;
  } catch {
    return null;
  }
}
```

If you would rather keep the experiment's blast radius small, the alternative is to refuse to dispatch when the container cannot be resolved **before** consuming the Prebid capability, so an unresolvable slot falls through to Prebid.js instead of going blank.

crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -2374,9 +2429,55 @@ mod tests {
 
         assert_eq!(registration.integration_id, APS_INTEGRATION_ID);
         assert_eq!(registration.proxies.len(), 1);
+        assert_eq!(registration.head_injectors.len(), 1);
```
prk-Jr, line 2438: 🔧 **wrench** — Nothing pins the default (`trusted_server`) mode to an *empty* head insert.

`publisher_native_config_registers_runner_mode_without_renderer_route` asserts the marker **is** emitted in native mode, but no test asserts it is **absent** by default. Drop or invert the `.then()` predicate at `aps.rs:1255` and every default deployment emits the marker, silently moving all APS rendering into the unsandboxed friendly frame. One boolean guards a security-relevant default, so it should be pinned from both sides.

I verified this is load-bearing rather than decorative: with the assertion below added, inverting that predicate to `!=` makes `enabled_config_registers_renderer_proxy` fail; without it, the inverted build passes the whole suite.

**Proposed fix** (apply manually — a one-click `suggestion` here would not survive `cargo fmt`, which rewraps the method chain across three lines; the block below is the rustfmt-clean form):

```rust
        assert_eq!(registration.head_injectors.len(), 1);
        let document_state = IntegrationDocumentState::default();
        let context = IntegrationHtmlContext {
            request_host: "publisher.example",
            request_scheme: "https",
            origin_host: "origin.example",
            document_state: &document_state,
        };
        assert!(
            registration.head_injectors[0]
                .head_inserts(&context)
                .is_empty(),
            "should not inject the native-mode marker in the default mode"
        );
```

`IntegrationDocumentState` and `IntegrationHtmlContext` are both already imported by this PR. Verified against this head: `cargo fmt --all -- --check`, `cargo clippy-fastly`, and `cargo test-fastly` (1936 passed, 0 failed) all pass with it applied.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -277,6 +312,207 @@ export function consumeApsPrebidRenderer(adId: string, expected: ApsPrebidRender
...
+        if (activeFrames.get(container) === iframe) activeFrames.delete(container);
+        iframe.remove();
+        if (warning) log.warn(warning);
+        resolve(false);
+        return;
+      }
+
+      for (const child of Array.from(container.children)) {
+        if (child !== iframe) child.remove();
+      }
```
prk-Jr, line 508: 🤔 **thinking** — Native mode leaves the Universal Creative port unanswered *and* deletes GAM's own creative iframe.

On commit this sweep removes every other child of the resolved div — on the GAM path that includes GAM's `google_ads_iframe_*`. On decline, GAM's iframe survives but the Universal Creative `render` promise never settles: no `Prebid Response` is posted (the new tests pin this with `expect(portMessages).toEqual([])`) and `stopImmediatePropagation()` already prevented Prebid.js from answering.

So in both branches, GAM-side impression counting and viewability measurement are left in an undefined state. Rollout step 5 in `docs/guide/integrations/aps.md` says to confirm impression reporting, but the doc never states the mechanism — that the friendly frame tears GAM's iframe out from under it on success, and strands the UC frame on failure. Worth naming explicitly there, since an operator validating the cohort needs to know what they are looking for.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -277,6 +312,207 @@ export function consumeApsPrebidRenderer(adId: string, expected: ApsPrebidRender
...
+      node.parentElement === frameDocument.body
+    ) {
+      node.style.display = 'block';
+    }
+  };
+  Array.from(frameDocument.body.children).forEach(normalizeFrame);
+  new frameWindow.MutationObserver((records) => {
+    for (const record of records) record.addedNodes.forEach(normalizeFrame);
+  }).observe(frameDocument.body, { childList: true });
+}
```
prk-Jr, line 456: ♻️ **refactor** — This reimplements in 22 lines of JS what the Rust renderer document does in one line of CSS.

`APS_RENDERER_DOCUMENT` gained `<style>html,body{margin:0;padding:0}body>iframe{display:block}</style>` in this same PR. The friendly frame reaches the identical result with inline styles plus a `MutationObserver` that is never disconnected. `body>iframe` selects exactly the same nodes as the observer's `parentElement === frameDocument.body` check, so the CSS is equivalent, and it applies to frames the runner adds later without any observation machinery.

**Proposed fix** (apply manually — touches three separate hunks in this file, so it can't be one contiguous `suggestion`):

```ts
frameDocument.write(
  '<!doctype html><html><head><meta charset="utf-8">' +
    '<style>html,body{margin:0;padding:0}body>iframe{display:block}</style>' +
    '</head><body></body></html>'
);
frameDocument.close();
```

then delete `prepareApsRunnerDocument` and its call site. The browser spec's `creativeDisplay: "block"` and `bodyMargin: "0px"` assertions cover the behavior either way, so the change is test-guarded.

crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -1225,6 +1246,22 @@ impl IntegrationProxy for ApsRendererIntegration {
...
+    }
+
+    fn head_inserts(&self, _ctx: &IntegrationHtmlContext<'_>) -> Vec<String> {
+        (self.rendering_mode == ApsRenderingMode::PublisherNative)
+            .then(|| {
+                "<meta name=\"trusted-server-aps-rendering-mode\" content=\"publisher_native\">"
+                    .to_string()
+            })
+            .into_iter()
+            .collect()
```
prk-Jr, line 1262: 🤔 **thinking** — `publisher_native` combined with `allow_script_creatives = true` is publisher-origin script execution with no guard and no operator signal.

The friendly frame deliberately has no sandbox, so a `script` tagType creative runs at publisher origin with full DOM, cookie, and storage access — any APS bidder that wins gets that. The docs describe the tradeoff honestly and rollout step 7 clearly intends the combination to be reachable, so I am not proposing to reject it in config validation; that would contradict the stated plan.

What is missing is any signal when both are on. A `log::warn!` at registration when `rendering_mode == ApsRenderingMode::PublisherNative && config.allow_script_creatives` would put it in operator logs rather than only in prose, which matters for a mode whose whole point is to be enabled on a small cohort and then reviewed.

docs/guide/integrations/aps.md
```diff
@@ -167,11 +191,11 @@ It deliberately omits `allow-same-origin`, so APS and bidder execution remains b
...
 
 ### GAM and Universal Creative
 
-For initial navigation and page-bids, Trusted Server publishes the same descriptor in `window.tsjs.bids`. The source-checked Prebid Universal Creative bridge accepts requests only from the iframe that owns the matching `hb_adid`, validates the complete envelope, and returns a static dynamic-renderer program that creates the same opaque renderer iframe.
+For initial navigation and page-bids, Trusted Server publishes the same descriptor in `window.tsjs.bids`. In `publisher_native` mode the ownership-checked bridge resolves the publisher div and starts the friendly-frame runner without sending a Universal Creative renderer response; in `trusted_server` mode it uses the static dynamic renderer described below. The source-checked Prebid Universal Creative bridge accepts requests only from the iframe that owns the matching `hb_adid`, validates the complete envelope, and returns a static dynamic-renderer program that creates the same opaque renderer iframe.
```
prk-Jr: ⛏ **nitpick** — The trailing sentence here is still unconditional and now contradicts the native-mode clause added immediately before it.

As written, the paragraph says native mode does *not* send a Universal Creative renderer response, and then says the bridge "returns a static dynamic-renderer program that creates the same opaque renderer iframe" with no mode qualifier. Merging the two orderings removes the contradiction:

```suggestion
For initial navigation and page-bids, Trusted Server publishes the same descriptor in `window.tsjs.bids`. The source-checked Prebid Universal Creative bridge accepts requests only from the iframe that owns the matching `hb_adid` and validates the complete envelope. In `trusted_server` mode it then returns a static dynamic-renderer program that creates the same opaque renderer iframe; in `publisher_native` mode it resolves the publisher div and starts the friendly-frame runner instead, and never returns a Universal Creative renderer response.
```

Verified with the pinned `docs` Prettier: `prettier --check .` passes with this applied.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -1697,29 +1698,49 @@ export function installTsRenderBridge(): void {
       e.stopImmediatePropagation();
       if (!messageSourceBelongsToAdUnit(e.source, prebidRendererEntry.adUnitCode)) return;
       const renderer = validateApsRenderer(prebidRendererEntry.renderer);
-      const rendererUrl = apsRendererUrl();
-      if (!renderer || !rendererUrl) return;
-      if (!hasConsumedPrebidApsIdCapacity(consumedPrebidApsIds, adId)) return;
+      if (!renderer || !hasConsumedPrebidApsIdCapacity(consumedPrebidApsIds, adId)) return;
```
prk-Jr, line 1703: 📝 **note** — The `apsRendererUrl()` check moved to *after* capability consumption.

It now lives inside the `trustedServer` callback, i.e. after `consumeApsPrebidRenderer` and `recordConsumedPrebidApsId` on the next two lines. Previously the handler returned early on a missing renderer URL without consuming anything. On an origin where `apsRendererUrl()` returns `undefined` (an opaque `window.location.origin`), the ad ID is now permanently burned in `consumedPrebidApsIds` and `markUsed` is skipped.

Practical impact is close to zero — a retry would have failed the same way, and `stopImmediatePropagation()` had already blocked Prebid.js in the old code too. Recording it because the state machine changed, not because I think it needs a fix in this PR.

prk-Jr (changes requested):
## Summary

An opt-in `publisher_native` APS rendering mode that hands the server-selected bid to APS's fixed Prebid creative runner inside a publisher-origin friendly frame, keeping the opaque static renderer as the default. The ownership discipline is careful and well tested; the blocking concerns are slot-container resolution on real GAM pages and an unpinned security-relevant default.

> 1 of the inline comments below carries a one-click GitHub `suggestion` — use **Commit suggestion** to apply it as a commit on the PR branch. The remaining comments describe the fix in prose because the change touches multiple files, spans several hunks, or would not survive `cargo fmt` as a single contiguous replacement.

## Blocking

### wrench

- **`findApsContainer` bypasses the codebase's slot-root resolution** — see inline at `crates/trusted-server-js/lib/src/integrations/aps/render.ts:54`
- **No test pins the default mode to an empty head insert** — see inline at `crates/trusted-server-core/src/integrations/aps.rs:2432`

## Non-blocking

### thinking / refactor / note

- **Native mode leaves the Universal Creative port unanswered and deletes GAM's own creative iframe** — see inline at `crates/trusted-server-js/lib/src/integrations/aps/render.ts:455`
- **`prepareApsRunnerDocument` reimplements the Rust document's one-line CSS** — see inline at `crates/trusted-server-js/lib/src/integrations/aps/render.ts:385`
- **`publisher_native` + `allow_script_creatives` has no guard or operator signal** — see inline at `crates/trusted-server-core/src/integrations/aps.rs:1255`
- **GAM doc paragraph's trailing sentence is still unconditional** — see inline at `docs/guide/integrations/aps.md:198`
- **Prebid path now consumes the capability before `apsRendererUrl()` is checked** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt/index.ts:1701`

- **Friendly-frame `document.open()/write()/close()` is proven on Chromium only.** `crates/trusted-server-integration-tests/browser/playwright.config.*` declares a single project, `chromium`. The premise of this PR is *observed* vendor behavior in a real browser, and synchronously writing into a freshly appended src-less iframe is exactly where Safari and Firefox have historically diverged on document-replacement timing — the initial `about:blank` navigation can land after the write. Worth running the new native spec against `webkit` and `firefox`, even if only locally, before the cohort test. This does not block the merge; it blocks trusting the result.

- **The one-shot ownership discipline is genuinely careful.** The dispatch `Symbol` plus the `releaseNativeDispatch` stale-completion guard, the "an invalid replacement still cancels the older pending frame" invariant, deferring `markUsed()` until the runner actually loads, and the decline-never-falls-back-to-the-opaque-renderer rule each have a matching contract test. `expect(portMessages).toEqual([])` is the right way to pin the no-Universal-Creative-response contract — it asserts the absence that the whole mode depends on, rather than only asserting the happy path.

## CI Status

- integration tests: PASS
- CodeQL: PASS
- Analyze (actions): PASS
- cargo test: PASS (required)
- Analyze (rust): PASS
- cargo fmt: PASS (required)
- vitest: PASS

crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -277,6 +312,207 @@ export function consumeApsPrebidRenderer(adId: string, expected: ApsPrebidRender
   return true;
 }
 
+/** Whether the server explicitly selected the opt-in publisher-native runner mode. */
+export function isPublisherNativeApsRendering(): boolean {
```
aram356: 🔧 **wrench** — The rendering-mode switch trusts any `<meta name="trusted-server-aps-rendering-mode" content="publisher_native">` found in `document.head`. Meta injection is exactly the injection class that survives a publisher CSP that blocks scripts, so on a *default-mode* deployment, markup injected into head flips APS rendering from the opaque sandboxed route into the unsandboxed publisher-origin friendly frame — the one place this PR's otherwise thorough validation chain reads an unauthenticated, content-forgeable signal.

The codebase already has a non-forgeable, equally CSP-inert channel: `IntegrationHeadInjector::tsjs_script_tag_attributes()` places `data-` attributes on the server-injected bundle tag, and `gpt/index.ts` already reads `data-ts-gam-attribution` through `document.currentScript` captured at module init.

**Fix**: emit the mode as a script-tag attribute and capture it at module load, instead of (or as a required corroboration of) the meta marker:

```rust
// aps.rs — on ApsRendererIntegration
fn tsjs_script_tag_attributes(&self) -> Vec<(&'static str, &'static str)> {
    (self.rendering_mode == ApsRenderingMode::PublisherNative)
        .then_some(("data-ts-aps-rendering-mode", "publisher_native"))
        .into_iter()
        .collect()
}
```

```ts
// render.ts — captured once at module init, like executingPublisherScript in gpt/index.ts
const nativeModeFromBundleTag =
  typeof document !== 'undefined' &&
  document.currentScript?.getAttribute('data-ts-aps-rendering-mode') === 'publisher_native';
```

crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -141,6 +153,9 @@ pub struct ApsConfig {
     /// Whether APS script creatives are eligible before winner selection.
     #[serde(default)]
     pub allow_script_creatives: bool,
+    /// Rendering owner for selected APS bids.
+    #[serde(default)]
+    pub rendering_mode: ApsRenderingMode,
```
aram356, line 158: 🔧 **wrench** — Nothing guards the `rendering_mode = "publisher_native"` + `allow_script_creatives = true` combination. In native mode a script-tagtype creative executes bidder-controlled markup with full publisher-origin privileges (the friendly frame has no sandbox at all). The docs mandate an ordering ("only after reviewing the friendly-frame security tradeoff, enable script creatives"), but that is a process-only control — an operator can enable both flags today and ship arbitrary third-party script into the publisher origin.

**Fix**: `ApsConfig` already has a `Validate` schema function — reject the combination there while the experiment is unvalidated:

```rust
if config.rendering_mode == ApsRenderingMode::PublisherNative && config.allow_script_creatives {
    return Err(ValidationError::new(
        "allow_script_creatives is not supported with rendering_mode = \"publisher_native\"",
    ));
}
```

If a validated cohort is expected to need the combination later, a warn-level startup log would be the minimum — but a hard rejection is safer until the friendly-frame script path has actually been validated.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -277,6 +312,207 @@ export function consumeApsPrebidRenderer(adId: string, expected: ApsPrebidRender
...
+  try {
+    rendering = renderApsPublisherNative({ slotId, renderer });
+  } catch {
+    releaseNativeDispatch(slotId, dispatch);
+    log.warn('APS native renderer: failed to start publisher-origin frame');
+    return Promise.resolve(false);
+  }
+
+  return rendering.then((accepted) => {
+    if (!releaseNativeDispatch(slotId, dispatch)) {
```
aram356, line 421: ♻️ **refactor** — This warning fires on the *normal* supersede path, not just genuinely surprising stale completions. When dispatch B replaces dispatch A, B's `cancelPendingApsRendering` settles A with `accepted = false`, and A's `.then` then always fails `releaseNativeDispatch` (B has replaced the symbol) — so every replacement logs "ignored stale completion" (the 'cancels a pending runner when a newer dispatch replaces it' test exercises exactly this). A cancelled `false` completion is expected; nothing was ignored.

**Fix**: only warn when a stale completion would otherwise have reported success:

```ts
return rendering.then((accepted) => {
  if (!releaseNativeDispatch(slotId, dispatch)) {
    if (accepted) log.warn('APS native renderer: ignored stale completion');
    return false;
  }
  return accepted;
});
```

crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -2319,7 +2372,9 @@ mod tests {
 
     #[test]
     fn registers_and_serves_only_static_renderer_route() {
```
aram356, line 2380: ♻️ **refactor** — Test gap: nothing asserts that `TrustedServer` mode emits *no* head marker. `head_inserts` returning empty in the default mode is the guard that keeps every default deployment on the sandboxed opaque route — a flipped comparison in `head_inserts` would inject the marker everywhere and silently switch all APS rendering to the unsandboxed native path client-side, while every existing test still passes.

**Fix**: one assertion in this test (or in `publisher_native_config_registers_runner_mode_without_renderer_route` with a `TrustedServer` integration):

```rust
assert!(
    integration.head_inserts(&context).is_empty(),
    "should emit no rendering-mode marker in the default mode"
);
```

crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -277,6 +312,207 @@ export function consumeApsPrebidRenderer(adId: string, expected: ApsPrebidRender
...
+    }
+  };
+  Array.from(frameDocument.body.children).forEach(normalizeFrame);
+  new frameWindow.MutationObserver((records) => {
+    for (const record of records) record.addedNodes.forEach(normalizeFrame);
+  }).observe(frameDocument.body, { childList: true });
+}
+
+/** Render the exact selected response through APS's fixed runner in a friendly iframe. */
+export function renderApsPublisherNative({
```
aram356: ⛏ **nitpick** — `renderApsPublisherNative` and `RenderApsPublisherNativeOptions` (line 380) are exported but consumed only inside this module; the tests exercise them through `dispatchApsRendering`. Making them module-private keeps the dispatch guard (`nativeDispatches` supersede bookkeeping) the only entry point, so a future caller cannot bypass it.

crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -1193,7 +1211,10 @@ impl IntegrationProxy for ApsRendererIntegration {
     }
 
     fn routes(&self) -> Vec<IntegrationEndpoint> {
```
aram356, line 1213: ⛏ **nitpick** — The renderer route is now gated on `rendering_mode` in two places: here in `routes()`, and in `register()`, which skips `.with_proxy()` entirely in native mode. Either alone suffices, and the two must stay in sync. If belt-and-braces is intended, a short comment cross-referencing the other gate would keep a future cleanup from removing the wrong one; otherwise, gating only in `register()` and leaving `routes()` unconditional is simpler.

aram356 (changes requested):
## Summary

Opt-in `publisher_native` rendering mode for selected APS bids: config enum + head marker on the Rust side, a shared rendering-owner dispatcher across the direct, server/GPT, and Prebid paths, and a friendly-frame runner with fail-closed, no-fallback semantics. The envelope validation, one-shot ownership (capability consume + tombstone before dispatch), and preserved publisher content on failure are consistent across all three paths, and coverage is thorough (unit contract tests plus a real Playwright friendly-frame proof). Requesting changes for the mode-switch trust model and the missing config guard below; the remaining findings are non-blocking.

## Blocking

### 🔧 wrench

- **Rendering-mode switch is a content-injectable DOM marker**: `isPublisherNativeApsRendering()` trusts any matching `<meta>` in `document.head`, so injected markup can flip a default-mode deployment into the unsandboxed friendly frame (crates/trusted-server-js/lib/src/integrations/aps/render.ts:316).
- **No config guard for `publisher_native` + `allow_script_creatives = true`**: nothing in code enforces the rollout ordering the docs mandate for the highest-risk combination (crates/trusted-server-core/src/integrations/aps.rs:158).

## Non-blocking

### ♻️ refactor

- **Spurious "ignored stale completion" warning on every supersede**: the normal replacement path always logs it (crates/trusted-server-js/lib/src/integrations/aps/render.ts:372).
- **Rust test gap**: no assertion that `TrustedServer` mode emits no head marker (crates/trusted-server-core/src/integrations/aps.rs:2374).

### 📝 note

- **Mode flip strands live sessions and cached HTML**: switching `trusted_server` → `publisher_native` removes `/integrations/aps/renderer` while already-served pages (no marker) still target it — those slots fail closed after the 10s ready timeout until reload. Fail-closed is the right behavior, but a sentence in the Rollout or Troubleshooting section of `docs/guide/integrations/aps.md` about in-flight sessions and HTML caches would save an operator a confusing debugging session.

### ⛏ nitpick

- **Unused exports**: `renderApsPublisherNative` / `RenderApsPublisherNativeOptions` (crates/trusted-server-js/lib/src/integrations/aps/render.ts:409).
- **Renderer route gated in two places**: `routes()` and `register()` must stay in sync (crates/trusted-server-core/src/integrations/aps.rs:1213).

## CI Status

- fmt: PASS
- clippy (all six targets): PASS
- rust tests (fastly/axum/cloudflare/spin/CLI/parity): PASS
- js tests (vitest): PASS
- CodeQL / docs + TS format: PASS

crates/trusted-server-js/lib/src/integrations/aps/render.ts
ChristianPavilonis: Implemented source-aware native-container resolution. It now handles logical GPT IDs, div IDs, `-container` wrappers, and dynamic prefixes; it uses the message source to disambiguate candidates and fails closed when no unique target exists. Tests cover each shape.

crates/trusted-server-core/src/integrations/aps.rs
ChristianPavilonis: Added a default-mode regression test. The default `TrustedServer` registration emits neither a native-mode script attribute nor a head insert, which pins the opaque renderer as the default.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
ChristianPavilonis: Kept the native runner as the sole rendering owner. On success it replaces GAM\u0027s creative iframe; on failure it preserves publisher content and does not answer Universal Creative, avoiding duplicate ownership. The docs now state this behavior and call out GAM impression and viewability validation as follow-up work.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
ChristianPavilonis: Retained DOM style assignments and the observer. The restrictive CSP fixture blocks the inline style approach in the friendly frame, while the DOM assignments work. The observer also covers runner-created child frames under that policy.

crates/trusted-server-core/src/integrations/aps.rs
ChristianPavilonis: Added a startup warning when `publisher_native` and `allow_script_creatives` are both enabled. The configuration remains allowed for the planned cohort after explicit operator review.

docs/guide/integrations/aps.md
ChristianPavilonis: Updated the paragraph to distinguish the two modes. `trusted_server` returns the opaque renderer program; `publisher_native` starts the friendly-frame runner and returns no Universal Creative response.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: No change here. The capability remains one-shot and the handler already prevents a Prebid fallback. Missing renderer URLs have no retry contract, so preserving the prior consumption ordering is not needed for this experiment.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
ChristianPavilonis: Replaced the meta marker with `data-ts-aps-rendering-mode="publisher_native"` on the server-generated TSJS script. The module captures `document.currentScript` once during initialization, so later DOM injection cannot switch the mode.

crates/trusted-server-core/src/integrations/aps.rs
ChristianPavilonis: Added a startup warning rather than a hard rejection. The experiment intentionally permits this reviewed cohort configuration, and the warning makes publisher-origin script execution visible to operators.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
ChristianPavilonis: Changed stale-completion logging to warn only when the stale attempt reports success. Normal superseded `false` completions no longer log a warning.

crates/trusted-server-core/src/integrations/aps.rs
ChristianPavilonis: Added the default-mode assertion. It verifies that the default integration emits no native-mode attribute or head marker.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
ChristianPavilonis: Made `renderApsPublisherNative` and its options type module-private. All native dispatches now go through the ownership guard.

crates/trusted-server-core/src/integrations/aps.rs
ChristianPavilonis: Retained both route gates as defense in depth. `register()` avoids attaching the proxy in native mode, while `routes()` independently ensures the renderer endpoint is absent if registration changes later.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -277,6 +368,202 @@ export function consumeApsPrebidRenderer(adId: string, expected: ApsPrebidRender
...
+ */
+export function dispatchApsRendering({
+  slotId,
+  renderer: input,
+  source,
+  trustedServer,
+}: DispatchApsRenderingOptions): boolean | Promise<boolean> {
+  // Every attempt supersedes a pending frame for this slot, including an invalid
+  // replacement that fails before a new frame can be created.
+  cancelPendingApsRendering(slotId, source);
```
prk-Jr, line 393: **HIGH — this changes the default `trusted_server` path, not just the experiment.**

`cancelPendingApsRendering()` runs unconditionally here, before `validateApsRenderer` and regardless of rendering mode. On `main`, the equivalent `pendingFrameCancels.get(container)?.()` sits *after* both the validation early-return and the `getElementById` early-return inside `renderApsCreative` — so on `main` a rejected descriptor never disturbs an in-flight frame.

**Failure scenario (default mode, no config change):** slot `S` has an APS frame in flight — created, `<iframe>` appended, waiting on the `renderer-ready` postMessage, a window of up to `RENDERER_READY_TIMEOUT_MS = 10_000`. A second dispatch arrives for `S` with a descriptor that fails validation (truncated or re-encoded `aaxResponse`, `bid.w` / `renderer.width` mismatch, non-standard base64, envelope cross-check failure). Before this PR: the invalid descriptor returned `false` and the in-flight valid frame went on to render. After: the in-flight frame is destroyed, `dispatch` returns `false`, and nothing replaces it — **permanently blank slot**. That is the failure class #418 / #958 have been chasing.

Verified with a scratch vitest pair comparing `main` semantics against this branch:

```
✓ BASELINE main behaviour: invalid second dispatch does NOT cancel the pending first frame
✗ PR behaviour: invalid second dispatch cancels the pending first frame in DEFAULT mode
  first frame still connected after invalid dispatch: false
```

No test covers this ordering in default mode — the test that does cover it (`lets an invalid replacement cancel an older pending runner`) runs with `publisherNativeRendering === true`, so the native intent is preserved by the gate below.

Second, sharper variant of the same root cause: this cancel resolves its container through the fuzzy `findApsContainer()`, while default-mode rendering resolves through the exact `document.getElementById()` — so the two can target **different elements**. Scratch-reproduced cross-slot destruction (dispatch for logical slot `homepage_header` with `divToSlotId {'div-header': 'homepage_header'}` returned `false` with `slot not found`, yet left `#div-header iframes now: 0` — it killed a frame in a slot it never rendered into). Latent rather than routinely reachable, but the same-`slotId` case above is directly reachable.

Fix applied and verified in a scratch worktree: full suite `863 passed`, prettier clean, both scratch cases green (`first frame still connected after invalid dispatch: true`), native-mode supersede test still passing.

```suggestion
  // Every native attempt supersedes a pending frame for this slot, including an
  // invalid replacement that fails before a new frame can be created. Default mode
  // keeps renderApsCreative's semantics, where a rejected descriptor never disturbs
  // an in-flight frame.
  if (publisherNativeRendering) cancelPendingApsRendering(slotId, source);
```

crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -277,6 +368,202 @@ export function consumeApsPrebidRenderer(adId: string, expected: ApsPrebidRender
...
+      const frameDocument = iframe.contentDocument;
+      if (!frameWindow || !frameDocument) {
+        fail();
+        return;
+      }
+
+      frameDocument.open();
+      frameDocument.write(
+        '<!doctype html><html><head><meta charset="utf-8"></head><body></body></html>'
+      );
```
prk-Jr, line 543: **MEDIUM — the friendly frame silently drops `no-referrer`, and inherits the publisher's CSP.**

`trusted_server` mode serves the renderer document with three protections (see `aps.rs:1236-1239`): `referrer-policy: no-referrer`, `x-content-type-options: nosniff`, and `APS_RENDERER_CSP = "default-src 'none'; …"` plus a CSP-level `sandbox`. This `document.write`n `about:blank` frame inherits the **publisher document's** referrer policy and CSP instead. The docs disclose the sandbox loss ("this friendly frame deliberately has no opaque-origin sandbox") but say nothing about referrer or CSP.

**Failure scenario:** publisher page runs the common default (`strict-origin-when-cross-origin`) or something looser (`unsafe-url`, `no-referrer-when-downgrade`). Every creative subresource, tracking pixel and click beacon the APS runner loads now carries the publisher origin — or under `unsafe-url` the full page URL — as `Referer`, and `Origin: https://publisher.example` instead of the sandbox's `Origin: null`. On a publisher with no CSP at all, the creative runs with no `connect-src` / `img-src` / `frame-src` restriction whatsoever, where `trusted_server` mode pinned it to `default-src 'none'` plus `https:` allowlists. This PR's own browser fixture demonstrates the inheritance — it had to add `frame-src https://creative.example` to the *page* CSP for the creative to load (`aps-renderer.spec.ts:~1029`).

The referrer half restores in one line (verified: `863 passed`, prettier clean):

```suggestion
      frameDocument.write(
        '<!doctype html><html><head><meta charset="utf-8">' +
          '<meta name="referrer" content="no-referrer"></head><body></body></html>'
      );
```

The CSP half can't be a one-liner — it needs a decision on how much of `APS_RENDERER_CSP` the runner tolerates without `sandbox`. Either add a `<meta http-equiv="Content-Security-Policy">` mirroring `APS_RENDERER_CSP` minus the `sandbox` directive, or state the CSP-inheritance delta explicitly in the security paragraph alongside the sandbox one.

docs/guide/integrations/aps.md
```diff
@@ -235,12 +264,12 @@ Use fictional values in source-controlled configuration and fixtures. Supply con
 
 ### Winner targets but does not render
 
-- Confirm `GET /integrations/aps/renderer` returns HTML with its CSP and `Referrer-Policy: no-referrer`.
-- Confirm publisher CSP permits `frame-src 'self'`.
+- In `trusted_server` mode, confirm `GET /integrations/aps/renderer` returns HTML with its CSP and `Referrer-Policy: no-referrer`, and that publisher CSP permits `frame-src 'self'`.
+- In `publisher_native` mode, confirm the mode `<meta>` marker is present, the slot receives a hidden friendly iframe, and `https://client.aps.amazon-adsystem.com/prebid-creative.js` is not blocked by CSP. The static renderer route is intentionally absent in this mode.
```
prk-Jr, line 268: **MEDIUM — this troubleshooting step points at a marker that was deliberately removed.**

`head_inserts` returns `Vec::new()`, and the test at `aps.rs:2483` asserts `"should not inject a forgeable native-mode marker"`. There is no `<meta>` marker to find. `4b8e04e4` did emit `<meta name="trusted-server-aps-rendering-mode" content="publisher_native">`; `69b30c0d Harden APS publisher-native rendering` replaced it with the script-tag attribute and updated the main paragraph (~line 54, now correct) but left this bullet behind.

**Consequence:** an operator debugging a non-rendering native cohort follows this step, finds no `<meta>`, and concludes the mode never activated — when the real signal is `data-ts-aps-rendering-mode="publisher_native"` on the `#trustedserver-js` script tag.

Verified `docs` prettier clean with this wording:

```suggestion
- In `publisher_native` mode, confirm the TSJS bundle tag carries `data-ts-aps-rendering-mode="publisher_native"`, the slot receives a hidden friendly iframe, and `https://client.aps.amazon-adsystem.com/prebid-creative.js` is not blocked by CSP. The static renderer route is intentionally absent in this mode.
```

crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -38,6 +43,92 @@ type ValidatedRendererCacheEntry = {
   renderer: ApsRendererV1;
 };
 const validatedRendererCache = new WeakMap<object, ValidatedRendererCacheEntry>();
+const nativeDispatches = new Map<string, symbol>();
+const publisherNativeRendering =
+  typeof document !== 'undefined' &&
+  document.currentScript?.getAttribute(APS_RENDERING_MODE_ATTRIBUTE_NAME) === 'publisher_native';
```
prk-Jr, line 49: **MEDIUM (design question) — "one owner per slot" is enforced per-bundle, not per-page.**

`nativeDispatches` and `publisherNativeRendering` here, and `activeFrames` / `pendingFrameCancels` at lines 31-32, are all module-scope. There is no `src/integrations/aps/index.ts`, so `render.ts` is inlined into every importing entry point as a separate esbuild IIFE. The docs' claim — "This one-owner behavior avoids a second render path" — holds inside a bundle, not across the page.

Confirmed by building and grepping the emitted bundles (`node build-all.mjs`):

```
tsjs-core.js   (19342 B): mode-attr 1  native-renderer-str 5  runner-url 1
tsjs-gpt.js    (37989 B): mode-attr 1  native-renderer-str 5  runner-url 1
tsjs-prebid.js (22545 B): mode-attr 1  native-renderer-str 0  runner-url 0
```

Each dist file is its own `(function(){"use strict"; … })();`, so the scopes do not merge when the Rust server concatenates them at request time.

**Mechanism:** on a native-mode page where both `core`'s `requestAds()` and the `gpt` bridge target slot `S`, `cancelPendingApsRendering` in the `gpt` copy cannot see the `core` copy's pending cancel. Two friendly frames get appended to the same container and **both load `prebid-creative.js`, so both can fire impression/billing beacons** before the winner's `finish(true)` strips siblings. The DOM self-heals (the loser sees `!iframe.isConnected` and resolves `false`), but the duplicated vendor runner load is not undone.

**What I could not prove:** that `core`'s `requestAds` and the `gpt` bridge both activate for the same bid on the same page load — they read different bid sources (the live `/auction` POST vs. the frozen page-load `window.tsjs.bids`). So the mechanism is real but the trigger is unproven. Flagging as a design question, not a confirmed live bug.

**Fix shape** (not one-click — needs a new file plus import rewrites in three modules): promote the frame registries to a single page-scoped owner, e.g. lazily hung off `window.tsjs` (`tsjs.__apsFrames ??= new WeakMap()`) the way `apsPrebidRenderers` already is, or add `src/integrations/aps/index.ts` and register APS as a real JS module so exactly one copy ships.

**Related, LOW:** the `document.currentScript` gate on lines 47-49 survives tree-shaking into `tsjs-prebid.js` (grep count 1) even though the dispatcher and `renderApsPublisherNative` do not (count 0). `prebid` is a **deferred** module and `tsjs_deferred_script_tags` (`html_processor.rs:373`) does not carry the attribute; the external R2 `trusted-prebid.js` is loaded by a tag TS doesn't control at all. So that copy's `publisherNativeRendering` is permanently `false`. Harmless today, but if anyone later calls the dispatcher from the prebid module, native mode is silently off there — and since `trusted-prebid.js` ships to R2 outside the deployer, the breakage would be invisible. Worth a comment on the constant, or resolving the mode from a shared `window.tsjs` field set by whichever bundle reads the tag first.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -277,6 +368,202 @@ export function consumeApsPrebidRenderer(adId: string, expected: ApsPrebidRender
...
+          new frameWindow.CustomEvent('prebid/creative/render', {
+            detail: { aaxResponse: renderer.aaxResponse, seatBidId: renderer.bidId },
+          }),
+        ],
+        store: new Map([['listeners', new Map()]]),
+      });
+
+      runner = frameDocument.createElement('script');
+      runner.src = APS_PREBID_CREATIVE_RUNNER_URL;
+      runner.addEventListener('load', commit, { once: true });
```
prk-Jr, line 558: **MEDIUM (design) — this commits on runner *script load*, not on creative render, and the commit destroys the GAM iframe.**

`runner.addEventListener('load', commit)` resolves the dispatch `true` the moment `prebid-creative.js` finishes downloading. `finish(true)` then removes every other child of the container and reveals the frame.

**Failure scenario:** the runner loads (200 OK, `load` fires) but cannot render — malformed `aaxResponse` that got past TSJS's checks, a creative subresource blocked by the publisher CSP, an APS-side error. Result: a visible empty 300×250 frame, the publisher's original content **and the GAM / Universal Creative iframe** removed from that div, no UC response ever posted (the capability was already consumed), and no fallback by design. Net: blank paid slot *plus* a destroyed GAM impression.

Both halves are explicit in the PR's own docs — "Trusted Server treats runner load as successful handoff" and "This removes the GAM or Universal Creative iframe when it is inside that div" — so this is a deliberate choice, not an oversight. But the combination (script-load as the commit signal + sibling destruction + no fallback) is the highest-blast-radius decision in the PR, and worth settling before any cohort rollout.

**Fix shape** (not one-click — restructures the promise plumbing): `prepareApsRunnerDocument` already installs a `MutationObserver` on the runner frame's body (line ~453). Reuse it as the commit signal — commit on runner `load` **and** the first node appended to the frame body, with the existing 10s timeout as the backstop — so a runner that loads but produces nothing declines the bid and leaves the GAM iframe intact.

crates/trusted-server-core/src/integrations/aps.rs
```diff
@@ -1262,6 +1305,11 @@ pub fn register_providers(
             "APS debug mode is ON — raw request and response data, including creative markup, will be included in client-visible /auction responses"
         );
     }
+    if config.rendering_mode == ApsRenderingMode::PublisherNative && config.allow_script_creatives {
+        log::warn!(
+            "APS publisher-native rendering with script creatives is ON; selected bidder scripts execute with publisher-origin privileges"
+        );
+    }
```
prk-Jr, line 1312: **MEDIUM (question) — should `publisher_native` + `allow_script_creatives` be a hard validation failure rather than a warning?**

In native mode the frame is same-origin with the publisher, so a `tagType: "script"` creative injected by the vendor runner executes with full publisher-origin privileges: publisher DOM, `document.cookie` (including the Trusted Server EC ID cookie), and same-origin `fetch` to publisher endpoints. The only guard on that combination is this log line, and the PR explicitly locks the combination in as supported (`publisher_native_script_creatives_remain_available_for_controlled_validation`, lines 2508-2530; line 791 is the only place `allow_script_creatives` gates anything).

`ApsConfig` already has a `#[validate(schema(...))]` hook (used by `validate_inventory_identity_override`). Would it be better to make this combination a hard config-validation failure behind a third explicit opt-in, so it cannot be reached by config drift or a stale cached config blob? A warning in edge logs is easy to miss for a blast radius this large.

No suggestion block — this contradicts one of the PR's own tests, so it needs an answer rather than a patch.

crates/trusted-server-integration-tests/browser/tests/shared/aps-renderer.spec.ts
```diff
@@ -4,6 +4,8 @@ import { expect, test, type Page } from "@playwright/test";
 import { runtimeUrl } from "../../helpers/state.js";
 
 const RUNNER_URL = "https://client.aps.amazon-adsystem.com/prebid-creative.js";
+const PUBLISHER_CORE_URL =
+    "https://client.aps.amazon-adsystem.com/trusted-server-core.js";
```
prk-Jr, line 8: **LOW — this serves Trusted Server's own bundle from Amazon's real CDN hostname.**

`https://client.aps.amazon-adsystem.com/trusted-server-core.js` is a fabricated path on a real third-party vendor domain, used only to satisfy the fixture's `script-src https://client.aps.amazon-adsystem.com`. It reads as though Trusted Server ships a bundle from Amazon's CDN.

Cheaper and clearer to use a fictional host:

```ts
const PUBLISHER_CORE_URL = "https://tsjs.example/trusted-server-core.js";
```

Not offered as a one-click suggestion because it also needs `https://tsjs.example` added to the `script-src` list in this test's CSP header (~line 1029) — a one-hunk suggestion would break the fixture.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -1697,29 +1698,50 @@ export function installTsRenderBridge(): void {
...
+          }
+        },
+      });
+      if (typeof dispatched === 'boolean') {
+        if (dispatched) markUsed();
+      } else {
+        void dispatched.then((accepted) => {
+          if (accepted) markUsed();
+        });
       }
```
prk-Jr, line 1747: **LOW — `dispatchApsRendering` returning `boolean | Promise<boolean>` forces every caller to branch on `typeof`.**

This call site `typeof`-switches; the other (`core/request.ts:55-61`) wraps in `Promise.resolve` and discards. Having the dispatcher always return `Promise<boolean>` (`return Promise.resolve(trustedServer(renderer))` on the default path) would reduce both call sites to a plain `.then`, and would remove the only reason this block needs the duplicated `markUsed()`.

Behaviour-neutral; safe to defer.

prk-Jr (changes requested):
Staff-level review pass. Reviewed in a clean detached worktree at head `13a6ecf3`, with every claim below scratch-verified before posting.

## What this PR does, mechanically

Adds a second, opt-in rendering owner for server-selected APS bids. Rust side: a strict `ApsRenderingMode` enum (`trusted_server` default / `publisher_native`) on `ApsConfig`; in native mode `register()` **drops** the `/integrations/aps/renderer` proxy route entirely and instead registers `ApsRendererIntegration` as an `IntegrationHeadInjector` whose only job is to stamp `data-ts-aps-rendering-mode="publisher_native"` onto the server-generated TSJS bundle `<script>` tag (`head_inserts` deliberately returns empty — the earlier forgeable `<meta>` marker was removed in `69b30c0d`). JS side: `render.ts` gains a `dispatchApsRendering()` fan-out that reads that attribute once at module-eval time via `document.currentScript`, then routes each validated descriptor to exactly one owner — the existing opaque-sandbox `renderApsCreative`, or the new `renderApsPublisherNative()` which builds an unsandboxed same-origin friendly iframe, seeds the frame's account-keyed `_aps` queue with a `prebid/creative/render` CustomEvent, and loads Amazon's `prebid-creative.js`. All three call sites are rewired through the dispatcher; in native mode the GPT paths send no Universal Creative response and never fall back.

## Verification run locally

| Check | Result |
|---|---|
| `npx vitest run` | **PASS** — `44 files`, `863 tests`, no type errors |
| `npm run format` (JS) / (docs) | **PASS** / **PASS** |
| `cargo fmt --all -- --check` | **PASS** |
| `clippy-fastly` / `-axum` / `-cloudflare` / `-cloudflare-wasm` / `-spin-native` / `-spin-wasm` | **PASS** ×6 |
| `cargo test -p trusted-server-core --target aarch64-apple-darwin` | **PASS** — `1982 passed`, + 2 doc-tests |
| `node build-all.mjs` | **PASS** — 13 modules |
| GitHub CI (`gh pr checks`) | **19/19 pass**, incl. cross-adapter parity, browser integration, CodeQL |

`cargo test-fastly` (Viceroy) was not run — all Rust changes are in `trusted-server-core`, which was exercised natively instead, and GitHub's `cargo test` job is green.

## Verdict

The experiment itself is well-built: descriptor validation, capability-consumption ordering, supersede semantics and fail-closed paths are unusually careful, the security tradeoff is documented honestly rather than buried, and native-mode coverage is strong (unit + jsdom + a real Playwright fixture with a publisher CSP). Rust-side changes are minimal and parity-safe.

Requesting changes on **one** issue: an experiment behind a non-default config flag must not change the default path, and this one does. `dispatchApsRendering` moves the pending-frame cancel ahead of validation for **all** operators, including everyone still on `trusted_server` — I proved by scratch test that a rejected descriptor now destroys an in-flight valid render that `main` would have completed. Blank-slot regression on the shipping path, no test covers it, one-line fix that keeps all 863 tests green. The two other suggestion-block findings (inherited referrer policy, stale docs bullet) are mechanical and both verified — worth riding along.

The three design findings (per-bundle rather than per-page slot ownership; commit-on-script-load destroying the GAM iframe; `publisher_native` + `allow_script_creatives` guarded only by a log line) need not block the merge, but all three should be answered before any traffic is pointed at `publisher_native` — particularly commit-on-script-load, which is the decision most likely to cost impressions in a live cohort.

## Checked and found clean

- **Cross-adapter parity.** No `stream_response`, no `fastly::*`, no platform-gated API. `tsjs_script_tag_attributes` is an existing adapter-agnostic core mechanism (`registry.rs:1063`, consumed at `html_processor.rs:359`). All six clippy targets and the parity job green.
- **The third-render-context URL-rewrite hazard does not apply.** Native mode touches neither `rewrite_creative_html` nor `rewrite_inline_creative_html` — no `adm` HTML is injected. The runner receives the base64 `aaxResponse` envelope and loads `renderer.creativeUrl`, which `validCreativeUrl` already requires to be absolute HTTPS, credential-free, and non-publisher-origin. No relative-vs-absolute trap.
- **`hb_adid` non-uniqueness.** The GPT server path still resolves via `slotIdForMessageSource(e.source)` and rejects on `matchedBid.hb_adid !== adId`; requesting-slot-first ordering and its comment untouched. The Prebid path still gates on `messageSourceBelongsToAdUnit`. The new `source` plumb-through into `findApsContainer` reuses the existing `contentWindow === source` idiom.
- **`sourceMatchedCandidates` fallback-to-all-candidates.** Tried to turn this into a wrong-slot-render finding and could not substantiate it — both native call sites validate the source upstream, and `uniqueSlotCandidate` requires exactly one match, so ambiguity falls through to the next strategy rather than guessing. Dropped.
- **No publisher-side changes required.** The attribute lands on TS's own generated `<script id="trustedserver-js">`.
- **Mode-attribute forgeability.** Captured once at IIFE eval from `document.currentScript`, and the immediate bundle tag is neither `async` nor `defer` (`tsjs.rs:43-46`), so later-injected markup cannot flip the mode — covered by test.
- **No real operator values** in docs, config or tests (`example-account`, `publisher.example`, `creative.example`, fictional throughout). One nit noted inline.
- **Serde contract.** `snake_case`, `#[default] TrustedServer`, `#[serde(default)]`, unknown values reject — covered by test.
- **Project conventions.** No `unwrap()` in prod code, no `anyhow`, no thiserror, no `println!`, no wildcard imports, comments above code, tests colocated. JS tests use `vi.spyOn`/`vi.resetModules` (no `vi.mock()` factory, so `vi.hoisted()` doesn't apply).
- **TDZ / cleanup / leak audit of `renderApsPublisherNative`.** `cleanup()` only ever runs after the `timeoutId` assignment, `settled` guards re-entry, the timeout is the guaranteed terminal path so `nativeDispatches` entries always release, and `activeFrames.set` correctly follows the previous frame's cancel. The un-`disconnect()`ed `MutationObserver` dies with the frame's browsing context on `iframe.remove()` — not a leak. Dropped.
- **Cache/rollback semantics of a mode change.** The route-disappears-under-cached-HTML hazard is documented in both the Migration and Rollout sections — good catch by the author.

## Pre-existing, not this PR

The APS renderer branch never calls `safelyRecordCreativeRequest` in **either** mode, so APS renders are invisible to `gptDiagnosticsRecorder` — `main` returns early before that point the same way. Native-mode renders inherit the blind spot rather than creating it. Worth a follow-up issue if the cohort needs render-rate telemetry, since blank-slot triage here has historically leaned on `tsjs.renders` vs `tsjs.bids`.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
```diff
@@ -38,6 +43,92 @@ type ValidatedRendererCacheEntry = {
   renderer: ApsRendererV1;
 };
 const validatedRendererCache = new WeakMap<object, ValidatedRendererCacheEntry>();
+const nativeDispatches = new Map<string, symbol>();
+const publisherNativeRendering =
```
aram356, line 47: 🌱 **seedling** — The deferred `tsjs-prebid.js` bundle contains its own copy of this module (it imports `registerApsPrebidRenderer`), and deferred tags generated by `tsjs_deferred_script_tags()` do not carry `tsjs_script_tag_attributes()` — so that copy's `publisherNativeRendering` is always `false`. Safe today because dispatch only ever runs from the immediate bundle (core/gpt share the attributed tag), but the invariant is implicit: a future `dispatchApsRendering` call from a deferred module would silently render in default mode even on a native-mode deployment.

A short comment here naming the invariant — or applying the attributes to deferred tags server-side — would pin it down before someone trips over it.

crates/trusted-server-core/src/integrations/aps.rs
aram356, line 1213: ⛏ **nitpick** — Carry-over from the previous review: the renderer route is still gated on `rendering_mode` both here in `routes()` and in `register()` (which skips `.with_proxy()` in native mode). Either alone suffices, and the two must stay in sync. If belt-and-braces is intended, a short comment cross-referencing the other gate would keep a future cleanup from removing the wrong one.

aram356 (approved):
## Summary

Re-review at head `07dfc189f`. Every finding from the previous review is resolved: the rendering-mode signal moved from the injectable head `<meta>` marker to the server-owned `data-ts-aps-rendering-mode` bundle-tag attribute captured once at module init, with a regression test proving post-init injected meta tags and spoofed script tags are ignored, and the Playwright proof now runs under a CSP without `'unsafe-inline'`. The script-creatives combination is settled with a warn-level startup log plus a test documenting the controlled-validation intent. The stale-completion warning, default-mode marker assertions, mode-flip/cached-HTML docs, and module-private renderer internals are all in place. The `findApsContainer` rework (requesting-frame disambiguation, `-container` inner resolution, unique-candidate dynamic prefixes) is conservative where candidates are ambiguous and is covered by new unit tests on both the render and bridge sides.

The findings below are non-blocking.

## Non-blocking

### ❓ question

- **Foreign-author "probe" commit on the branch** (`d97bda69f`, authored by `prk-Jr`): a 1.1MB commit duplicating main's content (ESI designs, template cache) under the message "probe", later netted out by the `07dfc189f` main merge. Verified it leaks nothing into the PR — the diff vs main is exactly the 10 expected APS files (+1324/−59) — and squash-merge keeps it out of main's history. Was this push intentional? Dropping it with a rebase would keep the branch history honest; otherwise a quick confirmation here is enough.

### 🌱 seedling

- **Deferred-bundle copy of `render.ts` cannot see the mode attribute**: the deferred `tsjs-prebid.js` bundles its own copy of this module, and deferred tags from `tsjs_deferred_script_tags()` do not carry `tsjs_script_tag_attributes()` (crates/trusted-server-js/lib/src/integrations/aps/render.ts:47).

### ⛏ nitpick

- **Renderer route gated in two places**: `routes()` and `register()` both gate on the mode and must stay in sync (crates/trusted-server-core/src/integrations/aps.rs:1213).

## CI Status

- fmt: PASS
- clippy (all six targets): PASS
- rust tests (fastly/axum/cloudflare/spin/CLI/parity): PASS
- js tests (vitest): PASS
- integration + browser integration tests (including the APS publisher-native Playwright proof): PASS
- CodeQL / docs + TS format: PASS

crates/trusted-server-js/lib/src/integrations/aps/render.ts
ChristianPavilonis: Fixed in `28a1b7b7e`. Early pending-frame cancellation now runs only in `publisher_native`. Default mode validates first, so a rejected replacement leaves a valid in-flight opaque frame alone. I added a regression test that completes the original frame after an invalid replacement; the existing native invalid-supersede behavior remains covered.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
ChristianPavilonis: Accepted both parts in `28a1b7b7e`. The friendly-frame document now sets `<meta name="referrer" content="no-referrer">` before loading the runner. The browser test asserts that neither the runner request nor the creative-frame request carries `Referer`. The guide now states that the initial friendly document inherits publisher CSP. I did not add another CSP because the publisher policy intentionally controls resource loading for this experiment.

docs/guide/integrations/aps.md
ChristianPavilonis: Fixed in `28a1b7b7e`. Troubleshooting now points operators to `data-ts-aps-rendering-mode="publisher_native"` on `#trustedserver-js` and no longer mentions the removed `<meta>` marker.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
ChristianPavilonis: Confirmed that the core and GPT IIFEs have separate owner state. I am leaving this as a follow-up because we do not have a demonstrated supported flow where direct `requestAds()` and the GPT bridge render the same slot, and page-global ownership would require defining winner semantics across those bid sources. Native rollout must not target the same slot through both paths concurrently. If mixed delivery becomes supported, it should first add a cross-bundle reproduction and then move ownership to page-scoped state.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
ChristianPavilonis: No code change here. Runner load remains the documented handoff signal for this experiment. A first DOM mutation would not prove that the creative rendered and could commit on a runner placeholder or script node instead. The guide now repeats that runner load is not proof of paint. APS impression, viewability, and completion semantics remain required validation before cohort rollout.

crates/trusted-server-core/src/integrations/aps.rs
ChristianPavilonis: Keeping the warning rather than making this a hard validation failure. Both settings are explicit, the combination has a startup warning, and the staged experiment intentionally keeps script creatives available only after the iframe-only cohort passes security review. Deployment guidance remains to start with `allow_script_creatives = false` and enable it only for controlled validation.

crates/trusted-server-integration-tests/browser/tests/shared/aps-renderer.spec.ts
ChristianPavilonis: Fixed in `28a1b7b7e`. The mocked Trusted Server bundle now uses `https://tsjs.example/trusted-server-core.js`; only the actual APS runner keeps the Amazon hostname. The CSP fixture was updated accordingly.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: No change here. The union reflects the current contract: default rendering returns synchronously, while native rendering completes asynchronously. Always returning a promise would change the timing of the default-path `markUsed()` call, so I am leaving that API migration out of this experiment.

crates/trusted-server-js/lib/src/integrations/aps/render.ts
ChristianPavilonis: Confirmed and leaving this as a future guardrail. The deferred Prebid bundle only validates and registers APS capabilities today; the attributed immediate GPT bundle performs dispatch. Any future deferred-bundle dispatch must first share the captured mode and owner state rather than relying on this module-local constant.

crates/trusted-server-core/src/integrations/aps.rs
ChristianPavilonis: The two gates are intentional defense in depth. `register()` avoids attaching the proxy in native mode, while `routes()` independently guarantees that the endpoint remains absent if registration changes later. The native-mode test asserts both conditions, so I am retaining them without a production-code cross-reference comment.

prk-Jr (approved):

PR #1043, opened 2026-08-19
Add a pluggable Edge Cookie provider seam with the built-in HMAC provider

jwrosewell:
> **This description has been edited.** The Edge Cookie selector now follows
> the one configuration convention the whole stack lands, so the table names
> and a few of the rules have changed. If you read an earlier version of this
> text, re-read "How a provider is selected". Everything else about the seam,
> the identifier bounds, the provider codes and the identity graph is the same
> work described the same way.
>
> What changed in the pull request since the last version of this description:
>
> 1. A provider's settings sit in `[ec.<name>]` rather than
>    `[ec.providers.<name>]`. The word `providers` is gone from the file, here
>    and in every other provider type.
> 2. A table's name is the implementation it configures, unless the table
>    carries an `implementation` line. That is new, and it lets a deployment
>    run an implementation under a name of its own.
> 3. Provider names and implementation ids must be snake_case, and a name
>    outside that is refused at startup.
> 4. The deprecated `[ec] passphrase` form migrates to `[ec.hmac] passphrase`
>    rather than to `[ec.providers.hmac] passphrase`.
>
> The convention itself, what it is checked against and how to move an existing
> file are written up in
> [`docs/guide/configuration-rules.md`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/docs/guide/configuration-rules.md),
> which lands with the integration seam (#1094).

This pull request is the first of six stacked pull requests and the only one
with no parent, so it depends on nothing beyond `main` and can merge on its
own. The stack is #1043, #1044, #1045, #1046, #1047 and #1094, each targeting
`main` and each branch building on the one before. The first five decompose #838
as requested in the #986 review, and #1094 opens the integration seam on top of
them. A later pull request's own change shows when its head branch is
compared with the previous pull request's head branch.

The design specs for the series are carried by the spec pull request (#1084).
The spec for this pull request is
[`2026-07-30-pluggable-providers-design.md`](https://github.com/jwrosewell/trusted-server/blob/split/6-integration-seam/docs/superpowers/specs/2026-07-30-pluggable-providers-design.md),
which covers both this pull request (the Edge Cookie seam) and the device and
location seam (#1044). The spec is revised to match this implementation, with a
revision record listing every divergence from the earlier draft and why.

## What this pull request does

Edge Cookie identity generation becomes a selectable provider behind the
`EdgeCookieProvider` trait in
[...]

PR #1044, opened 2026-08-19
Add device and geo provider selection with the host-signal Edge Cookie provider

jwrosewell:
> **This description has been edited.** The device and location selectors now
> follow the one configuration convention the whole stack lands, so a few
> table and provider names have changed. If you read an earlier version of
> this text, the changed names are listed below and nothing else about the two
> selectors or the host-signal provider has moved.
>
>
> 1. The host-signal Edge Cookie provider is `host_signals`, not
>    `host-signals`. Every name a person types is snake_case now.
> 2. Its settings sit in `[ec.host_signals]`, not
>    `[ec.providers.host-signals]`, and the built-in HMAC provider's sit in
>    `[ec.hmac]`. The word `providers` is gone from the file.
> 3. `[device]` and `[geo]` follow the same convention as `[ec]`, so a
>    provider's settings, where it has any, go in `[device.<name>]` and
>    `[geo.<name>]`, and a table may carry an `implementation` line to run an
>    implementation under a name of its own.
>
> file are written up in

Stacks on the Edge Cookie provider seam (#1043), and depends on its
`EdgeCookieProvider` trait and provider-code registry, which the host-signal
provider below uses. The stack has six pull requests
second, and the first five decompose #838 as requested in the #986 review.
[Compare `split/1-ec-provider` with `split/2-device-geo`](https://github.com/jwrosewell/trusted-server/compare/split/1-ec-provider...split/2-device-geo)

The spec for this pull request, shared with #1043, is
[`2026-07-30-pluggable-providers-design.md`](https://github.com/jwrosewell/trusted-server/blob/split/6-integration-seam/docs/superpowers/specs/2026-07-30-pluggable-providers-design.md),
which covers the Edge Cookie, device and location providers together.

Device classification and geolocation become selectable providers, mirroring
the Edge Cookie seam and using the same configuration shape.

- `[device] provider` selects the classifier. The default `builtin` reads the
  User-Agent alone and makes no host call. The opt-in `fastly` provider
  strengthens the browser and bot gate with the host's TLS JA4 and HTTP/2
  signals
  ([`crates/device/fastly`](https://github.com/jwrosewell/trusted-server/tree/split/2-device-geo/crates/device/fastly)).
- `[geo] provider` selects geolocation. In this pull request the host
  platform's lookup remains the default, matching the behavior before the
  selector existed, with `provider = "none"` as an explicit opt-out that sends
  no client IP to any host geo service. The switch to no geolocation by default
  lands with the permission model (#1045), together with the permission
  baseline that makes a deployment without geolocation workable, so this pull
  request alone changes no deployment's geo behavior.
- Every adapter (Fastly, Axum, Cloudflare, Spin) routes its host geo through
  the same `build_geo_provider` selector, so `[geo] provider` behaves the same
  everywhere rather than on Fastly alone.
- The provider tables (`[device]`, `[geo]`, `[ec.hmac]`, `[ec.host_signals]`)
  reject keys they do not know at startup, so a mistyped setting fails loudly
  instead of silently leaving a default in place. A table the selector does not
  name is refused the same way, and provider names must be snake_case.
- The host-signal Edge Cookie provider ships as an opt-in built-in. It creates
  the identifier from the host TLS and HTTP/2 signals plus the client IP, and
  with no host signals at all it defers with a warning rather than falling back
  to an identifier made from the IP alone. Its identifiers carry the `hs00~`
  code from the provider-code registry that #1043 introduces, which fixes the
  collision defect the review of #838 found, where host-signal identifiers
  shared the HMAC format and key space. Its `[ec.host_signals] passphrase`
  names a secret-store key, as the HMAC passphrase does. Push validation checks
  the name as a key reference, and the resolved value is held to the same
  32-byte minimum. Whether host TLS and HTTP/2 processing ships in the series
  is a policy question put to the task force in #1071, which is the separate
  design that sign-off row 22 asks for. The proposal there is to close the row
  with the capability opt-in, both uses permission-gated, and the policy
  expressed in `permissions.yaml` rather than compiled into the build.

## How it was verified

Every job `main`'s CI runs was run locally against `37eadaed4`, on Windows and
Clippy with warnings denied on the Axum, Cloudflare, Spin and Fastly adapters
and on both wasm targets, the Axum, Cloudflare and Spin adapter suites, the
cross-adapter parity suite, the benchmark smoke run, the release wasm builds
for Spin and Fastly, and the CLI, OpenRTB codegen, `format-docs` and template
cache harness jobs that only run on Linux. The core suite passes with 2,763
tests natively and 2,757 under Viceroy.

CI on that head was green across all 19 checks, being
[Run Tests](https://github.com/IABTechLab/trusted-server/actions/runs/34890494586),
[Run Format](https://github.com/IABTechLab/trusted-server/actions/runs/34890494507),
[Integration Tests](https://github.com/IABTechLab/trusted-server/actions/runs/34890494643)
and
[CodeQL Advanced](https://github.com/IABTechLab/trusted-server/actions/runs/34890494713).

On the head this pull request shows now, `558bde54c`, every CI check passes: [cargo test](https://github.com/IABTechLab/trusted-server/actions/runs/35028991454/job/104582774995), [the CLI tests](https://github.com/IABTechLab/trusted-server/actions/runs/35028991454/job/104582774941), [format-docs](https://github.com/IABTechLab/trusted-server/actions/runs/35028991468/job/104582774558), [the integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/35028991443/job/104585395474), [vitest](https://github.com/IABTechLab/trusted-server/actions/runs/35028991454/job/104582774616) and [CodeQL](https://github.com/IABTechLab/trusted-server/runs/104582941157).

References #780 and #781. Decomposes #838. Spec baseline from #986.


aram356:

Issue #778, opened 2026-06-18
Make Edge Cookie identity pluggable and opt-in


jwrosewell:
**User story**

As a deployer, I want to choose how the Edge Cookie identity is derived, or run with none, so that identity is a deliberate configuration choice rather than a fixed behavior.

**Description**

The Edge Cookie identity is currently hard-wired as an HMAC-SHA256 value over the client IP. Introduce an `EdgeCookieProvider` interface selected by configuration, with a built-in HMAC provider that preserves today's behavior. There is no default selection, so with no provider chosen Trusted Server runs statelessly. Trusted Server stays neutral on policy here, and the deployer decides whether to derive an identity and how.

**Done when**

- An `EdgeCookieProvider` interface is selected by `[ec] provider`.
- A built-in HMAC provider reproduces the current identifier.
- With no provider selected, no Edge Cookie is created and the request proceeds.
- Identifier comparison is a provider operation, so a provider whose identifiers wrap the same payload in different envelopes can compare by payload.
- A provider can return response headers (for example to request more client evidence on a later request).

**References**

- Part of #777. Implemented by the neutral-providers PR (stacked on the EdgeZero series, #725).
- Related: #712 (Edge Cookie key system, Phase 2).

PR #1045, opened 2026-08-19
Add the permission model with the Privacy Taxonomy vocabulary

jwrosewell:
> **This description has been edited.** Permission signals are now selected
> the way every other provider type in the stack is selected, so the key and
> two of the provider names have changed. If you read an earlier version of
> this text, re-read "How the signal providers are selected". Nothing about
> precedence, withdrawal, the terms document list or the permission model
> itself has moved.
>
>
> 1. The order and the set of signal providers are given by
>    `[permission_signal] provider`, not `[permission_signal] sources`. It is
>    the same key name every provider type uses.
> 2. Two provider names are snake_case, so `gpp-sale-opt-out` is
>    `gpp_sale_opt_out` and `us-privacy` is `us_privacy`. `gpc` and `tcf` are
>    unchanged. The crate directories on disk keep their own spelling.
> 3. A provider with settings would put them in
>    `[permission_signal.<name>]`. None of the four shipped here has any, so
>    none of them needs a table.
>
> file are written up in

Stacks on device and location provider selection (#1044), and depends on its
`[geo] provider` selector, whose default this pull request changes to no
geolocation. Through #1044 this pull request also depends on the Edge Cookie
provider seam (#1043) for the `EdgeCookieProvider` trait, which it gates on
permissions. The stack has six pull requests
third, and the first five decompose #838 as requested in the #986 review.
[Compare `split/2-device-geo` with `split/3-permissions`](https://github.com/jwrosewell/trusted-server/compare/split/2-device-geo...split/3-permissions)

The spec for this pull request is
[`2026-07-30-permission-model-design.md`](https://github.com/jwrosewell/trusted-server/blob/split/6-integration-seam/docs/superpowers/specs/2026-07-30-permission-model-design.md),
revised to match this implementation, with revision records listing every
divergence and why, and section 13 covering the permission signal providers.
Reader documentation lands with this pull request at
[`docs/guide/permission-model.md`](https://github.com/jwrosewell/trusted-server/blob/split/3-permissions/docs/guide/permission-model.md)
and
[`docs/guide/permission-signals.md`](https://github.com/jwrosewell/trusted-server/blob/split/3-permissions/docs/guide/permission-signals.md).

Permissions become the primitive that gates identity features, and consent is
one of several ways a permission is established, alongside a country baseline,
an opt-out signal and configuration.

- [`permissions.rs`](https://github.com/jwrosewell/trusted-server/blob/split/3-permissions/crates/trusted-server-core/src/permissions.rs)
  resolves a per-request permission state, being the country and region
  baseline from `permissions.yaml` amended by the signals the request carries.
  `permissions.yaml` is compiled into the build, and the repository sample is
  [`config/permissions/vanilla.yaml`](https://github.com/jwrosewell/trusted-server/blob/split/3-permissions/config/permissions/vanilla.yaml).
  Permission names follow the IAB Privacy Taxonomy Data Uses.

### How the signal providers are selected

Signals are read by permission signal providers, which are crates outside core
under
[`crates/permission-signal/`](https://github.com/jwrosewell/trusted-server/tree/split/3-permissions/crates/permission-signal),
being Global Privacy Control, the GPP sale opt-out, the US Privacy string and
TCF v2, each implementing the `PermissionSignalProvider` trait that core
defines in
[`permission_signal/mod.rs`](https://github.com/jwrosewell/trusted-server/blob/split/3-permissions/crates/trusted-server-core/src/permission_signal/mod.rs).

```toml
[permission_signal]
provider = ["gpc", "gpp_sale_opt_out", "us_privacy", "tcf"]
```

An adapter selects them once at startup from `[permission_signal] provider`,
refusing an unknown or repeated name there rather than on the first request,
and carries them on the request services. An unknown name is reported with the
names the build does have, so a typo is answered with the list to pick from.

Precedence is the configured order. For each permission, core asks the
providers in the order `[permission_signal] provider` lists them. Each answers
grant, revoke or neutral, a neutral answer leaves the prior value standing, and
the last provider with an opinion decides. Leaving the list out runs every
provider the adapter links, in the adapter's order, and an empty list runs
none, which leaves every permission at its country and region baseline. Every
adapter offers the four in the same default order, being `gpc`,
`gpp_sale_opt_out`, `us_privacy` and `tcf`. Global Privacy Control, a browser
setting with no interface of its own, is asked first, and the three schemes
that carry a choice someone made through an interface are asked after it, which
means an answer given at a prompt amends the header the visitor arrived with. A
deployment that wants an opt-out to stand lists it after `tcf`.

### The rest of the model

- A consent record that is present but cannot be decoded revokes everything
  ahead of the providers, whichever of them are configured, rather than
  degrading to the no-signal baseline.
- Leaving `gpc` off the list stops that provider, but the consent pipeline
  still turns the same header into a US Privacy opt-out for a US privacy state
  when the consent setting `gpc_implies_optout` is on, which it is by default,
  and the `us_privacy` provider then acts on that record. A deployment that
  wants the header to have no effect turns that setting off as well.
- Cross-scheme tests in
  [`crates/trusted-server-adapter-axum/tests/permission_signals.rs`](https://github.com/jwrosewell/trusted-server/blob/split/3-permissions/crates/trusted-server-adapter-axum/tests/permission_signals.rs)
  assemble the four real providers and cover the default order against each
  opt-out, the reversed order, a provider left off the list, and one opt-out
  removed while the others stand.
- The TCF purpose to Data Use mapping lives in the TCF provider crate
  ([`crates/permission-signal/tcf/src/mapping.rs`](https://github.com/jwrosewell/trusted-server/blob/split/3-permissions/crates/permission-signal/tcf/src/mapping.rs)),
  in code with tests, because which purpose grants which Data Use is that
  scheme's own meaning rather than a deployment's policy. `permissions.yaml`
  keeps what a deployment decides, being which opt-out sources revoke which
  Data Uses and whether a TCF record answers at all. The purposes are the IAB
  TCF Europe purposes and what they grant are IAB Tech Lab Privacy Taxonomy
  Data Uses, so the mapping is checked against the industry's own documents
  rather than against us. Two purposes have no Data Use yet, so the crate
  carries the proposed `necessary.operations.storage` key for purpose 1 and the
  TCF identifier `select-basic-content` for purpose 11 until the taxonomy adds
  them.
- The permission state also carries `tdls`, the terms documents the request's
  data is available under, as the providers declare them, in provider order
  with duplicates removed, and the page receives `{"set":[],"tdls":[]}`. A
  permission says what may be done with the data and not on what basis, and a
  recipient offered data needs both, because it has to decide whether the terms
  are ones it accepts and whether it may pass the data on. An empty list says
  no terms were declared, which is not the same as terms permitting anything.
  Each entry addresses a published document that must never be edited once
  published, which is why a version belongs in its address, and the name
  matches the `tdl` member the
  [Data Labels work](https://github.com/jwrosewell/data-labels/blob/main/OpenRTB.md)
  puts on an OpenRTB node. None of the four schemes here declares terms, so the
  list is empty on every shipped path, and the tests drive it with a provider
  that declares one, through the production assembly.
- The four schemes that ship are a starting set rather than the list. Model
  Terms for Marketing, where a publisher and the parties it passes data to
  agree to be bound by a published set of terms, is the next provider and
  arrives in a following pull request, as a crate like the four here with no
  change to core.
  [`docs/guide/permission-signals.md`](https://github.com/jwrosewell/trusted-server/blob/split/3-permissions/docs/guide/permission-signals.md)
  says so too, so a reader is not left thinking the four are fixed.
- Destructive withdrawal is narrow. A provider answers withdrawal separately
  through `withdraws`, core scopes the answer to the storage baseline, and of
  the four only TCF answers it. Only a TCF record refusing storage in a
  jurisdiction whose baseline did not grant storage therefore expires the
  cookie and writes the identity-graph tombstone. Opt-outs suppress use, with
  Edge Cookie headers stripped and nothing sent beyond the edge, but never
  destroy an issued identifier, so lifting the opt-out restores the identity.
  This differs from `main`, where `has_explicit_ec_withdrawal` in
  [`consent/mod.rs`](https://github.com/IABTechLab/trusted-server/blob/main/crates/trusted-server-core/src/consent/mod.rs)
  also withdraws in a US privacy state for a Global Privacy Control signal, a
  GPP sale opt-out or a US Privacy sale opt-out. Under this pull request those
  opt-outs suppress use and keep the cookie, and the withdrawal tests `main`
  wrote around a California Global Privacy Control opt-out now use a TCF record
  that refuses storage.
- Sharing beyond the edge, being the bidstream `user.id`, the identify response
  and partner pull sync, requires both storage and personalized-ad selection
  (`StoreOnDevice` and `SelectPersonalisedAds` in `ec_sharing_allowed`), the
  same pair that gates bidstream EIDs, so a storage-only grant keeps
  first-party use while withholding partner sharing.
- The Edge Cookie gate moves from raw consent to the permission model. A
  provider declares `required_permissions()` and core runs it only when every
  one is set.
- The permission baseline and the jurisdiction to assume when no country
  resolves are declared together on the top node of the `rules` tree in
  `permissions.yaml`, so there is always a defined baseline and one file states
  the policy for both. A failed geo lookup is distinct from an unmatched
  country. It resolves at the requires-signal floor instead of the declared
  default and is logged at error level. Geolocation is now off by default, and
  a deployment that runs an Edge Cookie provider with no location provider must
  set `[geo] assume_single_jurisdiction = true`, acknowledging that every
  request is treated as the declared jurisdiction.
- `permissions.yaml` rules use an explicit per-permission acquisition map
  (`granted`, `requires_signal`, `denied`). Parsing rejects an unknown group,
  Data Use, acquisition or revoke keyword, a top node missing `group` or
  `jurisdiction`, and two place codes that differ only by case. An EU-27 plus
  EEA coverage test locks the `gdpr-eu` mapping.

## Changed from an earlier description

An earlier version of this text described precedence as a rule fixed in code,
where an opt-out always beat a consenting TCF record. Precedence is now the
configured order described above. Under the default order, a visitor who sends
`Sec-GPC: 1` and then consents through a consent management platform has the
Data Uses that the TCF record consents to set, which is the opposite of the old
outcome, and a deployment gets the old outcome back by listing the opt-out
after `tcf`. The pinning tests for the old rule are replaced by the
cross-scheme tests named above.

## How it was verified

Every job `main`'s CI runs was run locally against `e3bb4b977`, on Windows and
Clippy with warnings denied on the Axum, Cloudflare, Spin and Fastly adapters,
on both wasm targets and on the four permission signal crates, the Axum,
Cloudflare and Spin adapter suites, the cross-adapter parity suite, the
benchmark smoke run, the release wasm builds for Spin and Fastly, and the CLI,
OpenRTB codegen, `format-docs` and template cache harness jobs that only run on
Linux. The core suite passes with 2,851 tests natively and 2,845 under Viceroy.
The head then gained one commit, `88a7482f8`, which touches only
`tools/permissions-inspector/wasm`, a crate that is deliberately its own
workspace, so no workspace job's result changes. That crate builds clean for
wasm32-unknown-unknown and is rustfmt clean.

CI on that head was green across all 20 checks, being
[Run Tests](https://github.com/IABTechLab/trusted-server/actions/runs/34893309552),
[Run Format](https://github.com/IABTechLab/trusted-server/actions/runs/34893310592),
[Integration Tests](https://github.com/IABTechLab/trusted-server/actions/runs/34893309808),
[Permissions Inspector](https://github.com/IABTechLab/trusted-server/actions/runs/34893309663)
and
[CodeQL Advanced](https://github.com/IABTechLab/trusted-server/actions/runs/34893309244).

On the head this pull request shows now, `5c9f27e6c`, every CI check passes: [cargo test](https://github.com/IABTechLab/trusted-server/actions/runs/35028998507/job/104582798688), [the CLI tests](https://github.com/IABTechLab/trusted-server/actions/runs/35028998507/job/104582798844), [format-docs](https://github.com/IABTechLab/trusted-server/actions/runs/35028998512/job/104582798521), [the integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/35028998517/job/104584951513), [vitest](https://github.com/IABTechLab/trusted-server/actions/runs/35028998507/job/104582798750) and [CodeQL](https://github.com/IABTechLab/trusted-server/runs/104582979072).

## Framing

Privacy is a spectrum and technology is neutral. This model encodes no
jurisdiction's law. The deployer brings the policy in `permissions.yaml` and
configuration, decides their own baselines, and the code makes those decisions
inspectable and enforced. Trust comes from that flexibility being respected,
not from constraint.

References #778. Decomposes #838. Spec baseline from #986.


aram356:

jwrosewell:
To make the permissions-file discussion concrete, #1045 now carries the policy made real rather than argued in the abstract:

- No policy ever ships with Trusted Server, and the file now says so in its own header. The builder of a deployment chooses the permissions file compiled into their image, and the operator can overlay their own at configuration time.
- The repository's samples live in `config/permissions/`, each naming itself with a top-level `name:` line (the plain baseline is `Vanilla`), and choosing one is always an explicit choice.
- A permissions inspector at `tools/permissions-inspector/` lets anyone, policy people included, load a sample or paste their own file and see exactly which Data Uses are set for a visitor in a given place with given consent signals, with the reason per Data Use and each definition quoted from the IAB Tech Lab Privacy Taxonomy. Step 3 is computed by trusted-server-core itself compiled to WebAssembly, and the page states the version, branch, commit and date of the engine it runs, so the page and the code cannot drift apart.

That is the case for a standalone YAML document in the taxonomy's own vocabulary, separate from the code and from any one program's configuration surface: a compliance owner can read it, a tool can inspect it, and the same file works in a browser page and in the compiled server. A temporary hosting of the inspector is at https://jwrosewell.github.io/trusted-server/tools/permissions-inspector/ until the repository hosts its own.

PR #1046, opened 2026-08-19
Add the client-set Edge Cookie value path

jwrosewell:
> **This description has been edited.** The demonstration provider's
> configured name has changed to fit the one configuration convention the whole
> stack lands. If you read an earlier version of this text, that name is the
> only thing that moved. The resolve endpoint, its checks, the marker cookie
> and what is deliberately left for v2 are all the same.
>
>
> 1. The demonstration provider is selected as `client_fixed`, not
>    `client-fixed`. Every name a person types in the configuration file is
>    snake_case now. A deployment still carrying the old spelling stops at
>    startup with a message naming `client_fixed`, rather than being told to
>    add a settings table for a provider that needs none.
> 2. The cargo feature that compiles it in keeps its own spelling,
>    `client-fixed-demo`, because a build feature is not something a deployment
>    writes in `trusted-server.toml`. The two spellings are deliberate and the
>    documentation says which is which.
> 3. Any settings a client-cycle provider has would go in `[ec.<name>]`, as
>    with every other Edge Cookie provider. The demonstration provider has
>    none, so it needs no table.
>
> file are written up in

Stacks on the permission model (#1045), and depends on it because the resolve
endpoint applies the same permission gate as organic generation. Through the
stack this pull request also depends on the Edge Cookie provider seam (#1043),
whose `EdgeCookieProvider` trait gains the client resolve method here. The
stack has six pull requests (#1043, #1044, #1045, #1046, #1047, #1094), each
targeting `main`, with this one fourth, and the first five decompose #838 as
requested in the #986 review.
[Compare `split/3-permissions` with `split/4-client-resolve`](https://github.com/jwrosewell/trusted-server/compare/split/3-permissions...split/4-client-resolve)

The spec for this pull request is
[`2026-07-30-client-cycle-ec-resolve-design.md`](https://github.com/jwrosewell/trusted-server/blob/split/6-integration-seam/docs/superpowers/specs/2026-07-30-client-cycle-ec-resolve-design.md),
the threat-model draft revised to the implemented state. Its revision record
maps each requirement to what v1 implements and what deliberately waits for the
first vendor scheme.

## Why this path matters

A client-cycle provider establishes the identifier through a browser round
trip. The page script obtains or derives a value, which for a real vendor is a
signed envelope from that vendor's identity system, posts it to
`POST /_ts/api/v1/ec/resolve`, and the provider verifies that value before the
edge sets it as the Edge Cookie. The vendor Edge Cookie provider proposed in the
51Degrees module issue (#1072) works client-side by design, so this endpoint is
on the series' critical path rather than deferred.

- `EdgeCookieProvider` gains `resolve_from_client` with a default that creates
  nothing, so server-side providers are untouched.
- The resolve endpoint enforces the same rules as organic generation plus its
  own. The permission gate applies unchanged, being the provider's complete
  declaration rather than a hard-coded storage check, and the provider's
  response headers pass the same reserved-surface check before any is kept. The
  request must carry an `Origin` on the publisher's domain (403 otherwise) and
  a `text/plain` or `application/json` body (415 otherwise, 413 over 64 KiB). A
  created identifier must fit the global identifier bounds (400) and must not
  silently replace a different identity already on the request (409).
- A created identifier is persisted to the identity graph before the cookie is
  set, keyed by the provider's canonical form, so withdrawal reaches a
  client-set identity the same way it reaches an edge-created one. With no
  graph available the endpoint sets no cookie at all, which is stricter than
  organic generation, where the identifier is kept and only the row write is
  skipped. A graph write failure answers 503 with no cookie. A closed
  permission gate, no configured provider, no graph, or a provider that creates
  nothing answers 204 with no Edge Cookie. The provider's own response headers
  go out only with the 200 that sets the cookie, or with the 204 when the
  provider creates nothing. Every response the handler builds carries
  `Cache-Control: no-store`, and a provider or configuration error goes to the
  adapter's error response instead.
- The Edge Cookie stays HttpOnly. A non-HttpOnly companion marker cookie
  (`ts-ecr=1`, no identity content) tells the page script a resolve succeeded,
  fixing the earlier defect where the script checked for a cookie it could
  never read and so posted on every page view. The marker expires together with
  the Edge Cookie on withdrawal. A Rust test checks that the marker name and
  the demonstration's fixed word match the page-script source.
- The `client_fixed` demonstration provider is compiled only behind the
  `client-fixed-demo` cargo feature, and production builds reject the selection
  at startup. A fixed shared word is not an identity. The demonstration
  provider's cookie carries the registry code (`cfix~an-ec`), so even
  demonstration identities are provider-namespaced. The spelling this provider
  was selected by before the snake_case rule, `client-fixed`, is refused at
  startup with a message naming `client_fixed`, so a deployment carrying the
  old name is told what to write rather than being asked for a settings table
  the provider does not have.
- The Fastly adapter routes the endpoint with the same bot-gated identity graph
  as generation. The other adapters deliberately do not route the endpoint yet,
  matching `identify` and `batch-sync`, which need the same platform key-value
  wiring those adapters lack. The Spin adapter's route list records it.
- The full anti-replay reservation design from the spec is deliberately not in
  v1. The reservation design needs the real vendor envelope's unique id and
  session binding, and the spec keeps the design verbatim as the bar for that
  implementation.

## How it was verified

Every job `main`'s CI runs was run locally against `2c34be429`, on Windows and
Clippy with warnings denied on the Axum, Cloudflare, Spin and Fastly adapters,
on both wasm targets and on the four permission signal crates, the Axum,
Cloudflare and Spin adapter suites, the cross-adapter parity suite, the
benchmark smoke run, the release wasm builds for Spin and Fastly, and the CLI,
OpenRTB codegen, `format-docs` and template cache harness jobs that only run on
Linux. The core suite passes with 2,885 tests natively and 2,879 under Viceroy.
The head then gained the merge `b7dd13f8e`, which touches only
`tools/permissions-inspector/wasm`, a crate that is deliberately its own
workspace, so no workspace job's result changes. That crate builds clean for
wasm32-unknown-unknown and is rustfmt clean.

CI on that head was green across all 20 checks, being
[Run Tests](https://github.com/IABTechLab/trusted-server/actions/runs/34894669983),
[Run Format](https://github.com/IABTechLab/trusted-server/actions/runs/34894670097),
[Integration Tests](https://github.com/IABTechLab/trusted-server/actions/runs/34894670099),
[Permissions Inspector](https://github.com/IABTechLab/trusted-server/actions/runs/34894670101)
and
[CodeQL Advanced](https://github.com/IABTechLab/trusted-server/actions/runs/34894670111).

Endpoint tests in
[`crates/trusted-server-core/src/ec/resolve.rs`](https://github.com/jwrosewell/trusted-server/blob/split/4-client-resolve/crates/trusted-server-core/src/ec/resolve.rs)
cover the origin, content type, body size, conflict, no-graph and marker
behaviors, and `client_set_value_round_trips_through_the_ec_scenario` drives a
client identifier through deferral, resolve, cookie set and verbatim
read-back. The endpoint tests in that file do not drive the identifier-bound
400 or the graph-write 503 answer.

On the head this pull request shows now, `ea716f5a1`, every CI check passes: [cargo test](https://github.com/IABTechLab/trusted-server/actions/runs/35029005084/job/104582818922), [the CLI tests](https://github.com/IABTechLab/trusted-server/actions/runs/35029005084/job/104582818866), [format-docs](https://github.com/IABTechLab/trusted-server/actions/runs/35029005036/job/104582818344), [the integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/35029005064/job/104585420816), [vitest](https://github.com/IABTechLab/trusted-server/actions/runs/35029005084/job/104582818782) and [CodeQL](https://github.com/IABTechLab/trusted-server/runs/104583018847).

References #778. Decomposes #838. Spec baseline from #986.


aram356:

Issue #846, opened 2026-07-06
Move secrets from trusted-server.toml to secret store


aram356:
### Description

Example: proxy passphrase

PR #1048, opened 2026-08-19
Add configurable trusted client-IP header support for Fastly

### LINKED ISSUE #1040: Client IP resolves to the fronting edge node, not the reader, when TS runs behind another Fastly service
## Summary

On the Fastly adapter, the client IP is taken only from `req.get_client_ip_addr()`. Fastly defines that as the immediate downstream peer. When a Trusted Server Compute service runs behind another Fastly service (service chaining), that peer is the fronting service's edge node, so every request presents the POP's IP instead of the reader's.

Geo lookup and EC ID generation both read that same single value, so both are wrong together. It fails silently: the site works, ads serve, and nothing errors.

## Evidence

Measured against a live TS deployment (`x-ts-version: 404`) from one residential client IP. Same service, same passphrase, same source IP, same request. The only difference is whether the request went straight to the Compute service or through the fronting Fastly delivery service. Publisher identity redacted.

| Route | `x-geo-city` | `x-geo-coordinates` | `x-geo-region` |
| --- | --- | --- | --- |
| `ts.prospect-a.com` (direct, 1 hop) | southampton | 40.89,-72.4 | NY |
| `www.prospect-a.com` (chained, 2 hops) | **newark** | **40.77,-74.17** | **NJ** |

Three runs each, no variance. `x-served-by` names an EWR node on both, and 40.77,-74.17 is the EWR POP itself. The one-hop row is the reader's ISP geo; the two-hop row is the datacenter.

Note that the chained response carries two `x-served-by` entries and the direct response one, which is the extra hop showing up in the response.

## Why this reaches EC and not just geo

Geo and EC consume the same value:

- `client_info_from_request` sets `client_ip` from `req.get_client_ip_addr()` (`crates/trusted-server-adapter-fastly/src/platform.rs:708`)
- `FastlyPlatformGeo::lookup` geolocates that value (`crates/trusted-server-adapter-fastly/src/platform.rs:690-694`)
- `EcContext` reads the same `services.client_info().client_ip`, normalizes it, and HMACs it (`crates/trusted-server-core/src/ec/mod.rs:210-212` and `275-281`, `crates/trusted-server-core/src/ec/generation.rs:91`)

So the geo divergence above is direct evidence that the EC HMAC input diverged. Behind a fronting service, the EC hash prefix is derived from the edge node, and collapses to roughly one value per POP.

## Impact

- **Cluster classification.** `evaluate_cluster` counts KV keys sharing the 64-hex prefix to detect NAT and office networks (`crates/trusted-server-core/src/ec/kv.rs:696-697`). One prefix per POP means `cluster_size` grows toward every EC ID ever minted there, against a default `cluster_trust_threshold` of 10 (`crates/trusted-server-core/src/settings.rs:472`). Readers progressively misclassify as shared networks rather than individuals. The mechanism is working correctly on a constant input.
- **Pull-sync rate limiting.** `pull_rate_limit_key` is `pull:{source_domain}:{ec_hash}` (`crates/trusted-server-core/src/ec/pull_sync.rs:286`), so a per-user limit becomes a single bucket shared by every reader through that POP.
- **Consent jurisdiction.** Consent context takes geo as an input (`crates/trusted-server-core/src/ec/mod.rs:214-222`), so jurisdiction is derived from the POP location rather than the reader's. Benign for a single-country deployment, wrong for anything EU-facing.

## What is not affected

This is not an identity merge. `generate_ec_id` appends a random 6-character suffix, so each minted EC ID is still unique, KV keys stay distinct, and returning readers are recognized from the `ts-ec` cookie regardless of IP. The damage is confined to anything derived from the HMAC prefix or from geo.

## Root cause

Per Fastly's documentation, `client.ip` is always the immediate peer, and across a shield or a chain the second service sees the first. The original client IP survives in the `Fastly-Client-IP` header, which "doesn't change when being forwarded from one Fastly server to another." The Fastly adapter never reads that header.

- https://www.fastly.com/documentation/reference/vcl/variables/client-connection/client-ip/
- https://www.fastly.com/documentation/reference/http/http-headers/Fastly-Client-IP/
- https://www.fastly.com/documentation/guides/concepts/service-chaining/

## Scope

Fastly adapter only. The Cloudflare adapter reads `cf-connecting-ip` (`crates/trusted-server-adapter-cloudflare/src/platform.rs:727`) and the Spin adapter has an equivalent. Fastly is the one adapter with no header path.

This is not limited to Fastly-behind-Fastly. Any deployment where TS sits behind a CDN has the same defect, which covers the publisher-fronted topologies TS documents as supported.

## Fix

Deliberately not "read `Fastly-Client-IP` instead." That header is not protected at the Fastly edge (per Fastly: "if a client sets this header themselves, we will use it"), and it is not in `SPOOFABLE_FORWARDED_HEADERS` (`crates/trusted-server-core/src/http_util.rs:40-45`). Preferring it unconditionally would let a caller choose their own EC identity by setting a request header, which is worse than the current bug.

The fix needs a trust boundary. Tracked separately.

## Reproducing

With TS deployed behind a fronting Fastly service, request the same path directly against the Compute service and through the front door from the same client, and compare `x-geo-city` / `x-geo-coordinates`. Note that responses with status 401 suppress geo headers by design (`crates/trusted-server-adapter-fastly/src/middleware.rs:162-175`), so the request has to reach a 200.

### LINKED ISSUE #1041: Add a configurable trusted client-IP header for CDN-fronted deployments
## Problem

Trusted Server on Fastly resolves the client IP solely from `req.get_client_ip_addr()` (`crates/trusted-server-adapter-fastly/src/platform.rs:708`), which is the immediate downstream peer. Any deployment with a CDN or another Fastly service in front therefore sees the fronting node's IP rather than the reader's, which corrupts geo, EC ID generation, cluster classification, and consent jurisdiction. Measured evidence is in the linked bug.

The Cloudflare and Spin adapters already read a client-IP header. Fastly has no header path at all.

## Why this is not a one-line change

The obvious fix, preferring `Fastly-Client-IP`, is unsafe on its own. Fastly does not protect that header at ingress: "The value is not protected from modification at the edge of the Fastly network, so if a client sets this header themselves, we will use it." It is also absent from `SPOOFABLE_FORWARDED_HEADERS` (`crates/trusted-server-core/src/http_util.rs:40-45`), so TS forwards whatever arrives.

Because that value becomes the HMAC input for EC ID generation (`crates/trusted-server-core/src/ec/generation.rs:91`), trusting it unconditionally would let any caller select their own EC identity, and drive the cluster counter, by setting a request header. That is a worse failure than the bug it fixes.

The Cloudflare adapter's use of `cf-connecting-ip` is safe only because Cloudflare overwrites that header at ingress. Fastly does not, so mirroring the Cloudflare pattern onto the Fastly adapter would look consistent while being unsound.

## Proposal

A configurable trusted client-IP header, opt-in, with the trust boundary established by a shared secret set at the front door rather than by header presence.

- New optional config. Absent by default, so existing deployments keep current behavior and there is no silent change to EC ID derivation.
- When configured, the adapter reads the client IP from the named header only if the accompanying secret matches, and otherwise falls back to `req.get_client_ip_addr()`.
- When not configured, the header name is added to `SPOOFABLE_FORWARDED_HEADERS` so it is stripped on entry.
- Values that fail to parse as an IP fall back rather than erroring.

Using a shared secret rather than `Fastly-FF` or peer detection keeps one mechanism working across every fronting topology TS supports: a Fastly delivery service, Akamai, Cloudflare, or CloudFront. Fastly's own guidance for validating a Fastly hop is `fastly.ff.visits_this_service`, a VCL variable with no confirmed Compute equivalent, and `Fastly-FF` can be present on requests originating outside the network.

## Known limitation to document

Fastly's no-code request routing provides no place to inject headers. If a deployment fronts TS that way and the client IP is not preserved by the platform, there is no header for TS to read and this mechanism cannot help. Whether request routing preserves the client IP is unverified and worth measuring separately, since it determines whether that topology is viable for EC at all.

## Acceptance

- Config absent: behavior identical to today, header stripped on entry.
- Config present with a valid secret: geo and EC derive from the forwarded IP.
- Config present with a missing, wrong, or malformed secret: falls back to the peer IP, no error.
- Documentation covers the front-door requirement, since a reader without a correctly configured front door is the spoofing hole.

## References

- https://www.fastly.com/documentation/reference/http/http-headers/Fastly-Client-IP/
- https://www.fastly.com/documentation/reference/http/http-headers/Fastly-FF/
- https://www.fastly.com/documentation/reference/vcl/variables/client-connection/client-ip/

Bug and measured evidence: #1040

prk-Jr:
## Summary

On Fastly, Trusted Server takes the client IP from `req.get_client_ip_addr()`, which Fastly defines as the **immediate peer**. Behind a fronting Fastly or CDN service that peer is the fronting edge node, so geolocation, EC identity derivation, and consent jurisdiction all describe the POP instead of the reader. It fails silently — pages render, ads serve, nothing errors. Measured evidence is in #1040.

Reading `Fastly-Client-IP` on its own would be worse than the bug: Fastly does not protect it at ingress, so any caller could choose the address that feeds EC identity, geo, and bot classification.

This PR adds an **opt-in** `[trusted_client_ip]` section. The forwarded address is accepted only when the same request carries a shared secret that only the front door knows. Every other outcome falls back to the peer address without rejecting the request.

Closes #1041
Fixes #1040

---

## What changes for existing deployments

**With no `[trusted_client_ip]` section, client-IP resolution is byte-identical to today.** One behavior change lands regardless of configuration:

`fastly-client-ip` is added to `SPOOFABLE_FORWARDED_HEADERS`, so the Fastly adapter strips it at request entry. Trusted Server never read that header, so nothing internal changes — but the origin request reuses the inbound header set, so **Trusted Server no longer forwards an inbound `Fastly-Client-IP` to the publisher origin**. Confirm whether the origin reads it for geolocation, fraud checks, or logging before merging.

## Deploy order

> [!WARNING]
> `Settings` uses `deny_unknown_fields`. Once a pushed config blob contains `[trusted_client_ip]`, **any binary predating this PR fails to load it and returns its startup-error response.**
>
> - **Rolling out:** deploy the code to every instance **first**, then `ts config push`.
> - **Rolling back:** restore a config without the section **first**, then roll the binary back.
>
> Getting this order wrong takes the service down rather than degrading it. The unconfigured direction is safe — `skip_serializing_if = "Option::is_none"` keeps the key out of the blob entirely, and a compat test pins that against a model of the pre-change schema.

## How resolution works

Per request, in `resolve_and_sanitize_client_ip`:

1. Read `auth_header`. Must be exactly one UTF-8 value.
2. Compare against `shared_secret` in constant time (SHA-256 + `subtle::ct_eq`).
3. Read `ip_header`. Must be exactly one UTF-8 value.
4. Parse as a bare IPv4 or IPv6 address. No trimming, no port, no zone suffix.
5. Strip both trust headers plus the static spoofable list, then route.

Resolution and sanitization are fused into one function so the ordering cannot be reversed by a future edit — resolution has to observe the headers before sanitization removes them.

| Request state | Address used |
| --- | --- |
| One matching auth value and one bare IP value | Forwarded reader |
| Auth value missing, empty, wrong, duplicated, or not UTF-8 | Immediate peer |
| IP value missing, duplicated, not UTF-8, or not a bare IP | Immediate peer |
| Request bypassed the front door | Immediate peer |
| No `[trusted_client_ip]` section | Immediate peer |

No combination rejects the request. A rotated secret, renamed header, or misconfigured front door degrades to current behavior. Fallbacks taken while a config is present log at debug — not warn, so a direct caller cannot drive log volume with junk headers. Neither the secret nor the address is logged.

## Front-door requirement

Code configuration alone does not enable this. The fronting VCL service must set both headers, and must remove client-supplied copies. Full setup in [`docs/guide/fastly.md`](docs/guide/fastly.md#cdn-fronted-client-ip); the shape is:

```vcl
sub vcl_recv {
  if (fastly.ff.visits_this_service == 0) {
    # Client-supplied copies never survive, on any route.
    unset req.http.X-TS-Client-IP;
    unset req.http.X-TS-Client-IP-Auth;

    # Stamp only on the Trusted Server route, so the secret never reaches
    # another backend.
    if (req.http.host == "www.example.com") {
      set req.http.X-TS-Client-IP = client.ip;
      set req.http.X-TS-Client-IP-Auth =
        table.lookup(ts_private_config, "trusted_client_ip_secret");
    }
  }
}
```

Four things this shape gets right, all explained in the docs:

- **A dedicated `x-` header name rather than `Fastly-Client-IP`.** `fastly-client-ip` is still accepted and suits a dedicated service, but on a shared service a dedicated name means the front door never modifies `Fastly-Client-IP`, so security rules, rate limiters, logging formats, and vendor snippets that read it keep working unchanged.
- **`unset` before `set`.** Trusted Server ignores the forwarded address whenever either header carries more than one value. Without the `unset`, a reader can send its own copy, force the fallback, and keep its real address out of geolocation and bot classification. This is anti-evasion, not tidiness.
- **The route condition wraps only the `set` lines.** Client-supplied values die on every route; the secret is added only on the route that reaches Trusted Server, so it never reaches another backend or another backend's logs.
- **No `req.restarts` guard.** A restart can change which backend a request reaches. Re-running the block each pass re-evaluates the route condition, so a stamp made before a restart cannot follow the request to a different backend.

The secret comes from a private (write-only) edge dictionary, not inline VCL — inline VCL is readable by anyone with service-configuration access and is preserved in every version diff.

## Confirm the topology first

| Topology | `client.ip` at the front door | Result |
| --- | --- | --- |
| Reader → fronting Fastly service → Trusted Server | The reader | Correct reader address |
| Reader → another CDN → Fastly → Trusted Server | That CDN's node | **Wrong address, and authenticated as valid** |
| Reader → Trusted Server directly | Not applicable | Falls back to the immediate peer |
| Fastly no-code request routing | No injection point | Mechanism unavailable |

Row 2 is the only topology that fails with a wrong value instead of falling back. If another CDN precedes Fastly, restrict direct access to the Fastly front door and derive `ip_header` from that CDN's protected reader-IP value rather than from `client.ip`.

Fastly no-code request routing provides no point to inject headers, so this mechanism cannot help that topology.

## Validation

Startup rejects, without exposing the secret in the error:

- `shared_secret` shorter than 32 bytes, or containing any byte outside `!`–`~` (whitespace, tab, control, DEL, non-ASCII)
- the documented placeholder value, via `reject_placeholder_secrets`
- `ip_header` that is neither `fastly-client-ip` nor `x-`-prefixed; `auth_header` that is not `x-`-prefixed
- the two header names being equal, or either reusing a Trusted Server internal header name

`Redacted<String>` keeps the secret out of debug output and validation errors.

## Secret storage

Redaction does not move the value into a platform secret store. With the current configuration architecture, `ts config push` serializes `shared_secret` into the Trusted Server application-config blob, so access to that config store must be restricted — it is the only gate on forging the reader address.

Fastly recommends Secret Store rather than Config Store for sensitive values. Migrating this field requires a shared-schema change to hold a secret-store reference and resolve it in the Fastly request path. The broader migration of existing Trusted Server passphrases and secrets is tracked by #846 and is outside this PR.

The fronting copy is a separate concern: a Fastly VCL service cannot read a Compute Secret Store, so it must obtain the identical value through a VCL-accessible mechanism such as a write-only edge dictionary.

## Scope notes

- **Only Fastly resolves client IP from this section.** Cloudflare, Spin, and Axum validate it and strip the configured headers before routing — so a shared multi-adapter config cannot leak the authentication value into publisher or integration handlers — but keep using their own runtime client address.
- **`X-Forwarded-For` is unchanged.** An earlier partial hardening was removed after review because it covered Fastly and Spin but not Cloudflare and Axum. Consistent reconstruction from authoritative `ClientInfo` should be handled separately.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/settings.rs` | Validated, redacted `[trusted_client_ip]` config; constant-time authentication; 32-byte ASCII-graphic minimum; placeholder rejection; `skip_serializing_if` for blob compatibility. |
| `crates/trusted-server-core/src/http_util.rs` | Treat `fastly-client-ip` as spoofable after authenticated consumption; shared sanitizer for configured trust headers. |
| `crates/trusted-server-adapter-fastly/src/platform.rs` | `resolve_client_ip`: exactly one authenticated IPv4/IPv6 value, peer fallback, debug-level fallback categories with no secret or address logged. |
| `crates/trusted-server-adapter-fastly/src/compat.rs` | `resolve_and_sanitize_client_ip` fuses resolution ahead of stripping so the order cannot be reversed. |
| `crates/trusted-server-adapter-fastly/src/main.rs` | Resolve once; propagate through authoritative `ClientInfo`. |
| `crates/trusted-server-adapter-fastly/src/middleware.rs` | Use authoritative `ClientInfo` for response geolocation, including authoritative absence. |
| `crates/trusted-server-adapter-fastly/src/app.rs` | Assert the resolved address reaches request-scoped services and the EC finalization context. |
| `crates/trusted-server-adapter-{axum,cloudflare,spin}/src/middleware.rs` | New `SanitizeRequestMiddleware` that strips the configured trust headers. Client-IP resolution unchanged. |
| `crates/trusted-server-adapter-{axum,cloudflare,spin}/src/app.rs` | Register `SanitizeRequestMiddleware` as the outermost middleware, so nothing else observes the authentication header. |
| `trusted-server.example.toml` | Commented `[trusted_client_ip]` example. |
| `docs/guide/configuration.md` | Fields, validation, deploy-order warning, env overrides, fallback table, cross-adapter sanitization, config-blob storage, origin-forwarding note. |
| `docs/guide/fastly.md` | Front-door setup, dedicated header-name guidance, per-line VCL rationale, failure-mode and topology tables, verification procedure, no-code routing limitation. |
| `docs/superpowers/specs/…-design.md`, `docs/superpowers/plans/…` | Security and data-flow design; implementation and verification plan. |

## Test plan

- [x] `cargo test-fastly`, `cargo test-axum`, `cargo test-cloudflare`, `cargo test-spin`
- [x] `cargo clippy-fastly`, `-axum`, `-cloudflare`, `-cloudflare-wasm`, `-spin-native`, `-spin-wasm`
- [x] JS tests and format (`npx vitest run`, `npm run format`) on pinned Node 24.12.0
- [x] Docs format (`cd docs && npm run format`)
- [x] WASM release build: `cargo build --package trusted-server-adapter-fastly --release --target wasm32-wasip1`
- [x] Local Viceroy: authenticated forwarded IP selects the configured Sydney geo fixture; missing or incorrect authentication falls back to the peer-IP San Francisco fixture without rejecting the request.
- [ ] **Live Fastly service-chain validation against the two-hop topology.** Not yet run — the checks above do not prove the handoff end to end. Verification procedure is in `docs/guide/fastly.md`: compare `x-geo-city` / `x-geo-coordinates` through the front door against a direct request, then replay with duplicate and junk trust headers to confirm the front door's `unset` holds.
- [ ] Confirm whether the publisher origin reads `Fastly-Client-IP` (see *What changes for existing deployments*).

## Checklist

- [x] Uses repository `log` macros, not `println!`
- [x] New code has tests
- [x] No real secrets or credentials committed


crates/trusted-server-core/src/settings.rs
```diff
@@ -1911,13 +1913,119 @@ pub struct TesterCookieConfig {
...
+
 #[derive(Debug, Default, Clone, Deserialize, Serialize, Validate)]
 #[serde(deny_unknown_fields)]
 pub struct Settings {
     #[validate(nested)]
     pub publisher: Publisher,
     #[serde(default)]
     pub tester_cookie: TesterCookieConfig,
+    /// Optional authenticated trusted client IP forwarding configuration.
+    #[serde(default)]
```
ChristianPavilonis: 🔧 **P1: Disabled configuration breaks older binaries during rollout or rollback**

When `[trusted_client_ip]` is absent, this field is `None`, but `#[serde(default)]` only affects deserialization. Derived serialization still writes `"trusted_client_ip": null` into every blob produced by `ts config push`. `TrustedServerAppConfig::serialize` delegates directly to `Settings::serialize`, while the base revision's `Settings` uses `deny_unknown_fields` and does not know this key.

Pushing an otherwise unchanged config before upgrading all instances, or rolling back after that push, makes old instances reject the blob and return startup-error HTTP 500 responses even though the feature was never enabled. The existing `AuctionConfig` compatibility logic explicitly omits default new fields for this reason.

Please omit `None` during serialization and add a regression test proving a default payload lacks this key and remains accepted by a schema matching the base revision. If the field is configured, document that operators must restore a compatible blob before rolling back.

```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
```

ChristianPavilonis (changes requested):
## Summary

Requesting changes for one high-severity config rollout and rollback compatibility issue. The trusted client-IP runtime behavior and validation otherwise looked sound at the reviewed revision.

crates/trusted-server-core/src/settings.rs
```diff
@@ -2596,13 +2598,126 @@ pub struct TesterCookieConfig {
...
+
+    if ip_header == auth_header {
+        return Err(ValidationError::new("identical_trusted_client_ip_headers"));
+    }
+
+    for header in [&ip_header, &auth_header] {
+        if matches!(header.as_str(), "x-ts-tls-protocol" | "x-ts-tls-cipher") {
+            return Err(ValidationError::new("reserved_trusted_client_ip_header"));
+        }
+    }
```
aram356, line 2676: ♻️ **refactor** - Reserved-header validation covers only the two TLS bridge headers. Every other internal trust header passes the `x-` prefix rule: `ip_header = "x-geo-info-available"`, `"x-forwarded-for"`, or `"x-ts-ec"` are all accepted, and the sanitizer then strips platform-injected internal signals before routing. On Cloudflare the finalize middleware reads `x-geo-info-available` from the request before sanitization, so a colliding trust header would double as a geo signal. `INTERNAL_HEADERS` in `constants.rs` is exactly this reserved set and already contains both TLS bridge names, so the existing reserved-header test passes unchanged.

```suggestion
    for header in [&ip_header, &auth_header] {
        if crate::constants::INTERNAL_HEADERS.contains(&header.as_str()) {
            return Err(ValidationError::new("reserved_trusted_client_ip_header"));
        }
    }
```

If you take this, consider hoisting the path into a `use crate::constants::INTERNAL_HEADERS;` import and adding one validation test case for a non-TLS reserved name; both live outside this hunk, so the suggestion cannot carry them.

.gitignore
```diff
@@ -52,6 +52,7 @@ src/*.html
 
 /guest-profiles
 /benchmark-results/**
+/.worktrees/
```
aram356, line 55: ⛏ **nitpick** - Duplicate entry: line 41 ("Agent implementation worktrees") already ignores `/.worktrees/` and was present before this PR. This second copy looks like a merge artifact; the plan doc records the decision as retaining the existing ignore, not adding another one.

```suggestion
```

crates/trusted-server-adapter-axum/src/middleware.rs
```diff
@@ -35,7 +36,11 @@ impl FinalizeResponseMiddleware {
 
 #[async_trait(?Send)]
 impl Middleware for FinalizeResponseMiddleware {
-    async fn handle(&self, ctx: RequestContext, next: Next<'_>) -> Result<Response, EdgeError> {
+    async fn handle(&self, mut ctx: RequestContext, next: Next<'_>) -> Result<Response, EdgeError> {
+        sanitize_trusted_client_ip_headers(
```
aram356, line 40: 🤔 **thinking** - The "strip trust headers before routing" invariant now lives inside `FinalizeResponseMiddleware`, a response-finalization middleware, and holds only because this middleware happens to be registered first in each adapter's router. Nothing enforces that ordering: a middleware registered ahead of it later, or a reorder during a refactor, would silently re-expose the auth header to earlier request handling, and the middleware's name gives no hint that it carries a security invariant. Consider a dedicated request-sanitization middleware, or at minimum a comment at each registration site pinning the ordering requirement. The same applies to the Cloudflare and Spin copies of this change.

crates/trusted-server-adapter-fastly/src/platform.rs
```diff
@@ -694,7 +695,53 @@ impl PlatformGeo for FastlyPlatformGeo {
...
+) -> Option<IpAddr> {
+    let Some(config) = config else {
+        return peer_ip;
+    };
+    let Some(auth_candidate) = single_utf8_header(req, &config.auth_header) else {
+        log::debug!("Trusted client IP: auth header is missing, duplicated, or not UTF-8");
+        return peer_ip;
+    };
+    if !config.authenticates(auth_candidate) {
+        log::debug!("Trusted client IP: auth header did not match the configured shared secret");
```
aram356, line 730: 🌱 **seedling** - Debug-level fallback logging avoids client-driven log volume, but it also means a rotated-but-unsynced secret or a renamed front-door header silently degrades geo, EC derivation, and bot protection to Fastly edge-node addresses. A follow-up counter (for example, a fallback-with-config-present rate in telemetry) or a synthetic check that asserts the forwarded address round-trips would make that failure mode observable in production. Not for this PR.

aram356 (changes requested):
## Summary

Adds authenticated trusted client-IP forwarding for Fastly with peer fallback, cross-adapter trust-header stripping, and validated configuration. No blocking defects found in the resolution path: authentication precedes IP parsing, the comparison is constant-time over fixed-size digests, malformed or duplicated input fails closed to the peer address, no secret or header value is logged, and the config-blob rollback hazard is regression-tested against a modeled base-revision schema. The findings below are hardening and maintainability items.

> 2 of the inline comments below carry a one-click GitHub `suggestion`. Use **Commit suggestion** (or **Add suggestion to batch**) to apply them as commits on the PR branch. The remaining comments describe concerns in prose because the change is design-level or spans files outside the diff.

## Non-blocking

### ♻️ refactor

- **Reserve all internal trust headers, not just the TLS bridge pair** - see inline at `crates/trusted-server-core/src/settings.rs:2672`

### ⛏ nitpick

- **Duplicate `/.worktrees/` entry** - see inline at `.gitignore:55`

### 🤔 thinking

- **Trust-header sanitization depends on unstated middleware ordering** - see inline at `crates/trusted-server-adapter-axum/src/middleware.rs:40`

### 🌱 seedling

- **No operator-visible signal when the trust handshake degrades** - see inline at `crates/trusted-server-adapter-fastly/src/platform.rs:730`

## CI Status

- cargo fmt: PASS (required)
- cargo test: PASS (required)
- integration tests: PASS
- vitest: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- Analyze (javascript-typescript): PASS (both runs)
- Analyze (actions): PASS

crates/trusted-server-core/src/settings.rs
prk-Jr: Applied, widened past the suggestion: the check now tests `INTERNAL_HEADERS.contains(&header.as_str())` with a top-level `use crate::constants::INTERNAL_HEADERS;` import. Renamed the test to `trusted_client_ip_rejects_reserved_internal_headers` and added the three non-TLS cases you called out — `x-forwarded-for`, `x-geo-info-available` as `ip_header`, and `x-ts-ec` as `auth_header`. Updated the reserved-header paragraph in `docs/guide/configuration.md` so it no longer enumerates only the TLS pair.

Commit: 3329711d

.gitignore
prk-Jr: Removed. Line 41 already covers it; the plan doc records retaining that entry, not adding a second one. Commit: 3329711d

crates/trusted-server-adapter-axum/src/middleware.rs
prk-Jr: Took the dedicated-middleware option on all three adapters. `SanitizeRequestMiddleware` (Cloudflare, Spin, Axum) now owns the strip and is registered outermost, ahead of `FinalizeResponseMiddleware`; `FinalizeResponseMiddleware::handle` is back to taking `ctx` by value and its doc no longer claims to be outermost. The strip test moved with it (`sanitize_middleware_strips_configured_trust_headers_before_routing`).

To be explicit about what this does and does not buy: a dedicated middleware is still registration-order dependent — nothing in the type system stops someone from registering ahead of it. What changed is that the invariant is now named and stated in two places instead of hidden in a response-finalization middleware: the type's doc comment says it must stay first and why, and each registration site in `app.rs` carries a comment pinning the requirement. Fastly is unaffected — it resolves and strips at the `main.rs` entry point before dispatch, so it never relied on middleware ordering for this.

Commit: 3329711d

crates/trusted-server-adapter-fastly/src/platform.rs
prk-Jr: Agreed it is a real gap and out of scope here. Tracked as #1067 with both options you named — an alertable fallback-with-config-present signal, or a synthetic check that asserts the forwarded address round-trips — plus a docs task for `docs/guide/fastly.md`.

docs/superpowers/specs/2026-08-19-trusted-client-ip-header-design.md
```diff
@@ -0,0 +1,223 @@
...
+`reject_placeholder_secrets` startup gate also
+rejects the placeholder secret published in the example configuration and
+guides, so a copied config cannot ship a publicly known secret.
+
+To ensure request entry can remove the fields without deleting a required HTTP
+field, `ip_header` must either be `fastly-client-ip` or begin with `x-`, and
+`auth_header` must begin with `x-`. Comparison is case-insensitive after parsing
+through `http::HeaderName`. The two Fastly-injected TLS bridge fields
+(`x-ts-tls-protocol` and `x-ts-tls-cipher`) are forbidden for either setting
+because the entry point owns and re-injects them after sanitization. These rules
```
ChristianPavilonis, line 70: 🔧 **P3: Design spec understates the reserved-header restriction**

The final validator rejects every name in `INTERNAL_HEADERS`, not only the two TLS bridge fields described here. A maintainer following this new design document could choose `x-forwarded-for`, `x-geo-info-available`, or `x-ts-ec` and get a startup failure even though the document appears to allow it. The public configuration guide and `trusted_client_ip_rejects_reserved_internal_headers` already describe and test the broader contract.

```suggestion
through `http::HeaderName`. Every name in `INTERNAL_HEADERS` is forbidden for
either setting. This includes `x-forwarded-for`, `x-geo-info-available`,
`x-ts-ec`, `x-ts-tls-protocol`, and `x-ts-tls-cipher`.
```

ChristianPavilonis (approved):
## Summary

Reviewed `3329711dbd7724d0c16178d2ec9d79518852ef90` against `2c1e0825d248361331441f312b7728d4c3041006`. The authenticated resolution, peer fallback, `ClientInfo` propagation, and cross-adapter sanitization paths looked sound. Approving with one low-severity documentation correction inline.

ChristianPavilonis:
@prk-Jr to confirm how vcl path works / clarify language in pr description

crates/trusted-server-adapter-axum/src/middleware.rs
```diff
@@ -7,20 +7,58 @@ use edgezero_core::http::{HeaderValue, Response};
...
+/// Outermost middleware: strips the configured client-IP trust headers from the
+/// request before any inner middleware or handler observes them.
+///
+/// Must stay the first middleware registered in [`crate::app`]. Registering
+/// another middleware ahead of it would re-expose the shared-secret
+/// authentication header to request handling. Only the Fastly adapter consumes
+/// these headers for client-IP resolution; every other adapter removes them so
+/// a shared configuration cannot leak the secret into publisher or integration
+/// request handling.
+pub struct SanitizeRequestMiddleware {
```
aram356: 👍 **praise** - Moving the strip into a dedicated middleware with the invariant stated in the doc comment and repeated at every registration site is a better resolution than the comment-only option the review offered. It also has a nice side effect on Cloudflare: the `x-geo-info-available` read in the finalize middleware now happens after sanitization instead of before it.

crates/trusted-server-adapter-fastly/src/compat.rs
```diff
@@ -63,9 +77,32 @@ pub(crate) fn sanitize_fastly_forwarded_headers(req: &mut fastly::Request) {
     }
 }
 
+/// Resolve the trusted client IP, then strip every trust and spoofable header.
+///
+/// Resolution has to observe the trust headers *before* sanitization removes
+/// them. Both steps live behind this one call so that ordering is structural
+/// rather than a convention the entry point has to remember.
+pub(crate) fn resolve_and_sanitize_client_ip(
```
aram356: 👍 **praise** - Binding resolution and sanitization behind one call so their ordering is structural rather than a convention the entry point has to remember is the right shape for a security-sensitive read-then-strip sequence, and the test proving the trust headers are read before they are removed pins it in place.

docs/guide/fastly.md
```diff
@@ -72,6 +72,180 @@ When you're ready to use your own domain:
...
+}
+```
+
+Attach an edge dictionary named `ts_private_config` to the fronting service and
+store the secret under the key `trusted_client_ip_secret`. Create the dictionary
+as write-only so the value cannot be read back through the API or the web
+interface. If the key is absent, the lookup yields no matching value, the
+authentication check fails, and Trusted Server falls back to the peer address.
+
+Four details in that example carry weight:
```
aram356: 👍 **praise** - Documenting why each VCL guard exists, including the shielding rationale for keeping the `unset` lines inside the `fastly.ff.visits_this_service` check and the reasoning for dropping the `req.restarts` guard, is what makes this example safe to adapt rather than just safe to paste. The topology table's honest disclosure of the one configuration that authenticates a wrong address instead of falling back is exactly the caveat an operator needs before enabling this.

aram356 (approved):

PR #1054, opened 2026-08-21
Add managed LiveRamp RampID integration

### LINKED ISSUE #355: Investigate and document LiveRamp integration

prk-Jr:
## Summary

- Add managed LiveRamp RampID support to Prebid so operators can configure IdentityLink centrally while preserving publisher User ID modules.
- Carry opaque `liveramp.com` EIDs through the existing auction, consent, edge-cookie, and EC/KV paths, with generated-artifact coverage for exact default TCF behavior.
- Keep ATS Direct audience segments and server-side identity resolution out of scope; live validation remains credential-gated.

## Status

Code complete; live LiveRamp validation pending IABTechLab/uid2-optout#385.

## Changes

| File | Change |
| ---- | ------ |
| `.cargo/config.toml` | Add host-target clippy aliases for the CLI and OpenRTB codegen crates. |
| `CLAUDE.md` | Include the new host-target aliases in the documented CI gates. |
| `crates/trusted-server-cli/src/prebid_bundle.rs` | Reject LiveRamp bundle configuration that explicitly omits `identityLinkIdSystem`. |
| `crates/trusted-server-core/src/consent/mod.rs` | Cover source-agnostic EID consent gating with multiple identity sources. |
| `crates/trusted-server-core/src/ec/prebid_eids.rs` | Verify opaque LiveRamp envelopes survive bounded edge-cookie persistence. |
| `crates/trusted-server-core/src/integrations/prebid.rs` | Add typed LiveRamp configuration, validation, defaults, and browser injection. |
| `crates/trusted-server-js/lib/build-prebid-external.mjs` | Bundle IdentityLink and `tcfControl` with generated Prebid artifacts. |
| `crates/trusted-server-js/lib/src/integrations/prebid/index.ts` | Install and protect the operator-managed IdentityLink entry while preserving publisher entries. |
| `crates/trusted-server-js/lib/test/build-prebid-external.test.mjs` | Cover generated bundle metadata and explicit IdentityLink selection. |
| `crates/trusted-server-js/lib/test/integrations/prebid/index.test.ts` | Cover ownership, queue ordering, API guards, EID transport, privacy, and idempotence. |
| `crates/trusted-server-js/lib/test/integrations/prebid/user_id_modules.test.ts` | Cover the LiveRamp module mapping and default preset. |
| `crates/trusted-server-js/lib/test/prebid-artifact-integration.test.mjs` | Exercise the real bundle and shim, including partial `userSync` updates and auction EID output. |
| `crates/trusted-server-js/lib/test/prebid-consent-enforcement.test.mjs` | Verify Purpose 1, Purposes 3/4, and LiveRamp vendor consent independently against generated artifacts. |
| `docs/guide/configuration.md` | Document LiveRamp settings, defaults, and validation constraints. |
| `docs/guide/integrations/prebid.md` | Document setup, ownership limits, exact TCF defaults, network/storage effects, failure modes, and live validation. |
| `docs/superpowers/plans/2026-08-21-liveramp-integration.md` | Record the reviewed implementation, correction, and verification plan. |
| `docs/superpowers/specs/2026-08-21-liveramp-integration-design.md` | Specify architecture, consent behavior, security boundaries, tests, and excluded ATS Direct scope. |
| `trusted-server.example.toml` | Add an example managed LiveRamp configuration. |

## Closes

Closes #355

## Test plan

- [x] `cargo clippy-fastly && cargo clippy-axum && cargo clippy-cloudflare && cargo clippy-cloudflare-wasm && cargo clippy-spin-native && cargo clippy-spin-wasm && cargo clippy-cli && cargo clippy-codegen`
- [x] JS build: `cd crates/trusted-server-js/lib && node build-all.mjs`
- [x] JS lint/format: `cd crates/trusted-server-js/lib && npm run lint && npm run format`
- [x] CLI tests: `./scripts/test-cli.sh`
- [x] Other: generated Prebid artifact tests cover managed configuration and the independent TCF matrix.

## Hardening note

This change does not add config-derived regular-expression or pattern compilation. Invalid enabled LiveRamp configuration is rejected by typed validation and covered by Rust and CLI regression tests.

## Checklist

- [x] Changes follow [CLAUDE.md](/CLAUDE.md) conventions.
- [x] No new `unwrap()` calls were introduced in production code.
- [x] No direct stdout logging was introduced in production code.
- [x] New behavior has automated coverage.
- [x] No secrets, credentials, live Placement IDs, or RampID envelope values are committed.

## Acceptance boundary

RampID identity envelopes traverse the existing Prebid auction path; ATS Direct audience segments are not passed by this implementation.


jwrosewell:
Flagging a design-guidance conflict before this goes further, with a concrete way to resolve it.

This PR puts a named vendor into `trusted-server-core`. In the current diff, `crates/trusted-server-core/src/integrations/prebid.rs` gains `default_liveramp_expires_days()`, `default_liveramp_refresh_in_seconds()`, and `validate_liveramp_placement_id()`; `crates/trusted-server-core/src/consent/mod.rs` and `crates/trusted-server-core/src/ec/prebid_eids.rs` gain `liveramp.com` handling in the consent and Edge Cookie identity paths.

That runs against the project's own provider guidance. The pluggable providers design spec (`docs/superpowers/specs/2026-07-30-pluggable-providers-design.md`, on `specs/providers-and-permissions`) states as a goal that "a deployment picks an implementation per concern (including none) **without a code change to Trusted Server core**", and as a non-goal that "**no vendor provider ships in this epic** beyond the host-platform implementations named below". The same principle is what #838 was asked to satisfy: vendor implementations live in their own crates and are injected by the adapter, so core never names a vendor.

The awkward part is the sequencing. The seam that would host this cleanly is sitting reviewed-ready in #1043 (the pluggable Edge Cookie provider seam) and #1045 (the permission model), which restructure the very files this PR edits: `consent/mod.rs`, `ec/prebid_eids.rs`, and the EC identity path. Whichever lands second pays for the collision.

Suggested way forward, which I think helps this PR rather than blocks it:

1. Merge #1043 first. It is green, mergeable, and adds `[ec] provider` with `[ec.providers.<key>]` configuration blocks plus adapter injection.
2. Re-express the RampID work as a vendor provider in its own crate, selected by configuration, with the `liveramp.com` specifics out of `trusted-server-core`. The EID plumbing that is genuinely generic (opaque envelope pass-through, partner registry handling) stays in core, vendor-neutral.
3. That leaves operators able to enable RampID by configuration alone, and leaves core free of any single vendor's naming, which is the outcome the spec asks for.

Happy to do the adaptation work myself if that is useful: the provider seam is mine and I can put up the RampID-as-provider version against #1043 quickly.

Raised with AI assistance under my direction; the file and line references above were checked against the current diff.

jevansnyc:
Noted James, thanks. Let us take a look at this and hold off on the merge for now.

jwrosewell:
The vendor-neutral `managed_user_ids` surface addresses the point I raised. Core no longer names a vendor, and RampID is reachable by configuration alone. Thank you for turning that round quickly.

One consequence of the rework deserves a second look before this merges.

`2f89a222` removes the `ts prebid bundle` guard that cross-checked the managed entry against `identityLinkIdSystem`, deleting 98 lines from `crates/trusted-server-cli/src/prebid_bundle.rs`. The stated reason is that a name-to-module table in the CLI would be vendor knowledge duplicating the registry the bundle generator already reads, and would go stale against newly released Prebid modules.

That reason does not hold, because the table already exists and it is not vendor knowledge. `crates/trusted-server-js/lib/src/integrations/prebid/user_id_modules.json` already carries exactly the mapping a check would need.

```json
{
  "moduleName": "identityLinkIdSystem",
  "configNames": ["identityLink"],
  "eidSources": ["liveramp.com"],
  "importPath": "prebid.js/modules/identityLinkIdSystem.js"
}
```

A check can resolve each `managed_user_ids[].name` against `configNames` across that registry, then confirm the resolved `moduleName` appears in the bundle manifest. Section 4 of your own design spec notes the manifest already identifies which User ID modules were compiled into the bundle, so both inputs are present. That is a join between two sources the project already maintains. No vendor name enters the CLI. It cannot go stale against newly released Prebid modules either, because it reads the same file the bundle generator reads, so adding a module to the registry extends the check automatically.

Without the guard, an operator who configures `name = "identityLink"` against a bundle built without `identityLinkIdSystem` gets no error at build, none at startup, and nothing from Trusted Server at runtime. The module is never installed, no EIDs appear, and nothing reports why. That failure is silent and it gets discovered in production.

There is also an inconsistency worth resolving. #1043 fails at startup when a selected provider has no configuration block, when a configured block is left unreferenced, and when a block carries an unknown key, on the stated grounds that misconfiguration should be loud. This PR documents the equivalent pairing as unvalidated operator responsibility. I do not think both positions can be right in the same identity area.

Suggested resolution, which leaves your design untouched. Restore the guard against the registry rather than against a vendor name. The configuration surface stays exactly as you have it, and the operator still owns the choice. They just find out at build time instead of in production.

Raised with AI assistance under my direction. The commit, file, and registry references above were checked against the current diff.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -356,6 +446,144 @@ export function auctionBidsToPrebidBids(
...
+  managedUserIds: InjectedManagedUserId[] | undefined,
+  setConfig: typeof pbjs.setConfig,
+  getConfig: PrebidGetConfig | undefined
+): ManagedTcfConsentActivation | undefined {
+  const tcfApi =
+    typeof window === 'undefined' ? undefined : (window as { __tcfapi?: unknown }).__tcfapi;
+  if (
+    !managedUserIds?.length ||
+    typeof window === 'undefined' ||
+    typeof tcfApi !== 'function' ||
```
ChristianPavilonis, line 705: **[P1] Do not seed managed IDs before a late-loading CMP is discoverable**

TCF activation performs a one-time check for `window.__tcfapi`. If an asynchronous CMP installs it after the deferred shim runs, managed IDs are seeded with Prebid’s GDPR handler disabled. Using the generated artifacts, installing a Purpose-1-denying CMP immediately after the shim still produced a request to `api.rlcdn.com` without TCF parameters. Defer managed-ID initialization until CMP discovery completes, or reconfigure/fail closed when the API appears.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -356,6 +446,144 @@ export function auctionBidsToPrebidBids(
...
+
+  if (effectiveConsentManagement !== undefined && !isRecord(effectiveConsentManagement)) {
+    log.error('[tsjs-prebid] effective consentManagement configuration is not mergeable');
+    return undefined;
+  }
+
+  const activation: ManagedTcfConsentActivation = { acceptCmpEvents: true };
+  try {
+    const effectiveConsent = effectiveConsentManagement ?? {};
+    if (Object.prototype.hasOwnProperty.call(effectiveConsent, 'gdpr')) {
```
ChristianPavilonis: **[P1] Treat Prebid’s legacy consent shape as publisher-owned**

Prebid 10.26 supports legacy TCF configuration directly under `consentManagement`, such as `{ cmpApi: "static", consentData: ... }`. Because this check only recognizes an own `gdpr` property, the shim appends `gdpr: { cmpApi: "iab" }`; Prebid then interprets the object as the modern shape and ignores the publisher’s top-level static consent. A generated-artifact reproduction with legacy static consent denying Purpose 1 and `__tcfapi` granting it made the LiveRamp request. Recognize legacy TCF configuration here and in `publisherClaimsGdprOwnership`.

crates/trusted-server-cli/src/prebid_bundle.rs
```diff
@@ -120,11 +122,85 @@ fn npm_prebid_bundle_args(request: &PrebidBundleGenerateRequest) -> Vec<String>
...
+        .map(|config_name| {
+            let mut candidates = registry
+                .modules
+                .iter()
+                .filter(|entry| entry.config_names.iter().any(|name| name == config_name))
+                .map(|entry| entry.module_name.clone())
+                .collect::<Vec<_>>();
+            candidates.sort();
+            candidates.dedup();
+
```
ChristianPavilonis, line 224: **[P2] Reject multiple aliases resolving to the same User ID module**

Each managed name is accepted independently even when multiple names resolve to one module. The registry maps both `sharedId` and `pubCommonId` to `sharedIdSystem`, so configuring both passes validation. Prebid registers one submodule with those name/alias values and selects the first matching configuration, silently ignoring the other managed entry. Reject duplicate resolved `module_name` values or canonicalize aliases before generation.

ChristianPavilonis (changes requested):
## Summary

Requesting changes for two consent-enforcement regressions and one managed User ID validation issue.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
prk-Jr: Confirmed and fixed in d269ff8f.

Managed entries now stay out of every configuration Prebid sees until CMP discovery concludes, rather than seeding and hoping to reconfigure later. Discovery resolves on whichever comes first:

- `window.__tcfapi` is already callable at shim time — activate and seed synchronously, as before.
- A CMP assigns `window.__tcfapi` later — `watchForLateTcfApi` installs a configurable accessor, restores the plain property on the first function assignment, then activates and seeds. If a CMP redefines the property outright, the descriptor check leaves it alone and the fallback below still reads the live API.
- No CMP by the first `requestBids` — a conforming CMP installs its stub before vendor tags request bids, so that is the last useful moment to conclude none is coming. Managed entries seed there with no TCF activation, which is the correct posture for a page that genuinely has no CMP.

While discovery is pending, `normalizePublisherConfig` passes publisher config through untouched; managed entries are merged onto the effective configuration once it resolves, so nothing the publisher set is lost.

Two regression tests cover it (`defers managed User ID seeding until a late CMP installs __tcfapi`, `seeds managed User IDs at the first auction when no CMP ever appears`), and `prebid-artifact-integration.test.mjs` now asserts the deferral against the real generated bundle before settling it. Both fail against the previous behaviour.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
prk-Jr: Confirmed and fixed in d269ff8f.

Verified against the pinned artifact: `modules/consentManagementTcf.ts:200` reads the config as `config.gdpr || config.usp || config.gpp ? config.gdpr : config`, so a legacy object carries no namespace key and appending `gdpr` demotes every legacy setting.

Ownership detection now mirrors that rule rather than probing for an own `gdpr` property, and is shared by `activateManagedUserIdTcfConsent` and `publisherClaimsGdprOwnership`:

- A truthy own `gdpr`, `usp`, or `gpp` selects the namespaced shape; the publisher owns TCF only when `gdpr` itself is truthy. A `usp`-only or `gpp`-only config still gets the automatic collector, since Prebid would leave TCF unconfigured there.
- Any other non-empty object is a legacy top-level TCF config and is left alone.
- Namespace probing uses `hasOwnProperty` deliberately, so a `consentManagement` object whose descriptors cannot be read still throws into the existing fail-safe path instead of being silently treated as empty.

Recognizing the legacy shape exposed a second defect on the retirement path that I fixed alongside it: `mergeConfig` deep-merges onto the current config, so retiring with `gdpr: { enabled: false }` survives a legacy publisher merge and Prebid then reads `{ enabled: false }` and disables the TCF module entirely — fail-open, and worse than the original bug. `removeAutomaticGdprNamespace` now drops the key with a `setConfig` topic replacement after the merge, so the publisher’s legacy config governs.

Three regression tests: `leaves a legacy top-level TCF consent configuration to the publisher`, `retires automatic IAB consent when a late setConfig uses the legacy TCF shape`, `drops the automatic gdpr namespace when a merge uses the legacy TCF shape`. All three fail against the previous behaviour. `activates managed GDPR consent alongside a publisher USP namespace` guards the non-regression.

crates/trusted-server-cli/src/prebid_bundle.rs
prk-Jr: Confirmed and fixed in d269ff8f.

The checked-in registry has exactly one such module today: `sharedIdSystem` <- `sharedId`, `pubCommonId`. `resolve_managed_user_id_modules` now runs `reject_managed_user_id_module_collisions` over the resolved set and fails before generation, naming both managed names, the shared module, and the registry path. Rejecting rather than canonicalizing, since collapsing two entries would silently pick one operator configuration over the other — the same ambiguity, just moved earlier.

Two existing tests encoded the old behaviour and are now rejection tests:

- `managed_names_resolve_aliases_to_registered_modules` -> split into `managed_name_resolves_an_alias_to_its_registered_module` (single alias still resolves) and `managed_names_sharing_one_module_are_rejected`.
- `run_bundle_accepts_two_aliases_backed_by_one_module` -> `run_bundle_rejects_two_aliases_backed_by_one_module`, which also asserts nothing is generated and the configured bundle hash and SRI are left untouched.

Docs updated in `docs/guide/integrations/prebid.md` and `docs/guide/configuration.md`.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -419,6 +532,14 @@ pub struct PrebidIntegrationConfig {
...
     #[validate(custom(function = "validate_excluded_gam_ad_unit_path_suffixes"))]
     pub excluded_gam_ad_unit_path_suffixes: Vec<String>,
+    /// Prebid User ID modules that Trusted Server installs and keeps installed.
+    ///
+    /// Each entry is forwarded to Prebid.js verbatim; publisher-configured
+    /// entries with other names are preserved. Names must be unique, and no two
+    /// names may resolve to the same Prebid User ID submodule.
+    #[serde(default)]
+    #[validate(nested, custom(function = "validate_unique_managed_user_id_names"))]
+    pub managed_user_ids: Vec<PrebidManagedUserIdConfig>,
```
ChristianPavilonis, line 547: :wrench: **[P1] Production head injection drops every managed User ID**

This field is parsed, but the production path never serializes it. `register_for_plan` calls `for_browser_plan`, which stores `head_inserts_for_plan` output in `planned_head_inserts`; that payload has no `managed_user_ids`. `head_inserts` then returns the planned output before reaching the new managed-ID serializer. As a result, `window.__tsjs_prebid.managedUserIds` is absent in a running service, so the browser shim installs nothing even when configuration and bundle generation succeed.

Add managed IDs to the `head_inserts_for_plan` payload, preferably through one shared payload builder, and add a regression test through `register_for_plan` or `IntegrationRegistry::with_plan`.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -1190,6 +1783,11 @@ export function installPrebidNpm(config?: Partial<PrebidNpmConfig>): typeof pbjs
   // every unowned bidder is left untouched.
   pbjs.requestBids = function (requestObj?: Parameters<typeof originalRequestBids>[0]) {
     log.debug('[tsjs-prebid] requestBids called');
+    // A conforming CMP installs `__tcfapi` before vendor tags request bids, so
+    // an auction is the last useful moment to conclude that none is coming.
+    const settleCmpDiscovery = resolveManagedUserIdCmpDiscovery;
+    resolveManagedUserIdCmpDiscovery = undefined;
+    settleCmpDiscovery?.();
```
ChristianPavilonis: :wrench: **[P1] The first auction ends CMP discovery too early**

If `__tcfapi` is absent when the first `requestBids` runs, this settles discovery and seeds managed IDs without TCF. An asynchronously loaded CMP can appear moments later, but the watcher is already gone and consent activation never occurs. Against the generated bundle and shim at this head, I queued `requestBids`, installed a Purpose-1/vendor-denying CMP 25 ms after shim evaluation, and observed one request to `api.rlcdn.com`, an `idl_env` cookie, and no effective `consentManagement`.

Let that auction proceed without managed IDs while discovery remains unresolved. Seed only after a callable TCF API, existing publisher-owned consent configuration, or an explicit operator policy confirming TCF does not apply. Add a generated-artifact test with a queued auction followed by late CMP installation.

ChristianPavilonis (changes requested):
## Review summary

Reviewed `d269ff8f9e3c78e7f024f29cd349739723279f51` against `705a69505bf6bf1eb5b498a88606d1473fdc6ea1`. The bundle validation and focused tests pass, but the production Rust injection path drops the new configuration, and the first-auction CMP cutoff still permits managed identity resolution before a late CMP can enforce denied consent.

crates/trusted-server-core/src/integrations/prebid.rs
prk-Jr: Fixed in 66da8064. Both planned and legacy injection now use the same browser payload builder, including managed User ID parameters and camelCase storage fields. Planned routing still supplies serverSideBidders separately.

Added a regression through register_for_plan and its registered head injector, covering nested parameters, optional fields, refreshInSeconds, script escaping, and managed IDs with auction routing disabled. The regression failed against the previous production payload and now passes. All required Rust test and Clippy gates pass.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
prk-Jr: Fixed in 66da8064. requestBids no longer concludes discovery from an absent CMP, and failure to install the property watcher no longer seeds IDs. Auctions continue with managed IDs deferred until a callable TCF API or publisher-owned TCF configuration is available; setConfig, mergeConfig, and subsequent auctions recheck readiness.

Late seeding preserves the full publisher userSync configuration and uses Prebid's autoRefresh for that seed, restoring the publisher's refresh policy afterward. This also handles Prebid initialization completing before CMP arrival.

The generated-artifact regression queues an auction and installs a Purpose-1/vendor-denying CMP 25 ms after shim evaluation. It reproduced the vendor request before the fix; now it verifies no identity request or storage and active consentManagement. Positive tests verify late-granted consent before and after the first auction. All 967 JS tests, bundle build, and formatting checks pass (local JS tests used Node 25 with native web storage disabled because the installed jsdom dependency fails to load under pinned Node 24).

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -361,6 +451,309 @@ export function auctionBidsToPrebidBids(
...
+      return originalTcfApi.call(window, command, version, guardedCallback, parameter);
+    };
+
+    const tcfWindow = window as typeof window & { __tcfapi: TcfApi };
+    tcfWindow.__tcfapi = guardedTcfApi;
+    try {
+      setConfig({
+        consentManagement: {
+          ...effectiveConsent,
+          gdpr: { cmpApi: 'iab' },
```
ChristianPavilonis, line 773: **[P1] Fail closed when the automatic CMP lookup times out**

The automatic configuration sets only `cmpApi: 'iab'`. Prebid defaults `defaultGdprScope` to `false`, so if a callable CMP bootstrap stub never returns consent, its 10-second timeout installs null consent as `gdprApplies: false`. `tcfControl` then allows the managed ID module to run.

Against the generated bundle at `66da8064d81f3ba9da969eb048857933743234c8`, an unresponsive `__tcfapi` produced a request to `api.rlcdn.com` without TCF parameters and wrote `idl_env`, retry, source, and `ts-eids` cookies after the timeout. This makes a stalled CMP equivalent to consent. Configure the automatic policy to fail closed and add a generated-artifact regression covering an unresponsive callable stub.

```suggestion
          gdpr: { cmpApi: 'iab', defaultGdprScope: true },
```

ChristianPavilonis (dismissed):
## Summary

Reviewed `66da8064d81f3ba9da969eb048857933743234c8`. The late-CMP changes handle delayed API discovery, but the automatic TCF policy still fails open when a callable CMP stub does not return consent before Prebid's timeout.

ChristianPavilonis (commented):
@prk-Jr the changes in prebid/index.ts seem to have privacy/demand implications that I'm not sure about so the following is worth calling out and perhaps needs product decision.

## Consent-timeout policy worth discussing

I dismissed my earlier change request because this is better treated as a product/privacy policy decision than an unconditional implementation defect. The reproduction still stands: when `__tcfapi` is callable but never returns consent, Prebid's default 10-second timeout produces null consent with `gdprApplies: false`; the managed IdentityLink module can then call `api.rlcdn.com` without TCF parameters and write its identity cookies.

There are several defensible policies:

### 1. Leave the current behavior unchanged

**Pros**

- Preserves Prebid's standard timeout semantics.
- Avoids suppressing demand or analytics during a CMP outage.
- Requires no additional lifecycle/state handling.

**Cons**

- A stalled CMP is treated like GDPR does not apply, even though the callable API indicates that a CMP was expected.
- An operator-managed identity module may contact its vendor and write storage without an affirmative jurisdiction or consent result.
- This failure mode should be explicitly documented so operators understand the compliance tradeoff.

### 2. Set `defaultGdprScope: true` for the automatic TCF configuration

**Pros**

- Small, direct change that fails closed after the timeout.
- Responsive CMP behavior is unchanged: `gdprApplies: false` still permits activity, while valid grants and denials are enforced normally.
- Prevents managed identity resolution when consent state is unavailable.

**Cons**

- The blast radius is broader than managed User IDs. `tcfControl` can also block client-side bidders, analytics, EID transmission, user data, and other controlled Prebid activity.
- A broken CMP could therefore reduce or eliminate monetization, including for non-EU users if the CMP never reports `gdprApplies: false`.
- The page remains restricted until a later CMP callback updates the consent handler.

### 3. Defer only managed User ID seeding until a terminal CMP result

Seed managed entries only after `gdprApplies: false`, `eventStatus: 'tcloaded'`, or `eventStatus: 'useractioncomplete'`.

**Pros**

- Gives operator-managed IDs a fail-closed guarantee without changing Prebid's timeout policy for publisher bidders and analytics.
- Keeps the monetization impact limited to the feature being introduced here.
- A later valid CMP event can still activate the managed entries.

**Cons**

- Requires more state coordination than the one-line scope change, especially for synchronous callbacks, late publisher ownership, listener retirement, and recovery after the first auction.
- Managed IDs remain unavailable for the page lifetime if the CMP never reaches a terminal state.
- Needs generated-artifact coverage for timeout, late recovery, denial, and `gdprApplies: false`.

### 4. Make the timeout posture operator-configurable

For example, expose a managed-ID consent policy selecting Prebid-default, global fail-closed, or managed-only deferral behavior.

**Pros**

- Makes the compliance/revenue decision explicit for each deployment.
- Supports operators with different jurisdiction and CMP guarantees.

**Cons**

- Expands configuration and documentation substantially for a narrow feature.
- A permissive default can still surprise operators; a restrictive default can still surprise them operationally.
- Adds more combinations that require browser-artifact testing.

My preference would be option 3 if managed identities need a stronger guarantee without changing the rest of the publisher's Prebid failure behavior. Leaving the current behavior can also be intentional, but the timeout semantics and resulting vendor/storage activity should then be documented explicitly.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -192,29 +217,89 @@ export function collectBidders(adUnits: Array<{ bids?: Array<{ bidder?: string }
...
-    ...new Set(
-      userIds
-        .map((entry) => entry?.name)
-        .filter((name): name is string => typeof name === 'string' && name.length > 0)
-    ),
-  ].sort();
+  const managedNames = new Set(managedUserIds.map((managed) => managed.name));
+  const retained = configuredUserIdEntries(config.userSync.userIds).filter(
+    (entry) => !managedNames.has(entry.name)
+  );
```
aram356: 🔧 **wrench** — Case-variant publisher entry silently defeats managed User ID ownership.

Prebid matches `userSync.userIds` entry names to submodules **case-insensitively** and takes the **first** matching config entry (pinned prebid.js, `modules/userId/index.ts:1069-1070`: `value?.toLowerCase() === query.toLowerCase()` followed by `configs.find(...)`). This filter compares names **case-sensitively**, and retained publisher entries are placed **before** managed entries. So `pbjs.setConfig({userSync: {userIds: [{name: 'IdentityLink', params: {pid: 'publisher-value'}}]}})` survives the filter, lands ahead of the managed `identityLink` entry, and Prebid's `configs.find` selects the publisher's entry for `identityLinkIdSystem` — the operator-managed configuration is silently ignored. That breaks the documented invariant that calls which add, remove, or replace a managed name are normalized back to the operator-managed values.

```suggestion
  const managedNames = new Set(managedUserIds.map((managed) => managed.name.toLowerCase()));
  const retained = configuredUserIdEntries(config.userSync.userIds).filter(
    (entry) => !managedNames.has(entry.name.toLowerCase())
  );
```

Scratch-verified: prettier check, full vitest suite (967/967, including the real-bundle ownership tests), and `node build-all.mjs` all pass with this applied. A regression test asserting that a case-variant publisher entry (`IdentityLink`) is filtered would pin this down; it lives outside this diff range, so it is left to you to add.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -222,6 +221,113 @@ fn extract_prebid_error_message(
...
+///
+/// Prebid keys `userSync.userIds` by entry name, so two entries sharing a name
+/// give one submodule two conflicting configurations with no defined winner.
+fn validate_unique_managed_user_id_names(
+    entries: &[PrebidManagedUserIdConfig],
+) -> Result<(), ValidationError> {
+    let mut seen = HashSet::with_capacity(entries.len());
+    let Some(duplicate) = entries
+        .iter()
+        .find(|entry| !seen.insert(entry.name.as_str()))
```
aram356: ♻️ **refactor** — Duplicate-name validation should use Prebid's case-insensitive matching rule.

This validator uses exact string comparison, so `name = "exampleId"` and `name = "exampleid"` both pass core validation — yet both address the same Prebid submodule, and Prebid takes the first matching entry with no defined winner. The `ts prebid bundle` registry check catches the pair only when the operator runs the CLI (and its exact-match lookup rejects a case variant as "unknown" rather than as a duplicate); runtime config validation alone lets the pair through to the browser. The doc on `PrebidIntegrationConfig.managed_user_ids` already promises that no two names may resolve to the same submodule. Names are validated ASCII-only, so `to_ascii_lowercase` matches Prebid's rule exactly.

```suggestion
/// Rejects a managed User ID list whose names address one module twice.
///
/// Prebid matches `userSync.userIds` entry names to submodules
/// case-insensitively and takes the first matching entry, so two entries whose
/// names differ only by case give one submodule two conflicting configurations
/// with no defined winner.
fn validate_unique_managed_user_id_names(
    entries: &[PrebidManagedUserIdConfig],
) -> Result<(), ValidationError> {
    let mut seen = HashSet::with_capacity(entries.len());
    let Some(duplicate) = entries
        .iter()
        .find(|entry| !seen.insert(entry.name.to_ascii_lowercase()))
```

Scratch-verified against the full CI-equivalent gate: `cargo fmt --all -- --check`, all six adapter clippy aliases, `cargo test-fastly` / `test-axum` / `test-cloudflare` / `test-spin`, and the cross-adapter parity suite all pass with this applied. A `managed_user_ids_reject_a_case_variant_module_name` test would cover the new rejection path; it lives outside this suggestion range.

crates/trusted-server-core/src/ec/prebid_eids.rs
```diff
@@ -800,6 +800,29 @@ mod tests {
         );
     }
 
+    #[test]
+    fn ingest_liveramp_eid_cookie_preserves_the_opaque_envelope() {
```
aram356, line 804: 🤔 **thinking** — Real RampID envelopes vs the 512-byte EC value cap.

This test proves a 20-byte placeholder envelope survives ingestion, but `MAX_UID_LENGTH` is 512 bytes (`kv_types.rs:35`) and real ATS envelopes routinely run several hundred characters — some exceed 512. An oversized envelope is silently dropped at EC persistence (the auction path is unaffected), which is a degradation the docs' "Degraded behavior" table does not list. Two cheap follow-ups: measure a real envelope's length during the credential-gated live validation (one extra bullet in its step list), and/or add an "envelope exceeds the EC value cap" row to the degraded-behavior table.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -1120,6 +1513,234 @@ export function installPrebidNpm(config?: Partial<PrebidNpmConfig>): typeof pbjs
...
+      }
+      if (tcfApiInstalled || publisherConsentConfigured) activateAndSeedManagedUserIds();
+    };
+    trySeedManagedUserIds();
+    if (managedUserIdsDeferred) {
+      awaitingLateConsent = true;
+      // Auctions may run before an asynchronous CMP arrives. Keep managed IDs
+      // deferred even when the property cannot be watched; later configuration
+      // and auction calls recheck discovery without treating absence as consent.
+      watchForLateTcfApi(trySeedManagedUserIds);
```
aram356: 🤔 **thinking** — `watchForLateTcfApi` accessor is never dismantled on the publisher-config seeding path.

`watchForLateTcfApi` returns its `settle` restorer, but this call site discards it. When seeding later resolves through publisher consent configuration (no CMP ever appearing), the get/set accessor pair stays installed on `window.__tcfapi` for the life of the page. Behavior is preserved for readers, and a late CMP assignment still settles correctly, but the accessor-typed property is observable (for example via `Object.getOwnPropertyDescriptor`) and could surprise CMP stubs or diagnostics that expect a plain data property. Worth either capturing the settle function and calling it once seeding concludes through another path, or a short comment stating the residue is accepted.

crates/trusted-server-core/src/integrations/prebid.rs
```diff
@@ -222,6 +221,113 @@ fn extract_prebid_error_message(
 #[cfg(test)]
 const GPC_US_PRIVACY: &str = "1YYN";
 
+/// Rejects a Prebid User ID identifier that Prebid.js could not address.
+///
+/// Applies only the constraints Prebid itself imposes on a `userSync.userIds`
+/// entry name and on a storage key: a non-empty, untrimmed-free ASCII token.
+/// Anything narrower would encode one vendor's rules into core.
```
aram356: ⛏ **nitpick** — Garbled doc phrase "untrimmed-free ASCII token".

```suggestion
/// entry name and on a storage key: a non-empty ASCII token with no
/// surrounding whitespace. Anything narrower would encode one vendor's rules
/// into core.
```

Scratch-verified: `cargo fmt --all -- --check` and `cargo clippy-fastly` pass (pure comment change). For awareness only: the `value.trim() == value` clause below is technically redundant with the byte-class check, since whitespace is outside the allowed byte set; fine to keep as belt-and-braces.

aram356 (changes requested):
## Summary

High-quality implementation of vendor-neutral managed Prebid User ID modules with LiveRamp RampID as the worked example. The fail-closed CMP-discovery machinery is carefully built, and the test depth (real-bundle TCF purpose/vendor matrix, ownership-transfer races, hostile-proxy edge cases, fail-closed CLI registry validation) is unusually strong. One correctness gap in the ownership invariant drives the verdict; the remaining findings are non-blocking.

> 3 of the 5 inline comments below carry a one-click GitHub `suggestion` (all scratch-verified against this head: `cargo fmt` + all six adapter clippy aliases + all four adapter test aliases + parity suite for the Rust changes; prettier + full vitest suite (967/967) + `node build-all.mjs` for the TypeScript change; verified in isolation and as a batch). Use **Commit suggestion** or **Add suggestion to batch** to apply them. The other two comments are observations with the reasoning in prose.

## Blocking

### 🔧 wrench

- **Case-variant publisher entry silently defeats managed User ID ownership** — see inline at `crates/trusted-server-js/lib/src/integrations/prebid/index.ts:292`

## Non-blocking

### ♻️ refactor

- **Duplicate-name validation should use Prebid's case-insensitive matching rule** — see inline at `crates/trusted-server-core/src/integrations/prebid.rs:284`

### 🤔 thinking

- **Real RampID envelopes vs the 512-byte EC value cap** — see inline at `crates/trusted-server-core/src/ec/prebid_eids.rs:804`
- **`watchForLateTcfApi` accessor is never dismantled on the publisher-config seeding path** — see inline at `crates/trusted-server-js/lib/src/integrations/prebid/index.ts:1740`

### ⛏ nitpick

- **Garbled doc phrase "untrimmed-free ASCII token"** — see inline at `crates/trusted-server-core/src/integrations/prebid.rs:227`

## CI Status

- integration tests: PASS
- vitest: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- cargo test: PASS (required)
- cargo fmt: PASS (required)
- Analyze (actions): PASS

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
prk-Jr: Confirmed and fixed, with one extension.

Verified the mechanism against the pinned bundle: `modules/userId/index.ts:1069` builds `matchesName` as `[name, aliasName].some(value => value?.toLowerCase() === query.toLowerCase())`, and `configs.find` takes the first match. A retained `IdentityLink` ahead of the managed `identityLink` does win.

The extension is that the same rule matches on `aliasName` too, and one module in our registry has a live alias: `sharedIdSystem` declares `aliasName: 'pubCommonId'`, which `user_id_modules.json` already records as a second `configNames` entry. So a publisher `pubCommonId` entry defeats a managed `sharedId` exactly the way a case variant does. Lowercasing alone would have left that half open, so ownership now filters on the module's full name set:

- new `userIdConfigNameAliases(configName)` in `user_id_modules.ts` resolves a config name to every lowercased `configNames` entry for its module
- `withManagedUserIds` builds its filter set from that and compares retained entry names lowercased

Regression tests in `index.test.ts`, each verified red at `66da8064` and green with the fix:

- `filters a case-variant publisher entry that Prebid would match to the managed module` — publisher supplies `IdentityLink` and `IDENTITYLINK`
- `filters a publisher entry naming a managed module by its Prebid alias` — managed `sharedId`, publisher `pubCommonId`
- `normalizes a later case-variant publisher identityLink update` — post-install `setConfig`

The `ts prebid bundle` registry check had the same exact-match assumption; see the reply on `prebid.rs:284` for that half.

The ownership paragraph in `docs/guide/integrations/prebid.md` now states Prebid's matching rule rather than implying exact string equality.

crates/trusted-server-core/src/integrations/prebid.rs
prk-Jr: Applied as suggested, plus two follow-ons.

`managed_user_ids_reject_a_case_variant_module_name` sits next to the existing repeated-name test: `name = "exampleId"` plus `name = "exampleid"` must fail validation. Verified red with `entry.name.as_str()` restored, green with `to_ascii_lowercase()`.

I did deviate from the suggestion in one small way. For an exact duplicate the message reads fine, but for a case variant `managed Prebid User ID module \`exampleid\` is configured more than once` prints one spelling and does not look repeated against the operator's TOML, so the check would be hard to act on. The message now ends with `(names are matched case-insensitively)`.

Second: you noted the `ts prebid bundle` registry check rejects a case variant as unknown rather than as a duplicate. That is fixed too. `resolve_managed_user_id_modules` now compares with `eq_ignore_ascii_case`, so a case variant resolves to the module Prebid would actually select, and `reject_managed_user_id_module_collisions` reports it as a module collision naming the shared module instead of "not registered in …". Two tests: `managed_name_resolves_a_case_variant_to_its_registered_module` and `managed_names_differing_only_by_case_are_rejected_as_a_module_collision`.

Worth recording for the record: the sibling doc on `PrebidIntegrationConfig::managed_user_ids` promises that no two names may resolve to the same submodule, and that promise *is* kept — by `reject_managed_user_id_module_collisions` in the CLI, which resolves through `user_id_modules.json` and already rejects `sharedId` + `pubCommonId`. Core deliberately keeps treating names as opaque, since it has no module registry.

crates/trusted-server-core/src/ec/prebid_eids.rs
prk-Jr: Took both follow-ups, with one correction to the premise: the auction path is not unaffected.

`MAX_UID_LENGTH` also gates the client EID path into `/auction`. `parse_client_auction_uid` drops a uid longer than the cap (`auction/endpoints.rs:542`) and `merge_auction_eids` drops it again on merge (`endpoints.rs:588`). So an oversized envelope never reaches Prebid Server either — it is not only a persistence degradation.

Changes in `docs/guide/integrations/prebid.md`:

- Degraded-behavior row: `The resolved envelope exceeds the 512-byte EID value cap` → dropped from both the `/auction` payload and EC persistence; the auction continues.
- Credential-based validation gained a step 3: record the resolved envelope's byte length — length only, never the value — and confirm it is at or below the cap. Later steps renumbered.

I did not raise the constant. It is the shared partner-UID cap across the pixel, batch, and pull sync paths plus `KvEntry` schema validation, so moving it is a KV-wide decision and not something to infer from one vendor's envelope size. If the live measurement comes back above 512, that is the data point for a tracked issue.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
prk-Jr: Captured the restorer, and the exercise turned up two further problems worth spelling out.

**The restorer itself.** `activateAndSeedManagedUserIds` now calls and clears it as soon as it flips `managedUserIdsDeferred`, so the accessor is dismantled whichever path resolves seeding. Test: `retires the late-CMP watch once publisher consent configuration seeds the IDs` — no CMP ever appears, seeding resolves through publisher consent configuration, and the test asserts the accessor is installed while deferred, gone after seeding, and that a CMP arriving afterwards lands as a plain own value. Red at `66da8064`.

**`watchForLateTcfApi` could hide a live CMP.** Its accessor pair starts with `stored === undefined`, so installing it over a `window.__tcfapi` that is already present would make every later reader see `undefined`. Not reachable at `66da8064`, because the old logic only deferred when no callable API existed — but it becomes reachable with the consent change on the other thread, where seeding can defer with a CMP present. Added an early return when `__tcfapi` is already callable: discovery has resolved in that case and there is nothing to watch for.

**A retired subscription could leak.** This one was mine, in the new consent gate, caught by a reviewer pass after I wrote it. `removeListener` needs the listener id, which TCF only delivers with a callback. If the subscription was retired before the CMP's first event — a real async CMP present, publisher consent configuration claiming ownership first — the old code set `settled` and then returned early on every later callback, so the id was never captured and `removeEventListener` was never sent at all. The CMP kept a dead listener for the page lifetime. The callback now captures the id before any early return and performs the deferred removal on the CMP's first event.

Two tests for it, both red without the fix:
- `removes a retired consent subscription once the CMP finally answers` (unit, new `pendingCmp()` helper that registers without answering — the existing `terminalCmp()` answers synchronously and structurally cannot reach this state)
- `removes its own pending consent subscription when publisher configuration takes over` (real bundle; asserts every registered subscription was retired and none left)

crates/trusted-server-core/src/integrations/prebid.rs
prk-Jr: Applied as suggested.

Kept the `value.trim() == value` clause as belt-and-braces, as you suggested.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
prk-Jr: Went with option 3 from your follow-up comment: managed User ID seeding now waits for a terminal CMP result, and Prebid's timeout semantics are left alone for everything else on the page.

First, your reproduction is exact. Against `66da8064` with a callable `__tcfapi` that never answers, waiting past Prebid's default timeout (`libraries/consentManagement/cmUtils.ts:147`, `DEFAULT_CONSENT_TIMEOUT = 10000`) produces a real request to `https://api.rlcdn.com/api/identity/envelope?pid=999`. That is now a test.

**Implementation** (`integrations/prebid/index.ts`)

- `isTerminalTcfResult` treats `gdprApplies: false`, `eventStatus: 'tcloaded'`, and `eventStatus: 'useractioncomplete'` as terminal. `cmpuishown`, a missing status, and a listener the CMP refuses are not.
- `awaitTerminalTcfConsent` subscribes with `__tcfapi('addEventListener', 2, …)`, reports once, removes its listener on the terminal event, and prefers the live API for removal so a replaced bootstrap stub does not strand it.
- A refusal (`success === false`) keeps managed IDs deferred but reopens the wait, so a later configuration or auction call can subscribe again — a TCF stub commonly gives way to the real CMP, and latching the flag would have deferred managed IDs for the page lifetime with no path back.
- `trySeedManagedUserIds` no longer treats a callable API as discovery. It seeds on a publisher-owned `consentManagement` (the publisher's own timeout posture, unchanged) or on a terminal result; otherwise it subscribes once and waits. `setConfig`, `mergeConfig`, and auction calls still recheck, and the existing late-`__tcfapi` watch still covers a CMP that installs itself after the shim.

I did not take option 2. `defaultGdprScope: true` synthesizes GDPR applicability, which contradicts what `prebid.md` already promises ("Trusted Server does not synthesize consent or GDPR applicability"), and its blast radius is the publisher's whole Prebid stack rather than the feature being introduced here.

**Generated-artifact coverage** (`prebid-consent-enforcement.test.mjs`, against the real built bundle)

- `never calls the vendor when the CMP stalls past the Prebid consent timeout` — waits 11s past the real timeout; asserts no `api.rlcdn.com` request, no `idl_env`, no `_lr_retry_request`, and no managed entry in the effective config. This is the direct reproduction; red at `66da8064` with the `api.rlcdn.com` URL in the received array.
- `never seeds a managed module while the CMP has not answered` — also asserts no automatic `cmpApi: 'iab'` claim.
- `never seeds a managed module while the CMP UI still awaits the user` — `eventStatus: 'cmpuishown'`.
- `resolves managed IDs once the user completes the CMP UI` — `cmpuishown` then `useractioncomplete`; envelope resolved, `idl_env` written.
- `resolves managed IDs when the CMP reports that GDPR does not apply` — `gdprApplies: false` with a non-terminal status, proving out-of-scope alone settles.

Denial stays covered by the existing Purpose 1 and vendor 97 tests: a settled denial still seeds the managed entry and `tcfControl` blocks the vendor call, which is the enforcement path this integration relies on.

The harness now distinguishes "the CMP never answers" from "only Prebid's own subscription is delayed" (`deferPrebidCmpResponse`), since the shim's consent gate subscribes before Prebid does. Two existing tests moved onto that: the delayed-initial-response test uses the new flag, and the listener-retirement test asserts the invariant (every subscription registered was retired, none left) instead of a bare count of 1, because this page now has two subscriptions.

Two lifecycle bugs in the above were caught by a review pass after I wrote it and are fixed: a subscription retired before the CMP's first event could never be removed (the listener id only arrives with a callback), and a refusal latched the wait permanently. Both have regression tests, red without the fix. Details on the `index.ts:1740` thread.

**Tradeoff, documented** in a new `A settled CMP result, not a callable API` section: a CMP that never settles leaves managed identity unresolved for the page lifetime; auctions, publisher bidders, and analytics are unaffected. The degraded-behavior table gained a matching row.

ChristianPavilonis (approved):
The latest changes address my consent-timeout concern by deferring managed User ID seeding until the CMP returns a terminal result, while preserving publisher-owned Prebid timeout behavior. The generated-bundle regression covers the original stalled-CMP reproduction, and the current checks pass.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -192,29 +217,96 @@ export function collectBidders(adUnits: Array<{ bids?: Array<{ bidder?: string }
...
+  managedUserIds: InjectedManagedUserId[]
+): PbjsConfig {
+  if (!hasUserIdsPath(config)) return config;
+
+  // Prebid resolves a `userSync.userIds` entry to a submodule on either its
+  // name or its alias, case-insensitively, and takes the first matching entry.
+  // A retained publisher entry sits ahead of the managed one, so it must be
+  // filtered on any spelling Prebid would resolve to the same submodule —
+  // otherwise the publisher's configuration silently wins.
+  const managedNames = new Set(
```
aram356, line 334: 🌱 **seedling** — Alias-set collisions between two managed entries still pass runtime validation.

Core's validator now rejects case-variant duplicates, and `ts prebid bundle` rejects `sharedId` plus `pubCommonId`, but an operator running a prebuilt external bundle (never invoking the CLI) can still configure two managed entries whose names share one submodule's alias set. This code appends both; Prebid takes the first matching entry and silently drops the second operator configuration, which is the same failure class the CLI gate exists to prevent. Since the alias sets are already computed here, the shim could drop-and-log a managed entry whose alias set collides with an earlier one, closing the last gap in the ownership story. Follow-up sized; not blocking this PR.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -1120,6 +1625,286 @@ export function installPrebidNpm(config?: Partial<PrebidNpmConfig>): typeof pbjs
...
+      }
+      // A publisher-owned GDPR configuration carries the publisher's own
+      // timeout posture, so seeding under it leaves Prebid's semantics alone.
+      if (publisherConsentConfigured || terminalTcfConsentSettled) {
+        activateAndSeedManagedUserIds();
+        return;
+      }
+      // Automatic TCF activation is ours, so it fails closed: subscribe once
+      // and wait for a settled CMP result rather than for a callable API.
+      if (awaitingTerminalTcfConsent) return;
```
aram356: 🤔 **thinking** — An orphaned terminal-consent subscription has no path back.

Once `awaitingTerminalTcfConsent` is set, every recheck returns here. The refusal path reopens the wait, but a non-compliant CMP stub that accepts `addEventListener` silently and never replays it to the real CMP leaves the flag armed for the page lifetime: managed IDs stay deferred even after a working CMP replaces the stub, and no later `setConfig`, `mergeConfig`, or auction call can re-subscribe. Spec-compliant stubs queue and replay, so this is a broken-stub scenario, and the failure direction is safe (no identity, no privacy exposure). Worth considering: remember which `__tcfapi` function the subscription was made against and re-subscribe when a recheck observes a different one.

aram356 (approved):
## Summary

All five findings from the previous review are resolved at this head, each with tests: alias-aware case-insensitive ownership in the shim, case-insensitive duplicate validation in core and the CLI registry lookup, the watcher restorer captured and retired on every seeding path, the envelope-cap documentation (with the corrected /auction-path premise), and the doc wording. The additional fail-closed handling of a callable-but-silent CMP closes a real gap beyond the original findings and is reproduced against the real bundle.

The two comments below are non-blocking observations for follow-up; neither affects the verdict.

## Non-blocking

### 🌱 seedling

- **Alias-set collisions between two managed entries still pass runtime validation** — see inline at `crates/trusted-server-js/lib/src/integrations/prebid/index.ts:297`

### 🤔 thinking

- **An orphaned terminal-consent subscription has no path back** — see inline at `crates/trusted-server-js/lib/src/integrations/prebid/index.ts:1862`

## CI Status

All 19 checks pass at this head, including the required set (`cargo fmt`, `cargo test`, `format-typescript`, `format-docs`) and the non-required gates (vitest, cross-adapter parity, browser and EC-lifecycle integration tests, CodeQL analyses).

docs/guide/integrations/prebid.md
```diff
@@ -506,6 +614,203 @@ Example EID source mapping:
...
+| LiveRamp cannot recognize the browser                    | IdentityLink yields no EID; the auction continues                                                  |
+| LiveRamp network resolution fails                        | The current auction continues without RampID                                                       |
+| The managed module is missing from the bundle            | Existing diagnostics report the missing module; auctions continue                                  |
+| The origin is not approved by LiveRamp                   | Resolution yields no usable EID; the auction continues                                             |
+| EC/KV is unavailable                                     | A current-request EID can still reach `/auction`; persistence degrades                             |
+| The resolved envelope exceeds the 512-byte EID value cap | The envelope is dropped from both the `/auction` payload and EC persistence; the auction continues |
+| The CMP is callable but never returns a settled result   | The managed entry is never seeded; no vendor call, no identity storage, and the auction continues  |
+
```
aram356, line 861: ⛏ **nitpick** — Degraded-behavior table lacks the two runtime behaviors this revision added.

The docs cover the CLI-time rejection of two managed names resolving to one module, but not what the runtime now does when such a pair reaches a prebuilt bundle (later entry dropped with a logged error), nor the silent-stub replacement recovery. Both are useful when debugging why managed IDs did or did not seed.

```suggestion
| The CMP is callable but never returns a settled result   | The managed entry is never seeded; no vendor call, no identity storage, and the auction continues  |
| Two managed entries address one Prebid submodule         | The later entry is dropped with a logged error; the first entry's configuration takes effect       |
| A silent CMP stub is replaced by a working CMP           | The shim re-subscribes to the replacement; a CMP that already answered once keeps its own wait     |
```

Verified: `cd docs && npm run format` passes with this applied, and the table needs no re-padding.

aram356 (approved):
## Summary

Both findings from the previous review are resolved at this head, each beyond what was asked: runtime alias-collision handling with a drop-and-log guard keyed on the submodule, backed by a new suite that verifies the checked-in registry against the pinned prebid.js modules' actual registrations, and terminal-consent recovery that re-subscribes when a silent, never-answered CMP stub is replaced while keeping the wait of a CMP that already answered. The listener-removal change to the subscribing API is correct: TCF listener ids are namespaced per API, and removal through a replacement could delete another consumer's listener.

The two merges from main integrate cleanly: none of the lines they changed in this PR's files touch managed-ID or consent code, the final diff stays scoped to the feature, and the shim size-bound raise tracks the merged feature set while still failing loudly on a Prebid value-import.

The one comment below is a docs nitpick with a one-click suggestion; it does not affect the verdict.

## Non-blocking

### ⛏ nitpick

- **Degraded-behavior table lacks the two new runtime behaviors** — see inline at `docs/guide/integrations/prebid.md:780`

## CI Status

All 20 checks pass at this head, including the required set (`cargo fmt`, `cargo test`, `format-typescript`, `format-docs`) and the non-required gates (vitest, cross-adapter parity, browser and EC-lifecycle integration tests, CodeQL analyses).

PR #1069, opened 2026-08-24
Add request phase timing design spec

### LINKED ISSUE #1068: Create eng spec for better server side timing metrics and observability
### Description

we need better server side timing metrics and observability

jevansnyc:
Closes #1068.

Design spec for per-request phase attribution: TS-emitted `Server-Timing` subtimings plus Tinybird access telemetry percentiles.

## What the spec covers

- A core `RequestTimings` type (closed phase enum, saturating monotonic spans, cheap-clone handle) collected on every request with no flag, carried through request extensions, the streaming body closure, and post-send emission.
- Header spans for the pre-first-byte phases: `ts-total`, `ts-appbuild`, `ts-filter`, `ts-geo`, `ts-kv`, `ts-origin`, `ts-c2`. Appended in `apply_finalize_headers`, gated by a new `observability.server_timing_enabled` flag. Browsers expose the header via `PerformanceResourceTiming.serverTiming`, so publisher RUM tooling picks up the breakdown with no integration work.
- Body-phase fields the header cannot carry (`auction_wait_ms`, `stream_ms`, `resp_bytes`) flow to Tinybird only.
- Access telemetry: extends the reserved `access_logs_raw` datasource with phase columns, `route_class`, `c2_state`, `ts_version`, and `pop`; one sampled NDJSON row per request emitted strictly after the last body byte, reusing the auction sink pattern; unwires the settings guard that rejects `tinybird.access_enabled`.
- A rider that dedupes the two per-request geo hostcalls into one (finalize reads the stashed request-phase result).
- One percentile endpoint pipe, and a standalone performance dashboard (separate from the revenue and auction dashboards, lives in the telemetry repo).
- Rollout order, including the one external unknown: verifying the fronting delivery layer passes an appended `Server-Timing` through.

Motivation: a production stall window (publisher redacted in the spec) added a uniform ~600 ms to every application-path request and could not be attributed without a live probing session. With this in place the next window names its phase in one response header or one dashboard query.

Spec only; no runtime code changes in this PR. Implementation plan follows.

### DIFF
--- a/docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
+++ b/docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
@@ -1,8 +1,8 @@
 # Request phase timing: Server-Timing subtimings and access telemetry
 
 **Date:** 2026-08-24
-**Status:** Approved design, pending implementation plan.
-**Scope:** `trusted-server-core`, Fastly and Axum adapters, `tinybird/` schema and pipes,
+**Status:** Approved design, revised for review round 1, pending implementation plan.
+**Scope:** `trusted-server-core`, Fastly and Axum adapters, `tinybird/` schema,
 performance dashboard (separate repo).
 
 ---
@@ -31,13 +31,21 @@ about itself.
 
 ## 2. Goals
 
-1. Every response attributes its own server time by phase, in a standard header that
-   browsers expose to JavaScript (`PerformanceResourceTiming.serverTiming`), so any RUM
-   tool the publisher already runs picks up the breakdown with zero integration work.
+1. Every normal application response attributes its own server time by phase in a
+   standard header. Browsers expose the values to same-origin JavaScript via
+   `PerformanceResourceTiming.serverTiming`, so RUM tooling that reads that API can
+   surface the breakdown. Whether a given vendor or the publisher's own monitoring
+   extension actually collects it is verified separately in rollout; the publisher
+   extension needs a small change to render it.
 2. The same numbers flow to Tinybird so we hold p50/p95/p99 per phase, per route class,
    per PoP, per deployed version, and a future stall window self-diagnoses in one query.
-3. Zero cost to TTFB: collection is a handful of monotonic clock reads; telemetry
-   emission happens strictly after the last body byte.
+3. No additional awaited I/O before first byte. The pre-send cost is a handful of
+   monotonic clock reads, one small allocation at entry, and rendering one header;
+   telemetry emission happens strictly after the last body byte.
+
+Scope note: phases cover the application lifecycle after T0. The `/health` and
+`/_ts/debug/ja4` short-circuits, config-store open failures, and request-conversion
+failures bypass the lifecycle and emit nothing.
 
 ## 3. Non-goals
 
@@ -47,8 +55,9 @@ about itself.
   regardless of which filter runs; vendor identity stays out of headers and telemetry.
 - No Cloudflare or Spin emission wiring in v1. Core collection is adapter-neutral; those
   adapters can wire emission later without core changes.
-- No rollup materialized views. The raw datasource plus one endpoint pipe is v1;
-  rollups only if panel latency demands them.
+- No Tinybird endpoint pipe and no rollup materialized views in v1. Grafana queries the
+  datasource through the ClickHouse connector, matching the auction dashboards; rollups
+  only if panel latency demands them.
 - No sampling of the header. The header is all-traffic when enabled; only Tinybird rows
   sample.
 
@@ -56,159 +65,243 @@ about itself.
 
 ```
 adapter entry (T0)
-  |  RequestTimings::new() -> request extensions
+  |  RequestTimings::new() -> shared handle
   v
-app construction ................ ts-appbuild   (adapter)
-pre-route request filters ....... ts-filter     (adapter wrapper)
-geo lookup (single, deduped) .... ts-geo        (adapter; result stashed)
-EC KV before send ............... ts-kv         (core: ec/finalize.rs)
-template cache lookup ........... ts-c2         (core: publisher.rs)
-origin fetch to resp headers .... ts-origin     (core: publisher.rs)
+app construction ................ ts-appbuild        (adapter)
+pre-route request filters ....... ts-filter          (adapter wrapper)
+geo lookup (single, deduped) .... ts-geo             (adapter; result carried forward)
+template cache lookup ........... ts-template-cache  (core: publisher.rs)
+origin fetch to resp headers .... ts-origin          (core: publisher.rs)
+EC identity KV, pre-send ........ ts-kv              (core: KV abstraction)
   |
-finalize middleware:
-  append Server-Timing header ... ts-total + all recorded spans
+send_edgezero_response, immediately before into_parts():
+  mark_headers_ready() snapshot (unconditional)
+  append Server-Timing header (flag-gated, skipped on shared-cacheable responses)
   |
 headers committed; body streams
-  auction hold at seam .......... auction_wait_ms  (row only)
+  auction hold at seam .......... auction_wait_ms    (row only)
   stream duration, bytes ........ stream_ms, resp_bytes (row only)
   |
 post-send (adapter main):
-  sample gate -> one NDJSON row -> Tinybird Events API (best effort)
+  sample gate -> one NDJSON row -> Tinybird Events API
+  bounded response await, 2xx validated
 ```
 
-Collection is always-on and flag-free. Two independent flags gate emission: the header
-(`observability.server_timing_enabled`) and the telemetry row (`tinybird.access_enabled`).
+Collection is always-on and flag-free, including the `mark_headers_ready()` snapshot.
+Two independent flags gate emission: the header (`observability.server_timing_enabled`)
+and the telemetry row (`tinybird.access_enabled`).
 
 ## 5. `RequestTimings` (core)
 
 New module `crates/trusted-server-core/src/request_timing.rs`.
 
-- `Phase`: a closed enum: `AppBuild`, `Filter`, `Geo`, `EcKv`, `Origin`, `C2Lookup`,
-  `AuctionWait`, `Stream`. Header rendering covers the first six plus the derived
-  total; the last two are row-only.
-- Inner state: one fixed-size array of `Option<Duration>` slots indexed by phase, plus
-  `t0: Instant`, plus `resp_bytes: Option<u64>`. Phases that repeat within a request
-  (geo, KV) accumulate by saturating addition into the same slot.
-- Sharing: `RequestTimings` is a cheap-clone handle, `Arc<Mutex<Inner>>`. It must cross
-  three boundaries: request extensions (adapter entry to core handlers), the streaming
-  body closure (records body-phase spans after the response object has been handed
-  off), and the adapter's post-send emission read. A poisoned or contended lock must
-  never fail a request: all recording methods are infallible and drop the sample on
-  lock failure.
+- `Phase`: a closed enum: `AppBuild`, `Filter`, `Geo`, `EcKv`, `Origin`,
+  `TemplateCacheLookup`, `AuctionWait`, `Stream`. Header rendering covers the first six
+  plus the stored total; the last two are row-only.
+- Inner state: one fixed-size array of `Option<Duration>` slots indexed by phase,
+  `t0: Instant`, `headers_ready_total: Option<Duration>`, and
+  `resp_bytes: Option<u64>`. Phases that repeat within a request (geo, KV) accumulate
+  by saturating addition into the same slot.
+- `mark_headers_ready()`: stores `t0.elapsed()` once at the response-commit boundary,
+  unconditionally, before either emission flag is consulted. The header renders this
+  stored value as `ts-total`; the telemetry row reads the same stored value as
+  `time_elapsed_ms`. The two surfaces cannot disagree, and the row stays correct when
+  the header flag is off. Full request duration is captured separately at row build as
+  `request_elapsed_ms` (`t0.elapsed()` after the body has streamed).
+- Sharing: `RequestTimings` is a cheap-clone handle, `Arc<Mutex<Inner>>`. It crosses
+  three boundaries: adapter entry to core handlers, the streaming body closure (records
+  body-phase spans after the response object has been handed off), and the adapter's
+  post-send emission read. A poisoned or contended lock must never fail a request: all
+  recording methods are infallible and drop the sample on lock failure.
 - Recording API: `timings.record(Phase::Geo, dur)` and a scope guard
   `timings.span(Phase::Origin)` that records on drop. Guards use saturating duration
   math; a non-monotonic reading records zero rather than panicking.
 - Rendering: `server_timing_value(&self) -> Option<String>` produces
   `ts-total;dur=41.2, ts-appbuild;dur=18.4, ts-filter;dur=9.1` with durations in
-  milliseconds at one decimal. Phases never recorded are omitted. `ts-total` is
-  elapsed-since-`t0` at render time, which by construction is the last mutation before
-  send. Returns `None` when nothing was recorded (defensive; `t0` always exists).
+  milliseconds at one decimal. Phases never recorded are omitted. Returns `None` when
+  `mark_headers_ready()` has not run.
 
 `Instant` is already used freely in the guest (`publisher.rs`, `auction/telemetry.rs`),
 so no new clock abstraction is needed.
 
 ## 6. Span taxonomy and recording sites
 
-| Entry         | Measures                                                                  | Site                                                                                    |
-| ------------- | ------------------------------------------------------------------------- | --------------------------------------------------------------------------------------- |
-| `ts-total`    | adapter entry (post health short-circuit) to header emission              | derived at render                                                                       |
-| `ts-appbuild` | config-store open, Settings parse, orchestrator + registry + router build | Fastly `main.rs` around `open_trusted_server_config_store()` + `build_app_with_state()` |
-| `ts-filter`   | pre-route request filters, end to end (backend ensure, secret read, POST) | around `run_pre_route_filters` (`app.rs:751`)                                           |
-| `ts-geo`      | geo hostcall (single after dedupe; accumulates if any path still repeats) | `build_ec_request_state` (`app.rs:410`)                                                 |
-| `ts-kv`       | EC identity KV operations before send                                     | `ec/finalize.rs` ingest path, EC-route KV lookups                                       |
-| `ts-origin`   | publisher backend send to response headers available (C1 hit or miss)     | `publisher.rs` around the origin `send` (~line 4496)                                    |
-| `ts-c2`       | template cache `lookup_or_reserve`, including the hit-path full-body read | `publisher.rs` around the lookup (~line 4391)                                           |
+| Entry               | Measures                                                                              | Site                                                                                    |
+| ------------------- | ------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------- |
+| `ts-total`          | T0 to `mark_headers_ready()` at the response-commit boundary                          | stored snapshot                                                                         |
+| `ts-appbuild`       | config-store open, Settings parse, orchestrator + registry + router build             | Fastly `main.rs` around `open_trusted_server_config_store()` + `build_app_with_state()` |
+| `ts-filter`         | pre-route request filters, end to end (backend ensure, secret read, POST)             | around `run_pre_route_filters` (`app.rs:751`)                                           |
+| `ts-geo`            | geo hostcall (single after dedupe; accumulates if any path still repeats)             | `build_ec_request_state` (`app.rs:410`)                                                 |
+| `ts-kv`             | EC identity KV operations before response send (see enumeration below)                | the shared KV abstraction                                                               |
+| `ts-origin`         | publisher backend send to response headers available (read-through cache hit or miss) | `publisher.rs` around the origin `send`                                                 |
+| `ts-template-cache` | template cache `lookup_or_reserve`, including the hit-path full-body read             | `publisher.rs` around the lookup                                                        |
+
+Naming follows the completed template-cache terminology migration (`x-ts-template-cache`
+is the emitted header on `main`; `c2` naming is retired).
+
+`ts-kv` is instrumented at the shared identity-graph/KV abstraction
+(`KvIdentityGraph` and the platform KV store wrapper), not at individual call sites, so
+new callers cannot silently escape the span. Included pre-send operations: EC
+generation `create_or_revive`, identify-path graph reads and evaluation, finalize-path
+`ingest_eid_cookies`/`upsert_partner_ids` and withdrawal tombstones, consent-store
+reads on consent routes, and batch-sync graph access when it runs before send.
+Explicitly excluded: pull-sync work, which runs strictly after `send_to_client` and is
+invisible to both surfaces. The timings handle reaches `ec_finalize_response` through
+an existing context or a parameter object; that function already has the repository
+maximum of seven arguments and does not gain an eighth.
 
 Row-only fields:
 
-| Field             | Measures                                                 | Site                                              |
-| ----------------- | -------------------------------------------------------- | ------------------------------------------------- |
-| `auction_wait_ms` | hold at the `</body>` seam waiting on dispatched auction | `collect_stream_auction` / seam wait in publisher |
-| `stream_ms`       | headers committed to last body byte                      | adapter around the body-stream drive              |
-| `resp_bytes`      | bytes written to the client body                         | same                                              |
+| Field                | Measures                                                 | Site                                              |
+| -------------------- | -------------------------------------------------------- | ------------------------------------------------- |
+| `auction_wait_ms`    | hold at the `</body>` seam waiting on dispatched auction | `collect_stream_auction` / seam wait in publisher |
+| `stream_ms`          | headers committed to last body byte                      | adapter around the body-stream drive              |
+| `resp_bytes`         | bytes written to the client body                         | same                                              |
+| `request_elapsed_ms` | T0 to row build (full request duration, post-body)       | row assembly                                      |
 
-## 7. Header emission
+`auction_wait_ms` is nested inside `stream_ms`. Queries and panels must never stack or
+sum them as siblings; the derived quantities are `stream_other_ms = stream_ms -
+auction_wait_ms` and `unattributed_ms = max(time_elapsed_ms - (appbuild + filter + geo
 
-The last step of `apply_finalize_headers` (`middleware.rs:197`) appends one
-`Server-Timing` header from `server_timing_value()`, gated on
-`observability.server_timing_enabled`. Append semantics, never insert: an
-origin-supplied Server-Timing survives, and the fronting delivery layer's own entries
-(`time-elapsed`, `hit-state`) are additive per the header's list semantics.
+- kv + origin + template_cache), 0)`, computed at query time.
 
-Emission runs after the finalize geo resolution and after every configurable response
-mutation, so the header reflects everything that happened before send. Both finalize
-sites (the router middleware and the adapter entry-point fallback for streaming
-responses) emit through the same helper; the `HEADER_X_TS_FINALIZED` marker already
-prevents double-finalization, which also prevents a doubled header.
+## 7. Freeze point and header emission
+
+The freeze-and-emit point is `send_edgezero_response` (Fastly `main.rs`), immediately
+before `response.into_parts()`. This is the single choke point every send path shares,
+and it runs after everything that can still mutate the response: the router middleware,
+entry-point finalize (`apply_finalize_headers`, asset-policy reapplication), EC
+finalization and its KV work, and terminal filter/privacy effects.
+`apply_finalize_headers` itself does not emit; the `HEADER_X_TS_FINALIZED` sentinel
+marks middleware finalization, not header commitment, and must not be treated as the
+timing boundary.
+
+At the freeze point, in order: `mark_headers_ready()` (unconditional), then, gated on
+`observability.server_timing_enabled`, append one `Server-Timing` header from
+`server_timing_value()`. Append semantics, never insert: an origin-supplied
+Server-Timing survives, and the fronting delivery layer's own entries (`time-elapsed`,
+`hit-state`) are additive per the header's list semantics.
+
+Header emission is skipped when the response is shared-cacheable (a public
+`Cache-Control` with positive freshness, or any surrogate/CDN cache directive): a
+shared cache would replay one request's timings for the object's full TTL. The
+long-lived immutable `tsjs` asset route is the concrete case. The snapshot and the
+telemetry row are unaffected by this skip.
+
+The Axum adapter applies the same rule at its equivalent last point before response
+serialization, using the same core helper.
 
 ## 8. Geo lookup dedupe (rider)
 
 Today every dispatched request pays two geo hostcalls for one answer: request-phase in
 `build_ec_request_state` (`app.rs:410`) and response-phase in
-`FinalizeResponseMiddleware` (`middleware.rs:83`, alternate site `main.rs:285`). The
-first lookup stashes the resolved `GeoInfo` in request extensions; the finalize path
-reads the stashed value and only falls back to a live lookup when the request-phase
-never ran. This halves exposure to a degraded geo subsystem, which is one of the three
-candidate dependencies for the observed stall window. The 401 short-circuit behavior
+`FinalizeResponseMiddleware` (`middleware.rs:83`, alternate site `main.rs:285`).
+
+Plain request extensions cannot carry the result out: the middleware moves the request
+context into `next.run(ctx)` and holds only the response afterward. The resolved geo
+travels instead through the response path, either on the existing `EcFinalizeState`
+response extension or a dedicated response extension installed by the dispatch layer,
+as a three-state value: `NotAttempted`, `Attempted(None)` (lookup ran and failed, do
+not retry), and `Resolved(GeoInfo)`. The finalize path consumes the carried value and
+performs a live lookup only in the `NotAttempted` state. The 401 rule
 (`resolve_geo_for_response` skips lookup for unauthorized responses) is preserved.
 
 ## 9. Access telemetry row
 
-Extends the reserved `tinybird/datasources/access_logs_raw.datasource` (in-repo only,
-no deployed pipes over it, so this is an edit rather than a migration).
+Extends the reserved `tinybird/datasources/access_logs_raw.datasource`.
 
-Kept columns: `event_ts`, `method`, `path`, `status`, `time_elapsed_ms` (now defined as
-`ts-total`), `cache_state`, `country`, `sample_rate`, `event_date`, 30-day TTL.
+Kept columns: `event_ts`, `method`, `status`, `time_elapsed_ms` (defined as the
+`mark_headers_ready()` snapshot), `cache_state`, `country`, `sample_rate`,
+`event_date`, 30-day TTL.
+
+Removed: raw `path`. Route identifiers like `/_ts/admin/ec/{id}` would otherwise put
+EC identifiers into a 30-day dataset, and publisher paths carry unbounded cardinality.
+Replaced by `route_template`: for named routes, the matched route-table pattern
+verbatim (parameters stay as placeholders); for the publisher fallback, the bounded
+normalized form produced by the existing auction-telemetry path normalizer. A
+serialization test proves a literal EC identifier never appears in an emitted row.
 
 Added columns:
 
 ```
-`route_class`     LowCardinality(String),      -- publisher_html | tsjs | integration_proxy | ec | auction_api | other
-`appbuild_ms`     Nullable(UInt32),
-`filter_ms`       Nullable(UInt32),
-`geo_ms`          Nullable(UInt32),
-`kv_ms`           Nullable(UInt32),
-`origin_ms`       Nullable(UInt32),
-`c2_ms`           Nullable(UInt32),
-`auction_wait_ms` Nullable(UInt32),
-`stream_ms`       Nullable(UInt32),
-`resp_bytes`      Nullable(UInt64),
-`c2_state`        LowCardinality(Nullable(String)),  -- x-ts-c2-cache value
-`ts_version`      LowCardinality(String),
-`pop`             LowCardinality(Nullable(String))   -- FASTLY_POP
+`publisher_domain`      LowCardinality(String),      -- row identity, matches auction schema
+`env`                   LowCardinality(String),      -- x-ts-env value
+`route_class`           LowCardinality(String),      -- publisher_html | tsjs | integration_proxy | ec | auction_api | other
+`route_template`        String,                      -- bounded, normalized; replaces path
+`appbuild_ms`           Nullable(UInt32),
+`filter_ms`             Nullable(UInt32),
+`geo_ms`                Nullable(UInt32),
+`kv_ms`                 Nullable(UInt32),
+`origin_ms`             Nullable(UInt32),
+`template_cache_ms`     Nullable(UInt32),
+`auction_wait_ms`       Nullable(UInt32),
+`stream_ms`             Nullable(UInt32),
+`request_elapsed_ms`    Nullable(UInt32),
+`resp_bytes`            Nullable(UInt64),
+`template_cache_state`  LowCardinality(Nullable(String)),  -- x-ts-template-cache value
+`ts_version`            LowCardinality(String),
+`pop`                   LowCardinality(Nullable(String))   -- FASTLY_POP
 ```
 
 Null means the phase did not run. `route_class` is assigned by the adapter at handler
-selection (the named-route table knows which handler it picked), not by path regex.
-Percentile analysis groups by `route_class`; `path` stays for drill-down only.
+selection, not by path regex. `ts_version` plus `pop` is not globally unique across
+deployments; `publisher_domain` plus `env` provides row identity.
+
+Sorting key: `(event_date, publisher_domain, route_class, pop, status)`, aligned with
+the analysis dimensions. This replaces the reserved key
+`(event_date, path, status, method)`. Rollout step 4 verifies whether the reserved
+datasource was ever deployed to the remote workspace; if it was, this schema ships as
+a versioned replacement datasource with a cutover, not an in-place edit.
 
 ## 10. Emission mechanics
 
 - Point: Fastly `main.rs`, strictly after `send_edgezero_response` returns (the body
   has fully streamed). The row can never affect TTFB or block delivery.
 - Sampling: uniform per-request decision against `tinybird.access_sample_rate`
-  (validation `0.0..=1.0` already exists). No client stickiness.
-- Transport: one NDJSON row to the Tinybird Events API, mirroring the auction sink
-  (`tinybird.rs`): same `api_host`, reserved `access_dataset` and
-  `access_token_secret`, 2 s first-byte and between-bytes timeouts, `max_body_bytes`
-  guard, best-effort with a warn log on failure, no retry.
+  (validation `0.0..=1.0` already exists). No client stickiness. Sampled-out requests
+  are silent; every other drop (row build failure, send failure, non-2xx) logs a
+  warning naming the reason, so lost diagnostic rows are visible.
+- Transport: one NDJSON row to the Tinybird Events API: same `api_host`, reserved
+  `access_dataset` and `access_token_secret`, 2 s first-byte and between-bytes
+  timeouts, `max_body_bytes` guard, no retry.
+- Delivery confirmation: unlike the auction sink, which starts `send_async` and drops
+  the pending response (it runs before delivery completes and cannot afford to wait),
+  the access emitter runs after the client has the full response and therefore awaits
+  the bounded ingest response and validates 2xx. A non-2xx or timeout logs a warning
+  with the status. This is what makes the warning semantics real rather than
+  aspirational.
+- Budget: at `access_sample_rate = 1.0` this adds one backend request per request to
+  the service, after delivery. The sample rate is the budget control; 1.0 is a
+  diagnosis setting, not a steady state.
 - Settings: the guard that rejects `tinybird.access_enabled = true` "until an emitter
-  is wired" (`settings.rs:4123` test) flips into a wiring test in the same change, so
-  the flag and the emitter land together. `access_enabled = true` requires
-  `tinybird.enabled` and a non-empty `api_host`.
-- Axum adapter: emits the header via the same core helper; does not emit Tinybird rows
-  in v1 (local dev has no Tinybird backend; the sink stays behind the Fastly adapter).
+  is wired" flips into a wiring test in the same change. `access_enabled = true`
+  requires `tinybird.enabled` and a non-empty `api_host`.
+- Axum adapter: emits the header only; no Tinybird rows in v1.
 
-## 11. Pipe and dashboard
+## 11. Dashboard and query model
 
-- `tinybird/pipes/access_phase_stats.pipe`: endpoint returning p50/p95/p99 of each
-  phase column grouped by `route_class` and hour, filterable by `pop` and `ts_version`.
-- Dashboard: a new standalone `grafana/dashboards/edge-performance.json` in the
-  telemetry repo (`trusted-server-tinybird`), performance only, no panels shared with
-  the revenue and auction dashboards. Panels: phase percentiles by route class, stacked
-  phase breakdown over time, PoP split, version overlay, C2 hit-rate, and a stall
-  panel: rows with `time_elapsed_ms > 500` grouped by dominant phase. Queries hit the
-  raw datasource with `$__timeFilter(event_ts)`.
+No endpoint pipe in v1. Grafana queries `access_logs_raw` directly through the
+ClickHouse connector with `$__timeFilter(event_ts)`, matching the auction dashboards.
+
+Dashboard: a new standalone `grafana/dashboards/edge-performance.json` in the
+telemetry repo (`trusted-server-tinybird`), performance only, no panels shared with
+the revenue and auction dashboards. Panels:
+
+- Phase percentiles (p50/p95/p99) by `route_class`, per phase column.
+- Stacked phase breakdown over time using the non-overlapping set: `appbuild_ms`,
+  `filter_ms`, `geo_ms`, `kv_ms`, `origin_ms`, `template_cache_ms`, and derived
+  `unattributed_ms`. `auction_wait_ms` and derived `stream_other_ms` chart in a
+  separate body-phase panel and never stack with pre-header phases.
+- PoP split, `ts_version` overlay, template-cache state rates.
+- Stall panel: rows with `request_elapsed_ms > 500` (post-body total, so body-only
+  stalls are caught) grouped by dominant phase, where `unattributed_ms` competes as a
+  phase so the panel cannot confidently blame a small measured span while most time is
+  uninstrumented.
+
+Sampling semantics for every aggregate: `sample_rate` must be operationally stable
+within any queried window. Panels either filter to a single `sample_rate` value or
+apply inverse-probability weights (`count() * 1/sample_rate`) for volume panels;
+pooled unweighted quantiles across a rate change are documented as invalid.
 
 ## 12. Config surface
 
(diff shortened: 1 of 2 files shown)


docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,287 @@
...
+about itself.
+
+## 2. Goals
+
+1. Every response attributes its own server time by phase, in a standard header that
+   browsers expose to JavaScript (`PerformanceResourceTiming.serverTiming`), so any RUM
+   tool the publisher already runs picks up the breakdown with zero integration work.
+2. The same numbers flow to Tinybird so we hold p50/p95/p99 per phase, per route class,
+   per PoP, per deployed version, and a future stall window self-diagnoses in one query.
+3. Zero cost to TTFB: collection is a handful of monotonic clock reads; telemetry
```
jevansnyc: These claims are too absolute. Health, JA4, config-store failures, and request-conversion failures bypass this lifecycle, while clock reads, locking, allocation, and header rendering still happen before TTFB. `Server-Timing` also makes values available to browser APIs but does not guarantee that an existing RUM vendor collects them automatically. Please scope this to normal application responses after T0 and say `no additional awaited I/O; measured negligible pre-send CPU`, with the exact RUM integration and any `Timing-Allow-Origin` requirements verified separately.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,287 @@
...
+
+```
+adapter entry (T0)
+  |  RequestTimings::new() -> request extensions
+  v
+app construction ................ ts-appbuild   (adapter)
+pre-route request filters ....... ts-filter     (adapter wrapper)
+geo lookup (single, deduped) .... ts-geo        (adapter; result stashed)
+EC KV before send ............... ts-kv         (core: ec/finalize.rs)
+template cache lookup ........... ts-c2         (core: publisher.rs)
```
jevansnyc: The repository has completed the C2 -> template-cache terminology migration: current code emits `x-ts-template-cache`, and tests require `x-ts-c2-cache` to be absent. Please rename this phase and every related header, column, state, and dashboard label to `TemplateCacheLookup` / `ts-template-cache` / `template_cache_ms` / `template_cache_state`.

jevansnyc (commented):
The direction is sound, but I found correctness blockers around the finalization boundary and total-time snapshot, plus naming, privacy, telemetry-delivery, configuration, and query-model concerns. I would address the blocking items before treating the design as implementation-ready. Inline comments contain concrete suggested changes.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,287 @@
...
+  off), and the adapter's post-send emission read. A poisoned or contended lock must
+  never fail a request: all recording methods are infallible and drop the sample on
+  lock failure.
+- Recording API: `timings.record(Phase::Geo, dur)` and a scope guard
+  `timings.span(Phase::Origin)` that records on drop. Guards use saturating duration
+  math; a non-monotonic reading records zero rather than panicking.
+- Rendering: `server_timing_value(&self) -> Option<String>` produces
+  `ts-total;dur=41.2, ts-appbuild;dur=18.4, ts-filter;dur=9.1` with durations in
+  milliseconds at one decimal. Phases never recorded are omitted. `ts-total` is
+  elapsed-since-`t0` at render time, which by construction is the last mutation before
```
jevansnyc: `ts-total` needs a stored pre-send snapshot rather than a live elapsed-time calculation. The access row is read after streaming, so recomputing then would include body time and disagree with the header; if header emission is disabled but access telemetry is enabled, no pre-send snapshot is created at all. Add an unconditional `headers_ready_total`/`mark_headers_ready()` value at the response-commit boundary and have both surfaces consume it. A separate post-body `request_elapsed_ms` can represent full request duration.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,287 @@
...
+| ------------- | ------------------------------------------------------------------------- | --------------------------------------------------------------------------------------- |
+| `ts-total`    | adapter entry (post health short-circuit) to header emission              | derived at render                                                                       |
+| `ts-appbuild` | config-store open, Settings parse, orchestrator + registry + router build | Fastly `main.rs` around `open_trusted_server_config_store()` + `build_app_with_state()` |
+| `ts-filter`   | pre-route request filters, end to end (backend ensure, secret read, POST) | around `run_pre_route_filters` (`app.rs:751`)                                           |
+| `ts-geo`      | geo hostcall (single after dedupe; accumulates if any path still repeats) | `build_ec_request_state` (`app.rs:410`)                                                 |
+| `ts-kv`       | EC identity KV operations before send                                     | `ec/finalize.rs` ingest path, EC-route KV lookups                                       |
```
jevansnyc: This scope misses KV operations on the suspected path: EC generation performs `create_or_revive` before finalization, identify performs graph reads/evaluation, and batch sync/auction enrichment have additional graph access. Please instrument the shared identity-graph/KV abstraction or explicitly enumerate every included and excluded call site. Also, `ec_finalize_response` already has the repository maximum of seven arguments, so the timing handle should travel through a parameter object or existing context rather than another argument.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,287 @@
...
+
+| Field             | Measures                                                 | Site                                              |
+| ----------------- | -------------------------------------------------------- | ------------------------------------------------- |
+| `auction_wait_ms` | hold at the `</body>` seam waiting on dispatched auction | `collect_stream_auction` / seam wait in publisher |
+| `stream_ms`       | headers committed to last body byte                      | adapter around the body-stream drive              |
+| `resp_bytes`      | bytes written to the client body                         | same                                              |
+
+## 7. Header emission
+
+The last step of `apply_finalize_headers` (`middleware.rs:197`) appends one
```
jevansnyc: Blocking: `apply_finalize_headers` is not the last response mutation in the current Fastly lifecycle. After this middleware returns, `main.rs` still applies asset policy, EC finalization/KV work, and terminal DataDome/privacy effects before splitting and sending the response. As written, `ts-kv` is missing, `ts-total` is early, and a later effect can overwrite the header. Freeze and emit the timing immediately before `response.into_parts()`, after all terminal response effects; the finalized sentinel does not mean headers are committed.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,287 @@
...
+sites (the router middleware and the adapter entry-point fallback for streaming
+responses) emit through the same helper; the `HEADER_X_TS_FINALIZED` marker already
+prevents double-finalization, which also prevents a doubled header.
+
+## 8. Geo lookup dedupe (rider)
+
+Today every dispatched request pays two geo hostcalls for one answer: request-phase in
+`build_ec_request_state` (`app.rs:410`) and response-phase in
+`FinalizeResponseMiddleware` (`middleware.rs:83`, alternate site `main.rs:285`). The
+first lookup stashes the resolved `GeoInfo` in request extensions; the finalize path
```
jevansnyc: This handoff does not work through ordinary request extensions: the outer middleware moves the request context into `next.run(ctx)` and receives only a response afterward. Use a shared cell installed before dispatch, or carry geo through a response/`EcFinalizeState` extension. Model `NotAttempted`, `Attempted(None)`, and `Resolved` separately so a failed lookup is not repeated, while preserving the 401 no-lookup rule.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,287 @@
...
+never ran. This halves exposure to a degraded geo subsystem, which is one of the three
+candidate dependencies for the observed stall window. The 401 short-circuit behavior
+(`resolve_geo_for_response` skips lookup for unauthorized responses) is preserved.
+
+## 9. Access telemetry row
+
+Extends the reserved `tinybird/datasources/access_logs_raw.datasource` (in-repo only,
+no deployed pipes over it, so this is an edit rather than a migration).
+
+Kept columns: `event_ts`, `method`, `path`, `status`, `time_elapsed_ms` (now defined as
```
jevansnyc: Please do not retain raw `path`. Routes such as `/_ts/admin/ec/{id}` can put an EC identifier directly into this 30-day dataset, and publisher paths can carry other high-cardinality identifiers. Store a bounded normalized route template (the existing auction telemetry normalizer is a useful starting point) and add a serialization test proving a literal EC ID never appears. The row should also identify publisher/service/environment; `ts_version` plus PoP is not globally unique across deployments.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,287 @@
...
+## 10. Emission mechanics
+
+- Point: Fastly `main.rs`, strictly after `send_edgezero_response` returns (the body
+  has fully streamed). The row can never affect TTFB or block delivery.
+- Sampling: uniform per-request decision against `tinybird.access_sample_rate`
+  (validation `0.0..=1.0` already exists). No client stickiness.
+- Transport: one NDJSON row to the Tinybird Events API, mirroring the auction sink
+  (`tinybird.rs`): same `api_host`, reserved `access_dataset` and
+  `access_token_secret`, 2 s first-byte and between-bytes timeouts, `max_body_bytes`
+  guard, best-effort with a warn log on failure, no retry.
```
jevansnyc: Mirroring the current auction sink cannot provide the stated warning semantics: it starts `send_async` and drops the pending response, so it cannot observe HTTP 4xx/5xx, timeout, schema rejection, quarantine, or ingestion success. Either await a bounded response after client delivery and validate 2xx, or explicitly define delivery as unconfirmed and add ingestion-freshness/quarantine monitoring. Please also account for the extra backend-request budget and expose a dropped-row reason/count, since the slowest high-fanout requests are otherwise the most likely to lose their diagnostic row.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,287 @@
...
+  is wired" (`settings.rs:4123` test) flips into a wiring test in the same change, so
+  the flag and the emitter land together. `access_enabled = true` requires
+  `tinybird.enabled` and a non-empty `api_host`.
+- Axum adapter: emits the header via the same core helper; does not emit Tinybird rows
+  in v1 (local dev has no Tinybird backend; the sink stays behind the Fastly adapter).
+
+## 11. Pipe and dashboard
+
+- `tinybird/pipes/access_phase_stats.pipe`: endpoint returning p50/p95/p99 of each
+  phase column grouped by `route_class` and hour, filterable by `pop` and `ts_version`.
```
jevansnyc: Please define sampling semantics for these aggregates. If `access_sample_rate` changes within a queried window, pooled unweighted quantiles and counts overrepresent periods with the higher rate. Require a stable rate per bucket, group/filter by `sample_rate`, or use inverse-probability weighting; traffic-volume panels need weighted counts.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,287 @@
...
+- Dashboard: a new standalone `grafana/dashboards/edge-performance.json` in the
+  telemetry repo (`trusted-server-tinybird`), performance only, no panels shared with
+  the revenue and auction dashboards. Panels: phase percentiles by route class, stacked
+  phase breakdown over time, PoP split, version overlay, C2 hit-rate, and a stall
```
jevansnyc: `auction_wait_ms` is nested inside `stream_ms`, so these cannot be stacked or summed as independent phases. Define `stream_other_ms = stream_ms - auction_wait_ms`, and add `unattributed_ms = max(headers_ready_total - sum(non-overlapping pre-header spans), 0)` so the dominant-phase panel does not confidently blame a tiny measured phase while most time is uninstrumented. If the stall panel should catch body-only stalls, it must use a post-send total rather than pre-header `time_elapsed_ms`.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,287 @@
...
+  panel: rows with `time_elapsed_ms > 500` grouped by dominant phase. Queries hit the
+  raw datasource with `$__timeFilter(event_ts)`.
```
jevansnyc: This contradicts the endpoint-pipe design above: choose either a published, scoped endpoint for Grafana or direct raw-datasource queries, rather than specifying both. The current datasource sort key `(event_date, path, status, method)` is also poorly aligned with route-class/PoP/version queries and embeds the high-cardinality path problem. Redefine it before first deployment, or document a compatible migration/cutover if the datasource already exists remotely.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,287 @@
...
+[observability]
+# Append TS phase timings to the Server-Timing response header.
+server_timing_enabled = false  # example default
+
+[tinybird]
+access_enabled = false         # requires tinybird.enabled and api_host when true
+access_sample_rate = 0.0       # 1.0 during active diagnosis, dial down after
+```
+
+New `ObservabilitySettings` struct with the single boolean, default off, standard
```
jevansnyc, line 438: Please specify rollback/config compatibility. `Settings` uses `deny_unknown_fields`, so serializing a default top-level `[observability]` table can make the previous binary reject the config during rollback. Omit the default table during serialization, add an old-schema compatibility test, and document binary/config deployment order. The current override path also cannot reliably create a missing nested leaf, so the disabled setting likely needs to exist in base TOML and be pushed explicitly.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,287 @@
...
+
+## 16. Overhead
+
+Roughly ten monotonic clock reads and one ~130-byte header per request; one sampled
+HTTP POST after the response has fully streamed. No allocation in the hot path beyond
+the one `Arc` at entry and the rendered header string at finalize.
+
+## 17. Risks and open questions
+
+- The fronting delivery layer's Server-Timing handling is unverified until the first
```
jevansnyc: The delivery layer is not the only external/security question. Public `Server-Timing` values can disclose protection latency/presence and template-cache behavior, and a fronting cache can replay stale timing headers unless storage/replay behavior is defined. Add a threat-model decision (public vs gated, quantized, or temporary), a cache strip/replace policy, and explicit validation of the actual RUM agent and same-origin/cross-origin behavior.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Scoped in 77b92f40: Goals rewritten to 'no additional awaited I/O before first byte' with pre-send CPU named; a scope note excludes health/JA4/config-failure/conversion-failure paths; the RUM claim now says browsers expose the values and vendor collection is verified separately in rollout (the monitoring extension needs a small change to render it).

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Renamed throughout in 77b92f40: ts-template-cache, template_cache_ms, template_cache_state, Phase::TemplateCacheLookup, plus a note recording the completed terminology migration. No c2 naming remains in the spec.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 77b92f40: unconditional mark_headers_ready() stores the snapshot at the freeze point before either emission flag is consulted; header ts-total and row time_elapsed_ms both consume the stored value; request_elapsed_ms added at row build for full post-body duration.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 77b92f40: ts-kv is instrumented at the shared KV abstraction (KvIdentityGraph plus the platform wrapper) so call sites cannot drift, with an explicit include list (create_or_revive, identify reads, finalize ingest and tombstones, consent reads, pre-send batch sync) and an explicit exclusion (pull sync, post-send). The handle reaches ec_finalize_response via an existing context or parameter object; the function keeps seven arguments.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Verified against main and confirmed: apply_finalize_headers runs at main.rs:288 while ec_finalize_response runs at :302, with sends at :216/:231/:253. Fixed in 77b92f40: freeze and emission move into send_edgezero_response immediately before into_parts(), after asset policy, EC finalization, and terminal filter/privacy effects; the spec now states the finalized sentinel is not the timing boundary, and an adapter test asserts ts-kv captures EC finalize work.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 77b92f40: the resolved geo travels on the response path (EcFinalizeState or a dedicated response extension) as NotAttempted / Attempted(None) / Resolved, live lookup only on NotAttempted, failed lookups never retried, 401 no-lookup rule preserved. The request-extensions mechanism is gone from the spec.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 77b92f40: raw path removed from the row. route_template carries the matched route-table pattern for named routes and the auction-telemetry normalizer output for publisher fallback, with a serialization test proving a literal EC identifier never appears. publisher_domain and env added for row identity alongside ts_version and pop.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 77b92f40: the access emitter awaits the bounded ingest response and validates 2xx (it runs after client delivery, so unlike the auction sink it can afford to), logs a reason for every non-sampling drop, and the spec notes the added backend-request budget with sample_rate as the control and 1.0 called out as a diagnosis setting.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 77b92f40: aggregates require an operationally stable sample_rate per queried window; panels filter to one rate or apply inverse-probability weights for volume; pooled unweighted quantiles across a rate change are documented as invalid.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 77b92f40: the nesting is documented at the schema, stream_other_ms and unattributed_ms are defined as query-time derivations, the stacked panel uses only the non-overlapping set with unattributed_ms included, and the stall panel keys on request_elapsed_ms (post-body) with unattributed competing as a phase.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 77b92f40: the endpoint pipe is dropped from v1 and Grafana queries the datasource directly via the ClickHouse connector, matching the auction dashboards. Sorting key redefined to (event_date, publisher_domain, route_class, pop, status); rollout step 4 verifies whether the reserved datasource was ever deployed remotely and ships a versioned replacement with cutover if so.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 77b92f40: deploy order binary-then-config and rollback order config-then-binary are specified; serialization omits the default table with a compatibility test against the previous schema; the key ships present-but-false in base operator TOML because the env overlay cannot create a missing leaf.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Addressed in 77b92f40 in two parts. The public all-traffic exposure is recorded as a decision with rationale (durations only, delivery layer already exposes hit-state and time-elapsed, vendor identity masked) rather than left open. The cache-replay half was a real gap: header emission is now skipped on shared-cacheable responses, with the immutable tsjs route named as the concrete case; RUM agent verification is rollout step 3.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,406 @@
...
+**Date:** 2026-08-24
+**Status:** Approved design, revised for review round 1, pending implementation plan.
+**Scope:** `trusted-server-core`, Fastly and Axum adapters, `tinybird/` schema,
+performance dashboard (separate repo).
+
+---
+
+## 1. Problem
+
+On 2026-08-21 a production deployment (publisher redacted, `prospect-a.com`) showed an
```
jevansnyc: Repository documentation rules require example or fictional information and specifically recommend example-domain names. Please replace `prospect-a.com` with something under `.example`, such as `prospect-a.example`.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,406 @@
...
+
+Attributing that window required a live probing session: route-by-route bisection,
+cookie-deletion experiments, and an eight-agent code trace. The trace found no
+unconditional await on the path that could cost 570 ms, and exactly two
+config-conditional candidates (the pre-route request filter's synchronous verification
+POST, and EC identity KV writes before send), plus one dependency shared by every
+application route (two geo hostcalls per request). We could not tell which one stalled,
+because nothing in the response says where server time went.
+
+The Compute CPU budget is ~50 ms per request, so any large `time-elapsed` is wall-clock
```
jevansnyc: This remains too absolute. Large `time-elapsed` strongly suggests time outside active guest CPU, but it is not a dependency await *by definition*; platform scheduling, hostcall queueing, and the VCL phase where `time.elapsed` is observed also matter. Please qualify this and identify the exact delivery/VCL measurement point used for comparison.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,406 @@
...
+| Field                | Measures                                                 | Site                                              |
+| -------------------- | -------------------------------------------------------- | ------------------------------------------------- |
+| `auction_wait_ms`    | hold at the `</body>` seam waiting on dispatched auction | `collect_stream_auction` / seam wait in publisher |
+| `stream_ms`          | headers committed to last body byte                      | adapter around the body-stream drive              |
+| `resp_bytes`         | bytes written to the client body                         | same                                              |
+| `request_elapsed_ms` | T0 to row build (full request duration, post-body)       | row assembly                                      |
+
+`auction_wait_ms` is nested inside `stream_ms`. Queries and panels must never stack or
+sum them as siblings; the derived quantities are `stream_other_ms = stream_ms -
+auction_wait_ms` and `unattributed_ms = max(time_elapsed_ms - (appbuild + filter + geo
```
jevansnyc: Blocking: `auction_wait_ms` is not universally nested in `stream_ms`. On a Fastly shared-template miss, `publisher_response_into_streaming_response` calls `buffer_publisher_response_async` before returning the response, so the auction can be collected before headers commit. In that mode `auction_wait_ms > stream_ms` is possible, the wait is included in pre-header `unattributed_ms`, and `stream_other_ms` can go negative. Split this into pre-header versus in-stream auction wait, or record an explicit placement/body mode and derive conditionally. Test ordinary streaming, template-cache miss/hit, and Axum buffering separately.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,406 @@
...
+
+- kv + origin + template_cache), 0)`, computed at query time.
```
jevansnyc: The blank line makes `- kv` render as a Markdown bullet and breaks the formula. Separately, every phase column is nullable, so ClickHouse arithmetic returns `NULL` whenever any phase did not run. Specify `coalesce(column, 0)` for every nullable term, use `greatest(..., 0)` for subtraction, and add query tests covering sparse phase combinations.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,406 @@
...
+marks middleware finalization, not header commitment, and must not be treated as the
+timing boundary.
+
+At the freeze point, in order: `mark_headers_ready()` (unconditional), then, gated on
+`observability.server_timing_enabled`, append one `Server-Timing` header from
+`server_timing_value()`. Append semantics, never insert: an origin-supplied
+Server-Timing survives, and the fronting delivery layer's own entries (`time-elapsed`,
+`hit-state`) are additive per the header's list semantics.
+
+Header emission is skipped when the response is shared-cacheable (a public
```
jevansnyc: Blocking: this predicate does not guarantee that request-specific timings cannot be cached. It misses `max-age=N` and `s-maxage=N` without `public`, heuristic caching, and fronting-VCL cache overrides; conversely, `no-cache`, `max-age=0`, and zero/no-store CDN directives do not fit the positive-freshness rule cleanly. Use a conservative typed condition that emits only on responses conclusively `private`/`no-store`, or make the delivery VCL inject/strip timing. Rollout needs a MISS->HIT replay test, not only one pass-through request.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,406 @@
...
+`Cache-Control` with positive freshness, or any surrogate/CDN cache directive): a
+shared cache would replay one request's timings for the object's full TTL. The
+long-lived immutable `tsjs` asset route is the concrete case. The snapshot and the
+telemetry row are unaffected by this skip.
+
+The Axum adapter applies the same rule at its equivalent last point before response
```
jevansnyc: Axum cannot currently apply semantically equivalent phase boundaries. It builds app state once at startup, buffers the full upstream body inside its HTTP client, and buffers publisher/auction work before finalization. Its `/health` route is also inside middleware, and router-generated 404/405 responses may bypass finalization. Either document adapter-specific meanings/exclusions and expected missing phases, add an Axum terminal hook/earlier snapshot, or remove Axum parity from v1.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,406 @@
...
+## 8. Geo lookup dedupe (rider)
+
+Today every dispatched request pays two geo hostcalls for one answer: request-phase in
+`build_ec_request_state` (`app.rs:410`) and response-phase in
+`FinalizeResponseMiddleware` (`middleware.rs:83`, alternate site `main.rs:285`).
+
+Plain request extensions cannot carry the result out: the middleware moves the request
+context into `next.run(ctx)` and holds only the response afterward. The resolved geo
+travels instead through the response path, either on the existing `EcFinalizeState`
+response extension or a dedicated response extension installed by the dispatch layer,
```
jevansnyc: Please require the dedicated response extension rather than allowing `EcFinalizeState` as an alternative. Asset fallback performs `build_ec_request_state` (including geo) and then intentionally returns without `EcFinalizeState`; using that state for geo would repeat the hostcall. Attach `GeoLookupState` on every exit after an attempted lookup, and time legitimate finalizer fallback lookups so admin/batch/error paths do not hide degraded geo time inside `unattributed_ms`.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,406 @@
...
+Kept columns: `event_ts`, `method`, `status`, `time_elapsed_ms` (defined as the
+`mark_headers_ready()` snapshot), `cache_state`, `country`, `sample_rate`,
+`event_date`, 30-day TTL.
+
+Removed: raw `path`. Route identifiers like `/_ts/admin/ec/{id}` would otherwise put
+EC identifiers into a 30-day dataset, and publisher paths carry unbounded cardinality.
+Replaced by `route_template`: for named routes, the matched route-table pattern
+verbatim (parameters stay as placeholders); for the publisher fallback, the bounded
+normalized form produced by the existing auction-telemetry path normalizer. A
+serialization test proves a literal EC identifier never appears in an emitted row.
```
jevansnyc: The existing auction path normalizer does not make arbitrary publisher paths privacy-safe or low-cardinality. It redacts obvious long numeric/UUID/token segments but preserves short account IDs, usernames, email addresses, search terms, and arbitrary slugs verbatim. Use a coarse fixed publisher template or an allowlisted/template matcher, and add adversarial path tests rather than testing only the named admin EC route.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,406 @@
...
+`request_elapsed_ms`    Nullable(UInt32),
+`resp_bytes`            Nullable(UInt64),
+`template_cache_state`  LowCardinality(Nullable(String)),  -- x-ts-template-cache value
+`ts_version`            LowCardinality(String),
+`pop`                   LowCardinality(Nullable(String))   -- FASTLY_POP
+```
+
+Null means the phase did not run. `route_class` is assigned by the adapter at handler
+selection, not by path regex. `ts_version` plus `pop` is not globally unique across
+deployments; `publisher_domain` plus `env` provides row identity.
```
jevansnyc: `publisher_domain + env` is not unique across Fastly services, and version numbers are service-local. Add `service_id` or another immutable deployment identifier. Also define trusted typed sources: production has no `x-ts-env` value, operator headers can override managed response headers, and `template_cache_state` should come from a typed response extension rather than the public header. The source/meaning of `cache_state` also needs definition because front-cache hits may bypass the guest entirely.

jevansnyc (commented):
Round 2: the first-round corrections materially improve the design. The freeze point, stored snapshot, template-cache naming, rollback plan, and confirmed ingestion response are now much clearer. I still see three implementation-blocking issues: buffered publisher paths invalidate the body-phase math, the Tinybird schema/query model is invalid as written, and the cache-replay exclusion is not conservative enough. The remaining inline comments cover geo/context propagation, privacy, adapter parity, validation, and operational behavior.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,406 @@
...
+
+Sorting key: `(event_date, publisher_domain, route_class, pop, status)`, aligned with
```
jevansnyc: Blocking Tinybird schema issue: `pop` is `Nullable(String)` but is included in `ENGINE_SORTING_KEY`; Tinybird sorting-key columns must be non-nullable. Make it non-null with an `unknown` sentinel or remove it from the key. The layout also omits `env` and `event_ts` even though Grafana filters by time; either include the commonly filtered tenant/environment/time dimensions in an appropriate order or add an `event_date` predicate, then validate with `EXPLAIN` before committing to raw-table queries. See https://www.tinybird.co/docs/sql-reference/performance/nullable-columns

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,406 @@
...
+the analysis dimensions. This replaces the reserved key
+`(event_date, path, status, method)`. Rollout step 4 verifies whether the reserved
+datasource was ever deployed to the remote workspace; if it was, this schema ships as
+a versioned replacement datasource with a cutover, not an in-place edit.
+
+## 10. Emission mechanics
+
+- Point: Fastly `main.rs`, strictly after `send_edgezero_response` returns (the body
```
jevansnyc: The post-send builder needs a surviving data and transport context. The request is consumed by router dispatch, the response/extensions are consumed by `into_parts()`, and `RuntimeServices` currently survives only through `EcFinalizeState`, which asset/admin/error paths may not carry. Define an `AccessTelemetrySnapshot` built before `into_parts()` containing method, status, route metadata, trusted dimensions and typed states, plus a transport context available to every scoped response. `send_edgezero_response` should return byte count and success/partial/error outcome rather than logging failures and returning `()`.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,406 @@
...
+  (validation `0.0..=1.0` already exists). No client stickiness. Sampled-out requests
+  are silent; every other drop (row build failure, send failure, non-2xx) logs a
+  warning naming the reason, so lost diagnostic rows are visible.
+- Transport: one NDJSON row to the Tinybird Events API: same `api_host`, reserved
+  `access_dataset` and `access_token_secret`, 2 s first-byte and between-bytes
+  timeouts, `max_body_bytes` guard, no retry.
+- Delivery confirmation: unlike the auction sink, which starts `send_async` and drops
+  the pending response (it runs before delivery completes and cannot afford to wait),
+  the access emitter runs after the client has the full response and therefore awaits
+  the bounded ingest response and validates 2xx. A non-2xx or timeout logs a warning
```
jevansnyc, line 389: At sample rate 1.0, a Tinybird outage causes one backend request and potentially one warning per client request, with each sandbox remaining alive for the timeout. This consumes backend-request/dynamic-backend capacity and can create a log storm even though client delivery is complete. Define rollout abort thresholds, warning suppression/circuit-breaking behavior, and resource-limit monitoring. Also specify ordering relative to pull-sync: snapshot post-body fields immediately after send, but do not accidentally include pull-sync in `request_elapsed_ms` or delay it behind a multi-second telemetry await. See https://www.fastly.com/documentation/reference/compute/errors/

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,406 @@
...
+  with the status. This is what makes the warning semantics real rather than
+  aspirational.
+- Budget: at `access_sample_rate = 1.0` this adds one backend request per request to
+  the service, after delivery. The sample rate is the budget control; 1.0 is a
+  diagnosis setting, not a steady state.
+- Settings: the guard that rejects `tinybird.access_enabled = true` "until an emitter
+  is wired" flips into a wiring test in the same change. `access_enabled = true`
+  requires `tinybird.enabled` and a non-empty `api_host`.
```
jevansnyc: Validate every field the access emitter uses when enabled: `secret_store`, `access_dataset`, `access_token_secret`, API host, and body limit. Decide whether `access_enabled = true` with `access_sample_rate = 0` is a valid armed-but-silent state or a configuration error. Also note that the required `tinybird.enabled` currently activates the auction sink, so access-only telemetry is not actually independent and may unexpectedly enable auction emission. Split the master flags or document/test the coupling.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,406 @@
...
+  separate body-phase panel and never stack with pre-header phases.
+- PoP split, `ts_version` overlay, template-cache state rates.
+- Stall panel: rows with `request_elapsed_ms > 500` (post-body total, so body-only
+  stalls are caught) grouped by dominant phase, where `unattributed_ms` competes as a
+  phase so the panel cannot confidently blame a small measured span while most time is
+  uninstrumented.
+
+Sampling semantics for every aggregate: `sample_rate` must be operationally stable
+within any queried window. Panels either filter to a single `sample_rate` value or
+apply inverse-probability weights (`count() * 1/sample_rate`) for volume panels;
```
jevansnyc: The inverse-probability estimator is `sum(1.0 / sample_rate)` per emitted row, not `count() * 1/sample_rate` unless the query is already grouped or filtered to one rate. For quantiles, require a strict single-rate filter or specify a supported weighted-quantile function and its weight conversion.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,406 @@
...
+
+## 14. Testing
+
+- Core unit tests: phase accumulation, saturating math, `mark_headers_ready()`
+  idempotence and both-surface consistency, render format (one decimal, omission of
+  unrecorded phases), row serialization shape.
+- Adapter tests (Fastly via Viceroy, Axum native): header present and well-formed on a
+  publisher route with the flag on; absent with the flag off; absent on a
+  shared-cacheable response (tsjs route) with the flag on; exactly one header across
+  all send paths; append preserves a pre-existing Server-Timing value; `ts-kv`
```
jevansnyc: `exactly one header` conflicts with append semantics plus preservation of an origin-supplied `Server-Timing`: the final map may legitimately contain multiple field values. Assert exactly one TS-owned metric set (for example, one `ts-total`) and preservation of every pre-existing value, rather than exactly one physical header field.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 5557ff8d: prospect-a.example.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 5557ff8d: reworded to 'strongly suggests wall-clock time outside active guest CPU' with dependency awaits as the leading explanation and platform scheduling/hostcall queueing as residuals, and the measurement point named as the fronting delivery layer's time-elapsed entry observed at its deliver phase.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Confirmed against the shared-template miss path (buffer_publisher_response_async runs before the response returns, so the wait completes pre-header). Fixed in 5557ff8d: the row carries body_mode (streamed|buffered) and auction_wait_placement (pre_header|in_stream|none), the recorder takes placement explicitly, derivations are conditional per mode, and body-mode tests cover ordinary streaming, the buffered template miss, and Axum.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 5557ff8d: the formula is now inline prose (no bullet break), every query-time term is wrapped in coalesce(column, 0), subtractions in greatest(..., 0), and query tests cover sparse phase combinations and both placements.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 5557ff8d: the predicate inverts to conservative, header emission only when Cache-Control is conclusively private or no-store via the existing cache_control_headers_are_private_or_no_store predicate; bare max-age, s-maxage, heuristic-cacheable, and everything else emit nothing. Excluded routes still report through the row. Rollout step 2 adds the MISS-then-HIT replay test alongside the pass-through check.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 5557ff8d: new section 8a defines Axum semantics honestly instead of claiming parity: ts-appbuild absent (startup state), body_mode always buffered with pre_header placement, a terminal middleware freeze point covering router-generated 404/405, and /health excluded by route match. Cloudflare/Spin stay collection-only.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 5557ff8d: dedicated GeoLookupState response extension required, attached on every exit that attempted a lookup including the asset fallback; EcFinalizeState is explicitly rejected as a carrier for exactly the reason you gave. Legitimate finalizer fallback lookups are timed into ts-geo so degraded geo cannot hide in unattributed_ms.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 5557ff8d: the auction normalizer is explicitly ruled out for publisher paths. Publisher fallback uses a coarse fixed template (first allowlisted segment plus /* when deeper); named routes use the route-table pattern verbatim. Tests are adversarial: EC id on the admin route, an email in a segment, search-term segments, overlong segments.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 5557ff8d: service_id (FASTLY_SERVICE_ID) added as the immutable deployment identity; env sourced from typed settings state, template_cache_state from the typed response extension, never from public headers; cache_state dropped entirely with the guest-visibility rationale documented (fronting-cache hits produce no row by construction).

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 5557ff8d: every dimension column is non-nullable with an unknown sentinel; sorting key is (event_date, service_id, publisher_domain, env, route_class, pop, status); panel queries carry $__timeFilter(event_ts) plus an event_date predicate; rollout step 4 validates the panel queries with EXPLAIN before the dashboard is committed.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 5557ff8d: AccessTelemetrySnapshot is built unconditionally at the freeze point before into_parts(), carrying method, status, route metadata, typed dimension states, and the transport context, precisely because request, response, and EcFinalizeState do not all survive on every path. send_edgezero_response now returns bytes written plus success/partial/error instead of ().

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Addressed in 5557ff8d with one honest disagreement: a cross-request circuit breaker is not implementable on per-request isolates (no shared state survives), so the spec names that as a non-goal and defines the real controls: the 2 s bounded await, one warning per request, access_sample_rate pushed by config as the abort lever, and ingestion-freshness monitoring on the Tinybird side for quarantine/schema rejection. Ordering is specified: request_elapsed_ms snapshots immediately after send, pull-sync runs unchanged next, telemetry emission last, so pull-sync is neither delayed nor included.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 5557ff8d: full validation matrix when access_enabled (api_host, secret_store, dataset, token secret, positive max_body_bytes); armed-but-silent (sample_rate = 0) is a configuration error; and the coupling you identified is resolved by demoting tinybird.enabled to transport-only and adding tinybird.auction_enabled (default true, preserving existing configs) so access and auction emission are independently switchable, with a two-direction decoupling test.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 5557ff8d: the estimator is stated as sum(1.0 / sample_rate) over emitted rows, count()/rate valid only under a single-rate filter, and quantile panels require a strict single-rate filter.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in 5557ff8d: the test asserts exactly one TS-owned metric set (a single ts-total) and preservation of every pre-existing Server-Timing value, rather than one physical header field.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,524 @@
...
+  stored value as `ts-total`; the telemetry row reads the same stored value as
+  `time_elapsed_ms`. The two surfaces cannot disagree, and the row stays correct when
+  the header flag is off. Full request duration is captured separately as
+  `request_elapsed_ms`, snapshotted immediately after the body-stream drive returns and
+  before any other post-send work, so pull-sync and telemetry emission are never
+  included in it.
+- Sharing: `RequestTimings` is a cheap-clone handle, `Arc<Mutex<Inner>>`. It crosses
+  three boundaries: adapter entry to core handlers, the streaming body closure (records
+  body-phase spans after the response object has been handed off), and the adapter's
+  post-send emission read. A poisoned or contended lock must never fail a request: all
```
ChristianPavilonis: **P1: Contended samples will block rather than drop**

`std::sync::Mutex::lock()` waits on contention. The implementation plan's `if let Ok(inner) = lock()` only handles poisoning, so it does not satisfy this requirement. Specify `try_lock()` here and in the plan if contended samples must be discarded rather than delaying the request.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,524 @@
...
+| `ts-geo`            | geo hostcall (single after dedupe; accumulates if any path still repeats)             | `build_ec_request_state` (`app.rs:410`) and timed finalizer fallback lookups            |
+| `ts-kv`             | EC identity KV operations before response send (see enumeration below)                | the shared KV abstraction                                                               |
+| `ts-origin`         | publisher backend send to response headers available (read-through cache hit or miss) | `publisher.rs` around the origin `send`                                                 |
+| `ts-template-cache` | template cache `lookup_or_reserve`, including the hit-path full-body read             | `publisher.rs` around the lookup                                                        |
+
+Naming follows the completed template-cache terminology migration (`x-ts-template-cache`
+is the emitted header on `main`; `c2` naming is retired).
+
+`ts-kv` is instrumented at the shared identity-graph/KV abstraction
```
ChristianPavilonis: **P1: There is no single KV abstraction covering this taxonomy**

EC graph operations use `KvIdentityGraph`, while consent persistence uses `PlatformKvStore` through `RuntimeServices`. Graphs are also constructed independently in request setup, identify, admin lookup, batch sync, and finalization. The plan currently adds timings only to `KvIdentityGraph` and mentions construction in `main.rs`, which misses consent operations and several graph paths. Define both a timed identity-graph construction path and a `PlatformKvStore` timing decorator, with pull-sync explicitly using an untimed instance.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,524 @@
...
+## 8a. Adapter phase semantics
+
+The Fastly adapter is the reference implementation of the taxonomy. Axum differs
+structurally and its emissions are defined accordingly rather than pretending parity:
+
+- `ts-appbuild` is absent: Axum builds application state once at startup.
+- `body_mode` is always `buffered`: the Axum HTTP client buffers upstream bodies, so
+  `stream_ms` measures buffered-body write-out and `auction_wait_placement` is always
+  `pre_header`.
+- The freeze point is an Axum-terminal middleware layer that runs after all response
```
ChristianPavilonis: **P1: Router middleware cannot provide this Axum freeze point**

Axum passes `RouterService` directly to `AxumDevServer`. Router-generated 404/405 responses bypass router middleware, as the Fastly middleware already documents, and middleware returns before Axum serializes or writes the body. This design needs an outer service wrapper in the Axum adapter or a narrower v1 scope that excludes router-generated responses and body write-out timing.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,524 @@
...
+
+Kept columns: `event_ts`, `method`, `status`, `time_elapsed_ms` (defined as the
+`mark_headers_ready()` snapshot), `sample_rate`, `event_date`, 30-day TTL.
+
+Removed: raw `path`. Route identifiers like `/_ts/admin/ec/{id}` would otherwise put
+EC identifiers into a 30-day dataset, and publisher paths carry unbounded cardinality
+and user-generated content (search terms, usernames, emails in slugs). Replaced by
+`route_template`:
+
+- Named routes: the matched route-table pattern verbatim, parameters left as
```
ChristianPavilonis, line 285: **P1: The matched route pattern does not survive dispatch**

The entry point retains the raw method and path, but not the matched `NAMED_ROUTES` pattern. A handler enum cannot reconstruct it because some handlers serve multiple patterns and aliases. Add a typed route-metadata response extension, populated by the selected route handler and consumed at the freeze point.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,524 @@
...
+  route, an email address in a path segment, search-term-shaped segments, and
+  overlong segments must all normalize to bounded, content-free templates.
+
+Added columns (all dimension columns non-nullable with an `unknown` sentinel, because
+ClickHouse sorting keys cannot contain nullable columns):
+
+```
+`service_id`            LowCardinality(String),      -- FASTLY_SERVICE_ID; immutable deployment identity
+`publisher_domain`      LowCardinality(String),      -- matches auction schema
+`env`                   LowCardinality(String),      -- from typed settings state, never from a response header
```
ChristianPavilonis: **P1: `Settings` has no typed environment field**

The current settings model cannot supply this value. Fastly only derives staging from `FASTLY_IS_STAGING` while finalizing headers. Add and validate an environment setting, or revise the source contract to use an adapter-owned Fastly environment value rather than typed settings state.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,524 @@
...
+`filter_ms`             Nullable(UInt32),
+`geo_ms`                Nullable(UInt32),
+`kv_ms`                 Nullable(UInt32),
+`origin_ms`             Nullable(UInt32),
+`template_cache_ms`     Nullable(UInt32),
+`auction_wait_ms`       Nullable(UInt32),
+`stream_ms`             Nullable(UInt32),
+`request_elapsed_ms`    Nullable(UInt32),
+`resp_bytes`            Nullable(UInt64),
+`template_cache_state`  LowCardinality(String),      -- from the typed response extension, not the public header
```
ChristianPavilonis, line 316: **P1: The typed template-cache state does not exist**

`TemplateCacheResponseState` is private to `publisher.rs`, and `set_template_cache_response_state` writes only `x-ts-template-cache`. The companion plan does not add a response extension or modify this path. Define a typed extension here and require every state transition to update both the managed header and the extension.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,524 @@
...
+reserved key `(event_date, path, status, method)`. Rollout step 4 verifies whether the
+reserved datasource was ever deployed to the remote workspace; if it was, this schema
+ships as a versioned replacement datasource with a cutover, not an in-place edit.
+
+## 10. Emission mechanics
+
+- `AccessTelemetrySnapshot`: built unconditionally at the freeze point, before
+  `into_parts()` consumes the response. It captures method, status, route metadata
+  (`route_class`, `route_template`), typed dimension states (`env`,
+  `template_cache_state`, geo country), and the transport context needed to emit
```
ChristianPavilonis: **P1: The freeze point has no emission services on every path**

`send_edgezero_response` currently receives only the response and filter effects. `RuntimeServices` is built inside route handlers and survives through `EcFinalizeState` only on some paths, so asset, admin, and error responses cannot call the planned access emitter. The proposed snapshot type in the plan also omits transport/services. Define a cloneable emission context that is attached to every dispatched response, or retain adapter services from entry through post-send.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
```diff
@@ -0,0 +1,524 @@
...
+  (backend spec, secret store name, dataset, token secret). It exists because nothing
+  else survives to post-send on every path: the request is consumed by dispatch, the
+  response by `into_parts()`, and `EcFinalizeState` is absent on asset, admin, and
+  error paths.
+- `send_edgezero_response` returns the delivery outcome (bytes written and
+  success/partial/error) instead of `()`; the outcome feeds `resp_bytes` and lets a
```
ChristianPavilonis: **P1: Fastly cannot report these delivery states uniformly**

The streaming branch can count successful writes and detect truncation with a counting writer. The buffered branch calls `send_to_client()`, which returns no delivery result, so it cannot distinguish success, partial delivery, or error. Define separate observable semantics for buffered and streamed bodies, and specify how buffered byte counts are captured before consuming `EdgeBody`.

ChristianPavilonis (changes requested):
## Summary

The observability goal and Fastly freeze point fit the project, but the design and companion implementation plan assume several carriers and adapter hooks that the current code does not provide. The inline findings identify the architectural changes needed before implementation can proceed without losing routes or reporting misleading data.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in a845f007: access is exclusively try_lock() in both spec and plan; contention and poison both drop the sample instead of waiting.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in a845f007 with your mechanism: a TimedKvStore decorator implementing PlatformKvStore wraps the store handed to request-scoped consumers; every request-path KvIdentityGraph construction site (request setup, identify, admin lookup, batch sync, finalization) receives the timed store; consent-store access uses the same decorator; pull-sync explicitly constructs from the untimed store, with a test asserting it records nothing. One clarification now codified in the spec: the decorator measures store-call latency only, reads no payload, and the emitted surfaces carry no consent or identity data, so this feature needs no consent gate; the consent-store coverage here is about timing that I/O, not about requiring consent to measure.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in a845f007: the Axum freeze point is an outer service wrapper around the RouterService inside AxumDevServer, not router middleware, so router-generated 404/405 and body serialization are covered; /health excluded by path match inside the wrapper. Plan Task 10 updated to match.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in a845f007: a typed RouteMetadata response extension { route_class, route_template } is attached by each named-route handler wrapper with its matched route-table pattern verbatim (multi-pattern handlers attach the one that matched), fallback and tsjs handlers attach class plus coarse template, and the freeze point consumes the extension. No handler-enum reconstruction.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in a845f007 with your second option: env is adapter-owned, derived from the same FASTLY_IS_STAGING input that drives x-ts-env, values production|staging|unknown. Settings gains no environment field.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in a845f007: TemplateCacheResponseState becomes a typed response extension, and every transition that writes x-ts-template-cache sets the header and the extension together so they cannot drift; the row reads the extension. Plan Task 7 now includes the publisher.rs change.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in a845f007: the emitter's transport context is adapter-owned and route-independent; the Events API target derives from settings once at entry in main.rs and the send uses the stateless FastlyPlatformHttpClient directly, so asset, admin, and error responses emit without RuntimeServices or EcFinalizeState. emit_access_event's signature in the plan takes the client and target, not RuntimeServices.

docs/superpowers/specs/2026-08-24-request-phase-timing-design.md
jevansnyc: Fixed in a845f007: per-mode delivery semantics. Streamed bodies get a counting writer reporting bytes plus Complete/Partial/Error; buffered bodies capture the byte count from the body length before send_to_client() and report complete-on-return with no partial detection, with body_mode keeping the regimes distinguishable in analysis.

ChristianPavilonis (approved):

PR #1070, opened 2026-08-25
Allow ESI shared-template caching behind basic auth

prk-Jr:
## Summary

- Allow ESI shared-template caching when Trusted Server has already validated the request Basic Auth credential at the edge.
- Preserve the fail-closed bypass for publisher-origin credentials and ambiguous repeated Authorization fields.
- Carry the server-only authentication decision through all four adapter middleware stacks without modifying the forwarded header.

## Changes

| File | Change |
| ---- | ------ |
| crates/trusted-server-core/src/auth.rs | Add the request marker, fail closed on repeated Authorization fields, and add marker lifecycle regressions. |
| crates/trusted-server-core/src/publisher.rs | Exempt exactly one edge-terminated credential from cache bypass and independently reject repeated values. |
| crates/trusted-server-adapter-fastly/src/middleware.rs | Pass the mutable request through Basic Auth middleware. |
| crates/trusted-server-adapter-axum/src/middleware.rs | Pass the mutable request through Basic Auth middleware. |
| crates/trusted-server-adapter-cloudflare/src/middleware.rs | Pass the mutable request through Basic Auth middleware. |
| crates/trusted-server-adapter-spin/src/middleware.rs | Pass the mutable request through Basic Auth middleware. |

## Test plan

- [x] cargo test-fastly && cargo test-axum
- [x] cargo test-cloudflare && cargo test-spin
- [x] All six target-specific Clippy gates
- [x] cargo fmt --all -- --check
- [x] Native core suite: 2,216 tests
- [x] Regression tests cover pass-through, edge-terminated, and repeated Authorization behavior
- [ ] JS tests
- [ ] JS format
- [ ] Docs format
- [ ] Manual staging validation

## Checklist

- [x] Changes follow CLAUDE.md conventions
- [x] No unwrap calls added to production code
- [x] No println calls added
- [x] New code has regression tests

No linked issue for this PR, per request.


crates/trusted-server-core/src/publisher.rs
```diff
@@ -4343,7 +4343,22 @@ pub async fn handle_publisher_request(
...
+    // A credential this edge already terminated is exempt. Basic auth runs as
+    // middleware ahead of routing, so reaching here on a gated path means the same
+    // handler already validated the request; every reader that can look the template
+    // up has satisfied that same credential. Without the marker the credential is
+    // pass-through to the origin and still disqualifies. See
+    // [`crate::auth::EdgeTerminatedAuthorization`].
+    let authorization_value_count = req.headers().get_all(header::AUTHORIZATION).iter().count();
+    let request_had_authorization = match authorization_value_count {
+        0 => false,
+        1 => req
```
ChristianPavilonis, line 4349: 🔧 **Bind the exemption to the Authorization value that passed authentication**

This checks only that one `Authorization` value exists and that the marker is present. Request filters run after `AuthMiddleware`, and DataDome can replace `Authorization` through `X-DataDome-request-headers` because that name is not forbidden by `is_forbidden_filter_header`. The replacement leaves this marker intact, so a publisher-origin credential can be treated as edge-terminated and reach the shared-template lookup/store path.

Please store a digest of the uniquely validated raw value in `EdgeTerminatedAuthorization` and require it to match here, or clear the marker whenever a later filter mutates `Authorization`. A regression should cover authentication followed by header replacement.

crates/trusted-server-core/src/auth.rs
```diff
@@ -11,6 +11,46 @@ use crate::settings::Settings;
...
+/// the credential is pass-through, and the template cache continues to refuse it.
+///
+/// # Invariants
+///
+/// The private field makes [`enforce_basic_auth`] the only code that can produce
+/// this marker. It grants shared-template eligibility to a request that would
+/// otherwise be refused, so being unforgeable outside this module is the whole
+/// point: a caller cannot assert "already authenticated" without having actually
+/// checked. [`enforce_basic_auth`] also clears any inherited marker before it
+/// decides, so the value can never outlive the check that produced it.
```
ChristianPavilonis: ⛏ **Nitpick: shorten the marker documentation**

Most of this repeats the function documentation and explains privacy that the type already enforces. The invariant fits in a few lines.

```suggestion
/// Marks a request whose single `Authorization` value Trusted Server validated.
///
/// The shared template cache may exempt this value from its normal authorization
/// bypass. [`enforce_basic_auth`] clears any existing marker before checking and
/// inserts a new marker only after successful authentication.
```

crates/trusted-server-core/src/publisher.rs
```diff
@@ -4343,7 +4343,22 @@ pub async fn handle_publisher_request(
...
     // below needs it, and an authorized response must never become a shared
     // template.
-    let request_had_authorization = req.headers().contains_key(header::AUTHORIZATION);
+    //
+    // A credential this edge already terminated is exempt. Basic auth runs as
+    // middleware ahead of routing, so reaching here on a gated path means the same
+    // handler already validated the request; every reader that can look the template
+    // up has satisfied that same credential. Without the marker the credential is
+    // pass-through to the origin and still disqualifies. See
+    // [`crate::auth::EdgeTerminatedAuthorization`].
```
ChristianPavilonis: ⛏ **Nitpick: keep the cache-gate explanation local and short**

The longer authentication argument already lives in `auth.rs`. This site only needs to explain the local distinction.

```suggestion
    // Capture cache eligibility before the origin send consumes the request. A marked,
    // uniquely valued Authorization header was already validated by edge auth;
    // unmarked or repeated values remain pass-through and bypass sharing.
```

crates/trusted-server-adapter-fastly/src/middleware.rs
```diff
@@ -124,8 +124,11 @@ impl AuthMiddleware {
 
 #[async_trait(?Send)]
 impl Middleware for AuthMiddleware {
-    async fn handle(&self, ctx: RequestContext, next: Next<'_>) -> Result<Response, EdgeError> {
-        match enforce_basic_auth(&self.settings, ctx.request()) {
+    async fn handle(&self, mut ctx: RequestContext, next: Next<'_>) -> Result<Response, EdgeError> {
+        // Takes the request mutably because `enforce_basic_auth` marks requests
+        // whose credential it consumed itself; the shared template cache gate
+        // reads that marker later.
+        match enforce_basic_auth(&self.settings, ctx.request_mut()) {
```
ChristianPavilonis: ⛏ **Nitpick: remove the repeated middleware comment**

The mutable request and call name already make this clear, while `enforce_basic_auth` owns the full explanation. I would remove this comment here and from the Axum, Cloudflare, and Spin copies.

```suggestion
        match enforce_basic_auth(&self.settings, ctx.request_mut()) {
```

ChristianPavilonis (approved):
## Summary

The edge-auth marker is carried consistently through all four adapters, and the focused cache and authentication tests pass. I found two medium-risk follow-ups around preserving the marker's meaning and documenting the new cache exception. The remaining comments are cleanup nitpicks for the repeated and overly long explanations.

### Operator documentation

`docs/guide/configuration.md:1612` still says authorization bypasses the template cache unconditionally. It should distinguish pass-through and repeated values, which still bypass, from one edge-validated Basic credential, which may now share a template. Please also document that the credential remains forwarded and that an origin using it must declare `Vary: Authorization`.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -4343,7 +4343,22 @@ pub async fn handle_publisher_request(
...
+    let authorization_value_count = req.headers().get_all(header::AUTHORIZATION).iter().count();
+    let request_had_authorization = match authorization_value_count {
```
aram356: ♻️ **refactor** - `request_had_authorization` no longer means what its name says: a marked single-credential request *does* have an `Authorization` header, yet the flag is `false`. The variable now means "authorization disqualifies template sharing", which is exactly the semantics the sibling two lines down spells as `cookie_disqualifies`. Renaming restores the symmetry:

```rust
let authorization_disqualifies = match authorization_value_count {
    0 => false,
    1 => req
        .extensions()
        .get::<crate::auth::EdgeTerminatedAuthorization>()
        .is_none(),
    _ => true,
};
```

Apply manually (cannot be auto-applied as a suggestion because the other two uses at `publisher.rs:4407` and `publisher.rs:4653`, and ideally the `template_cache_ttl` parameter at `publisher.rs:6006`, sit outside the diff hunks).

crates/trusted-server-core/src/auth.rs
```diff
@@ -24,22 +64,49 @@ const BASIC_AUTH_REALM: &str = r#"Basic realm="Trusted Server""#;
...
+/// call's own decision. Nothing else about the request is touched — in
+/// particular the `Authorization` header is left in place and still reaches the
+/// publisher origin.
+///
+/// That last point is a stated assumption: a credential this edge terminates is
+/// treated as reader-neutral, which holds unless the origin *also* authenticates
+/// on the same header. An origin that does so declares `Vary: Authorization`,
+/// which the template-cache store refuses as an uncovered `Vary` name. An origin
+/// that varies on `Authorization` without declaring it would defeat any HTTP
+/// cache, and is out of scope here.
```
aram356, line 73: 🤔 **thinking** - this claim is conditional on operator config: it is only true while the operator does not list `authorization` in `template_cache_vary`. Validation rejects `cookie` explicitly (`creative_opportunities.rs:476-481`) but permits `authorization`; covering it makes the store accept such responses and embeds raw credential bytes in cache-key material via `values_from(req.headers())`. No cross-serving results (keys then vary per credential value), so this is not exploitable, but the invariant the doc leans on is not enforced. Proposed hardening, which makes the doc claim true by construction:

```rust
if names.iter().any(|name| name.eq_ignore_ascii_case("authorization")) {
    return Err(
        "template_cache_vary must not include Authorization; shared templates are keyed \
         on the edge-terminated credential decision, not the header value"
            .to_string(),
    );
}
```

Apply manually (the fix lives in `creative_opportunities.rs` plus a validation test, outside this PR's diff).

crates/trusted-server-core/src/auth.rs
```diff
@@ -24,22 +64,49 @@ const BASIC_AUTH_REALM: &str = r#"Basic realm="Trusted Server""#;
...
+        let Some(handler) = settings.handler_for_path(path)? else {
+            if Settings::is_admin_path(path) {
+                return Err(Report::new(TrustedServerError::Configuration {
+                    message: format!("Admin path `{path}` has no configured handler"),
+                }));
+            }
+            return Ok(None);
+        };
+        handler
     };
```
aram356: ⛏ **nitpick** - the scoping block is unnecessary: non-lexical lifetimes already end the `path` borrow at its last use, so the function compiles identically without it, and the comment teaches future readers a borrow rule that does not exist (`handler` borrows `settings`, and `path` is dead well before `req.extensions_mut()`). Verified in a scratch worktree: with the block removed, `cargo check` passes on the fastly (wasm32-wasip1), axum, and cloudflare targets, and `cargo fmt --check` + `cargo clippy-fastly` are clean.

```suggestion
    let path = req.uri().path();
    let Some(handler) = settings.handler_for_path(path)? else {
        if Settings::is_admin_path(path) {
            return Err(Report::new(TrustedServerError::Configuration {
                message: format!("Admin path `{path}` has no configured handler"),
            }));
        }
        return Ok(None);
    };
```

aram356 (approved):
## Summary

Allows ESI shared-template caching when Trusted Server itself terminated the request's Basic Auth credential at the edge. The design is sound: the `EdgeTerminatedAuthorization` marker is unforgeable outside `auth.rs` (private unit field, cleared on entry, inserted only on the validated path), the publisher gate independently disqualifies repeated `Authorization` values even when a marker is present, the template-cache key's raw `path_and_query` is the same raw path the handler regex gated (so every reader of a gated key passed the same first-match handler and credential), and `template_fingerprint` hashing the complete typed settings means a credential rotation or gate removal rotates every key and orphans gated-era templates. Regression coverage on both sides of the seam is thorough.

> 1 of the inline comments below carries a one-click GitHub `suggestion` (use **Commit suggestion** to apply it as a commit on the PR branch). The remaining comments describe the fix in prose because the change touches lines outside the diff or other files.

## Non-blocking

### ♻️ refactor

- **`request_had_authorization` no longer means what its name says** - see inline at `crates/trusted-server-core/src/publisher.rs:4354`

### 🤔 thinking

- **The `Vary: Authorization` refusal claim is conditional on operator config** - see inline at `crates/trusted-server-core/src/auth.rs:76-81`

### ⛏ nitpick

- **The `handler` scoping block is unnecessary and its comment states a borrow constraint that does not exist** - see inline at `crates/trusted-server-core/src/auth.rs:97-110` (suggestion)

- 🌱 **The middleware → router → publisher marker seam is untested end-to-end** - `auth.rs` tests prove the marker is inserted, and the publisher tests inject it with `for_test()`. The seam between them, extensions surviving `ctx.into_request()` through each adapter's dispatch path, is exactly the contract the design depends on, and no test exercises it. One adapter dispatch-path test (the fastly `app.rs` full-dispatch test module or the parity suite already have the scaffolding) asserting that a gated ESI request stores on the cold pass and hits on the warm pass would lock the contract against a future edgezero change that rebuilds the request and drops extensions.

- 📌 **Path-regex basic auth is alias-sensitive, and this PR's premise leans on the gate** - pre-existing, not introduced here: handler regexes match the raw path (`^/secure` does not match `/%73ecure/x`) while most origins decode, so an unauthenticated request can fetch gated origin content through the proxy under a percent-encoded alias. The shared-template keys also use the raw path, so no stored gated template is readable cross-key and the caching addition itself is safe, but the doc's "every reader able to look up a template has authenticated" premise is only as strong as the gate. Worth a docs note that whole-site staging gates should use an alias-proof pattern (`^/`), which is the intended use case anyway.

## CI Status

- cargo fmt: PASS (required)
- cargo test: PASS (required)
- vitest: PASS
- integration tests: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS

crates/trusted-server-core/src/publisher.rs
prk-Jr: Fixed. EdgeTerminatedAuthorization now stores a SHA-256 digest of the exact raw unique Authorization value, and the publisher requires the current unique value to match it. A regression test confirms replacing the header after authentication bypasses cache lookup.

crates/trusted-server-core/src/auth.rs
prk-Jr: Shortened the marker documentation to the local invariant and noted that it is digest-bound to the authenticated header value.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Replaced this with a concise cache-eligibility comment covering the unchanged authenticated value versus unmarked, replaced, or repeated Authorization values.

crates/trusted-server-adapter-fastly/src/middleware.rs
prk-Jr: Removed this comment from the Fastly middleware and the equivalent Axum, Cloudflare, and Spin implementations.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Renamed the flag and related helpers to authorization_disqualifies throughout the cache lookup/store gate and its tests.

crates/trusted-server-core/src/auth.rs
prk-Jr: Template cache vary validation now rejects Authorization case-insensitively, with normal-case and mixed-case tests. The operator documentation was updated as well.

crates/trusted-server-core/src/auth.rs
prk-Jr: Removed the extra scope and explanatory comment; non-lexical lifetimes keep the borrow valid. The affected test and clippy gates pass.

prk-Jr:
Addressed the body-level review findings as well:

- Added a Fastly dispatch-seam regression covering AuthMiddleware → router → publisher cache behavior: cold `miss-stored`, warm `hit`, and one origin request.
- Documented raw-path regex alias sensitivity, including `path="^/"` for whole-site staging.
- Clarified Authorization forwarding/cache behavior and prohibited Authorization in `template_cache_vary`.

Verification completed: Fastly, Axum, Cloudflare, and Spin test suites; integration parity; formatting; and all adapter clippy aliases.

PR #1077, opened 2026-08-26
Reject disallowed first-party proxy targets before signing

### LINKED ISSUE #1035: Enforce proxy.allowed_domains when minting signed URLs at /first-party/sign
## Description

Security-review follow-up (finding TS-10). Scope of this issue is strictly domain-allowlist enforcement — rate limiting, auth, and Origin checks are explicitly out of scope.

`/first-party/sign` (`handle_first_party_proxy_sign`, `crates/trusted-server-core/src/proxy.rs:1644`) mints signed `/first-party/proxy?tsurl=…&tstoken=…` URLs for any absolute `http(s)` URL submitted via `GET ?url=` or JSON `POST`. The only host check at mint time is the `rewrite.exclude_domains` **denylist**. The `proxy.allowed_domains` **allowlist** (`crates/trusted-server-core/src/settings.rs:1567`) is consulted only later, at fetch time, in `proxy_with_redirects` (`crates/trusted-server-core/src/proxy.rs:1322`; redirect hops at `:1483`).

Consequence: the sign endpoint acts as a signing oracle for arbitrary off-list hosts. When an allowlist is configured the minted tokens fail later at fetch time, but nothing rejects them at mint time and operators get no signal that clients are requesting signatures for hosts outside policy.

## Proposed change

- In `handle_first_party_proxy_sign`, validate the normalized target host against `proxy.allowed_domains` using the same matching semantics as the fetch path (exact match or `*.suffix`), and reject non-allowlisted hosts with `403` before signing.
- Apply the check on both input forms (`GET ?url=` and JSON `POST`), including the protocol-relative (`//host/...`) branch, after normalization.
- Keep existing semantics otherwise: an empty `allowed_domains` list continues to mean open mode (no restriction), and the `rewrite.exclude_domains` denylist check stays as-is.
- Log rejections with the offending host so operators can see off-policy signing attempts.

## Done when

- `/first-party/sign` refuses to mint tokens for hosts not covered by a non-empty `proxy.allowed_domains`, on both GET and POST forms.
- Fetch-time enforcement in `proxy_with_redirects` is unchanged.
- Unit tests cover: allowlisted host, non-allowlisted host, `*.suffix` wildcard match, protocol-relative input, and empty-allowlist (open-mode) passthrough, for both GET and POST.
- `docs/guide/api-reference.md` documents that the allowlist is enforced at mint time as well as fetch time.

## Affected area

Core

Related: #414 (added the fetch-time allowlist), #982 (signing broker for opaque-origin creative iframes shares this surface).

ChristianPavilonis:
## Summary

- Reject valid HTTP and HTTPS targets at `/first-party/sign` when their parsed host does not match `proxy.allowed_domains`, before expiry or signature work begins.
- Keep dynamic images and iframes blocked after a signing `403`, while preserving the existing direct-load fallback for network errors, malformed responses, and other HTTP failures.
- Add GET and POST regression coverage and update operator documentation to describe signing, initial-fetch, redirect, and open-mode behavior.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/proxy.rs` | Reuses one parsed-host policy for signing and fetching, logs rejected hosts, and adds the fourteen-case GET and POST matrix. |
| `crates/trusted-server-core/src/error.rs` | Generalizes the allowlist violation wording beyond redirects. |
| `crates/trusted-server-core/src/settings.rs` | Documents the full allowlist policy and updates the open-mode log message. |
| `crates/trusted-server-js/lib/src/integrations/creative/proxy_sign.ts` | Returns explicit signed, fallback, and blocked outcomes. |
| `crates/trusted-server-js/lib/src/integrations/creative/dynamic_src_guard.ts` | Suppresses native resource assignment for blocked outcomes. |
| `crates/trusted-server-js/lib/test/integrations/creative/proxy_sign.test.ts` | Covers successful signing, policy rejection, malformed responses, and non-policy fallback. |
| `crates/trusted-server-js/lib/test/integrations/creative/image.test.ts` | Verifies a blocked assignment preserves the previous image source. |
| `crates/trusted-server-js/lib/test/integrations/creative/iframe.test.ts` | Verifies a blocked iframe source is not applied directly. |
| `docs/guide/api-reference.md` | Documents signing-time enforcement, the `403` response, and the actual `{ href, base }` response shape. |
| `docs/guide/configuration.md` | Describes allowlist matching across signing, initial fetches, and redirects. |
| `docs/guide/first-party-proxy.md` | Documents signing policy, browser rejection behavior, and open mode. |
| `trusted-server.example.toml` | Updates the `allowed_domains` comment to cover signing and fetching. |
| `CHANGELOG.md` | Records the security fix under Unreleased. |
| `docs/superpowers/specs/2026-08-26-first-party-sign-allowlist-enforcement-design.md` | Adds the approved behavior and completion criteria. |
| `docs/superpowers/plans/2026-08-26-first-party-sign-allowlist-enforcement.md` | Adds the approved implementation and validation checklist. |

## Scope

The change spans core signing policy, the creative browser runtime, regression tests, and maintained documentation because returning `403` earlier would otherwise make the browser fall back to the rejected raw URL. It does not change adapters, routes, authentication, CORS, DNS or IP policy, token format, expiry, dependencies, or the configuration schema.

## Closes

Closes #1035

## Test plan

- [x] Other: `cargo test-cloudflare`, `cargo test-spin`, parity tests, all target-specific Clippy gates, JavaScript build, documentation lint/build, and `git diff --check`

## Checklist

- [x] No `unwrap()` in production code; `expect("should ...")` is used where applicable
- [x] Uses logging macros, not `println!`
- [x] New code has tests

### DIFF
--- a/crates/trusted-server-core/src/error.rs
+++ b/crates/trusted-server-core/src/error.rs
@@ -76,8 +76,8 @@ pub enum TrustedServerError {
     #[display("Forbidden: {message}")]
     Forbidden { message: String },
 
-    /// A redirect destination was blocked by the proxy allowlist.
-    #[display("Redirect to `{host}` blocked: host not in proxy allowed_domains")]
+    /// A proxy host was blocked by `proxy.allowed_domains`.
+    #[display("Proxy host `{host}` blocked: host not in proxy.allowed_domains")]
     AllowlistViolation { host: String },
 
     /// Settings parsing or validation failed.
--- a/crates/trusted-server-core/src/proxy.rs
+++ b/crates/trusted-server-core/src/proxy.rs
@@ -1280,11 +1280,11 @@ fn append_ec_id(req: &Request<EdgeBody>, target_url_parsed: &mut url::Url) {
     }
 }
 
-/// Returns `true` when a redirect to `host` should be followed.
+/// Returns `true` when `host` is permitted by the proxy host policy.
 ///
 /// When `allowed_domains` is empty every host is permitted (open mode).
 /// When non-empty the host must match at least one pattern via [`is_host_allowed`].
-fn redirect_is_permitted<S: AsRef<str>>(allowed_domains: &[S], host: &str) -> bool {
+fn is_host_permitted<S: AsRef<str>>(allowed_domains: &[S], host: &str) -> bool {
     allowed_domains.is_empty()
         || allowed_domains
             .iter()
@@ -1352,7 +1352,7 @@ async fn proxy_with_redirects(
             }));
         }
 
-        if !redirect_is_permitted(redirect_policy.allowed_domains, host) {
+        if !is_host_permitted(redirect_policy.allowed_domains, host) {
             log::warn!(
                 "request to `{}` blocked: host not in proxy allowed_domains",
                 host
@@ -1513,7 +1513,7 @@ async fn proxy_with_redirects(
                 }));
             }
         };
-        if !redirect_is_permitted(redirect_policy.allowed_domains, next_host) {
+        if !is_host_permitted(redirect_policy.allowed_domains, next_host) {
             log::warn!(
                 "redirect to `{}` blocked: host not in proxy allowed_domains",
                 next_host
@@ -1673,7 +1673,8 @@ pub async fn handle_first_party_click(
 ///
 /// # Errors
 ///
-/// Returns an error if JSON parsing fails, the URL cannot be parsed, or the URL uses an unsupported scheme.
+/// Returns an error if JSON parsing fails, the URL cannot be parsed, the URL uses an
+/// unsupported scheme, the URL lacks a host, or the host violates `proxy.allowed_domains`.
 pub async fn handle_first_party_proxy_sign(
     settings: &Settings,
     _services: &RuntimeServices,
@@ -1745,6 +1746,21 @@ pub async fn handle_first_party_proxy_sign(
     if scheme != "http" && scheme != "https" {
         return Err(Report::new(TrustedServerError::Proxy {
             message: "unsupported scheme".to_string(),
+        }));
+    }
+
+    let host = parsed.host_str().ok_or_else(|| {
+        Report::new(TrustedServerError::Proxy {
+            message: "missing host".to_string(),
+        })
+    })?;
+    if !is_host_permitted(&settings.proxy.allowed_domains, host) {
+        log::warn!(
+            "sign request for `{}` blocked: host not in proxy.allowed_domains",
+            host
+        );
+        return Err(Report::new(TrustedServerError::AllowlistViolation {
+            host: host.to_string(),
         }));
     }
 
@@ -2208,8 +2224,8 @@ mod tests {
         build_asset_proxy_target_url, clear_s3_credentials_cache_for_tests,
         handle_asset_proxy_request, handle_first_party_click, handle_first_party_proxy,
         handle_first_party_proxy_rebuild, handle_first_party_proxy_sign, is_host_allowed,
-        proxy_request, rebuild_response_with_body, reconstruct_and_validate_signed_target,
-        redirect_is_permitted, stream_asset_body,
+        is_host_permitted, proxy_request, rebuild_response_with_body,
+        reconstruct_and_validate_signed_target, stream_asset_body,
     };
     use crate::cache_policy::{CachePolicy, EdgeCacheHeader};
     use crate::constants::{HEADER_ACCEPT, HEADER_X_FORWARDED_FOR};
@@ -2287,6 +2303,16 @@ mod tests {
             .header(http::header::CONTENT_TYPE, "application/json")
             .body(EdgeBody::from(body.to_string()))
             .expect("should build http post request")
+    }
+
+    fn build_proxy_sign_request(method: &Method, uri: &str, target: &str) -> HttpRequest<EdgeBody> {
+        if method == Method::GET {
+            let mut serializer = url::form_urlencoded::Serializer::new(String::new());
+            serializer.append_pair("url", target);
+            return build_http_request(method.clone(), format!("{uri}?{}", serializer.finish()));
+        }
+
+        build_http_post_json_request(uri, &serde_json::json!({ "url": target }))
     }
 
     fn build_http_post_streaming_request(uri: impl AsRef<str>) -> HttpRequest<EdgeBody> {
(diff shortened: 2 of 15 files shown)


trusted-server.example.toml
```diff
@@ -116,7 +116,7 @@ enabled = false
 
 [proxy]
 # certificate_check = true
-# Required for integrations.prebid.external_bundle_url and first-party proxy redirects.
+# Permits first-party proxy signing, initial fetch targets, and redirect targets.
```
aram356: ⛏ **nitpick** — The previous comment told operators this setting is required for `integrations.prebid.external_bundle_url`, and that requirement still holds: `validate_external_bundle_config` in `crates/trusted-server-core/src/integrations/prebid.rs` fails deploy validation when the bundle URL is configured and `proxy.allowed_domains` is empty or does not match the bundle host. With the mention dropped, an operator whose deploy fails on that check gets no hint from this template. The same omission applies to the `allowed_domains` **Purpose** line in `docs/guide/configuration.md` and the `Proxy::allowed_domains` doc comment in `settings.rs` — worth carrying the mention there too.

```suggestion
# Permits first-party proxy signing, initial fetch targets, and redirect targets.
# Also required when integrations.prebid.external_bundle_url is configured.
```

crates/trusted-server-js/lib/src/integrations/creative/dynamic_src_guard.ts
```diff
@@ -84,12 +86,19 @@ export function createDynamicSrcProxy<E extends ElementWithSrc>(
...
         if (!current || current.requestId !== requestId) return;
         assignments.delete(element);
-        const finalUrl = signed || raw;
-        if (signed) {
+        if (result.outcome === 'blocked') {
+          log.warn(`${options.logPrefix}: blocked dynamic ${options.resourceName} ${attr}`, {
+            raw,
+          });
+          return;
+        }
```
aram356, line 98: 🤔 **thinking** — The blocked outcome suppresses the native assignment for the setter/`setAttribute` interception paths, but elements that arrive via the MutationObserver (a creative injecting `` through `innerHTML`/`document.write`) already have the raw attribute natively set — the browser starts that load before signing resolves, and on `blocked` the guard returns leaving the rejected URL in place. The server-side refusal to sign and fetch is the real enforcement boundary, so this is a scope observation rather than a bug — and clearing unconditionally here would clobber the legitimately-preserved previous `src` that the image test pins down. Might be worth a short code comment (or a line in `first-party-proxy.md`) acknowledging that markup-inserted resources are not retracted.

crates/trusted-server-js/lib/src/integrations/creative/proxy_sign.ts
```diff
@@ -54,13 +61,13 @@ export async function signProxyUrl(raw: string): Promise<string | null> {
     });
     if (!resp.ok) {
       log.warn('tsjs-creative: sign HTTP error', resp.status);
-      return null;
+      return resp.status === 403 ? { outcome: 'blocked' } : FALLBACK;
```
aram356, line 64: 🤔 **thinking** — This keys the blocked outcome on status alone, so a 403 emitted by anything in front of the handler (WAF, bot manager, auth middlebox) is also treated as a policy rejection and the resource renders nothing, where it previously degraded to a direct load. Fail-closed is a defensible default for this fix; if it ever bites on a bot-manager-fronted deployment, checking for the policy error's response shape alongside the status would disambiguate.

crates/trusted-server-core/src/error.rs
```diff
@@ -76,8 +76,8 @@ pub enum TrustedServerError {
     #[display("Forbidden: {message}")]
     Forbidden { message: String },
 
-    /// A redirect destination was blocked by the proxy allowlist.
-    #[display("Redirect to `{host}` blocked: host not in proxy allowed_domains")]
+    /// A proxy host was blocked by `proxy.allowed_domains`.
+    #[display("Proxy host `{host}` blocked: host not in proxy.allowed_domains")]
```
aram356, line 80: 🌱 **seedling** — `AllowlistViolation` falls through `user_message()`'s generic arm, so the now-documented 403 responds with the body "An internal error occurred". Since this PR promotes the 403 to a documented, client-interpreted policy signal, a specific safe message (e.g. "Forbidden: target host not permitted by proxy.allowed_domains") would ease creative and integration debugging. `user_message()` is outside this diff, so future work.

crates/trusted-server-core/src/proxy.rs
```diff
@@ -1748,6 +1749,21 @@ pub async fn handle_first_party_proxy_sign(
         }));
     }
 
+    let host = parsed.host_str().ok_or_else(|| {
+        Report::new(TrustedServerError::Proxy {
+            message: "missing host".to_string(),
+        })
+    })?;
```
aram356, line 1756: 📝 **note** — For `http`/`https` URLs the `url` crate rejects empty hosts at parse time (`EmptyHost`), so `host_str()` is always `Some` once the scheme check above has passed — this branch is defensively unreachable. Good practice to keep in security code; no action needed.

aram356 (approved):

prk-Jr (approved):

crates/trusted-server-core/src/proxy.rs
ChristianPavilonis: Agreed. Keeping the defensive host check unchanged.

crates/trusted-server-js/lib/src/integrations/creative/proxy_sign.ts
ChristianPavilonis: Keeping status-only handling here. The approved design treats a signing-endpoint 403 as fail-closed, while the server deliberately returns a generic body. Matching the body would require a new response contract and could turn an infrastructure 403 into a direct-load bypass.

crates/trusted-server-js/lib/src/integrations/creative/dynamic_src_guard.ts
ChristianPavilonis: Documented this limitation in `first-party-proxy.md` in a128575a6. The guide now distinguishes intercepted assignments from markup inserted through `innerHTML` or `document.write`, and states that a later 403 does not retract an existing attribute. Runtime behavior remains unchanged so a previous `src` is preserved.

trusted-server.example.toml
ChristianPavilonis: Addressed in a128575a6. The example now keeps the signing, initial-fetch, and redirect description while restoring the Prebid requirement. I also added the requirement to the `Proxy::allowed_domains` docs and configuration guide, including the empty-list deploy-validation behavior.

crates/trusted-server-core/src/error.rs
ChristianPavilonis: Leaving the generic body unchanged in this PR. The design explicitly preserves it, and the server warning already records the rejected host for operators. A client-visible policy response can be considered separately if a concrete debugging need appears.

Issue #1072, opened 2026-08-25
51Degrees Identifier (51Did) Edge Cookie vendor provider, registry code 51dd


jwrosewell:
This issue tracks the first real vendor Edge Cookie provider: the 51Degrees Identifier (51Did) provider, registry code `51dd` (reserved in [provider-code-registry.md](https://github.com/jwrosewell/trusted-server/blob/split/1-ec-provider/docs/superpowers/specs/provider-code-registry.md)).

## Shape

- A vendor crate under `crates/edgecookie/`, injected by the adapter, configured by an `[ec.providers.51did]`-style block that core captures as raw values and never interprets. Core stays vendor-neutral; the provider ships nothing into core.
- The provider works client-side by design, through the client-cycle path (#1046): the page script obtains the identifier and posts it to `POST /_ts/api/v1/ec/resolve`, and the provider verifies the payload before minting. This is the general pattern for any solution that supplies a web-browser-unique identifier, whether from a new web platform feature or from a user-installed extension, and the demo provider (`cfix`) exists to exercise exactly this round trip.
- Minted identifiers are `51dd~<value>`. Inside the envelope the value format is the provider's own, constrained only by the global identifier bounds (at most 256 bytes, alphabet `A-Za-z0-9._~-`), so binary payloads are carried base64url-encoded (standard base64 uses `+/=`, which are outside the alphabet).

## Acceptance bar

- Verification of the posted payload measured against the client-cycle spec's retained reservation design (audience binding, expiry, unique id, session binding, replay handling), which the spec keeps verbatim as the bar for the first vendor scheme.
- The page module performs the live permission check before vendor contact required by the client-cycle spec §4, and is injected only when the provider is selected.
- The verbatim round-trip proof every provider owes: mint, cookie, read-back, identity-graph key, and withdrawal, all under the `51dd~` code.
- A real-browser integration round trip in the browser suite.

## Declared interest

This is 51Degrees' own provider, filed by 51Degrees. The seam it uses is the same one available to every vendor, and the provider adds no vendor code to core, which is the pattern we are asking the project to hold every vendor to.

Produced with AI assistance under James Rosewell's direction; reviewed before filing.

PR #1047, opened 2026-08-19
Add the provider documentation set and finish the decomposition

jwrosewell:
> **This description has been edited.** The documentation now describes the
> provider selectors in the one configuration convention the whole stack lands,
> so several table and provider names have changed. If you read an earlier
> version of this text, re-read "What this pull request does". The CI matrix
> change and the two small code changes are the same.
>
> What changed in the pull request since the last version of this description:
>
> 1. The configuration reference and the Edge Cookie guide describe `[ec.hmac]`
>    and `[ec.host_signals]`, not `[ec.providers.hmac]` and
>    `[ec.providers.host-signals]`. The word `providers` is gone from the file
>    and every name a person types is snake_case.
> 2. The same pages describe the `implementation` line, which lets a deployment
>    run an implementation under a name of its own, and the startup refusals
>    that come with the convention, being a table the selector does not name, an
>    implementation this build does not have, and a setting a provider does not
>    know.
> 3. The example configuration carries a commented `[ec.host_signals]` block
>    rather than `[ec.providers.host-signals]`.
> 4. The page that states the convention on its own,
>    [`docs/guide/configuration-rules.md`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/docs/guide/configuration-rules.md),
>    lands with the integration seam (#1094), because that is where the
>    remaining provider types arrive. The pages here point at it.

Stacks on the client-set Edge Cookie value path (#1046), and depends on #1043
to #1046 together, because this pull request documents what they build. The
stack has six pull requests (#1043, #1044, #1045, #1046, #1047, #1094), each
targeting `main`, with this one fifth and the last of the five that decompose
the provider work in #838, as requested in the #986 review.
[Compare `split/4-client-resolve` with `split/5-response-hook-docs`](https://github.com/jwrosewell/trusted-server/compare/split/4-client-resolve...split/5-response-hook-docs)
to see only this pull request's change.

The design specs for the series are carried by the spec pull request (#1084),
including the two this pull request used to carry:

- [`2026-07-30-provider-migration-rollout-design.md`](https://github.com/jwrosewell/trusted-server/blob/split/6-integration-seam/docs/superpowers/specs/2026-07-30-provider-migration-rollout-design.md),
  the migration and rollout draft revised against the implemented series. Its
  sign-off ledger is kept intact, with each row the series implements marked
  with its pull request, so the task force can ratify rows rather than argue
  them again.
[...]

PR #1084, opened 2026-08-27
Add the integration provider seam design spec

jwrosewell:
> **This description has been edited.** The specs now describe one
> configuration convention for every provider type, and the auction side is
> part of it. If you read an earlier version of this text, re-read "The one
> configuration convention" and "The auction in the seam spec". The list of
> documents and the governance question below are unchanged.
>
>
> 1. Section 3.4 of the integration seam spec no longer says the auction stays
>    closed. The ad server, plain OpenRTB, Prebid Server and APS become
>    provider types that an integration registers, so the three things that
>    section recorded as closed are now open.
> 2. Every spec that shows configuration follows one convention. A provider
>    type is a top-level table, a `provider` key inside it selects what runs,
>    and `[<type>.<name>]` holds that provider's settings.
> 3. The spellings in the specs move with it. `[auction.providers]` becomes
>    `[demand]`, `[auction] mediator` becomes `[adserver]`, `profile` becomes
>    `implementation`, `[ec.providers.<name>]` becomes `[ec.<name>]`,
>    `[permission_signal] sources` becomes `[permission_signal] provider`, and
>    an integration is switched on by being named in `[integration] provider`
>    rather than by `enabled = true`.
> 4. Names are snake_case throughout, so `host-signals`, `client-fixed`,
>    `gpp-sale-opt-out` and `us-privacy` read `host_signals`, `client_fixed`,
>    `gpp_sale_opt_out` and `us_privacy`.
> 5. The migration and rollout spec records that the auction keys change
>    without a deprecation period, and the sign-off ledger rows that named the
>    old spellings name the new ones.
> 6. Each spec's revision record carries a row for the change, so a reviewer
>    can see what moved and why without reading the whole document again.

Carries the design specs for the provider series and for the integration seam,
seven documents and no code. It targets `main` directly, so the diff is only
those files, and the specs are reviewed here before the code that implements
them in #1043 to #1047 and #1094.

## The documents

| Spec | What it defines | Implemented by |
| --- | --- | --- |
| [`2026-08-27-integration-provider-seam-design.md`](https://github.com/jwrosewell/trusted-server/blob/split/6-integration-seam/docs/superpowers/specs/2026-08-27-integration-provider-seam-design.md) | Opening the integration registry, so a vendor integration ships in its own crate with its browser script, its deploy rules, and its identity, location, device, demand and ad server providers carried on its registration | #1094 |
| [`2026-07-30-pluggable-providers-design.md`](https://github.com/jwrosewell/trusted-server/blob/split/6-integration-seam/docs/superpowers/specs/2026-07-30-pluggable-providers-design.md) | Edge Cookie, device and location providers, selected by `[ec] provider`, `[device] provider` and `[geo] provider` | #1043 and #1044 |
| [`2026-07-30-permission-model-design.md`](https://github.com/jwrosewell/trusted-server/blob/split/6-integration-seam/docs/superpowers/specs/2026-07-30-permission-model-design.md) | The jurisdiction permission model, the `rules:` tree in the permissions file, and permission signal providers run in a configured order | #1045 |
| [`2026-07-30-client-cycle-ec-resolve-design.md`](https://github.com/jwrosewell/trusted-server/blob/split/6-integration-seam/docs/superpowers/specs/2026-07-30-client-cycle-ec-resolve-design.md) | Client-cycle Edge Cookie providers and the hardened resolve endpoint | #1046 |
| [`2026-07-30-provider-migration-rollout-design.md`](https://github.com/jwrosewell/trusted-server/blob/split/6-integration-seam/docs/superpowers/specs/2026-07-30-provider-migration-rollout-design.md) | Migration and rollout for the provider and permission model, with the series' sign-off ledger | #1043 to #1047 |
| [`2026-07-30-integration-response-header-hook-design.md`](https://github.com/jwrosewell/trusted-server/blob/split/6-integration-seam/docs/superpowers/specs/2026-07-30-integration-response-header-hook-design.md) | The response-header hook, kept as the design bar for its first consumer | Not implemented |
| [`provider-code-registry.md`](https://github.com/jwrosewell/trusted-server/blob/split/6-integration-seam/docs/superpowers/specs/provider-code-registry.md) | The append-only registry of the four-character codes that namespace Edge Cookie identifiers | #1043 |

## The one configuration convention

The specs used to show four ways of switching the same kind of thing on. An
integration had `enabled = true` inside its own block, a demand source was a
key under `[auction.providers]`, a permission signal was a name in a `sources`
list, and an Edge Cookie provider was `[ec] provider` naming a nested
`[ec.providers.<name>]` block. Every spec now shows one shape:

```toml
[<type>]
provider = "<name>"          # a string where one runs, a list where several run

[<type>.<name>]              # only when that provider has settings
setting = "value"
```

The types named across the specs are `ec`, `geo`, `device`,
`permission_signal`, `demand`, `adserver` and `integration`. A table's name is
the implementation it configures, unless the table carries an `implementation`
line, which is how two instances of one implementation run side by side under
names a deployment chooses. Names are snake_case.

The specs also state what this buys, which is the part a reviewer should
weigh. A table the `provider` line does not name stops startup rather than
sitting unused, an implementation this build does not have stops startup with
the ones it does have listed, a setting a provider does not know is refused
rather than dropped, and demand and ad server endpoints must be HTTPS, or HTTP
to a loopback host. The operator-facing version of the same rules lands with
the code as
[`docs/guide/configuration-rules.md`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/docs/guide/configuration-rules.md)
in #1094.

The spellings change in every document that shows configuration, which is the
pluggable providers spec (`[ec.<name>]` and `host_signals`), the permission
model spec (`[permission_signal] provider` with `gpp_sale_opt_out` and
`us_privacy`), the client-cycle resolve spec (`client_fixed`), the migration
and rollout spec, and the integration seam spec.

## The auction in the seam spec

Section 3.4 has been rewritten. The version before this one recorded the
auction as closed, because `main` had replaced the auction provider table with
a compiled plan (#1016) and the plan kept three things in core, being a fixed
list of three OpenRTB profiles, special cases for the `prebid-server` and `aps`
profiles by name, and `adserver_mock` as the only ad server the plan would
accept.

Section 3.4 now opens those three. The ad server, plain OpenRTB, Prebid Server
and APS become provider types that an integration registers on its builder, and
a deployment selects them with `[demand] provider` and `[adserver] provider`.
The compiled plan stays, and what it compiles changes from a fixed profile list
to whatever implementations the registered builders offer, so core names no
vendor. The APS row in section 4 and the corresponding row in section 9 no
longer say that an auction-side vendor needs a change to the plan. What they
now say is that an auction-side vendor needs its own crate and a registration,
on the same terms as an identity, location or device provider.

The bid renderer contract is unchanged from the previous revision. It is an
open descriptor, a type tag plus the payload the demand provider supplies, with
the same serialized form, so the response a page receives does not change.

The migration and rollout spec records the cost. The auction keys change with
no deprecation period, which is a deliberate choice for a major release, and a
configuration still carrying `[auction.providers]` or `[auction] mediator` is
refused with a message naming where the setting moved to.

## Governance question this raises

The seam spec's section 2 states a principle for the task force rather than for
the compiler. Tech Lab engineering owns core and reviews vendor crates, and
vendors ship and maintain their own integrations. That is the change that
removes the per-vendor core work the project pays for today, most recently the
LiveRamp module (#1054). Opening the auction side widens it, because a demand
source or an ad server is now a vendor crate too.

## Declared interest

51Degrees is a vendor and would use this seam for its own integration (#1072).
The reasoning here applies to every vendor on the same terms, ourselves
included, and our own provider follows the same route.


aram356 (changes requested):
## Summary

Thank you for this, and for §8 in particular. Publishing what implementing the seam actually found, including three items that say your own work is not finished, is the right instinct and it is what makes the design half of this trustworthy.

The design is accepted. §3.6's rule is correct, and it is the rule the work below implements. Things the host supplies are platform services. Things a vendor supplies are capabilities of that vendor's module. What is being changed is the **order** and the **structure**, not the direction.

Five changes to sequencing and structure follow. Tell me where the reasoning is wrong and I will change my view, but I want these settled before more code lands on the current shape. The two items that need wider sign-off are named at the end.

## 1. One `crates/trusted-server-integrations` crate first, with a discovering `build.rs`

Move all thirteen integrations into a single crate and generate the registration table at build time by scanning the crate's own directories, in the manner `crates/trusted-server-js/build.rs` already scans `dist/tsjs-*.js`.

This is the change that delivers the neutrality goal. Core stops naming vendors in `builders()`, in `validate_enabled_integrations`, and in `migration_guards.rs`, and it stops naming them because nobody writes the list, not because the list moved. Three of §1's four hardcoded vendor tables become generated output.

It also resolves several of §8's findings without further design:

- §8.2, the hand-carried module hash and its line-ending fragility, goes away because hashes are generated exactly as core's are today.
- §8.6, a module's `validate` running nowhere, is solvable in the generated table.
- §1.2, the compile-time JS map, is solved by one scan that finds Rust and TypeScript together.
- The `migration_guards.rs` drift is structurally impossible once generated. Worth noting that this drift is not hypothetical: `osano` is registered in `builders()` and has **no entry in `migration_guards.rs` at all** on `main`. Nobody noticed because both lists are hand-written.

`prebid` stays in core as protocol support, as §3.4 already says.

The runtime injection seam in §3.1 is not required for any of the thirteen integrations you have, because they are all compiled from source in this repository. It becomes the right mechanism when a genuinely external vendor crate exists, and it should be added then, with that vendor as its first consumer.

## 2. Crate layout follows from that

§4 proposes nine vendor crates, one PR each. We are not doing that now. One crate holds all thirteen integrations, and a vendor gets its own crate when that vendor takes over maintaining it, which none of them does today. Every integration on `main` imports the same roughly twelve crates, all of which core already depends on, so nothing about dependencies argues for splitting them either. Splitting later is a directory move against a module that already has clean boundaries.

Two consequences for what is in #1094 now:

- **Directory name equals package name, as everywhere else in the repository.** The crate is `crates/trusted-server-integrations`. `crates/geo/fastly` as package `trusted-server-geo-fastly` breaks that convention, and `crates/geo/` is a category folder holding one item.
- **Host code stays in its adapter.** `FastlyPlatformGeo` is 52 lines that already live at `crates/trusted-server-adapter-fastly/src/platform.rs:687`, and the Fastly adapter is its only possible consumer. Extracting it into a crate gains nothing and contradicts §3.6's own rule that host-supplied things are platform services. The same applies to the 103-line Fastly device provider. `crates/edgecookie/` contains only a README and should be deleted until a vendor crate lands. `seam-probe` is a test fixture and should not be a permanent workspace member.

## 3. Permissions policy belongs in `trusted-server.toml`; `permissions.yaml` should not exist

Policy belongs in a `[permissions]` section of `trusted-server.toml`, beside the existing `[consent]`. We should not add a second configuration format, and policy should not be compiled into the binary.

The permission-model spec's §3.1 gives the reason for YAML. The draft required policy to flow through the runtime config pipeline, and that activation apparatus does not exist. That conflates two things. `ts config push` **does** exist. It publishes `trusted-server.toml` as a blob envelope, and `TrustedServerAppConfig` already carries the entire `Settings` struct. What does not exist is the staged multi-revision activation protocol the draft designed, with fleet quiescence, admission leases, and a hash-linked journal. No other setting needs that machinery. The publisher domain, the proxy secret, the EC passphrase, and every integration config all push atomically today, and policy is not different in kind.

What the move gains:

- One place to configure everything. Today policy is the only control that bypasses `ts config validate` and `ts config push`, and it is the most compliance-sensitive one.
- The validation we already have, being `deny_unknown_fields`, `Validate`, and startup errors, instead of a bespoke parser reimplementing duplicate-key and case-collision checks.
- A publisher can correct a jurisdiction rule with a config push rather than rebuilding and redeploying a WebAssembly binary. For a compliance control that may have to change on a regulator's timeline, that difference is the point.
- §3.1's own recorded defect, that the `include_str!` path reaches above the crate root, disappears.

Alongside the move, please trim the fifty-three Data Uses that have no signal mapping and no enforcement point. Only two carry enforcement weight today. Moving fifty-three inert flags into `trusted-server.toml` would carry the misleading operator view across rather than fix it, and the spec already acknowledges this departs from its own rule that a permission appears only with both a mapping and an enforcement point.

## 4. The EC provider spec rebases onto 1 and 3

Rewritten against a seam that already exists and policy that lives in settings, the EC provider work is materially smaller than #1043 as it stands:

- No `RuntimeServices::ec_provider` slot to add and then delete, and no `resolved_ec_provider` second field. Note that #1094 currently has both, plus the registration path, so identity has three injection paths where §3.6 specifies one. §3.6's sentence that the `ec_provider` slot "goes" is not yet honored.
- No per-adapter resolution divergence. Related, and worth fixing wherever this lands: on #1094 `resolve_geo_provider` and `resolve_device_provider` return `Result` and fail loudly on a bad selector, while `resolve_ec_provider` returns a bare `Option` and falls back silently. §3.6 says an unresolvable selector is a startup error.
- **No double configuration migration.** This is the strongest reason for the reordering. As it stands, #1043 migrates operators from `[ec] passphrase` to `[ec.providers.hmac]`, and §3.6 with sign-off row 8 then migrates them again to `[integrations.<id>]`. Two breaking configuration changes in one release cycle, the second withdrawing the first. Reordered, operators migrate once, into the final shape.

## 5. Some of this belongs in edgezero, not Trusted Server

`edgezero-core` already owns the platform layer, being `app_config`, `blob_envelope`, `config_store`, `secret_store`, `key_value_store`, `store_registry`, the router, the middleware, and the four adapters. `Hooks::stores()` returning `StoresMetadata` is already the seam `run_app` uses to inject store registries. Before we build more platform machinery in this repository, we should be explicit about which side of that line each piece sits on.

- **The Fastly composition gap (§8 item 7) is probably an edgezero fix rather than a Trusted Server one.** `trusted-server-adapter-fastly` wraps `edgezero-adapter-fastly`. If a deployment composes its modules through `Hooks`, the pattern already exists upstream and no Trusted Server adapter needs a new library target. Adding a `[lib]` here would work around the absence of something that may already exist one level down.
- **`RuntimeServices` overlaps edgezero substantially.** It carries a config store, a secret store, a KV store, an HTTP client, and a backend. `edgezero-core` models all five. Adding identity, device, and geo provider slots to it builds up a second platform abstraction beside edgezero's rather than using edgezero's.
- **`PlatformGeo` has no edgezero counterpart.** There is no geo abstraction in `edgezero-core` today. Either a geo lookup is a host capability edgezero should expose the way it exposes a KV store, or it stays here. Whichever it is, that should be a stated call rather than a default that follows from where the code happens to have been written.
- **`TrustedServerApp` implements `Hooks::name()` and `Hooks::routes()` but not `Hooks::stores()`** (`crates/trusted-server-adapter-cloudflare/src/app.rs:332`). The store-injection seam edgezero already provides is unused here, which is worth resolving before we design another one.

Which of these should move is a question for whoever owns edgezero. I would rather ask it now than build a second mechanism here and answer it afterwards. It also makes change 1 smaller: if composition runs through `Hooks`, the integrations crate needs less new machinery than either this spec or #1094 assumes.

## Spec corrections, verified against the tree

- **Five of the seven specs are statused "Implemented"** for code that is not on `main`. `EdgeCookieProvider`, `PermissionSet`, and `resolve_from_client` are all absent from `origin/main`. Merging this set as written would tell every later reader that shipped behavior exists when it does not. Please status everything unmerged as "Proposed".
- **§4, "thirteen vendor files" in `migration_guards.rs`, is wrong.** There are eighteen vendor files across twenty `integrations/` entries; `nextjs` alone contributes six. Thirteen is the count of builders. And as above, `osano` has no entry at all, which §4's migration plan depends on.
- **§3.5, GPT diagnostics "called by name from all four adapters"**, undercounts. There are nine `prepare_request` call sites plus one `finalize_response`, so the hook work in §3.5 is larger than stated.
- **§8.3, "a provider is resolved more than once per request", does not hold on `main`.** No integration or proxy calls geo. A real double resolution does exist, but it is core against core: `handle_auction` at `auction/endpoints.rs:262` and the adapter's `build_ec_context` both call `lookup` on `POST /auction`. The underlying gap, that there is no per-request provider context, is real and worth keeping.
- **§8.5 understates the Spin gap.** Spin does not merely skip inline admin stubs; it skips all six first-party route bindings (proxy, click, sign, proxy-rebuild), which carry real traffic. Cloudflare, by contrast, has no health route and so covers everything.
- **Line references to correct:** `settings.rs:166` should be `:215`; `auction/mod.rs:49` points at the function header rather than the list; `publisher.rs:4361` should be `:4369`.

## §8 item 7 is the load-bearing finding

Your assessment is right and the cause is more fundamental than the text states. The Fastly adapter has **no `[lib]` section and no `src/lib.rs`**, and every module is declared bare `mod`, so nothing in the crate is externally reachable. This is not a visibility modifier to widen; it is a missing library target. Axum, Cloudflare, and Spin each expose `pub fn routes_with_settings` and `pub struct TrustedServerApp`; Fastly exposes neither.

As things stand the seam is proven on the dev server and unreachable on the primary deployment target. §6's acceptance criteria should require the round trip on Fastly specifically, not on any adapter. Under change 1 this stops blocking the immediate work, since in-tree integrations compile in and nothing needs runtime injection. It must still be fixed before the first external vendor crate, and per change 5 the fix may belong in edgezero rather than here.

## Disposition of the open pull requests

#1043 through #1047 and #1094 are **superseded rather than rejected**. The design in §3.6 is accepted, and most of the provider logic carries over onto the reordered base. Please do not read this as the work being discarded.

The order, with the first two independent and able to run in parallel:

1. One `crates/trusted-server-integrations` crate with a discovering `build.rs`, against `main`.
2. Permissions into `trusted-server.toml`, `permissions.yaml` deleted.
3. The EC provider work, rebased onto 1 and 2.
4. Device and geo selection, in the same shape.
5. Per-vendor crate extraction, if and when a vendor takes over maintaining one.

Happy to discuss who does what. The Fastly composition gap is core infrastructure rather than vendor contribution, and per change 5 it may not belong in this repository at all.

## The jurisdiction default should not name a country

`[geo] default_country` is required in every mode, and it must resolve to a rule in the policy file. There are three groups today, being `gdpr-eu`, `gdpr-uk`, and `us-opt-out`, so a deployment with global traffic and no geo provider has to nominate one country and then apply that country's law to every visitor. That is the wrong shape. It makes a neutral product take a position on whose rules apply, and the operator's honest answer is usually "I do not know where this visitor is", not "treat everyone as French".

The default should be a policy baseline rather than a country. Two forms worth considering, and I do not mind which:

- A named group, so an operator selects `gdpr-eu` or a strictest-available baseline directly without claiming a location.
- A per-permission baseline written inline, for an operator who wants to state exactly what is allowed with no signal.

Either way the value says what is permitted, not where the visitor supposedly is. The requires-signal floor already exists for a failed lookup, so the machinery for a location-free baseline is there.

This also removes most of what I was going to send to the task force. The UK row shipping `necessary.operations.storage: granted` while the EU requires a signal, and Australia mapping to `us-opt-out`, are still policy positions that need someone with the relevant expertise to confirm with a citation, and the shipped table should carry that citation. But nobody has to ratify a default that no longer asserts a jurisdiction.

The crate layout, the configuration format, and the merge order are not task force questions. They are engineering calls and we should settle them here.

One further item to fix wherever the permission work lands: §7 of the permission-model spec records that publisher navigation and page-bids attach `user.id` under the provider gate rather than the sharing pair, so a storage-only grant can still place the EC identifier in `user.id` on those paths while `/auction` pair-gates it. That is an open inconsistency in the sharing boundary, and it should be closed rather than tracked.

jwrosewell:
Thank you for this. It is a serious review and it moves the design forward. Taking your points in order.

**1. One integrations crate with a discovering `build.rs`.** Agreed as the destination. Generating the registration table removes a whole class of drift rather than one instance of it, and the hand-enumerated shape it replaces is the existing Trusted Server convention our PRs followed rather than introduced. It is out of scope for these PRs and the current specification though, so we suggest iterating toward it rather than blocking on it. The halfway house to move to now is the one that most helps code review, being vendors actually separated, with a hard build failure when a vendor change reaches into core. Reviewers then get a clean line between vendor changes and core changes today, and the generated registration lands as its own later step. The runtime injection seam is already built in #1094, so nothing there needs to wait for an external vendor.

**2. Crate layout.** Here we differ. Whether a vendor wrote the code, or the Tech Lab team wrote it for them, should not decide whether it gets its own crate. Separating per vendor now makes ownership legible, and a crate boundary is enforced by the compiler, while a CODEOWNERS entry only assigns review ownership and constrains nothing about what code can reach. The maintainer can be the Tech Lab team today. If the Tech Lab team built the Permutive integration, the Permutive crate is theirs until Permutive take it over, and the handover then changes one line, not the structure. We take the point that host code stays in its adapter, which §3.6 already says, and we will follow the repository's crate-naming convention. The disagreement is only on timing.

**3. Permissions.** This one is fundamental for us and not something we can change 🙂. Deleting the dedicated permissions file breaks a deployment model that is very hard to change later.

The transparency point first. A single, dedicated permissions policy that anyone can open and read is how Trusted Server shows exactly what a publisher permits, on its own rather than buried beside proxy secrets and backend hosts.

Then the mechanism. Trusted Server should not decide policy for the deployer, and there should be no automatic default. Selecting the policy is a conscious choice when an environment is set up for the first time. The entity that compiles a deployment provides the policy it wants as part of the build, which is why the policy has to be a build-time artifact and not only a runtime config section, and that same entity can expose configuration override so the operator of the compiled variant can adjust it without rebuilding. Consider a provider in Europe that compiles Trusted Server with a European base and ships it to its customers, who get an interface to tune it. Two actors and two moments, with the integrator setting the base at build time and the operator adjusting it at runtime. A single runtime config section cannot express that, and neither can a compiled-only policy, so both layers are required. The repository should also carry multiple policy files for different test scenarios, because exercising the permission model across scenarios is part of the robustness of the service.

Three things follow. Policy is not locked in the binary forever. The compiled base is the layer these PRs implement, and the operator override is the follow-on the permission-model spec already records as deferred, so a jurisdiction rule can change on a regulator's timeline without a rebuild. What the shipped jurisdiction rules should say, the UK storage row and the country baselines included, is a policy question for the group to confirm with citations, which is your point in the jurisdiction section too, and we would take it there rather than settle it in code review. And which file carries the compiled-in base and which one arrives through configuration at runtime is not something engineers need to debate. That choice sits with the people deploying, and testing proves every mutation of base and override works.

**4. The EC provider and one injection path.** Agreed. §3.6 specifies one identity injection path and we will honor it, collapsing to the single registration path. An unresolvable selector will fail loudly, while an absent selector keeps today's behavior of running with no Edge Cookie provider. Rebased onto a seam that already exists, the double configuration migration also goes, so operators migrate once.

**5. What belongs where.** The distinction that matters is which repository these live in. The geo provider, the device provider and the Edge Cookie provider are part of Trusted Server, not EdgeZero. EdgeZero supplies the host plumbing, and the provider capabilities are Trusted Server's domain, which is where a vendor goes to plug in. They are also deliberately off by default. Only User-Agent device detection runs by default, geo resolves no location until a provider is configured, and the Edge Cookie provider is selected by configuration, which is the same conscious-choice principle as point 3. Vendor code already lives in Trusted Server as the registered integrations, and the provider seam exists so a vendor can be added the same way. We are asking to use it for the purpose it was built. Where you are right is the platform plumbing rather than the providers. The config, secret and KV stores that `RuntimeServices` carries are things the host already models, and we should route those through the existing store seam rather than a second abstraction beside it.

**Spec corrections.** Taken, and thank you for checking against the tree.

- Statuses. Some specs are marked Implemented for code that is not yet on `main`. We marked them Implemented because the code is in the PRs, which was the wrong signal, and we will restatus everything unmerged as Proposed.
- Counts. These were taken from the builder registrations rather than a file count, which undercounts the real files and call sites. We will correct them against the tree.
- The `osano` entry missing from the migration-guard hand-list is real, and exactly the drift that change 1 removes structurally. There will be more of these as we iterate, which is the argument for generating the list.
- The Fastly composition point and §6's acceptance criteria. Agreed the round trip should be proven on the primary target, not only the dev server, and we will come back with the specifics.

**On merging.** We agree with most of this and will make the changes. The differences are narrow, being per-vendor separation now rather than later and keeping a dedicated permissions file. Rather than treat the current PRs as superseded and reset, could we agree a concrete sequence and owners for landing this so the provider work merges soon? We would like to keep this a direct back and forth to a merge. Thanks again for the depth of the review.

docs/superpowers/specs/2026-07-30-client-cycle-ec-resolve-design.md
```diff
@@ -0,0 +1,260 @@
...
+
+1. **Reject cross-site requests with an origin check.** v1 authorizes a
+   request only when its `Origin` header names the publisher's configured
+   domain or a subdomain of it. A missing or foreign `Origin` is rejected
+   with `403`. Browsers always send `Origin` on POST `fetch`, so its
+   absence means a non-browser caller, which has no business on a
+   page-script endpoint. The 2026-07-31 draft asked for an exact origin
+   allowlist as new configuration plus an optional session-bound CSRF
+   token. v1 derives the allowed set from `publisher.domain` (existing,
+   operator-trusted configuration) and admits the publisher's own
```
ChristianPavilonis: 🔧 **P1: Do not authorize every publisher subdomain to set identity**

Authorizing every origin below `publisher.domain` recreates the identity-fixation path for a delegated, user-controlled, or compromised sibling subdomain. Such a page passes this suffix check and can issue the simple credentialed POST described in the threat model, while the EC cookie is scoped to `.publisher.domain`. Control of the DNS namespace does not imply that every subdomain is a trusted identity-setting origin. Please use an explicit serialized-origin allowlist and match scheme, normalized host, and effective port exactly. Add rejection cases for a sibling subdomain, the wrong scheme, and the wrong port.

docs/superpowers/specs/2026-07-30-pluggable-providers-design.md
```diff
@@ -0,0 +1,653 @@
...
+  outbound response, and the client-cycle resolve path returns them, which
+  is how a client-side provider requests further evidence from the page.
+  The draft banned the field when nothing consumed it. The consumer landed
+  in the same series, satisfying the rule the ban enforced.
+- `IdentityInput.permissions` / `IdentityInput.consent`: **shipped and
+  populated.** The organic create path passes the request's resolved
+  permission state and consent context so a provider can read them for
+  behavior beyond gating. The gate itself has already run before `generate`
+  is called, so a provider cannot use the fields to authorize itself.
+- `required_permissions` on `DeviceProvider` and `PlatformGeo`: **present,
```
ChristianPavilonis, line 221: 🔧 **P1: Enforce or reject device-provider permission declarations**

`DeviceProvider::required_permissions` is public but deliberately has no enforcement point, while the integration seam exposes `DeviceProvider` as an external vendor capability with request evidence and a possible shared backend. An external provider can therefore declare that its use requires permission and still run on every request. Operator selection is not a per-request consent decision. Please enforce the declaration before invoking a device provider, or reject nonempty device requirements until that gate exists. If the built-in classifier must run without advertising permissions, split that security role into a separate, narrowly scoped authority with restricted evidence and no general vendor egress.

docs/superpowers/specs/2026-07-30-integration-response-header-hook-design.md
```diff
@@ -0,0 +1,1219 @@
...
+  `pragma`, `via`, `sec-ch-ua`, and `sec-ch-ua-full-version-list`. Core removes
+  leading and trailing optional whitespace from each field value, rejects a
+  value containing invalid field-value octets, and combines all field lines
+  (including empty values) in received order with the two literal bytes `, `.
+  This one normalized value is then parsed where the mapping above requires
+  parsing and is then bounded. Commas inside an individual value are not split
+  and reserialized.
+- Every other admitted value-bearing source header in the exact mapping above
+  is singleton. Zero lines means omit the DataDome field. Exactly one valid
+  line is OWS-normalized and processed. Two or more lines, even identical,
```
ChristianPavilonis: 🔧 **P1: Duplicate request headers must not disable DataDome**

Two instances of any admitted singleton header, including `user-agent`, `origin`, or `authorization`, currently skip the entire DataDome call through the fail-open path. Header multiplicity is attacker-controlled, and this spec requires adapters to preserve repeated field lines, so a bot can deliberately add a duplicate and receive the publisher response without validation. Please reject ambiguous security-singleton requests at the HTTP boundary, or omit an optional ambiguous field while still calling DataDome. Do not convert multiplicity into a protection bypass. Add duplicate-header cases to every adapter's qualification suite.

docs/superpowers/specs/2026-07-30-client-cycle-ec-resolve-design.md
```diff
@@ -0,0 +1,260 @@
...
+## 4. Requirements on the page script
+
+- **The re-post guard must not depend on reading an HttpOnly cookie.**
+  Implemented as the draft's first option. The resolve response sets a
+  non-HttpOnly companion marker cookie (`ts-ecr=1`) carrying no identity,
+  which is the only signal the page has that a resolve succeeded. The
+  marker shares the Edge Cookie's scope and lifetime and is expired
+  together with it on withdrawal, so a visitor who later re-establishes
+  the permission can resolve again.
+- **The page leg is permission-gated before vendor contact.** The demo
```
ChristianPavilonis, line 204: 🔧 **P2: Define the browser permission-gating mechanism**

This requires the vendor module to check the server's resolved permissions before vendor contact, but the design does not say how those permissions reach TSJS or combine with a fresh CMP check. Integration modules are self-executing IIFEs selected by provider configuration, while the endpoint gate runs only after the module may already have derived identity or contacted its vendor. A CMP signal alone also cannot reproduce the server's jurisdiction policy. Please define the browser-side authorization mechanism before the first vendor module lands, require module initialization to be side-effect free until that check passes, and add a browser test proving the vendor endpoint receives no request when permission is absent or withdrawn, including after BFCache restoration.

docs/superpowers/specs/2026-07-30-client-cycle-ec-resolve-design.md
```diff
@@ -0,0 +1,260 @@
...
+> owner hash has the `Set-Cookie` re-emitted (lost-response recovery),
+> anyone else gets a terminal response with no cookie. The same-identity
+> no-op first checks the family revocation record, and "revocation wins" is
+> enforced by a CAS conditioned on the family epoch read at the start.
+
+## 4. Requirements on the page script
+
+- **The re-post guard must not depend on reading an HttpOnly cookie.**
+  Implemented as the draft's first option. The resolve response sets a
+  non-HttpOnly companion marker cookie (`ts-ecr=1`) carrying no identity,
```
ChristianPavilonis, line 187: 🔧 **P2: Bind the resolve marker to the selected provider**

`ts-ecr=1` records only that some client-cycle provider resolved in the past. On a switch from provider A to provider B, A's coded EC becomes unrecognized, but this marker remains until withdrawal or its one-year expiry. B's page script therefore sees the marker and never calls resolve, even though the migration contract says the switch restarts identity. Please bind the marker to the selected provider code and behavior revision, or expire it whenever the incoming EC is not recognized by the selected provider. Add an A-to-B client-provider switch test.

docs/superpowers/specs/2026-08-27-integration-provider-seam-design.md
```diff
@@ -0,0 +1,422 @@
...
+the test covers both tables.
+
+### 3.4 Auction providers
+
+Give `AuctionOrchestrator` the same treatment as the registry: a public
+provider-builder type and a second input, so an auction-side vendor such as
+APS can register from its own crate. Prebid stays in core as protocol
+support rather than as a vendor integration.
+
+The bid renderer contract is generalized in the same change. Today
```
ChristianPavilonis, line 169: 🔧 **P2: Add a browser renderer registration seam before moving APS**

Generalizing Rust's `BidRenderer` is not enough to move APS out whole. TSJS core still imports `parseApsRendererDescriptor` from APS in `core/auction.ts`, imports APS dispatch in `core/request.ts`, and fixes `AuctionBidRenderer` to `ApsRendererV1` in `core/types.ts`. Because `build-all.mjs` builds `core/index.ts` as a self-contained IIFE, those static imports remain in `tsjs-core.js` independently of carried integration modules. Please specify a browser renderer registry keyed by descriptor type. Integration modules should register their parser and dispatcher, core should reject unknown types safely, and a cross-language test should prove APS can be removed without changing or breaking the core bundle.

ChristianPavilonis (changes requested):
## Summary

Requesting changes for three security or authorization failures and three contract gaps in the normative design. These findings are distinct from the existing feedback on sequencing, implementation status, Fastly composition, policy placement, and jurisdiction defaults.

Reviewed `180060402f91b4efa96f5823da8934d19c1c78f8` against `d516a9e94249e10cbc36e41beb4269f9255cf407`.

jwrosewell:
Thanks. Great review and I've checked the points. All six hold up. Five are changing in this stack, and one is real but is new design work that these PRs are not the place for. In order.

**P1, subdomain identity fixation. Agreed, and it is worse than you say.** The check in `ec/resolve.rs` is a suffix match on `publisher.domain`, and it also accepts an `http://` origin and discards the port before comparing, so all three of your rejection cases pass today. The Edge Cookie really is `Domain=.publisher.domain`, from `ec_cookie_domain()` in `settings.rs`, so an identity authorized from a sibling subdomain does land on the apex. The justification I put in the spec, that the sibling concern belongs to `Sec-Fetch-Site` and not to a suffix match on the apex, does not hold, because the two admit the same set of origins. I've withdrawn that sentence.

The fix is the same-origin test of RFC 6454, so the request `Origin` and each allowed origin are compared as the scheme, host and port triple of §4 under the §5 rule, with a missing port meaning the scheme's default, and a value that is not a serialized origin under §6.1 never matches. Your three rejection cases are tests. One refinement on the default. A strict allowlist would mean the endpoint does nothing until an operator configures it, so the default is the single exact origin `https://{publisher.domain}` and nothing else, with an optional configured list adding further exact origins for a publisher whose pages are served from `www` or another domain. That also closes the open question I had left at §7.5, so the two are one piece of work.

One more point on severity. The origin check is defense in depth here, not the primary control. The primary control is the provider's envelope verification with audience and session binding, which §3.9 parks out of v1. An exact allowlist does not help against a subdomain that legitimately serves publisher pages and is later compromised, so §3.9 still has to land with the vendor scheme.

**P1, device provider permissions. Agreed.** `DeviceProvider::required_permissions` has no production reader anywhere, and the only place `detect` is called runs before permissions are assembled at all. Before the module seam that was a documented no-op, because core and the host supplied the only two providers. With the seam a vendor crate can implement the trait, so the method now reads as a promise that is not kept.

I've taken your second option. A nonempty `required_permissions` from a module-supplied device provider is rejected at registry resolution and fails loudly at startup, rather than being silently ignored. That closes the false assurance now. Enforcing the declaration properly, which is where this has to end up, needs a per-request device gate that does not exist yet, so I'd rather refuse the declaration than pretend to honor it.

**P1, duplicate headers and DataDome. Agreed on the risk, and I've changed the approach rather than patched the rule.** One correction on the framing first. Nothing skips the call today because a header arrived twice. `protection.rs` reads each mapped field through a single `headers().get()` in `header_value` and goes on to call the Protection API however many lines arrived, and the only skips it has are its own rules, being the test bypass, the OPTIONS preflight and the configured exclusions. The clause in my spec would have created the duplicate-header bypass rather than described one, so this is caught before it ships rather than found in running code. Good catch.

I haven't taken either remedy exactly, because both still have core deciding what to do with a vendor's evidence. Core should not invent header normalization for a vendor, and it certainly should not decide to skip a vendor call. Core passes the request headers through as received, repeated field lines preserved, and the vendor integration decides how to interpret multiplicity, because that is the vendor's detection logic and not core's business. The one field I am leaving out of that is `content-length`, which is HTTP framing handled by the host HTTP layer beneath Trusted Server and which no integration reads as a value, so there is no vendor decision in it.

The wider point, and it is the same separation argument running through this whole review. The DataDome logic sits in core today. Vendor detection logic of that kind belongs in the vendor's own module, behind the seam, not in the shared engine. Getting it there is the direction, and this rule is a small example of why.

**P2, the `ts-ecr` marker. Agreed.** Read-back, the identity-graph key and withdrawal are all bound to the provider code envelope. The marker is the one identity-adjacent artifact left unnamespaced, a bare `ts-ecr=1` that only withdrawal expires, so after a switch a visitor sits with no identity rather than a restarted one. The provider-switching section of the pluggable-providers spec lists read-back, the browser cookie and the graph rows and never mentions the marker, so that is a hole in my text as much as in the code.

I've taken your second remedy rather than your first. Core expires the marker on any request carrying a `ts-ec` that the selected provider does not own. Binding the value to the provider code would work only if every vendor's page script compares the value correctly, and the demo script we would be handing vendors as the pattern does not compare it at all, so making correctness depend on each vendor's copy of that check is the weaker half. Expiring it in core does not depend on anyone's page script.

**P2, browser permission gating. Agreed, the mechanism was undefined. It is defined and delivered now, in the permissions PR (#1045).** There are two deliveries. On the server, integration request filters receive the permission state resolved for the request, next to the geo result they already get, on the Fastly adapter, which is the only adapter that runs the filter step today. On the page, that same state arrives as `window.tsjs.permissions`, an object `{"set": [...]}` naming the Data Uses set for the request. Under inline assembly it goes in as a script at the open of `<head>`, before the bundle. Under a shared template the head is shared across visitors and cannot carry anything request-scoped, so the value is spliced into the per-request `</body>` seam script instead, and a permissions-only seam script is emitted even when the ad stack did not run, so a bot-classified visitor, or one for whom the auction was gated off, still gets the state resolved for the request rather than nothing.

Because the arrival point moves, core exposes `tsjs.whenPermissions()`, a promise that resolves when the value arrives, or with the empty state at DOMContentLoaded if nothing arrived, which a module reads as nothing set. That gives you side-effect free until the check passes. A page module waits on it and does nothing with identity and contacts no vendor before it resolves. You are also right that a CMP signal cannot reproduce the server's jurisdiction policy, which is precisely why permissions rather than consent are the primitive here. The server's resolved decision is the authority and an in-page CMP read is a withdrawal re-check layered under it, never a substitute.

A page module is treated the same way as any other provider. It declares the permissions it requires and checks them against the state it is handed before it does anything. The client-fixed demo script does exactly that now, declaring `necessary.operations.storage` (the same Data Use its server-side provider requires, held in step by a test) and posting only when the promise resolves with it in the set. The page itself renders with that state passed to everything running on it, so what ran and why is visible to anyone looking, which is the point of the trust in the name. Your browser test, meaning no vendor request when a required permission is unset or withdrawn, including after BFCache restoration, is a test of a module's declaration, so it gets written with the first module that declares one.

**P2, the browser renderer seam. You are right, and it is out of scope for this stack.** `core/auction.ts` imports `parseApsRendererDescriptor` from the APS integration, `core/request.ts` imports the APS dispatch, and `core/types.ts` fixes `AuctionBidRenderer` to `ApsRendererV1`. Two things to add, and both help your case. There is no `integrations/aps/index.ts` at all, so APS is not a discovered browser module in the first place, and that code is pre-existing on `main` and byte identical on this branch, so no PR in this stack introduces or changes it.

I've corrected the §4 APS migration row so it names the browser-side work as well as the Rust side, and added the coupling as a §8 finding. Designing the browser renderer registry belongs with the APS migration, not here, and I have filed it as #1111 so it is not lost in this thread.

**On scope.** This stack is not adding new designs or features. It implements the seam that was agreed, and where it overclaims I've corrected the text rather than widened the specs to cover work these PRs will not deliver. The browser renderer registry is genuinely new design work. It is a good agenda item for the next meeting and I'd rather settle it there than in a review thread.

Five changes are in the stack now, being the exact origin check with its rejection tests, the device permission rejection, the marker expiry, the DataDome pass-through rule, and the page and filter carriage of the resolved permissions. The other one is a spec correction that says plainly what is not yet designed.

aram356 (changes requested):
## Permissions policy should use the existing typed app-config path

Requesting changes on the `permissions.yaml` decision.

The argument for a dedicated YAML file does not establish a technical guarantee. YAML is editable before a build, and `include_str!` only records a snapshot in that binary. It does not prove who selected the policy, that the repository file matches the deployed binary, or that a later build or deployment preserves it. If an immutable distributor floor is required, that needs an explicit trust model—such as signed build provenance, an exposed policy digest, and a merge rule under which runtime configuration can only tighten the floor. The file format itself does not provide that.

The current design also does not yet deliver the two-layer model described in the discussion: `permissions.yaml` is compiled in, while runtime overrides are deferred. As written, operational policy changes require rebuilding and redeploying, and the most compliance-sensitive configuration bypasses `ts config validate`, `ts config diff`, and `ts config push`.

We already have an appropriate configuration path:

- Add `permissions: PermissionsConfig` to `Settings`, beside `consent`.
- Serialize and deserialize it through `TrustedServerAppConfig`.
- Validate group references, permission identifiers, acquisition values, normalized location keys, and provider enforcement coverage in `validate_settings_for_deploy`.
- Publish it in the existing `BlobEnvelope` with `ts config push`.
- Construct `PermissionMaps` from `settings.permissions` once at startup.

The TOML can represent the same concepts: a location-independent `default_group`, named groups or baselines, country/region-to-group rules, and consent-signal mappings. A `default_group` also avoids pretending that a visitor with unknown geography belongs to a made-up `default_country`.

Please revise the spec so the effective operational policy lives under `[permissions]` in `trusted-server.toml` and participates in the existing typed configuration lifecycle. If a compiled distributor floor is still required, specify it as a separate optional layer and define exactly:

1. Who is trusted to choose it.
2. How a deployed binary proves which floor it contains.
3. Whether runtime configuration may only tighten it—for example, `granted < requires_signal < denied`.
4. How the effective merged policy and its digest are exposed for audit.

Without those guarantees, a separate compiled YAML file adds a second configuration system and a rebuild requirement without delivering the immutability or auditability used to justify it.

aram356 (commented):
We can design the provider interface to support vendor-owned crates without turning every existing integration into one prematurely. A single crate now does not close the door—it preserves the option until there is a real owner and reason to extract.

aram356 (changes requested):
## Vendor-owned crates should be an extension path, not the default migration

This is blocking because §4 makes separate vendor crates the required structure for integrations that are currently maintained together in this repository. I agree with the goal of allowing vendors to own their integrations independently. I do not agree that we should create those ownership boundaries before the owners and independent lifecycles actually exist.

A crate boundary should represent a real maintenance boundary: an independent owner, release process, dependency or licensing requirements, security-response responsibility, and compatibility commitments. The current integrations are reviewed under the same project governance, released with Trusted Server, and use largely the same dependencies. Moving each one into its own crate does not make it vendor-maintained. Unless a vendor has explicitly accepted those responsibilities, the Trusted Server team still owns the integration—only now it also owns another manifest, release surface, CI path, dependency-update stream, and compatibility obligation.

Separate Rust crates provide compile-time package boundaries. They do not by themselves provide runtime isolation, independent deployment, or the ability to add an integration without rebuilding Trusted Server. If all of these crates remain workspace dependencies compiled into the same binary, most deployment characteristics remain unchanged while repository and release complexity increase.

The current integrations should instead live as vendor modules in one crate:

```text
crates/trusted-server-integrations/
  src/
    vendor_a/
    vendor_b/
    vendor_c/
```

That crate can still provide the intended provider interface, generated registration, and shared conformance tests. The interface should also allow an independently maintained external crate to implement the same contract. A vendor contribution can therefore start as a module in the shared crate or as a genuinely external implementation, depending on who owns its lifecycle.

An integration should be extracted into its own crate when at least one concrete boundary exists, for example:

1. The vendor commits to owning releases, compatibility testing, and security fixes.
2. The integration has materially different dependency or licensing requirements.
3. It needs an independent release cadence.
4. A real external integration demonstrates that the provider seam requires the separation.

Whether Trusted Server supports independently maintained vendor integrations is a product and governance decision. Whether every existing integration needs a separate crate is an engineering packaging decision, and the current ownership and release model does not justify it.

Please revise the spec to use one `trusted-server-integrations` crate for the integrations maintained here, with vendor-specific crates supported as an optional extraction path when real owners and lifecycle boundaries emerge. This preserves the vendor-owned extension model without creating a separate package for every hypothetical future owner.

aram356 (dismissed):
## Blocking: define and sequence the EdgeZero boundary before adding another platform layer

This is blocking. §3.6 states the right high-level rule—host-supplied capabilities are platform services—but the design does not carry that rule through to repository ownership or implementation sequence. Trusted Server already runs on EdgeZero. If we add the missing host-service and composition machinery here first, we create a second platform abstraction that every adapter and future integration must understand.

The overlap is concrete on current `main`:

- `edgezero-core` owns `Hooks`, `RequestContext`, app configuration, `BlobEnvelope`, config/secret/KV store registries, proxying, routing, and middleware.
- `Hooks::stores()` is already the portable declaration that EdgeZero adapter runners use to provision and inject stores.
- `RequestContext` already exposes named and default config, secret, and KV handles.
- Each EdgeZero adapter already owns extraction of its native request context.
- Trusted Server's `RuntimeServices` separately carries config store, secret store, KV store, HTTP client, backend, geo lookup, and client information.

That leaves the boundary ambiguous: some platform services flow through EdgeZero, while others are rebuilt and selected in Trusted Server. The provider seam should not make that duplication permanent.

### What belongs in EdgeZero

1. **Portable host-service access.** Config, secret, and KV handles must continue to flow through EdgeZero's store registries and `RequestContext`. Generic outbound HTTP/proxy access and platform transport/backend capabilities should also have one EdgeZero-owned contract. Trusted Server may use thin type adapters where its error or data types differ, but it should not own a second service registry or a second adapter-selection mechanism for the same host resources.

2. **Normalized host evidence.** Client IP, TLS protocol/cipher, JA4/HTTP2 evidence, and host-provided geographic data are facts extracted by the deployment adapter. The portable evidence types and capability-presence metadata belong in `edgezero-core`; the Fastly, Cloudflare, Spin, and Axum implementations belong in their EdgeZero adapters. Trusted Server should consume normalized evidence rather than reaching into platform SDKs or inventing a separate cross-adapter host context.

3. **Generic adapter lifecycle/composition hooks.** If loading an application-supplied module requires a new startup, request-extension, or composition hook, the generic hook belongs in `edgezero_core::app::Hooks` and must be honored by all four EdgeZero adapter runners. Fastly cannot have a separate composition rule. The payload carried through that hook can be Trusted Server-specific, but the adapter lifecycle mechanism cannot be.

4. **Host capability discovery.** EdgeZero should expose which optional host capabilities are available so an application can fail startup when a selected provider requires unavailable evidence or transport. The provider requirement remains a Trusted Server concept; the authoritative statement of what the current adapter supplies belongs to EdgeZero.

### What remains in Trusted Server

- `IntegrationRegistration`, vendor JavaScript, prepare/finalize hooks, auction providers, configuration selection, and conformance tests.
- `EdgeCookieProvider`, `GeoProvider`, and `DeviceProvider` behavior, including vendor calls, selection, per-request sharing, and permission enforcement.
- HMAC identity, the User-Agent classifier, and other Tech Lab-owned implementations.
- Permission and jurisdiction policy.
- Auction telemetry and other ad-tech-specific request/response behavior.

The important distinction is that a **host geo lookup** belongs in EdgeZero, while a **vendor geo provider that interprets or augments geo data** belongs behind the Trusted Server integration seam. Likewise, raw TLS/JA4/H2 evidence belongs in EdgeZero, while device classification based on that evidence remains Trusted Server behavior. Identity remains a Trusted Server provider, while the KV and secret-store plumbing it consumes remains EdgeZero infrastructure.

### Required changes to this spec

Please make the dependency direction and sequencing normative:

1. Define the missing portable host-evidence, transport, capability-discovery, and lifecycle/composition contracts in EdgeZero.
2. Land and version those EdgeZero changes first.
3. Upgrade Trusted Server to that EdgeZero version and consume the contracts through `Hooks` and `RequestContext`.
4. Make `TrustedServerApp` declare its stores through `Hooks::stores()` instead of leaving the existing EdgeZero injection seam unused.
5. Keep `RuntimeServices`, if it remains, as a Trusted Server request/provider context over EdgeZero handles and normalized evidence—not as an independently owned platform layer.
6. Require the external-registration round trip on Fastly, Axum, Cloudflare, and Spin through the same lifecycle contract. Any Trusted Server-specific library entry point can remain here, but it must not compensate for a missing generic EdgeZero adapter hook.

Until those prerequisites and ownership lines are recorded, §3.6 permits two competing platform seams and leaves each implementation PR to decide ad hoc what belongs upstream. That is exactly the architectural drift this design is supposed to prevent.

jwrosewell:
## 1. The permissions policy location ([review](https://github.com/IABTechLab/trusted-server/pull/1084#pullrequestreview-5081423596))

We considered moving the policy into
`[permissions]` in `trusted-server.toml` and we are not going to, for a
reason that sits before the engineering trade-offs, which is who this
file is for.

The permission policy is the most sensitive document in a deployment. It
states, per jurisdiction, what is allowed and what is not. That is
decided first by the builder of the deployment and then, in
configuration, by its operator. Whether their decision follows a law or
departs from one is theirs to make and not ours as developers, and the
file's job is to let the people who hold that role perform it, and
demonstrate what they decided, easily and simply. The people who author,
review and sign off that policy are compliance and policy people, not
developers.
They can open a single YAML document whose names come from the IAB
Privacy Taxonomy, read it top to bottom, and be walked through it line
by line in a review. Asking them to maintain it as a section of the
operator's TOML, between backend hosts and cache settings, hands their
document to engineers. Keeping the policy understandable and editable by
non-developers, even in raw form, is a key feature of Trusted Server
over other solutions, and for content this sensitive that feature is
worth more than uniform plumbing.

There is a larger reason the policy must stand apart from the operator
config and from any Rust concept. A policy file whose vocabulary is the
IAB Tech Lab Privacy Taxonomy and whose format is plain YAML is a
portable document, not a setting. It is entirely possible, and in my
opinion likely, that files like this end up authored and hosted by a
central authority, a trade body or in time a regulator, and applied
across many implementations of Trusted Server, including implementations
in other languages and other environments doing a similar job. A
publisher then takes the file for their jurisdiction from their trade
body rather than writing one, and there may even be competing files
serving different requirements within a jurisdiction. A document with
that future cannot be a section of one program's TOML. Folding it into
the operator config would tie an industry document to a single
implementation's configuration surface, and the value of a standard,
readable, separate file carrying the Tech Lab's own taxonomy should not
be understated.

On provenance, the compiled file carries exactly the provenance the rest
of the binary carries. Every policy change is a commit with an author, a
pull-request review and a permanent history, and a release is built from
a known commit, so the questions of who selected the policy and what a
binary contains have the same answers for this file as for every line of
code in the project. A `ts config push` has none of that. A runtime push
appears in no repository history and is reviewed by nobody before it
takes effect. Moving the compiled base from the reviewed path into the
operator's TOML would remove the audit trail it has today, not add one.
For the base, a deliberate, reviewed, versioned change is the feature.

To keep the whole position in one place, the builder of a deployment
chooses the permissions file compiled into their image, and the operator
of that image can overlay their own permissions file at configuration
time, the follow-on our 31 August reply described and the
permission-model spec records, so a jurisdiction rule can still change
on a regulator's timeline without a rebuild. The overlay will need its
own trust model when it lands, which is the one your review lists. What
we are declining is narrower than it may have read: making the
operator's TOML the policy document's home.

The layering you describe, a distributor floor with tighten-only merges
and an exposed digest, is a real design space and stays open as a
separate optional layer if a deployment needs it. Nothing in v1
forecloses it, and we agree it would need the trust model you list
before it could claim immutability.

Your point about `default_country` is taken, in a different form. You
are right that treating an unknown visitor as being in a chosen country
is a fiction. A `default_group` setting does not remove the fiction,
because the country answers two questions for an unknown visitor, the
permission baseline and which consent handling applies, and a group name
answers only the first, so a default country would survive alongside it.
The fix belongs in the policy hierarchy itself, and working through
your comment improved the file beyond the default problem. The rules
already run country then region, so the rules became one uniform shape.
At every level a `group` line is the fallback for that level, and
everything beside it is a place. The top level's `group` is, explicitly,
the policy for a visitor whose location cannot be known, and the regions
sit under their country so the reader sees the precedence instead of
inferring it from slash-joined keys:

```yaml
rules:
  group: gdpr-eu       # when no country can be resolved
  jurisdiction: gdpr   # inherited by every country not overriding it,
                       # and the consent handling when no country resolves
  GB: gdpr-uk          # United Kingdom, inherits gdpr
  US:                  # United States
    group: us-opt-out  # any state not listed below
    jurisdiction: non-regulated
    CA:
      group: us-opt-out
      jurisdiction: us-state   # the state is the node itself
```

That is the structure of the shipped file. The US states share a baseline
group and differ in which consent machinery applies, which the
`jurisdiction` override expresses without repeating the group.

A state falls back to its country and a country falls back to the top
level, the same rule at every level, and the file now shows that shape
to the person reading it, which matters because that person is a policy
owner and not a developer.
The same attribute goes further than the unknown case. `jurisdiction` is
inherited down the tree the same way `group` falls back up it, so the
policy document also states which consent regime applies to each place,
and the `consent.gdpr.applies_in` and `consent.us_states.privacy_states`
lists have retired from the operator config into it. One document now carries
the whole of the policy, baselines and regime applicability together,
which is exactly what a trade body or regulator hosting such a file
would need it to carry. The reserved words can never collide with the
place codes, because ISO country and region codes are at most three
characters.

A file whose top level lacks `group` or `jurisdiction` is rejected at
startup, exactly as a missing `default_country` was before it retired,
so there is still no default that nobody chose. The policy document is
self-contained, and the person responsible for the policy can see the
whole of it, the unknown-location case included, in one document they
can read and edit without asking a developer what it means. That change
is on #1045 now, and the spec here is corrected to match.

Since writing this, #1045 has gone further to make the position
concrete: the repository's sample permissions files live in
config/permissions with display names, the file header states that no
policy ever ships with Trusted Server, and a permissions inspector at
tools/permissions-inspector lets anyone load a sample and see which
Data Uses are set for a visitor, computed by trusted-server-core itself
compiled to WebAssembly. The details are in
https://github.com/IABTechLab/trusted-server/pull/1045#issuecomment-5515594008.

jwrosewell:
## 2. Vendor crates ([review](https://github.com/IABTechLab/trusted-server/pull/1084#pullrequestreview-5081481967), and [your comment](https://github.com/IABTechLab/trusted-server/pull/1084#pullrequestreview-5081480212))

We are keeping one crate per vendor as the target structure in §4, and
here is the case for it, taking your points in turn.

The affirmative reason is cadence and ownership, not packaging taste.
Core is IAB Tech Lab's, and the aim is for it to be small and to change
slowly. How slowly is a decision for the group, and the seam is what
makes a cadence as slow as a couple of releases a year possible at all. The vendor integrations are vendors regardless of who typed the
code. Fastly's device evidence is Fastly's, DataDome's detection logic
is DataDome's, and the CMS platform providers we expect next will belong
to their platforms. Those parties ship on their own clocks, and the
structure should let each of them move without touching core or each
other.

On runtime isolation, you are right that separate crates give no runtime
isolation and no ability to add an integration without rebuilding, and
nobody has claimed they do. The boundary is about ownership, review and
change cadence, which a compiler-enforced package boundary serves and a
module directory does not.

On overhead, §4 keeps the vendor crates inside this workspace and this
repository, so there is one CI, one release and one dependency-update
stream exactly as today. A crate adds a manifest and a directory, not a
release surface. A separate release surface appears only when a vendor
takes their crate out of the workspace, and at that point it is the
goal, not a cost.

On "a single crate preserves the option", the project's own history says
otherwise. A shared crate is how melding happens: shared private
helpers, cross-vendor imports, one review queue, and after a while no
boundary anyone can enforce. That is the condition the seam exists to
cure, and putting nine vendors into one `trusted-server-integrations`
crate recreates it one level down. The option that matters is the one a
hard boundary preserves, which is that any vendor's code can be handed
to that vendor whole.

On ownership, we agree an unowned crate is not vendor-maintained by
virtue of its manifest, and the fix is the one Prebid has run for
years, where every adapter names its maintainer, visibly, as a
requirement. §4's migration PRs give each vendor crate a maintainers
declaration the same way, the spec now requires it, and unlike Prebid's
single module tree, the crate boundary adds compiler-enforced
separation, per-crate code ownership and independent versioning on top
of the named maintainer. Where no vendor steps up for
an existing integration, deciding whether Tech Lab keeps it or retires
it is a governance question, and it is one we have already put to Tech
Lab management rather than one this spec can settle.

This shape is the established pattern, not an invention of this spec,
and the pages below show it directly, vendor after vendor in one
repository. The Rust ecosystem builds this way as a matter of course,
with [tokio](https://github.com/tokio-rs/tokio) and
[wasmtime](https://github.com/bytecodealliance/wasmtime/tree/main/crates)
each being one repository and one workspace of many crates (wasmtime
carries 38), built and released through one CI. OpenTelemetry's
collector-contrib holds more than 240 components in one repository, its
[exporter directory](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter)
reads as a vendor list, and its
[CODEOWNERS](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/.github/CODEOWNERS)
names owners per component with the vendors' own people among them. Home
Assistant runs per-integration code owners across the more than a
thousand vendor integrations in its
[components directory](https://github.com/home-assistant/core/tree/dev/homeassistant/components).
Prebid.js requires every adapter in its
[modules directory](https://github.com/prebid/Prebid.js/tree/master/modules)
to name its maintainer. Hard package boundaries plus named owners inside
one repository is the shape these projects converged on at scale, and it
is exactly what §4 describes.

§4 stays as written, one migration PR per integration, and the seam is
proven by an implementation outside core either way.

jwrosewell:
## 3. The EdgeZero boundary ([review](https://github.com/IABTechLab/trusted-server/pull/1084#pullrequestreview-5081543112))

Two corrections to the premise, one agreement in direction, and a
disagreement on sequencing.

First, these PRs add no platform machinery. `RuntimeServices` and every
platform service handle it carries exist on `main` today and predate
this stack. What the stack adds to it are the resolved Edge Cookie
provider, the device provider, and a host signals handle whose capture
happens in the Fastly adapter and the opt-in Fastly device crate, both
host-specific, with core gaining only the neutral evidence types. The
stack threads what exists and composition stays in the adapters where it
already lives. §3.6 states the rule that host-supplied capabilities are
platform services, and nothing in these seven PRs moves against it. So
the two-layer situation you describe is real, and it is not something
this stack created or worsens, which makes it a strange thing to block
this stack on.

Second, geo is not a host service in this design. It is a provider a
deployer selects, like identity and device. A host geo lookup is one
data source that one implementation of that provider may consume, a
publisher who wants a different geo service plugs one in through the
same seam, and we do not assume the host has geo at all. §3.6's list
had loosely named "the host geo lookup" among the platform services and
has been tightened to say this precisely. Geographic data therefore does
not belong on a list of normalized host evidence, and folding it into a
platform contract would take a deployer's choice away and hand it to the
platform.

The agreement is that the transport-level evidence, client IP, TLS,
JA4 and HTTP/2, can reasonably live behind an EdgeZero contract in time. Those
types live in Trusted Server today because EdgeZero has no evidence
contract to hand them to, they are thin, their capture is
host-specific, and nothing in this stack hardens against moving them. When EdgeZero has the contract,
migrating them is mechanical.

The disagreement is the sequencing. Defining, landing and versioning
those contracts is work in EdgeZero, to be proposed there and brought to
the Task Force like any other dependency change. This design does not
gate on it, because nothing here needs it. The seam composes through
what exists on `main` today, and the later migration you describe
touches the adapters, not the seam contract. Holding seven reviewed PRs
against future releases of a separate repository inverts the dependency,
since the seam works without the new contracts and the new contracts
would change nothing a vendor sees.

There is also a decision here that is not a review thread's to make.
Taking a hard dependency on new contracts in a repository outside IAB
Tech Lab's governance is a dependency decision for the Task Force, and
it should be put there as a proposal rather than imposed as a
precondition on reviewed PRs.

So the concrete outcome is this. §3.6 keeps its rule and now records
that the transport evidence types are candidates to migrate behind an
EdgeZero evidence contract when one exists, landed on this PR. The EdgeZero
contract work is yours to propose in that repository on its own
timeline, and the dependency question goes to the Task Force. The seam
proceeds on `main` as it stands.

jwrosewell:
For reviewers of this spec: the permission model it governs can now be seen working rather than read about. The inspector at https://jwrosewell.github.io/trusted-server/tools/permissions-inspector/ runs trusted-server-core compiled to WebAssembly against the repository's sample policy and shows which Data Uses are set for any visitor location and consent signals, with the reason per Data Use, the taxonomy definition behind each one, and the engine's version, branch and commit stated on the page. It ships in #1045 at `tools/permissions-inspector`, and the hosting above is temporary from my fork until the repository hosts its own.

docs/superpowers/specs/2026-07-30-client-cycle-ec-resolve-design.md
jwrosewell: **Agreed, and it is worse than the review says. Changed in this stack.**

The check in `ec/resolve.rs` was a suffix match on `publisher.domain`, and it also accepted an `http://` origin and discarded the port before comparing, so all three of your rejection cases passed. The Edge Cookie really is `Domain=.publisher.domain` (from `ec_cookie_domain()` in `settings.rs`), so an identity authorized from a sibling subdomain does land on the apex. The justification I had written into the spec, that the sibling concern belongs to `Sec-Fetch-Site` rather than to a suffix match, does not hold because the two admit the same set of origins. That sentence is withdrawn.

**What it does now.** The comparison is the same-origin test of RFC 6454 §5, so the request `Origin` and each allowed origin match only when their scheme, host and port triples (§4) are equal, a missing port means the scheme's default, and a value that is not a serialized origin under §6.1 never matches. See `crates/trusted-server-core/src/ec/resolve.rs:247-285` (`origin_is_publisher` and `origins_match`).

**On the default.** A strict allowlist would mean the endpoint does nothing until an operator configures it, so the default is the single exact origin `https://{publisher.domain}` and nothing else, with an optional configured list adding further exact origins for a publisher serving pages from `www` or another domain. That also closes the open question left at §7.5, so the two are one piece of work.

**Your three rejection cases are tests**, in `ec/resolve.rs`:
- `resolve_rejects_a_publisher_subdomain_origin` (line 890)
- `resolve_rejects_a_plain_http_origin` (line 915)
- `resolve_rejects_a_non_default_port_origin` (line 940)
- plus `resolve_rejects_a_missing_or_foreign_origin` (863), `resolve_accepts_a_configured_extra_origin` (965) and `origins_match_follows_rfc_6454` (994)

**One point on severity.** The origin check is defense in depth, not the primary control. The primary control is the provider's envelope verification with audience and session binding, which §3.9 parks out of v1. An exact allowlist does not help against a subdomain that legitimately serves publisher pages and is later compromised, so §3.9 still has to land with the vendor scheme.

Full response to all six points is in [this comment](https://github.com/IABTechLab/trusted-server/pull/1084#issuecomment-5497421669). Apologies that it went there rather than into these threads, which left them looking unanswered.

docs/superpowers/specs/2026-07-30-pluggable-providers-design.md
jwrosewell: **Agreed. Changed in this stack.**

`DeviceProvider::required_permissions` had no production reader anywhere, and the only place `detect` is called runs before permissions are assembled at all. Before the module seam that was a documented no-op, because core and the host supplied the only two providers. With the seam a vendor crate can implement the trait, so the method read as a promise that was not kept.

**I have taken your second option.** A nonempty `required_permissions` from a module-supplied device provider is now rejected at registry resolution and fails loudly at startup, rather than being silently ignored. See `crates/trusted-server-core/src/integrations/registry.rs:965`, which refuses the registration when `provider.required_permissions() != PermissionSet::none()`.

That closes the false assurance now. Enforcing the declaration properly, which is where this has to end up, needs a per-request device gate that does not exist yet, so I would rather refuse the declaration than pretend to honor it. The built-in User-Agent classifier declares no permissions and so is unaffected.

Full response to all six points is in [this comment](https://github.com/IABTechLab/trusted-server/pull/1084#issuecomment-5497421669).

docs/superpowers/specs/2026-07-30-integration-response-header-hook-design.md
jwrosewell: **Agreed on the risk. I have changed the approach rather than patched the rule.**

One correction on the framing first. Nothing skips the call today because a header arrived twice. `protection.rs` reads each mapped field through a single `headers().get()` in `header_value` and goes on to call the Protection API however many lines arrived, and the only skips it has are its own rules, being the test bypass, the OPTIONS preflight and the configured exclusions. The clause in my spec would have **created** the duplicate-header bypass rather than described one, so this is caught before it ships rather than found in running code. Good catch.

**What the spec says now.** I have not taken either remedy exactly, because both still have core deciding what to do with a vendor's evidence. Core should not invent header normalization for a vendor, and it certainly should not decide to skip a vendor call. Core passes the request headers through as received, preserving received field-line order and value octets, and the vendor integration decides how to interpret multiplicity, because that is the vendor's detection logic and not core's business.

The one field left out of that is `Content-Length`, which is HTTP framing handled by the host HTTP layer beneath Trusted Server and which no integration reads as a value, so there is no vendor decision in it.

**The wider point**, and it is the same separation argument running through this review, is that the DataDome logic sits in core today. Vendor detection logic of that kind belongs in the vendor's own module behind the seam, not in the shared engine. Getting it there is the direction, and this rule is a small example of why.

Full response to all six points is in [this comment](https://github.com/IABTechLab/trusted-server/pull/1084#issuecomment-5497421669).

docs/superpowers/specs/2026-07-30-client-cycle-ec-resolve-design.md
jwrosewell: **Agreed, the mechanism was undefined. It is defined and delivered now, in the permission model PR (#1045).**

There are two deliveries. On the server, integration request filters receive the permission state resolved for the request, next to the geo result they already get, on the Fastly adapter, which is the only adapter that runs the filter step today. On the page, that same state arrives as `window.tsjs.permissions`, an object `{"set": [...]}` naming the Data Uses set for the request.

Under inline assembly it goes in as a script at the open of `<head>`, before the bundle. Under a shared template the head is shared across visitors and cannot carry anything request-scoped, so the value is spliced into the per-request `</body>` seam script instead, and a permissions-only seam script is emitted even when the ad stack did not run, so a bot-classified visitor, or one for whom the auction was gated off, still gets the state resolved for the request rather than nothing.

**Side-effect free until the check passes.** Because the arrival point moves, core exposes `tsjs.whenPermissions()`, a promise that resolves when the value arrives, or with the empty state at DOMContentLoaded if nothing arrived, which a module reads as nothing set. See `crates/trusted-server-js/lib/src/core/permissions.ts:71`. A page module waits on it and does nothing with identity and contacts no vendor before it resolves. The client-fixed demo module does exactly that, at `crates/trusted-server-js/lib/src/integrations/ec_client_fixed/index.ts:45-50`, declaring `necessary.operations.storage` (the same Data Use its server-side provider requires, held in step by a test) and posting only when the promise resolves with it in the set.

You are also right that a CMP signal cannot reproduce the server's jurisdiction policy, which is precisely why permissions rather than consent are the primitive here. The server's resolved decision is the authority and an in-page CMP read is a withdrawal re-check layered under it, never a substitute.

**On your browser test.** No vendor request when a required permission is unset or withdrawn, including after BFCache restoration, is a test of a module's declaration, so it gets written with the first module that declares one.

Full response to all six points is in [this comment](https://github.com/IABTechLab/trusted-server/pull/1084#issuecomment-5497421669).

docs/superpowers/specs/2026-07-30-client-cycle-ec-resolve-design.md
jwrosewell: **Agreed. Changed in this stack.**

Read-back, the identity-graph key and withdrawal are all bound to the provider code envelope. The marker was the one identity-adjacent artifact left unnamespaced, a bare `ts-ecr=1` that only withdrawal expired, so after a switch a visitor sat with no identity rather than a restarted one. The provider-switching section of the pluggable-providers spec listed read-back, the browser cookie and the graph rows and never mentioned the marker, so that was a hole in my text as much as in the code.

**I have taken your second remedy rather than your first.** Core expires the marker on any request carrying a `ts-ec` that the selected provider does not own. Binding the value to the provider code would work only if every vendor's page script compares the value correctly, and the demo script we would be handing vendors as the pattern does not compare it at all, so making correctness depend on each vendor's copy of that check is the weaker half. Expiring it in core does not depend on anyone's page script.

Expiry is asserted at `crates/trusted-server-core/src/ec/finalize.rs:1231` and `:1266`, which check the response carries `ts-ecr=` with `Max-Age=0`.

Full response to all six points is in [this comment](https://github.com/IABTechLab/trusted-server/pull/1084#issuecomment-5497421669).

docs/superpowers/specs/2026-08-27-integration-provider-seam-design.md
jwrosewell: **You are right, and it is out of scope for this stack. Leaving this thread open deliberately.**

Confirming your reading. `core/auction.ts` imports `parseApsRendererDescriptor` from the APS integration, `core/request.ts` imports the APS dispatch, and `core/types.ts` fixes `AuctionBidRenderer` to `ApsRendererV1`.

Two things to add, and both help your case. There is no `integrations/aps/index.ts` at all, so APS is not a discovered browser module in the first place. And that code is pre-existing on `main` and byte identical on this branch, so no PR in this stack introduces or changes it.

**What I have done.** Corrected the §4 APS migration row so it names the browser-side work as well as the Rust side, and added the coupling as a §8 finding. Designing the browser renderer registry belongs with the APS migration rather than here, so I have filed it as #1111 to make sure it is not lost in a review thread.

I am not resolving this one, because the deferral is a judgment you should get to disagree with. If you would rather the registry were specified before this spec merges, say so and I will take it back into scope.

Full response to all six points is in [this comment](https://github.com/IABTechLab/trusted-server/pull/1084#issuecomment-5497421669).

PR #1090, opened 2026-08-28
Support analytics adapters in generated Prebid bundles

### LINKED ISSUE #1085: Support analytics adapters in generated Prebid bundles
## User story

As a publisher using `ts prebid bundle`, I want to select Prebid analytics adapters alongside bidder and User ID modules so that replacing the publisher's Prebid bundle does not leave configured `pbjs.enableAnalytics` providers unregistered.

## Current behavior

`[integrations.prebid.bundle]` accepts only `adapters` and `user_id_modules`. The CLI and `build-prebid-external.mjs` have no supported way to import analytics adapter modules. If publisher code enables an adapter that was present in its original bundle, Prebid logs `no analytics adapter found in registry` after Trusted Server substitutes the generated bundle.

One observed case requires the upstream `atsAnalyticsAdapter.js` module, registered as `atsAnalytics`. A second configured provider, `mavenDistributionAnalyticsAdapter`, is not present in the repository's pinned Prebid.js package and needs an actionable unsupported-module error unless custom adapter support is designed separately.

## Proposed behavior

Add an optional typed configuration field:

```toml
[integrations.prebid.bundle]
analytics_adapters = ["atsAnalyticsAdapter"]
```

Then:

- parse and validate the field in `ts prebid bundle`;
- pass the selection to the JavaScript bundle generator;
- resolve modules only from the pinned `prebid.js/modules` directory and reject missing or escaping paths;
- statically import selected analytics modules into the generated entry point;
- include selected module names in the generated bundle manifest;
- retain the existing bundle hash and SRI update flow; and
- document how module filenames map to the adapter names used by `pbjs.enableAnalytics`.

Do not accept arbitrary local paths or remote URLs as part of this issue. Publisher-specific analytics adapters need a separate design with reproducible builds and an explicit trust policy.

## Completion criteria

- A bundle generated with `atsAnalyticsAdapter` registers `atsAnalytics`, and enabling it no longer produces a missing-registry error.
- An unavailable module fails the build with the config field, requested name, and supported-module guidance in the error.
- Omitting `analytics_adapters` preserves current output behavior.
- CLI parsing, generator output, manifest data, invalid module names, and runtime registration have automated tests.
- The Prebid integration and configuration guides describe the field and the upstream-only limitation.

ChristianPavilonis:
## Summary

- Adds analytics adapter selection to generated Prebid bundles so publisher calls to `pbjs.enableAnalytics` can use adapters such as ATS without missing-registry errors.
- Replaces the separate bidder and User ID settings with one typed module map covering bidder, User ID, and analytics modules.
- Validates every selected module against the locked Prebid.js package before generating source or invoking Vite, then records module stems and registered runtime codes separately.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-cli/src/prebid_bundle.rs` | Parses the typed module map, rejects removed fields, sends one JSON request to the generator, and validates the versioned manifest before updating bundle hashes. |
| `crates/trusted-server-js/lib/build-prebid-external.mjs` | Validates locked-package metadata and exports, generates selected module imports, and writes nested versioned manifests. |
| `crates/trusted-server-js/lib/src/integrations/prebid/index.ts` | Reads the nested browser manifest and uses registered bidder codes for runtime diagnostics. |
| `crates/trusted-server-js/lib/src/integrations/prebid/user_id_modules.json` | Keeps User ID registry metadata while removing generator-controlled import paths. |
| `crates/trusted-server-js/lib/src/integrations/prebid/user_id_modules.ts` | Removes the obsolete `importPath` field from the registry type. |
| `crates/trusted-server-js/lib/test/build-prebid-external.test.mjs` | Covers request parsing, package validation, metadata and export containment, generated imports, and manifest output. |
| `crates/trusted-server-js/lib/test/integrations/prebid/index.test.ts` | Covers nested manifest parsing and independent bidder and User ID diagnostics. |
| `crates/trusted-server-js/lib/test/prebid-artifact-integration.test.mjs` | Proves ATS registration, publisher queue enablement, auction behavior, network isolation, omitted analytics, and watchdog queue draining with production artifacts. |
| `docs/guide/cli.md` | Documents the typed module configuration and generated manifest workflow. |
| `docs/guide/integrations/prebid.md` | Documents exact module stems, analytics provider ownership, ATS configuration, and upstream-only restrictions. |
| `docs/superpowers/plans/2026-08-28-prebid-bundle-module-map.md` | Records the reviewed implementation and validation plan. |
| `docs/superpowers/specs/2026-08-28-prebid-bundle-module-map-design.md` | Records the implemented configuration, validation, manifest, and runtime design. |
| `trusted-server.example.toml` | Updates the sample Prebid bundle configuration to use the typed module map. |

## Scope

This pull request changes the Rust command-line interface, Node.js bundle generator, browser diagnostics, tests, and operator documentation because they share one configuration and manifest contract. It is limited to modules shipped by the locked Prebid.js package. It does not add custom module paths, download third-party adapters, or move `pbjs.enableAnalytics` options out of publisher code.

## Closes

Closes #1085

## Test plan

- [x] Other: `cargo test-cloudflare`, `cargo test-spin`, all six target-matched Clippy commands, parity tests, CLI tests, JavaScript build, documentation build, a real bundle smoke test, and the unsupported-module failure path.

## Checklist

- [x] No `unwrap()` in production code; new fallible paths return actionable errors
- [x] No runtime logging was added; existing project logging conventions are unchanged
- [x] New code has tests


docs/guide/integrations/prebid.md
```diff
@@ -154,16 +158,33 @@ the generated manifest. Upload the generated JavaScript file manually, set
...
+effective module lists and the bidder and analytics runtime codes. Regenerating
+a bundle changes its content-addressed filename, SHA-256, and SRI when its
+contents change.
+
+The external artifact contains Prebid core, consent modules, and the selected
+modules. The separate deferred `tsjs-prebid` shim installs the `trustedServer`
+adapter on the same `window.pbjs` object and processes the publisher queue. A
+bundle generated before the shim split still carries a baked-in shim, so upgrade
+that bundle with the server and push its new hash and SRI. The sentinel
+`window.__tsjsPrebidShimInstalled` prevents duplicate shim installation.
```
prk-Jr, line 187: 🤔 **thinking** — Two separate breaks land here with no migration section for operators.

Issue #1085 asked for an additive `analytics_adapters` field and stated that omitting it should preserve current output. This PR instead removes `adapters` and `user_id_modules` and requires full upstream stems, so `adapters = ["rubicon"]` has to become `bidder = ["rubiconBidAdapter"]`. That is a defensible design and the rejection error in `load_bundle_config` names the replacement field path — but it never shows the short-name-to-stem expansion, and these guides only describe the new shape. An operator upgrading has no page that says what to change.

Separately, `getExternalBundleManifest()` now requires `schemaVersion === 1`. The shim ships with the server deploy while the external bundle is a manual rebuild and re-upload, so there is a window where the hosted bundle stamps the old flat shape and reads as unstamped. `client_side_bidders` adapter verification and User ID coverage diagnostics go quiet behind a single warning. That is clearly deliberate — `test/integrations/prebid/index.test.ts` covers the old-schema manifest explicitly — and auctions are unaffected, but nothing tells the operator to rebuild.

Both facts fit in the section that already carries the shim-split migration note:

```suggestion
bundle generated before the shim split still carries a baked-in shim, so upgrade
that bundle with the server and push its new hash and SRI. The sentinel
`window.__tsjsPrebidShimInstalled` prevents duplicate shim installation.

### Upgrading from `bundle.adapters` and `bundle.user_id_modules`

`ts prebid bundle` rejects the removed `adapters`, `user_id_modules`, and
`analytics_adapters` keys. Move each list under
`[integrations.prebid.bundle.modules]` and expand short adapter names to full
upstream stems — `adapters = ["rubicon"]` becomes
`bidder = ["rubiconBidAdapter"]`. `client_side_bidders` keeps the runtime code
(`rubicon`), not the stem.

The manifest is also schema-versioned now. A bundle built before `schemaVersion`
existed reads as unstamped, so the shim logs one warning and skips
`client_side_bidders` adapter verification and User ID coverage diagnostics
until that bundle is rebuilt and re-uploaded. Auction behavior is unaffected.
```

_(Verified: applying this introduces no new `docs` prettier deviation.)_

crates/trusted-server-js/lib/test/prebid-artifact-integration.test.mjs
```diff
@@ -73,155 +78,356 @@ afterAll(() => {
...
+    consoleWarnings,
+    unexpectedTransports,
+    blockedResourceRequests: pageWindow.__blockedResourceRequests,
+  };
+}
+
+function expectNoUnexpectedNetworkActivity(stubs) {
+  expect(stubs.unexpectedTransports).toEqual([]);
+  expect(stubs.blockedResourceRequests).toEqual([]);
+}
```
prk-Jr, line 201: ♻️ **refactor** — The PR description lists "network isolation" as one of the properties this suite proves, but this helper only inspects `unexpectedTransports` (websocket and eventsource, which throw on construction anyway) and jsdom's `blockedResourceRequests`. Every `fetch`, XHR, `sendBeacon`, and `Image` request the artifacts make is recorded into `stubs.requests` and then never looked at, so a bid or analytics adapter that beacons a vendor endpoint would pass this suite silently. That matters more now that the bundle can carry analytics adapters.

I instrumented the suite at this head and dumped `stubs.requests`: it is genuinely clean today — `[{"transport":"fetch","url":"https://pub.example.com/auction"}]` in both auction tests and `[]` in the watchdog test — so the assertion below passes as written. I then injected a synthetic `https://vendor.example.net/beacon` and confirmed it fails all three call sites, so it is not a no-op.

```suggestion
function expectNoUnexpectedNetworkActivity(stubs) {
  expect(stubs.unexpectedTransports).toEqual([]);
  expect(stubs.blockedResourceRequests).toEqual([]);
  // Every artifact-initiated call must stay first-party. Without this, an
  // analytics or bid adapter that beacons a vendor endpoint passes silently.
  expect(
    stubs.requests
      .map((entry) => requestUrl(entry.url ?? entry.resource))
      .filter((url) => !url.startsWith('https://pub.example.com/'))
  ).toEqual([]);
}
```

`requestUrl` is declared below this point but is a hoisted function declaration, so the reference resolves. _(Verified: prettier clean, 5/5 tests pass.)_

crates/trusted-server-js/lib/test/prebid-artifact-integration.test.mjs
```diff
@@ -73,155 +78,356 @@ afterAll(() => {
...
+      expect(pageWindow.__tsjs_prebid_bundle.modules.analytics).toEqual([]);
+      expect(pageWindow.__tsjs_prebid_bundle.runtimeCodes.analytics).toEqual([]);
+      expectNoUnexpectedNetworkActivity(stubs);
+    } finally {
+      vi.useRealTimers();
+      dom.window.close();
+    }
+  });
+
+  it('records hashes and SRI for the no-analytics artifact', () => {
```
prk-Jr, line 417: ⛏ **nitpick** — The name says "the no-analytics artifact", but the body calls `expectManifest` on both artifacts before getting to the hash and SRI checks.

```suggestion
  it('records manifests, hashes, and SRI for both artifacts', () => {
```

_(Verified: prettier clean, 5/5 tests pass.)_

prk-Jr (approved):
## Summary

Typed module map, versioned manifest, and analytics support for `ts prebid bundle`. The resolver hardening is the strongest part of this change and I have no correctness objections to the implementation. Everything below is non-blocking: one documentation gap on the operator upgrade path, one test assertion that is weaker than the PR description claims, two maintainability items, and one follow-up that is out of scope for this PR.

> 3 of the inline comments below carry a one-click GitHub `suggestion` — use **Commit suggestion** (or **Add suggestion to batch**) to apply them. Each was applied to a scratch worktree at this head and verified before posting: `npx prettier --check`, `npx vitest run test/prebid-artifact-integration.test.mjs` (5/5), and `docs` prettier via the pinned `docs/node_modules` binary. The three together produce no drift beyond the approved ranges. The remaining comments are prose because the fix spans multiple hunks or files.

## Non-blocking

### 🤔 thinking

- **Operator upgrade path is not written down anywhere but a CLI error string** — see inline at `docs/guide/integrations/prebid.md:185`

### ♻️ refactor

- **`expectNoUnexpectedNetworkActivity` never asserts on `stubs.requests`** — see inline at `crates/trusted-server-js/lib/test/prebid-artifact-integration.test.mjs:198`
- **Metadata directory is re-read once per selected module** — see Cross-cutting below
- **Pinned Prebid version hardcoded in three assertions** — see Cross-cutting below

### ⛏ nitpick

- **Test name describes one artifact, body asserts both** — see inline at `crates/trusted-server-js/lib/test/prebid-artifact-integration.test.mjs:417`

### 📌 out of scope

- **Analytics adapters compile into a bundle with no TCF/GPP purpose enforcement** — see Cross-cutting below

### 👍 praise

- **Resolver hardening** — see Cross-cutting below

- ♻️ **Metadata directory is re-read once per selected module** — `resolveOneModule` calls `fs.readdirSync(context.metadataDir)` inside the per-module loop (`crates/trusted-server-js/lib/build-prebid-external.mjs:307`), and that directory holds 665 entries in the pinned package. The readdir itself is the right call — it is an exact-case membership test, which `fs.existsSync` would not give you on a case-insensitive filesystem, and the `ATSanAlyticsAdapter` test depends on that behaviour — it just does not need repeating per module. Hoist it into `resolveBundleModules` beside the `canonicalMetadataDir` validation and thread a `Set` through `context`:

  ```js
  // resolveBundleModules, after canonicalMetadataDir is validated:
  const metadataEntries = new Set(fs.readdirSync(canonicalMetadataDir));
  // ...include metadataEntries in the context object passed to resolveOneModule

  // resolveOneModule:
  if (!context.metadataEntries.has(metadataFilename)) {
    throw unsupportedModuleError(definition, stem, context.prebidVersion);
  }
  ```

  Apply manually — this touches two non-adjacent hunks, so it cannot be a single `suggestion`.

- ♻️ **Pinned Prebid version hardcoded in three assertions that are not about the version** — `crates/trusted-server-js/lib/test/build-prebid-external.test.mjs:52`, `:671`, `:739`, and `crates/trusted-server-js/lib/test/prebid-artifact-integration.test.mjs:279` all pin `10.26.0`. A prebid.js bump breaks four assertions whose subject is module resolution, not the version number. `verifyPrebidPackageVersion` is already exported and already imported in the first file — deriving it once at module scope (`const prebidVersion = verifyPrebidPackageVersion();`) and substituting keeps the version-interpolation assertion at `:671` meaningful while removing the coupling. The `writeVersionFixtures` literals are correct as they are; those are the subject of their own tests.

  Apply manually — spans two files.

- 📌 **Analytics adapters now compile into a bundle with no TCF/GPP purpose enforcement** — `renderExternalEntry` imports `consentManagementTcf`, `consentManagementGpp`, and `consentManagementUsp` (`crates/trusted-server-js/lib/build-prebid-external.mjs:526-528`) but not `tcfControl` or `gppControl`. The `consentManagement*` modules read and expose the consent string; the `*Control` modules are what actually gate vendor activity on denied purposes. That gap predates this PR, but this PR is what makes it straightforward to compile an identity-analytics vendor into the bundle. Publisher JavaScript owns `enableAnalytics`, so a publisher can gate it themselves — but the operator has no way to add the enforcement modules through `bundle.modules`, since they are not bidder, userId, or analytics component types. I could not find an open issue tracking this. Worth filing one before analytics ships to a GDPR publisher; no change requested here.

- 👍 **Resolver hardening** — `realpath` containment checks on both the metadata file and the resolved package export, with a symlink-escape test for each; exact-case metadata matching; the lockfile-versus-installed version check with actionable `npm ci` guidance; and the removal of the old silent fallback that stamped the module stem as a bidder code when metadata was missing. Asserting `createGeneratedPaths` and `buildBundle` were `not.toHaveBeenCalled()` on every failure path is a precise way to pin the "fail before Vite" ordering, and it is done consistently across six tests.

## CI Status

- integration tests: PASS
- CodeQL: PASS
- format-docs: FAIL (required)
- Analyze (rust): PASS
- cargo test: PASS (required)
- vitest: PASS
- Analyze (actions): PASS
- cargo fmt: PASS (required)

`format-docs` is two whitespace-only deviations in `docs/guide/integrations/prebid.md`, both introduced by `1e52d790`: a missing blank line after the `### Browser configuration options` heading (line 79), and a doubled blank line before `## External Bundle Generation` (line 144). Reviewer confirmed those are the only two, and that fixing them makes `cd docs && npm run format` pass. They were deliberately excluded from this review's findings at the reviewer's request, but `format-docs` is a required check, so merge stays blocked until they are fixed.

crates/trusted-server-cli/src/prebid_bundle.rs
```diff
@@ -29,19 +29,76 @@ fn cli_error<T>(message: impl Into<String>) -> CliResult<T> {
...
+        D: Deserializer<'de>,
+    {
+        let value = String::deserialize(deserializer)?;
+        Self::new(value).map_err(serde::de::Error::custom)
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct PrebidBundleModules {
```
aram356, line 67: 🔧 **wrench** — The runtime config still models the old `bundle` shape, so the config this PR documents cannot be validated, pushed, or loaded.

`PrebidBundleBuildConfig` in `crates/trusted-server-core/src/integrations/prebid.rs:376-386` still declares only `adapters` and `user_id_modules` under `#[serde(deny_unknown_fields)]`, and every core deserialization of `PrebidIntegrationConfig` (integration registration, `validate_settings_for_deploy`, the config payload) goes through it. I built `ts` from this head and ran `ts config validate` on `trusted-server.example.toml` with the new `[integrations.prebid.bundle.modules]` block uncommented:

```
[ts] typed app-config failed validation: trusted_server: Configuration error: Integration 'prebid' configuration could not be parsed
exit 2
```

Same result with `enabled = false` and `enabled = true`; the old `adapters` shape validates. So an operator following the new guides cannot `ts config validate`, `ts config push`, or start the server with that config. CI stayed green because the core fixtures (`crates/trusted-server-core/src/test_support.rs:27` and `crates/trusted-server-core/src/config.rs:1278`) still use `bundle.adapters`.

**Proposed fix** (apply manually — touches `trusted-server-core`, outside this diff):

```rust
/// CLI build inputs retained in app config but ignored safely by the runtime.
#[derive(Debug, Clone, Default, Deserialize, Serialize)]
#[serde(deny_unknown_fields)]
pub struct PrebidBundleBuildConfig {
    /// Typed Prebid.js module selections consumed by `ts prebid bundle`.
    #[serde(default)]
    pub modules: PrebidBundleModulesConfig,
}

/// Exact Prebid.js module stems selected by `ts prebid bundle`.
///
/// The CLI validates these values; the runtime only needs to parse them so a
/// config carrying build inputs stays loadable.
#[derive(Debug, Clone, Default, Deserialize, Serialize)]
#[serde(deny_unknown_fields)]
pub struct PrebidBundleModulesConfig {
    /// Bidder adapter module stems, for example `rubiconBidAdapter`.
    #[serde(default)]
    pub bidder: Vec<String>,
    /// User ID module stems; `None` selects the generator's curated preset.
    #[serde(default)]
    pub user_id: Option<Vec<String>>,
    /// Analytics adapter module stems; `None` and `[]` both select none.
    #[serde(default)]
    pub analytics: Option<Vec<String>>,
}
```

Then switch the two fixtures to `[integrations.prebid.bundle.modules]` / `bidder = ["exampleBidderBidAdapter"]` (and the JSON equivalent in `config.rs`). Verified in a scratch worktree: `cargo fmt --check`, `cargo clippy-fastly`, `cargo clippy-axum`, 2644 core tests under Viceroy, and 181 CLI tests pass, and the new-shape probe config validates.

One consequence worth stating in the migration note: with this strict mirror, a deployed config that still carries `bundle.adapters` is rejected by the **server** at startup after upgrade, not only by `ts prebid bundle`. That is consistent with the no-compat decision in the spec, but the docs currently frame the break as CLI-only. The alternative is to keep `adapters` / `user_id_modules` as tolerated deprecated fields in core and reject them only in the CLI.

docs/guide/integrations/prebid.md
```diff
@@ -74,7 +77,6 @@ provider = "pbs-main"
 ```
 
 ### Browser configuration options
```
aram356, line 79: 🔧 **wrench** — `format-docs` (a required check) fails on this file: prettier wants a blank line between the heading and the table. This and the companion suggestion at line 144 are both needed for the check to pass.

```suggestion
### Browser configuration options

```

docs/guide/integrations/prebid.md
```diff
@@ -138,10 +141,11 @@ Browser `timeout_ms`/`debug` never inherit a server provider timeout or profile
 debug value. Enabling the browser integration does not create a server provider,
 and a `prebid-server` provider can exist independently from browser injection.
 
+
 ## External Bundle Generation
```
aram356, line 145: 🔧 **wrench** — Second half of the `format-docs` failure: prettier collapses the doubled blank line before this heading. _(Verified together with the suggestion at line 79: `prettier --check` clean.)_

```suggestion
## External Bundle Generation
```

docs/guide/integrations/prebid.md
```diff
@@ -503,8 +532,38 @@ Example EID source mapping:
...
+})
+```
+
+`atsAnalyticsAdapter` is the module stem, while `atsAnalytics` is the registered
+runtime provider. Trusted Server imports the module but does not call
+`pbjs.enableAnalytics`.
+
+Only analytics modules shipped by the pinned Prebid package can be selected.
+Prebid.js 10.26.0 does not include `mavenDistributionAnalyticsAdapter`. Custom
+files, local paths, URLs, and automatic downloads are not supported.
```
aram356, line 566: ⛏ **nitpick** — "Prebid.js 10.26.0 does not include `mavenDistributionAnalyticsAdapter`" is a claim about one pinned version and becomes wrong or unverifiable on the next bump. Describing the failure behavior keeps the guidance true regardless of the pin. The same pattern appears at line 540 ("For the pinned Prebid.js 10.26.0 package").

```suggestion
Only analytics modules shipped by the pinned Prebid package can be selected.
When a requested module such as `mavenDistributionAnalyticsAdapter` is absent
from that package, the generator fails with the config field, the requested
stem, and the pinned Prebid version. Custom files, local paths, URLs, and
automatic downloads are not supported.
```

_(Verified: introduces no new prettier deviation.)_

trusted-server.example.toml
```diff
@@ -467,10 +467,20 @@ client_side_bidders = []          # bidders running via native Prebid.js adapter
...
-# [integrations.prebid.bundle]
-# adapters = ["rubicon"]
-# user_id_modules = ["sharedIdSystem"]
+# Per-bidder / per-zone param overrides (canonical rule form):
+# [[integrations.prebid.bid_param_override_rules]]
+# when.bidder = "examplebidder"
+# when.zone = "header"
+# set = { placementId = "_abc" }
+#
+# Bundle build inputs consumed by `ts prebid bundle`, not by the edge runtime.
```
aram356, line 476: ⛏ **nitpick** — This override-rules example is unrelated to the module map and is not mentioned in the PR description. The keys are valid (`when.bidder`, `when.zone`, `set` match `BidParamOverrideRule`), so it is harmless, but it widens the diff of a config-contract change. Either drop it here or call it out in the description.

```suggestion
# Bundle build inputs consumed by `ts prebid bundle`, not by the edge runtime.
```

crates/trusted-server-js/lib/build-prebid-external.mjs
```diff
@@ -34,247 +37,523 @@ const PREBID_LIVE_INTENT_STANDARD = path.join(
...
+function resolveSpecifierFromInstalledPackage(specifier) {
+  return require.resolve(specifier, { paths: [__dirname] });
+}
 
-  if (entry.moduleName === 'liveIntentIdSystem') {
-    return;
+function resolveOneModule(stem, definition, context) {
+  const metadataFilename = `${stem}.json`;
+  const metadataPath = path.join(context.metadataDir, metadataFilename);
+  const metadataEntries = fs.readdirSync(context.metadataDir);
```
aram356, line 307: ⛏ **nitpick** — `fs.readdirSync(context.metadataDir)` runs once per selected stem, so the default User ID preset plus a few bidders reads the 665-entry directory around fifteen times. Read it once in `resolveBundleModules` (after `canonicalMetadataDir` is established) and pass `metadataEntries` through `context`. Apply manually — the change spans `resolveBundleModules` and this function.

aram356 (changes requested):
## Summary

Typed module map, upstream-only resolution, and the schema-versioned manifest are coherent across the CLI, generator, and shim, and the test coverage is thorough. Two blocking problems: the core runtime config still models the old `bundle.adapters` shape, so the config this PR documents cannot be validated, pushed, or loaded; and the required `format-docs` check fails on `docs/guide/integrations/prebid.md`.

> 4 of the inline comments below carry a one-click GitHub `suggestion` — use **Commit suggestion** (or **Add suggestion to batch** for several at once) to apply them as commits on the PR branch. The remaining comments describe the fix in prose because the change touches other files or lines outside the diff and can't be auto-applied.

## Blocking

### 🔧 wrench

- **Core runtime config still rejects `bundle.modules`** — see inline at `crates/trusted-server-cli/src/prebid_bundle.rs:67`
- **`format-docs` (required) fails: missing blank line after heading** — see inline at `docs/guide/integrations/prebid.md:79`
- **`format-docs` (required) fails: extra blank line before heading** — see inline at `docs/guide/integrations/prebid.md:144`

## Non-blocking

### ⛏ nitpick

- **Version-specific analytics claim will rot** — see inline at `docs/guide/integrations/prebid.md:564`
- **Unrelated override-rules example added to the config template** — see inline at `trusted-server.example.toml:470`
- **Metadata directory re-read per selected stem** — see inline at `crates/trusted-server-js/lib/build-prebid-external.mjs:307`

- 🌱 **Shim parses `runtimeCodes.analytics` but never uses it** — `getExternalBundleManifest()` validates and types the analytics runtime codes, then nothing reads them. The bidder codes drive the `client_side_bidders` check; the analytics codes could drive an equivalent one. A follow-up could wrap `pbjs.enableAnalytics` and log an operator-facing error naming `[integrations.prebid.bundle.modules].analytics` when the requested provider is not in the manifest, since Prebid's own `no analytics adapter found in registry` message does not mention the config key that fixes it.

## CI Status

- integration tests: PASS
- CodeQL: PASS
- format-docs: FAIL (required)
- Analyze (rust): PASS
- cargo test: PASS (required)
- vitest: PASS
- Analyze (actions): PASS
- cargo fmt: PASS (required)

The github-code-quality bot's `fakeBundleMetadata` comment is stale; that helper no longer exists at this head.

crates/trusted-server-js/lib/test/build-prebid-external.test.mjs
```diff
@@ -2,20 +2,503 @@
...
+  const metadataDir = path.join(packageDir, 'metadata', 'modules');
+  const target = path.join(packageDir, 'dist', 'src', 'public', 'exampleModule.js');
+  fs.mkdirSync(metadataDir, { recursive: true });
+  fs.mkdirSync(path.dirname(target), { recursive: true });
+  fs.writeFileSync(path.join(metadataDir, 'exampleModule.json'), JSON.stringify(metadata));
+  fs.writeFileSync(target, 'export {};\n');
+  return { temp, packageDir, metadataDir, target };
+}
+
+function fakeBundleMetadata() {
```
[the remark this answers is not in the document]
ChristianPavilonis: This was already addressed before the current review pass: `fakeBundleMetadata` is no longer present at the current head.

docs/guide/integrations/prebid.md
ChristianPavilonis: Added an explicit upgrade section in `7c012daf`. It covers the strict migration from `bundle.adapters` / `bundle.user_id_modules` to exact stems under `bundle.modules`, and explains how pre-schema manifests disable bidder and User ID diagnostics until the external bundle is rebuilt.

crates/trusted-server-js/lib/test/prebid-artifact-integration.test.mjs
ChristianPavilonis: Updated `expectNoUnexpectedNetworkActivity` in `7c012daf` to inspect every recorded fetch, XHR, beacon, and image URL and reject any non-first-party origin. The production artifact test passes with the stronger assertion.

crates/trusted-server-js/lib/test/prebid-artifact-integration.test.mjs
ChristianPavilonis: Renamed the test in `7c012daf` to state exactly what it checks: both manifests, plus the no-analytics artifact hash and SRI.

crates/trusted-server-cli/src/prebid_bundle.rs
ChristianPavilonis: Fixed in `7c012daf`. Core now models the strict nested `bundle.modules` shape, the shared fixtures use it, and regression tests cover successful parsing plus rejection of `adapters`, `user_id_modules`, `analytics_adapters`, and unknown nested fields with Prebid both enabled and disabled.

docs/guide/integrations/prebid.md
ChristianPavilonis: Fixed the missing blank line in `7c012daf`. The full docs Prettier check now passes.

docs/guide/integrations/prebid.md
ChristianPavilonis: Removed the extra blank line in `7c012daf`. The full docs Prettier check now passes.

docs/guide/integrations/prebid.md
ChristianPavilonis: Reworded this in `7c012daf` so the guidance describes pinned-package availability and the generated missing-module error without hardcoding a claim about Prebid.js 10.26.0.

trusted-server.example.toml
ChristianPavilonis: Removed the unrelated override-rules example from this PR in `7c012daf`.

crates/trusted-server-js/lib/build-prebid-external.mjs
ChristianPavilonis: Hoisted the exact-case metadata directory read into `resolveBundleModules` in `7c012daf` and pass a shared `Set` into each module resolution. The generator and production artifact tests pass.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -56,38 +56,58 @@ const pbjs: PbjsGlobal = (
...
-    userIdModules: sanitizeManifestList(manifest.userIdModules),
+    schemaVersion: 1,
+    modules: {
+      bidder: parseManifestList(modules.bidder),
+      userId: parseManifestList(modules.userId),
+      analytics: parseManifestList(modules.analytics),
+    },
+    runtimeCodes: {
+      bidder: parseManifestList(runtimeCodes.bidder),
+      analytics: parseManifestList(runtimeCodes.analytics),
```
aram356, line 109: 🌱 **seedling** — `runtimeCodes.analytics` is parsed and validated here but never read anywhere. The bidder list drives the `client_side_bidders` check in `installPrebidNpm`; the analytics list has no equivalent consumer.

Not for this PR, and nothing here is wrong: stamping the codes is what makes a later check possible. Recording it so the asymmetry is deliberate rather than forgotten.

A follow-up could wrap `pbjs.enableAnalytics` and log an operator-facing error when the requested provider is absent from the manifest. Prebid's own `no analytics adapter found in registry` message names the provider but not the config key that fixes it, which is exactly the gap `[integrations.prebid.bundle.modules].analytics` closes. That would give analytics the same misconfiguration feedback bidders already get.

aram356 (approved):

PR #1094, opened 2026-08-31
Open the integration seam so a vendor module can live outside core

jwrosewell:
> **This description has been edited.** The pull request now lands one
> configuration convention for everything a deployment can switch on, and the
> auction side is inside it. If you read an earlier version of this text, the
> sections worth re-reading are "One convention for every provider", "The
> auction side becomes provider types", "Why this rather than what `main` does
> today" and the migration table. The rest is the same work described the same
> way.
>
>
> 1. The ad server, plain OpenRTB, Prebid Server and APS became provider types
>    that an integration registers, rather than names written into core. The
>    earlier text said the auction was not opened. That is no longer true.
> 2. `[auction.providers]` became `[demand]` and `[auction] mediator` became
>    `[adserver]`. `profile` became `implementation`, and what sat in
>    `profile_config` now sits flat in the provider's own table.
> 3. The word mediator is gone. It is "ad server" in prose and `adserver` in
>    configuration, and the auction response metadata that read
>    `parallel_mediation` now reads `parallel_adserver`.
> 4. An integration is switched on by being named in `[integration] provider`
>    rather than by `enabled = true` inside its own block, and the table is
>    `[integration]` rather than `[integrations]`.
> 5. `[ec.providers.<name>]` became `[ec.<name>]`, and
>    `[permission_signal] sources` became `[permission_signal] provider`.
> 6. The demand seam's two request hooks hand an implementation only the
>    parts of a request it owns, the extension objects and the outbound
>    headers, rather than the whole request, and the driver refuses an
>    implementation that writes `ext.trusted_server`.
> 7. Every page integration refuses a setting it does not know, which the
>    configuration rules page promised and nine integrations did not do.
> 6. Every name a person types is snake_case, so `host-signals`,
>    `client-fixed`, `gpp-sale-opt-out` and `us-privacy` became `host_signals`,
>    `client_fixed`, `gpp_sale_opt_out` and `us_privacy`.
> 7. A new reference page,
>    [Configuration Rules](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/docs/guide/configuration-rules.md),
>    states the convention, what is checked before a request is served, and how
>    to move a file from the previous layout.
>
> There is no deprecation period for the auction keys. A configuration still
> carrying `[auction.providers]` or `[auction] mediator` is refused at startup
> with a message naming where the setting moved to.

Stacks on the provider documentation set (#1047), the last of the five pull
requests that decompose #838, so this pull request sits on the whole provider
series. It also depends directly on the `EdgeCookieProvider` and
`DeviceProvider` traits from the Edge Cookie provider seam (#1043) and the
device and location seam (#1044), whose providers a registration can now carry,
and on the permission model (#1045), where an unset `[geo] provider` makes no
host geo call. The stack has six pull requests
sixth.
[Compare `split/5-response-hook-docs` with `split/7-integration-seam-impl`](https://github.com/jwrosewell/trusted-server/compare/split/5-response-hook-docs...split/7-integration-seam-impl)

The design specs for the series are carried by the spec pull request (#1084),
which has no code. This pull request implements
[`2026-08-27-integration-provider-seam-design.md`](https://github.com/jwrosewell/trusted-server/blob/split/6-integration-seam/docs/superpowers/specs/2026-08-27-integration-provider-seam-design.md),
where sections 3.1 to 3.6 define the seam and section 8 records what
implementing it found, so read #1084 first and this pull request as the answer
to it.

## What this does and why

[`trusted-server-core`](https://github.com/jwrosewell/trusted-server/tree/split/7-integration-seam-impl/crates/trusted-server-core)
carries nine vendor integrations as core code because every place an
integration plugs in is closed. The builder table is private, browser
JavaScript is fixed at compile time, deploy validation names each vendor's
configuration type, the auction names its vendors in core, and two vendors
reach into core through named types. Every new vendor is therefore another core
change, most recently the LiveRamp module (#1054).

This pull request opens those places. An integration can now ship in its own
crate with its Rust, its browser script, its configuration type, its deploy
rules and its tests, and an adapter composes that crate into a deployment at
startup without core naming the vendor. No existing integration changes
behavior, the served script keeps its exact bytes and its `?v=` hash, and the
built-in set still registers through the same path a vendor crate would use.

What is now composable:

| Seam | How a crate outside core reaches it |
| --- | --- |
| Integration registration | `IntegrationBuilder::new(id, source, build, validate)`, passed to an adapter's `routes_with_registrations`, which builds the registry with `IntegrationRegistry::with_plan_and_registrations` |
| Browser script | `.with_js_module(CarriedJsModule { source, sha256 })` on the registration |
| Deploy validation | the builder's validate function, run by `validate_settings_for_deploy_with` |
| Request preparation | `.with_request_preparer(...)`, run by `IntegrationRegistry::prepare_request` |
| Identity, location and device | `.with_ec_provider(...)`, `.with_geo_provider(...)` and `.with_device_provider(...)`, selected by `[ec] provider`, `[geo] provider` and `[device] provider` |
| Auction demand | `.with_demand(...)`, selected by `[demand] provider` |
| The ad server | `.with_adserver(...)`, selected by `[adserver] provider` |

### One convention for every provider

Everything a deployment can switch on is a provider, and every provider type is
configured the same way.

```toml
[<type>]
provider = "<name>"          # a string where one runs, a list where several run

[<type>.<name>]              # only when that provider has settings
setting = "value"
```

The types are `ec`, `geo`, `device`, `permission_signal`, `demand`, `adserver`
and `integration`. A type that runs one provider takes a string, a type that
runs several takes a list, and a provider with nothing to set needs no table at
all. A table's name is the implementation it configures, unless the table
carries an `implementation` line, which is how two Prebid Servers run side by
side under names a deployment chooses for itself. Every name a person types is
snake_case.

Four things stop a deployment rather than being tolerated:

- A `[<type>.<name>]` table that its type's `provider` does not name.
- A `provider` entry, or an `implementation` line, naming an implementation
  this build does not have. The message lists the ones it does have.
- A setting the chosen provider does not know, because every provider rejects
  keys it does not recognize.
- A `demand` or `adserver` endpoint that is not HTTPS. Plain HTTP is allowed
  only to `127.0.0.1`, `::1` or `localhost`, so a local test stack runs without
  certificates and nothing leaves the machine unencrypted.

The rules that hold for every type live in one place,
[`crates/trusted-server-core/src/provider_table.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/provider_table.rs),
so a new provider type inherits them instead of writing them again. The whole
convention, the checks and the move from the previous layout are written up for
operators in
[`docs/guide/configuration-rules.md`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/docs/guide/configuration-rules.md).

### The auction side becomes provider types

The ad server, plain OpenRTB, Prebid Server and APS are now implementations
that an integration registers through `.with_demand(...)` and
`.with_adserver(...)` in
[`crates/trusted-server-core/src/integrations/mod.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/integrations/mod.rs),
selected by `[demand] provider` and `[adserver] provider`. The four that ship
here, in
[`openrtb.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/integrations/openrtb.rs),
[`prebid_server.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/integrations/prebid_server.rs),
[`aps.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/integrations/aps.rs)
and
[`adserver_mock.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/integrations/adserver_mock.rs),
go through the same registration a vendor crate uses.

What an implementation may decide is stated once, in
[`crates/trusted-server-core/src/auction/demand.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/auction/demand.rs).
A shared OpenRTB driver builds every standard request field, sends the request
and enforces the privacy rules, and an implementation chooses only its field
policy, its own extensions, its outbound headers and how its responses are
read. The compiled plan in
[`crates/trusted-server-core/src/auction/plan.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/auction/plan.rs)
resolves each selected name to a registered implementation, so
`crates/trusted-server-core/src/auction/profile.rs` and its fixed list of three
profiles are deleted.

`[auction]` keeps the settings that belong to the auction itself, being
`enabled`, the whole-auction timeout and creative handling, with
`[auction.bidders.<code>] provider` naming the demand source a browser bidder
code is sent to.

### The provider interfaces are asynchronous, and providers get the services

Before this change every provider trait method was synchronous while every
platform service was asynchronous, and a provider was handed no services at
all, so a provider that needed to call a backend, read a key-value store or
fetch a secret could not be written. The methods that do the work on all three
provider interfaces (`EdgeCookieProvider::generate`, `DeviceProvider::detect`
and `PlatformGeo::lookup`) are now asynchronous and take `&RuntimeServices`,
and so is `resolve_from_client`, which has to verify what the browser posted.
Every implementation was converted, the built-in ones included. Finalizing the
Edge Cookie response is asynchronous too and takes the services, because
`main`'s orphaned identifier recovery asks the selected provider for a
replacement identifier.

The traits keep their `Send + Sync` bound and use `#[async_trait(?Send)]`, as
`PlatformHttpClient` in this repository already does, so a provider stays safe
to share while its future stays on one thread.

Two tests in
[`crates/trusted-server-core/src/ec/mod.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/ec/mod.rs),
`a_provider_reads_a_platform_service_through_the_services_it_is_given` and
`a_geo_provider_reads_a_platform_service_through_the_services_it_is_given`,
drive a provider that reads a value from the configuration store it is handed,
through the production path, and assert the value reaches the output. They
exist because once everything compiled and passed, nothing anywhere read the
services parameter, and a parameter no code exercises is not a working seam.

### Two supporting changes

The bid renderer is no longer an enum with a single APS variant. It is a type
tag plus the payload the demand provider supplies, serialized flat so the
response a page receives is byte for byte what it was, and `ApsRendererV1`
moves into the APS integration. DataDome's cache and origin-path marker becomes
a neutral `PersonalizedResponse` request extension that any integration may
set, so core acts on the marker without knowing which integration asked for it.

### One design decision made while stacking this on #1045

Two designs for `[geo]` met when this branch moved onto the stack, and they
disagreed about what an unset selector means. The permission model (#1045)
treats unset as resolving nothing and making no host geo call, and this branch
had treated unset as leaving the host's own lookup in place. The permission
model's meaning is kept, because a deployment that has not asked for a host geo
lookup should not be making one. Unset and `"none"` therefore both resolve to
`DisabledGeo`, `"platform"` is the explicit opt-in to the adapter's own lookup,
and any other value names a module that declared a location provider.

### Where to start

1. **[`crates/trusted-server-adapter-axum/tests/seam_probe.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-adapter-axum/tests/seam_probe.rs).**
   Every seam is driven end to end from a crate core does not know about,
   through the Axum adapter's real router. The tests turn on observable
   outcomes, being the bytes served, the JSON a route returns and the error a
   startup or deploy check produces, rather than on a function having been
   called. Start here, because a seam is only proven by an implementation that
   is not the built-in one.
   `ec_provider_generates_an_identifier_with_the_modules_prefix` shows the
   module's own Edge Cookie provider ran, because its identifier starts
   `seam-probe-`, which the built-in HMAC provider cannot produce, and
   `ec_selector_naming_a_module_starts_the_adapter` starts the Axum router with
   that provider selected.
2. **[`crates/integrations/seam-probe/src/lib.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/integrations/seam-probe/src/lib.rs)**,
   the fixture those tests drive. It is a workspace member and a dev-dependency
   of the Axum adapter only, so no production build reaches it. One
   registration carries a browser module, a proxy route, a request preparer,
   location, Edge Cookie and device providers and its own configuration block.
3. **[`crates/trusted-server-core/src/provider_table.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/provider_table.rs)**,
   the table every provider type shares, and
   [`auction/plan.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/auction/plan.rs),
   which resolves `[demand]` and `[adserver]` names to registered
   implementations and reports an unknown one by listing the ones this build
   has.
4. **[`crates/trusted-server-core/src/integrations/registry.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/integrations/registry.rs)**,
   `with_plan_and_registrations`. Building the registry refuses a duplicate id,
   naming both sources. It checks the SHA-256 of a carried browser module, and
   resolves `[ec] provider`, `[geo] provider` and `[device] provider` against
   the modules that declared each capability.
5. **[`crates/trusted-server-core/src/tsjs_bundle.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/tsjs_bundle.rs)**,
   new. Composition of the served script moves out of `trusted-server-js` and
   into core, keyed on content rather than on ids, because a module a vendor
   crate carries is not in the compile-time map. The byte rule is unchanged,
   being core first then each part joined by `;\n`. The tests in that file
   compare the composed bytes and hash against
   `trusted_server_js::concatenate_modules` and
   `trusted_server_js::concatenated_hash`, which is the evidence that no
   current `?v=` value moves.
6. **[`crates/trusted-server-core/src/config.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/config.rs)**.
   The hand-written list of vendor configuration types is replaced by a loop
   over the builders, and the auction plan is compiled and checked on the same
   path, so a bidder route naming a demand source `[demand] provider` does not
   select is caught before a push.
7. **[`crates/trusted-server-core/src/auction/types.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/auction/types.rs)**,
   `BidRenderer`. This is the riskiest single change, because the wire shape
   `{"type":"aps", ...}` has to survive byte for byte.
   `aps_renderer_serializes_to_versioned_camel_case_contract` pins the
   serialized form and `renderer_bid_id_key_matches_the_serialized_form` pins
   the one field the publisher reads by key.

[`crates/trusted-server-core/build.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/build.rs)
is also new. The migration guard used to list core source files by hand with
`include_str!`, so a vendor moving out of core would break the build rather
than a test. The list is now generated from the source tree, so a file that
leaves core leaves the guard with nobody editing anything.

## Why this rather than what `main` does today

**Core names vendors it has no business knowing.**
[`crates/trusted-server-core/src/auction/plan.rs` on `main`](https://github.com/IABTechLab/trusted-server/blob/main/crates/trusted-server-core/src/auction/plan.rs)
holds `MOCK_MEDIATOR_ID = "adserver_mock"` and refuses any other ad server. It
special-cases the profile ids `prebid-server` and `aps` by name when it
compiles a plan, and
[`crates/trusted-server-core/src/auction/profile.rs` on `main`](https://github.com/IABTechLab/trusted-server/blob/main/crates/trusted-server-core/src/auction/profile.rs)
holds the three profiles a deployment may choose, being `standard`,
`prebid-server` and `aps`, as a fixed list with a closed enum behind it. A
vendor cannot add a demand source or an ad server without a pull request
against core, which is the cost this series exists to remove.

**The same idea is spelled four different ways.** An integration is switched on
by `enabled = true` inside its own block. A demand source is switched on by a
key appearing in `[auction.providers]`. A permission signal is switched on by
its name appearing in a `sources` list. An Edge Cookie provider is switched on
by `[ec] provider` naming a nested `[ec.providers.<name>]` block. Two of those
four are on `main` today. The other two arrived inside this stack, in the Edge
Cookie provider seam (#1043) and the permission model (#1045), which shows how
quickly a fifth spelling appears when no rule says what a spelling looks like.
One convention now covers all of them, and the next provider type gets the same
one without anyone deciding again.

**For the people who run a deployment**, one convention plus refusal at startup
changes what a mistake costs:

- A table left behind after a provider is switched off is caught rather than
  quietly ignored. On `main` an `[integrations.datadome]` block that outlives
  its integration keeps every setting it had, and nothing tells the next reader
  whether those settings are live. A leftover `[auction.providers.pbs-old]`
  entry is worse, because on `main` the entry is what switches the demand
  source on, so a stale one is a live bidder nobody meant to keep. Now a table
  has to be named by a `provider` line or the deployment does not start.
- A misspelled setting fails instead of being dropped. Every provider rejects
  keys it does not know, so `timeout_millis` where the setting is `timeout_ms`
  stops the deployment rather than silently leaving the default in place, which
  is the kind of fault that otherwise turns up much later in a latency graph.
- Switching a provider is a one-line change that reads plainly in review.
  `provider = ["pbs_main"]` becoming `provider = ["pbs_main", "aps_main"]` says
  what happened. On `main` the same intent is a new nested block, a `profile`
  string, a `profile_config` object and a `protocol` key, and a reviewer has to
  reconstruct the outcome from four places.
- An endpoint cannot quietly be plain HTTP. Demand and ad server endpoints must
  be HTTPS, or HTTP to a loopback host for a local test stack.
- An old file is not half accepted. A configuration still carrying
  `[auction.providers]` or `[auction] mediator` is refused, and the error names
  the table the setting moved to instead of leaving the operator to search for
  it.

**For the people who write providers**, a provider plugs in through one
registration and inherits every check above without writing any of them. Core
code names no vendor, so shipping a demand source, an ad server, an identity
provider, a location provider or a device provider means publishing a crate,
not editing core and waiting for a core release. The seam probe proves this
from the outside, because it is a crate core does not know about and it reaches
every one of those seams through the real router.

**What this costs.** These are breaking changes, and the auction keys get no
deprecation period. That is a deliberate choice for a major release, taken
because the alternative is carrying two spellings of every selector
indefinitely, and because a configuration refused with a message naming the fix
is cheaper to migrate than one that half works.

## Moving from the previous layout

| Previous | Now |
| --- | --- |
| `[integrations.<id>]` with `enabled = true` | `<id>` in `[integration] provider`, and `[integration.<id>]` only for settings |
| `[ec.providers.<name>]` | `[ec.<name>]` |
| `[permission_signal] sources` | `[permission_signal] provider` |
| `host-signals`, `client-fixed`, `gpp-sale-opt-out`, `us-privacy` | `host_signals`, `client_fixed`, `gpp_sale_opt_out`, `us_privacy` |
| `[auction.providers.<id>]` with `protocol`, `profile` and `profile_config` | `[demand] provider` and `[demand.<name>]`, with `implementation` and the settings flat in the table |
| `profile = "standard"` | `implementation = "openrtb"` |
| `[auction] mediator = "adserver_mock"` and `[integrations.adserver_mock]` | `[adserver] provider = "adserver_mock"` and `[adserver.adserver_mock]` |
| `[integrations.aps] rendering_mode` | `rendering_mode` in the `[demand.<name>]` table of the `aps` provider |
| `[debug.auction_html_comment_options] include_mediator_response` | `include_adserver_response` |

The word mediator goes with those keys. It is "ad server" in prose and
`adserver` in configuration, and the auction response metadata that read
`parallel_mediation` now reads `parallel_adserver`.

A file in the old shape is not read on a best-effort basis. `[auction]
providers` and `[auction] mediator` are both refused with a message naming
where the setting moved to.

## How it was verified

The head this pull request shows now is `ede61f685`. The gates below were run
locally on `6a7cbb91d`, and the four commits after it change only test code
and one comment, each verified by the gate that reaches it.

On Windows, `cargo fmt --all --check`, Clippy with warnings denied on the
Fastly, Axum, Cloudflare and Spin adapters on their native and wasm targets
and on the codegen and parity crates, the core suite (2,982 tests and 18
doctests), the Axum, Cloudflare and Spin adapter suites, the seam-probe
fixture's tests, the cross-adapter parity suite, the Fastly release wasm
build, and the docs Prettier check, ESLint and VitePress build with the
pinned Prettier all pass. The CLI crate does not build on a Windows host,
because a dependency uses an unstable Windows feature, so its tests and its
Clippy job ran under WSL, where they pass at 80 unit tests and 7 environment
overlay tests. The Fastly adapter's own tests run only under Viceroy on
wasm32-wasip1, where they pass at 192, and the device, geo, OpenRTB and
JavaScript crates pass under Viceroy as well.

Three of those runs found fixtures the native gates could not reach and that
this pull request had left on the old layout: two `ts audit` tests counted a
commented example the template now carries, the CLI environment overlay test
still built its project on `[auction.providers]`, and three Fastly fixtures
did the same. Each is fixed in a commit of its own on this head.

On CI, every check passes on `ede61f685` except CodeQL:
[cargo test](https://github.com/IABTechLab/trusted-server/actions/runs/35033487384/job/104597167492),
[the Axum tests on Ubuntu](https://github.com/IABTechLab/trusted-server/actions/runs/35033487384/job/104597167555),
[the CLI tests](https://github.com/IABTechLab/trusted-server/actions/runs/35033487384/job/104597167535),
[format-docs](https://github.com/IABTechLab/trusted-server/actions/runs/35033487319/job/104597167452),
[the integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/35033487318/job/104599327315)
and
[vitest](https://github.com/IABTechLab/trusted-server/actions/runs/35033487384/job/104597167290).
[CodeQL](https://github.com/IABTechLab/trusted-server/runs/104597325349)
reports one high alert on a log line in the orchestrator that names the ad
server and its timeouts. The query reaches that line by tracing a test
helper's secret store into the log, and the line logs neither request data
nor a secret, so the alert stands as reported rather than being suppressed.

The CI workflow runs the seam-probe fixture crate's own tests, but no clippy
alias or workflow step names that crate, so its source is not linted in CI.

## What is not in this change

The design in #1084 is wider than this pull request. Every part of it this pull
request does not deliver is listed here.

**No implementation has physically moved out of core yet.** The auction, Edge
Cookie, location and device seams are open and the built-in implementations go
through them, but they still ship inside
[`crates/trusted-server-core`](https://github.com/jwrosewell/trusted-server/tree/split/7-integration-seam-impl/crates/trusted-server-core).
Each vendor moves in its own pull request after this one, which is the point of
doing the seam once.

**Nothing is refused for a missing host signal (section 3.6).** A registration
declares no host-signal requirements, so a provider that needs a signal the
running adapter does not expose is not turned back at startup.

**The acceptance round trips run on Axum, not Fastly (section 6, items 1 and
2).** Both items ask for the round trip on the Fastly adapter, and the tests
here run through the Axum adapter, because the Fastly adapter cannot yet take a
vendor crate (below). Item 1 also asks that a module's hooks run in the right
order, and the tests show the preparer runs before routing and exactly once,
but no test checks the order of several hooks on one registration.

**There is no finalize hook (section 3.5).** The prepare half is done, and
`IntegrationRegistry::prepare_request` replaced the direct
`gpt_diagnostics::prepare_request` calls in all four adapters. The finalize
half is not. The registration builder has no finalize method, and the GPT
diagnostics decision is still a named type in
[`html_processor.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/html_processor.rs)
and
[`publisher.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/publisher.rs),
so a GPT move needs that hook first.

**A module's own deploy rules still run nowhere an operator can reach (section
8, items 1 and 6).** `ts config validate` and `ts config push` go through
`TrustedServerAppConfig`, which calls `validate_settings_for_deploy` with no
extra builders, so only core's rules run on the path an operator uses. The
registry calls a builder's build function and not its validate function, so a
vendor whose checks live in `validate` has them enforced on no path unless the
deployment's own code calls `validate_settings_for_deploy_with` with that
builder. The seam probe works around this by repeating its check inside its
build function, which every vendor would have to copy.
[`docs/guide/integration-guide.md`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/docs/guide/integration-guide.md)
records the gap, and it needs a decision, either the CLI is built per
deployment with its vendor crates, or the registry runs `validate` when it
builds.

**A Fastly deployment still cannot take a vendor crate (section 8, item 7).**
The Axum, Cloudflare and Spin adapters expose `build_state_with_registrations`
and `routes_with_registrations` publicly. The Fastly adapter's
`build_state_with_registrations` is `pub(crate)` and the crate has no library
target, so composing a module into a Fastly deployment means editing that
adapter. Fastly is the primary deployment target, so this gap decides whether
the seam is usable in production or only in the dev server, and closing it
should come before the first vendor is asked to use the seam.

**A provider can still be resolved more than once per request (section 8, item
3).** On `POST /auction` location is looked up to build the Edge Cookie context
and again in `handle_auction`, and the seam probe's proxy route calls its
location provider as well. This pull request adds no per-request provider
context, so a module sharing one backend across identity, location and device
has nowhere to hang a single call per request.

**Request preparers still cover different routes on each host (section 8, item
5).** Replacing the vendor-named call with the registry call did not make the
covered routes the same. Axum runs preparers once before routing, Cloudflare in
its shared handler wrapper and its fallback dispatch, Fastly after its early
admin dispatch and in its fallback, and Spin only in the auction handler, the
page-bids handler and the fallback. A module that strips its own reserved query
or cookie is therefore protected on a different set of routes depending on the
host.

**One core reader still reaches into an APS payload (section 8, item 4).** The
`hb_adid` fallback in
[`publisher.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/publisher.rs)
reads the APS renderer's bid id. It goes through the neutral accessor
`BidRenderer::payload_field`, which reads one field instead of cloning the
whole payload, but `publisher.rs` still imports `APS_RENDERER_TYPE` and
`APS_RENDERER_BID_ID_KEY`, so the APS move needs a neutral answer for that
fallback and not only the renderer contract.

**Core TypeScript still imports APS directly (section 8, item 8).**
[`auction.ts`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-js/lib/src/core/auction.ts)
and
[`types.ts`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-js/lib/src/core/types.ts)
import and name the APS renderer, so generalizing the Rust renderer does not by
itself move APS out.

**DataDome is still named in core outside its own integration.**
[`publisher.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/publisher.rs)
reads `DataDomeClientTagSuppressed` and
[`html_processor.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/src/html_processor.rs)
imports it, so that DataDome's head injection can leave its client-side tag
out. That marker decides nothing about caching or the origin path any more,
which the neutral `PersonalizedResponse` marker now does, and it moves out with
the DataDome integration, as the code comment at the read site in
`publisher.rs` says.

**The `ts audit` vendor table is unchanged.** The
[`ts audit`](https://github.com/jwrosewell/trusted-server/tree/split/7-integration-seam-impl/crates/trusted-server-cli/src/commands/audit)
command keeps its own vendor detection patterns outside the registry, and how
the CLI learns a vendor's detection pattern from a crate is still open.

**One deliberate deviation from the spec.** Section 4 says the source-file
guard should drop the nine vendors' files. This pull request instead generates
the whole guard list from the source tree through the new
[`crates/trusted-server-core/build.rs`](https://github.com/jwrosewell/trusted-server/blob/split/7-integration-seam-impl/crates/trusted-server-core/build.rs),
so a file joins or leaves the guard on its own and no vendor move has to touch
it. The goal is the same and the mechanism is different.

**What the round trip does not exercise.** The seam probe declares a proxy, a
carried browser module, a request preparer, and location, Edge Cookie and
device providers. It declares no head injector, attribute rewriter, script
rewriter, HTML post processor or request filter, and it uses neither the
deferred nor the standalone script delivery flag, so the round trip does not
exercise those from outside core.


aram356:

jwrosewell:
**Sequencing note on the `rc/202608` release branch and this stack.**

While checking today's activity I found that `rc/202608` (cut from `main` at d516a9e94 on 28 August, currently at 07dfc1c6d) carries 67 merged PRs, including #940, #823, #1016, #1036, #1039 and #1079, in 643 commits. `main` is an ancestor of it, so it can land on `main` as a fast-forward at any time.

A merge simulation of `rc/202608` against this PR's head (b557d87e4) conflicts in 33 files, listed below. The same simulation against #1084's head (180060402) is clean, because that branch is specs only.

- `crates/trusted-server-adapter-axum/src/app.rs`
- `crates/trusted-server-adapter-axum/tests/routes.rs`
- `crates/trusted-server-adapter-cloudflare/src/app.rs`
- `crates/trusted-server-adapter-fastly/src/platform.rs`
- `crates/trusted-server-adapter-spin/src/app.rs`
- `crates/trusted-server-adapter-spin/src/platform.rs`
- `crates/trusted-server-core/src/auction/README.md`
- `crates/trusted-server-core/src/auction/formats.rs`
- `crates/trusted-server-core/src/auction/mod.rs`
- `crates/trusted-server-core/src/auction/orchestrator.rs`
- `crates/trusted-server-core/src/auction/test_support.rs`
- `crates/trusted-server-core/src/config.rs`
- `crates/trusted-server-core/src/config_payload.rs`
- `crates/trusted-server-core/src/ec/device.rs`
- `crates/trusted-server-core/src/integrations/aps.rs`
- `crates/trusted-server-core/src/integrations/datadome/protection.rs`
- `crates/trusted-server-core/src/integrations/mod.rs`
- `crates/trusted-server-core/src/integrations/registry.rs`
- `crates/trusted-server-core/src/platform/traits.rs`
- `crates/trusted-server-core/src/proxy.rs`
- `crates/trusted-server-core/src/publisher.rs`
- `crates/trusted-server-core/src/response_privacy.rs`
- `crates/trusted-server-integration-tests/fixtures/configs/trusted-server.integration.toml`
- `docs/guide/api-reference.md`
- `docs/guide/auction-orchestration.md`
- `docs/guide/configuration.md`
- `docs/guide/ec-setup-guide.md`
- `docs/guide/error-reference.md`
- `docs/guide/fastly.md`
- `scripts/template-cache-local-test.sh`
- `trusted-server.example.toml`

Reproduce:

```
git fetch upstream main rc/202608
git merge-tree --write-tree --name-only upstream/rc/202608 b557d87e4
```

The request is the one we made on #940. Please land this stack (#1043 to #1047, #1084 and this PR, all based on `main`) before `rc/202608` merges to `main`, or if the release branch has to go first, say so here so we rebase once against a known base rather than chase one that moves as it takes more PRs. Every one of the 33 files is one we would have to re-resolve by hand, and the count rises with each PR merged into the release branch.

PR #1106, opened 2026-09-01
Match CSS url() quoting rules when rewriting creative styles

### LINKED ISSUE #1114: Creative CSS and srcset rewriters mishandle malformed values
## Description

Two parsers in `crates/trusted-server-core/src/creative.rs` handle malformed input in ways that diverge from how a browser reads the same markup.

**`rewrite_style_urls`** decides whether a `url()` value is quoted by looking at its first character, then steps back one **byte** from the closing paren on the assumption that a matching quote is there. Nothing verifies it. When the preceding character is multi-byte, the resulting slice is not on a character boundary. When the quotes simply do not match, the closing delimiter is silently rewritten to match the opening one.

It also ends the value at the first `)`, but a quoted CSS string may legally contain one — so `url("https://cdn.example/a)b.png")` is truncated, and the intended URL is left unrewritten.

Values it cannot faithfully resolve are still rewritten: CSS escapes are not resolved here, and a raw newline (which preprocessing also produces from a carriage return or form feed) makes the value a bad string the browser discards. In both cases a proxy URL is emitted for a resource the page never requests.

**`split_srcset_candidates`** re-derives per-candidate facts — whether the candidate uses the `data:` scheme, and whether whitespace has appeared — from the whole candidate prefix at every comma, allocating a fresh lowercased copy each time. Those facts belong to the candidate, not to each comma, so a candidate containing many commas rescans and reallocates repeatedly. The creative module's test suite took 64s on one such input; it now takes 0.46s.

## Expected behavior

- A `url()` value is treated as quoted only when a matching closing quote is present, and slicing never lands mid-character.
- A quoted value ends at its closing quote, so an inner `)` is part of the URL.
- A value whose bytes cannot be resolved to the URL the browser will request is passed through untouched.
- Candidate state in `srcset` is tracked as the scan advances rather than re-derived per comma.

## Affected area

HTML processing / JS injection

## Done when

- [ ] A malformed `url()` value round-trips unchanged instead of being mangled or slicing mid-character
- [ ] A quoted value containing `)` is rewritten in full
- [ ] Values carrying an escape or a raw newline are left alone
- [ ] `srcset` scanning is linear in the candidate length
- [ ] Each behaviour has a test that fails when its fix is removed

prk-Jr:
## Summary

- The CSS `url()` rewriter decided the extent of a value by scanning for quote and paren positions. That is not how a browser reads a declaration, so some values were rewritten in ways a browser would not, and others were left unrewritten — most visibly a quoted value containing `)`, which the scanner truncated at the first `)` after `url(`, leaving the intended URL unproxied.
- The scanner is replaced by a token-level walk of the CSS grammar (`cssparser`), so the extent of a value and the resolution of its escapes come from the same rules a browser applies. The walk is property-agnostic: it rewrites URL references wherever they appear rather than from an enumerated list of URL-bearing properties, and it leaves fragment-only references such as `filter: url(#blur)` alone.
- `split_srcset_candidates` re-derived per-candidate facts from the whole candidate prefix at each comma. Those facts belong to the candidate, so they are now tracked as the scan advances.

## Output safety

Style blocks retain raw `&tstoken=` query separators so signed proxy URLs remain valid. CSS output is limited to 10 MiB during construction, including unchanged text and CSS-serialized strings; no oversized complete output is allocated before checking. The same bound applies to temporary CSS output inside HTML. Whole-response CSS refusals propagate as errors; inline CSS refusals drop the affected styling.

## Changes

| File | Change |
| ---- | ------ |
| `Cargo.toml`, `crates/trusted-server-core/Cargo.toml` | Add `cssparser` as a direct dependency. It is already in the lockfile via `lol_html`, so no new package is vendored |
| `crates/trusted-server-core/src/creative.rs` | `rewrite_style_urls`: replace positional quote scanning with a `cssparser` grammar walk. Escapes are already resolved when the value is read, and a malformed value — which the tokenizer reports as a bad URL or bad string, exactly what a browser discards — is left untouched rather than guessed at |
| `crates/trusted-server-core/src/creative.rs` | Coverage the walk adds: `src()` alongside `url()`, bare-string candidates in `image-set()`, `@import` preludes in both the `url()` and bare-string forms, and `var()` / `env()` fallbacks, which are substituted in place |
| `crates/trusted-server-core/src/creative.rs` | A rewritten reference keeps the shape it was read in — `url()` as `url()`, `src()` as `src()`, a bare string as a bare string — because the forms are not interchangeable to a browser. Only the value inside is replaced, and it is re-quoted; anything not rewritten keeps its original bytes |
| `crates/trusted-server-core/src/creative.rs` | New `MAX_CSS_NESTING_DEPTH`: the CSS is supplied by the upstream creative, and a stack overflow aborts the guest, so the recursion the walk performs is bounded. The bound counts scopes the walk recurses into, not how a URL is spelled, so `url(x)` and `url("x")` are admitted at the same depth. Nesting past it discards the whole stylesheet rather than passing the deeper bytes through, which would turn the bound into a way around the rewrite |
| `crates/trusted-server-core/src/creative.rs` | `rewrite_css_body` returns `Result`. A stylesheet the depth bound refuses previously left the CSS proxy path serving an empty `200`, indistinguishable from a stylesheet the origin legitimately served empty. It now reports the refusal so `finalize_proxied_response` sets a status, matching the oversized-body path. The markup path still drops only the offending `<style>` block or attribute, and the log names which |
| `crates/trusted-server-core/src/creative.rs` | `split_srcset_candidates`: derive candidate scheme and whitespace state as the scan advances instead of from the candidate prefix at each comma; corrected the doc note, which described behavior the function did not have |
| `crates/trusted-server-core/src/creative.rs` | Tests for quoted, unquoted, unterminated, mismatched and escaped values; `src()`, `image-set()`, `@import` and fallback coverage; the depth boundary in every spelling of a URL; and the multi-comma `data:` srcset case |

## Behavior

| Input | Rewritten as |
| ----- | ------------ |
| `url("https://cdn.example/a)b.png")` | whole value proxied, inner paren percent-encoded |
| `url("https://t.example/\70 ixel.gif")` | proxied — the escape is resolved before the decision |
| `src("https://cdn.example/a.woff2")` | proxied, re-emitted as `src()` |
| `image-set("https://cdn.example/a.png" 1x)` | bare string proxied, still a bare string |
| `@import "https://cdn.example/x.css";` | proxied |
| `image-set(var(--c, "https://cdn.example/a.png") 1x)` | fallback proxied where it sits |
| `--c:"https://cdn.example/a.png"` used as `image-set(var(--c) 1x)` | untouched — resolving one declaration against another is the cascade's job, not a rewriter's |
| `filter:url(#blur)` | untouched — a fragment is not a fetch |
| `url('/local/a.png")` | untouched; the tokenizer reports a bad string, as a browser would |
| `url("https://cdn.example/a` + newline + `b)` | untouched; a newline has already ended the string |
| `srcset="data:...;base64,,,,, 1x, /b.png 2x"` | unchanged grouping — pinned by test |
| CSS nested past `MAX_CSS_NESTING_DEPTH` | stylesheet discarded; the CSS proxy path answers with a status rather than an empty `200` |

## Earlier verification against production CSS

Ran `rewrite_css_body` from `main` and from this branch over an identical corpus of **51 real stylesheets (1.75 MB)**: 20 captured from the configured publisher origin, plus 31 widely-used public stylesheets from `fonts.googleapis.com`, `cdnjs.cloudflare.com` and `cdn.jsdelivr.net`. The public ones are there because the publisher's own bundles carry almost no `url()` at all, so a corpus limited to them cannot exercise the rewrite at all — which is what the earlier, smaller run of this check ended up showing.

| Result | Count |
| ------ | ----- |
| Byte-identical output, `main` vs branch | 46 / 51 |
| Differing output | 5 / 51 |
| Differences that are anything other than quoting | 0 |
| Files where the set of proxied targets differs | 0 / 51 |
| Files rejected by the depth bound | 0 / 51 |

The 5 differing files differ only in that this branch emits `url("…")` where `main` emitted `url(…)`; the value inside is identical, and quoting is the documented normalization. Across the 6 files that contain absolute references, both versions proxy the same **183 targets** — none newly rewritten, none newly missed. That includes a real `@import url("https://…")` and 165 unquoted absolute `url()` references.

Corpus coverage, for what this does and does not attest to: 8,258 `var(`, 165 unquoted absolute `url(`, 63 quoted `url("`, 2 `url('`, 2 protocol-relative `url(//`, 1 `@import`. No stylesheet in the sample used `image-set()` or `src()`, so those paths rest on unit tests rather than on this differential.

## Closes

Closes #1114

## Test plan

- [x] `cargo test-fastly`, `cargo test-axum`, `cargo test-cloudflare`, `cargo test-spin`
- [x] JS tests
- [x] JS format
- [x] Docs format
- [x] Cross-adapter parity suite: 13 passed
- [x] Other: differential run of `main` and branch over a 51-file production CSS corpus (above)

## Checklist

- [x] New code has tests


prk-Jr:
Closes #1114

crates/trusted-server-core/src/creative.rs
```diff
@@ -73,64 +73,394 @@ pub(super) fn to_abs(settings: &Settings, u: &str) -> Option<String> {
...
+/// the one measured. It is still far above any real stylesheet, where nesting
+/// is a declaration list inside a handful of blocks. Entering this many
+/// scopes costs one more stack frame than the bound, since the outermost walk
+/// has entered none.
+///
+/// An iterative walk would need no bound, but the tokenizer only exposes a
+/// block through a closure, so each nested slice would be re-read from its
+/// start — quadratic in the input, which [`MAX_REWRITABLE_BODY_SIZE`] allows
+/// to be 10 MB. Recursing stays linear and bounds the stack instead.
+const MAX_CSS_NESTING_DEPTH: usize = 64;
```
aram356, line 102: 🔧 **wrench** — The depth cap admits a different nesting level depending on whether the URL is quoted, and the difference silently discards the whole stylesheet.

`Token::UnquotedUrl` is a terminal token — it never recurses, so `url(https://…)` is rewritten at any depth. But `url("https://…")` arrives as `Token::Function("url")`, takes the `CssToken::UrlFunction` arm, hits the `depth >= MAX_CSS_NESTING_DEPTH` guard, and sets `depth_exceeded`, which discards the entire stylesheet. `image-set()` behaves the same way.

Reproduced at exactly `MAX_CSS_NESTING_DEPTH` (64) nested `a{` blocks:

| input at depth 64 | result |
| --- | --- |
| `url(https://cdn.example/a.png)` | rewritten; stylesheet intact (459 bytes out) |
| `url("https://cdn.example/a.png")` | whole stylesheet discarded (287 bytes in, 0 bytes out) |
| `image-set("https://cdn.example/a.png" 1x)` | whole stylesheet discarded |

So the constant advertises 64 for one spelling of a URL and effectively 63 for another spelling of the same URL. A stylesheet is kept or dropped based on quoting, which is not a distinction the doc comment claims to make.

This passes CI because `rewrite_style_urls_rewrites_at_the_deepest_supported_nesting` (line 2428) pins only the unquoted form. Swapping that test's `url(https://cdn.example/a.png)` for the quoted `url("https://cdn.example/a.png")` fails on this branch, which is the cheapest way to confirm the above before changing anything.

The fix I would suggest is hoisting the single depth check above the `match found` so every recursing arm (`UrlFunction`, `ImportPrelude`, `Block`) shares one bound, rather than three separate guards that the terminal `Url` arm bypasses. Roughly:

```rust
// One bound, checked before any arm that recurses.
if matches!(
    found,
    CssToken::UrlFunction(_) | CssToken::ImportPrelude | CssToken::Block(..)
) && depth >= MAX_CSS_NESTING_DEPTH
{
    self.depth_exceeded = true;
    return;
}

match found {
    // … arms with their individual depth checks removed …
}
```

That still leaves `url(unquoted)` rewritable one level deeper than `url("quoted")`, since only one of them opens a scope — so whichever shape you choose, it is worth extending the depth test to cover the quoted and `image-set()` forms so the bound the constant advertises is the bound every spelling actually gets.

Apply manually — can't be auto-applied as a `suggestion` because the change spans three arms of the `match` plus a companion test edit outside this hunk.

crates/trusted-server-core/src/creative.rs
```diff
@@ -73,64 +73,394 @@ pub(super) fn to_abs(settings: &Settings, u: &str) -> Option<String> {
...
+        allows_import_rules,
+        out: String::with_capacity(style.len() + 16),
+        write_pos: 0,
+        depth_exceeded: false,
+    };
+    let mut input = cssparser::ParserInput::new(style);
+    let mut parser = cssparser::Parser::new(&mut input);
+    rewriter.walk(&mut parser, 0, BareStringUrls::Never);
+    if rewriter.depth_exceeded {
+        log::warn!("Rejecting a stylesheet nested past the supported depth");
```
aram356: ❓ **question** — Is a zero-byte 200 the right thing to put on the wire when the depth cap trips?

To be clear on the part I agree with: fail-closed over leaking is the right call, and the reasoning in this doc comment for why partial rewriting would defeat the purpose is sound. My question is only about what the client ends up seeing.

Traced through `proxy.rs:666`, a proxied `text/css` response that trips the cap is returned as **HTTP 200 with an empty body** — indistinguishable from a stylesheet the origin legitimately served as empty. Confirmed by driving `CreativeCssProcessor` directly: 1026 bytes in, 0 bytes out, no error.

The `log::warn!` on the next line is the only signal, and it is edge-side. A publisher seeing an unstyled creative has nothing in the response that distinguishes "we rejected this stylesheet" from "upstream sent nothing", which makes this a hard failure to attribute in production.

Would an error status — letting `finalize_proxied_response` decide, the way the oversized-body path already does — be preferable here? Or, if 200-with-empty-body is deliberate, a distinguishing response header would at least make the rejection attributable from a HAR capture.

Not arguing this is likely to fire: I verified that realistic nesting (`@media` + `@supports` + `@layer` + `@container` + `image-set`) sits nowhere near 64, and that 500 sibling blocks are fine. This is about diagnosability when it does.

aram356 (changes requested):
## Summary

This is a larger change than the description suggests: commits from `a8d4e099e` onward replace the hand-rolled CSS scanner with a `cssparser`-based grammar walk, adding `url()`/`src()`, `image-set()` bare-string candidates, `@import` preludes, and `var()`/`env()` fallback following. The token-level, property-agnostic design is the right shape for this problem — it covers URL-bearing properties nobody had to enumerate, while correctly leaving fragment-only references like `filter:url(#blur)` alone.

Two findings below. The depth cap admits a different nesting level depending on whether a URL is quoted, and the depth rejection is operationally invisible on the wire.

Both are prose-only: the wrench fix restructures a `match` across three arms and needs a companion test change, so it can't be expressed as a single contiguous `suggestion`.

Verified locally against the PR head: `cargo fmt --all -- --check`, `clippy-fastly`, `test-fastly` (2301 passed), `test-axum`, `check-cloudflare`, `check-spin`, and the cross-adapter parity suite (13 passed) all pass.

Three things I checked and am explicitly **not** raising, since each looked like a finding and turned out not to be:

- `@import layer(base) "url.css"` looked like a missed egress leak, but the CSS grammar requires the URL first — that prefix form is invalid and browsers do not fetch it.
- `&` becoming `&amp;` in rewritten `<style>` output is real, but `git show origin/main` confirms the handler is byte-identical to `main`. Pre-existing, not this PR.
- The `cssparser` dependency is free: `cssparser 0.36.0` is already in `main`'s lockfile via `lol_html` (`Cargo.lock:2911`). No new packages, and deleting the hand-rolled scanner makes the shipped WASM code sections roughly 19.7 KB smaller.

I also probed parse-error recovery specifically (bad-url tokens, unterminated strings, stray `)`, unmatched `}`, embedded NUL, CDO/CDC) and could not construct an input where a URL after the error point escapes rewriting.

## Blocking

### 🔧 wrench

- **Depth cap rejects at a different depth depending on whether the URL is quoted** — see inline at `crates/trusted-server-core/src/creative.rs:93`

### ❓ question

- **Blanking the whole stylesheet is invisible to the client** — see inline at `crates/trusted-server-core/src/creative.rs:159`

## Non-blocking

### 📝 note

- **PR description no longer matches the change** — see the Cross-cutting section below

- 📝 **PR description no longer matches the change** — The Summary and Changes tables describe a `css_string_end` helper and positional quote-scanning, which commit `a8d4e099e` deleted. The "Verification against production CSS" section describes a differential run over a parser version that is no longer what ships, so that evidence no longer covers the code under review. Worth refreshing before merge so the squashed commit message describes what actually landed — and worth re-running the production-CSS differential against the grammar walk, since that check is genuinely valuable and currently attests to superseded code.

## CI Status

At the time of review, checks on `441c831` were still running (the head is a fresh merge commit). No failures observed; the findings above are independent of CI.

- CodeQL: SKIPPED
- Analyze (actions): PASS
- Analyze (rust): PENDING
- Analyze (javascript-typescript): PENDING
- cargo test: PENDING (required)
- cargo test (ts CLI, native): PENDING
- cargo test (cross-adapter parity): PENDING
- cargo check (cloudflare native + wasm32-unknown-unknown): PENDING
- cargo check/build/test (spin native + wasm32-wasip1): PENDING
- cargo fmt: PENDING (required)
- format-typescript: PENDING (required)
- format-docs: PENDING (required)
- vitest: PENDING

crates/trusted-server-core/src/creative.rs
prk-Jr: Fixed in 60b5e992.

Confirmed the asymmetry before changing anything, the way you suggested: swapping the deepest-nesting test to the quoted form fails on the old code, with an empty output.

One correction to the diagnosis, since it changed what needed touching. CI was not passing only because `rewrite_style_urls_rewrites_at_the_deepest_supported_nesting` pinned the unquoted form. A second test, `rewrite_style_urls_rejects_a_quoted_url_past_the_supported_depth`, pinned the quoted rejection explicitly — asserting "should count the quoted url() parser frame against the depth bound". So the asymmetry was deliberate rather than unnoticed, and the fix had to replace that test'''s premise, not just add coverage for a gap.

I went past the hoist, because the hoist keeps the cliff you flagged at the end. Every recursion the walk makes now runs through one `descend()`, and `url()`/`src()` is no longer charged a level when its argument is a single string. That grammar is terminal — it never recurses into `walk` — so charging it bought no stack safety and cost only the quoting-dependent cliff. `src(var(…))` does recurse, so it still goes through `descend`.

`image-set()` still costs a level, and that is now pinned deliberately rather than incidentally: it holds a value list the walk has to read, so it is a scope. The new test asserts it is admitted at `MAX_CSS_NESTING_DEPTH - 1` and refused at the cap, so the boundary is a property of the grammar rather than of how a URL is spelled — which is what the constant now documents.

Tests: `rewrite_style_urls_rewrites_at_the_deepest_supported_nesting` loops `url(x)`, `url("x")`, `url('''x''')` and `src("x")` at exactly the cap. The old quoted-rejection test is replaced by the `image-set()` boundary test.

crates/trusted-server-core/src/creative.rs
prk-Jr: Took the first option — an error status, matching the oversized-body precedent. Fixed in 60b5e992.

`rewrite_css_body` now returns `Result<String, CssRewriteError>`, and `CreativeCssProcessor::process_chunk` maps that to an `io::Error`, so a refusal lands in the same `TrustedServerError::Proxy` path `MAX_REWRITABLE_BODY_SIZE` already takes and `finalize_proxied_response` decides the status. Pinned by `css_processor_reports_a_stylesheet_nested_past_the_supported_depth`, which asserts the processor reports rather than serving an empty body.

I left the markup path dropping rather than erroring, and documented why on each entry point: a `<style>` block is one part of a document whose rest is still rewritten, so failing the whole page over one block trades a styling loss for a page loss. Your diagnosability point applies there too, though, so the warn now names what it dropped — `<style> block` or `style attribute` — instead of one generic line covering both.

prk-Jr:
Both non-blocking items addressed.

**PR description** — rewritten. It described the `css_string_end` helper and positional quote scanning, both deleted in a8d4e099. It now describes the grammar walk, the coverage it adds (`src()`, `image-set()`, `@import`, `var()`/`env()` fallbacks), the shape-preserving re-emit, the depth bound, and the two fixes from this review round.

**Production-CSS differential** — re-run against the grammar walk, and widened, because the old corpus could not have caught anything. The publisher'''s own Next.js bundles carry 2 `url()` references across 20 files and 0 absolute ones, so a corpus limited to them exercises no rewriting in either version. I kept those 20 and added 31 widely-used public stylesheets (`fonts.googleapis.com`, `cdnjs.cloudflare.com`, `cdn.jsdelivr.net`) that actually contain absolute references.

Method: two detached worktrees at `origin/main` and this branch, an identical dump harness in each calling `rewrite_css_body` over the same 51 files (1.75 MB), outputs diffed.

| Result | Count |
| ------ | ----- |
| Byte-identical output | 46 / 51 |
| Differing output | 5 / 51 |
| Differences that are anything other than quoting | 0 |
| Files where the set of proxied targets differs | 0 / 51 |
| Files rejected by the depth bound | 0 / 51 |

The 5 that differ differ only in that this branch emits `url("…")` where `main` emitted `url(…)` — the value inside is byte-identical, and re-quoting is the documented normalization. Across the 6 files carrying absolute references both versions proxy the same 183 targets: none newly rewritten, none newly missed. That set includes a real `@import url("https://…")` and 165 unquoted absolute `url()` references.

Stating the gap rather than leaving it implied: the corpus holds 8,258 `var(`, 165 unquoted absolute `url(`, 63 quoted `url("`, 2 `url('`, 2 protocol-relative `url(//` and 1 `@import`, but no `image-set()` and no `src()` anywhere. Those two paths rest on unit tests, not on this differential.

crates/trusted-server-core/src/creative.rs
```diff
@@ -73,64 +73,440 @@ pub(super) fn to_abs(settings: &Settings, u: &str) -> Option<String> {
...
+/// `src()` is not one — that grammar is terminal, so it costs no recursion and
+/// is not charged a level. Without that exemption the bound would depend on
+/// how a URL is spelled, admitting `url(https://…)` where it rejects the
+/// identical `url("https://…")`, and rejection discards the whole stylesheet.
+const MAX_CSS_NESTING_DEPTH: usize = 64;
+
+/// Rewrites URL references inside a CSS string to the first-party proxy.
+///
+/// Covers every form the browser fetches: `url()` and `src()`, a bare string
+/// candidate in `image-set()`, and an `@import` prelude string.
```
ChristianPavilonis, line 107: **[P2] Preserve signed query separators inside `<style>`**

When the newly covered forms reach the existing `<style>` handler at line 1263, `ContentType::Text` HTML-escapes the generated URL's `&tstoken=` as `&amp;tstoken=`. A `<style>` element is raw text, so the browser does not decode that entity before CSS parsing. It requests a URL with an `amp;tstoken` parameter, and `/first-party/proxy` rejects it for missing `tstoken`.

This PR makes the problem affect bare `image-set()` candidates, `src()`, and bare `@import` strings that previously remained direct. Please replace the style text without HTML body escaping and add a `rewrite_creative_html` test asserting that the result contains `&tstoken=` and not `&amp;tstoken=`.

crates/trusted-server-core/src/creative.rs
```diff
@@ -1014,7 +1435,7 @@ impl StreamProcessor for CreativeCssProcessor<'_> {
             let css = String::from_utf8(std::mem::take(&mut self.buffer))
                 .map_err(|e| io::Error::other(format!("Invalid UTF-8 in CSS: {e}")))?;
 
-            let rewritten = rewrite_css_body(self.settings, &css);
+            let rewritten = rewrite_css_body(self.settings, &css).map_err(io::Error::other)?;
```
ChristianPavilonis, line 1483: **[P2] Bound transformed CSS output**

The processor caps buffered input at 10 MiB, but `rewrite_css_body` can expand every short URL into a much longer signed proxy URL and return an output far above that limit. Repeated newly supported `image-set("https://...")` candidates make a permitted stylesheet a practical WASM memory-exhaustion input.

Checking the length after rewriting would still allow the large allocation. Please enforce the limit while building `CssUrlRewriter::out` and return an error once the output budget is exceeded. The same bound should cover CSS rewritten inside proxied HTML, where the temporary CSS string is allocated before the outer HTML output limit runs.

ChristianPavilonis (approved):
## Summary

Re-reviewed PR head `60b5e992c004969836febef3efd5340ceaa85d1c`.

The latest commit fixes the inconsistent nesting boundary and now reports rejected CSS responses through the proxy error path. I left two P2 findings inline.

Non-blocking documentation fix: `crates/trusted-server-core/README.md:37` still documents `rewrite_css_body` as returning `String`; it now returns `Result<String, CssRewriteError>`.

All checks on this revision pass.

aram356 (changes requested):
## Summary

Re-reviewed at `60b5e992`. Both findings from my previous pass are fixed, and I verified each by assertion rather than from the commit message: the depth bound now admits `url(x)`, `url("x")`, `url('x')` and `src("x")` at the same level, with `image-set()` still costing one because it genuinely opens a scope; and a refused stylesheet now reaches `finalize_proxied_response` as an error, which maps to a 502 instead of an empty 200. The `descend()` chokepoint is a better fix than the one I proposed, since it makes the bound structural rather than something each new arm has to remember.

Not approving, because of one regression in the paths this PR newly covers.

@ChristianPavilonis raised this as P2 and I want to add a reproduction, because the framing matters: I saw the `&amp;` escaping on my first pass and set it aside as pre-existing on `main`, which was the wrong call. It is pre-existing for `url()`, but `image-set()`, `src()` and bare `@import` strings were passed through untouched before this PR, so they loaded correctly. Now they are rewritten into URLs the proxy rejects. Those three forms regress from working to broken, and that is this PR's to fix.

## Blocking

### 🔧 wrench

- **Newly covered `<style>` forms emit `&amp;tstoken=` and fail at the proxy** — `crates/trusted-server-core/src/creative.rs:1263`, detailed below

- 🔧 **Newly covered `<style>` forms emit `&amp;tstoken=` and fail at the proxy** — `ContentType::Text` HTML-escapes the signed URL, and this PR makes that break forms which previously worked.

`ContentType::Text` entity-escapes `&`, but a `<style>` element is raw text — the HTML parser does not decode entities inside it, so the CSS parser sees `&amp;tstoken=` literally. The browser requests `…&amp;tstoken=…`, `/first-party/proxy` sees a parameter named `amp;tstoken` and no `tstoken`, and rejects the request.

Confirming @ChristianPavilonis's P2 with a reproduction, and correcting my own earlier read of it. I saw this escaping on my first pass and set it aside as pre-existing on `main`. That was right for `url()` — which was already rewritten there and already broken this way — but wrong as a reason to leave it. `image-set()`, `src()` and bare `@import` strings were **not** rewritten before this PR, so they reached the browser as direct third-party URLs and loaded. Now they are rewritten into URLs the proxy rejects, so they regress from working to broken:

| `<style>` content at `60b5e992` | emits |
| --- | --- |
| `image-set("https://cdn.example/a.png" 1x)` | `&amp;tstoken=` — asset fails to load |
| `src("https://cdn.example/f.woff2")` | `&amp;tstoken=` — font fails to load |
| `@import "https://cdn.example/x.css";` | `&amp;tstoken=` — stylesheet fails to load |

`ContentType::Html` inserts the replacement without entity-escaping, which is the correct content model for a raw-text element:

```rust
                text!("style", |t| {
                    let s = t.as_str();
                    let rewritten = rewrite_style_urls(settings, s, base_origin);
                    if rewritten != s {
                        t.replace(&rewritten, ContentType::Html);
                    }
                    Ok(())
                }),
```

Verified in a scratch worktree at this head: with that one-line change, all four forms above emit a raw `&tstoken=` and none emit `&amp;tstoken=`, and the full `cargo test-fastly` suite still passes (2302 tests, 0 failures).

On the injection question that `ContentType::Html` naturally raises — I checked, and it does not open one here. A `</style>` sequence inside a CSS string terminates the element identically before and after this change, because the HTML tokenizer ends raw text at that literal regardless of how the replacement was inserted; I confirmed the escaping form is byte-identical in a case the rewriter leaves untouched. Guarding creative markup against that is `sanitize_creative_html`'s job, and it strips `<style>` outright.

Worth pairing with a `rewrite_creative_html` test asserting the output contains `&tstoken=` and not `&amp;tstoken=` — no current test covers the `<style>` path's query separators, which is why CI is green on this.


- 📝 **Output growth bound (@ChristianPavilonis's second P2) not assessed here** — I did not evaluate it, so nothing in this review should be read as clearing it. It is plausible on its face: every rewritten reference expands a short URL into a longer signed one, so a stylesheet that passes the 10 MiB input cap can exceed it after rewriting. Worth a reply from the author either way.

## CI Status

All checks pass on `60b5e992`. Locally against the same commit: `cargo fmt --all -- --check`, `clippy-fastly`, `test-fastly` (2302 passed), `test-axum`, `check-cloudflare`, `check-spin`, and the cross-adapter parity suite (13 passed) are all green. The regression below is not caught by any existing test, which is part of the finding.

- Analyze (actions): PASS
- Analyze (rust): PASS
- CodeQL: PASS
- cargo fmt: PASS (required)
- cargo test: PASS (required)
- integration tests: PASS
- vitest: PASS

crates/trusted-server-core/src/creative.rs
```diff
@@ -827,7 +1248,7 @@ fn rewrite_creative_html_impl(
                 // Inline style url(...)
                 element!("[style]", |el| {
                     if let Some(st) = el.get_attribute("style") {
-                        let rewritten = rewrite_style_urls(settings, &st, base_origin);
+                        let rewritten = rewrite_style_attribute_urls(settings, &st, base_origin);
```
aram356, line 1296: 🔧 **wrench** — Newly covered `<style>` forms emit `&amp;tstoken=` and are rejected by the proxy.

_(Anchored here because the line this is about — the `<style>` handler's `t.replace(&rewritten, ContentType::Text)` at `creative.rs:1263` — sits outside this diff's hunks and GitHub rejects a comment there with "line could not be resolved". This line is its sibling: the attribute handler that calls the same rewriter, and the nearest resolvable anchor.)_

`ContentType::Text` entity-escapes `&`, but a `<style>` element is raw text — the HTML parser does not decode entities inside it, so the CSS parser sees `&amp;tstoken=` literally. The browser requests `…&amp;tstoken=…`, `/first-party/proxy` sees a parameter named `amp;tstoken` and no `tstoken`, and rejects it.

Confirming @ChristianPavilonis's P2 with a reproduction, and correcting my own earlier read. I saw this escaping on my first pass and set it aside as pre-existing on `main`. That was right for `url()` — already rewritten there, already broken this way — but wrong as a reason to drop it. `image-set()`, `src()` and bare `@import` strings were **not** rewritten before this PR, so they reached the browser direct and loaded. Now they are rewritten into URLs the proxy rejects:

| `<style>` content at `60b5e992` | emits |
| --- | --- |
| `image-set("https://cdn.example/a.png" 1x)` | `&amp;tstoken=` — asset fails to load |
| `src("https://cdn.example/f.woff2")` | `&amp;tstoken=` — font fails to load |
| `@import "https://cdn.example/x.css";` | `&amp;tstoken=` — stylesheet fails to load |

Those three regress from working to broken, which is what makes this the PR's to fix rather than inherited.

`ContentType::Html` inserts the replacement without entity-escaping, the correct content model for a raw-text element. At `creative.rs:1263`:

```rust
                text!("style", |t| {
                    let s = t.as_str();
                    let rewritten = rewrite_style_urls(settings, s, base_origin);
                    if rewritten != s {
                        t.replace(&rewritten, ContentType::Html);
                    }
                    Ok(())
                }),
```

Verified in a scratch worktree at this head: with that one-line change all four forms emit a raw `&tstoken=`, none emit `&amp;tstoken=`, and `cargo test-fastly` still passes (2302 tests, 0 failures).

On the injection question `ContentType::Html` naturally raises — it does not open one here. A `</style>` inside a CSS string terminates the element identically before and after, because the HTML tokenizer ends raw text at that literal however the replacement was inserted; I confirmed the output is byte-identical in a case the rewriter leaves untouched. Guarding creative markup against that is `sanitize_creative_html`'s job, and it strips `<style>` outright.

Worth pairing with a `rewrite_creative_html` test asserting the output contains `&tstoken=` and not `&amp;tstoken=` — no current test covers this path's query separators, which is why CI is green on it.

Apply manually — can't be offered as a one-click `suggestion`, since the line it would replace is the unresolvable 1263 rather than this anchor.

crates/trusted-server-core/src/creative.rs
prk-Jr: Fixed in 60780997. The `<style>` text handler now uses `ContentType::Html`, preserving raw signed query separators.

`rewrites_style_block_url_variants` covers `url()`, `src()`, bare `image-set()` candidates, and bare `@import` strings through `rewrite_creative_html`. Each case asserts `&tstoken=` is present and `&amp;tstoken=` is absent. The regression failed before the fix and passed afterward; the full local CI gates also passed.

crates/trusted-server-core/src/creative.rs
prk-Jr: Fixed in 60780997. `CssUrlRewriter` now writes through a bounded `fmt::Write` implementation that checks each append before adding bytes, including CSS string serialization and unchanged prefixes/tails. Initial and subsequent requested capacity growth are capped at the 10 MiB budget.

Exceeding the limit produces `CssRewriteError::OutputTooLarge`. Whole CSS responses propagate the error through the processor; style blocks and attributes use the same bounded output and retain their existing drop-on-refusal behavior.

Tests cover URL expansion, exact-limit acceptance, rejection during serialization, unchanged trailing text, processor error propagation, and both inline CSS helpers. The expansion regression failed before the fix. All local CI gates passed.

crates/trusted-server-core/src/creative.rs
prk-Jr: Applied in 60780997: the `<style>` replacement uses `ContentType::Html`. The HTML-level regression covers all four forms (`url()`, `src()`, bare `image-set()`, and bare `@import`) and checks the raw `&tstoken=` separator and absence of `&amp;tstoken=`.

The regression reproduced the broken output before the change and passes with the fix. The full Fastly suite and all other required local gates passed.

crates/trusted-server-core/src/creative.rs
```diff
@@ -1531,19 +2827,151 @@ mod tests {
...
+            "should bound reserved capacity too"
+        );
+        assert!(
+            std::fmt::Write::write_str(&mut out, "").is_ok(),
+            "should permit an empty append at the limit"
+        );
+    }
+
+    #[test]
+    fn css_output_budget_includes_unchanged_trailing_text() {
```
aram356: ⛏ **nitpick** — These budget tests cost ~19s of CI time, and the cheaper technique is already in this commit.

Four of the new tests allocate 10 MiB strings (`" ".repeat(MAX_REWRITABLE_BODY_SIZE)`), several per test in some cases. Measured on this branch:

```
cargo test-fastly -p trusted-server-core css_output  →  19.34s   (4 tests)
```

The whole `trusted-server-core` suite went from roughly 3s to 78s. That lands on every CI run and every local `cargo test-fastly`, and under Viceroy those allocations are especially slow.

To be clear about scope: the tests are correct, and the coverage is worth keeping — this is only about how it is paid for.

The cheaper approach is already in this same commit. `css_output_checks_each_serialized_fragment_before_appending` (line 2847) constructs the type directly:

```rust
        let mut out = super::CssOutput {
            value: String::new(),
            limit: 8,
        };
```

Because `limit` is a plain field rather than hardwired to the constant, the same boundary — the last byte that fits, the first that does not, the escape-aware append — can be exercised at an 8-byte limit for effectively zero cost. `css_output_budget_includes_unchanged_trailing_text`, `inline_css_output_uses_the_same_budget` and `css_output_expansion_is_rejected_before_the_body_limit` all look convertible that way.

The one worth its cost as-is is `css_processor_reports_output_expansion_and_accepts_the_exact_limit`, since it drives `CreativeCssProcessor` end-to-end and the real constant is the thing under test there.

Non-blocking — correctness is fine either way.

crates/trusted-server-core/src/creative.rs
```diff
@@ -73,64 +74,473 @@ pub(super) fn to_abs(settings: &Settings, u: &str) -> Option<String> {
...
+    }
+    rewriter
+        .out
+        .write_str(&style[rewriter.write_pos..])
+        .map_err(|_| CssRewriteError::OutputTooLarge)?;
+    Ok(rewriter.out.value)
+}
+
+/// Checks each append before allocating output, including CSS string escapes.
+struct CssOutput {
```
aram356: 📝 **note** — Worth restoring a commit body for `60780997` before merge.

`CssOutput` is the most subtle thing in this commit, and the design decisions behind it are not self-evident from the code:

- the budget is enforced **during** construction rather than checked afterward, so the oversized allocation never happens
- `fmt::Write` is the interface specifically so `cssparser::serialize_string` is bounded too, escapes included
- a partial write is deliberately left in the buffer on failure, because `self.error` makes the whole rewrite return `Err` and the buffer is discarded — it is never served

The subject line `Bound CSS output and preserve signed style URLs` names both changes accurately, but `60780997` has no body, unlike the earlier commits on this branch. `a8d4e099` and `60b5e992` both carry explanations of *why*, and `60b5e992`'s body is what let me verify its claims quickly on the previous pass rather than reverse-engineering intent from the diff.

Since this branch will land as a merge, that reasoning is worth preserving in history for whoever next touches the write path.

Non-blocking, and purely about the commit message — the code is fine.

aram356 (approved):

crates/trusted-server-core/src/creative.rs
aram356, line 2882: ⛏ **nitpick** — These budget tests cost ~19s of CI time, and the cheaper technique is already in this commit.

Four of the new tests allocate 10 MiB strings (`" ".repeat(MAX_REWRITABLE_BODY_SIZE)`), several per test in some cases. Measured on this branch:

```
cargo test-fastly -p trusted-server-core css_output  →  19.34s   (4 tests)
```

The whole `trusted-server-core` suite went from roughly 3s to 78s. That lands on every CI run and every local `cargo test-fastly`, and under Viceroy those allocations are especially slow.

To be clear about scope: the tests are correct, and the coverage is worth keeping — this is only about how it is paid for.

The cheaper approach is already in this same commit. `css_output_checks_each_serialized_fragment_before_appending` (line 2847) constructs the type directly:

```rust
        let mut out = super::CssOutput {
            value: String::new(),
            limit: 8,
        };
```

Because `limit` is a plain field rather than hardwired to the constant, the same boundary — the last byte that fits, the first that does not, the escape-aware append — can be exercised at an 8-byte limit for effectively zero cost. `css_output_budget_includes_unchanged_trailing_text`, `inline_css_output_uses_the_same_budget` and `css_output_expansion_is_rejected_before_the_body_limit` all look convertible that way.

The one worth its cost as-is is `css_processor_reports_output_expansion_and_accepts_the_exact_limit`, since it drives `CreativeCssProcessor` end-to-end and the real constant is the thing under test there.

Non-blocking — correctness is fine either way.

crates/trusted-server-core/src/creative.rs
aram356, line 210: 📝 **note** — Worth restoring a commit body for `60780997` before merge.

`CssOutput` is the most subtle thing in this commit, and the design decisions behind it are not self-evident from the code:

- the budget is enforced **during** construction rather than checked afterward, so the oversized allocation never happens
- `fmt::Write` is the interface specifically so `cssparser::serialize_string` is bounded too, escapes included
- a partial write is deliberately left in the buffer on failure, because `self.error` makes the whole rewrite return `Err` and the buffer is discarded — it is never served

The subject line `Bound CSS output and preserve signed style URLs` names both changes accurately, but `60780997` has no body, unlike the earlier commits on this branch. `a8d4e099` and `60b5e992` both carry explanations of *why*, and `60b5e992`'s body is what let me verify its claims quickly on the previous pass rather than reverse-engineering intent from the diff.

Since this branch will land as a merge, that reasoning is worth preserving in history for whoever next touches the write path.

Non-blocking, and purely about the commit message — the code is fine.

aram356 (approved):

Issue #1081, opened 2026-08-27
Improvements to TS_CONSOLE for ad observability


jevansnyc:
### Description

TS Console (sometimes called "tracer") doesn't provide completely accurate wording and is missing important features such as linking a creative in the DOM to the diagnostics menu. We need to update ts_console to provide better creative observability. Updated requirements include: 

1. Assign numerical value to the creative on page to the side menu
2. Inject auction timing metrics as per PR #1075 and #1076 
3. Label auction as SSAT, TS auction or client side auction 
4. Inject winning bidder and bid price 
5. Remove the 1x1 or 1 language as its every creative and not helpful 
6. "Box" should be "size filled" 
7. Define what is "response" and "render" and normalize against #2 but probably better to use the new auction timing metrics

ChristianPavilonis:
@ChristianPavilonis to review and create dictionary

PR #1076, opened 2026-08-26
Add auction timeline offsets spec amendment

jevansnyc:
Spec-first follow-up to #1074, targeting the feature branch so it lands with (or after) the base spec rather than against main.

Adds **section 18** to the request phase timing design: three T0-anchored auction milestones so the auction's timeline and the request's timeline finally share a clock.

## Problem

Two clocks that never meet: `auction_events_raw` measures the auction internally (`total_time_ms`, per-provider `provider_response_time_ms`) on a clock that starts at auction creation; the access row is T0-anchored but only records `auction_wait_ms` (blocked time at collect). Nothing can answer: when did the auction start relative to request entry, when did the final bid land, and when was targeting committed toward GAM.

## Design

- Three first-call-wins marks on `RequestTimings` (same style as `mark_headers_ready()`): dispatched (bid requests left the edge), resolved (final bid or timeout), committed (`write_bids_to_state` returned; targeting available to the response pipeline in both buffered and streaming modes).
- Four additive columns on `access_logs_raw`: `auction_dispatched_ms` / `auction_resolved_ms` / `auction_committed_ms` (Nullable UInt32; null = no auction ran) plus `auction_id` (join key to the per-bidder auction dataset; `none` sentinel).
- No header emission, no config surface, no new emission path: values ride the existing snapshot and the existing `tinybird.access_enabled` gate. Additive schema evolution with JSONPaths + FORWARD_QUERY, checked with `tb --cloud deploy --check`.
- "Committed toward GAM" is defined as edge-side commit: TS never calls GAM server-side; the browser GPT call carries the targeting, and that half of the timeline stays client-measured.

## Why it matters

This is the overlap proof: a client-side wrapper cannot dispatch until the browser boots (t~3000ms on measured prospect pages); the server-side auction dispatches while the origin fetch is in flight. One access row then reads as a timeline (dispatch at t=D, resolve at t=R, commit at t=C, headers at t=H), with `R - D` joining per-bidder detail via `auction_id`, and `auction_wait_ms` finally interpretable next to it: `(R - D) - auction_wait_ms` approximates how much of the auction was absorbed by work the request needed anyway.

[...]

Issue #1050, opened 2026-08-20
Create debug endpoint for mobile user to trace ad rendering


aram356:
### Description

Create debug endpoint for mobile user to trace ad rendering

aram356:
@jevansnyc Please let us know if you want engineers to start the spec

aram356:
@prk-Jr to write spec

aram356:
@prk-Jr Some notes from @jevansnyc 

1. Network information (as example see https://www.fastly-debug.com)
2. Auction information per ad slot (similar to ts-tracer and metrics in telemetry PR)
3. End-user cookie information

PR #1107, opened 2026-09-01
Design mobile ad-rendering trace endpoint

### LINKED ISSUE #1108: SPEC Specify mobile ad-rendering trace endpoint
## Context

Issue #1050 requests a mobile-friendly debug endpoint for tracing ad rendering, but the endpoint contract, privacy boundary, auction evidence model, and user journey require design agreement before implementation begins.

## Scope

- define the `/_ts/trace` mobile activation, reproduction, report, and export journey
- define privacy-safe network and Trusted Server cookie-health fields
- define separate SSAT/Trusted Server auction, GPT delivery, and creative-rendering evidence
- define exact transport and opaque correlation contracts without a database or server-side report store
- define security, truncation, failure, test, rollout, and implementation-sequencing requirements

## Acceptance criteria

- [ ] Design spec is reviewed and approved
- [ ] Scope ownership between #1050, #1081, and #1076 is explicit
- [ ] Implementation can proceed in independently reviewable phases

This ticket covers design approval only. Implementation remains tracked by #1050 and follow-up implementation PRs.

prk-Jr:
## Summary

- establish a reviewable design boundary before implementing the mobile ad-rendering trace requested by #1050
- define a mobile-first reproduction and export journey that does not require credentials, developer tools, a target URL, or a server-side report database
- separate privacy-safe server-auction, GPT delivery, and creative-rendering evidence so the report does not claim correlations it cannot prove

## Changes

| File | Change |
| ---- | ------ |
| `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md` | Defines UX, routing, configuration, schemas, live auction transport, opaque slot correlation, privacy, failure handling, testing, rollout, acceptance criteria, and implementation sequencing. |

## Closes

Closes #1108

Implementation remains tracked by #1050. Related observability and timing follow-ups remain tracked by #1081 and #1076.

## Test plan

- [ ] `cargo test-fastly && cargo test-axum` — not run; documentation-only change
- [ ] `cargo clippy-fastly && cargo clippy-axum` — not required for the documentation-only change
- [x] JS tests: `cd crates/trusted-server-js/lib && npx vitest run` — 45 files and 893 tests passed under pinned Node 24.12.0
- [ ] WASM build: `cargo build --package trusted-server-adapter-fastly --release --target wasm32-wasip1` — not run; documentation-only change
- [ ] Manual testing via `fastly compute serve` — not applicable
- [x] Other: independent adversarial design review approved; `git diff --check main...HEAD` passed

## Checklist

- [x] No `unwrap()` in production code — no production code changed
- [x] Uses `tracing` macros (not `println!`) — no logging code changed
- [x] New code has tests — no code added; the spec defines the required implementation test strategy


docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1667 @@
...
+The model has deliberately lower cardinality and sensitivity than the existing
+telemetry and OpenRTB objects:
+
+- `diagnostic_auction_id` is a fresh opaque `ts-auc-...` correlation token. When
+  trace capture is active under the two-part gate in section 5.2, it is minted
+  once when an eligible auction is observed, before dispatch, and is retained
+  for zero-bid, skipped, dispatch-failed, execution-failed, and abandoned
+  outcomes. It is never `AuctionRequest.id`, the telemetry UUID, a provider
+  request ID, or an identifier joinable to user-bearing logs.
+- Auction and slot tokens are the fixed prefixes `ts-auc-` and `ts-slot-`
```
jevansnyc: **Token shape contradicts the existing producer.**

This mandates `ts-auc-` followed by a canonical lowercase hyphenated UUID v4, and says validators reject every other shape. The producer in `crates/trusted-server-core/src/publisher.rs:5286` emits `Uuid::new_v4().simple()`, which is unhyphenated.

That value is the same GPT opportunity marker line 936 says the token attaches to, so the correlation join would never match and every report would render `Unknown`.

Either relax the validator to accept both forms, or change the producer and say so explicitly here.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1667 @@
...
+
+`TraceAuctionEvidenceV1` is a server-produced public model defined in section
+9.4. `TraceSlotCorrelationV1` is the browser-produced exact-token sidecar
+defined in section 9.4.1. `TraceGptDiagnosticsV1` is a separate trace-owned
+projection sourced only from `GptDiagnosticsExportV1`. It contains:
+
+- `schema_version: 1` and `source_schema_version: 1`;
+- the source `capturedAt` value;
+- `page.origin` after validation and `page.pathname` fixed to `/[redacted]`;
+- field-for-field allowlisted copies of the current v1 slots, requests,
```
jevansnyc: **This allowlist and acceptance criterion 8 cannot both hold.**

`TraceGptDiagnosticsV1` is defined as a field-for-field copy of the current v1 request cycle. That cycle carries `adManager.creativeId`, `lineItemId`, `campaignId`, and `advertiserId` (`core/types.ts:149`). Section 12.1 and acceptance criterion 8 forbid creative identifiers in the export.

As written this is unsatisfiable. Please name the excluded fields here rather than leaving "field-for-field" to be reconciled against section 12 at implementation time.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1667 @@
...
+
+Every adapter implements the following order:
+
+1. Parse the method, canonical host/origin, path, query, and bounded headers
+   required for route safety.
+2. Classify an exact Trusted Server reserved path.
+3. For a trace path, terminate locally after only trace-specific validation and
+   bounded request-context inspection, including an optional read-only platform
+   geo lookup used solely for the displayed setup request.
+4. For all other paths, continue through the adapter's ordinary event context,
```
jevansnyc, line 530: **Trace paths terminate before authentication.**

Step 3 terminates trace paths locally, and step 4 is where authentication happens for everything else. That means `/_ts/trace*` bypasses an operator's Basic Auth handler.

`crates/trusted-server-core/src/auth.rs:498-523` documents that a broad `^/_ts` handler is intended to challenge that whole namespace, and that the project deliberately refuses to carve out exemptions. This design carves one out by ordering, and section 12 does not mention it.

This needs to be an explicit decision in the security section, not a side effect of pipeline placement.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1667 @@
...
+toggling diagnostics through a top-level GET while preserving a one-URL,
+one-tap mobile workflow. This control does not defend against code already
+executing on the publisher origin.
+
+### 5.2 Reuse the existing diagnostics session
+
+The endpoint reuses `__Host-ts-console` and the existing GPT diagnostics
+activation semantics rather than creating a second `ts-trace` session. The
+cookie remains host-only, `Secure`, `HttpOnly`, and `SameSite=Lax`.
+`POST /_ts/trace/enable` sets it with a fixed 30-minute `Max-Age` and does not
```
jevansnyc: **The 30-minute `Max-Age` does not actually bound the session.**

The design leans on this lifetime to bound accidental private/no-store operation. But it is on the same `__Host-ts-console` cookie name that the existing `?ts_console=1` flow sets with no `Max-Age` (`crates/trusted-server-core/src/gpt_diagnostics.rs:30`).

Any later query activation overwrites the bounded cookie with an unbounded session cookie, so the guarantee this paragraph claims does not survive normal use of the existing flow. Worth either reconciling the two writers or using a distinct cookie name.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1667 @@
...
+  request ID, or an identifier joinable to user-bearing logs.
+- Auction and slot tokens are the fixed prefixes `ts-auc-` and `ts-slot-`
+  followed by a canonical lowercase hyphenated UUID v4. Validators reject every
+  other shape; tokens are not silently shortened or normalized.
+- `slot_number` is a one-based ordinal over the exact post-conversion
+  `AuctionRequest.slots` sequence observed by orchestration. It is display-only
+  and is never used to map a response back to pre-conversion client input.
+  `slot_ref` is a fresh auction-local opaque token carried with that slot. Core
+  creates it for initial-navigation and SPA auctions. For a TSJS `/auction`
+  request, TSJS creates it only after `buildAdRequest` has finished grouping and
```
jevansnyc: **No stated gate on the TSJS side.**

TSJS is told to mint `trace_slot_ref` on every outgoing `/auction` unit, with no condition attached. The browser can only observe `window.__tsjs_gpt_diagnostics_active`; it cannot see `trace_page_enabled`.

That contradicts line 1471, which requires that disabling `trace_page_enabled` removes all capture behavior even when a `?ts_console=1` cookie is live. Please state what TSJS checks before minting.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1667 @@
...
+`TraceGptDiagnosticsV1`, with the GPT projection sourced from
+`GptDiagnosticsExportV1`. The viewer rejects every unknown outer,
+server-auction, correlation, GPT-projection, or GPT-source version with an
+actionable message rather than guessing. A future TS Console successor requires
+an additive source compatibility change and, if the public projection changes,
+a new trace-envelope version.
+
+`auction_coverage.capture_status` describes only what reached the browser
+collector: `not_observed` means no valid server-auction record arrived, not that
+no server auction ran. `partial` requires at least one retained record plus a
```
jevansnyc: **`capture_status` has no value for one reachable case.**

Consider: records arrived, then size truncation removed all of them. `partial` requires at least one retained record. `unavailable` requires a projection, transport, or validation issue, and `record_evicted` is not in that list. `complete` requires a retained record.

An implementer lands on `not_observed`, which line 683 defines as meaning no valid record arrived. That is a false statement in this case. Either add `record_evicted` to the `unavailable` issue list or add a status for it.

jevansnyc (commented):
Review of the design spec. No code changes here, so this is internal consistency plus whether the stated contracts hold against what is in the repo today (`gpt_diagnostics.rs`, `publisher.rs`, `auth.rs`, `prebid_eids.rs`, and the TS diagnostics store/types).

Seven issues inline. The first three change the design rather than the wording:

1. The `ts-auc-` token shape does not match the existing producer, so correlation never joins.
2. `TraceGptDiagnosticsV1` as specified cannot satisfy acceptance criterion 8.
3. Trace paths terminate ahead of authentication, which carves an exemption out of the `^/_ts` namespace that `auth.rs` says should not exist.

The remaining four are bounded-scope corrections to the cookie lifetime claim, the TSJS gate, a `capture_status` gap, and redaction consistency.

Mechanical checks came back clean: cookie names (`ts-ec`, `ts-eids`, `ts-tester`, `__Host-ts-console`), the 8 KiB `ts-eids` cap (`MAX_EIDS_COOKIE_BYTES`), and the callback-issue `reason` values all match what the spec assumes.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1667 @@
...
+| Slots                                         | 64                                                           |
+| Request cycles                                | 10 per slot before total-size truncation                     |
+| Callback issues                               | 128                                                          |
+| Attribution issues                            | 128                                                          |
+| Requested slot sizes                          | 16 per cycle                                                 |
+| Ad Manager yield-group or company IDs         | 8 of each per cycle                                          |
+| Creative-failure enums                        | 16 per cycle                                                 |
+| Origin                                        | 255 UTF-8 bytes                                              |
+| GPT pathname in trace projection              | Exact literal `/[redacted]`                                  |
+| Slot element ID and ad-unit path              | 512 UTF-8 bytes each                                         |
```
jevansnyc: **Redaction is inconsistent across path-bearing fields.**

`page.pathname` is pinned to `/[redacted]` on the grounds that paths can carry secrets, but `slotElementId` and `adUnitPath` are retained at 512 bytes each. In practice both are usually templated from the page path.

So criterion 8's "no exact page paths" holds for the `page` object only. Either apply the same reasoning to these two fields or narrow the criterion to say what it actually guarantees.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [e3f371f8c](https://github.com/IABTechLab/trusted-server/commit/e3f371f8cb102d4b7f713d375f8587a34c6f01fd). Section 9.4 now specifies the existing producer format: `ts-auc-` plus a lowercase, unhyphenated UUID v4. New slot tokens retain the hyphenated `crypto.randomUUID()` format. Validators check version/variant and compare tokens verbatim; the auction producer does not need a format change. The verification requirements now include producer-to-sidecar joins.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [e3f371f8c](https://github.com/IABTechLab/trusted-server/commit/e3f371f8cb102d4b7f713d375f8587a34c6f01fd). Section 9.3 explicitly excludes the entire `adManager` object (all eight identity fields) and `previousCreativeId`. Derived `responseClass` and `creativeChanged` remain eligible. The builder omits these identifiers and the viewer rejects stored reports containing them; the obsolete identity-array bounds are removed. Sentinel-based projection/export checks are specified.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [e3f371f8c](https://github.com/IABTechLab/trusted-server/commit/e3f371f8cb102d4b7f713d375f8587a34c6f01fd). Sections 8 and 12.5 now require configured Basic Authentication before trace responses or mutations, including assets, disabled routes, and unsupported methods. Protected assets and challenges are private/no-store. The credential-free journey is explicitly conditional on operator rules leaving trace routes public; broad namespace rules remain enforced.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [e3f371f8c](https://github.com/IABTechLab/trusted-server/commit/e3f371f8cb102d4b7f713d375f8587a34c6f01fd). Section 5.2 now requires both endpoint and query activation writers to share `Max-Age=1800`, including when the trace page is disabled. Explicit activation restarts the lifetime; ordinary requests do not refresh it. The spec distinguishes browser cookie persistence from a hard capture deadline and documents legacy session cookies and already-loaded pages. Tests cover activation through both writers in either order.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [e3f371f8c](https://github.com/IABTechLab/trusted-server/commit/e3f371f8cb102d4b7f713d375f8587a34c6f01fd). Section 5.2 now defines the server-provided boolean `window.__tsjs_trace_active`. Both TSJS callers require `=== true` before minting tokens or creating trace collection state; the existing GPT diagnostics flag alone is insufficient. Document activation also honors the effective diagnostics decision, including query disable and navigation eligibility. Server API requests independently recheck configuration and cookie state. Rollback wording now distinguishes new requests from already-loaded documents.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [e3f371f8c](https://github.com/IABTechLab/trusted-server/commit/e3f371f8cb102d4b7f713d375f8587a34c6f01fd). Section 9.3 now includes eviction in the zero-retained-record `unavailable` case. Coverage is recomputed after memory eviction and snapshot truncation: losing every record produces `unavailable` with `record_evicted`, while retaining records produces `partial`. The unconditional eviction-to-partial wording is corrected, and an all-record eviction fixture is required. Missing optional transport is also distinguished from supplied invalid evidence.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [e3f371f8c](https://github.com/IABTechLab/trusted-server/commit/e3f371f8cb102d4b7f713d375f8587a34c6f01fd). Sections 9.1, 9.3, and 12.1 now exclude slot `slotElementId` and `adUnitPath`, plus `slotElementId` in callback and attribution issues. These fields are omitted rather than hashed or truncated, and the viewer rejects stored copies containing them. Runtime slot/request numbers preserve grouping and supported exact-token joins. Tests cover path-bearing sentinel values across HTML, storage, copy/share, and direct export.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1852 @@
...
+  cookie using `X-TS-Trace-Action: end`, and returns a small local JSON result.
+  After explicit user confirmation, client JavaScript independently attempts
+  local report deletion and the end POST. Neither result gates the other.
+- For both POST paths, absent or exactly-zero `Content-Length` is accepted,
+  `Transfer-Encoding` is rejected, and the adapter reads at most one byte when
+  it must verify an absent length. Any body byte or positive/invalid length
+  returns local `413 Payload Too Large` without draining or processing an
+  unbounded body. The one-byte read inherits a maximum two-second adapter
+  request-body deadline; timeout returns local `408 Request Timeout` with no
+  mutation.
```
aram356: 🔧 **wrench** — The two-second request-body deadline and the one-byte read are not implementable on the pinned adapter stack.

The body is already fully buffered before any Trusted Server code runs:

- `edgezero-adapter-fastly/src/request.rs:455-457` (tag `v0.0.8`) does an unbounded, blocking `body.read_to_end(&mut bytes)` with no deadline and no size cap.
- `edgezero-adapter-cloudflare/src/request.rs:263` does `req.bytes().await` — whole-body buffering, no deadline.
- `edgezero_core::body::Body` exposes no timeout or deadline method at all. Its only bounded primitive is `into_bytes_bounded(max_size)`, which is a size cap, not a deadline.
- Fastly's entry point runs under `futures::executor::block_on` (`crates/trusted-server-adapter-fastly/src/main.rs:181`), so there is no async timer to hang a deadline on in the first place.

Consequences: “reads at most one byte” is not expressible (the bytes are already read), the `408 Request Timeout` can never be produced, and the “without draining or processing an unbounded body” anti-DoS property is already violated upstream before this code executes. The `413` itself is achievable after the fact by inspecting `Content-Length` and the buffered length.

The house pattern for size-limited POST bodies is `crates/trusted-server-core/src/auction/endpoints.rs:128-159` (256 KiB, `Content-Length` precheck then post-buffer length check) and `crates/trusted-server-core/src/ec/batch_sync.rs:174-180` (2 MiB).

Section 14.1 also requires a test for “the two-second deadline without an unbounded read” (line 1458) that cannot be written against this stack; that line should go too.

```suggestion
- For both POST paths, absent or exactly-zero `Content-Length` is accepted,
  `Transfer-Encoding` is rejected, and the adapter reads at most one byte when
  it must verify an absent length. Any body byte or positive/invalid length
  returns local `413 Payload Too Large` without draining or processing an
  unbounded body. Where a runtime exposes a request-body read deadline, the
  one-byte read uses a maximum of two seconds and a timeout returns local
  `408 Request Timeout` with no mutation. `edgezero_core::body::Body` exposes
  no such deadline today, so any adapter that cannot express one must instead
  bound the read structurally and document the gap rather than claim a
  timeout it does not enforce.
```

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1852 @@
...
+  pseudonymous network data, and that opaque server-auction outcomes become
+  visible to same-origin JavaScript. Enabling the option is the deployment's
+  explicit acceptance of those bounded disclosures.
+- `GET /_ts/trace` returns the setup/report shell without changing cookies or
+  browser storage at the HTTP layer. After load, the explicitly included viewer
+  script may remove a rejected or expired local entry. Unrelated query
+  parameters do not activate or deactivate tracing and are not reflected into
+  the page or export.
+- `HEAD /_ts/trace` returns the GET status and headers without a body or state
+  mutation.
```
aram356: 🔧 **wrench** — Without explicit per-adapter `HEAD` registration, `HEAD /_ts/trace` is proxied to the publisher origin.

`HEAD` is a member of `publisher_fallback_methods()` (`crates/trusted-server-adapter-fastly/src/app.rs:313`), and the route registration loop maps every non-primary fallback verb to the publisher fallback handler (`app.rs:1322-1326`). So a `HEAD` on a GET-only named route does not return a local 405 — it goes upstream. The behavior is pinned by the regression test `dispatch_head_on_named_get_route_falls_through_to_publisher_fallback` (`app.rs:2557`).

That directly contradicts section 8 (“None falls through to the publisher origin”, line 512) and section 13 (“Unsupported method: local `405`; never publisher fallback”, line 1380) — and it is the same origin-leak class the rest of section 8 is written to prevent.

```suggestion
- `HEAD /_ts/trace` returns the GET status and headers without a body or state
  mutation. This requires registering `HEAD` explicitly on every trace path in
  every adapter: `HEAD` is a member of `publisher_fallback_methods()`
  (`crates/trusted-server-adapter-fastly/src/app.rs:313`), and the registration
  loop maps each non-primary fallback verb to the publisher fallback
  (`app.rs:1322-1326`), so an unregistered `HEAD` on a GET-only route is
  proxied to the publisher origin rather than answered locally. The guard test
  `dispatch_head_on_named_get_route_falls_through_to_publisher_fallback`
  (`app.rs:2557`) pins that behavior.
```

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1852 @@
...
+  malformed, cross-site, or duplicate control values return local `403` without
+  mutation. This deliberately targets current supported mobile browsers rather
+  than weakening the check for legacy clients.
+- The canonical request origin comes from adapter-owned inbound URL/scheme and
+  validated authority data, never an arbitrary forwarded header. Both it and
+  the single parsed `Origin` header are serialized with lowercase host and
+  default ports removed before exact comparison. Invalid or multi-valued host,
+  authority, scheme, or origin input fails closed.
+- Unsupported methods on a shell or state-changing path return a local 405
+  Method Not Allowed response with the path-specific `Allow` header.
```
aram356: 🔧 **wrench** — A router-level 405 supplies neither the `Allow` header nor any of the section 12.3 hardening headers.

Two separate gaps:

1. **No `Allow` header.** `EdgeError::MethodNotAllowed` folds the allowed verbs into a JSON body message, and its `IntoResponse` sets no `Allow` header. The only `Allow`-emitting 405 in this repo is hand-rolled in `deny_admin_diagnostic_fallback` (`crates/trusted-server-core/src/ec/admin.rs:179-196`), and it hardcodes `GET`.
2. **No hardening headers.** A router-level 405 is returned *before* the middleware chain runs, so it never reaches `FinalizeResponseMiddleware` and receives no CSP, `X-Content-Type-Options`, `Referrer-Policy`, or cache-control treatment. The test `dispatch_unregistered_method_returns_405_at_router_level` (`crates/trusted-server-adapter-fastly/src/app.rs:2599-2626`) asserts exactly this, with the comment “router-level 405 bypasses FinalizeResponseMiddleware”.

Since section 12.3 requires those headers on every trace response, trace routes must emit the 405, its path-specific `Allow`, and the hardening headers themselves rather than inheriting a router 405.

```suggestion
- Unsupported methods on a shell or state-changing path return a local 405
  Method Not Allowed response with the path-specific `Allow` header. The shared
  router does not supply either property: its `MethodNotAllowed` response
  carries no `Allow` header, and a router-level 405 is returned before
  `FinalizeResponseMiddleware` runs, so it receives none of the section 12.3
  hardening headers
  (`crates/trusted-server-adapter-fastly/src/app.rs:2599-2626`). Trace routes
  must therefore emit the 405, its `Allow` value, and the hardening headers
  themselves, as `deny_admin_diagnostic_fallback` already does for the admin
  namespace (`crates/trusted-server-core/src/ec/admin.rs:179-196`).
```

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1852 @@
...
+the raw error. A malformed envelope is rejected as a whole. A network failure
+before an envelope arrives is recorded separately by the browser as
+`evidence_transport_failed`.
+
+Initial-navigation and SPA slot definitions carry their token on the exact slot
+object TSJS already consumes:
+
+```text
+AuctionSlot.ext.trusted_server.trace_slot_ref: string
+```
```
aram356, line 995: 🤔 **thinking** — `AuctionSlot.ext` reads as an existing type; it is net-new on both sides.

`AuctionSlot` is a TypeScript interface at `crates/trusted-server-js/lib/src/core/types.ts:24-30` with exactly five members (`id`, `gam_unit_path`, `div_id`, `formats`, `targeting`) and no `ext`. There is no `pub struct AuctionSlot` in Rust at all; the producer is `build_slot_json` (`crates/trusted-server-core/src/publisher.rs:5612-5636`), which emits a fixed `serde_json::json!` object.

The change is genuinely safe and additive — `build_slot_json` builds free-form JSON rather than a derived struct, and TSJS assigns slots wholesale (`integrations/gpt/index.ts:809`, `:1725`) without validating or stripping unknown members, so an added `ext` flows through harmlessly. The claim that the ordinary fields “remain unchanged” is accurate against the five keys actually emitted. Worth saying “add” rather than describing it as present, since it is a two-sided edit.

````suggestion
Initial-navigation and SPA slot definitions carry their token on the exact slot
object TSJS already consumes. `AuctionSlot` is the TypeScript interface at
`crates/trusted-server-js/lib/src/core/types.ts:24`; it has no `ext` member
today, so this change adds one on both sides — an optional `ext?` on the
interface, and the matching nested key in `build_slot_json`
(`crates/trusted-server-core/src/publisher.rs:5612`), which builds a free-form
`serde_json::json!` object rather than a derived struct:

```text
AuctionSlot.ext.trusted_server.trace_slot_ref: string
```
````

File: docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md [hunk 12]
```diff
@@ -0,0 +1,1852 @@
...
+  for zero-bid, skipped, dispatch-failed, execution-failed, and abandoned
+  outcomes. It is never `AuctionRequest.id`, the telemetry UUID, a provider
+  request ID, or an identifier joinable to user-bearing logs.
+- Auction tokens are `ts-auc-` followed by a lowercase UUID v4 in the existing
+  producer's 32-hex-digit unhyphenated form (`Uuid::new_v4().simple()`). Slot
+  tokens are `ts-slot-` followed by a canonical lowercase hyphenated UUID v4,
+  matching `crypto.randomUUID()`. Both validators enforce UUID version 4 and
+  the RFC variant and reject every other shape. Tokens are compared verbatim,
+  never normalized during validation or correlation; no producer format change
+  is required for the existing GPT auction opportunity marker.
```
aram356: 🤔 **thinking** — The slot-token half describes a producer that does not exist, and the verbatim-comparison rule conflicts with the existing recorder.

The auction-token half is exactly right: `crates/trusted-server-core/src/publisher.rs:5184` emits `format!("ts-auc-{}", uuid::Uuid::new_v4().simple())`.

The slot-token half is prospective, not descriptive — there is no `ts-slot-` producer anywhere in the tree, and `crypto.randomUUID()` is not called anywhere under `crates/trusted-server-js/lib/src/`.

More substantively: “Tokens are compared verbatim, never normalized” does not hold against the existing store. `normalizedAuctionId` (`crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts:203`) trims surrounding whitespace and applies a 256-byte cap before storing the marker, and it is what `recordTrustedServerOpportunity` writes (`store.ts:361`). The verbatim rule needs to be stated against the value as that function returns it, or the join will be specified against a value the recorder never stored.

```suggestion
- Auction tokens are `ts-auc-` followed by a lowercase UUID v4 in the existing
  producer's 32-hex-digit unhyphenated form (`Uuid::new_v4().simple()`,
  `crates/trusted-server-core/src/publisher.rs:5184`). Slot tokens are new in
  this design: no `ts-slot-` producer exists today, and `crypto.randomUUID()`
  is not yet called anywhere in `crates/trusted-server-js/lib/src/`. They are
  `ts-slot-` followed by a canonical lowercase hyphenated UUID v4, matching
  `crypto.randomUUID()`. Both validators enforce UUID version 4 and
  the RFC variant and reject every other shape. Tokens are compared verbatim,
  never normalized during validation or correlation; no producer format change
  is required for the existing GPT auction opportunity marker. Note that the
  existing recorder stores the marker through `normalizedAuctionId`
  (`crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts:203`),
  which trims surrounding whitespace, so the verbatim-comparison rule must be
  applied to the value as that function returns it.
```

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1852 @@
...
+  `ts-ec-extra` remains unrelated. No malformed unrelated pair poisons a
+  reserved-cookie result.
+- Count exact, case-sensitive reserved names before passing values to existing
+  parsers. Zero occurrences is `absent`; more than one is `duplicate`,
+  regardless of whether one value would otherwise be valid. Duplicate
+  precedence is therefore diagnostic rather than first- or last-value
+  selection.
+- Per-value limits are 512 bytes for `ts-ec`, 8 KiB for `ts-eids`, and 16 bytes
+  each for `ts-tester` and `__Host-ts-console`. A single value beyond its limit
+  is `present_invalid/oversized`; it does not change the other three states.
```
aram356, line 696: 🤔 **thinking** — This list mixes one inherited cap with three new ones.

Only the 8 KiB EID cap exists today: `MAX_EIDS_COOKIE_BYTES` at `crates/trusted-server-core/src/ec/prebid_eids.rs:25`. There is no 512-byte `ts-ec` cap anywhere in the codebase — it is introduced here, and it is roughly seven times the only valid length, since `is_valid_ec_id` (`crates/trusted-server-core/src/ec/generation.rs:151-172`) accepts exactly 71 characters (`{64 hex}.{6 alnum}`).

That is defensible as an outer guard before parsing, but presenting it alongside the genuinely-existing EID cap reads as though both are established behavior.

```suggestion
- Per-value limits are 512 bytes for `ts-ec`, 8 KiB for `ts-eids`, and 16 bytes
  each for `ts-tester` and `__Host-ts-console`. Only the 8 KiB EID cap is
  inherited from existing code
  (`crates/trusted-server-core/src/ec/prebid_eids.rs:25`); the others are new
  bounds introduced here. The 512-byte `ts-ec` cap is an outer guard before
  parsing, well above the 71 characters `is_valid_ec_id` actually accepts
  (`crates/trusted-server-core/src/ec/generation.rs:151`). A single value
  beyond its limit
  is `present_invalid/oversized`; it does not change the other three states.
```

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1852 @@
...
+  `application/json; charset=utf-8` for enable, end, and state results.
+- `X-Content-Type-Options: nosniff`
+- `Referrer-Policy: no-referrer`
+- The Content Security Policy specified below.
+- `Permissions-Policy: camera=(), microphone=(), geolocation=(), payment=(),
+usb=()`
+
+```text
+default-src 'none'; script-src 'self'; style-src 'self'; base-uri 'none';
+object-src 'none'; frame-ancestors 'none'; form-action 'none'; connect-src 'self'
```
aram356: 🤔 **thinking** — This CSP has one certain breakage and two risks against the behavior sections 6.3, 6.4, and 13 require.

**Certain — favicon.** `default-src 'none'` with no `img-src` blocks every image load, including the browser's automatic `/favicon.ico` request and any `<link rel="icon">`. On a user-facing mobile page this logs a CSP violation on every load. Either add `img-src 'self' data:` or state explicitly that the page ships iconless.

**Risk — `blob:` download.** Section 13 makes JSON download the last-resort fallback when clipboard and Web Share are unavailable, and section 6.2 offers direct download when storage rejects the snapshot. The existing analogue (`crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/api.ts:242-259`) does `createObjectURL` → `anchor.href` → `click()`. A same-origin `<a download>` to a `blob:` URL is generally not blocked in current browsers, but any implementation that instead does `fetch(blobUrl)` or assigns `iframe.src` is blocked outright by `default-src 'none'`. A last-resort fallback should not depend on which shape the implementer picks — adding `blob:` to the relevant directive makes the intent explicit.

**Risk — inline styles.** `style-src 'self'` forbids both `<style>` blocks and `style="..."` attributes, and it also blocks `element.style.x = ...` assignments from JavaScript. Section 6.3's `aria-live` status region and progressive disclosure commonly toggle `style.display`. The spec handles the inline-*script* case explicitly (line 1333); the inline-*style* case is unaddressed. Either state that all styling is class-toggle-only, or widen the directive.

`connect-src 'self'` is sufficient for the enable/end/state fetches, Web Share is a JS API that CSP does not govern, and `form-action 'none'` is correct given activation is fetch-only.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1852 @@
...
+publicly cacheable.
+
+### 12.5 Operator authentication policy
+
+Trace routing preserves the existing `auth.rs` namespace contract: every
+matching operator Basic Authentication rule is enforced, including `^/_ts` or
+`^/`. Classification may run early, but does not authorize a request. Only
+authentication is factored ahead of trace handling; ordinary event, identity,
+filter, and auction processing stays outside trace routes. A challenge exposes
+no setup context, changes no cookie, and is terminally private/no-store.
```
aram356: 🤔 **thinking** — Two accuracy issues in the auth contract, plus an existing precedent that does exactly what 5.1 forbids.

**“Every matching rule is enforced” is not how selection works.** `handler_for_path` (`crates/trusted-server-core/src/settings.rs:3182-3192`) returns on the **first** match, and `settings.rs:3306-3309` states it outright: “handler selection is first-match-wins over operator regexes, so a narrow handler can shadow the admin namespace”. A narrow rule ordered before a broad `^/` means only the narrow one is ever checked. The section 14.2 test is still satisfiable, but an implementer reading this could assume rules compose.

**There is no fail-closed backstop for `/_ts/trace`.** `enforce_basic_auth` fails closed on an unmatched path only when `Settings::is_admin_path` holds (`crates/trusted-server-core/src/auth.rs:90-97`), and that is exactly `path == "/_ts/admin" || path.starts_with("/_ts/admin/")` (`settings.rs:3198-3200`). An unmatched `/_ts/trace` returns `Ok(None)` — which is the intended public behavior, but it means the “namespace contract” framing overstates the protection.

**Existing carve-out worth an explicit “do not copy”.** Fastly already has an early reserved-route short-circuit for `GET /_ts/debug/ja4` (`crates/trusted-server-adapter-fastly/src/main.rs:99-115`) that loads settings, serves or 404s, and returns — bypassing `enforce_basic_auth` entirely, since auth only runs later in `AuthMiddleware`. That is precisely the ordering exemption section 5.1 rules out, so the new classifier should call it out rather than inherit the pattern.

The good news for the design: `enforce_basic_auth` is fully synchronous, needs only `&Settings` and `&mut Request`, and does no I/O or async secret resolution (`auth.rs:79-82`), so invoking it inside an early classifier is genuinely workable.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1852 @@
...
+  is displayed, including after an earlier local-deletion failure.
+
+These are product limits, not assumptions about browser quota. A storage write
+failure is handled even when the report is below the application limit.
+
+Runtime limits are part of the v1 contract:
+
+| Value                                           | Limit                                                        |
+| ----------------------------------------------- | ------------------------------------------------------------ |
+| Container nesting                               | 8 levels                                                     |
```
aram356: 📝 **note** — This cap has exactly zero headroom.

Applying the section 9.5 rule (the `TraceReportV1` object is level 1; entering an object or an array adds one), the deepest legitimate path in the GPT projection lands on level 8:

```text
L1 TraceReportV1 object
L2 gpt_diagnostics object
L3 slots array
L4 slot object
L5 requests array
L6 request cycle object
L7 requestedSizes array
L8 Size [w, h] array
```

`Size` is `readonly [number, number]` (`crates/trusted-server-js/lib/src/core/types.ts:2`), so that eighth level is real, not hypothetical. The server-auction side tops out at level 7.

This is valid today, but any future nested object inside a request cycle — and `GptDiagnosticsRequestCycle` already contains `durations` as a nested object (`types.ts:208`) — pushes a legitimate report to level 9 and rejects it. Worth either raising the cap or noting the constraint beside the source type.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1852 @@
...
+  invalid client token causes core to mint a server token with no browser
+  correlation; it never changes ordinary auction acceptance. TSJS uses
+  `crypto.randomUUID()` and, if unavailable, omits the client token rather than
+  using weak randomness. Raw publisher slot IDs, ad-unit paths, and internal
+  impression IDs are not copied into this model. Neither the ordinal nor token
+  is a creative number.
+- `source` is assigned by the server call site: initial document auction is
+  `initial_navigation_ssat`, `/_ts/page-bids` is `spa_page_bids`, and
+  `POST /auction` is `auction_api`. Browser `requestPath` does not determine or
+  override this value.
```
aram356: 📝 **note** — These names diverge from the existing `AuctionSource` in two of three cases.

`crates/trusted-server-core/src/auction/telemetry.rs:28-42` already defines the same three-way split with the string forms `initial_navigation`, `spa_navigation`, and `auction_api`. This spec uses `initial_navigation_ssat`, `spa_page_bids`, and `auction_api` — only the last matches.

That is legitimate for a separate public model with its own stability guarantees, and the new names are arguably clearer. Flagging it only so the divergence is a recorded decision rather than an accident, since the two vocabularies will sit side by side when someone correlates a trace report against a telemetry dashboard.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1852 @@
...
+  beside the winning-bid map before initial ad initialization; the corresponding
+  request-scoped slot definitions already carry
+  `ext.trusted_server.trace_slot_ref`. Failed, abandoned, skipped, and zero-bid
+  outcomes still inject their bounded evidence when the publisher document can
+  be delivered.
+- **SPA page-bids:** `/_ts/page-bids` adds an optional, namespaced
+  `trace_auction` transport envelope beside its existing bid result. TSJS
+  validates and records it and consumes each returned slot's
+  `ext.trusted_server.trace_slot_ref` before triggering ad initialization. Both
+  the envelope and slot extensions are absent when the trace gate is inactive.
```
aram356: 📝 **note** — The deprecated page-bids alias is still live and is not covered here.

`PAGE_BIDS_LEGACY_PATH = "/__ts/page-bids"` (`crates/trusted-server-core/src/publisher.rs:6334`) is still registered and still serves the same `{slots, bids}` body; its removal is tracked separately by #970, and the current TSJS bundle retries against it when the canonical path is unusable. A deployment still receiving traffic on the alias would silently produce no SPA trace evidence.

Either add the alias to this bullet or state that trace evidence is canonical-path-only and the alias reports `not_observed`.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1852 @@
...
+9.4. `TraceSlotCorrelationV1` is the browser-produced exact-token sidecar
+defined in section 9.4.1. `TraceGptDiagnosticsV1` is a separate trace-owned
+projection sourced only from `GptDiagnosticsExportV1`. It contains:
+
+- `schema_version: 1` and `source_schema_version: 1`;
+- the source `capturedAt` value;
+- `page.origin` after validation and `page.pathname` fixed to `/[redacted]`;
+- explicit field-by-field projections of current v1 slots, requests, callback
+  issues, attribution issues, coverage, and metadata, subject to the following
+  exclusions and the bounds/truncation below.
```
aram356, line 757: 📝 **note** — “Explicit field-by-field projections” is the right architecture, but this prose is not a usable allowlist.

Checked against the real type: roughly twenty members of `GptDiagnosticsRequestCycle` and `GptDiagnosticsSlotExport` (`crates/trusted-server-js/lib/src/core/types.ts:202-247`) appear in neither the include list nor the exclusion list — among them `requestPath`, `requestIntentId`, `delivery`, `binding`, `durations`, `trustedServerOpportunity`, `trustedServerCreativeFailures`, `loadObservedBeforeRender`, and the visibility percentages.

No active privacy defect: the three path-bearing fields (`page.pathname`, `slotElementId`, `adUnitPath`) are all correctly identified and excluded, and the unlisted members are bounded enums, numbers, and booleans. `requestPath` is worth explicit treatment either way, since section 9.4 already reasons about it (line 840) without placing it in the projection.

The practical risk is that an implementer builds the allowlist from this paragraph instead of from the type and leaves those fields unclassified. Worth stating that the allowlist is derived from the source type, with this list as the exclusion set.

aram356 (changes requested):
## Summary

A design-only PR adding a single 1852-line spec for a `/_ts/trace` mobile
diagnostics endpoint. The document is unusually well-grounded: field names,
constants, and several non-obvious hazards (the `AuctionRequest.id` leak, the
JA4/H2 fingerprint exclusion, bootstrap-fallback argument tolerance, the
unpopulated `asn`) are verbatim correct against the code. All seven findings
from the previous round are genuinely resolved in `e3f371f8c`, and I re-verified
each rather than re-raising it.

The blocking findings below are places where the spec mandates behavior the
pinned platform cannot express, or where it assumes adapter defaults that do the
opposite of what sections 8, 12.3, and 13 require. Because this is a design
document, each one is cheaper to fix now than after it becomes four
implementation PRs.

> 6 of the inline comments below carry a one-click GitHub `suggestion` —
> use **Commit suggestion** (or **Add suggestion to batch**) to apply them. The
> remaining comments describe the change in prose because the fix is a new
> validation hook or spans more than one contiguous range.

## Blocking

### 🔧 wrench

- **Two-second request-body deadline is unimplementable** — see inline at
  `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md:494`
- **`HEAD /_ts/trace` is proxied to the publisher origin** — see inline at
  `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md:468`
- **Router-level 405 carries no `Allow` and no hardening headers** — see
  inline at
  `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md:506`
- **The mandated configuration validation cannot fire** — see
  Cross-cutting below

## Non-blocking

### 🤔 thinking / 📝 note

- **`AuctionSlot.ext` presented as an existing type** — see inline at
  `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md:908`
- **Slot tokens described as existing; verbatim-comparison rule conflicts with
  `normalizedAuctionId`** — see inline at
  `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md:819`
- **Inherited and new cookie caps are presented as one list** — see inline
  at
  `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md:654`
- **CSP blocks the favicon, and leaves `blob:` and inline styles unaddressed**
  — see inline at
  `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md:1326`
- **Auth contract overstates rule composition; existing JA4 route is an
  auth-bypass precedent** — see inline at
  `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md:1364`
- **Container-nesting cap has zero headroom** — see inline at
  `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md:1092`
- **`source` enum diverges from the existing `AuctionSource`** — see inline
  at
  `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md:841`
- **Deprecated `/__ts/page-bids` alias is uncovered** — see inline at
  `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md:996`
- **GPT projection prose is not a usable allowlist** — see inline at
  `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md:715`

- 🔧 **The configuration validation this spec mandates cannot fire as a
  `validator` rule.** Section 8 (line 452) requires that
  “`trace_page_enabled = true` requires `enabled = true`; invalid
  combinations fail configuration validation”, and section 14.1 (line 1433)
  requires a test that configuration “rejects trace-page enablement without
  GPT diagnostics”.

  One correction first, in fairness to the design: adding `trace_page_enabled`
  to TOML *today* is already a hard error on both the enabled and disabled
  paths, because `validate_disabled_schema`
  (`crates/trusted-server-core/src/settings.rs:247-253`) forgives only errors
  beginning `"missing field "`, and an unknown-field error is not forgiven.
  That part is correctly fail-closed.

  The real gap is narrower but still real. Once `trace_page_enabled` is a
  legitimate field, `IntegrationSettings::get_typed` returns before validation:

  ```rust
  // crates/trusted-server-core/src/settings.rs:348-350
  if !config.is_enabled() {
      return Ok(None);
  }

  config.validate().map_err(|err| { /* :352 */ })?;
  ```

  So the exact combination the spec wants rejected — `enabled` omitted (serde
  default `false`) together with `trace_page_enabled = true` — resolves to
  `Ok(None)` and a `#[validate(schema(...))]` rule never runs. Note
  `#[validate(schema(...))]` does otherwise work in this crate
  (`settings.rs:2737`), so the failure mode is silent rather than obvious.

  The precedent that fits is `validate_js_asset_proxy_config`
  (`crates/trusted-server-core/src/config.rs:280-299`): it reads the raw JSON
  and calls `validate()` *outside* the enabled gate, runs from both the deploy
  (`config.rs:250`) and runtime (`config.rs:271`) paths, and is proven by
  `validate_rejects_invalid_disabled_js_asset_proxy_assets`
  (`config.rs:1424`). Naming that pattern here would keep an implementer from
  writing a rule that never fires.

  On sourcing: the “invalid enabled config must not be silently
  logged-and-disabled” rule is not actually in `AGENTS.md` or
  `CONTRIBUTING.md`. Its canonical statement is a HIGH-severity finding in
  `docs/superpowers/specs/2026-03-11-production-readiness-report-design.md:217-233`.
  If this spec relies on it as normative, that is worth making explicit, since
  it was never promoted into the contributor docs.

- 📝 **Section 12.4's cache invariant is already implemented, in two
  places on Fastly.** The spec asks that “tests must prove that late
  response-header handlers cannot make traced content publicly cacheable”
  (line 1355). That guarantee exists today:
  `apply_response_headers_with_cache_privacy`
  (`crates/trusted-server-core/src/response_privacy.rs:163-172`) skips operator
  `response_headers` entries for `Cache-Control` and edge-cache header names
  whenever the response is already uncacheable. On Fastly there is a second
  layer after it — `apply_terminal_response_effects`
  (`crates/trusted-server-adapter-fastly/src/main.rs:371-392`) re-runs the
  privacy guards, because EC finalize and filter effects can add `Set-Cookie`
  later; the regression test is
  `late_filter_effects_cannot_make_an_assembled_response_public`. Citing both
  would let the implementation plan reuse the mechanism instead of rebuilding
  it. Note also that `apply_finalize_headers` is terminal on Axum, Cloudflare,
  and Spin but **not** on Fastly.

- 📝 **A reserved-namespace classifier already exists, at the fallback
  boundary.** Section 8 requires rejecting trailing slashes, extra segments,
  repeated separators, encoded separators, and ambiguous dot segments beneath
  the reserved namespace (lines 509-514). `deny_admin_diagnostic_fallback`
  (`crates/trusted-server-core/src/ec/admin.rs:179-196`) already solves that
  shape for `/_ts/admin`, including a bounded percent-decode-to-fixed-point
  with `MAX_PERCENT_DECODE_ROUNDS = 4`. It runs first inside each adapter's
  fallback rather than at the front door, but it is the pattern to lift
  forward rather than re-derive.

- 📝 **Adapter-parity caveat for section 8.** Several existing `/_ts/*`
  routes (`/_ts/api/v1/*`, `/_ts/set-tester`, `/_ts/clear-tester`,
  `/_ts/debug/ja4`) are registered only on Fastly. Section 8 requires the trace
  namespace on all four adapters, so trace routing cannot follow that
  precedent — worth stating, since it affects the sequencing in section 17.

## CI Status

All checks passing on `e3f371f8c`.

- integration tests: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS
- cargo fmt: PASS
- cargo test: PASS
- vitest: PASS
- format-typescript: PASS
- format-docs: PASS

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [14c274cb](https://github.com/IABTechLab/trusted-server/commit/14c274cbae3e33b338fe6119e98ec4470d4b075d). Section 8 now specifies header prechecks and `Body::into_bytes_bounded(0)` for application-level empty-body acceptance. It explicitly documents upstream buffering on the pinned adapters and removes the one-byte transport-read, two-second deadline, and local 408 guarantees, including the impossible test requirement. I did not retain the suggested structural-read guarantee because it would still overstate what the pinned stack can enforce; transport limits require a separate adapter/upstream change.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [14c274cb](https://github.com/IABTechLab/trusted-server/commit/14c274cbae3e33b338fe6119e98ec4470d4b075d). Section 8 now requires explicit local HEAD handling on every trace path in all four adapters before ordinary dispatch. Read routes return GET-equivalent headers without a body; POST-only actions return a bodyless local 405. The spec cites the existing Fastly fallback regression and adds per-path HEAD parity requirements.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [14c274cb](https://github.com/IABTechLab/trusted-server/commit/14c274cbae3e33b338fe6119e98ec4470d4b075d). Sections 8 and 12.3 now require trace-owned local 405 responses before router dispatch, with `Allow: GET, HEAD` for read routes or `Allow: POST` for actions, plus no-store and all trace error hardening headers. Parity tests must assert these properties. The text distinguishes the router gap from Fastly entry-point finalization, which can still add ordinary headers later.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [14c274cb](https://github.com/IABTechLab/trusted-server/commit/14c274cbae3e33b338fe6119e98ec4470d4b075d). Section 9.4.1 now explicitly describes the two-sided addition: optional `ext?` on the TypeScript `AuctionSlot` interface and the corresponding nested key in Rust `build_slot_json`. It identifies the existing free-form JSON producer rather than implying an existing Rust slot type or extension.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [14c274cb](https://github.com/IABTechLab/trusted-server/commit/14c274cbae3e33b338fe6119e98ec4470d4b075d). Section 9.4 now identifies slot tokens and the TSJS randomUUID call as new work. For GPT projection, token validation and verbatim comparison start from the value returned by existing `normalizedAuctionId`; trace validation performs no additional normalization or repair. The existing unhyphenated auction-token producer format is preserved.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [14c274cb](https://github.com/IABTechLab/trusted-server/commit/14c274cbae3e33b338fe6119e98ec4470d4b075d). Section 9.2 now distinguishes the inherited 8 KiB `MAX_EIDS_COOKIE_BYTES` cap from the three new trace-inspection bounds. It explains that the 512-byte EC limit is only an outer guard and does not broaden the existing exact 71-character validity rule.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [14c274cb](https://github.com/IABTechLab/trusted-server/commit/14c274cbae3e33b338fe6119e98ec4470d4b075d). Section 12.3 now specifies a fixed data-URL favicon with `img-src data:`, external CSS with class/hidden toggles, and no inline styles or style-property writes. It also fixes the download mechanism as Blob object URL → `<a download>` → click → later revocation, explicitly excluding blob fetches and frames. Browser tests must exercise that mechanism under the exact CSP, without widening connect/frame permissions.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [14c274cb](https://github.com/IABTechLab/trusted-server/commit/14c274cbae3e33b338fe6119e98ec4470d4b075d). Section 12.5 now states first-match-wins handler selection, including narrow-rule shadowing and public behavior when no rule matches. It limits the unmatched-path fail-closed backstop to admin, requires synchronous `enforce_basic_auth` before trace responses, and explicitly warns against copying the JA4 early-return bypass. Parity requirements now include rule-shadowing coverage.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [14c274cb](https://github.com/IABTechLab/trusted-server/commit/14c274cbae3e33b338fe6119e98ec4470d4b075d). Section 9.5 raises the container cap to 10 and updates the depth-accounting rule consistently. It records the actual eight-level `requestedSlotSizes` tuple path, requires current-projection and over-depth fixtures, and makes clear that headroom does not allow unknown properties or bypass schema compatibility review.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [14c274cb](https://github.com/IABTechLab/trusted-server/commit/14c274cbae3e33b338fe6119e98ec4470d4b075d). Section 9.4 now records this as an intentional trace-owned vocabulary and explicitly maps existing telemetry `initial_navigation`, `spa_navigation`, and `auction_api` to the public names. It forbids changing telemetry vocabulary or serializing the telemetry enum directly into the trace report.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [14c274cb](https://github.com/IABTechLab/trusted-server/commit/14c274cbae3e33b338fe6119e98ec4470d4b075d). The SPA transport contract now covers both `/_ts/page-bids` and the live `/__ts/page-bids` alias. Both must emit the same gated evidence and slot extensions; the test requirements explicitly cover equivalent behavior and the TSJS retry against the alias.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [14c274cb](https://github.com/IABTechLab/trusted-server/commit/14c274cbae3e33b338fe6119e98ec4470d4b075d). Section 9.3 now contains an exhaustive property allowlist derived from current source interfaces, including nested binding, durations, coverage, and metadata. It explicitly classifies `requestPath` as a bounded intent enum and `requestIntentId` as a local sequence. Callback reasons have a concrete six-value allowlist despite the source string type. Unknown properties remain rejected, and tests must classify each source member as copied, transformed, or excluded.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1982 @@
...
+  following the header-precheck/body-size-check pattern in
+  `crates/trusted-server-core/src/auction/endpoints.rs`. This is an application
+  acceptance limit, not a transport read or allocation limit. Pinned EdgeZero
+  v0.0.8 buffers the Fastly body with blocking `read_to_end` and the Cloudflare
+  body with `req.bytes().await` before core handling. Spin also buffers the
+  body; Axum buffers JSON bodies but can stream other content types. `Body`
+  exposes no read deadline, and Fastly uses `futures::executor::block_on` without a timer.
+  Therefore v1 promises neither a one-byte transport read, a two-second timeout,
+  nor a local 408. Transport-level size/deadline protection requires a separate
+  adapter/upstream change; document the pre-buffering limitation at deployment.
```
aram356: 🔧 **wrench** — The prescribed mechanism cannot produce the `413` this same bullet mandates, and it misreads an empty body on Axum.

The rewrite correctly drops the unimplementable one-byte read and two-second deadline. The replacement mechanism has three verified problems:

1. **`into_bytes_bounded` cannot return 413.** On overflow it returns `Err(EdgeError::bad_request("request body too large"))` (`edgezero-core/src/body.rs:80` for the buffered variant, `:90` for the streaming one). `EdgeError::BadRequest` maps to `StatusCode::BAD_REQUEST` (`edgezero-core/src/error.rs:182`), and `EdgeError` has no `413`/`PayloadTooLarge` variant at all. Following this instruction literally yields `400`, contradicting the `413` required three lines above.
2. **The cited precedent does not use that API.** `crates/trusted-server-core/src/auction/endpoints.rs` does a `Content-Length` precheck (`:126-146`) and then `body.into_bytes().unwrap_or_default()` with an explicit length check (`:148-159`), hand-building `StatusCode::PAYLOAD_TOO_LARGE` both times. `into_bytes_bounded` has zero callers anywhere in this repository. So the sentence names an API and a precedent that disagree with each other.
3. **`into_bytes()` returns `None` for streamed bodies.** `Body::into_bytes` yields `Some` only for `Body::Once` (`edgezero-core/src/body.rs:61-66`). These POSTs send an empty body and therefore carry no `Content-Type`, which is precisely the case Axum streams rather than buffers (`edgezero-adapter-axum/src/request.rs:21-35` — it buffers only `application/json` and `application/*+json`). Copying the precedent's `unwrap_or_default()` would read a **non-empty** streamed body as empty and accept it, which is the exact failure this bullet exists to prevent.

The bullet already notes that Axum can stream non-JSON bodies, so the facts are present; the prescribed mechanism just does not account for them. The emptiness check needs to handle the buffered and streaming variants explicitly and fail closed when the body cannot be read.

```suggestion
- For both POST paths, accept absent or exactly-zero `Content-Length` only
  when the body is empty; reject `Transfer-Encoding`, positive/invalid lengths,
  and any actual body bytes with local `413 Payload Too Large` and no mutation.
  Precheck headers, then drain the body and reject a non-empty one, mirroring
  the header-precheck/body-size-check pattern in
  `crates/trusted-server-core/src/auction/endpoints.rs`, which builds its own
  `StatusCode::PAYLOAD_TOO_LARGE` response. Two adapter details constrain the
  mechanism. `Body::into_bytes_bounded` reports overflow as
  `EdgeError::bad_request`, which maps to `400`, and `EdgeError` has no `413`
  variant, so it cannot produce the status required here. `Body::into_bytes`
  returns `None` for the streaming variant, so an
  `into_bytes().unwrap_or_default()` would read a non-empty streamed body as
  empty; these POSTs carry no `Content-Type`, which is exactly the case Axum
  streams. The emptiness check must therefore handle both the buffered and
  streaming variants explicitly and fail closed when the body cannot be read.
  This is an application
  acceptance limit, not a transport read or allocation limit. Pinned EdgeZero
  v0.0.8 buffers the Fastly body with blocking `read_to_end` and the Cloudflare
  body with `req.bytes().await` before core handling. Spin also buffers the
  body; Axum buffers JSON bodies but can stream other content types. `Body`
  exposes no read deadline, and Fastly uses `futures::executor::block_on` without a timer.
  Therefore v1 promises neither a one-byte transport read, a two-second timeout,
  nor a local 408. Transport-level size/deadline protection requires a separate
  adapter/upstream change; document the pre-buffering limitation at deployment.
```

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1982 @@
...
+object-src 'none'; frame-ancestors 'none'; form-action 'none'; connect-src 'self';
+img-src data:
+```
+
+The shell supplies a fixed data-URL favicon; `img-src data:` allows it without
+an automatic publisher `/favicon.ico` fetch. All styles live in the fixed CSS
+asset: toggle classes or the `hidden` attribute, with no inline style attributes,
+style blocks, or JavaScript style-property writes. JSON download uses a Blob
+object URL assigned directly to an `<a download>` followed by a click, then
+revokes the URL after the download has started. Do not fetch the blob URL or
```
aram356: 📝 **note** — “After the download has started” leaves the revoke trigger undefined.

`URL.revokeObjectURL` called synchronously after `anchor.click()` truncates or cancels the download in some browsers, because the download may not have taken a reference to the blob yet. The usual safe pattern is to revoke from a `setTimeout` after the click rather than on the next statement.

This matters more than it looks: section 13 makes JSON download the last-resort fallback when clipboard and Web Share are unavailable, and section 6.2 makes it the recovery path when browser storage rejects a valid snapshot. A truncated file on that path loses the report the user reproduced the bug to obtain. Worth naming the trigger here, since section 14.3 asks for tests of download behavior and “has started” is not a testable condition.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
```diff
@@ -0,0 +1,1982 @@
...
+usb=()`
+
+```text
+default-src 'none'; script-src 'self'; style-src 'self'; base-uri 'none';
+object-src 'none'; frame-ancestors 'none'; form-action 'none'; connect-src 'self';
+img-src data:
+```
+
+The shell supplies a fixed data-URL favicon; `img-src data:` allows it without
+an automatic publisher `/favicon.ico` fetch. All styles live in the fixed CSS
```
aram356: ⛏ **nitpick** — The causality here is stated backwards, though the outcome is right.

`img-src data:` does not suppress the automatic `/favicon.ico` request. What suppresses it is declaring an icon at all: a browser falls back to `/favicon.ico` only when the document declares no `<link rel="icon">`. The CSP directive's role is narrower — it permits the declared `data:` URL to load, which `default-src 'none'` would otherwise block.

Both halves of the mechanism are correct and the directive is right to be `data:` without `'self'`, since no same-origin image is loaded. Only the explanation is inverted, e.g. “the `<link rel="icon">` data URL suppresses the implicit `/favicon.ico` fallback; `img-src data:` permits it to load.”

aram356 (changes requested):
## Summary

Second review pass, against head `7a82a944a` (base `main`). The previous pass
raised twelve findings; `14c274cba` addresses all twelve, and I verified each
fix against the code at this head rather than taking the replies at face value.

Two are worth calling out as substantively verified rather than merely
annotated. The new section 9.3 allowlist table is genuinely exhaustive: diffing
it mechanically against `GptDiagnosticsRequestCycle` gives 30 real members, 28
allowlisted, and exactly `adManager` and `previousCreativeId` excluded, with no
listed name that does not exist on the real type; the coverage keys, counters,
metadata, `binding`, and `durations` rows all match their interfaces exactly.
And the container-nesting cap moved 8 to 10, which restores two levels of
headroom over the deepest legitimate path (`gpt_diagnostics.slots[].requests[].requestedSizes[][w]`, level 8).

One new blocking finding, introduced by the body-handling rewrite itself: the
replacement mechanism cannot produce the `413` the same bullet mandates, cites
a precedent that uses a different API, and misreads an empty body on the one
adapter that streams it. Details inline.

> 1 of the inline comments below carries a one-click GitHub `suggestion`.
> The other two are prose: one names a trigger condition, one corrects an
> explanation.

## Blocking

### 🔧 wrench

- **Body-emptiness mechanism cannot return `413` and misreads streamed bodies**
  — see inline at
  `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md:516`

## Non-blocking

### 📝 note / ⛏ nitpick

- **Blob revoke trigger is undefined** — see inline at
  `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md:1432`
- **Favicon suppression is attributed to CSP rather than the `<link>` element**
  — see inline at
  `docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md:1428`

- 📝 **Prior-round findings verified as resolved.** Recorded so the next
  pass does not re-litigate them: the unimplementable two-second body deadline
  and one-byte read are gone; configuration validation now specifies a
  raw-config hook modelled on `validate_js_asset_proxy_config`
  (`crates/trusted-server-core/src/config.rs:280-299`) and correctly explains
  why a schema validator never fires for a disabled integration; `HEAD` now
  requires explicit registration on all four adapters, citing
  `dispatch_head_on_named_get_route_falls_through_to_publisher_fallback`; the
  405 contract now requires the trace responder to supply `Allow` and the
  section 12.3 hardening itself rather than inheriting a router error; the CSP
  gained `img-src data:` and a class-toggle-only styling rule; the auth section
  now states first-match-wins and that the fail-closed backstop covers only
  `/_ts/admin`; the deprecated `/__ts/page-bids` alias is covered, and the TSJS
  retry it refers to is real (`crates/trusted-server-js/lib/src/integrations/gpt/index.ts:1675-1692`);
  and the slot-token, cookie-cap, `AuctionSlot.ext`, and `source`-enum
  paragraphs now distinguish new design from existing code. I re-verified the
  underlying code claims at this head.

- 📝 **One claim in the rewrite is accurate and load-bearing enough to
  keep.** The statement that Spin buffers the body while Axum buffers only JSON
  checks out against the pinned dependency: Spin reads the full body into
  `Body::Once` unconditionally (`edgezero-adapter-spin/src/request.rs:74-80`),
  and Axum branches on content type
  (`edgezero-adapter-axum/src/request.rs:21-35`). That asymmetry is what drives
  the blocking finding above, so it is worth keeping the sentence even after
  the mechanism changes.

## CI Status

No failing checks. Several are still running against the merge commit pushed
shortly before this review; they are recorded as pending rather than treated as
findings.

- cargo fmt: PASS
- cargo test: PASS
- vitest: PASS
- format-typescript: PASS
- format-docs: PASS
- Analyze (actions): PASS
- Analyze (rust): PENDING
- Analyze (javascript-typescript): PENDING
- integration tests: PENDING
- CodeQL: SKIPPED

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [155d3b09f](https://github.com/IABTechLab/trusted-server/commit/155d3b09ffb0fe637cf59f6972cf327be5ce7cc9). Section 8 now explicitly handles Body::Once and Body::Stream: empty chunks are skipped, only clean EOF proves a stream empty, the first body bytes produce an explicitly built local 413, and read errors fail closed with local 400 and no mutation. The auction precedent is limited to header/size checks and explicit response construction; the spec warns against both treating into_bytes() None as empty and propagating bounded-read overflow as the response. Adapter buffering limitations remain documented. Sections 14.1 and 14.2 require buffered/streamed, read-error, and no-Content-Type Axum coverage. Docs formatting and git diff --check pass.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [155d3b09f](https://github.com/IABTechLab/trusted-server/commit/155d3b09ffb0fe637cf59f6972cf327be5ce7cc9). Section 12.3 now schedules URL.revokeObjectURL via setTimeout for 1000 ms after the anchor click, forbids synchronous revocation, and requires each download to clean up its own URL. The delay is explicitly not a download-completion signal. Section 14.3 requires fake-timer coverage, including repeated downloads; section 14.4 requires complete-file browser checks under the exact CSP for storage-failure recovery and clipboard/Web Share fallback.

docs/superpowers/specs/2026-09-01-mobile-ad-render-trace-endpoint-design.md
prk-Jr: Addressed in the design spec in [155d3b09f](https://github.com/IABTechLab/trusted-server/commit/155d3b09ffb0fe637cf59f6972cf327be5ce7cc9). Section 12.3 now attributes suppression of the implicit /favicon.ico fallback to the declared <link rel="icon"> data URL. It separately explains that img-src data: permits that URL to load despite default-src 'none'. Section 14.4 also requires a browser check for the declared favicon and absence of the implicit request.

PR #1109, opened 2026-09-01
Harden the Google Tag Manager proxy's upstream and tag selection

### LINKED ISSUE #1115: GTM proxy resolves the tag from the request rather than from configuration
## Description

In `crates/trusted-server-core/src/integrations/google_tag_manager.rs`, `build_target_url` consults the validated `container_id` only when the request carries **no** query string:

```rust
if let Some(query) = req.uri().query() {
    target_url = format!("{}?{}", target_url, query);
} else if path.ends_with("/gtm.js") {
    target_url = format!("{}?id={}", target_url, self.config.container_id);
}
```

Any query string therefore wins, and the response is re-served from the publisher's own origin as `application/javascript` with `Cache-Control: public, max-age=<cache_max_age>`. The same applies to `/gtag/js`, which answers `200` for tag ids that do not exist, so upstream cannot be relied on to reject an unknown one.

Related problems in the same file:

- `upstream_url` is validated only as a URL. It may be plaintext, may embed a username and password, and may carry a wildcard host — and it is handed to the browser in the redirect for an unconfigured tag, so anything in it is disclosed. The proxy fetch is also unpinned: `follow_redirects` defaults on and an empty `allowed_domains` means any host.
- Responses passed through unmodified may carry `Set-Cookie`, `Strict-Transport-Security` or `Clear-Site-Data`, all of which act against the serving origin.
- `GTM_URL_PATTERN` matched on host alone, so *any* path on the Google hosts was rewritten to a first-party URL — but only five paths have routes, so the rest became first-party 404s. On the live GTM script this produced three broken rewrites, two of them landing inside minified JavaScript expressions.

## Expected behavior

- `gtm.js` serves the configured container regardless of what the request asks for.
- `gtag/js` serves only tag ids the operator has configured; any other id is not served from this origin.
- The upstream fetch is pinned to HTTPS and to the configured hosts.
- Passed-through responses cannot set state against the publisher's origin.
- Only routed paths are rewritten first-party.

## Affected area

Integrations (prebid, lockr, permutive, etc.)

## Done when

- [ ] A client-supplied `id` cannot change which container `gtm.js` serves
- [ ] `gtag/js` is limited to configured tag ids, with an operator-visible upgrade note
- [ ] `upstream_url` must be HTTPS with a literal host and no credentials
- [ ] Proxy fetches are pinned to the configured hosts
- [ ] `Set-Cookie` / HSTS / `Clear-Site-Data` are stripped from passthrough responses
- [ ] Rewriting is limited to routed paths, with tests covering prefix collisions and quoting styles

prk-Jr:
## Summary

- The GTM proxy decided which tag to fetch from the client's query string rather than from configuration, so the tag this origin re-served as first-party JavaScript was the caller's choice, not the publisher's.
- The upstream fetch was unpinned: no scheme requirement and an empty domain allowlist, which `is_host_permitted` treats as "any host", while redirects are followed by default.
- Responses passed through unmodified could carry upstream headers that act against the serving origin.

## Operator note — read before upgrading

**If your pages load `gtag/js` with a measurement id different from your container, add those ids to the new `allowed_tag_ids`.**

```toml
[integrations.google_tag_manager]
container_id = "GTM-XXXXXX"
allowed_tag_ids = ["G-XXXXXXXX", "AW-000000"]
```

Nothing breaks if you miss one — the tag is answered with a redirect to Google and still loads for the visitor. But it is then served third-party, so it loses the ad-blocker resilience this integration exists to provide, and a `<link rel="preload">` for it pays an extra round trip. Find the ids your pages use:

```bash
curl -s https://www.example.com/ | grep -oE 'gtag/js\?id=[A-Za-z0-9-]+'
```

This was confirmed against a live deployment: its pages preload two GA4 ids distinct from the container, and both fell back to the redirect until they were listed.

## Behavior

| Request | Before | After |
| ------- | ------ | ----- |
| `gtm.js?id=<anything>` | fetches the id the client named | fetches `container_id`; other params (`l`, `gtm_auth`, `gtm_preview`, `gtm_cookies_win`) preserved |
| `gtag/js?id=<configured>` | proxied | proxied, query rebuilt canonically |
| `gtag/js?id=<other>` | proxied | 302 to upstream, `private, no-store` |
| `gtag/js?ID=<a>&id=<b>` | forwarded verbatim | single validated id sent |
| `/collect`, `/g/collect` | passed through | passed through, minus `Set-Cookie` / `Strict-Transport-Security` / `Clear-Site-Data` |
| script-body URL rewriting | any path on the Google hosts | only the five routed paths |

Tag ids are matched exactly; a different capitalization is treated as unconfigured and redirected. `gtag/js` answers `200` for ids that do not exist, so upstream cannot be relied on to reject an unknown tag.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/integrations/google_tag_manager.rs` | `build_target_url` returns a `GtmTarget` (`Proxy`/`Redirect`); `canonical_query` rebuilds the query so the validated id is the one sent; `tag_id_is_allowed` gates `gtag/js`; `redirect_to_upstream` serves the non-matching case |
| | New `allowed_tag_ids` config plus `validate_allowed_tag_ids`; `validate_https_upstream` requires an https URL with a literal (non-wildcard) host |
| | `build_proxy_config` adds `.with_https_only()` and `.with_allowed_domains(...)`, pinned to the configured upstream and the GA beacon host |
| | `strip_upstream_first_party_headers` applied before any branch, so it covers passthrough and the upstream-error early return |
| | `GTM_URL_PATTERN` narrowed to the routed paths with a terminator guard; host literals consolidated onto constants |
| | `handle` split into `empty_response`, `content_length_rejection`, `payload_size_status`, `rewritten_script_response` |
| `crates/trusted-server-core/src/proxy.rs` | `ASSET_PROXY_STRIP_RESPONSE_HEADERS` renamed to `pub(crate) FIRST_PARTY_PASSTHROUGH_STRIP_HEADERS` and shared, so the two passthrough paths cannot drift |
| `docs/guide/integrations/google_tag_manager.md` | Documents `allowed_tag_ids`, the https requirement, exact matching, the redirect behavior, and an upgrade note |
| `trusted-server.example.toml` | Adds `allowed_tag_ids` with guidance |

34 tests added, covering id clamping and parameter pollution (duplicate, case-varied and percent-encoded keys), the allowlist and exact matching, the redirect, host pinning, header stripping through `handle`, the rewriter's routed-path and terminator behavior, and the config validators.

## Closes

Closes #1115

## Test plan

- [x] `cargo test-fastly && cargo test-axum` (also `test-cloudflare`, `test-spin`)
- [x] `cargo clippy-fastly && cargo clippy-axum` (also `clippy-cloudflare`, `clippy-cloudflare-wasm`, `clippy-spin-native`, `clippy-spin-wasm`, `trusted-server-cli`)
- [x] JS tests and format
- [x] Docs format
- [x] Cross-adapter parity suite: 13 passed
- [x] Manual testing via `fastly compute serve` — verified id clamping, the allowlist before and after configuring real tag ids, exact matching, all three parameter-pollution shapes, the redirect chain through headless Chrome, and beacon passthrough

## Checklist

- [x] New code has tests


prk-Jr:
Closes #1115

crates/trusted-server-core/src/integrations/google_tag_manager.rs
```diff
@@ -281,37 +504,302 @@ impl GoogleTagManagerIntegration {
...
+    /// container this origin serves, and `allowed_tag_ids` widens `gtag/js`
+    /// only, so serving a second container here would quietly extend a setting
+    /// documented for the other endpoint. A page that needs another container
+    /// is a configuration change rather than a request parameter.
+    fn container_target(&self, base: &str, query: Option<&str>) -> GtmTarget {
+        GtmTarget::Proxy(Self::with_query(
+            base.to_owned(),
+            &Self::canonical_query(query, Some(&self.config.container_id)),
+        ))
+    }
```
aram356, line 595: 🤔 **thinking** — `is_rewritable_url` does not consult the `id` param, so on a page running two GTM containers both `gtm.js?id=` URLs are rewritten first-party and the second container's request is answered with the first container's script: container 1 double-fires and container 2 silently never loads. `gtag/js` handles the analogous case with a redirect that keeps the tag alive third-party; here the substitution leaves no signal anywhere — `script_target`'s redirect arm logs, `container_target` does not. Pinning is a defensible, documented choice, so no objection to the behavior itself, but a value-free warn gives operators a way to notice the misconfiguration. The suggestion deliberately does not echo the client-controlled id, matching the redirect log's practice.

```suggestion
    fn container_target(&self, base: &str, query: Option<&str>) -> GtmTarget {
        if Self::requested_tag_id(query).is_some_and(|tag_id| tag_id != self.config.container_id) {
            // The id value is client-controlled, so it is not echoed here.
            log::warn!("Ignoring a gtm.js id that differs from the configured container");
        }
        GtmTarget::Proxy(Self::with_query(
            base.to_owned(),
            &Self::canonical_query(query, Some(&self.config.container_id)),
        ))
    }
```

_(compile-verified only — `cargo fmt --check`, `cargo clippy-fastly`, and `cargo check-fastly`/`check-axum`/`check-cloudflare` pass with this applied; please re-run the matching `cargo test-*` aliases after applying)_

crates/trusted-server-core/src/integrations/google_tag_manager.rs
```diff
@@ -54,34 +65,77 @@ static GTM_CONTAINER_ID_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
...
+    Regex::new(r"^(?:G|GT|GTM|AW|DC|UA)-[A-Za-z0-9-]{1,30}$")
+        .expect("GTM tag ID regex should compile")
+});
+
+/// Host alternation shared by the URL patterns below.
+const GTM_URL_HOSTS: &str =
+    r"(?:https?:)?//(?:www\.(?:googletagmanager|google-analytics)\.com|analytics\.google\.com)";
+/// The paths this integration actually routes, mirroring
+/// [`GoogleTagManagerIntegration::is_rewritable_url`].
+/// The paths this integration routes, plus the empty path.
```
aram356: ⛏ **nitpick** — Two stacked summary lines, an editing leftover — and the first one ("mirroring `is_rewritable_url`") is now inaccurate, since the pattern also routes the empty path which `is_rewritable_url` does not. Keep the second line only:

```suggestion
/// The paths this integration routes, plus the empty path.
```

crates/trusted-server-core/src/integrations/google_tag_manager.rs
```diff
@@ -192,12 +362,43 @@ pub struct GoogleTagManagerIntegration {
...
+        let mut domains = vec![
+            GA_COLLECT_HOST.to_string(),
+            format!("*.{GA_COLLECT_DOMAIN}"),
+            ANALYTICS_HOST.to_string(),
+        ];
+        if let Some(host) = url::Url::parse(&config.upstream_url)
+            .ok()
+            .and_then(|upstream| upstream.host_str().map(str::to_owned))
+        {
+            domains.push(host);
```
aram356, line 401: 🌱 **seedling** — With the default upstream this list includes `www.googletagmanager.com`, but an operator pointing `upstream_url` at their own tagging domain removes it — so a custom upstream that 302s a script fetch to Google fails closed, with no setting to widen the fetch allowlist. Likely acceptable (failing closed is the design, and server-side tagging domains normally serve the script themselves), but worth confirming it is intended. Relatedly, `proxy_config_pins_the_hosts_the_upstream_fetch_may_reach` only covers the default upstream; a companion test asserting a custom upstream's host lands in the allowlist would pin this.

aram356 (approved):

crates/trusted-server-core/src/integrations/google_tag_manager.rs
prk-Jr: Fixed in 3bc43039 — kept the second line only.

crates/trusted-server-core/src/integrations/google_tag_manager.rs
prk-Jr: Applied in 3bc43039. Confirmed the mechanism: `is_rewritable_url` keys only on host and path, so with two containers on a page both `gtm.js?id=` URLs are rewritten first-party and the second is answered with the first container's script. Pinning stays — `allowed_tag_ids` is documented as widening `gtag/js` only — but the mismatch is no longer silent. Took the suggestion as written, without echoing the client-controlled id, and updated the doc comment, which claimed the id "is not consulted at all" and is now read for the warning.

`cargo fmt --check`, all six clippy aliases, and `cargo test-fastly / test-axum / test-cloudflare / test-spin` pass.

crates/trusted-server-core/src/integrations/google_tag_manager.rs
prk-Jr: Fail-closed is intended. An operator repointing `upstream_url` is naming the host that serves the script, so widening the fetch allowlist back to Google would undo the pin on operator input; adding a setting for it is a config decision beyond this PR, and a server-side tagging domain that cannot serve the script itself is a deployment we would rather see fail loudly than proxy around.

Pinned both halves in 3bc43039 with `the_proxy_allowlist_permits_a_custom_upstream_host`: the custom host lands in the allowlist, and `www.googletagmanager.com` no longer matches any entry, so a later change that quietly re-adds it trips the test.

crates/trusted-server-core/src/integrations/google_tag_manager.rs
```diff
@@ -54,34 +65,75 @@ static GTM_CONTAINER_ID_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
...
-/// - `//www.google-analytics.com/g/collect`
-/// - `https://analytics.google.com/g/collect`
-/// - `//analytics.google.com/g/collect`
+/// Covers the prefixes Google serves from `/gtag/js`: GA4 (`G-`), Google Tag
+/// (`GT-`), Tag Manager (`GTM-`), Google Ads (`AW-`), Floodlight (`DC-`) and
+/// legacy Analytics (`UA-`, which carries a property suffix). The pattern
+/// guards against typos in operator config; it is not a security boundary,
+/// because the allowlist itself decides which IDs may be served.
+static GTM_TAG_ID_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r"^(?:G|GT|GTM|AW|DC|UA)-[A-Za-z0-9-]{1,30}$")
```
ChristianPavilonis: 🔧 **Valid Merchant Center tag IDs cannot be allowlisted**: This prefix set omits `MC-`. Google documents `MC-ABCD1234` as a Merchant Center destination used with a Google tag, and deployed pages request `gtag/js?id=MC-...`. Adding one to `allowed_tag_ids` therefore fails configuration validation; omitting it forces the third-party redirect, which a strict `script-src 'self'` policy blocks. Please add `MC` to the regex, validation message, documentation, and positive validation test.

docs/guide/integrations/google_tag_manager.md
```diff
@@ -37,20 +37,51 @@ Add the GTM configuration to `trusted-server.toml`:
...
-| `container_id`         | string  | Yes      | Your GTM Container ID (e.g., `GTM-A1B2C3`)                              |
-| `upstream_url`         | string  | No       | Custom upstream URL (default: `https://www.googletagmanager.com`)       |
-| `cache_max_age`        | number  | No       | Cache duration in seconds (default: `900`, range: `60`-`86400`)         |
-| `max_beacon_body_size` | number  | No       | Max POST body size in bytes (default: `65536`, range: `1024`-`1048576`) |
+| Field                  | Type    | Required | Description                                                                        |
+| ---------------------- | ------- | -------- | ---------------------------------------------------------------------------------- |
+| `enabled`              | boolean | No       | Enable/disable integration (default: `false`)                                      |
+| `container_id`         | string  | Yes      | Your GTM Container ID (e.g., `GTM-A1B2C3`)                                         |
+| `upstream_url`         | string  | No       | Custom upstream URL, must be `https` (default: `https://www.googletagmanager.com`) |
```
ChristianPavilonis: 🔧 **Document the new bare-origin requirement**: The validator now also rejects paths, queries, fragments, credentials, and wildcard hosts. Before this PR, a path-based value such as `https://tags.example.com/gateway` passed validation and produced targets below that path; after upgrading, an enabled deployment using it fails integration registration. Please state that this must be a credential-free HTTPS bare origin with a literal host, and include the breaking validation change in the upgrade note.

ChristianPavilonis (approved):
## Summary

Reviewed `3bc430396b4d577bd29ca1a2f860a93999749234` against `640e93d1389a0fcf1ea6dab1e2d8ea026c5f2045`. The tag-selection clamp, upstream restrictions, response-header stripping, and routed-path rewriting are sound. I found two limited compatibility and upgrade-guidance issues, noted inline.

crates/trusted-server-core/src/integrations/google_tag_manager.rs
prk-Jr: Fixed in f3ee3b7d. Added `MC` to the regex alternation, the validation message, the doc comment, the positive validation test, and both docs surfaces (`allowed_tag_ids` upgrade note and `trusted-server.example.toml`).

Also added `gtag_js_serves_an_allowlisted_merchant_center_tag_first_party`, which asserts `Proxy` rather than `Redirect` so the off-origin fallback can't come back silently.

Two notes on the reasoning, since they change how the fix is justified in the code:

**This is a regression, not a missing feature — worse than your write-up implies.** Base `640e93d1` had no allowlist and no redirect at all (`grep -c` for `allowed_tag_ids|tag_id_is_allowed|Redirect` on the base file returns `0`); `build_target_url` proxied whatever `id` arrived. So `MC-` was served first-party before this PR and only breaks after it. The URL rewriter matches on host+path and is `id`-agnostic, so it does capture the request — the full chain is real:

```
MC- validate() => Err(invalid_tag_id: "allowed_tag_ids entries must be G-, GT-, GTM-, AW-, DC- or UA- tag IDs")
MC- gtag/js target => Some(Redirect("https://www.googletagmanager.com/gtag/js?id=MC-ABCD1234"))
```

**Google does not actually document `MC-`, though.** I checked: [Destination ID: Definition](https://support.google.com/tagmanager/answer/12324787?hl=en) lists only `AW-`/`G-`; [Analyze existing tag configurations](https://developers.google.com/tag-platform/devguides/existing) lists `GT`/`G`/`AW`/`DC`/`UA`; the [Merchant Center Google tag page](https://support.google.com/merchants/answer/16649809?hl=en) shows no ID format at all. `MC-` is real and served from `/gtag/js` — Merchant Center conversion tracking, injected by storefront platforms — but it's corroborated by practitioner sources, not Google's docs.

That matters because "Google documents it" would be the obvious reason to keep it, and it isn't true, so someone could later verify the docs and remove the prefix as bogus. The doc comment therefore states it's absent from Google's tables and explains why it's listed anyway.

Verified: `MC-ABCD1234` accepted; `MC-`, `mc-abcd`, `XX-1234` still rejected (case-sensitive, consistent with the existing prefixes).

docs/guide/integrations/google_tag_manager.md
prk-Jr: Fixed in f3ee3b7d. The `upstream_url` table row now states it must be a credential-free `https` bare origin with a literal host, and there's a new "Upgrading: `upstream_url` must now be a bare origin" section carrying the breaking change. Same constraint added above the `upstream_url` line in `trusted-server.example.toml`.

Confirmed the break the way you described it: base had only `#[validate(url)]` and built `format!("{}/gtm.js", upstream_base)`, so `https://tags.example.com/gateway` validated and produced targets below that path.

I ran the validator over the whole surface rather than writing the table from reading the code, so each row is verified:

```
https://tags.example.com/gateway   => must be a bare origin, with no path, query or fragment
https://tags.example.com?env=stage => must be a bare origin, with no path, query or fragment
https://tags.example.com#anchor    => must be a bare origin, with no path, query or fragment
https://user:pw@tags.example.com   => must not embed a username or password
https://*.example.com              => host must be a literal host, not a wildcard
https://tags.example.com/          => ACCEPTED
```

Two clarifications the note now makes explicit:

- **A trailing slash is still accepted.** The comment at the validator says "a trailing slash yields a double one", which reads as though it were rejected, but `path_is_bare` accepts `"/"` and `upstream_url()` trims it — target came out `https://tags.example.com/gtm.js`, no double separator. Worth documenting so operators don't strip a slash that was never a problem; the code comment's reasoning is just stale.

- **Where the failure lands.** `ts config push` runs the same validation via `TrustedServerAppConfig::new`, so the operator gets a clear error before the config reaches production — this is loud and fail-closed, not a silent prod break. But if a config does reach the runtime unvalidated, `IntegrationRegistry::new` propagates with `?` at `(builder.build)(settings)?`, so the whole registry fails to build, not just this integration. The note says both.

All gates green: fmt, six clippy targets, core 2412, `test-fastly` under viceroy, axum/cloudflare/spin, parity 13, docs prettier (pinned 3.8.1).

PR #1113, opened 2026-09-02
Harden the Edge Cookie withdrawal write path

### LINKED ISSUE #1116: Withdrawal tombstones an identity the graph does not hold
## Description

On a consent withdrawal, `ec_finalize_response` writes a tombstone row for whatever identifier the request's `ts-ec` cookie names. `write_withdrawal_tombstone` performs an unconditional `write_entry(..., Overwrite)` in `crates/trusted-server-core/src/ec/kv.rs`, and the cookie is only shape-checked — `is_valid_ec_id` verifies 64 hex, a dot, and 6 alphanumerics, with no signature recomputation.

The tombstone exists to stop later reads of a real row by batch-sync clients during the 24-hour revocation window. Writing one for an identifier the graph never held enforces nothing, while still consuming a write and occupying a keyspace entry with a TTL. Since the identifier arrives in a client-supplied cookie, that write is the client's to trigger, and the random 6-character suffix is not covered by the HMAC, so each variation is a distinct row.

Recomputing the HMAC does not address it: the suffix is unsigned, so it can be varied freely once a valid prefix is known, and a user whose IP changed between issuance and withdrawal would fail the check and have a genuine opt-out discarded.

Determining existence needs care in both directions:

- A Fastly lookup is eventually consistent, so a stale "missing" would discard a withdrawal for an identity issued moments earlier.
- Counting keys by prefix is not an existence check — any longer key starting with the identifier answers for it.
- If existence cannot be determined at all, writing regardless restores the unconditional write whenever the store can be made to fail.

Separately, `log_id` truncates by byte with `get(..8).unwrap_or(ec_id)`. When byte 8 splits a character the fallback emits the **whole** identifier, and the Fastly lookup error embedded the full key in its message.

## Expected behavior

- Only an identity the graph holds is tombstoned.
- Existence is determined exactly and with a strongly consistent read.
- A store that cannot answer never causes a write for an identity that was never issued, and never silently discards a genuine withdrawal.
- The browser cookie is expired in every case — it is the primary enforcement.
- A redacted identifier is never emitted in full.

## Affected area

Core (Edge Cookies, GDPR)

## Done when

- [ ] A withdrawal for an identity the graph does not hold creates no row
- [ ] A genuine withdrawal still tombstones, verified through the full request path
- [ ] Existence is exact, not a prefix count, and pages are followed to exhaustion
- [ ] An unanswerable check writes nothing and is reported once with its reason
- [ ] Cookie expiry and header removal survive a store outage
- [ ] Redaction holds regardless of the identifier's encoding

prk-Jr:
## Summary

Consent withdrawal now writes a tombstone only for an identity held by this deployment. Unknown client-selected identifiers no longer create rows. The browser cookie is expired even when the store check or write fails.

Existence uses a strongly consistent exact-key check. An eventually consistent miss can hide a newly issued identity that has already been disclosed to a partner; skipping its withdrawal would let later batch sync keep that identity live. Cookie deletion alone does not close that server-side exposure.

## Implementation

- `EcKvStore::key_exists` requires strong consistency and exact equality. The Fastly adapter executes individual list pages, follows their cursors, and compares complete keys.
- Listing is bounded to four pages of 100 keys. `ItemNotFound` terminates immediately. Store errors and an exhausted page budget return an error without an eventual-lookup fallback or a blind write.
- Manual pagination avoids the Fastly SDK iterator path that can reissue a failed page.
- `write_withdrawal_tombstone` returns `Written` or `UnknownIdentity`; an inconclusive check remains an error. Errors are logged at the request boundary, while expected unknown identities log at debug.
- The same call takes the post-withdrawal snapshot write-back as a parameter, and the graph builds that snapshot from the entry it actually wrote. Every path out of the method, the error path included, hands the caller the state it must now hold.
- Identifier logging truncates by character and keeps complete identifiers out of locally constructed error messages.

The existence check and tombstone write remain non-atomic. An entry that expires between them can briefly be restored as a tombstone. The unconditional write preserves withdrawal across concurrent entry updates.

### Removing main's CAS tombstone path

The merge drops `tombstone_existing_from_snapshot`, its `MAX_CAS_RETRIES` loop, the `DisappearOnConflictEcKv` double, and six tests, replacing them with the strong existence check followed by an unconditional overwrite. This is a deliberate trade, not an artefact of the merge.

Withdrawal must always win over a concurrent write. Main's CAS loop could lose that race outright — `tombstone_existing_from_snapshot_returns_failed_after_cas_exhaustion` asserts exactly that losing case, where the row stays live with consent granted. The cost is that a row deleted between the check and the write is recreated as a tombstone. That tombstone denies consent, expires on `TOMBSTONE_TTL`, and cannot mint an identity, so it fails safe; the CAS loop's failure mode does not.

### `key_exists_confirmed` also gates orphan recovery

This helper has a second caller outside the withdrawal path: `confirm_then_recover_orphaned_ec` in `finalize.rs`. Moving it from a prefix count to an exact match changes behaviour there too, and the change is wanted. Under the prefix count, a longer key sharing the orphaned ID as a prefix reported `Ok(true)` and suppressed a legitimate identity rotation — the same prefix collision this PR closes on the withdrawal path.

## Validation

Regression coverage includes a lagging ordinary lookup with a successful strong check and subsequent batch-sync rejection; exact versus prefix matches; later-page and last-allowed-page matches; immediate error termination; page-budget exhaustion; and no ordinary lookup on the withdrawal path. Orphan recovery is covered against a prefix-neighbour key: with only a longer key seeded the orphan is proven absent, so recovery runs and the identity rotates. The Fastly backend also has local Viceroy coverage for real KV operations.

Full local validation: formatting, all six adapter clippy targets, Fastly/Axum/Cloudflare/Spin tests, cross-adapter parity, JS build/tests/format, and docs format.

## Integration considerations

This changes the withdrawal API and touches the finalize path shared with #901 and the provider stack (#1043–#1047, #1084, #1094). Preserve strong existence checking and error handling when reconciling that work; provider-specific identifier validation belongs at the appropriate caller boundary.

Dropping the snapshot write-back during that reconciliation would let post-send pull sync disclose a just-withdrawn identity to partners, so it is a build failure rather than a silent regression: both stale call shapes now fail to compile against the new signature. #901 is the one to watch — its `write_withdrawal_tombstone` returns `Result<(), _>` with no existence gate at all, so landing it over this branch would revert the fix outright rather than only lose the write-back.

Closes #1116


jwrosewell:
**Sequencing note on this PR and the provider stack.**

This PR and the open provider stack (#1043 to #1047, #1084, #1094) rework the same Edge Cookie finalize flow. A merge simulation of this PR's head (f6181d149, and identically its earlier head b5b68bba9) against six of the seven stack heads conflicts in one file, `crates/trusted-server-core/src/ec/finalize.rs`, in four regions, being two import collisions and two larger blocks of roughly 40 and 90 lines where this PR's withdrawal-path rework and the stack's permission-gate restructuring rewrite the same code. The spec-only #1084 merges clean, and every other file in this PR auto-merges.

Reproduce:

```
git fetch upstream main refs/pull/1113/head:pr-1113
git merge-tree --write-tree --name-only pr-1113 <stack head>
```

The request is the one we have made on #885, #940 and #1094. The stack has been open since 19 August, carrying work that has been under review since 2 July as #838, is green on required CI, and its branches are kept rebased close to `main`, so please land the stack first, or say here that this PR goes first so we rebase once against a known base. The tombstone tightening here sits directly on the finalize path the stack restructures, so sequencing the two deliberately keeps both reviews readable.

prk-Jr:
Closes #1116

crates/trusted-server-adapter-fastly/src/ec_kv.rs
```diff
@@ -117,24 +131,63 @@ impl EcKvStore for FastlyEcKvStore {
...
+            .iter()
+            // Unbounded here on purpose: `contains_exact_key` stops reading at
+            // its budget, so the page count lives with the code that is tested
+            // against it rather than being recomputed at the call site.
+            .map(|page| match page {
+                // A store that reports nothing under the prefix is telling us
+                // the key is absent, which is what `lookup` already does for a
+                // missing key. Treating it as a failure would send every
+                // withdrawal naming an unissued identity down the error path.
+                Err(fastly::kv_store::KVStoreError::ItemNotFound) => Ok(Vec::new()),
```
aram356: 🔧 **wrench** — Mapping `ItemNotFound` to an empty page does not end the listing; it makes the SDK re-issue the list call.

`fastly-0.12.1`'s `ListResponse::next()` declares `iterator_did_error` but **never assigns `true` to it** — the field appears only three times in the SDK (declaration at `kv_store.rs:139`, init at `:170`, read at `:185`). Its error path returns `Some(Err(..))` early, *before* the `keys.is_empty() -> None` check at `kv_store.rs:215-217` that would terminate iteration. So an error yielded from a page does not close the iterator: the next `next()` call issues another `handle.list(...)` with the same cursor.

Converting that error into `Ok(Vec::new())` here therefore does not mean "absent" — it means `contains_exact_key` sees a non-matching page and asks for another one, repeatedly, until the page budget trips.

I reproduced this with a harness replicating the SDK's `next()` control flow exactly:

```text
A: persistent ItemNotFound    -> result=Ok(Undetermined)          backend_calls=5
B: persistent TooManyRequests -> result=Err("TooManyRequests")    backend_calls=1
C: found on page 1            -> result=Ok(Found)                 backend_calls=1
```

Case A is the intended-benign path. `KvSysError::NotFound` maps to `KVStoreError::ItemNotFound` in the generic `From<KvSysError>` impl used by every KV operation including list (`fastly-0.12.1/src/kv_store/handle.rs:75`), so this arm is reachable. The outcome is the opposite of what the comment above it states: instead of avoiding the error path, a withdrawal naming an unissued identity costs 5 list round trips and *then* goes down the error path via `Undetermined`, into the `lookup_raw` fallback.

**Apply manually** — cannot be expressed as a single-file `suggestion`, because the durable fix is the one in the `kv.rs:666` comment (use `lookup`), which removes this method. If the `list` approach is kept, this arm must terminate the iteration rather than yield an empty page — e.g. by taking the pages via a `.take_while`/`scan` that stops on the first `ItemNotFound`, or by treating `ItemNotFound` as a definitive `Absent` result before `contains_exact_key` is entered.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -617,10 +643,33 @@ impl KvIdentityGraph {
...
+    ///
+    /// A value longer than any identifier this deployment could hold is
+    /// refused without a store round trip. That bound is about cost, not
+    /// correctness — the check is exact either way, so it does not care what
+    /// shape an identifier takes.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`TrustedServerError::KvStore`] on store error.
+    fn key_exists_confirmed(&self, ec_id: &str) -> Result<bool, Report<TrustedServerError>> {
```
aram356: 🔧 **wrench** — `EcKvStore::lookup` already answers this question in one round trip, and the strong-consistency argument for preferring `list` is not maintained end to end.

**Cost.** On the withdrawal response path at the edge:

| Case | Before this PR | After |
| --- | --- | --- |
| Identity held (the normal withdrawal) | 1 write | 1 list + 1 write = **2** |
| Identity not held | 1 write | **1** list |
| Degraded (list failing or over budget) | 1 write | up to 5 lists + 1 lookup + 1 write = **~7** |

`build_list` reads from the primary data source, which the SDK documents as the slower path (`fastly-0.12.1/src/kv_store.rs:495`).

**The consistency rationale.** The doc comment at `kv.rs:654-657` rejects `lookup` because a lagging read "would report a freshly written identity as absent, discarding a genuine withdrawal." But the fallback immediately below (`kv.rs:733-744`) does exactly that: when the exact check fails it consults `lookup_raw`, and treats `Ok(None)` as grounds for not writing. The guarantee is therefore held only on the happy path and abandoned precisely when the store is degraded. It is not load-bearing.

Note also that the check and the write are already non-atomic (acknowledged at `kv.rs:690-695`), so a strong read at the check does not close the TOCTOU window either way.

**What this buys.** `EcKvStore::lookup` is already on the trait (`kv_backend.rs:134`), already implemented by `FastlyEcKvStore` (`ec_kv.rs:64-88`), already maps `ItemNotFound -> Ok(None)`, and is already exposed here as `lookup_raw` (`kv.rs:201`). Existence becomes:

```rust
fn key_exists_confirmed(&self, ec_id: &str) -> Result<bool, Report<TrustedServerError>> {
    Ok(self.lookup_raw(ec_id)?.is_some())
}
```

Exact-key by construction, one round trip always, no prefix-collision concern. That change also deletes `contains_exact_key`, `ExactKeyMatch`, `EXACT_MATCH_PAGE_SIZE`, `EXACT_MATCH_MAX_PAGES`, the `Undetermined` third state, the `FastlyEcKvStore::key_exists` body that carries the `ec_kv.rs:160` defect, the new `EcKvStore::key_exists` trait method along with its five test-double implementations, and the `pub` widening of `crates/trusted-server-core/src/ec/kv_backend.rs`.

Worth weighing explicitly: the residual risk of using `lookup` is a missed tombstone for an identity issued within the replication lag *and* withdrawn in that same window. The browser cookie is expired unconditionally in every one of those cases (`finalize.rs:158`), which this code itself calls the primary enforcement, so the exposure is limited to the batch-sync window.

**Apply manually** — spans `kv.rs`, `kv_backend.rs`, and `ec_kv.rs` together; cannot be a single-file `suggestion`.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -123,6 +123,22 @@ impl fmt::Debug for KvIdentityGraph {
     }
 }
 
+/// Longest identifier this store will look for.
+///
+/// Bounds the work a caller-supplied cookie can ask for. It is not a format
+/// check: the existence check is exact, so it does not depend on the shape of
+/// an identifier.
+const MAX_EC_ID_LEN: usize = 256;
```
aram356: 🤔 **thinking** — This bound cannot be reached from any production caller, so it is not bounding client-supplied work.

`key_exists_confirmed` is private (`fn`, not `pub(crate)`), and its only non-test caller is `write_withdrawal_tombstone`. That in turn has exactly one production call site: `crates/trusted-server-core/src/ec/finalize.rs:167`. Every other call is inside a `#[cfg(test)]` module (`admin.rs:1126`, and the tests in this file).

That single production path sources its identifiers from `withdrawal_ec_ids` (`finalize.rs:205-221`), which gates both the cookie value and the active value through `is_valid_ec_id`. That predicate (`crates/trusted-server-core/src/ec/generation.rs:151-171`) requires exactly two dot-separated segments, a 64-character ASCII-hex first segment and a 6-character alphanumeric second — pinning every identifier that reaches this function to exactly 71 ASCII bytes. Never empty, never above 256.

So the doc comment here and the PR description both describe the bound as bounding "the work a caller-supplied cookie can ask for", but the cookie is already constrained to 71 characters two frames up, before the value arrives. The tests at `kv.rs:1558` and `kv.rs:1586` cover a branch production cannot take.

Not harmful — it is O(1) on a private function. But it reads as a load-bearing safety check and is not one, and if the `lookup`-based check above is adopted the length bound stops mattering entirely. Worth either dropping it or restating the comment as defence-in-depth for a hypothetical future caller.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -630,15 +679,79 @@ impl KvIdentityGraph {
...
+                // The raw form is deliberate: a row whose body no longer
+                // deserializes is still a row, and presence is all that matters.
+                //
+                // Failing to determine existence is an error, not a third
+                // outcome. A caller that only inspects the error case still
+                // reports it, where an extra `Ok` variant would be discarded in
+                // silence.
+                match self.lookup_raw(ec_id) {
+                    Ok(Some(_)) => {
+                        log::warn!(
```
aram356: ♻️ **refactor** — A confirmed identity that then gets written is a success, and logging it at `warn` inverts the level convention this PR establishes elsewhere.

This line is reached on the fallback path the comment directly above describes as the intended degraded behaviour, and it fires only when the lookup *succeeded* and the tombstone write proceeds. Compare `log_tombstone_outcome` in `finalize.rs`, added in this same PR, which puts the expected `UnknownIdentity` case at `debug` (`finalize.rs:186`) and reserves `error` for the case where nothing was recorded (`finalize.rs:196`). By that rubric a confirmed-and-written fallback belongs at `debug`.

Volume: `withdrawal_ec_ids` returns a `HashSet` of up to 2 identifiers (`finalize.rs:205-221`), and `apply_withdrawal_tombstones` calls the writer once per entry, so a store-wide list outage produces up to 2 `warn` lines per withdrawal request across every POP, unsampled — while operators are reading logs to find the actual fault. It scales with withdrawal traffic rather than total traffic, so it is bounded, but it is per-request noise at `warn` for a non-fault.

```suggestion
                        log::debug!(
```

_(Verified in a scratch worktree: `cargo fmt --all -- --check` clean, `cargo clippy-fastly` clean, `cargo test -p trusted-server-core --target wasm32-wasip1` 2280 passed.)_

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -630,15 +679,79 @@ impl KvIdentityGraph {
...
+                            log_id(ec_id),
+                        );
+                    }
+                    Ok(None) => {
+                        return Err(list_error.attach(
+                            "a lookup found nothing, but it may lag behind a recent write",
+                        ));
+                    }
+                    Err(lookup_error) => {
+                        return Err(lookup_error.attach(format!(
```
aram356: 📝 **note** — No leak here today; flagging the invariant this line now depends on.

`current_context()` returns `&TrustedServerError`, and `KvStore`'s `Display` renders the `message` field verbatim (`crates/trusted-server-core/src/error.rs:57-58`), so whatever a backend put in `message` lands in this string.

I traced every construction site reachable as `list_error` — it originates from `self.store.key_exists(ec_id)`, so on Fastly that is `ec_kv.rs:165`, `ec_kv.rs:176-179`, and `open_store()` at `ec_kv.rs:36`/`:41`. The first two redact through `log_id`; the third carries no identifier. I also checked the remaining `message:` constructions in `ec_kv.rs` (73, 115, 134, 190), `kv.rs` (172, 178, 184, 232, 238, 270, 768), and `kv_backend.rs:295`. All are redacted or identifier-free, so **there is no disclosure today**.

The structural point is that this converts a typed error into an opaque interpolated string, so the redaction invariant now rests on every present and future author of a `TrustedServerError::KvStore { message }` remembering to call `log_id`, with nothing at the type level enforcing it. A second backend implementing this trait would break it silently. If `key_exists` goes away per the `kv.rs:666` comment, so does this line.

Minor inconsistency alongside it: the `Ok(None)` arm just above attaches a static string and discards the lookup detail, while this arm interpolates the other error's rendering. Both are defensible, but they preserve different amounts of context.

aram356 (changes requested):
## Summary

Gating the withdrawal tombstone on an existence check is the right call, and the premise holds up: `handle_batch_sync` maps both `UpsertResult::NotFound` and `UpsertResult::ConsentWithdrawn` to the same `REASON_INELIGIBLE` (`crates/trusted-server-core/src/ec/batch_sync.rs:211`), so skipping the write for a row that does not exist costs no enforcement. The identifier-redaction sweep is thorough, and the negative-case tests are well chosen.

The objection is to the mechanism rather than the goal. Choosing a prefix `list` over the `lookup` already on the trait introduces a reachable defect in the only production backend, doubles the happy-path round trips on the withdrawal response path, and pulls in the paging machinery, the third `Undetermined` state, the new trait method, and the expanded public surface that come with it. Switching the check to `lookup_raw` resolves the first finding and deletes the rest.

Verified locally against `7b5ea5720`: `cargo clippy-fastly` clean, `cargo fmt --all -- --check` clean, `cargo test -p trusted-server-core --target wasm32-wasip1` 2280 passed, cross-adapter parity 13 passed.

> 1 of the inline comments below carries a one-click GitHub `suggestion` — use **Commit suggestion** to apply it. The remaining comments describe the fix in prose because the change spans multiple files or lines outside the diff and cannot be auto-applied.

## Blocking

### 🔧 wrench

- **`ItemNotFound` mapped to an empty page re-issues the list instead of ending it** — see inline at `crates/trusted-server-adapter-fastly/src/ec_kv.rs:160`
- **`lookup` answers existence in one round trip, and the strong-consistency rationale is not maintained end to end** — see inline at `crates/trusted-server-core/src/ec/kv.rs:666`
- **The only production `EcKvStore` implementation has no test coverage** — see Cross-cutting below

## Non-blocking

### 🤔 thinking / ♻️ refactor / 📝 note

- **`MAX_EC_ID_LEN` is unreachable from the production call graph** — see inline at `crates/trusted-server-core/src/ec/kv.rs:131`
- **`warn` for a successful, expected fallback inverts this PR's own level convention** — see inline at `crates/trusted-server-core/src/ec/kv.rs:735`
- **Redaction verified clean at every reachable construction site** — see inline at `crates/trusted-server-core/src/ec/kv.rs:746`

- 🔧 **The only production `EcKvStore` implementation has no test coverage.** `crates/trusted-server-adapter-fastly/src/ec_kv.rs` has no `#[cfg(test)] mod tests`, and `FastlyEcKvStore` is the sole non-test implementor of the trait — `crates/trusted-server-adapter-{axum,cloudflare,spin}` contain no `EcKvStore` implementation at all, and the only production construction sites are `crates/trusted-server-adapter-fastly/src/main.rs:445` and `:477`.

  The PR notes the pagination loop as a known gap, but the gap is wider than pagination. `contains_exact_key` is a pure function over an already-materialized page iterator and carries 8 tests; the untested code is the adapter glue that builds that iterator, decides what an `Err` page means, and converts `Undetermined` into a `Report` — which is exactly where the first finding above lives. A test double implementing the `Iterator<Item = Result<ListPage, KVStoreError>>` shape and asserting the round-trip count would have caught it without a live Fastly store.

## CI Status

- Analyze (actions): PASS
- Analyze (rust): PENDING
- CodeQL: SKIPPED
- cargo fmt: PASS (required)
- cargo test: PASS (required)
- vitest: PASS

No failing checks. The three pending gates were still in progress at review time.

crates/trusted-server-core/src/ec/kv.rs
prk-Jr: Adopted. `key_exists_confirmed` is now `Ok(self.lookup_raw(ec_id)?.is_some())` (kv.rs:654).

Agreed the strong read was not load-bearing: the `Ok(None)` arm of the fallback returned without writing, so the guarantee was held on the happy path and dropped exactly when the store was degraded. And the check/write pair is non-atomic regardless.

That swap deleted everything it was holding up: `contains_exact_key`, `ExactKeyMatch`, `EXACT_MATCH_PAGE_SIZE`, `EXACT_MATCH_MAX_PAGES`, the `Undetermined` state, `EcKvStore::key_exists` and its five test-double impls, the `pub` widening in kv_backend.rs, and the entire lookup fallback block — which is where the comments at :735 and :746 lived. Net -186 lines across the two commits.

The residual risk you named is now in the doc comment rather than implicit: an identity issued and withdrawn inside the replication lag is reported unknown and gets no tombstone; the cookie is expired unconditionally, so the exposure is the batch-sync window.

Added a round-trip budget test as a regression guard (`a_withdrawal_costs_one_read_whether_or_not_the_identity_is_held`) so a future change cannot quietly reintroduce a two-read gate.

Commit 0031fc3a.

crates/trusted-server-adapter-fastly/src/ec_kv.rs
prk-Jr: Confirmed against the SDK before acting on it. `ListResponse::next` (fastly-0.12.1/src/kv_store.rs:178-224) declares `iterator_did_error` but never assigns it, and the error arm `return Some(Err(e))` sits above the `keys.is_empty() -> None` check, so a page error leaves the iterator open with `next_cursor` and `data` untouched — the next call re-issues `handle.list(...)` with the same cursor. And the arm is reachable: `KvSysError::NotFound -> KVStoreError::ItemNotFound` is in the generic `From<KvSysError>` impl at kv_store/handle.rs:75, shared by every KV operation including list.

So the comment above that arm described the opposite of what it did. Resolved by the durable fix you pointed at rather than by terminating the iteration: `key_exists` is gone from the trait and from this backend, along with both `EXACT_MATCH_*` constants.

Commit 0031fc3a.

crates/trusted-server-core/src/ec/kv.rs
prk-Jr: Checked and it holds: `key_exists_confirmed` is private, its only non-test caller is `write_withdrawal_tombstone`, and that has one production call site (finalize.rs:167) whose identifiers come from `withdrawal_ec_ids`, gated through `is_valid_ec_id` — exactly 71 ASCII bytes, never empty.

`MAX_EC_ID_LEN` and the guard are dropped, along with the branch-only test. With the lookup-based check nothing is scanned, so there is no work for a bound to bound.

Commit 0031fc3a.

crates/trusted-server-core/src/ec/kv.rs
prk-Jr: The level convention point is right, and the line is gone rather than lowered — the whole lookup fallback block went with the switch to a lookup-based check, so there is no successful-fallback path left to log at any level.

`log_tombstone_outcome` in finalize.rs keeps the rubric you described: `debug` for the expected `UnknownIdentity`, `error` only when nothing was recorded.

Commit 0031fc3a.

crates/trusted-server-core/src/ec/kv.rs
prk-Jr: The line is gone with the fallback block, so both the interpolation and the asymmetry between the two arms go with it.

The structural point stands on its own though, and I have not addressed it: nothing at the type level stops a future `TrustedServerError::KvStore { message }` from carrying an unredacted identifier. Every present site is redacted or identifier-free — including the ones you traced — but that rests on authorial discipline. Worth a tracked follow-up (a newtype that redacts on construction, or a `Display` that does) rather than something to bolt onto this PR.

Commit 0031fc3a.

prk-Jr:
Addressed all five findings. Two commits: 0031fc3a swaps the check, cee87c20 adds the adapter coverage. Net -186 lines.

**Blocking**

- **ItemNotFound re-issues the list** — confirmed against fastly-0.12.1 before acting: `iterator_did_error` is declared but never assigned, and the error arm returns above the `keys.is_empty()` termination check, so a page error leaves the iterator open on the same cursor. Resolved by the durable fix rather than by terminating the iteration.
- **Use `lookup`** — adopted. `key_exists_confirmed` is now `Ok(self.lookup_raw(ec_id)?.is_some())`. That deleted `contains_exact_key`, `ExactKeyMatch`, both `EXACT_MATCH_*` constants, the `Undetermined` state, `EcKvStore::key_exists` and its five test doubles, the `pub` widening in kv_backend.rs, and the fallback block that hosted the :735 and :746 comments. The residual replication-lag exposure you named is documented in the doc comment.
- **No adapter test coverage** — `fastly.toml` already declares `ec_identity_store` for the local simulator and the adapter tests run under Viceroy, so `FastlyEcKvStore` is testable against a real KV store with no live service. Six tests added: unlinked store, missing key reading absent rather than failing, insert/lookup/delete round trip, `Add` precondition, generation mismatch, prefix counting.

**Non-blocking**

- `MAX_EC_ID_LEN` — dropped with its branch-only test. Verified your call-graph trace first (private fn, one production call site through `is_valid_ec_id`, fixed 71 bytes).
- `warn` level and the interpolated error string — both lines removed with the fallback block. The redaction invariant point survives its line; noted as a follow-up rather than bolted on here.

Also added a round-trip budget test pinning one read per withdrawal, so a future change cannot quietly restore a two-read gate.

**Verification:** `cargo fmt --all -- --check` clean; all six clippy targets clean; `cargo test-fastly` 172 + 2271 passed (adapter 166 -> 172 from the new tests, core 2280 -> 2271 as the removed tests exceed the one added); axum, cloudflare, spin suites pass; cross-adapter parity 13 passed.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -617,10 +636,25 @@ impl KvIdentityGraph {
...
+    /// Reads the key itself, so the answer is about that identity and no
+    /// other: counting keys by prefix would let a longer key carrying this one
+    /// as a prefix answer in its place. One round trip, and no bound on the
+    /// identifier is needed because nothing is scanned.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`TrustedServerError::KvStore`] on store error.
+    fn key_exists_confirmed(&self, ec_id: &str) -> Result<bool, Report<TrustedServerError>> {
+        Ok(self.lookup_raw(ec_id)?.is_some())
```
ChristianPavilonis: **P1 — An eventually consistent miss can discard a genuine withdrawal**

`lookup_raw` uses Fastly’s eventually consistent lookup path. If an identity is withdrawn shortly after issuance, a lagging POP can return `None`, causing line 703 to classify the real identity as `UnknownIdentity` and skip the tombstone.

This is not limited to the browser response. The issuance request can already have disclosed the EC ID through post-response pull sync (`crates/trusted-server-adapter-fastly/src/main.rs:221-222`). Once the original row becomes visible, a later batch sync sees `consent.ok == true` (`kv.rs:593-600`), accepts a partner mapping, and writes it with the one-year live-entry TTL (`kv.rs:620-624`). The missed withdrawal can therefore leave the identity live and writable for up to a year, with a later write refreshing that TTL.

Please use a strongly consistent exact-key existence operation. If paginated listing is restored, handle Fastly `ItemNotFound` by explicitly terminating the operation rather than yielding an empty iterator page, and return an error when the page budget is exhausted instead of falling back to eventual lookup. A regression test should model ordinary lookup missing a newly written row while the strongly consistent existence path still finds it.

ChristianPavilonis (changes requested):
## Summary

The write gate correctly prevents arbitrary client-selected identities from creating tombstones, and the redaction and local outcome handling look sound. I found one blocking correctness issue in the existence decision; details are inline.

crates/trusted-server-adapter-fastly/src/ec_kv.rs
prk-Jr: Updated in eea0f6ba. Strong listing is now driven one page at a time with `ListBuilder::execute()` and explicit cursor handling; it does not use the SDK iterator. `ItemNotFound` returns immediately, and other errors or the four-page budget return an error without an eventual-lookup fallback.

The adapter regression asserts that `ItemNotFound` causes exactly one fetch. Later-page matches, the last allowed page, and budget exhaustion are also covered. These tests and the full Fastly suite passed.

crates/trusted-server-core/src/ec/kv.rs
prk-Jr: The later correctness finding changes the conclusion in my earlier reply: cookie expiry does not bound the server-side exposure. A partner can already hold a freshly issued ID, and a missed withdrawal lets subsequent batch sync keep that row live.

In eea0f6ba, existence therefore uses a required strongly consistent exact-key operation. Fastly executes bounded individual list pages with explicit termination, so the iterator defect is avoided. There is no eventual-lookup fallback. The extra primary-store read is necessary for this guarantee; the tests assert one existence operation and zero ordinary lookups, without claiming that pagination is a single network round trip.

The replication-lag regression models an ordinary lookup miss, verifies that withdrawal still writes a tombstone, and confirms that subsequent batch sync returns `ConsentWithdrawn`. Full local CI gates passed.

crates/trusted-server-core/src/ec/kv.rs
prk-Jr: The unused `MAX_EC_ID_LEN` guard and its branch-only test remain removed. The final strong-list implementation in eea0f6ba bounds scanning directly with a four-page budget of 100 keys per page, rather than describing an unreachable identifier-length check as the work limit. Budget-boundary tests passed.

crates/trusted-server-core/src/ec/kv.rs
prk-Jr: This remains addressed in the final implementation (eea0f6ba): there is no successful-fallback path and no warning for one. An inconclusive strong check returns an error to the request boundary; an expected `UnknownIdentity` still logs at debug. The restored strong check does not reintroduce the removed warning. Full EC tests passed.

crates/trusted-server-core/src/ec/kv.rs
prk-Jr: The opaque error interpolation remains removed in eea0f6ba. The restored strong-existence path uses `log_id` in both its list-error context and page-budget error, and the new budget regression asserts that the full identifier is absent from the rendered report.

This preserves the current redaction behavior; it does not add type-level enforcement for every future `KvStore` message. That broader limitation remains as noted in the earlier reply.

crates/trusted-server-core/src/ec/kv.rs
prk-Jr: Fixed in eea0f6ba. `EcKvStore::key_exists` now requires strongly consistent exact-key existence. Fastly follows explicit list cursors with a four-page budget, terminates immediately on `ItemNotFound`, and returns an error on other failures or budget exhaustion. It never falls back to eventual lookup.

`withdrawal_does_not_lose_a_new_identity_to_lookup_lag` models ordinary lookup missing the issued row, verifies the tombstone is written, and checks that later batch sync returns `ConsentWithdrawn`. Adapter tests cover pagination and its termination cases. The full local adapter, lint, parity, JS, and formatting gates passed.

ChristianPavilonis (approved):
## Review summary

Reviewed `12f5f1e8e7102c8e2fe633e6907ab8d178db4860` against `640e93d1389a0fcf1ea6dab1e2d8ea026c5f2045`.

No actionable regressions found in the seven changed files. Traced the bounded, strongly consistent exact-key check through withdrawal, orphan recovery, request snapshots, and downstream identity consumers. The earlier lookup-lag and pagination findings are addressed.

## Safety proof

- Executed regression tests confirm that a lagging ordinary lookup does not discard withdrawal of a strongly confirmed identity, unknown keys do not create tombstones, and subsequent batch sync rejects the withdrawn identity.
- Executed finalization tests confirm that a failed existence check still expires the browser cookie, removes the EC response header, and invalidates the live snapshot.
- Exact-head CI logs confirm all 13 Fastly KV tests passed, including explicit cursor traversal, immediate error termination, page-budget exhaustion, and real simulator KV operations. The pinned Fastly SDK defaults list requests to strong consistency.

## Validation

- `cargo test -p trusted-server-core --locked ec::`: 321 passed.
- `cargo test -p trusted-server-core --locked --lib --quiet`: 2,367 passed.
- `cargo clippy -p trusted-server-core --locked --all-targets --all-features -- -D warnings`: passed.
- `cargo fmt --all -- --check`: passed.
- `cargo test-fastly --locked ec::`: blocked locally before tests by missing `wasm32-wasip1`; Viceroy is also unavailable locally. Inspected exact-head CI logs confirming Fastly KV tests and the EC lifecycle integration test passed.
- All reported PR checks passed. Reviewed existing reviews, comments, replies, and all six resolved threads; no duplicate findings.

## Residual risk

Live Fastly replication and concurrent deletion were not reproduced. The existence check and tombstone write remain non-atomic. An inconclusive check intentionally leaves withdrawal unrecorded server-side while expiring the browser cookie.

crates/trusted-server-adapter-fastly/src/ec_kv.rs
```diff
@@ -5,12 +5,56 @@
...
-use fastly::kv_store::{InsertMode, KVStore};
+use fastly::kv_store::{InsertMode, KVStore, KVStoreError, ListPage};
 use trusted_server_core::ec::kv_backend::{
     EcKvLookup, EcKvStore, EcKvWrite, EcKvWriteMode, EcKvWriteOutcome,
 };
+use trusted_server_core::ec::log_id;
 use trusted_server_core::error::TrustedServerError;
 
+/// Bounds the work a withdrawal can trigger when keys share its prefix.
+const EXISTENCE_PAGE_SIZE: u32 = 100;
```
aram356, line 16: 🤔 **thinking** — Worst-case eight strong list round-trips per withdrawal, on the response path.

`withdrawal_ec_ids` can yield two distinct IDs (the active EC and a different cookie EC), and each one can walk up to `EXISTENCE_MAX_PAGES` pages, so a single withdrawal can issue 2 × 4 = 8 strongly consistent list calls before the tombstone write. For a well-formed `{64hex}.{6alnum}` ID this is one page per ID in practice, so the normal case is fine — but this is synchronous work on the user-facing response path.

One easy reduction: the request only asks whether the key exists, yet it fetches pages of 100 keys.

```rust
const EXISTENCE_PAGE_SIZE: u32 = 1;
```

With an exact-match prefix, at most one key can equal the prefix, so a limit of 1 answers the question with materially less to transfer. The `EXISTENCE_MAX_PAGES` budget still protects against a short prefix producing a long listing. Worth measuring rather than taking on faith, but the current 100 does not buy anything the check uses.

crates/trusted-server-core/src/ec/kv_backend.rs
```diff
@@ -83,6 +83,18 @@ pub trait EcKvStore {
...
+    ///
+    /// A completed issuance must be visible even when [`Self::lookup`] lags.
+    /// Prefix matches are insufficient and an inconclusive check is an error.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`TrustedServerError::KvStore`] on store failure or when a
+    /// bounded check cannot determine existence. Never falls back to an
+    /// eventually consistent read.
+    fn key_exists(&self, key: &str) -> Result<bool, Report<TrustedServerError>>;
```
aram356, line 111: 🤔 **thinking** — This contract is Fastly-only and nothing enforces it.

The doc says the implementation must be exact, strongly consistent, and "Never falls back to an eventually consistent read." `FastlyEcKvStore` is currently the only production implementor — the Cloudflare, Spin, and Axum adapters wire no `EcKvStore` at all — so the guarantee holds today.

The risk is the next adapter. A Cloudflare Workers KV implementation is the obvious case: its list API is eventually consistent, so the natural implementation of this trait method would satisfy the type signature while silently violating the contract, and the failure mode is a withdrawal being dropped for a recently issued identity — exactly #1116, reintroduced on a different platform with all of this PR's tests still green (they run against `InMemoryEcKv`, which is trivially strong).

Consider making the consequence explicit in the trait doc, e.g. "An implementation whose listing is eventually consistent must not implement this by listing; doing so reintroduces #1116 — a withdrawal for a recently issued identity will be silently discarded." A shared conformance test that a new backend must pass would be stronger still, though that is reasonably a follow-up rather than this PR.

crates/trusted-server-adapter-fastly/src/ec_kv.rs
```diff
@@ -56,7 +100,7 @@ impl EcKvStore for FastlyEcKvStore {
                 return Err(
                     Report::new(err).change_context(TrustedServerError::KvStore {
                         store_name: self.store_name.clone(),
-                        message: format!("Failed to read key '{key}'"),
+                        message: format!("Failed to read key '{}'", log_id(key),),
```
aram356: ♻️ **refactor** — Stray trailing comma inside a single-argument `format!`.

`log_id(key),)` — the trailing comma is a leftover from the multi-argument form this replaced. `cargo fmt` accepts it, so it is cosmetic only, but it reads as an unfinished edit.

```suggestion
                        message: format!("Failed to read key '{}'", log_id(key)),
```

_Verified in a scratch worktree: applied alone, `cargo fmt --all -- --check` and `cargo clippy-fastly` both pass with no drift._

crates/trusted-server-adapter-fastly/src/ec_kv.rs
```diff
@@ -117,10 +173,7 @@ impl EcKvStore for FastlyEcKvStore {
             .execute()
             .change_context(TrustedServerError::KvStore {
                 store_name: self.store_name.clone(),
-                message: format!(
-                    "Failed to list keys with prefix '{}'",
-                    prefix.get(..8).unwrap_or(prefix),
-                ),
+                message: format!("Failed to list keys with prefix '{}'", log_id(prefix),),
```
aram356: ♻️ **refactor** — Same stray trailing comma as at line 103.

```suggestion
                message: format!("Failed to list keys with prefix '{}'", log_id(prefix)),
```

_Verified in a scratch worktree: applied alone (independently of the line 103 suggestion, since they sit in different diff hunks), `cargo fmt --all -- --check` and `cargo clippy-fastly` both pass with no drift._

aram356 (changes requested):
## Summary

Gates the Edge Cookie withdrawal tombstone on an exact, strongly consistent existence check, so a client-supplied `ts-ec` cookie can no longer mint a row for an identity the graph never held. The approach is sound and the new Viceroy-backed backend tests close the coverage gap the previous revision listed as unclosable. Verified locally at `12f5f1e8e`: 321 core EC tests, 13 new Fastly backend tests, `cargo fmt --check`, and `cargo clippy-fastly` all pass, and all 19 GitHub checks are green.

Requesting changes on three points: this merge silently removes `main`'s CAS-based tombstone path, `key_exists_confirmed` changed semantics for a second caller outside this PR's stated scope, and the interaction with another in-flight branch is a compile-silent hazard that would drop the new snapshot write-back.

> 2 of the inline comments below carry a one-click GitHub `suggestion`. The remaining comments describe the fix in prose because the finding spans a deletion, multiple files, or is cross-cutting and cannot be auto-applied.

## Blocking

### 🔧 wrench

- **Merge deletes `main`'s CAS tombstone path; unconditional overwrite can resurrect a deleted row** — see "Cross-cutting" below
- **`key_exists_confirmed` silently changed semantics for a second, unrelated caller** — see "Cross-cutting" below

### ❓ question

- **Cross-PR reconciliation of the withdrawal call site is a compile-silent hazard** — see "Cross-cutting" below

## Non-blocking

### 🤔 thinking / ♻️ refactor

- **Worst-case 8 strong list round-trips per withdrawal on the response path** — see inline at `crates/trusted-server-adapter-fastly/src/ec_kv.rs:16`
- **The strong-consistency guarantee is Fastly-only and unenforced** — see inline at `crates/trusted-server-core/src/ec/kv_backend.rs:96`
- **Stray trailing comma in single-argument `format!`** — see inline at `crates/trusted-server-adapter-fastly/src/ec_kv.rs:103` and `:176`

- 🔧 **Merge deletes `main`'s CAS tombstone path; unconditional overwrite can resurrect a deleted row.** This merge removes `tombstone_existing_from_snapshot`, its `MAX_CAS_RETRIES` loop, the `DisappearOnConflictEcKv` double, and six tests that existed on `main` (`git show origin/main:crates/trusted-server-core/src/ec/kv.rs`, around lines 959 and 2182-2300), replacing all of it with a check followed by an unconditional `EcKvWriteMode::Overwrite`.

  I confirmed the behavioural consequence by running it rather than reasoning about it. With a store double that deletes the row immediately after the existence check returns `true`, the withdrawal path reports `Written` and the row exists afterwards — a row that no longer existed is recreated:

  ```
  PROBE1 outcome=Written row_exists_after=true
  ```

  I do think the trade-off is defensible, and I am not asking you to restore the CAS loop. A resurrected *tombstone* denies consent and expires on `TOMBSTONE_TTL`, so it fails safe, and the unconditional write means a withdrawal can no longer lose a CAS race the way `main` could — `tombstone_existing_from_snapshot_returns_failed_after_cas_exhaustion` on `main` asserts exactly that losing case, where the row stayed live with consent granted.

  What blocks is that none of this appears in the PR description or the commit messages. A reviewer reading the diff sees ~130 lines of tested concurrency machinery disappear inside a merge commit with no statement of intent. Please state the removal and its rationale explicitly (the "withdrawal must always win over a concurrent write" argument is the right one), so the trade-off is a recorded decision rather than an artefact of the merge.

- 🔧 **`key_exists_confirmed` silently changed semantics for a second, unrelated caller** (`crates/trusted-server-core/src/ec/finalize.rs:223`, unchanged context so it carries no inline anchor).

  On `main` this helper was prefix-based:

  ```rust
  pub fn key_exists_confirmed(&self, ec_id: &str) -> Result<bool, Report<TrustedServerError>> {
      Ok(self.store.count_keys_with_prefix(ec_id, 1)? > 0)
  }
  ```

  This PR reimplements it as an exact match against strongly consistent state. That is correct and is the point of the change — but this call site, the orphan-recovery path, is not part of the PR's stated scope and its behaviour changes as a side effect.

  The change is an improvement here too: previously a longer key sharing this ID as a prefix would report `Ok(true)` and suppress a legitimate identity rotation, which is the same prefix-collision bug the PR fixes on the withdrawal path. So I am not asking you to revert it — I am asking that it be deliberate and covered.

  Please add a test asserting orphan recovery behaviour when only a *longer* key exists under the same prefix: with the exact check, `key_exists_confirmed` returns `false` and `recover_orphaned_ec` should run, where the old prefix logic would have taken the `Ok(true)` branch and skipped recovery. Mentioning the second caller in the PR description would also help, since the body currently describes this as a withdrawal-path change only.

- ❓ **Cross-PR reconciliation of the withdrawal call site is a compile-silent hazard.** Your reviewer note flags this; I verified it against the other branch's actual diff and it is worse than the note implies. That branch keeps this shape at the shared call site:

  ```rust
  if let Err(err) = graph.write_withdrawal_tombstone(kv_key) { ... }
  ```

  That still compiles against the new `Result<TombstoneOutcome, _>`. It discards `UnknownIdentity`, which is benign, but it also drops the `set_kv_snapshot` write-back this PR adds at `finalize.rs:316-331`. That write-back is what propagates the tombstone into `EcContext` so `dispatch_pull_sync` cancels — `pull_sync.rs` runs post-send and discloses the raw `ec_id` to partners, and `dispatch_pull_sync_skips_dispatch_when_ec_is_tombstoned_after_snapshot_capture` is the test covering exactly that. A merge resolved toward that side would therefore reintroduce partner disclosure of a just-withdrawn identity, with no compile error and no failing test in this PR.

  Two questions: what is the intended landing order between #901 and this PR; and can the outcome be made structurally impossible to discard — `#[must_use]` on `TombstoneOutcome` plus having `write_withdrawal_tombstone` take the context write-back as a closure, or returning a type the call site cannot ignore — so this converts into a compile error instead of a silent behaviour loss?

## CI Status

- Analyze (actions): PASS
- Analyze (rust): PASS
- cargo fmt: PASS (required)
- cargo test: PASS (required)
- CodeQL: PASS
- integration tests: PASS
- vitest: PASS

crates/trusted-server-adapter-fastly/src/ec_kv.rs
prk-Jr: Fixed in c4d569dd.

crates/trusted-server-adapter-fastly/src/ec_kv.rs
prk-Jr: Fixed in c4d569dd.

crates/trusted-server-core/src/ec/kv_backend.rs
prk-Jr: Added in 7212953a — an `# Implementing this method` section on the trait, close to your wording: an implementation whose listing or point read is eventually consistent must not answer from it, because that reintroduces #1116 on a new platform, and a backend with no strongly consistent read should return an error rather than a `false` the caller will trust.

Also noted there that the in-memory double is trivially strong, so the core tests cannot catch a backend that breaks this — a new backend has to be covered against its own platform. Agreed the shared conformance test is the stronger answer; leaving that as a follow-up, since there is no second implementor to write it against yet.

crates/trusted-server-adapter-fastly/src/ec_kv.rs
prk-Jr: Pushing back on this one — I would rather keep 100.

Two reasons. First, `limit` is a maximum, not a fetch size. Fastly SDK 0.12.1 `ListBuilder::limit`: "Set the maximum number of items included the response. The default limit is 100 items." With one matching key the response carries one key at either setting, so the normal case transfers exactly the same bytes. 100 is also the SDK's own default, so the constant costs nothing and documents the intent.

Second, limit 1 quietly rests on list ordering. "At most one key can equal the prefix" is true, but a limit-1 page returns *some* prefix match, not necessarily the exact one — that only follows if the listing is ordered, and `kv_store.rs` documents no ordering guarantee (no mention of sorted, lexicographic, or order anywhere in it). If the order is not lexicographic, limit 1 shrinks the effective scan budget from 400 keys to 4, and budget exhaustion is an error, i.e. a dropped withdrawal. That trades a bounded guarantee for an unstated assumption, in the direction of the bug this PR fixes.

On the round trips themselves: the bound is 2 x EXISTENCE_MAX_PAGES, so it is set by the page *count*, not the page size — cutting the size does not reduce it. Happy to revisit if you want the page budget itself lowered, or if there is an ordering guarantee documented somewhere I did not find.

prk-Jr:
Thanks for the review — the three body-level findings are addressed in c4d569dd (four commits on top of a4c61b5e).

## Cross-PR reconciliation with #1043 — now a compile error

Made structural rather than documented, along the lines you suggested. `write_withdrawal_tombstone` takes the write-back as a parameter:

```rust
pub fn write_withdrawal_tombstone(
    &self,
    ec_id: &str,
    record_snapshot: impl FnOnce(EcKvSnapshot),
) -> Result<TombstoneOutcome, Report<TrustedServerError>>
```

The graph now builds the snapshot itself, from the entry it actually wrote, and calls `record_snapshot` on every path out of the method including the error path. Verified by reproducing #1043's exact call shape in a scratch tree:

```
error[E0061]: this method takes 2 arguments but 1 argument was supplied
   --> crates/trusted-server-core/src/ec/finalize.rs:321:37
    |
321 |             if let Err(err) = graph.write_withdrawal_tombstone(ec_id) {
    |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^------- argument #2 is missing
```

`TombstoneOutcome` is also `#[must_use]` now. On its own that would not have caught #1043's shape — `if let Err(..)` counts as a use of the `Result` — but it did catch two existing tests discarding the outcome after `expect`, which now assert `Written`.

A side effect worth flagging: **#901 is caught too, and it is the more dangerous of the two.** Its `write_withdrawal_tombstone` returns `Result<(), _>` with no existence gate at all — the pre-fix shape — so landing it over this branch would revert the gate entirely, not just drop the write-back. It now fails to compile against both the signature and the call site.

On landing order: I have not fixed one, and it is worth deciding explicitly rather than by merge accident. The relevant facts are that #1043 bases on `main` and renamed `withdrawal_ec_ids` to `withdrawal_kv_keys`, so that hunk conflicts textually regardless and a human resolves it either way — the compile error is the backstop if it is resolved toward #1043's side. #901 stacks on `fix/no-op-kv-reads` rather than `main`. Since both stale shapes now fail to build against this one, no ordering silently loses the write-back; happy to sequence them however you prefer.

## `key_exists_confirmed`'s second caller — covered

Agreed on all counts, including that the change is an improvement at that call site and should not be reverted. Added `finalize_rotates_when_only_a_longer_key_shares_the_orphan_prefix` (e99a127f): seeding only a longer neighbouring key leaves the orphan proven absent, so `recover_orphaned_ec` runs and the identity rotates, while the neighbour is left untouched.

I checked that the test earns its keep rather than just passing. Reverting the helper to `count_keys_with_prefix` fails the rotation assertion on its own, with the existence pre-assert stripped out:

```
assertion `left != right` failed: a proven-absent orphan must rotate even with a prefix neighbour present
```

## CAS tombstone path removal

Your reading of the trade-off is the one I would defend, and you are right that it is currently invisible. Recording it here, and I will add it to the PR description.

The merge drops `tombstone_existing_from_snapshot`, its `MAX_CAS_RETRIES` loop, the `DisappearOnConflictEcKv` double and six tests, replacing them with a strong existence check followed by an unconditional overwrite. The trade is deliberate: withdrawal must always win over a concurrent write, and main's CAS loop could lose that race outright — `tombstone_existing_from_snapshot_returns_failed_after_cas_exhaustion` asserts exactly that losing case, leaving a live row with consent granted. The cost is the resurrection you reproduced: a row deleted between check and write comes back as a tombstone. That denies consent, expires on `TOMBSTONE_TTL`, and cannot mint an identity, so it fails safe.

## Validation

`cargo fmt`, all eight clippy targets, `test-fastly` under Viceroy (all 13 `ec_kv::tests`, including cursor traversal and page-budget exhaustion), `test-axum`, `test-cloudflare`, `test-spin`, cross-adapter parity, and 2,467 core lib tests — all passing. Each of the four commits was checked independently for `fmt` and `--all-targets` build, so no intermediate state is broken.

aram356 (approved):

crates/trusted-server-core/src/ec/finalize.rs
```diff
@@ -314,6 +283,84 @@ pub fn clear_ec_on_response(settings: &Settings, response: &mut Response<EdgeBod
...
+    // The identity-graph tombstone is the authoritative withdrawal marker
+    // for subsequent EC behavior.
+    if let Some(graph) = kv {
+        apply_withdrawal_tombstones(&ids_to_withdraw, |ec_id| {
+            // The graph hands back the post-withdrawal snapshot rather than
+            // leaving the caller to rebuild it, so post-send work that reads
+            // the context — pull sync discloses the raw EC ID to partners —
+            // sees the tombstone that was just written. Only the active ID has
+            // a snapshot in the context to correct.
+            let outcome = graph.write_withdrawal_tombstone(ec_id, |snapshot| {
```
aram356, line 321: ❓ **question** — Cross-PR reconciliation of the withdrawal call site (raised at the review-body level on the previous pass; recording it here so it has a thread that can be tracked and resolved).

**Original concern:** another in-flight branch kept a two-arm `if let Err(err) = graph.write_withdrawal_tombstone(kv_key)` at this shared call site. That still compiled against the then-new `Result<TombstoneOutcome, _>`, so a merge resolved toward that side would have silently dropped the `set_kv_snapshot` write-back — and that write-back is what stops post-send `dispatch_pull_sync` from disclosing a just-withdrawn `ec_id` to partners.

**Resolved by `8c2f161b3`.** Taking `record_snapshot` as a required parameter converts that from a silent behaviour loss into a build failure. I verified this rather than taking it on trust — injecting the old single-argument call shape gives:

```
error[E0061]: this method takes 2 arguments but 1 argument was supplied
    --> crates/trusted-server-core/src/ec/kv.rs:2562:25
     |
2562 |     if let Err(_e) = kv.write_withdrawal_tombstone(&ec_id) {
     |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^-------- argument #2 is missing
```

Worth noting the structure is stronger than `#[must_use]` alone: all fallible work sits in `tombstone_held_identity`, so `record_snapshot` is invoked unconditionally on every path including the error path — no early return can skip it.

Still worth confirming before merge: **#901 returns `Result<(), _>` with no existence gate at all**, so landing it over this branch would revert the fix outright rather than merely lose the write-back — and unlike the call-shape case above, that is not caught by the new compile error. What is the intended landing order between #901 and this PR?

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -132,6 +132,16 @@ impl fmt::Debug for KvIdentityGraph {
     }
 }
 
+/// Result of [`KvIdentityGraph::write_withdrawal_tombstone`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[must_use]
```
aram356, line 137: 🔧 **wrench** — Removal of `main`'s CAS tombstone path (raised at the review-body level on the previous pass; recording it here so it has a thread that can be tracked and resolved).

**Original concern:** this branch deletes `tombstone_existing_from_snapshot`, its `MAX_CAS_RETRIES` loop, the `DisappearOnConflictEcKv` double and six tests, replacing them with a strong existence check plus an unconditional `EcKvWriteMode::Overwrite` — all inside a merge commit, with no statement of intent. A reviewer saw ~130 lines of tested concurrency machinery vanish with no recorded rationale.

I confirmed the behavioural consequence by running it rather than reasoning about it: with a store double that deletes the row immediately after the existence check returns `true`, the path reports `Written` and the row exists afterwards (`row_exists_after=true`) — a deleted row is recreated.

**Resolved.** The PR body now carries a "Removing main's CAS tombstone path" section making it a recorded decision, and the argument is the right one: withdrawal must win over a concurrent write, and `main`'s CAS loop could lose that race outright — `tombstone_existing_from_snapshot_returns_failed_after_cas_exhaustion` asserted exactly that losing case, leaving the row live with consent granted. The residual TOCTOU fails safe by comparison: a resurrected tombstone denies consent, expires on `TOMBSTONE_TTL`, and cannot mint an identity.

No change requested — flagging that the trade-off is now documented rather than implicit.

crates/trusted-server-core/src/ec/finalize.rs
```diff
@@ -940,6 +1133,73 @@ mod tests {
         );
     }
 
+    #[test]
+    fn finalize_rotates_when_only_a_longer_key_shares_the_orphan_prefix() {
```
aram356, line 1139: 🔧 **wrench** — `key_exists_confirmed` semantics changed for a second caller (raised at the review-body level on the previous pass because the natural anchor, `finalize.rs:223`, was unchanged context that GitHub rejected; recording it here so it has a thread).

**Original concern:** on `main` this helper was `count_keys_with_prefix(ec_id, 1) > 0`. Reimplementing it as an exact match also changed behaviour at the orphan-recovery call site, which is outside the PR's stated withdrawal-path scope, and nothing covered it. The change is an improvement — a longer key sharing the orphan's ID as a prefix previously reported it as still held and suppressed a legitimate rotation, the same collision this PR closes on the withdrawal path — but it was silent.

**Resolved by `e99a127fb`,** which is this test. I verified it genuinely pins the behaviour rather than passing vacuously: reverting `key_exists_confirmed` to `main`'s prefix-count form makes it fail (panic, exit 134), and it passes again once restored.

crates/trusted-server-core/src/ec/finalize.rs
prk-Jr: **This PR lands first**, and it turns out the stack forces that rather than it being a judgement call — which is a better answer than the one I was going to give.

`#901` bases on `fix/no-op-kv-reads` (#900, still OPEN, `CHANGES_REQUESTED`), which bases on `fix/kv-eid-request-snapshot-ec-ttl` (#885, MERGED). So #901 sits behind one unmerged ancestor carrying requested changes, and #901 itself has no review decision yet. It cannot reach `main` ahead of this branch. #1043 is `CHANGES_REQUESTED` and last touched Sep 4, so the same holds there.

**Your risk statement is right, and worth stating one notch more precisely.** #901 is not entirely gate-free — `tombstone_unproven_missing` does call `key_exists_confirmed` before writing. What it lacks is a gate *in the function*, and its two other production callers have none:

- `admin.rs:1118` — `kv.write_withdrawal_tombstone(&ec_id)`
- `pull_sync.rs:1079` — `.write_withdrawal_tombstone(&ec_id)`

So the accurate failure mode for landing #901's body over this branch is: the admin and post-send pull-sync withdrawal paths can mint a tombstone row for an `ec_id` the store never held. On this branch all three call sites are covered because the check sits inside `tombstone_held_identity`.

And I agree the new compile error does not save us here. `E0061` fires on call *shape* only. A resolution that takes #901's function together with its own call sites compiles clean with the gate gone — that is a human-resolution requirement, not something the type system can carry. Which is the real argument for this order: with the gate on `main` first, any later resolution that drops it is a visible regression against `main` in a diff, rather than a gap that was never there to begin with.

**One thing worth flagging, because it cuts the other way: this PR does not subsume #901.** Its idempotence property is orthogonal and still needed. On this branch a repeated withdrawal re-passes the existence check (the tombstone row exists), then does another unconditional `Overwrite` at a full `TOMBSTONE_TTL`. I measured it rather than reading it — a store double recording the TTL of every `insert`, seeded live, withdrawn twice:

```
SCRATCH first=Written second=Written ttl_writes=[86400s, 86400s]
```

Two writes, each a fresh 86400s. So a withdrawal burst keeps pushing the tombstone's expiry out, which is exactly what #901's title is about and what its `tombstone_existing_from_snapshot_repeated_request_preserves_first_write` and `tombstone_existing_from_repeated_stale_miss_preserves_first_write` pin. There is no completion-marker machinery on this branch at all — `withdrawal_marker_exists` / `record_withdrawal_completion` do not exist here.

That suggests the post-rebase shape for #901, if it is useful to write it down now:

- **Keep** the marker machinery (`write_withdrawal_marker`, `withdrawal_marker_exists`, `record_withdrawal_completion`, plus the clear-on-revive and clear-on-delete paths). This is #901's actual contribution and nothing here replaces it.
- **Drop** `tombstone_unproven_missing` and its own `key_exists_confirmed` call as redundant — the method it calls is now gated, and `key_exists_confirmed` here is `self.store.key_exists(ec_id)` (exact match) rather than #901's `count_keys_with_prefix(ec_id, 1) > 0`, so the prefix-collision case is closed too.
- **Drop** the CAS machinery, per the trade-off recorded in the other thread.

Net effect: #901 becomes a marker check in front of the gated write, which is a much smaller diff than its current form.

For #1043 the mechanical picture is unchanged from your note — `finalize.rs:84` is still the single-argument `if let Err(err) = graph.write_withdrawal_tombstone(kv_key)`, so `E0061` catches it, and the `withdrawal_ec_ids` → `withdrawal_kv_keys` rename means that hunk conflicts textually regardless.

All 18 checks are green on `f99ac0ae`, including both integration suites and the Fastly EC lifecycle job.

PR #1074, opened 2026-08-26
Add request phase timing: Server-Timing subtimings and access telemetry

jevansnyc:
Closes #1068. Implements the design in #1069 (`docs/superpowers/specs/2026-08-24-request-phase-timing-design.md`); the implementation plan and the spec ride in the branch.

## What this adds

- `RequestTimings` (core): always-on per-request phase collection; `try_lock`-only, saturating, infallible by construction.
- `Server-Timing` header (`ts-total`, `ts-appbuild`, `ts-filter`, `ts-geo`, `ts-kv`, `ts-origin`, `ts-template-cache`) emitted at the send freeze point immediately before `into_parts()`, gated by `observability.server_timing_enabled` and restricted to conclusively private (`private`/`no-store`) responses so no shared cache can replay timings.
- Phase spans: app build, pre-route filters (vendor-masked), geo, origin fetch, template-cache lookup, and EC KV via a `TimedKvStore` decorator implementing both `PlatformKvStore` and `EcKvStore` (latency-only; reads no payloads). Pull-sync stores are explicitly untimed.
- Geo dedupe rider: one hostcall per request instead of two, via a `GeoLookupState` response extension (`NotAttempted` / `Attempted` / `Resolved`), 401 rule preserved.
- Body-phase capture: stream duration via a counting writer (`DeliveryResult::{Complete, Partial, Error}`), auction-wait with explicit placement (`in_stream` at the seam, `pre_header` on buffered paths), response bytes, and a post-body `request_elapsed_ms` that excludes pull-sync and telemetry.
- Access telemetry: an `AccessTelemetrySnapshot` built at the freeze point when `tinybird.access_enabled` is set, route templates emitted only from an operator-configured section allowlist (`observability.route_sections`, default empty, so no request-derived byte reaches the row; adversarially tested with usernames, slugs, tokens, EC ids, emails, and search terms), and a confirmed-delivery Tinybird sink (bounded await, streamed response, 2xx-validated, sampled by `tinybird.access_sample_rate`) that runs after client delivery and after pull-sync.
- `access_logs_raw` schema aligned column-for-column with the row producer; sorting key `(toDate(event_ts), service_id, publisher_domain, env, route_class, pop, status)`; there is no `event_date` column (Tinybird's Events API requires a JSONPath on every column, which a derived-default column cannot carry).
- Axum adapter emits the header via an outer service wrapper (router-generated 404/405 covered); operator docs in `docs/guide/configuration.md`.

## Review process

[...]

PR #1121, opened 2026-09-04
Improve GPT auction diagnostics observability

### LINKED ISSUE #1081: Improvements to TS_CONSOLE for ad observability

ChristianPavilonis:
## Summary

- Link every creative badge to the same stable `Ad #N` in the TS Console panel and export.
- Add SSAT, SPA Trusted Server, client-side, and genuinely competing-auction classification with winning bidder and bucketed price facts.
- Surface session-gated server auction timings alongside clearly named GAM callback timings, and clean up `1×1` and size terminology.

This is stacked on #1076, which is stacked on #1074. Retarget this PR to `main` after those dependencies merge.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/publisher.rs` | Serialize initial and SPA auction timing facts, preserve generation safety, and gate SPA diagnostics on the active console session. |
| `crates/trusted-server-js/lib/src/integrations/gpt/` | Carry immutable auction facts through the initial scheduler and SPA page-bids path. |
| `crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/` | Validate, retain, classify, clone, and present auction diagnostics with stable slot numbering. |
| `crates/trusted-server-js/lib/test/` and browser integration tests | Cover timing propagation, session gating, winner validation, classification, stable numbering, terminology, and export isolation. |
| `docs/guide/integrations/gpt-diagnostics.md` | Document auction labels, timing origins, winner/privacy boundaries, size behavior, and browser callback semantics. |

## Closes

Closes #1081

## Test plan

- [x] `cargo test-cloudflare`
- [x] JS tests: `cd crates/trusted-server-js/lib && npx vitest run` (898 passed, no type errors)
- [x] JS lint/build: `npm run lint && npm run build`
- [x] Docs format/lint/build: `cd docs && npm run format && npm run lint && npm run build`
- [x] Other: focused Next.js Playwright GPT diagnostics suite (3 passed)

## Checklist

- [x] Logging conventions preserved; no `println!` / `eprintln!` added
- [x] New code has tests


crates/trusted-server-core/src/publisher.rs
```diff
@@ -6600,12 +6744,29 @@ pub async fn handle_page_bids(
                 provider_responses: None,
                 services,
             };
-            match auction
+            let timing_started = web_time::Instant::now();
+            let auction_dispatched_ms = elapsed_millis(&timing_started);
```
prk-Jr, line 6748: 🤔 **thinking** — This field is a constant `0` on every SPA cycle: the clock starts on the line above and is read immediately, so `auction_dispatched_ms` can only ever be sub-millisecond. `auction_wait_ms` (`resolved - dispatched`) then becomes an exact duplicate of `auction_resolved_ms`, and the panel does render the zero — `formatMilliseconds(0)` returns `"0 ms"`, so every SPA cycle shows `SPA auction T0 → auction dispatched 0 ms`.

The anchor also excludes everything the handler already did before this line (CSRF gate, consent gate, slot matching, EC/KV lookups) — which is precisely the pre-dispatch cost the initial-document path *does* surface, because there `auctionDispatchedMs` is a real T0 offset. So the two paths look comparable in the UI while measuring different things, and a slow pre-dispatch phase is invisible on the SPA side.

The request already carries a T0-anchored `RequestTimings` in its extensions (inserted at `crates/trusted-server-adapter-fastly/src/main.rs:216` and `crates/trusted-server-adapter-axum/src/timing.rs:87`), so this path could use the same instrument it uses on the document path:

```rust
let timings = req
    .extensions()
    .get::<RequestTimings>()
    .cloned()
    .unwrap_or_default();
// … around the auction:
timings.mark_auction_dispatched(auction_id);
let result = auction.orchestrator.run_auction(&auction_request, &auction_context).await;
timings.record_auction_wait(AuctionWaitPlacement::PreHeader, wait_started.elapsed());
timings.mark_auction_resolved();
// after the bid map is built:
timings.mark_auction_committed();
let auction_diagnostics = gpt_diagnostics
    .browser_session_active()
    .then(|| BrowserAuctionDiagnostics::from_request_timings(&timings))
    .flatten();
```

That gives both paths one clock semantics, removes the parallel timing mechanism, and — because these are the same marks the Server-Timing header and auction telemetry read — makes the console numbers reconcilable with the logs instead of existing only in the JSON response. If you'd rather keep the local clock, then dropping `auctionDispatchedMs` (and with it the redundant `auctionWaitMs`) from the SPA payload is more honest than shipping a measured-looking `0`.

Apply manually — spans the timing setup, the `Ok(result)` arm, and the commit patch below, so it can't be expressed as one contiguous suggestion.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -6600,12 +6744,29 @@ pub async fn handle_page_bids(
...
+                    // the edge, so they must not fabricate dispatch timing evidence.
+                    if gpt_diagnostics.browser_session_active() {
+                        auction_diagnostics = Some(BrowserAuctionDiagnostics {
+                            auction_dispatched_ms: Some(auction_dispatched_ms),
+                            auction_resolved_ms: Some(auction_resolved_ms),
+                            auction_committed_ms: None,
+                            auction_wait_ms: Some(
+                                auction_resolved_ms.saturating_sub(auction_dispatched_ms),
+                            ),
+                            auction_wait_placement: Some("pre_header"),
```
prk-Jr, line 6767: ♻️ **refactor** — The wire values for this field now have two sources of truth: `BrowserAuctionDiagnostics::from_request_timings` maps the enum (`publisher.rs:3140-3148`), and this site hardcodes the string. The TS side pins them as a closed union (`'pre_header' | 'in_stream'` in `core/types.ts`) and `normalizedServerAuctionTimings` drops anything else, so a typo here degrades silently to `placement unknown` in the panel rather than failing anywhere.

**Proposed fix** (apply manually — the helper definition and this call site are in separate hunks):

```rust
const fn placement_wire(placement: AuctionWaitPlacement) -> &'static str {
    match placement {
        AuctionWaitPlacement::PreHeader => "pre_header",
        AuctionWaitPlacement::InStream => "in_stream",
    }
}
```

Then `from_request_timings` becomes `snapshot.auction_wait_placement.map(placement_wire)` and this line becomes `auction_wait_placement: Some(placement_wire(AuctionWaitPlacement::PreHeader))`, which also documents at the call site *which* placement the page-bids wait is being claimed as.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -3146,11 +3195,34 @@ impl AdBidsState {
...
+            .lock()
+            .expect("should lock auction diagnostics")
+            .clone();
+        let bids_script =
+            build_bids_script_with_diagnostics(&bid_map, auction_diagnostics.as_ref());
         *self.script.lock().expect("should lock bid script") = Some(bids_script);
         *self.bids.lock().expect("should lock bid map") = bid_map;
     }
 
+    fn set_auction_diagnostics(&self, timings: &RequestTimings) {
```
prk-Jr, line 3209: ♻️ **refactor** — This writes `*self.script.lock() = Some(build_bids_script_with_diagnostics(...))`, discarding whatever the cell already held. That is safe today only because `prepend_auction_debug_comment` (`publisher.rs:4138`) runs *after* it in `collect_stream_auction`; swap those two statements and the `ts-debug` comment silently disappears from the inline bids script, with no test covering the pair. `prepend_to_script` keeps `debug_prefix` for exactly this reason, but only `build_seam_script` re-applies it.

Cheapest fix is to record the ordering requirement next to the new write:

```suggestion
    /// Attach the server auction facts to the rendered bid script.
    ///
    /// Rebuilds the script cell from scratch, so it must run **before**
    /// [`Self::prepend_to_script`]: a debug comment already written into the
    /// script cell would be discarded here, and only the shared-template seam
    /// re-applies the retained `debug_prefix`.
    fn set_auction_diagnostics(&self, timings: &RequestTimings) {
```

If you'd rather have it enforced than documented, re-apply the retained prefix the way `build_seam_script` already does, so the write cannot lose it regardless of call order.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts
```diff
@@ -229,6 +234,36 @@ function cycleFacts(cycle: GptDiagnosticsRequestCycle): string[] {
...
+  if (cycle.auctionType) facts.push(`Auction type: ${auctionTypeLabel(cycle.auctionType)}`);
+  if (cycle.auctionWinner) {
+    facts.push(`Winning bidder: ${cycle.auctionWinner.bidder}`);
+    facts.push(`Winning bid price bucket: ${cycle.auctionWinner.priceBucket}`);
+  }
+  if (cycle.serverAuctionTimings) {
+    const timingOrigin =
+      cycle.serverAuctionTimingOrigin ??
+      (cycle.auctionType === 'trusted_server' ? 'spa_auction' : 'navigation');
+    const timingAnchor = timingOrigin === 'spa_auction' ? 'SPA auction T0' : 'Navigation T0';
```
prk-Jr, line 246: 🤔 **thinking** — `RequestTimings::t0` is stamped when the edge starts processing the request, not at the browser's `navigationStart`. Labeling it `Navigation T0` invites the operator to line these offsets up against browser Navigation Timing — the exact cross-clock subtraction the new documentation forbids two paragraphs earlier ("Server auction timing and browser GPT timing use separate clocks and are never subtracted from each other").

`Edge request T0` (or `Server request T0`) names what is actually being measured and keeps the SPA/initial distinction intact:

```ts
const timingAnchor = timingOrigin === 'spa_auction' ? 'SPA auction T0' : 'Edge request T0';
```

Apply manually — `test/integrations/gpt_diagnostics/overlay.test.ts` asserts the rendered `Navigation T0 → …` strings, so the rename has to land with those assertions.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
```diff
@@ -1025,6 +1118,18 @@ export class GptDiagnosticsStore {
...
 
+  private auctionType(
+    intent: PendingRequestIntent | undefined,
+    trustedServerEvidence: PendingSourceEvidence | undefined
+  ): GptDiagnosticsAuctionType | undefined {
+    const hasTrustedServerAuction = intent?.sources.has('trusted_server_direct') === true;
+    const hasClientSideAuction = intent?.sources.has('prebid_refresh') === true;
+    if (hasTrustedServerAuction && hasClientSideAuction) return 'competing';
+    if (hasClientSideAuction) return 'client_side';
+    if (hasTrustedServerAuction) return trustedServerEvidence?.auctionType ?? 'ssat';
```
prk-Jr, line 1129: 🤔 **thinking** — This is the one place the console infers rather than reports. Any direct evidence whose `auctionFacts` failed normalization is labelled `SSAT`, including a SPA request — your own test pins the behaviour (`auctionType: 'invalid'` → `'ssat'`), and the same fallback fires when the facts object is missing entirely. The panel then makes a positive claim about a path that was never observed, in a document that opens with "reports positive observations, not inferred ownership".

The producer side is already almost there: `diagnosticsAuctionFacts` always emits `auctionType` when `generation > 0`, and only returns `undefined` for the generation-0 case with no winner and no diagnostics. If it emitted `{ auctionType: 'ssat' }` for that case too, this line could be `return trustedServerEvidence?.auctionType;` and a malformed payload would stay unlabelled instead of being relabelled.

Apply manually — touches `gpt/index.ts`, this file, and the store/ad_init expectations together.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/presentation_helpers.ts
```diff
@@ -1,10 +1,29 @@
-import type { Size } from '../../core/types';
+import type { GptDiagnosticsAuctionType, Size } from '../../core/types';
 
 /** Formats CSS sizes consistently across diagnostics presentation surfaces. */
 export function formatSizes(sizes: ReadonlyArray<Size>): string {
   return sizes.map((size) => `${size[0]}×${size[1]}`).join(', ');
 }
 
+/** Hide GPT's ubiquitous 1×1 placeholder from presentation while retaining it in exports. */
+export function displayableGptFillSize(size: Size | undefined): Size | undefined {
```
prk-Jr, line 9: 🤔 **thinking** — Returning `undefined` for `1x1` means the badge and panel omit the `Fill` line entirely, which is indistinguishable from a cycle where `slotRenderEnded` carried no size at all. The export keeps the truth (and the browser spec now asserts that), but the console is the first place an operator looks, and "GPT told us 1x1" and "GPT told us nothing" are different diagnoses — the first is a live flexible/APS creative, the second is a missing observation.

Labeling rather than hiding keeps the noise down without erasing the fact, e.g. render `Fill 1x1 (GPT placeholder)` and leave the helper as a presentation predicate:

```ts
/** GPT's ubiquitous 1x1 fill size carries no layout information on its own. */
export function isGptFillSizePlaceholder(size: Size | undefined): boolean {
  return size?.[0] === 1 && size[1] === 1;
}
```

Apply manually — the badge and overlay call sites plus their assertions move with it.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -1084,13 +1110,26 @@ export function installTsAdInit(): void {
...
           const requestedSlotSizes = ts.gptSlotHandoffs?.[slotDivId2]?.formats;
           const opportunity = trustedServerOpportunity(bid);
-          ts.gptDiagnosticsRecorder?.recordTrustedServerOpportunity(
-            gptSlot,
-            slot.id,
-            opportunity,
-            bid.hb_auction_id,
-            requestedSlotSizes
-          );
+          const auctionFacts = diagnosticsAuctionFacts(generation, auctionDiagnostics, bid);
```
prk-Jr, line 1113: ⛏ **nitpick** — The `if (auctionFacts) … else …` branch repeats five arguments to make the sixth conditional, but `auctionFacts` is an optional parameter, so passing it through directly is equivalent:

```ts
ts.gptDiagnosticsRecorder?.recordTrustedServerOpportunity(
  gptSlot,
  slot.id,
  opportunity,
  bid.hb_auction_id,
  requestedSlotSizes,
  diagnosticsAuctionFacts(generation, auctionDiagnostics, bid)
);
```

I verified this in a scratch worktree rather than guessing: `npx tsc --noEmit` reports no new errors, but 8 assertions in `test/integrations/gpt/ad_init.test.ts` fail, because `toHaveBeenCalledWith` compares argument-list length and a trailing explicit `undefined` no longer matches a 5-argument expectation. So the branch currently exists only to preserve the call *arity* those assertions pin.

Apply manually — worth relaxing the assertions (they care about the facts payload, not the argument count) rather than keeping the duplicated call, but it needs the test file, so it can't be a one-click suggestion.

crates/trusted-server-integration-tests/browser/tests/nextjs/gpt-diagnostics.spec.ts
```diff
@@ -220,7 +220,7 @@ test.describe("GPT runtime diagnostics", () => {
         );
         await emit(page, "slotRenderEnded", "gpt-diagnostics-slot-primary", {
             isEmpty: false,
-            size: [300, 250],
+            size: [1, 1],
```
prk-Jr, line 223: ⛏ **nitpick** — This was the only place the browser suite asserted an exported GPT fill size, and it changed from `[300, 250]` to `[1, 1]` rather than gaining a case. The new "`1x1` stays in the JSON evidence" behaviour deserves end-to-end coverage, but so does the ordinary size, which is now only covered in vitest (`badges.test.ts`). Emitting the second cycle with a real size — or adding one `1x1` slot alongside the existing `[300, 250]` one — keeps both paths proven against a real GPT surface.

prk-Jr (commented):
## Summary

Solid, well-tested addition: stable `Ad #N` identity, auction classification, and server auction timings all land behind the existing activation gate, with a normalization boundary and tests for the malformed cases. No correctness, security, or WASM-compatibility problem found; all 14 CI checks pass. The findings below are all non-blocking — the substantive ones are about the SPA timing anchor and label accuracy in a tool whose value is precise facts.

> 1 of the inline comments below carries a one-click GitHub `suggestion` — use **Commit suggestion** to apply it as a commit on the PR branch. The remaining comments describe the fix in prose because the change touches test files, spans more than one hunk, or is a design choice rather than a mechanical edit.

## Non-blocking

### 🤔 thinking

- **SPA `auctionDispatchedMs` is always `0` by construction** — see inline at `crates/trusted-server-core/src/publisher.rs:6748`
- **`Navigation T0` overstates the anchor** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts:246`
- **The store infers `ssat` when auction facts are absent or malformed** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts:1129`
- **`1x1` suppression conflates "GPT reported 1x1" with "GPT reported nothing"** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/presentation_helpers.ts:9`

### ♻️ refactor

- **`set_auction_diagnostics` rebuilds the script cell and drops a debug prefix** — see inline at `crates/trusted-server-core/src/publisher.rs:3209` (suggestion)
- **Placement wire string has two sources of truth** — see inline at `crates/trusted-server-core/src/publisher.rs:6767`

### ⛏ nitpick

- **Duplicated 5-arg / 6-arg recorder call** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt/index.ts:1113`
- **Browser spec traded away its only non-`1x1` fill-size case** — see inline at `crates/trusted-server-integration-tests/browser/tests/nextjs/gpt-diagnostics.spec.ts:223`

### 👍 praise

- **Stable `Ad #N` on re-entry, and fail-closed dispatch gating** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts:1201`

- 📝 **Stacked base** — this targets `spec/auction-timeline-offsets` (stacked on #1076 → #1074). Retarget to `main` after those land, as the description says. Nothing in the diff depends on that ordering beyond the base itself.
- 📝 **Coverage of the server write path is complete** — `write_bids_to_state` has exactly two production call sites (`collect_non_html_auction`, `collect_stream_auction`) and both are now paired with `set_auction_diagnostics`, so there is no document path that commits bids without the timing facts. Verified by grep rather than assumed.

## CI Status

- integration tests: PASS
- cargo test: PASS
- cargo fmt: PASS
- vitest: PASS
- format-typescript: PASS
- format-docs: PASS

`gh pr checks --required` returned no names for this base, so none of the above are annotated as branch-protection-required; all of them are gates CLAUDE.md treats as PR gates, and all pass.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
```diff
@@ -1093,13 +1198,15 @@ export class GptDiagnosticsStore {
       }
     }
 
-    const runtimeSlotNumber = this.nextRuntimeSlotNumber;
-    this.nextRuntimeSlotNumber += 1;
+    const runtimeSlotNumber = existingNumber ?? this.nextRuntimeSlotNumber;
```
prk-Jr, line 1201: 👍 **praise** — Reusing the `WeakMap` number on re-entry is what finally makes the documented "`Ad #N` identity and request numbers remain stable and monotonic for the console lifetime" claim true, and it composes correctly with the surrounding bookkeeping: the eviction branch above is the only `slots.delete` site and it also splices `slotOrder` and shifts `slotActivityOrder`, so the reused number can't produce a duplicate `slotOrder` entry or clobber a live record. `requestNumbers` being a `WeakMap` too means the cycle numbering survives the same round trip.

Two other things worth calling out in this PR: gating the page-bids timings on the `Ok(result)` arm so a failure before any provider dispatch can't fabricate dispatch evidence (with a test that asserts exactly that), and the bounded normalization boundary in `normalizedServerAuctionTimings` / `normalizedAuctionWinner` — including the price-bucket shape check — backed by a malformed-input test.

prk-Jr (approved):
## Verdict: `APPROVE`

Follow-up to [the detailed review above](https://github.com/IABTechLab/trusted-server/pull/1121#pullrequestreview-5129685073), correcting its verdict: none of its findings are blocking (no 🔧 wrench, no ❓ question), and all 14 CI checks pass, so this should have been submitted as an approval rather than a comment.

Every inline comment there stands as written and remains worth reading — 4 🤔 thinking, 2 ♻️ refactor (one as a one-click `suggestion`), 2 ⛏ nitpick, 1 👍 praise — but each is a merge-can-proceed observation. The two most substantive, if you want to pick any of them up here rather than in a follow-up:

- SPA `auctionDispatchedMs` is a structural `0` (the clock is read on the line after it starts), which also makes `auctionWaitMs` a duplicate of `auctionResolvedMs` — `crates/trusted-server-core/src/publisher.rs:6748`.
- `Navigation T0` in the panel is the edge request-receipt offset, not the browser's `navigationStart` — `crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts:246`.

Approved on `ffc0438dd5f872e68c3ccad36d5c35bde47101d2`.

aram356:
@ChristianPavilonis please assign issue to this PR

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -1424,6 +1464,7 @@ export function installSpaAuctionHook(): void {
       if (inflight !== controller) return;
       ts.adSlots = data.slots;
       ts.bids = data.bids;
+      ts.auctionDiagnostics = data.auctionDiagnostics;
```
aram356, line 1458: 🔧 **wrench** — Stale navigation-T0 timings get relabelled as SPA-auction timings.

`onNavigate` increments `ts.navGeneration` at line 1440, **before** the page-bids fetch, but this line is the only place `ts.auctionDiagnostics` is replaced — and it runs only on the success path. Four early returns skip it:

- line 1458 — page-bids fetch returned no data
- lines 1460 and 1464 — a newer navigation superseded this one
- line 1483 — the `catch` for a network/abort failure

In that window `navGeneration > 0` while `ts.auctionDiagnostics` still holds the **initial document's navigation-T0** offsets. `diagnosticsAuctionFacts` then computes `isSpaAuction = generation > 0` → `auctionType: 'trusted_server'`, and `store.ts:679-684` derives `serverAuctionTimingOrigin: 'spa_auction'`. The panel renders `SPA auction T0 → auction resolved 840 ms` for offsets actually anchored to navigation T0 — wrong by the whole navigation prelude, and wrong in the direction that makes the server auction look slower than it was.

Reproduced against this head in a scratch worktree — SSAT bids plus navigation-T0 timings on the initial document, then one SPA navigation whose page-bids fetch rejects:

```
navGeneration = 1
auctionDiagnostics = {"auctionDispatchedMs":12,"auctionResolvedMs":840}
origin = spa_auction
```

This also undercuts the guarantee the PR adds to `docs/guide/integrations/gpt-diagnostics.md`: "Diagnostics retain that timing origin separately from the aggregate auction classification." The origin is retained separately from `auctionType`, but both are still derived from the same client-side `navGeneration` guess rather than from which endpoint actually produced the timings.

**Proposed fix** (apply manually — the right line is 1440, which is outside any RIGHT-side diff hunk, so this can't be a one-click `suggestion`). In `onNavigate`, immediately after the generation bump:

```ts
    ts.navGeneration = (ts.navGeneration ?? 0) + 1;
    // The previous route's server timings are anchored to a clock this route
    // never used. Drop them now so a failed or diagnostics-free page-bids
    // response cannot relabel them under this navigation's generation.
    ts.auctionDiagnostics = undefined;
```

Verified in a scratch worktree at this head: with this change the repro above flips to failing (the stale timings are cleared) and the full JS suite still passes — 900 passed, 0 failed.

A sturdier alternative, if you would rather not depend on client-side inference at all: have the server stamp the origin into `BrowserAuctionDiagnostics` (e.g. `timingOrigin: "navigation" | "spa_auction"`), since the server already knows which endpoint produced the timings. The client would then carry it through instead of deriving it from `navGeneration`.

crates/trusted-server-core/src/publisher.rs
aram356, line 6759: 🤔 **thinking** — `auctionDispatchedMs` is structurally always `0` here, and `auctionWaitMs` is therefore always exactly `auctionResolvedMs`.

`timing_started` is created on line 6747 and read on 6748, so `elapsed_millis` measures the gap between two adjacent statements. Unlike the navigation path — where `mark_auction_dispatched` is a genuine T0-anchored offset — there is no dispatch latency being captured on this path at all.

That makes two of the five server facts on the SPA path carry no information: a constant `0`, and a wait that duplicates the resolve offset. The new test asserts only `.is_some()` on each field, never a value, so nothing catches it:

```rust
assert!(
    auction_diagnostics
        .get(field)
        .and_then(serde_json::Value::as_u64)
        .is_some(),
    "page-bids diagnostics should include {field}"
);
```

Two ways out, either is fine:

- Anchor `timing_started` at handler entry (or at the earliest point the request is known to be an auction) so the dispatch offset measures real pre-auction work, matching the navigation path's meaning.
- Or omit `auctionDispatchedMs` and `auctionWaitMs` for page-bids and emit only `auctionResolvedMs` / `auctionCommittedMs`, so the console never shows a field that is structurally zero.

If you keep the current shape, consider asserting the actual expected values in the test rather than presence, so the constant-zero property is at least intentional and pinned.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -6600,12 +6744,29 @@ pub async fn handle_page_bids(
...
+                    // provider outcome. Failures can occur before any request leaves
+                    // the edge, so they must not fabricate dispatch timing evidence.
+                    if gpt_diagnostics.browser_session_active() {
+                        auction_diagnostics = Some(BrowserAuctionDiagnostics {
+                            auction_dispatched_ms: Some(auction_dispatched_ms),
+                            auction_resolved_ms: Some(auction_resolved_ms),
+                            auction_committed_ms: None,
+                            auction_wait_ms: Some(
+                                auction_resolved_ms.saturating_sub(auction_dispatched_ms),
+                            ),
```
aram356, line 6777: 🤔 **thinking** — This bypasses the `AuctionWaitPlacement` enum and hardcodes the wire string.

Every other producer records a placement as an enum variant, and `publisher.rs:3143-3144` is the single place that maps variants to wire strings:

```rust
AuctionWaitPlacement::PreHeader => "pre_header",
AuctionWaitPlacement::InStream => "in_stream",
```

Because this site writes `Some("pre_header")` directly, renaming either wire value updates the mapping but silently leaves this path emitting the old string — with no compile error to catch it. Routing through the same mapping keeps the two in sync by construction.

There is also a semantic question underneath: `pre_header` means "the auction was awaited before response headers were sent", which is a distinction about an HTML document response. `/_ts/page-bids` is a JSON endpoint with no streamed body, so neither variant really describes it — the field is always `pre_header` by construction and tells a reader nothing.

Consider either threading `AuctionWaitPlacement::PreHeader` through the existing mapping, or omitting `auctionWaitPlacement` on the page-bids path and letting the client's existing `'placement unknown'` branch in `overlay.ts` handle its absence.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
```diff
@@ -1084,13 +1110,26 @@ export function installTsAdInit(): void {
...
+            );
+          } else {
+            recorder?.recordTrustedServerOpportunity(
+              gptSlot,
+              slot.id,
+              opportunity,
+              bid.hb_auction_id,
+              requestedSlotSizes
+            );
+          }
```
aram356: ♻️ **refactor** — These two branches are behaviourally identical; only their arity differs.

`normalizedAuctionFacts(undefined)` returns `{}` (`store.ts:288-291`), and `recordTrustedServerOpportunity` declares `auctionFacts?: GptDiagnosticsAuctionFacts` — so passing `undefined` explicitly is exactly equivalent to omitting the argument. The `else` branch exists only to preserve a 5-argument call shape.

Collapsing to a single call drops ~12 lines and removes the now-unneeded `recorder` binding:

```ts
          const auctionFacts = diagnosticsAuctionFacts(generation, auctionDiagnostics, bid);
          ts.gptDiagnosticsRecorder?.recordTrustedServerOpportunity(
            gptSlot,
            slot.id,
            opportunity,
            bid.hb_auction_id,
            requestedSlotSizes,
            auctionFacts
          );
```

**Apply manually — can't be auto-applied as a `suggestion` because it also requires updating a second file.** I verified this in a scratch worktree: the collapse alone fails 8 tests in `crates/trusted-server-js/lib/test/integrations/gpt/ad_init.test.ts`, every one of them a `toHaveBeenCalledWith` assertion that now sees a trailing `undefined` as a 6th argument. No behavioural break — the fix is to add `undefined` as the trailing expected argument in those 8 assertions, the same way the PR already did for the two assertions it updated at `api.test.ts:132` and `api.test.ts:166`.

Non-blocking; entirely reasonable to leave as-is if you prefer keeping the test assertions untouched.

crates/trusted-server-integration-tests/browser/tests/nextjs/gpt-diagnostics.spec.ts
aram356: 🔧 **wrench** — This change removes the spec's only coverage of a normal fill size.

Both `size` emissions in this file changed from `[300, 250]` to `[1, 1]` (here and at line 282), so `1×1` is now the only size value anywhere in the browser spec. Two things are lost:

1. **No end-to-end check that a normal fill size survives to the export.** The `1×1` value is precisely the one the new `displayableGptFillSize` helper treats as special, so every remaining assertion exercises the suppressed path. A regression that dropped or mangled ordinary `size` values would pass this suite.

2. **The new suppression behaviour still isn't covered end to end.** The spec only reads `tsjs.gptDiagnostics.snapshot()`, which is the export surface — it never inspects badge or panel text. So the assertion here proves the export *retains* `[1, 1]`, but nothing proves the UI *hides* it. That's the actual claim the PR adds to the docs: "the UI suppresses that `1×1` text but the export retains the original `size: [1, 1]` evidence."

The unit tests do cover the helper (`badges.test.ts` asserts `'Filled · SSAT · Size filled 728×90'`), so this is about end-to-end coverage rather than the logic being unverified.

**Proposed fix** (apply manually — spans two emit sites plus new assertions, so it can't be expressed as a single-range `suggestion`): keep one slot emitting a real size and add the `1×1` case alongside it, rather than converting both. For example, leave `gpt-diagnostics-slot-primary` at `size: [300, 250]` and emit `size: [1, 1]` on `gpt-diagnostics-slot-secondary`, then assert both halves of the claim:

```ts
// export retains the raw GPT-reported placeholder
expect(secondary.requests[0]).toMatchObject({ size: [1, 1] });

// ... and the rendered badge omits it
const badgeText = await page.evaluate(
    () =>
        document
            .querySelector("#trusted-server-gpt-diagnostics")
            ?.shadowRoot?.querySelector(".tsgd-badge")?.textContent ?? "",
);
expect(badgeText).not.toContain("1×1");
```

Adjust the badge lookup to match however this suite already reaches the overlay root; the point is asserting the suppression, not the exact selector.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts
```diff
@@ -229,6 +234,36 @@ function cycleFacts(cycle: GptDiagnosticsRequestCycle): string[] {
   if (typeof cycle.trustedServerAuctionId === 'string' && cycle.trustedServerAuctionId.length > 0) {
     facts.push(`Trusted Server auction: ${cycle.trustedServerAuctionId}`);
   }
+  if (cycle.auctionType) facts.push(`Auction type: ${auctionTypeLabel(cycle.auctionType)}`);
+  if (cycle.auctionWinner) {
+    facts.push(`Winning bidder: ${cycle.auctionWinner.bidder}`);
```
aram356, line 239: 📝 **note** — This fallback is unreachable for any store-produced snapshot.

`store.ts:679-684` computes `serverAuctionTimingOrigin` from exactly the same condition that gates `serverAuctionTimings`:

```ts
const serverAuctionTimingOrigin: GptDiagnosticsServerAuctionTimingOrigin | undefined =
  trustedServerEvidence?.serverAuctionTimings === undefined
    ? undefined
    : trustedServerEvidence.auctionType === 'trusted_server'
      ? 'spa_auction'
      : 'navigation';
```

So whenever `cycle.serverAuctionTimings` is present — the condition guarding this whole block — `cycle.serverAuctionTimingOrigin` is present too, and the `??` right-hand side never evaluates.

No action needed; it's harmless defensive code. Flagging only because it reads as a live branch and duplicates the store's derivation rule in a second place, so a future change to that rule would need to be made twice to stay consistent. If you keep it, a short comment saying it's a defensive fallback for externally-supplied cycles would make the intent explicit.

crates/trusted-server-core/src/integrations/gpt_diagnostics.rs
```diff
@@ -279,12 +292,19 @@ pub fn prepare_request(
         replace_path_and_query(request, &clean_path)?;
     }
 
-    let mut decision = GptDiagnosticsRequestDecision::default();
+    let mut decision = GptDiagnosticsRequestDecision {
+        browser_session_active: integration_enabled
```
aram356, line 296: 🤔 **thinking** — `browser_session_active` relaxes all four eligibility guards, where only the navigation one needed relaxing.

`eligible_navigation` combines four conditions:

```rust
let eligible_navigation = request.method() == Method::GET
    && is_navigation_request(request)
    && !crate::publisher::is_prefetch_request(request)
    && !crate::publisher::is_bot_user_agent(request);
```

The new field is gated only on `integration_enabled` plus the cookie state, so it drops the prefetch and bot-user-agent guards along with the document-destination one. Confirmed against this head in a scratch test — a `User-Agent: Googlebot/2.1` request carrying a canonical console cookie yields:

```
active=false  browser_session_active=true
```

The doc comment explains the intent precisely — staying true "for non-document requests such as the SPA page-bids fetch" — and that only requires dropping `is_navigation_request` (and arguably the `GET` check). Keeping the prefetch and bot guards would preserve that intent while narrowing the surface:

```rust
let non_document_session_eligible = !crate::publisher::is_prefetch_request(request)
    && !crate::publisher::is_bot_user_agent(request);

let mut decision = GptDiagnosticsRequestDecision {
    browser_session_active: integration_enabled
        && non_document_session_eligible
        && directive == QueryDirective::Absent
        && cookie_state.occurrences == 1
        && cookie_state.canonical,
    ..GptDiagnosticsRequestDecision::default()
};
```

Apply manually if you want it — it restructures the surrounding binding rather than replacing a contiguous range.

To be clear about severity: I don't think this is exploitable today. The cookie is `__Host-`-prefixed, `HttpOnly`, `Secure`, and `SameSite=Lax`, so a crawler or third-party page shouldn't hold one, and `handle_page_bids` gates on `page_bids_request_allowed` before the decision is ever read. This is defence-in-depth, not a hole — reasonable to defer to a follow-up.

aram356 (changes requested):
## Summary

Solid, well-tested expansion of the GPT diagnostics console: stable `Ad #N` identity across badge/panel/export, auction classification, winner and bucketed-price facts, and server auction timings. The privacy boundary is careful — bounded string lengths, a validated price-bucket shape, and timing normalization at the diagnostics boundary.

Two blocking issues: the SPA timing origin can be derived wrongly after a failed or superseded navigation, labelling navigation-T0 offsets as SPA-auction offsets; and the browser spec loses its only normal-fill-size coverage.

> None of the inline comments below carry a one-click GitHub `suggestion`. Every fix either lands outside the diff hunks or spans a second file, so each is described in prose with the proposed code.

## Blocking

### 🔧 wrench

- **Stale navigation-T0 timings relabelled as SPA-auction timings** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt/index.ts:1467`
- **Browser spec loses all normal-fill-size coverage** — see inline at `crates/trusted-server-integration-tests/browser/tests/nextjs/gpt-diagnostics.spec.ts:223`

## Non-blocking

### 🤔 thinking

- **`auctionDispatchedMs` is structurally always `0` on the page-bids path** — see inline at `crates/trusted-server-core/src/publisher.rs:6748`
- **Page-bids hardcodes `auctionWaitPlacement` as a bare string literal** — see inline at `crates/trusted-server-core/src/publisher.rs:6766`
- **`browser_session_active` drops more eligibility guards than it needs to** — see inline at `crates/trusted-server-core/src/integrations/gpt_diagnostics.rs:296`

### ♻️ refactor

- **Redundant `if`/`else` around `recordTrustedServerOpportunity`** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt/index.ts:1115-1132`

### 📝 note

- **Dead timing-origin fallback in the overlay** — see inline at `crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts:239`

- 📝 **Verified during review, no action needed** — the following were checked against a scratch worktree at this head and are correct: slot-number reuse across eviction produces no `slotOrder`/`slotActivityOrder` duplicates at the `MAX_DIAGNOSTIC_SLOTS` boundary; `set_auction_diagnostics` cannot clobber the auction debug prefix (`prepend_auction_debug_comment` runs after it in `collect_stream_auction`, and `collect_non_html_auction` has no prepend); stripping the console cookie before `handle_request_cookies` is correctly ordered for the stated privacy goal; the added `prepare_request` call in `handle_page_bids` is genuinely idempotent via the extension cache, and all four adapters already call it pre-routing; the normalization boundary correctly rejects `-1.00`, `1e3`, `1.`, `.5` for `priceBucket` and negatives / `NaN` / `> u32::MAX` for timings; and `elapsed_millis` saturation (~49.7 days) is not a practical edge concern.

## CI Status

- cargo fmt: PASS
- cargo test: PASS
- format-docs: PASS
- format-typescript: PASS
- integration tests: PASS
- vitest: PASS

Branch protection reports no required checks on this branch, so none of the above are marked `(required)`.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Implemented in a758e161. The SPA clock now starts at page-bids handler entry, before validation and slot preparation, while `auctionWaitMs` remains dispatch-to-resolution. The Rust regression asserts monotonic milestones and verifies that wait is computed from the dispatch and resolution offsets. The UI and guide now call this anchor `SPA page-bids T0`.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Implemented in a758e161. Both diagnostics producers now use the shared `auction_wait_placement_wire` mapper. Page-bids retains `pre_header` because its auction completes before the JSON response headers are built; the guide now defines this as HTTP response-header placement rather than document placement.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Implemented in a758e161. Added the source-order invariant to `set_auction_diagnostics`: it must run before a debug prefix is prepended because rebuilding the script does not preserve a previously installed prefix.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts
ChristianPavilonis: Implemented in a758e161. Initial-document diagnostics now use `Edge request T0`; SPA diagnostics use `SPA page-bids T0`. Tests and the operator guide were updated to match the actual clock boundaries.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
ChristianPavilonis: Implemented in a758e161. Every built-in direct opportunity now carries explicit `{ auctionType: 'ssat' }` evidence, and the store no longer defaults missing or malformed direct facts to SSAT. The tests now cover valid no-winner SSAT evidence and the fail-closed malformed/missing cases.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/presentation_helpers.ts
ChristianPavilonis: Keeping the existing suppression policy because issue #1081 explicitly asks to remove repeated `1×1` wording from the first-look UI. Commit a758e161 adds browser coverage proving that GPT's raw `1×1` fact remains available in the snapshot/export while the rendered diagnostics suppress it, so the distinction is preserved for deeper inspection without restoring the UI noise.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Implemented in a758e161. The duplicated branches are now one optional-chained recorder call with `diagnosticsAuctionFacts(...)` passed directly, and the affected mock assertions expect the explicit trailing `undefined`.

crates/trusted-server-integration-tests/browser/tests/nextjs/gpt-diagnostics.spec.ts
ChristianPavilonis: Implemented in a758e161. The browser spec again exercises and exports an ordinary `300×250` fill, and a separate second cycle records `1×1`, verifies it remains in the snapshot/export, and verifies it is absent from the rendered diagnostics.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/store.ts
ChristianPavilonis: Thanks for verifying these invariants and the bounded normalization path. No change was needed for this praise-only thread; the related behavior remains covered by the existing tests.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Fixed in a758e161. An accepted pathname navigation now clears `ts.auctionDiagnostics` immediately after incrementing `navGeneration`, before starting page-bids work. Regression coverage verifies stale facts stay absent after failed requests and across superseded requests until the current response commits.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Fixed in a758e161. The page-bids clock begins at handler entry, so dispatch includes real pre-auction handler work. `auctionWaitMs` is still measured strictly from dispatch to resolution, and the Rust test now checks concrete ordering and subtraction rather than field presence alone.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Implemented in a758e161. Page-bids now routes `AuctionWaitPlacement::PreHeader` through the shared enum-to-wire mapper. I retained the placement because the wait occurs before this JSON response's headers are constructed; the guide now states that `pre_header` is relative to HTTP response headers, not specifically an HTML document.

crates/trusted-server-js/lib/src/integrations/gpt/index.ts
ChristianPavilonis: Implemented in a758e161. The redundant `if`/`else` and local recorder binding are gone; the optional facts value is passed to one recorder call. The arity-sensitive mock assertions were updated for the explicit trailing `undefined`.

crates/trusted-server-integration-tests/browser/tests/nextjs/gpt-diagnostics.spec.ts
ChristianPavilonis: Fixed in a758e161. The real-browser path again proves an ordinary `300×250` value reaches both the UI and export. A separate `1×1` cycle proves that the raw value reaches snapshot/export while remaining suppressed in the rendered UI.

crates/trusted-server-js/lib/src/integrations/gpt_diagnostics/overlay.ts
ChristianPavilonis: Acknowledged. This is intentionally unchanged because the fallback is harmless and no action was requested.

crates/trusted-server-core/src/integrations/gpt_diagnostics.rs
ChristianPavilonis: Keeping this unchanged. `browser_session_active` continues to represent canonical session activation, while page-bids separately enforces bot and prefetch eligibility before producing diagnostics. Narrowing the session-state field would mix those responsibilities without changing current behavior; this can be revisited if another consumer appears without its own eligibility gate.

docs/guide/integrations/gpt-diagnostics.md
```diff
@@ -129,6 +135,59 @@ step. When `slotRenderEnded` omits `isEmpty`, the result stays Rendered (fill un
...
+| Competing auctions  | Trusted Server direct and client-side Prebid auction evidence were both observed for the same GPT request. |
+
+Publisher refresh evidence does not by itself establish another auction. It remains
+visible in the request-path classification but does not turn an SSAT, TS auction, or
+client-side auction label into “Competing auctions.” Publisher-only refreshes and
+unattributed requests have no auction label because the available evidence does not
+establish an auction implementation.
+
+Server auction timing and browser GPT timing use separate clocks and are never
+subtracted from each other. Initial SSAT offsets use edge-request T0. SPA TS auction
```
aram356: 🔧 **wrench** — This sentence now describes a clock anchor the code does not use.

The new wording says SPA offsets use a clock started when the page-bids handler begins, "not the browser's navigation clock or **the edge's request-receipt time**". But `handle_page_bids` takes its collector from the request extension rather than constructing one (`publisher.rs:6686-6692`):

```rust
let timings = req
    .extensions()
    .get::<RequestTimings>()
    .cloned()
    .unwrap_or_default();
```

and that extension is inserted at edge-request receipt — Fastly in `adapter-fastly/src/main.rs:218`, before the router is even invoked, and Axum in `adapter-axum/src/timing.rs:89`, in middleware ahead of the handler. So on those adapters the SPA clock *is* the edge request-receipt time this sentence explicitly rules out. The previous round's reply described it as handler entry, which is what the fallback does but not what the extension path does.

This matters beyond wording, because it also makes the two UI anchors overclaim. `overlay.ts:246` renders:

```ts
const timingAnchor = timingOrigin === 'spa_auction' ? 'SPA page-bids T0' : 'Edge request T0';
```

Presented side by side, those read as two different clocks. On Fastly and Axum they are the same clock — edge-request T0 — measured on two different requests (the document navigation vs. the `/_ts/page-bids` fetch). An operator comparing a `spa_auction` number against a `navigation` number would reasonably assume the anchors differ in kind when they differ only in which request started them.

Suggested direction: say both paths anchor at edge-request T0 and that `serverAuctionTimingOrigin` distinguishes *which request* the offsets belong to. That distinction is genuinely useful and worth keeping — it is only the "different clock" framing that is wrong. Relabelling to something like `SPA request T0` / `Initial request T0` would carry the real meaning without implying a separate clock basis.

Apply manually — this is a prose rewrite plus a naming decision spanning the guide and `overlay.ts`, so it cannot be a single-range `suggestion`.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -6548,6 +6689,12 @@ pub async fn handle_page_bids(
         .cloned()
         .unwrap_or_default();
 
+    // Adapter fallbacks prepare this before routing. Keep this idempotent call as
```
aram356: 🔧 **wrench** — Two of the four adapters never insert this extension, so they silently get a different clock anchor.

`unwrap_or_default()` is doing real work here, not just satisfying the type. Neither the Cloudflare nor the Spin adapter references `RequestTimings` at all:

```
$ grep -rn "RequestTimings" crates/trusted-server-adapter-cloudflare/src/ crates/trusted-server-adapter-spin/src/
(no matches)
```

On those two adapters the fallback constructs a fresh collector, and `Default` is `new()`, which stamps `t0: Instant::now()` at call time (`request_timing.rs:427-431`). Since `mark_auction_dispatched()` then fires a short, mostly-synchronous stretch later in the same handler, `auctionDispatchedMs` lands at or near zero and `auctionWaitMs` collapses to `auctionResolvedMs` — exactly the structural-zero problem from the previous round, still present on half the adapters while Fastly and Axum now report genuine edge-anchored offsets.

The practical consequence is that two operators running the same console on different runtimes read materially different dispatch numbers, with nothing in the payload or the UI explaining why. The guide's single description of the anchor cannot be true for all four adapters at once.

Worth noting alongside this: the new monotonicity assertions only exercise the fallback. The page-bids test never inserts a `RequestTimings` extension, so

```rust
assert!(
    dispatched_ms <= resolved_ms && resolved_ms <= committed_ms,
    "page-bids auction milestones should be monotonic"
);
```

is checking the handler-entry clock. Monotonicity holds trivially under either anchor, so the edge-T0 path the guide describes currently has no coverage at all.

Two ways to close it, either is fine:

- Insert `RequestTimings` in the Cloudflare and Spin adapters the way Fastly and Axum do, so all four share one anchor and the guide's description is true everywhere.
- Or treat a missing extension as "no dispatch timing available" and omit `auctionDispatchedMs` / `auctionWaitMs` in that case, rather than reporting a structurally zero dispatch as though it were measured.

Whichever you pick, a page-bids test that inserts a deliberately pre-aged `RequestTimings` (constructed, then a short sleep, then attached to the request) would pin the intended behaviour — asserting `auctionDispatchedMs > 0` under the extension path, or its absence under the fallback.

Apply manually — this spans two adapter crates plus a test, so it cannot be a single-range `suggestion`.

aram356 (changes requested):
## Summary

Second pass, against `563670df9` (the branch was rebased and `Address GPT diagnostics review feedback` added since the last review at `ffc0438dd`).

Five of the seven earlier findings are fixed, two were declined with sound reasoning, and I verified each outcome against the new head rather than taking the replies at face value. The `1×1` browser coverage came back stronger than what I proposed, and dropping `?? 'ssat'` from the store so malformed direct facts fail closed was a good unprompted improvement.

Two new blocking findings, both consequences of the timing fix rather than pre-existing issues: the guide now describes a clock anchor the code does not use, and two of the four adapters silently fall back to a different anchor than the other two.

## Verification of the previous round

| Finding | Outcome |
| --- | --- |
| 🔧 Stale navigation-T0 timings relabelled as SPA-auction timings | **Fixed, re-verified.** The original repro now passes: after a failed page-bids fetch `ts.auctionDiagnostics` is `undefined`, and the superseded-navigation path clears too. `spa_hook.test.ts` pins both. |
| 🤔 `auctionDispatchedMs` structurally always `0` | **Partly fixed** — now a real offset on Fastly and Axum, still structurally zero on Cloudflare and Spin. See finding B. |
| 🤔 Hardcoded `auctionWaitPlacement` wire literal | **Fixed.** Shared `auction_wait_placement_wire` mapper, and page-bids routes `AuctionWaitPlacement::PreHeader` through it. |
| ♻️ Redundant `if`/`else` around the recorder call | **Fixed** as proposed; the arity-sensitive assertions now pin `{ auctionType: 'ssat' }` rather than a bare trailing `undefined`. |
| 🔧 Browser spec lost all normal-fill-size coverage | **Fixed, and better than proposed.** `300×250` is exercised and exported again, a separate `1×1` cycle proves the raw fact survives to snapshot/export while staying out of the rendered UI, and `captureClosedShadowRoots` is a tidy way to read the closed root. |
| 📝 Dead timing-origin fallback in the overlay | Declined — reasonable, it is harmless. |
| 🤔 `browser_session_active` drops prefetch and bot guards | **Declined, and the rebuttal checks out.** `publisher.rs:6863` gates the auction on `!is_bot && !is_prefetch`, so the diagnostics block is unreachable for a bot or prefetch. Withdrawing it. |

Local verification at this head: JS suite 899 passed / 0 failed, no type errors; the previous round's repro re-run against the new code; adapter `RequestTimings` wiring grepped across all four crates.

## Blocking

### 🔧 wrench

- **Guide contradicts the code on the SPA clock anchor** — see inline at `docs/guide/integrations/gpt-diagnostics.md:156`
- **Cloudflare and Spin silently fall back to a handler-entry clock** — see inline at `crates/trusted-server-core/src/publisher.rs:6689`

## CI Status

- cargo fmt: PASS
- cargo test: PASS
- format-docs: PASS
- format-typescript: PASS
- integration tests: PASS
- vitest: PASS

Branch protection reports no required checks on this branch, so none are marked `(required)`.

docs/guide/integrations/gpt-diagnostics.md
ChristianPavilonis: Fixed in f951955f. The guide now describes both initial-document and SPA offsets as using the adapter server-request clock, with the timing origin identifying which HTTP request owns the offsets. The panel labels now read `Initial document request T0` and `SPA page-bids request T0`, so they no longer imply different clock semantics. The overlay assertions and timing design spec were updated with the same terminology.

crates/trusted-server-core/src/publisher.rs
ChristianPavilonis: Fixed in f951955f. Cloudflare and Spin now attach `RequestTimings` near router entry for every non-health request, matching the request-scoped collector contract used by Fastly and Axum without adding header emission. The page-bids handler also fails closed: if an adapter omits the collector, it does not expose handler-local timing facts. Regression coverage verifies both adapters attach the collector, a pre-aged collector produces `auctionDispatchedMs > 0`, and a missing collector omits diagnostics.

crates/trusted-server-adapter-cloudflare/src/middleware.rs
```diff
@@ -46,6 +47,40 @@ impl Middleware for SanitizeRequestMiddleware {
...
+    #[must_use]
+    pub fn new() -> Self {
+        Self
+    }
+}
+
+#[async_trait(?Send)]
+impl Middleware for RequestTimingMiddleware {
+    async fn handle(&self, mut ctx: RequestContext, next: Next<'_>) -> Result<Response, EdgeError> {
+        if ctx.request().uri().path() != "/health"
```
aram356: 🔧 **wrench** — This middleware should live in EdgeZero, not as a per-adapter copy.

The implementation landed twice, byte for byte. Diffing the `RequestTimingMiddleware` block in this file against `trusted-server-adapter-spin/src/middleware.rs` returns no differences at all, tests included — same struct, same `handle`, same `/health` skip, same `request_timing_middleware_attaches_a_collector_except_for_health` loop over `[("/test", true), ("/health", false)]`.

Nothing in it is Trusted Server's concern. "Attach a per-request clock at router entry unless the path is the liveness probe" is a property of the edge framework, and EdgeZero already accepts middleware of exactly this kind: `edgezero-core/src/middleware.rs:74` ships a concrete `RequestLogger` alongside the `Middleware` trait, so this is not a category EdgeZero refuses. EdgeZero currently has no timing concept at all, which is precisely why each adapter is now carrying its own copy.

I checked that the move is actually possible rather than just desirable, because the collector has to move with the middleware — `RequestTimings` lives in `trusted-server-core`, and dependencies run Trusted Server → EdgeZero, never back. The split is cleaner than it looks:

- **Generic, moves to EdgeZero**: `new`, `record`, `span`, `mark_headers_ready`, `mark_request_elapsed`, `set_resp_bytes`, `server_timing_value`, `snapshot`, `PhaseSpan`, plus this middleware.
- **Domain-specific, stays here**: `Phase`'s eight variants (`EcKv`, `AuctionWait`, `Origin`, `TemplateCacheLookup`, …), `AuctionWaitPlacement`, `record_auction_wait`, `set_auction_id`, and the three `mark_auction_*` marks. A generic edge framework should not know about ad auctions or EC key-value stores, so the phase taxonomy needs to become a parameter (generic over a phase enum, or an index/string-keyed slot API) rather than moving as-is.

The dependency side is already in place: `edgezero-core` carries `http`, `web-time`, and `async-trait` today. `uuid` is the only addition the collector would suggest, and it is used solely by `set_auction_id`, which stays on this side — so it does not need to follow.

The one genuine coupling to unpick is `request_timing.rs:19`, `use crate::cache_policy::cache_control_headers_are_private_or_no_store`, called once at line 455 to decide whether `Server-Timing` may be appended. Either take a predicate at that boundary, or leave `append_server_timing_if_private` in Trusted Server and move only the collector.

Two practical notes on sequencing, since EdgeZero is a separate repo pinned at `tag = "v0.0.7"` in the root `Cargo.toml`. First, this needs an EdgeZero change, a tag, and a bump here, coordinated across four adapters — it is a migration, not a cleanup, and doing it as a follow-up PR stacked after this one is entirely reasonable. Second, if you want to unblock merging sooner without shipping the duplication, there is a smaller intermediate step available today: `trusted-server-core` already depends on `edgezero-core` for the `Middleware` trait and already has `async-trait`, and both adapters already depend on `trusted-server-core` — so hosting one copy in `trusted-server-core` right now is a pure move, no redesign, and it reduces the eventual EdgeZero extraction to relocating a single item instead of reconciling two copies.

Apply manually — this spans two adapter crates plus `request_timing.rs`, and the EdgeZero half lives in another repository, so it cannot be expressed as a `suggestion`.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -6538,15 +6679,21 @@ pub async fn handle_page_bids(
...
-        .unwrap_or_default();
+    // Keep a local collector for direct-handler callers, but only expose
+    // browser timings from an adapter-attached request clock. Otherwise an
+    // adapter that forgets the extension would silently report a different
+    // handler-entry timing origin.
+    let request_timings = req.extensions().get::<RequestTimings>().cloned();
+    let has_request_timings = request_timings.is_some();
+    let timings = request_timings.unwrap_or_default();
+
+    // Adapter fallbacks prepare this before routing. Keep this idempotent call as
```
aram356: 🌱 **seedling** — The fail-closed rule now guards this handler but not the other one that depends on it.

`has_request_timings` correctly stops `handle_page_bids` from publishing handler-local offsets when an adapter omits the collector. The initial-document path has no equivalent guard: `handle_publisher_request` at `publisher.rs:4266-4270` still does

```rust
let timings = req
    .extensions()
    .get::<RequestTimings>()
    .cloned()
    .unwrap_or_default();
```

and `set_auction_diagnostics` gates only on `diagnostics_active` and on whether a dispatch mark exists, never on where the clock came from. So the same defect this commit fixes for SPA offsets could reappear for initial-document offsets, reported under the `Initial document request T0` label.

Not reachable today, and I confirmed that rather than assuming it: all four adapters now attach the collector in production code (Fastly 3 insert sites, Axum 1, Cloudflare 1, Spin 1). So this is about where the invariant is written down, not a live bug — worth a follow-up issue rather than a change in this PR.

The reason it is worth recording: the invariant is now enforced in one of the two places that rely on it, and the unguarded side is the one with no test covering a missing collector. A future adapter, or a router path that bypasses the new middleware, would reintroduce it on the navigation side only. Applying the same `has_request_timings` gate in `handle_publisher_request` — or pushing the check down into `set_auction_diagnostics` so both callers inherit it — would make the rule hold structurally instead of per-call-site.

aram356 (changes requested):
## Summary

Third pass, against `f951955f5`. The two blocking findings from the previous round are both genuinely fixed, and I verified each against this head rather than from the replies: the guide and panel labels now describe the adapter server-request clock that the code actually uses, Cloudflare and Spin attach the collector, and `handle_page_bids` fails closed when an adapter omits it. The pre-aged-collector test closes the coverage gap I raised, and the timing design spec was updated unprompted so it does not go stale.

What remains is the shape of the fix rather than its correctness. `RequestTimingMiddleware` landed as two byte-identical copies in two adapter crates, which is the fourth instance of a duplication pattern this repo has been carrying, and the new middleware plus the collector it installs are framework-level concerns that belong in EdgeZero rather than in Trusted Server's adapters.

## Verification performed at this head

Local runs: `cargo test-cloudflare` (22 passed), `cargo test-spin` (44 + 37 passed), `cargo test-axum` (17 + 1 + 25 passed), core `page_bids` filter (25 passed, including the new fail-closed test), `page_bids_response_includes_auction_id_only_for_winning_bids` (passed, confirming `auctionDispatchedMs > 0` under the extension path), and the full JS suite (899 passed, no type errors). The first round's stale-timing repro still comes back clear, and both timing origins map to the corrected labels.

`git diff 563670df9 f951955f5` is exactly the nine files addressing the previous round, with no incidental drift, so the surfaces cleared in earlier passes are unchanged.

Things that looked like problems and are not: the hardcoded `"/health"` matches the existing literal in the Fastly and Axum paths rather than introducing it; Spin's second router at `app.rs:441` is the degraded-state 503 fallback where no auction runs; and the 2 ms `thread::sleep` in the async test has ample margin for a `> 0` millisecond assertion with nothing else pending on that runtime.

## Blocking

### 🔧 wrench

- **Move the request-timing middleware into EdgeZero instead of duplicating it per adapter** — see inline at `crates/trusted-server-adapter-cloudflare/src/middleware.rs:73`

## Non-blocking

### 🌱 seedling

- **The fail-closed timing invariant is enforced on only one of the two handlers that depend on it** — see inline at `crates/trusted-server-core/src/publisher.rs:6689`

- 📌 **The other three adapter middlewares carry the same duplication** — `SanitizeRequestMiddleware`, `FinalizeResponseMiddleware`, and `AuthMiddleware` are each duplicated across `trusted-server-adapter-cloudflare/src/middleware.rs` and `trusted-server-adapter-spin/src/middleware.rs` in the same way, and predate this PR. They are not this PR's to fix, but they are the reason the new middleware should not become a fourth instance: whatever mechanism carries `RequestTimingMiddleware` into shared code is the one that should eventually carry these too. Worth a follow-up issue so the pattern stops growing.

- 📝 **Where the generic/domain boundary falls, for whoever picks up the extraction** — I mapped `request_timing.rs` against `edgezero-core` to check the move is actually possible. Generic and movable: `new`, `record`, `span`, `mark_headers_ready`, `mark_request_elapsed`, `set_resp_bytes`, `server_timing_value`, `snapshot`, and `PhaseSpan`. Domain-specific and staying in Trusted Server: `Phase`'s eight variants (`EcKv`, `AuctionWait`, `Origin`, `TemplateCacheLookup` and the rest), `AuctionWaitPlacement`, `record_auction_wait`, `set_auction_id`, and the three `mark_auction_*` marks. `edgezero-core` already carries `http`, `web-time`, and `async-trait`, so `uuid` (used only by `set_auction_id`, which stays behind anyway) is the one dependency question, and it resolves itself once the auction marks stay on this side.

## CI Status

- cargo fmt: PASS
- cargo test: PASS
- format-docs: PASS
- format-typescript: PASS
- integration tests: PASS
- vitest: PASS

Branch protection reports no required checks on this branch, so none are marked `(required)`.

PR #1127, opened 2026-09-07
Forward trusted geo to Didomi notice loaders

### LINKED ISSUE #85: As publisher I would like to pass geo information to Didomi
These values will need to be passed into API or Backend calls such as these fields required by Didomi for consent. So what would happen in JS today as such:

window.didomiConfig = {
user: {
country: "US",
region: "CA"
}
}

will need to be passed in URL parameters as such:
[https://client.com/consent/{public](https://client.com/consent/%7Bpublic) API key}/loader.js?target_type=notice&target=4QS9p3Qn&country=US&region=CA

(https://developers.didomi.io/api-and-platform/domains/reverse-proxy/self-hosting)

prk-Jr:
## Summary

- Add opt-in trusted country and region parameters to Didomi notice-loader URLs so browser and edge-cache identities select the correct geographic notice.
- Canonicalize eligible loader requests through a private same-origin redirect, replace caller-supplied geo, and fail closed when complete trusted geo is unavailable.
- Preserve Didomi SDK cache metadata while bypassing cache and enforcing private responses for Didomi API traffic.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/integrations/didomi.rs` | Adds typed configuration, geo normalization, exact loader matching, canonical redirects, authoritative upstream headers, API cache privacy, and regression coverage. |
| `crates/trusted-server-core/src/config_payload.rs` | Verifies the new option survives the deployed configuration blob round trip. |
| `crates/trusted-server-adapter-fastly/src/platform.rs` | Verifies query encoding, order, and duplicates survive Fastly request conversion. |
| `trusted-server.example.toml` | Documents the disabled-by-default TOML option. |
| `crates/trusted-server-integration-tests/fixtures/configs/trusted-server.integration.toml` | Adds the option to the integration fixture. |
| `docs/guide/integrations/didomi.md` | Replaces the stale guide with the current TOML/typed-CLI configuration flow, corrects the inaccurate claim that publisher `Authorization` is forwarded, and documents geo, caching, limitations, and rollout checks. |
| `docs/superpowers/specs/2026-09-07-didomi-geo-design.md` | Records the reviewed design and acceptance criteria. |
| `docs/superpowers/plans/2026-09-07-didomi-geo-forwarding.md` | Records the completed implementation and verification plan. |

## Rollout considerations

- A notice-loader request with country but no valid region returns a private `503`. This is a deliberate availability decision because the canonical cache identity requires both fields; Fastly can omit region for locations without an ISO subdivision.
- The option defaults to `false`. Manual staging verification with real Fastly geo does not block merging the disabled feature, but it must be completed before enabling `geo_query_parameters` for production traffic.
- Follow-up #1151 tracks replacing legacy caller-supplied geo headers on other Didomi SDK paths. Follow-up #1152 tracks avoiding the first-load redirect by exposing trusted geo to the browser bootstrap. Follow-up #1153 tracks the pre-existing API path and method contract audit.

## Closes

Closes #85

## Test plan

- [x] `cargo clippy-cloudflare-wasm && cargo clippy-spin-native && cargo clippy-spin-wasm`
- [x] JS tests: `cd crates/trusted-server-js/lib && npx vitest run` (Node 24.12.0; 893 passed)
- [x] Adapter parity: `cargo test --manifest-path crates/trusted-server-integration-tests/Cargo.toml --test parity`
- [x] Focused Didomi tests: `cargo test --package trusted-server-core --target aarch64-apple-darwin didomi`
- [x] CLI and native integration-test package suites
- [x] WASM release build: `cargo build --package trusted-server-adapter-fastly --release --target wasm32-wasip1`
- [x] Local headless-Chrome verification through `ts dev proxy` and Viceroy: conflicting geo returned a private `307`, then the canonical loader returned `200`.
- [ ] Manual staging verification with real Fastly geo, browser redirects, cache reuse, and Didomi notice selection (required before production enablement, not before merge)

## Checklist

- [x] Uses repository `log` macros and no `println!`
- [x] New code has tests


ChristianPavilonis (approved):
## Summary

Reviewed the trusted geo forwarding flow through canonical redirects, upstream request construction, response privacy, and Fastly cache behavior. I found no actionable issues in revision `eb4e9431695fdcfcf65cdcb7bd2bc3d6cc72cd2d`.

crates/trusted-server-core/src/integrations/didomi.rs
```diff
@@ -100,6 +103,112 @@ enum DidomiBackend {
...
+        || !country.bytes().all(|byte| byte.is_ascii_alphabetic())
+        || matches!(country.as_str(), "XX" | "ZZ")
+    {
+        return Err(DidomiGeoError::InvalidCountry);
+    }
+
+    let Some(region) = geo.region.as_deref() else {
+        return Err(DidomiGeoError::MissingRegion);
+    };
+    let region = trim_ascii(region).to_ascii_uppercase();
```
aram356: ♻️ **refactor** — `trim_ascii` duplicates std's `str::trim_ascii`, stable since Rust 1.80 (the repo pins 1.95). Dropping the helper and calling the std method at both call sites removes six lines with no behavior change.

```suggestion
fn normalize_didomi_geo(geo: &GeoInfo) -> Result<DidomiGeo, DidomiGeoError> {
    let country = geo.country.trim_ascii().to_ascii_uppercase();
    if country.is_empty() {
        return Err(DidomiGeoError::MissingCountry);
    }
    if country.len() != 2
        || !country.bytes().all(|byte| byte.is_ascii_alphabetic())
        || matches!(country.as_str(), "XX" | "ZZ")
    {
        return Err(DidomiGeoError::InvalidCountry);
    }

    let Some(region) = geo.region.as_deref() else {
        return Err(DidomiGeoError::MissingRegion);
    };
    let region = region.trim_ascii().to_ascii_uppercase();
```

_(Scratch-verified: `cargo fmt --check`, clippy `-D warnings` on `trusted-server-core`, and the focused didomi/config tests pass with this applied.)_

crates/trusted-server-core/src/integrations/didomi.rs
```diff
@@ -100,6 +103,112 @@ enum DidomiBackend {
...
+    }
+
+    let Some(path) = consent_path.strip_prefix('/') else {
+        return false;
+    };
+    let mut segments = path.split('/');
+    let public_key = segments.next();
+    let file_name = segments.next();
+
+    public_key.is_some_and(|segment| !segment.is_empty())
```
aram356: ⛏ **nitpick** — `.` and `..` currently count as a `<public-key>` segment, so `/prefix/../loader.js` engages geo lookup and redirects to a path the browser normalizes away from the proxy prefix. The `proxy_path` validator already rejects dot segments; the matcher can mirror it:

```suggestion
    public_key.is_some_and(|segment| !segment.is_empty() && !matches!(segment, "." | ".."))
```

Worth also adding `(Method::GET, "/./loader.js")` and `(Method::GET, "/../loader.js")` to the `matches_only_exact_get_notice_loader_paths` rejection table (outside this suggestion's range).

_(Compile-verified only — fmt, clippy, and the focused didomi tests pass; please re-run the matching `cargo test-*` aliases after applying.)_

crates/trusted-server-core/src/integrations/didomi.rs
```diff
@@ -226,6 +353,29 @@ impl DidomiIntegration {
...
+        location: &str,
+    ) -> Result<http::Response<EdgeBody>, Report<TrustedServerError>> {
+        let mut response = http::Response::builder()
+            .status(http::StatusCode::TEMPORARY_REDIRECT)
+            .header(header::LOCATION, location)
+            .body(EdgeBody::empty())
+            .change_context(Self::error("Failed to build Didomi geo redirect"))?;
+        crate::response_privacy::enforce_terminal_private_cache_privacy(&mut response);
+        Ok(response)
+    }
```
aram356, line 376: 🤔 **thinking** — `add_cors_headers` puts `Access-Control-Allow-Origin: *` on every proxied SDK response, which exists to support pages that load the proxy cross-origin. These two synthesized responses skip it, so a loader embedded with `crossorigin` from a different origin would fail the CORS check on the geo `307`/`503` (every response in a redirect chain needs CORS headers in CORS mode), and the loader would break only when geo is enabled. In the primary same-origin topology this is a non-issue. Either call `Self::add_cors_headers` on both before enforcing cache privacy, or confirm cross-origin embedding is out of scope.

crates/trusted-server-core/src/integrations/didomi.rs
```diff
@@ -286,13 +436,41 @@ impl IntegrationProxy for DidomiIntegration {
...
+                Ok(None) => return Ok(Self::geo_failure_response("geo_unavailable")),
+                Err(_) => return Ok(Self::geo_failure_response("lookup_failed")),
+            };
+            let canonical = canonical_loader_url(&path, parts.uri.query(), &geo);
+            let incoming_path_and_query = parts
+                .uri
+                .path_and_query()
+                .map_or(path.as_str(), http::uri::PathAndQuery::as_str);
+            if incoming_path_and_query != canonical.browser_target {
+                return Self::redirect_response(&canonical.browser_target);
```
aram356, line 455: 🌱 **seedling** — redirect convergence assumes the client's geo is stable across consecutive requests. Egress IPs whose geo shifts between requests (carrier NAT pools, iCloud Private Relay) can bounce between differing canonical targets until the browser's redirect cap, and the notice then fails to load for that pageview. Bounded and inherent to the chosen redirect design, but worth a staging check with relay-style traffic and a line in the rollout checklist before production enablement.

aram356 (approved):

crates/trusted-server-core/src/integrations/didomi.rs
prk-Jr: Applied in 09624c3b. Helper dropped, `str::trim_ascii` at both call sites.

crates/trusted-server-core/src/integrations/didomi.rs
prk-Jr: Applied in 09624c3b. Matcher now rejects `.` and `..` as the public-key segment, mirroring `validate_proxy_path`, and `(Method::GET, "/./loader.js")` / `(Method::GET, "/../loader.js")` are in the `matches_only_exact_get_notice_loader_paths` rejection table.

crates/trusted-server-core/src/integrations/didomi.rs
prk-Jr: Took the "add the headers" branch rather than declaring cross-origin embedding out of scope, since these were the only SDK-path responses missing CORS and the breakage would have appeared only once geo forwarding was enabled. `Self::add_cors_headers` now runs on both synthesized responses before `enforce_terminal_private_cache_privacy` (that function only touches cache-control and validator headers, so there is no interaction). Asserted in `enabled_geo_redirects_noncanonical_loader_without_upstream_call` and `enabled_geo_fails_closed_without_complete_geo`. 09624c3b.

crates/trusted-server-core/src/integrations/didomi.rs
prk-Jr: No code change — the redirect design is deliberate, so this is a rollout concern rather than a defect. Added as rollout check 8 in `docs/guide/integrations/didomi.md`: test from egress whose geo can shift between requests (carrier NAT pools, relay-style proxies), and the note that such an egress can bounce between canonical targets until the browser redirect limit and fail to load the notice for that pageview. 09624c3b.

PR #1135, opened 2026-09-07
Make body hold parser-aware and stream Next.js processing

### LINKED ISSUE #850: Fix </body> hold-buffer scan to use parser context; make nextjs post-processor streaming-safe
### Description

From the 2026-07-05 performance & architecture review (P0 finding 4).

The `</body>` hold buffer does a raw-byte search for `</body` (`publisher.rs:914-966`). The literal appearing inside inline JSON/JS (Next.js `__next_f` payloads, serialized editor content) triggers the hold early, converting the tail-only hold into a near-full-page hold.

Separately, registering any HTML post-processor (currently only `nextjs`) buffers the whole document (`html_processor.rs:59-148`). Next.js publishers are a primary SSAT target and currently get the worst of both.

### Fix

- Derive the hold from lol_html's parser context instead of raw bytes.
- Make the nextjs post-processor streaming-safe, or explicitly document that combination as buffered mode.

### Effort / expected effect

- **Effort:** Medium
- **Expected effect:** Removes near-full-page holds on the primary target CMS.

_Sequencing item 4 of 10._

prk-Jr:
## Summary

- Use a parser-generated, request-specific seam to delay only the structural `</body>` tail, preventing inline JavaScript or JSON literals from holding most of the page.
- Replace whole-document Next.js post-processing with bounded per-document streaming that rewrites complete RSC groups and restores malformed, incomplete, or over-limit payloads unchanged.
- Preserve auction ordering and terminal telemetry across streaming, compressed, missing-body-end, and failure paths.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/html_processor.rs` | Insert a parser-confirmed deferred body marker and stream transformed output. |
| `crates/trusted-server-core/src/integrations/google_tag_manager.rs` | Adapt registration to the per-document processor interface. |
| `crates/trusted-server-core/src/integrations/mod.rs` | Define the per-document HTML stream processor factory contract. |
| `crates/trusted-server-core/src/integrations/nextjs/html_post_process.rs` | Remove the whole-document Next.js post-processor. |
| `crates/trusted-server-core/src/integrations/nextjs/mod.rs` | Register the bounded Next.js stream processor. |
| `crates/trusted-server-core/src/integrations/nextjs/rsc.rs` | Classify malformed and incomplete RSC T-chunks safely, including decoded character boundaries. |
| `crates/trusted-server-core/src/integrations/nextjs/rsc_placeholders.rs` | Capture RSC scripts into request-namespaced placeholders with bounded fallback. |
| `crates/trusted-server-core/src/integrations/nextjs/rsc_stream.rs` | Add ordered, bounded RSC grouping, rewriting, restoration, and streaming release. |
| `crates/trusted-server-core/src/integrations/nextjs/script_rewriter.rs` | Move Next.js script fragment state to the document scope. |
| `crates/trusted-server-core/src/integrations/nextjs/shared.rs` | Share URL rewrite support with the stream processor. |
| `crates/trusted-server-core/src/integrations/registry.rs` | Construct processor instances per document. |
| `crates/trusted-server-core/src/publisher.rs` | Collect auctions at the parser-confirmed seam while preserving early output and abandonment telemetry. |
| `docs/guide/integrations/nextjs.md` | Document bounded Next.js streaming and unchanged fallback behavior. |
| `docs/superpowers/specs/2026-09-07-850-parser-aware-body-hold-nextjs-streaming-design.md` | Add the reviewed design specification. |
| `docs/superpowers/plans/2026-09-07-850-parser-aware-body-hold-nextjs-streaming.md` | Add the reviewed implementation plan. |

## Closes

Closes #850

## Test plan

- [ ] JS tests: `cd crates/trusted-server-js/lib && npx vitest run` (blocked by the existing CommonJS/ESM incompatibility between `html-encoding-sniffer` and `@exodus/bytes`; no JS files changed)
- [x] Other: `cargo test-cloudflare`, `cargo test-spin`, `cargo clippy-cloudflare`, `cargo clippy-cloudflare-wasm`, `cargo clippy-spin-native`, and `cargo clippy-spin-wasm`

## Checklist

- [x] Uses repository logging macros (not `println!`)
- [x] New code has tests


crates/trusted-server-core/src/integrations/nextjs/rsc_stream.rs
```diff
@@ -0,0 +1,1031 @@
...
+                .unwrap_or_else(std::sync::PoisonError::into_inner)
+                .bypass_rsc = true;
+            return self.release_group(None).map(Some);
+        }
+        let payloads: Vec<&str> = self
+            .group
+            .iter()
+            .map(|payload| payload.original.as_str())
+            .collect();
+        match classify_rsc_group(&payloads, self.limit) {
```
ChristianPavilonis: **[P2] Classify unresolved RSC groups incrementally**

`resolve_group` calls `classify_rsc_group` after every payload, and that function rebuilds and rescans the entire logical group. At the allowed maximum of 256 payloads, an 8.96 MiB T-chunk causes roughly 1.15 GiB of cumulative copying and scanning.

A native debug probe with fixed 8.96 MiB content took 870 ms for 32 segments and 4,510 ms for 256 segments. Large but permitted responses can therefore exhaust an edge worker’s CPU budget before releasing output.

Please retain classifier state and feed each new payload once, or enforce a cumulative-work limit that restores the group before repeated scans become expensive.

crates/trusted-server-core/src/integrations/nextjs/shared.rs
```diff
@@ -13,9 +13,12 @@ use crate::host_rewrite::rewrite_bare_host_at_boundaries;
 // intentionally remain lazy statics instead of participating in
 // `Settings::prepare_runtime`.
 /// RSC push script call pattern for extracting payload string boundaries.
+///
+/// The initializer prologue may already have streamed before a fragmented
+/// `__next_f` identifier is recognized, so matching can start at the identifier.
 pub(crate) static RSC_PUSH_CALL_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(
-        r#"(?s)(?:(?:self|window)\.__next_f\.push|\(\s*(?:self|window)\.__next_f\s*=\s*(?:self|window)\.__next_f\s*\|\|\s*\[\]\s*\)\s*\.push)\(\[\s*1\s*,\s*(['"])"#,
+        r#"(?s)(?:(?:(?:self|window)\.)?__next_f\.push|(?:(?:\(\s*)?(?:self|window)\.)?__next_f\s*=\s*(?:self|window)\.__next_f\s*\|\|\s*\[\]\s*\)\s*\.push)\(\[\s*1\s*,\s*(['"])"#,
```
ChristianPavilonis: **[P2] Do not match `__next_f` as a suffix of arbitrary member expressions**

Making the `self.`/`window.` qualifier optional allows this regex to match expressions such as:

```js
other.__next_f.push([1, "https://origin.example.com/path"])
```

The production placeholder rewriter then treats that payload as Next.js Flight data and rewrites its URL. A probe changed the example above to use the proxy host. The base regex accepted only `self.__next_f` and `window.__next_f`.

Please keep qualified calls strict and support a streamed-away qualifier with a separate alternative anchored to the beginning of the claimed fragment. Negative tests should cover other object properties, identifier suffixes, and string/comment literals.

crates/trusted-server-core/src/integrations/nextjs/rsc_placeholders.rs
```diff
@@ -38,6 +22,108 @@ impl NextJsRscPlaceholderRewriter {
...
+            state.bypass_rsc = true;
+            return if was_buffered {
+                ScriptRewriteAction::replace(content.to_owned())
+            } else {
+                ScriptRewriteAction::Keep
+            };
+        }
+
+        let payload = &content[payload_start..payload_end];
+        let exceeds_limit = payload.len() > limit
```
ChristianPavilonis, line 81: **[P2] Do not apply the group limit across independent payloads in one parser call**

`captured_payload_bytes` includes every placeholder queued while `HtmlRewriterAdapter::process_chunk` is running. The downstream processor cannot resolve and decrement an already-complete group until the entire parser call returns.

As a result, two independent payloads that are individually below `max_combined_payload_bytes` enter document-wide bypass when they arrive in one source chunk, but rewrite successfully when the same HTML is delivered in separate chunks. With a 100-byte limit and two 65-byte payloads, a single call produced no proxy-host rewrites, while two calls rewrote both payloads.

Please enforce the cumulative limit on the downstream unresolved semantic group rather than the parser’s entire pending queue. The parser can retain an individual-script limit while the parser-to-processor queue is bounded separately.

ChristianPavilonis (changes requested):
## Summary

The parser-aware seam paths are well covered, but the new RSC streaming implementation introduces CPU-scaling, false-positive matching, and chunk-dependent fallback behavior that should be corrected before merging.

**[P3] Update the deprecation guidance to name an available replacement** (`crates/trusted-server-core/src/integrations/nextjs/html_post_process.rs:98`)

The public functions are still re-exported for compatibility, but their documentation and compiler deprecation notes tell callers to use `NextJsHtmlPostProcessor`, which this PR deletes.

Please point callers to the enabled Next.js streaming integration, or state that these compatibility functions have no direct public replacement.

crates/trusted-server-core/src/integrations/nextjs/shared.rs
aram356: 🔧 **wrench** — This widening has no left boundary, so it now matches `__next_f.push([1,"…")` after **any** receiver, and ordinary publisher scripts get their string literals rewritten.

Verified end-to-end through the real pipeline: `myAnalytics.__next_f.push([1,"https://origin.example.com/track"])` is captured into `captured_payloads`, replaced with `__ts_rsc_<ns>_0__`, and the stream processor substitutes a **host-rewritten** payload. Same for `foo.bar.__next_f.push(...)`, `window.myapp.__next_f.push(...)`, and — because the pattern also has no identifier boundary — `a__next_f.push(...)`. On `main` none of these match.

Failure scenario: a publisher ships any non-Next.js script containing `.__next_f.push([1,"…"])` (an analytics shim, a vendor bundle, a variable that happens to end in `__next_f`) → its string literal is silently URL-rewritten with no warning and no way for the publisher to opt out.

**No `suggestion` block here, deliberately.** I tried the obvious narrowing — restoring `(?:self|window)\.` on the first alternative — applied it in a scratch worktree and ran the gate. It **fails** `integrations::nextjs::tests::html_processor_rewrites_rsc_stream_payload_with_chunked_input` (SIGABRT under Viceroy). That test uses `chunk_size = 32`, which fragments the script text so `claimed_start` trims the `self.` prologue and the surviving claim genuinely begins at a bare `__next_f` — so the bare-receiver alternative **is** load-bearing, and narrowing the regex alone is wrong. I also confirmed a negative-lookbehind variant satisfies every case but does not compile: `regex` 1.12 rejects `(?<!…)` with a parse error.

**Proposed fix (apply manually — belongs in the caller, not the pattern).** The bare receiver is only legitimate when the claim was actually trimmed, i.e. when `prior_probe` is non-empty. When `prior_probe` is empty and `claimed_start == 0`, the text is a complete unfragmented script and should require the `self.`/`window.` receiver. Thread that distinction into matching rather than relaxing the pattern for both cases — for example keep two compiled patterns (strict for the untrimmed case, bare-permitted for the trimmed case) and select on `prior_probe.is_empty()`, or check the byte before `identifier_start` is not in `[A-Za-z0-9_.$]` before accepting a bare match.

Whichever shape you pick, a regression test for `myAnalytics.__next_f.push([1,"https://origin.example.com/x"])` staying byte-identical would lock this down — there is currently no test asserting a non-RSC `__next_f` receiver is left alone.

crates/trusted-server-core/src/integrations/nextjs/rsc.rs
```diff
@@ -189,19 +189,25 @@ fn consume_unescaped_bytes(s: &str, start_pos: usize, byte_count: usize) -> (usi
...
+
+pub(super) enum TChunkScan {
+    Complete(Vec<TChunkInfo>),
+    NeedMore,
+    Invalid,
 }
 
 /// Find all T-chunks in content, optionally skipping markers.
-fn find_tchunks_impl(content: &str, skip_markers: bool) -> Option<Vec<TChunkInfo>> {
+fn scan_tchunks_impl(content: &str, skip_markers: bool) -> TChunkScan {
```
aram356: 🔧 **wrench** — `scan_tchunks_impl` is the new entry point that reaches `EscapeSequenceIter`, and that iterator slices `&str` at indices it never checks for char boundaries. Publisher-controlled RSC text can therefore panic the HTML pipeline, which under `wasm32-wasip1` traps and returns 500 for the page.

Three sites in `EscapeSequenceIter::next`, all guarded on length only:

| Line | Slice | Input that panics |
| --- | --- | --- |
| `rsc.rs:156` | `&self.str_ref[self.pos..]` | `\x4é` |
| `rsc.rs:124` | `&self.str_ref[self.pos + 2..self.pos + 6]` | `\u12😀` |
| `rsc.rs:133` | `&self.str_ref[self.pos + 8..self.pos + 12]` | `\ud83d\u12😀` |

The `\x` guard `self.pos + 3 < self.bytes.len()` bounds length but not boundaries; the `\u` guards slice before validating the hex.

Reproduced end-to-end from ordinary origin HTML through `create_html_processor` + `StreamingPipeline` — not just by calling the iterator directly:

```
<html><body><script>self.__next_f.push([1,"1:T9,\x4ézzzzzzzz"])</script></body></html>

thread '…' panicked at crates/trusted-server-core/src/integrations/nextjs/rsc.rs:156:33:
start byte index 9 is not a char boundary; it is inside 'é' (bytes 8..10) of `1:T9,\x4ézzzzzzzz`
```

To be clear about scope: **this is pre-existing, not a regression.** `EscapeSequenceIter` is byte-identical at the merge base, and the same payload reached it on `main` via `rewrite_rsc_scripts_combined_with_limit`. It is raised here because this PR materially widens the exposure — `classify_rsc_group` now runs the scanner on borrowed publisher fragments *inside a `lol_html` text callback* (`rsc_placeholders.rs:114`), which is earlier and more reachable than the old end-of-document post-processor, and only needs `__next_f.push([1,"` plus an `N:TX,` prefix to trigger.

**Proposed fix (apply manually — three separate slices outside this PR's hunks):**

```rust
// \x: verify the advanced position is a char boundary before continuing.
if esc == b'x' && self.pos + 3 < self.bytes.len() {
    if !self.str_ref.is_char_boundary(self.pos + 4) {
        return None;
    }
    self.pos += 4;
    return Some(EscapeElement { byte_count: 1 });
}

// \u: use get() so a non-boundary end yields None instead of panicking.
if esc == b'u' && self.pos + 5 < self.bytes.len() {
    let Some(hex) = self.str_ref.get(self.pos + 2..self.pos + 6) else {
        return None;
    };
    // … and likewise for hex2 at self.pos + 8..self.pos + 12
```

with the same `get()` treatment for `self.str_ref[self.pos..]` at line 156. A `None` here should surface as `TChunkScan::Invalid`, which the streaming path already handles by restoring originals unchanged — the hydration-safe fallback this design already commits to.

Entirely reasonable to split this into a follow-up issue since it predates the PR; flagging it on the changed entry point so the decision is explicit rather than implicit.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -646,6 +647,12 @@ impl PublisherBodyProcessor {
         let is_html = is_html_content_type(&params.content_type);
         let is_rsc_flight =
             content_type_contains_ascii_case_insensitive(&params.content_type, "text/x-component");
+        let inline_seam_token = deferred_inline_seam_token(
```
aram356, line 655: 🤔 **thinking** — `PublisherBodyProcessor::new` mints the seam token as a side effect of construction and installs it as `deferred_inline_marker`, so `lol_html` will emit `<!--ts-inline-body-close-…-->` at the structural `</body>`. Whether that token is ever *removed* depends entirely on the caller remembering to call `take_inline_seam_token()` and wire the result into an `AuctionHoldState`. Only the lazy Fastly path does; `process_response_streaming_async` constructs the same processor and hands it to `process_body_chunks_async`, which has no seam controller — every processed byte goes straight to the writer.

I confirmed this is **not live today**: both call sites of `process_response_streaming_async` are reached only when `dispatched_auction` is `None` or `is_html` is false, so the token is `None` in both. No action needed for correctness right now.

The concern is that the invariant is invisible at the construction site and unenforced by types. Failure scenario for a future change: any caller that routes an HTML body with a pending auction through `process_response_streaming_async` ships the raw comment token to the browser *and* injects zero bids — silently, with no error and no telemetry signal, since the auction still collects normally at EOF.

Returning the token from the constructor rather than stashing it — `let (processor, seam_token) = PublisherBodyProcessor::new(…)` — would make discarding it a visible `let _ =` at the call site instead of an invisible omission. The design doc calls out this exact hazard ("This prevents any constructor that lacks a seam controller from producing an unresolved token"), so tightening it here would close the gap the spec anticipated.

crates/trusted-server-core/src/integrations/nextjs/rsc_stream.rs
```diff
@@ -0,0 +1,1031 @@
...
+                self.state
+                    .lock()
+                    .unwrap_or_else(std::sync::PoisonError::into_inner)
+                    .bypass_rsc = true;
+                self.release_group(None).map(Some)
+            }
+        }
+    }
+
+    fn release_bypass(&mut self, current: &[u8]) -> io::Result<Vec<u8>> {
```
aram356, line 375: 🤔 **thinking** — `release_bypass` drains **every** captured payload from document state and passes them all to `substitute_payloads`, which hard-errors when any placeholder is absent from the bytes it was given:

```
Err(Custom { kind: Other, error: "Next.js RSC captured placeholder is missing from held output" })
```

Reproduced at unit level: two payloads captured, the first classifies `Invalid` (setting bypass) while the second placeholder has not yet reached the output stream. The first `process_chunk` returns that error, and the next returns `"Next.js RSC generated placeholder remained after substitution"`. Both propagate as `io::Error` and abort the response — the opposite of the byte-preserving fallback this design promises for invalid RSC data.

Being straight about reachability: **I could not trigger this end-to-end.** Capture and placeholder emission are atomic within a single `rewrite_complete` call (`rsc_placeholders.rs:80-87`), so the parser cannot get ahead of the output stream, and two separate e2e attempts through `StreamingPipeline` both completed cleanly. So this reads as defense-in-depth that is currently unreachable rather than a live bug.

Worth either a comment recording *why* the invariant holds (capture and emission are atomic, so a drained payload is always present in the current or held bytes), or making the drain tolerant — skip placeholders not found in the current bytes and leave them queued for a later chunk — so the fallback degrades to unchanged bytes rather than a 500 if that atomicity ever changes. The same shape appears via `resolve_group`'s segment-limit branch at `rsc_stream.rs:288-298`, which sets `bypass_rsc` and calls `release_group` without draining, then relies on the loop's bypass check to call `release_bypass` with only the remainder of the current chunk.

crates/trusted-server-core/src/integrations/nextjs/rsc_placeholders.rs
```diff
@@ -38,6 +22,108 @@ impl NextJsRscPlaceholderRewriter {
...
+            FragmentCapture::Restore(restored) => {
+                state.bypass_rsc = true;
+                ScriptRewriteAction::replace(restored)
+            }
+            FragmentCapture::PassThrough => {
+                if is_last && content.len() > limit && content.contains("__next_f") {
+                    let unsafe_continuation = find_rsc_push_payload_range(content)
+                        .map(|(start, end)| {
+                            matches!(
+                                classify_rsc_group(&[&content[start..end]], usize::MAX),
```
aram356: ♻️ **refactor** — This inspection passes `usize::MAX` as the combined-payload bound, so the configured `max_combined_payload_bytes` is bypassed at exactly the point where the work happens inside a `lol_html` text callback. The only remaining guard is `MAX_REASONABLE_TCHUNK_LENGTH` (100 MB) in `scan_tchunks_impl`.

Failure scenario: an over-limit one-fragment script declaring `1:T5F5E100,` (99 MB) with short actual content makes `consume_unescaped_bytes` walk the entire fragment escape-by-escape before returning `NeedMore`. Bounded by fragment length, so not unbounded — but it is CPU spent inside the parser callback on data the configured limit already said was too large to process.

`limit` is already in scope in this function, and passing it here costs nothing while making the bound uniform with the stream-side call in `rsc_stream.rs:303`. The classification result is unchanged for any group that is actually within bounds.

_(compile-verified only — please re-run the matching `cargo test-*` alias after applying)_

crates/trusted-server-integration-tests/tests/parity.rs
```diff
@@ -919,3 +933,259 @@ async fn legacy_admin_aliases_are_denied_locally_not_proxied() {
...
+            response.status(),
+            200,
+            "{adapter} should serve fixture HTML"
+        );
+        let body = response
+            .into_body()
+            .into_bytes()
+            .expect("should buffer adapter output");
+        let html = String::from_utf8(body.to_vec()).expect("should emit UTF-8 HTML");
+        assert_eq!(
```
aram356, line 973: ⛏ **nitpick** — One `AtomicUsize` is shared across all three sequential router runs and asserted positionally as `index + 1`. It passes because the loop is sequential, but the assertion cannot distinguish per-adapter behaviour from the running total: if Axum dispatched two auctions and Cloudflare zero, the three checkpoints would still read 1, 2, 3 and the test would stay green — while a genuine parity violation went unreported.

`client.auction_requests.store(0, Ordering::SeqCst)` before each request plus `assert_eq!(…, 1)` makes each adapter's count independently meaningful and removes the positional coupling.

crates/trusted-server-integration-tests/tests/parity.rs
```diff
@@ -919,3 +933,259 @@ async fn legacy_admin_aliases_are_denied_locally_not_proxied() {
...
+            suffix < bids && bids < close,
+            "{adapter} should inject bids at the structural body close"
+        );
+        assert!(
+            html.contains("parity-creative"),
+            "{adapter} should include deterministic auction creative"
+        );
+        assert!(
+            html[bids..].ends_with("</script></body></html>"),
+            "{adapter} should place bid markup immediately before the body close"
```
aram356, line 1036: ⛏ **nitpick** — `expected_html` is seeded from whichever adapter runs first (Axum) and the other two are compared against it, so this assertion proves the three adapters *agree* rather than that any of them is right. The plan (Task 7, Step 5) asked for a comparison against the core expected bytes.

In practice the important bytes are pinned independently just above — the reconstructed Flight payload with its recomputed `T` length, script ordering, and the `</script></body></html>` tail — so a shared-core regression would still be caught by those. But if core output changed for all three adapters identically, this particular `assert_eq!` stays green. Worth a one-line comment saying that explicitly, so a future reader does not mistake it for an absolute oracle.

aram356 (changes requested):
## Summary

Parser-confirmed body seam and bounded Next.js streaming, verified by execution rather than inspection: with the Next.js processor registered a non-final chunk emits real output pre-EOF, and a false `</body>` literal inside RSC data now streams past while the auction is still pending. `BodyCloseHoldBuffer` and every auction-coordination scan for `</body` are gone, and all 19 CI checks pass.

Two findings block. The `RSC_PUSH_CALL_PATTERN` widening changes behaviour for scripts that are not Next.js RSC at all, and the T-chunk escape scanner panics on publisher-controlled input reachable through the new classification path. Neither comment carries a one-click `suggestion`: I applied the obvious regex narrowing in a scratch worktree and it **broke** `html_processor_rewrites_rsc_stream_payload_with_chunked_input`, so the fix is described in prose instead of proposed as bytes I could not stand behind.

## Blocking

### 🔧 wrench

- **Widened RSC push pattern rewrites unrelated publisher scripts** — see inline at `crates/trusted-server-core/src/integrations/nextjs/shared.rs:21`
- **Escape scanner panics on non-char-boundary indices** — see inline at `crates/trusted-server-core/src/integrations/nextjs/rsc.rs:210`

## Non-blocking

### 🤔 thinking

- **Deferred seam token is minted by construction but claimed by convention** — see inline at `crates/trusted-server-core/src/publisher.rs:650`
- **`release_bypass` requires every captured placeholder in the current chunk** — see inline at `crates/trusted-server-core/src/integrations/nextjs/rsc_stream.rs:346`

### ♻️ refactor

- **Parser-side classification passes `usize::MAX` instead of the configured bound** — see inline at `crates/trusted-server-core/src/integrations/nextjs/rsc_placeholders.rs:114`

### ⛏ nitpick

- **Parity test shares one auction counter across adapters** — see inline at `crates/trusted-server-integration-tests/tests/parity.rs:1112`
- **Parity test uses Axum as its own oracle** — see inline at `crates/trusted-server-integration-tests/tests/parity.rs:1176`

- 🏕 **`html_post_process.rs` is dead surface, and the PR description overstates its removal.** The description says the whole-document post-processor was removed; 231 lines went, but 674 remain. `post_process_rsc_html` and `post_process_rsc_html_in_place` have **zero callers anywhere in `crates/`** outside this module's own tests (verified by workspace grep) — roughly 150 lines of production code including a second full `lol_html` re-parse in `find_rsc_push_scripts`, plus ~400 lines of test module, all compiled into the wasm build behind `#[allow(deprecated)]`. Their doc comments still point at `NextJsHtmlPostProcessor`, which this PR deletes, so the surviving documentation describes a type that no longer exists. There are no callers left to migrate, so CLAUDE.md's "migrate callers and delete legacy APIs" applies cleanly. Deleting the module and the `mod.rs` re-export would drop ~674 lines of wasm-compiled dead weight; at minimum the stale doc comments need fixing.

- 📝 **The adapter regressions landed in a different crate than the plan specifies, leaving three CI gates blind to this path.** The plan (`docs/superpowers/plans/2026-09-07-850-parser-aware-body-hold-nextjs-streaming.md:839-891`, Task 7 Steps 5 & 7) prescribes "one buffered route regression in each adapter test module", lists `crates/trusted-server-adapter-{axum,cloudflare,spin}/tests/routes.rs` under **Modify**, and gives per-adapter verification commands. None of those three files changed — `git diff --name-only` shows zero `tests/routes.rs` in this PR. The single fixture lives only in `crates/trusted-server-integration-tests/tests/parity.rs`. Consequence: `cargo test-axum`, `cargo test-cloudflare`, and `cargo test-spin` (CI gate 3) do not exercise the Next.js/auction path at all, and the new `pub fn routes_with_settings_and_services` added to all three adapters is dead code within its own crate — its only consumer is the parity suite. The plan's completion review does disclose the relocation, but the Step 5/7 recipe and the **Modify** list were never corrected, so the plan still claims coverage that does not exist. Either move a fixture into each adapter's own tests, or amend the plan and note explicitly that per-adapter gates do not cover this path.

- 📌 **Escape-scanner panic is pre-existing on `main`, not introduced here — flagged because this PR widens its reach.** I verified `EscapeSequenceIter` is byte-identical at the merge base and that the same payload reached it there through `rewrite_rsc_scripts_combined_with_limit`, so this is not a regression and is legitimately out of scope for a fix in this PR. It is called out because the new `classify_rsc_group` call site runs the scanner on borrowed publisher fragments *inside a `lol_html` callback*, which is a materially wider and earlier exposure than the old post-processor path. A follow-up issue is the right home if you would rather not grow this PR.

## CI Status

- integration tests: PASS
- CodeQL: PASS
- cargo test: PASS (required)
- vitest: PASS
- Analyze (rust): PASS
- cargo fmt: PASS (required)
- Analyze (actions): PASS

crates/trusted-server-core/src/integrations/nextjs/rsc_stream.rs
prk-Jr: Fixed in `05eb493d`. `RscGroupClassifier` now retains state: each payload is appended and scanned once, chunk-content consumption and non-chunk segment inspection resume from `scan_from`/`segment_start`, and EOF reclassifies with nothing held back. The 8.96 MiB / 256-segment case went from ~2.4s to ~19ms, and the cost no longer scales with segment count.

crates/trusted-server-core/src/integrations/nextjs/shared.rs
prk-Jr: Fixed in `05eb493d`. `RSC_PUSH_CALL_PATTERN` requires a `self.`/`window.` receiver again. The streamed-away-receiver case is handled by a separate `RSC_PUSH_CALL_PATTERN_TRIMMED`, anchored to the start of the claimed fragment and only reachable once `receiver_context_is_flight_push` has proven the receiver out of band. A match whose preceding byte continues a member expression or identifier is rejected, so `myAnalytics.__next_f`, `foo.window.__next_f`, and `a__next_f` no longer match.

crates/trusted-server-core/src/integrations/nextjs/rsc_placeholders.rs
prk-Jr: Fixed in `05eb493d`. `max_combined_payload_bytes` bounds a single script and the downstream unresolved group, not the parser queue in aggregate. The parser-to-processor queue is bounded separately by the parser's own script-buffer budget plus a payload count, so two independent sub-limit payloads arriving in one source chunk rewrite instead of bypassing each other.

crates/trusted-server-core/src/integrations/nextjs/shared.rs
prk-Jr: Fixed in `05eb493d`. Your read on why the naive narrowing broke `html_processor_rewrites_rsc_stream_payload_with_chunked_input` was right — the bare form was load-bearing for a receiver that streams away before `__next_f` is recognised. That is now handled by retaining the receiver in a bounded context (`RSC_RECEIVER_CONTEXT_BYTES`) on the document state and using the anchored trimmed pattern only once that context proves it, rather than relaxing the main pattern. A match whose preceding byte continues a member expression or identifier is also rejected.

crates/trusted-server-core/src/integrations/nextjs/rsc.rs
prk-Jr: Fixed in `05eb493d`. All three sites stay on character boundaries now: `\x` is gated on `is_char_boundary(self.pos + 4)`, both `\u` reads use `str::get` so a body that straddles a character falls through to literal handling exactly as invalid hex does, and the non-ASCII tail advances one byte instead of slicing. Only the boundary is validated, not the hex digits, so the unescaped byte count that drives T-chunk length recomputation is unchanged for input that previously scanned.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Addressed in `05eb493d`. `PublisherBodyProcessor::new` now returns the deferred inline seam token, so a caller without a seam controller has to discard it visibly, and dropping one logs a warning — a debug assertion would be compiled out of the wasm builds that ship.

crates/trusted-server-core/src/integrations/nextjs/rsc_stream.rs
prk-Jr: Agreed it is unreachable today, and `05eb493d` documents why rather than changing behaviour: capture and placeholder emission are atomic inside `rewrite_complete`, so a queued payload's placeholder is always already in the held output or in `current`. Left as a hard error rather than a silent degrade, so a future change that breaks that atomicity surfaces instead of shipping wrong bytes.

crates/trusted-server-core/src/integrations/nextjs/rsc_placeholders.rs
prk-Jr: Fixed in `05eb493d` — the in-callback classification passes `limit`, so the bound is uniform with the stream-side call in `rsc_stream.rs`. Gate re-run, green.

crates/trusted-server-integration-tests/tests/parity.rs
prk-Jr: Fixed in `05eb493d`. `client.reset_auction_requests()` runs before each adapter and the assertion is `assert_eq!(client.auction_requests(), 1)`, so each adapter's count is independently meaningful and the positional coupling is gone.

crates/trusted-server-integration-tests/tests/parity.rs
prk-Jr: Addressed in `05eb493d` with the comment you suggested, stating this is cross-adapter agreement only and that the bytes that matter — the reconstructed Flight payload with its recomputed `T` length, script order, and the body-close tail — are pinned absolutely above. The fixture also moved into `test_support`, so each adapter now carries its own buffered regression in its own `tests/routes.rs` and `cargo test-{axum,cloudflare,spin}` exercise this path.

ChristianPavilonis (approved):
## Summary

Reviewed `481b24f7de7d7e6e3bb55d7fe59297383a2f95e1` against `6cae7f5da8911c746cf873581885f90c3820dd96`. The parser-confirmed auction seam and bounded Next.js RSC streaming paths are covered by focused WASM tests across supported encodings, and all reported CI checks pass. No actionable issues found.

crates/trusted-server-core/src/integrations/nextjs/rsc.rs
```diff
@@ -189,83 +206,161 @@ fn consume_unescaped_bytes(s: &str, start_pos: usize, byte_count: usize) -> (usi
...
 
-            search_pos = content_end;
-        } else {
+    let mut iter = match marker {
+        Some(marker) => EscapeSequenceIter::from_position_with_marker(content, chunk.pos, marker),
+        None => EscapeSequenceIter::from_position(content, chunk.pos),
+    };
+    while chunk.consumed < chunk.declared_length {
+        let position = iter.position();
+        if hold_back_partial_escape
```
aram356: 🔧 **wrench** — This guard breaks on *any* backslash within the last 12 bytes of the accumulated text, not on an escape that is actually incomplete. It is evaluated against the whole of `combined`, so a complete, self-contained payload whose final bytes contain ordinary escaped JSON never releases on `push` and waits for EOF.

Measured through `NextJsRscStreamProcessor::process_chunk` (one captured payload, one output chunk `<p>{placeholder}</p>`, 10 MiB limit):

| payload | emitted | `held_output` | group |
| --- | --- | --- | --- |
| `1:T3,abc` | 15 | 0 | 0 (released) |
| `1:T3,ab\n` | 3 | 49 | 1 (held) |
| `1:T9,{\"a\":\"b\"}` | 3 | 49 | 1 (held) |
| `1:T15,{\"a\":\"b\"}` + 12 clean bytes | 38 | 0 | 0 (released) |

Real Flight rows are JSON, so a `\"` or `\n` inside the last 12 bytes is close to guaranteed. Failure scenario: the group stays open past its own payload, `held_output` then absorbs every subsequent body byte until either EOF rescues it via `finalize()`, or `append_held` exceeds `self.limit` and `release_bypass` fires, setting document-wide `bypass_rsc` and dropping RSC rewriting for the remainder of the page. On a document with a substantial body after the first Flight script this converts working rewriting into a document-wide fallback, and in every case it buffers bytes that this PR set out to stream.

**No `suggestion` here, and the reason matters for whatever fix you pick.** This guard is also what currently prevents a classifier/rewriter length disagreement. The classifier scans `combined` without `RSC_MARKER`; the rewriter scans it with markers inserted at payload boundaries. A boundary landing inside an escape changes the unescaped count:

```
calculate_unescaped_byte_length(r"\x41\x42")                = 2
calculate_unescaped_byte_length_skip_markers("\\x41<M>x42")  = 4
```

The `\` before the marker is read as a literal byte and `x42` as three more. A sweep of escape-dense bodies at every two-way split found zero cases where the classifier says `CompleteRewritable` and the rewrite round-trip differs from its input, because the hold-back stops the scan first. So loosening the guard naively reopens a path that ships a wrong `T` header length to the browser, which is a hydration break rather than a safe fallback.

**Suggested direction (apply manually):** bound the hold-back to a genuinely partial escape rather than a fixed window. Hold back only when the trailing backslash run actually reaches the end of `content` and the bytes remaining are fewer than the escape starting at `position` requires, instead of firing on any backslash within `MAX_ESCAPE_SEQUENCE_BYTES` of the end. That keeps the length invariant intact for a truncated escape while letting a complete `\"` or `\n` two bytes from the end release immediately.

crates/trusted-server-core/src/integrations/nextjs/rsc_placeholders.rs
```diff
@@ -38,6 +25,128 @@ impl NextJsRscPlaceholderRewriter {
...
+                            )
+                        })
+                        .unwrap_or(true);
+                    state.bypass_rsc |= unsafe_continuation
+                        || state.captured_payload_bytes > 0
+                        || !state.captured_payloads.is_empty();
+                } else if !is_last && content.len() > limit {
+                    state.bypass_rsc = true;
+                }
+                ScriptRewriteAction::Keep
```
aram356, line 166: 🔧 **wrench** — This path always calls `find_rsc_push_payload_range`, which now requires a qualified `self.`/`window.` receiver. When the active claim is trimmed, `content` begins at a bare `__next_f`, so the match always fails and `.unwrap_or(true)` makes `unsafe_continuation` true, bypassing RSC rewriting for the whole document.

```
find_rsc_push_payload_range(r#"__next_f.push([1,"1:T3,abc"])"#)         = None
find_trimmed_rsc_push_payload_range(r#"__next_f.push([1,"1:T3,abc"])"#) = Some((18, 26))
```

Failure scenario: fragment A is `self.` (released, remembered as receiver context), fragment B is a trimmed claim `__next_f.push([1,"..."])` longer than `max_combined_payload_bytes`. `capture_fragment` returns `PassThrough`, this line uses the qualified pattern, gets `None`, and sets `bypass_rsc`. Before the receiver hardening the widened pattern matched the bare form here, so only that one oversized script lost rewriting; now the rest of the document loses it too.

`rewrite_complete` already selects the right pattern on `state.rsc_receiver_trimmed` at lines 45-49; this arm just needs the same selection. The replacement below also clears `rsc_receiver_trimmed` on this path, which is the one arm that currently leaves it set (lines 50, 126, and 180 all clear it) — otherwise the flag can survive into the next script element and make an ordinary unfragmented `self.__next_f.push(...)` match the anchored pattern, fail, and silently skip capture.

Verified in an isolated scratch tree against the full gate: `cargo fmt --all -- --check`, `cargo clippy-fastly`, `cargo clippy-axum`, `cargo test-fastly` (2696 passed), `cargo test-axum`, `cargo test-cloudflare`, `cargo test-spin`, and the cross-adapter parity suite all pass, with a byte-exact drift check confirming these are the bytes that were tested.

```suggestion
            FragmentCapture::PassThrough => {
                // A trimmed claim starts at a bare `__next_f`, which the
                // qualified pattern cannot match. Select the same way
                // `rewrite_complete` does, or every over-limit trimmed claim
                // falls through to `unwrap_or(true)` and bypasses the document.
                let trimmed = state.rsc_receiver_trimmed;
                state.rsc_receiver_trimmed = false;
                if is_last && content.len() > limit && content.contains("__next_f") {
                    let range = if trimmed {
                        find_trimmed_rsc_push_payload_range(content)
                    } else {
                        find_rsc_push_payload_range(content)
                    };
                    let unsafe_continuation = range
                        .map(|(start, end)| {
                            matches!(
                                classify_rsc_group(&[&content[start..end]], limit),
                                RscGroupStatus::NeedMore | RscGroupStatus::Invalid
                            )
                        })
                        .unwrap_or(true);
                    state.bypass_rsc |= unsafe_continuation
                        || state.captured_payload_bytes > 0
                        || !state.captured_payloads.is_empty();
                } else if !is_last && content.len() > limit {
                    state.bypass_rsc = true;
                }
                ScriptRewriteAction::Keep
            }
```

crates/trusted-server-core/src/integrations/nextjs/rsc_stream.rs
```diff
@@ -0,0 +1,1306 @@
...
+            let per = document.len().div_ceil(segments);
+            let payloads: Vec<&str> = document
+                .as_bytes()
+                .chunks(per)
+                .map(|chunk| std::str::from_utf8(chunk).expect("fixture should be ASCII"))
+                .collect();
+
+            let mut classifier = RscGroupClassifier::new(usize::MAX);
+            for payload in &payloads {
+                classifier.push(payload);
```
aram356, line 1398: 🤔 **thinking** — Two points about this test and the property it is meant to guarantee.

**The assertion is tautological.** `classify_rsc_group` (`rsc_stream.rs:755-764`) now constructs an `RscGroupClassifier`, pushes each payload, and finalizes, which is byte for byte what the left-hand side of this assertion does. It cannot detect an incremental/full-rescan divergence; only the `== CompleteRewritable` assert carries information. The neighbouring `classification_is_independent_of_payload_split` is a genuine property test and is fine. Getting the guarantee the commit message claims needs an independent oracle: `scan_tchunks` over the concatenation plus a fresh `SegmentInspector`, i.e. the pre-`05eb493d4` logic. Worth saying plainly that my own first differential pass missed the divergence below for exactly this reason, because I used the in-tree function as the reference.

**With a real oracle there is one disagreement class.** Reimplementing the pre-`05eb493d4` full rescan as the reference and diffing it against the incremental classifier surfaces `incremental = Invalid` where `full rescan = NeedMore`. Minimal case, the document `a:T3:Te,` delivered as three payloads:

```
["a", ":T3:", "Te,"]
  incremental: push("a") -> NeedMore; push(":T3:") -> Invalid (latched); finalize() -> Invalid
  full rescan: NeedMore
```

Cause: the trailing segment is inspected on every `push` (`rsc_stream.rs:722-724`). On the intermediate view `a:T3:`, `SegmentInspector::inspect` reaches `:T`, a hex run, then a visible non-comma, and returns `Invalid`, which latches via `self.invalid = true`. The old code ran `scan_tchunks` over the whole text first, found the header `3:Te,`, and short-circuited to `NeedMore` before that text was ever inspected as a non-chunk segment.

Two things bound how much this matters, which is why this is not filed as blocking. Across an exhaustive sweep (alphabet `{1, :, T, ',', a, x}`, lengths 3..=9, every two- and three-way split, ~9.4M classifier runs) there was **no** case where the full rescan reported `CompleteRewritable` or `CompleteUnrewritable` and the incremental reported `Invalid` — so no group the old code would have rewritten is lost. And no realistic Flight shape reproduced it (`3:T5,hello 4:T5,world`, `a:T4,abcd b:T4,efgh`, `1:HL[...]\n2:T4,abcd`, `2:T4,abcd3:T4,efgh` all had zero bad splits). The cost when it does fire is a premature latched `Invalid`, which sets document-wide `bypass_rsc` and silently drops RSC rewriting for the rest of the page. Hydration-safe, but a real loss of function from input that is merely incomplete rather than malformed.

crates/trusted-server-core/src/integrations/nextjs/shared.rs
```diff
@@ -13,19 +13,83 @@ use crate::host_rewrite::rewrite_bare_host_at_boundaries;
 // intentionally remain lazy statics instead of participating in
 // `Settings::prepare_runtime`.
 /// RSC push script call pattern for extracting payload string boundaries.
+///
+/// The `self.`/`window.` receiver is required. A fragmented script keeps its
+/// receiver out of the output stream via [`RSC_RECEIVER_CANDIDATES`] instead of
```
aram356: ⛏ **nitpick** — `[`RSC_RECEIVER_CANDIDATES`]` does not resolve; grepping `crates/` finds this line and nothing else. The mechanism this doc comment is describing is `receiver_context_is_flight_push` together with `RSC_RECEIVER_CONTEXT_BYTES`. `cargo doc` will warn on the unresolved intra-doc link.

crates/trusted-server-core/src/integrations/registry.rs
```diff
@@ -549,26 +551,22 @@ pub struct IntegrationHtmlContext<'a> {
...
-/// Trait for integration-provided HTML post-processors.
-/// These run after streaming HTML processing to handle cases that require
-/// access to the complete HTML (e.g., cross-script RSC T-chunks).
-pub trait IntegrationHtmlPostProcessor: Send + Sync {
-    /// Identifier for logging/diagnostics.
-    fn integration_id(&self) -> &'static str;
+/// Owned request data supplied when an integration creates an HTML stream processor.
+#[derive(Clone)]
+pub struct IntegrationHtmlStreamContext {
+    pub request_host: String,
```
aram356, line 558: ⛏ **nitpick** — The four new public fields (`request_host`, `request_scheme`, `origin_host`, `document_state`) have no doc comments, against the "each public item must have a doc comment" rule in CLAUDE.md. The struct itself is documented. This matches the pre-existing `IntegrationScriptContext` alongside it, so it reads as convention drift rather than a regression introduced here, but this is new public API surface.

aram356 (changes requested):
## Summary

Second review pass, against head `481b24f7d`. Every finding from the previous pass is addressed, and I re-verified the two blocking ones by running the original repros rather than reading the diff: all four regex false positives now return `None` with every legitimate receiver form still matching, and all three previously-panicking escape inputs now scan cleanly. The seam-token constructor change is a better fix than the one proposed. `html_post_process.rs` is fully deleted, each adapter now carries its own buffered regression that genuinely runs under its own CI alias, and the plan doc records each deviation instead of silently dropping it. I also reproduced the classifier performance claim independently: 8.75 MiB across 256 segments classifies in 26ms.

The hardening commit introduced two new defects, both in the fixes themselves. Neither corrupts a response — both degrade to the hydration-safe fallback — but the first one undoes the streaming behaviour this PR exists to deliver, so it should not merge as-is.

One of the inline comments carries a one-click `suggestion`; it passed the full CI-equivalent gate in isolation (`cargo fmt`, `clippy-fastly`, `clippy-axum`, `test-fastly`, `test-axum`, `test-cloudflare`, `test-spin`, and the cross-adapter parity suite), with a byte-exact drift check confirming the posted bytes are the bytes that were tested. The other blocking finding is prose-only on purpose: its fix interacts with a length-counting invariant described in that comment, and proposing bytes I had not proven safe would be worse than describing the constraint.

## Blocking

### 🔧 wrench

- **Escape hold-back defers every realistic Flight payload, defeating streaming** — see inline at `crates/trusted-server-core/src/integrations/nextjs/rsc.rs:302`
- **Over-limit trimmed claim always forces document-wide RSC bypass** — see inline at `crates/trusted-server-core/src/integrations/nextjs/rsc_placeholders.rs:132` _(carries a suggestion)_

## Non-blocking

### 🤔 thinking

- **Incremental classifier can latch `Invalid` where a full rescan reports `NeedMore`** — see inline at `crates/trusted-server-core/src/integrations/nextjs/rsc_stream.rs:1224`

### ⛏ nitpick

- **Split-invariance test compares the classifier against itself** — see inline at `crates/trusted-server-core/src/integrations/nextjs/rsc_stream.rs:1224`
- **Broken intra-doc link to a symbol that does not exist** — see inline at `crates/trusted-server-core/src/integrations/nextjs/shared.rs:18`
- **New public context fields carry no doc comments** — see inline at `crates/trusted-server-core/src/integrations/registry.rs:557`

- 🌱 **Lazy Fastly disconnect still records no abandonment telemetry, and this PR is not the cause.** A client disconnect mid-stream on the lazy path leaves `DispatchedAuctionGuard::drop` logging a `log::warn!` with no terminal `abandoned` row, so the SSP fan-out and its quota consumption vanish from telemetry. The write-sink driver now handles this correctly via the new `abandon_hold_auction(..., "stream_write_error")` wrapper. I want to be precise about scope, because the asymmetry invites the wrong conclusion: `origin/main` carries the byte-identical warn-only `Drop` impl, and `stream_write_error` does not appear anywhere on `main`. So this PR **added** abandonment telemetry to one driver and left the other exactly as it found it. That is a net improvement, not a regression, and it should not gate this PR. Worth a follow-up issue, with the note that a generator drop is not an `Err`, so the write-sink's `is_err()` branch will not port across directly: it needs a guard-side terminal emit or an explicit drop hook.

- 📝 **Worst-case concurrent RSC memory now compounds across four buffers.** The captured-payload queue bound moved from `max_combined_payload_bytes` (10 MiB default) to `ctx.max_buffered_script_bytes` (`publisher.max_buffered_body_bytes`, 16 MiB default), and `RscGroupClassifier` now retains `combined` for the lifetime of a group, which is a second copy of bytes already held in `self.group` (the previous `classify_rsc_group` built and dropped that string per call). Worst case, concurrently live: roughly 16 MiB queue + 10 MiB `held_output` + 10 MiB `combined` + 10 MiB `group`. I cannot demonstrate an OOM, and each individual bound is defensible, but the two changes compound and the total is worth confirming against the Fastly Compute heap budget rather than discovering at peak traffic.

- 📌 **`migration_guards.rs`'s allowlist is hand-maintained and silently incomplete.** Out of scope here and not caused by this PR, which correctly adds its one new core file (`rsc_stream.rs`) and removes the deleted one. Flagging because the guard's value depends on the list: roughly two dozen core modules are absent from it (`auction/plan.rs`, `config.rs`, `platform/template_cache.rs`, `price_bucket.rs`, among others), so a future file added without a list entry is unguarded with no failing test to say so. A directory-walk-based guard would remove the maintenance burden. None of the missing modules were touched by this PR.

## CI Status

All 20 checks reported by `gh pr checks` are PASS, including the four required gates. I also re-ran `cargo fmt --all -- --check` and `cargo clippy-fastly` against a clean export of this head, both clean, and `cargo test -p trusted-server-core --lib` reports 2702 passed / 0 failed.

- integration tests: PASS
- CodeQL: PASS
- cargo test: PASS (required)
- vitest: PASS
- cargo fmt: PASS (required)
- Analyze (rust): PASS
- Analyze (actions): PASS

crates/trusted-server-core/src/integrations/nextjs/rsc.rs
prk-Jr: Fixed in 66e17c480. The scanner now holds only escape prefixes that can still change their decoded length, including a high surrogate that can still join a low surrogate. Complete escapes and escaped JSON release on `process_chunk(..., false)`.

The marker invariant has a separate safeguard: when a script boundary interrupts an escape, combined rewriting preserves that group unchanged rather than inserting a marker into the escape and recomputing a different T length. Tests cover each escape split, identity-rewrite length preservation, immediate release, and subsequent body output larger than the hold limit without document bypass.

Local validation passed: all six target-specific Clippy aliases, all four adapter test aliases, 14 parity tests, 2,709 native core tests, 959 JS tests, JS build, and Rust/JS/docs formatting.

crates/trusted-server-core/src/integrations/nextjs/rsc_placeholders.rs
prk-Jr: Fixed in 66e17c480. The pass-through arm selects the trimmed matcher when the receiver was verified from released context, and clears `rsc_receiver_trimmed` before returning.

Added a regression that releases `self.`, passes through an oversized but complete trimmed claim, checks that document bypass and the trimmed flag are both clear, then verifies that the next ordinary qualified script is captured. The full local validation gates pass.

crates/trusted-server-core/src/integrations/nextjs/rsc_stream.rs
prk-Jr: Both points are addressed in 66e17c480. The reference helper now concatenates payloads, calls `scan_tchunks`, inspects non-chunk segments with fresh `SegmentInspector` instances, and checks header boundaries independently of `RscGroupClassifier`. The existing many-segment test uses that oracle; additional fixtures compare every two-/three-way split.

Malformed non-chunk text remains provisional while another payload can introduce a pending chunk. The exact `["a", ":T3:", "Te,"]` regression stays `NeedMore` without enabling document bypass. Malformed groups are restored unchanged at EOF or the existing resource limits; this deliberately delays their fallback, while complete valid groups still release immediately. Hard chunk errors and size overflow retain immediate rejection. The behavior change and bounded rescanning tradeoff are recorded in the plan. All local validation gates pass.

crates/trusted-server-core/src/integrations/nextjs/shared.rs
prk-Jr: Fixed in 66e17c480. The comment now describes retained released receiver context and links to `RSC_RECEIVER_CONTEXT_BYTES` and `receiver_context_is_flight_push`. `cargo doc -p trusted-server-core --no-deps --all-features` succeeds; unrelated documentation warnings remain elsewhere.

crates/trusted-server-core/src/integrations/registry.rs
prk-Jr: Fixed in 66e17c480. Added doc comments for all four `IntegrationHtmlStreamContext` fields, describing the publisher-facing host/scheme, origin host, and request-local shared document state. The core documentation build succeeds.

crates/trusted-server-core/src/integrations/nextjs/rsc_placeholders.rs
```diff
@@ -38,6 +25,135 @@ impl NextJsRscPlaceholderRewriter {
...
+                        .map(|(start, end)| {
+                            matches!(
+                                classify_rsc_group(&[&content[start..end]], limit),
+                                RscGroupStatus::NeedMore | RscGroupStatus::Invalid
+                            )
+                        })
+                        .unwrap_or(true);
+                    state.bypass_rsc |= unsafe_continuation
+                        || state.captured_payload_bytes > 0
+                        || !state.captured_payloads.is_empty();
```
dhruv8sh, line 162: 🔧 **wrench** — The oversized-fragment safety check only inspects the *first* `__next_f.push(...)` call in a fragment, so a second incomplete push call in the same oversized script is never classified.

```rust
let range = if trimmed {
    find_trimmed_rsc_push_payload_range(content)
} else {
    find_rsc_push_payload_range(content)
};
let unsafe_continuation = range
    .map(|(start, end)| {
        matches!(
            classify_rsc_group(&[&content[start..end]], limit),
            RscGroupStatus::NeedMore | RscGroupStatus::Invalid
        )
    })
    .unwrap_or(true);
```

`find_rsc_push_payload_range`/`find_trimmed_rsc_push_payload_range` (`shared.rs:73`, `:88`) both call `RSC_PUSH_CALL_PATTERN.captures(script)` — a single, first-match lookup. There is no loop over multiple `__next_f.push(...)` calls anywhere in this file. Next.js commonly batches several `self.__next_f.push(...)` calls inside one inline `<script>`. If that combined script exceeds `max_combined_payload_bytes` and contains more than one push call, only the first call's payload is ever classified as safe/unsafe here — a second push whose T-chunk is genuinely incomplete and continues into a later `<script>` is invisible to this check, so `bypass_rsc` stays `false` and that continuation gets captured and rewritten later as if it were a fresh, self-contained payload — corrupting its `T` header length. This is exactly the hydration-breaking corruption class this parser exists to prevent, and it's reachable with entirely ordinary content (no attacker needed) on any page whose batched script exceeds the configured limit.

Reproduced directly: with `max_combined_payload_bytes = 100`, a script containing `self.__next_f.push([1,"1:T3,abc"]);self.__next_f.push([1,"2:T64,<80 x's>"])` (first push complete, second push short of its declared 100-byte T-chunk) leaves `bypass_rsc == false`, when it should be `true`.

_No suggestion block: the fix needs the classification loop to scan every push call in the fragment, not a single-line change._

dhruv8sh (changes requested):

crates/trusted-server-core/src/integrations/nextjs/rsc_placeholders.rs
prk-Jr: Fixed in bd18e237d. The reported reproduction failed before the fix: the complete first push left `bypass_rsc` false despite the incomplete second payload.

The oversized-fragment check now scans subsequent push calls, bypassing on incomplete or invalid payloads. It advances past each payload string and permits a trimmed receiver only for the first call.

The regression covers incomplete, invalid, and complete second payloads with both intact and previously streamed receivers, and verifies whether later scripts remain capturable.

Validation: 110 Next.js tests passed on native and Fastly. The full repository CI gate list also passed: formatting, all six adapter Clippy checks, all four adapter test suites, parity tests, JS build/tests, and JS/docs formatting.

dhruv8sh (approved):

crates/trusted-server-core/src/integrations/nextjs/rsc.rs
```diff
@@ -86,6 +86,44 @@ impl<'a> EscapeSequenceIter<'a> {
     fn position(&self) -> usize {
         self.pos
     }
+
+    /// Whether appending bytes could change how the next escape is decoded.
+    fn has_partial_escape(&self) -> bool {
```
aram356: 📝 **note** — Recording the verification for this replacement, since it is the fix for the previous pass's blocking finding and the reasoning is easy to lose.

Behaviour confirmed by driving `NextJsRscStreamProcessor::process_chunk` directly with one captured payload and a single output chunk:

| payload | emitted | `held_output` | group |
| --- | --- | --- | --- |
| `1:T3,abc` | 15 | 0 | 0 (released) |
| `1:T3,ab\n` | 16 | 0 | 0 (released) |
| `1:T9,{\"a\":\"b\"}` | 25 | 0 | 0 (released) |
| `1:T3,ab\x` | 3 | 49 | 1 (held) |
| `1:T3,ab\u1` | 3 | 49 | 1 (held) |
| `1:T4,\ud83d` | 3 | 49 | 1 (held) |
| `1:T4,😀` | 24 | 0 | 0 (released) |

Escaped JSON and complete surrogate pairs release; only genuinely growable prefixes hold. Previously the middle rows emitted 3 bytes and held 49.

The surrogate branch is the part I most expected to be off by one, and it is exact. Enumerating all 65,536 code points, `\uDC00`-`\uDFFF` maps precisely to `{d,D}` at index 2 and `{c,d,e,f}` at index 3: zero non-low-surrogates wrongly accepted as growable, zero low surrogates wrongly rejected. A false reject here would have released early and drifted the decoded byte count.

The marker invariant this guard also protects did not reopen: roughly 1.05M `CompleteRewritable` marker round-trips against the real `RscUrlRewriter` across four request hosts, plus exhaustive two- and three-way splits over escape-dense bodies, produced zero T-length inconsistencies. A ~600k-body sweep also confirmed no prefix holds forever and that the classifier and the iterator never disagree about whether a sequence is an escape.

crates/trusted-server-core/src/integrations/nextjs/rsc.rs
```diff
@@ -455,6 +581,24 @@ pub(crate) fn rewrite_rsc_scripts_combined_with_limit(
             .collect();
     }
 
+    // Markers preserve script boundaries, but inserting one inside an escape
```
aram356: 🌱 **seedling** — Not for this PR; noting a coupling worth hardening later.

`&payloads[..payloads.len() - 1]` underflows if `payloads` is ever empty. Today it cannot be: the `is_empty()` and `len() == 1` paths both return before this point, so `len >= 2` holds, and I could not reach an underflow by any route. The invariant is sound as written.

The observation is that the precondition lives in two earlier returns rather than at the slice itself. A future refactor that adds a third early-exit shape, or reorders those guards, converts this into a panic on publisher-controlled input, which on `wasm32-wasip1` traps and becomes a 500 for the page. `payloads.split_last()` or a `let [.., _last] = payloads else { ... }` destructure would make the precondition local instead of inherited, whenever this hunk is next touched.

crates/trusted-server-core/src/integrations/nextjs/rsc_placeholders.rs
```diff
@@ -38,6 +25,148 @@ impl NextJsRscPlaceholderRewriter {
...
+            FragmentCapture::Restore(restored) => {
+                state.rsc_receiver_trimmed = false;
+                state.bypass_rsc = true;
+                ScriptRewriteAction::replace(restored)
+            }
+            FragmentCapture::PassThrough => {
+                let trimmed = state.rsc_receiver_trimmed;
+                state.rsc_receiver_trimmed = false;
+                if is_last && content.len() > limit && content.contains("__next_f") {
+                    let mut remaining = content;
```
aram356: 📝 **note** — Recording verification for the every-push loop, which is the fix for the previous pass's second blocking finding plus the additional defect found while fixing it.

The pattern selection is confirmed correct. On a trimmed claim the qualified matcher returns nothing while the trimmed one matches, which is precisely why the old unconditional call fell through to `unwrap_or(true)` and bypassed the whole document:

```
find_rsc_push_payload_range(r#"__next_f.push([1,"1:T3,abc"])"#)         = None
find_trimmed_rsc_push_payload_range(r#"__next_f.push([1,"1:T3,abc"])"#) = Some((18, 26))
```

`&remaining[end + 1..]` cannot panic. `end` is the closing-quote index returned from inside `while i < bytes.len()` at `bytes[i] == quote`, so `end < len` and `end + 1 <= len`, and the quote is always ASCII `'` or `"`, so `end + 1` is always a character boundary. Checked against scripts with a multi-byte character immediately before the closing quote, with the quote as the final byte (`end + 1 == len` exactly), with single and double quotes, and with a trailing four-byte character: every case boundary-safe, no panic. A wider fuzz over chained pushes, unterminated pushes, empty payloads, and payload text impersonating a push call added ~20k cases and 88 adversarial scripts, also with zero panics.

Resetting `receiver_trimmed` to false after the first iteration is the right model: only the first receiver in a fragment can have streamed in an earlier text node, and the qualified matcher additionally applies the `is_receiver_continuation` guard on the preceding byte, which the anchored trimmed matcher cannot. Keeping `trimmed` set beyond the first iteration would have been the unsafe direction.

crates/trusted-server-core/src/integrations/nextjs/rsc_stream.rs
```diff
@@ -0,0 +1,1480 @@
...
+        // in a full scan. Do not turn a provisional segment verdict into a
+        // document-wide bypass while more payloads can still arrive.
+        if self.invalid_segment && finalize {
+            return RscGroupStatus::Invalid;
+        }
+        match self
+            .inspector
+            .inspect(&self.combined[self.segment_start..], true)
+        {
+            HeaderSuffixStatus::Complete => {
```
aram356: 📝 **note** — Recording verification for the deferred-`Invalid` behaviour, which resolves the previous pass's observation that the incremental classifier could latch `Invalid` where a full rescan reported `NeedMore`.

The specific regression is fixed: `["a", ":T3:", "Te,"]` now returns `NeedMore` on every push and at `finalize()`, where it previously latched `Invalid` and enabled document-wide bypass.

More usefully, the divergence class is gone rather than just that case. Comparing the classifier against an independent full-rescan oracle, one that discovers chunks first and inspects every segment from scratch rather than re-invoking the incremental path, over all strings of length 1..=7 across two alphabets with every two- and three-way split: roughly 18M comparisons, zero mismatches, plus 400k randomized trials. This is the check the in-tree test could not make before, since the previous assertion compared the classifier against a wrapper around itself.

Deferring the verdict cannot hold unboundedly. Three independent bounds cut it off: `append_held` against the output limit, `RscGroupClassifier::push` against `max_combined_payload_bytes` (which returns `Invalid` and latches), and `resolve_group`'s `MAX_UNRESOLVED_RSC_PAYLOADS`. The tradeoff, that malformed groups now fall back later while valid complete groups still stream immediately, is recorded in the plan document.

crates/trusted-server-core/src/integrations/nextjs/rsc.rs
```diff
@@ -455,6 +581,24 @@ pub(crate) fn rewrite_rsc_scripts_combined_with_limit(
...
+    for payload in &payloads[..payloads.len() - 1] {
+        let mut iter = EscapeSequenceIter::new(payload);
+        loop {
+            if iter.has_partial_escape() {
+                return payloads
+                    .iter()
+                    .map(|payload| (*payload).to_owned())
+                    .collect();
+            }
+            if iter.next().is_none() {
```
aram356: 📌 **out of scope** — A real URL-corruption defect surfaced while sweeping this PR's rewrite path. It is **not** introduced or worsened here, and it should not gate this PR, but it deserves its own issue.

When a T-chunk's declared length lands inside a URL scheme, the colon is eaten. Reproduced with identity hosts, where the output should be byte-identical to the input:

```
in : 1:T7,{"u":"https://origin.example.com/a"}
out: 1:T7,{"u":"https//origin.example.com/a"}
```

`T7` makes `content_end` land inside `https:`, so the text splits into chunk content `{"u":"https` and remainder `://origin.example.com/a"}`. The remainder alone no longer looks scheme-qualified, so phase-2 bare-host rewriting in `rewrite_bare_host_at_boundaries` matches the host after `://` and emits `//` plus host, dropping the colon. A broken URL ships to the browser. Same shape for a truncated-scheme fragment: `ttps://origin.example.com/z` becomes `ttps//origin.example.com/z`.

Why it is out of scope for this PR, verified rather than assumed:

- `crates/trusted-server-core/src/host_rewrite.rs` is byte-identical to both the merge base and `origin/main` (`git diff` empty against each).
- It reproduces with a **single** payload, so no `RSC_MARKER`, no classifier, and no `has_partial_escape` involvement.
- I exported `origin/main` and ran the same input through its `rewrite_rsc_scripts_combined_with_limit`: identical corrupted output. Pre-existing, unchanged by this branch.

Worth noting why no test catches it: the recomputed T header matches the corrupted content, so the group is internally consistent and the classifier has nothing to object to. A fix probably needs the rewriter to avoid applying bare-host rewriting to a fragment whose leading bytes are the tail of a split scheme, or to rewrite before splitting at `content_end`.

aram356 (approved):
## Summary

Third pass, against head `481b24f7d` → `f93f80d99`. Both blocking findings from the previous pass are fixed, and I re-ran the original reproductions rather than reading the diff: escaped-JSON Flight payloads now release on `process_chunk` (25 bytes emitted, 0 held, previously 3 and 49), while genuinely truncated `\x`, `\u1`, and a lone high surrogate still hold back, so the length invariant the old guard was protecting survives the narrowing. The trimmed-claim bypass is fixed and extended beyond what was reported.

Nothing in this pass blocks. The comments below are two questions worth answering inline and three notes recording verification, one of which is a real defect that belongs in its own issue rather than here.

## Verification performed on this head

The two things most likely to have broken were the narrowed escape hold-back and the two main merges. Both came back clean at scale.

Marker and T-length invariant, the hole that could have reopened when `has_partial_escape` replaced the fixed twelve-byte window: roughly 1.05M `CompleteRewritable` marker round-trips against the real `RscUrlRewriter` across four request hosts (including hosts whose replacement begins with hex digits, the only way a rewrite could feed hex into a truncated escape), plus 17,725 exhaustive two-way and 15,562 three-way splits over escape-dense bodies. Zero T-length failures, zero arity failures. Classifier against an independent full-rescan oracle, not the in-tree helper: roughly 18M comparisons, zero mismatches. Escape self-consistency over ~600k bodies: no prefix holds forever, and the classifier and the iterator never disagree about whether a sequence is an escape. The new placeholder loop: ~20k fuzz cases plus 88 adversarial scripts, zero panics.

Merge correctness, given this repository's history of a merge silently reverting streaming work: both merges show an empty diff against the branch-side parent over `publisher.rs`, `html_processor.rs`, and `registry.rs` while main had added 1064 lines to `publisher.rs`, which is the shape a silent revert would take. It is not one. #1150 arrived via the earlier merge `3363ee49d`, so it was already branch-side, and independently its entire `publisher.rs` delta is inside `#[cfg(test)]` with zero production functions added. The two sides changed template-cache gating in disjoint code, so there was no conflict to mis-resolve. All 17 of main's added symbols are present at head, and `request_can_use_shared_template` now requires `AssemblyMode::Esi` and main's new `!cookie_disqualifies` as a conjunction, so the inline seam token still cannot enter a cached template.

Also confirmed: the two CI gates added to `AGENTS.md` since the last pass, `cargo clippy-cli` and `cargo clippy-codegen`, both pass on a clean export of this head, as does `cargo fmt --all -- --check`. `test_support.rs` remains gated `#[cfg(any(test, feature = "test-utils"))]` with the adapters depending on it only under `[dev-dependencies]`, so nothing test-only reaches a production wasm binary.

## CI Status

All 20 checks reported by `gh pr checks` are PASS, including the four required gates.

- integration tests: PASS
- CodeQL: PASS
- cargo test: PASS (required)
- vitest: PASS
- cargo fmt: PASS (required)
- Analyze (rust): PASS
- Analyze (actions): PASS

crates/trusted-server-core/src/integrations/nextjs/rsc.rs
prk-Jr: Confirmed; the partial-escape guard is unchanged by this follow-up. The full Fastly test suite passed locally (2,971 tests including doctests), including the Next.js escape and streaming coverage. The exhaustive sweep counts in this thread remain your verification results; I did not repeat those external sweeps.

crates/trusted-server-core/src/integrations/nextjs/rsc.rs
prk-Jr: Addressed in 7a690ed6f. The initial empty-input guard now uses split_last() and retains preceding_payloads. Both marker-count sizing and the escape scan use that slice directly, removing both len() - 1 calculations. Existing empty, single, and multi-payload coverage passes. All repository CI gates also passed locally; Vitest was run directly with the pinned Node 24 binary, and Fastly/Axum tests needed access outside the sandbox.

crates/trusted-server-core/src/integrations/nextjs/rsc_placeholders.rs
prk-Jr: Confirmed; no additional change needed here. The every-push loop still switches from the trimmed matcher to the qualified matcher after the first push. The full Fastly suite, including the Next.js placeholder coverage, passed again locally.

crates/trusted-server-core/src/integrations/nextjs/rsc_stream.rs
prk-Jr: Confirmed; the deferred-Invalid behavior and its existing regression coverage are unchanged. The full Fastly suite passed again locally, along with the native adapter tests and all 14 cross-adapter parity tests. I did not repeat the external exhaustive oracle sweep reported above.

crates/trusted-server-core/src/integrations/nextjs/rsc.rs
prk-Jr: Keeping this pre-existing defect out of this PR, as requested. I prepared a separate issue draft with the reproduction and acceptance criteria; it has not been filed yet. One correction from inspecting the implementation: the colon is dropped in phase 1, in shared.rs RscUrlRewriter::apply_rewrite. The regex captures an optional colon independently of the scheme, but the no-scheme replacement emits only slashes and the host. rewrite_bare_host_at_boundaries itself preserves the prefix before the host. The follow-up should cover split/truncated schemes, identity and different-length host replacements, escaped URLs, and T-length invariants.

prk-Jr: Following up: we are including this fix in this PR after all. Fixed in dcb7e5387.

The no-scheme branch in RscUrlRewriter::apply_rewrite now preserves the captured colon. Both new regressions failed before the fix with the expected colon loss, including the exact T7 identity-host reproduction, and pass afterward. Coverage includes six T-chunk boundaries in the scheme, single and combined payloads, escaped slashes, partial scheme prefixes, explicit ports, and identity/shorter/longer replacement hosts. The T headers and script boundaries remain unchanged in these boundary cases.

All repository-required local gates passed: target-matched clippy, Fastly tests (2,973 including doctests), native adapter tests, cross-adapter parity (14), JS build/tests (1,130), and Rust/JS/docs formatting. No separate issue is needed for this defect.

PR #1150, opened 2026-09-08
Add per-cookie template cache policy

### LINKED ISSUE #1138: Grow origin_is_cookie_independent into a per-cookie template cache policy
## Problem

`origin_is_cookie_independent` is all-or-nothing: either every cookie-bearing request bypasses the shared template cache (the default, which makes the cache nearly inert on real traffic because TS mints its own identity cookie), or all cookies are asserted irrelevant to origin HTML. Two common publisher deployment shapes cannot be expressed, and one of them makes a correct-looking config silently wrong.

### Case 1: publisher A/B testing framework (motivating case)

A common topology: the publisher's CDN tier sits downstream of TS (it is TS's origin) and translates an experiment-bucket cookie (say `ab_bucket`) into a variant request header (say `X-Exp-Variant`) before the origin. The origin varies its HTML on that header and declares it in `Vary`.

The operator does the apparently correct thing:

```toml
origin_is_cookie_independent = true
template_cache_vary = ["x-exp-variant", ...]
```

But `template_cache_vary` keys on request headers as TS receives them from the browser, and the browser sends only the cookie. The variant header does not exist yet at TS's hop, so it keys as absent for every reader and all experiment arms collapse into one shared template. The origin's `Vary: X-Exp-Variant` counts as covered by the configured key, so the drift guard does not refuse storage. Result: cross-served experiment arms, with a config that reads as correct.

### Case 2: small logged-in population

A mostly anonymous site with a small share of logged-in users whose account state is server-rendered. The operator wants those requests on the inline path and everyone else sharing templates. Today the only options are disabling sharing for everyone or asserting independence that is false for the logged-in slice.

## Proposal

Grow the boolean into a per-cookie policy with three behaviors:

```toml
# Named cookies whose values join the template cache key (per-variant signals).
template_cache_key_cookies = ["ab_bucket"]

# Named cookies whose presence forces the inline path: no lookup, no store.
template_cache_bypass_cookies = ["session"]

# Unchanged meaning, now scoped: every cookie not named above is asserted
# irrelevant to origin HTML. Default false, unchanged.
origin_is_cookie_independent = true
```

- Key cookies must be variant-shaped (bounded value sets: experiment arms, region buckets). Document that user-shaped identifiers do not belong in the key; the template stays reader-neutral and the key holds per-variant signals only, which is the existing key design rule.
- Bypass cookies mirror the existing disqualification: presence of any named cookie sets cookie-disqualified for both lookup and store.
- Unconfigured deployments are unchanged: conservative default, cache stays failing closed.
- The `Vary: Cookie` drift guard is unchanged and still refuses regardless of policy.

## Alternatives considered

- Deriving a synthetic request header from the cookie before key derivation: equivalent power, but it invents a header namespace; putting the cookie name in the key config is more direct.
- Keying on the full Cookie header: forbidden by design; every user is unique, so it is the per-user final-response cache under another name.

## Relation to existing work

Follows the shared-template eligibility gate introduced with the ESI/C2 work and the authorization carve-out from #1070; this is the cookie-side equivalent of that carve-out.

prk-Jr:
## Summary

- Separate shared HTML templates by bounded cookie variants, so experiment arms remain isolated when a downstream CDN derives a variant header after TS has selected its cache key.
- Route requests carrying configured session cookies inline without shared lookup or storage. Scope the existing cookie-independence assertion to unlisted cookies while preserving conservative defaults and unconditional `Vary: Cookie` refusal.
- Validate configuration and ambiguous prepared cookie input while tolerating compact JSON and comma-list values in ignored cookies. Preserve existing request preparation and cover variant isolation, session bypass, reader assembly, and rollback.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/creative_opportunities.rs` | Add optional key/bypass cookie lists, case-sensitive name validation, scoped independence documentation, and configuration tests. |
| `crates/trusted-server-core/src/cookies.rs` | Register the internal template-cookie policy module. |
| `crates/trusted-server-core/src/cookies/template_cache_policy.rs` | Evaluate all prepared cookie fields; distinguish absence from empty values; retain strict key-cookie values and duplicate rejection while allowing bounded JSON/comma-list syntax for ignored cookies. |
| `crates/trusted-server-core/src/platform/template_cache.rs` | Add framed, hashed cookie dimensions while preserving legacy keys when dimensions are empty; redact values in Debug output. |
| `crates/trusted-server-core/src/platform/mod.rs` | Export the cookie dimension type. |
| `crates/trusted-server-core/src/publisher.rs` | Reuse one cookie decision for lookup and storage; add end-to-end regressions, distinct bids on warm hits, and outbound Cookie assertions. |
| `crates/trusted-server-adapter-fastly/src/template_cache.rs` | Update the cache-key fixture with empty cookie dimensions. |
| `scripts/template-cache-local-test.sh` | Add repeatable cookie-selected runtime coverage in ESI and inline modes, including origin counts and content assertions. |
| `trusted-server.example.toml` | Document optional policies, bounded variants, downstream header coverage, and safe rollback. |
| `docs/guide/configuration.md` | Explain combined and independent deployment patterns, eligibility, limitations, and rollback. |
| `docs/superpowers/specs/2026-09-08-1138-per-cookie-template-cache-policy-design.md` | Record the reviewed behavior and request-preparation boundary. |
| `docs/superpowers/plans/2026-09-08-1138-per-cookie-template-cache-policy.md` | Record implementation steps and validation results. |

Ignored-value tolerance applies only when independence is asserted. Names, key values, and cookie framing remain validated; unmatched quotes, unsafe bytes, and comma-delimited cookie assignments still bypass. The origin must treat ignored values as opaque and parse semicolon-separated cookies independently; parsers that stop at nonstandard values cannot safely make this assertion.

## Closes

Closes #1138

## Test plan

Fresh review-readiness validation on 2026-09-10 used `1f56ee08cb25a35d79cccc19a442a698334a398c` plus the harness/documentation follow-up in `191380f0`. Production Rust and JavaScript are unchanged by that follow-up.

- [x] Fastly release artifact: `cargo build --package trusted-server-adapter-fastly --release --target wasm32-wasip1`.
- [x] Repeatable runtime test: `BID_DELAY=3 ./scripts/template-cache-local-test.sh esi` — 22 harness checks passed, including all 17 cookie-matrix requests.
- [x] Inline control: `BID_DELAY=3 ./scripts/template-cache-local-test.sh inline` — 9 harness checks passed, including all 17 cookie-matrix requests with an origin fetch on every request.
- [x] Negative control: a temporary harness copy with cookie keying removed failed on the first B request because it received the wrong variant HTML. No production code was changed for this control.
- [x] `cargo test-fastly` — 2,902 tests/doc-tests passed; 10 existing tests ignored.
- [x] `cargo test-axum`, `cargo test-cloudflare`, `cargo test-spin` — 41, 44, and 86 tests passed respectively.
- [x] `cargo test --manifest-path crates/trusted-server-integration-tests/Cargo.toml --test parity` — 13 passed.
- [x] All six target-specific Clippy aliases and `cargo fmt --all -- --check`.
- [x] JS build; Vitest under the pinned Node 24.12.0 from `crates/trusted-server-js/lib` — 45 files, 901 tests passed; JS and docs formatting.
- [x] Independent review of the harness extension; `bash -n scripts/template-cache-local-test.sh` and `git diff --check`.

The harness now returns distinguishable cookie-selected HTML and `Vary: X-Exp-Variant` without a client variant header. Both modes verify A/B isolation, absent/empty buckets, ignored compact JSON/comma-list cookies, warm session bypass including empty sessions, and cold session bypass without populating anonymous templates. Each request checks content, cache diagnostics, origin fetch counts, private response policy, and winning-bid assembly. Existing CI runs both harness modes.

These runtime runs invoke **Viceroy directly**, not `fastly compute serve`. The earlier headless-browser/Viceroy smoke test reported separate A/B `miss-stored` → `hit` sequences and session `bypass-request`, all HTTP 200, with page JavaScript disabled to keep cookies stable. Its exact tested commit was not recorded in the previous description; it is historical evidence, not a fresh browser run at the current head.

Viceroy/Fastly tests and Axum socket tests passed after granting the local certificate/keychain and loopback access needed outside the sandbox. An initial Vitest invocation used an incompatible Node environment; rerunning with the pinned Node executable from the library directory passed.

## Checklist

- [x] Changes follow [CLAUDE.md](/CLAUDE.md) conventions.
- [x] No new `unwrap()` in production code.
- [x] Uses `log` macros, not `println!`, per project conventions.
- [x] New code has tests.
- [x] No secrets or credentials committed.


crates/trusted-server-core/src/publisher.rs
```diff
@@ -11478,6 +11497,1001 @@ mod tests {
...
+                        cookie_policy_request(&[cookies.as_bytes()]),
+                        EdgeCacheHeader::SMaxageFallback,
+                    )
+                    .await
+                    .expect("should serve a reader with an existing EC");
+                    assert!(
+                        ec_context.kv_snapshot().entry_for(identity).is_some(),
+                        "should preload this reader's identity even during withdrawal on a hit"
+                    );
+                    let mut response = match finalizer {
```
ChristianPavilonis: **S1. Repeated finalizer switch**

The new withdrawal test repeats the existing `Finalizer::Streaming` and `Finalizer::Buffered` dispatch at line 9571. Both paths must keep testing streaming and buffered finalization, but duplicating this switch makes later finalizer changes easy to apply in only one location.

Consider extracting the finalizer dispatch into one test helper and calling it from both locations. This adds one helper and some parameter passing.

ChristianPavilonis (approved):
## Summary

Approved. The implementation is sound; one non-blocking code-smell note is attached inline.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -4280,20 +4281,27 @@ pub async fn handle_publisher_request(
...
+        });
+    let (cookie_disqualifies, cookie_values) = match evaluate_cookie_policy(
+        req.headers(),
+        key_cookie_names,
+        bypass_cookie_names,
+        cookie_independent,
+    ) {
+        TemplateCookieDecision::Bypass => (true, Vec::new()),
+        TemplateCookieDecision::Eligible(values) => (false, values),
+    };
```
aram356, line 4304: ♻️ **refactor** — A cookie bypass leaves no log at all. The key is `None`, so no reservation happens and the store-side `template_cache_ttl` never runs. Operators canarying this policy see `bypass-request` with no way to tell which of the six request gates fired, and this gate now has many more reasons to fire (bypass name, unlisted cookie, duplicate, malformed field). The method and request-semantics gates a few lines down already log this way; the message is static, so no cookie bytes reach the log.

```suggestion
    };
    if cookie_disqualifies && matches!(assembly_mode, AssemblyMode::Esi) {
        log::debug!(
            "template_cache bypass: {}",
            TemplateCacheBypassReason::CookieForwarded
        );
    }
```

_(scratch-verified: fmt, all six clippy aliases, all four adapter test aliases, and the parity suite)_

crates/trusted-server-core/src/creative_opportunities.rs
```diff
@@ -331,21 +331,29 @@ pub struct CreativeOpportunitiesConfig {
...
+    pub template_cache_key_cookies: Option<Vec<String>>,
+    /// Cookie names whose presence forces inline processing, with no lookup or store.
     ///
-    /// Setting `true` asserts the origin serves the same HTML with or without cookies.
-    /// It is not taken on trust alone — if the origin ever declares `Vary: Cookie`, the
-    /// response is refused regardless of this flag or the configured key. So a wrong
-    /// assertion is caught whenever the origin is honest about it, and this only widens
-    /// the window where the origin personalizes *silently*.
+    /// Empty values still count as present. Names must be unique and must not overlap
+    /// [`Self::template_cache_key_cookies`]. Unset or empty names no bypass cookies.
```
aram356: ⛏ **nitpick** — The sentence is broken: "Unset or empty names no bypass cookies."

```suggestion
    /// [`Self::template_cache_key_cookies`]. Unset or empty means no bypass cookies.
```

crates/trusted-server-core/src/cookies/template_cache_policy.rs
```diff
@@ -0,0 +1,457 @@
...
+                    dimension("z", Some(b"a=b%2F"))
+                ]),
+                "should preserve representation and ignore pair order and splitting"
+            );
+        }
+        for value in [b"".as_slice(), b"\"\"", b"!#$%&'()*+-./0123456789:<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[]^_`abcdefghijklmnopqrstuvwxyz{|}~"] {
+            let mut field = b"ab_bucket=".to_vec();
+            field.extend_from_slice(value);
+            assert_eq!(evaluate_cookie_policy(&headers(&[&field]), &names(&["ab_bucket"]), &[], false), TemplateCookieDecision::Eligible(vec![dimension("ab_bucket", Some(value))]), "should accept every cookie-octet and preserve quotes");
+        }
```
aram356: ⛏ **nitpick** — Line 338 is a 230-character `assert_eq!` that rustfmt gave up on because the byte-string literal cannot be broken. Binding the literal first lets rustfmt format the rest.

```suggestion
        let every_octet: &[u8] =
            b"!#$%&'()*+-./0123456789:<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[]^_`abcdefghijklmnopqrstuvwxyz{|}~";
        for value in [b"".as_slice(), b"\"\"", every_octet] {
            let mut field = b"ab_bucket=".to_vec();
            field.extend_from_slice(value);
            assert_eq!(
                evaluate_cookie_policy(&headers(&[&field]), &names(&["ab_bucket"]), &[], false),
                TemplateCookieDecision::Eligible(vec![dimension("ab_bucket", Some(value))]),
                "should accept every cookie-octet and preserve quotes"
            );
        }
```

_(scratch-verified: `cargo fmt --check`, `cargo clippy-fastly`, and the `template_cookie_policy` tests)_

docs/guide/configuration.md
```diff
@@ -1905,8 +1905,12 @@ template_cache_vary = [
...
 template_cache_max_age_seconds = 1200
 
-# Default false. Enable only after proving publisher HTML ignores Cookie.
-origin_is_cookie_independent = true
+# Optional bounded variant and personal-session policies; omitted means empty.
+template_cache_key_cookies = ["ab_bucket"]
+template_cache_bypass_cookies = ["session"]
+
+# Default false. Enable only after proving unlisted cookies do not change HTML.
+origin_is_cookie_independent = false
```
aram356: ⛏ **nitpick** — This primary example pairs both lists with `false`. With either list present, `false` bypasses every request carrying any unlisted cookie, which includes the TS identity cookie, so the example as shown shares almost nothing on real traffic. The harness and the downstream-CDN example further down both use `true`. Keeping `false` here is fine as the safe placeholder, but the comment should say what it costs.

```suggestion
# Default false. With the lists above, false still bypasses every request that
# carries any other cookie, including the TS identity cookie. Set true only
# after proving those unlisted cookies do not change origin HTML.
origin_is_cookie_independent = false
```

_(scratch-verified: prettier check)_

crates/trusted-server-core/src/cookies/template_cache_policy.rs
```diff
@@ -0,0 +1,457 @@
...
+                return TemplateCookieDecision::Bypass;
+            };
+            let (name, rest) = pair.split_at(separator);
+            let value = &rest[1..];
+            if !is_cookie_name(name) {
+                return TemplateCookieDecision::Bypass;
+            }
+            if parsed.insert(name, value).is_some() {
+                return TemplateCookieDecision::Bypass;
+            }
```
aram356: 🤔 **thinking** — A duplicate *unlisted* name bypasses even when independence is asserted. Browsers legitimately send same-name cookies when they exist for different `Path` or `Domain` scopes (RFC 6265 §5.4), so a site with a cookie set on both the apex and `www` bypasses every reader under a named policy. For an unlisted cookie the operator has already asserted the value is irrelevant to origin HTML, so which duplicate wins is irrelevant too. Only key-cookie duplicates are genuinely ambiguous, and bypass names bypass regardless of duplication.

Restricting the duplicate check to key names keeps the safety argument and recovers hit rate:

```rust
let keyed = key_names.iter().any(|item| item.as_bytes() == name);
if parsed.insert(name, value).is_some() && keyed {
    return TemplateCookieDecision::Bypass;
}
```

Apply manually — the spec (§6, "Any repeated cookie name anywhere in the request causes bypass") and the `a=1; a=1` rows in `template_cookie_policy_rejects_malformed_and_duplicate_fields` would need to change with it.

crates/trusted-server-core/src/cookies/template_cache_policy.rs
```diff
@@ -0,0 +1,457 @@
...
+) -> Result<(), String> {
+    let mut seen = HashSet::new();
+    for (field, names) in [
+        ("template_cache_key_cookies", key_names),
+        ("template_cache_bypass_cookies", bypass_names),
+    ] {
+        for name in names {
+            if !is_cookie_name(name.as_bytes()) {
+                return Err(format!("{field} contains invalid cookie name `{name}`"));
+            }
```
aram356, line 32: ♻️ **refactor** — Keying on `ts-ec` is a one-word config mistake that silently creates the per-user cache the design forbids, and the docs can only warn against it. TS knows its own identity cookie names (`COOKIE_TS_EC`, `COOKIE_TS_EIDS`, `COOKIE_SHAREDID` in `constants.rs`), so validation can reject them in the key list at load time. Bypassing on them is an odd but legitimate choice, so only the key list needs the guard.

```rust
if field == "template_cache_key_cookies"
    && [COOKIE_TS_EC, COOKIE_TS_EIDS, COOKIE_SHAREDID].contains(&name.as_str())
{
    return Err(format!(
        "{field} must not key on Trusted Server identity cookie `{name}`"
    ));
}
```

Apply manually — needs the `crate::constants` import and a new case in `template_cookie_policy_validates_names`, which live outside this range.

crates/trusted-server-core/src/publisher.rs
```diff
@@ -5710,14 +5719,11 @@ pub(crate) enum TemplateCacheBypassReason {
...
-    /// `Set-Cookie` itself when the session was established earlier. Named in §4 of
-    /// the design doc; disqualifying until the origin's `Vary` is verified to cover
-    /// it.
-    #[display("request carried Cookie and the origin's Vary does not cover it")]
+    /// The prepared request's cookies are disqualified by the configured policy.
+    ///
+    /// Named bypass cookies, unlisted cookies without an independence assertion,
+    /// or ambiguous input under a named policy prohibit both lookup and storage.
+    #[display("request cookies disqualified by template cache policy")]
     CookieForwarded,
```
aram356: ⛏ **nitpick** — `CookieForwarded` no longer describes the reason. Its display text now says "disqualified by template cache policy", and the variant fires for bypass names, unlisted cookies, duplicates, and malformed fields, not for forwarding. `CookiePolicy` would match the new meaning.

Apply manually — the rename also touches the `template_cache_ttl` return at line 6151 and the existing `a_forwarded_request_cookie_disqualifies_even_without_set_cookie` test, both outside this hunk.

aram356 (changes requested):
## Summary

Grows the cookie-independence boolean into a per-cookie policy: named key cookies become sorted, length-prefixed dimensions of the shared template key, named bypass cookies force the inline path, and the assertion narrows to unlisted cookies. The runtime logic holds up under tracing: one decision drives lookup, reservation, and store, so a bypass cookie can never read or populate a template; `Vary: Cookie` refusal is untouched; empty dimensions preserve legacy keys; the fingerprint picks up the new fields through settings serialization; Debug output redacts values. Tests cover the evaluator matrix, key isolation, both finalizers, EC withdrawal on warm hits, and the harness negative control, and CI runs both harness modes.

Nothing here is blocking on its own. The findings below are hit-rate, operator-safety, and readability improvements.

> 4 of the inline comments below carry a one-click GitHub `suggestion`. Use **Commit suggestion** (or **Add suggestion to batch**) to apply them. Each was applied in isolation and as a batch in a scratch worktree and passed `cargo fmt --check`, all six clippy aliases, all four adapter test aliases, the parity suite, and prettier. The remaining comments describe the fix in prose because the change spans ranges outside one hunk.

## Non-blocking

### ♻️ refactor

- **Log the cookie bypass reason at the request gate** — see inline at `crates/trusted-server-core/src/publisher.rs:4304`
- **Reject TS identity cookie names in the key list** — see inline at `crates/trusted-server-core/src/cookies/template_cache_policy.rs:28`

### 🤔 thinking

- **Duplicate unlisted names need not bypass under an independence assertion** — see inline at `crates/trusted-server-core/src/cookies/template_cache_policy.rs:73`

### ⛏ nitpick

- **Doc sentence for `template_cache_bypass_cookies`** — see inline at `crates/trusted-server-core/src/creative_opportunities.rs:344`
- **Unformatted 230-character assert** — see inline at `crates/trusted-server-core/src/cookies/template_cache_policy.rs:335`
- **Primary docs example pairs the lists with `false`** — see inline at `docs/guide/configuration.md:1912`
- **`CookieForwarded` variant name is stale** — see inline at `crates/trusted-server-core/src/publisher.rs:5727`

- 🌱 **No bound on key-cookie value length.** `is_cookie_value` accepts any cookie-octet run of any length, so a mistakenly keyed token or ID enters the key unchecked. The spec scopes cardinality controls out, but a cheap bypass above a fixed byte length (64 is generous for experiment arms and region buckets) would catch the accidental case without limiting real variant labels. Future work, not this PR.
- 🤔 **Named-policy strictness for unlisted values.** Under a named policy with independence `true`, every unlisted cookie value is still framing-checked in `is_ignored_cookie_value`: a raw space, a non-ASCII byte, or a backslash in any third-party cookie bypasses the whole request. That is stricter than the legacy boolean path and partially reintroduces the "nearly inert" problem this feature targets. The origin-parser rationale is documented in the spec and the guide, so this is a heads-up on expected hit rate during the canary rather than a request to change it.

## CI Status

- integration tests: PASS
- CodeQL: PASS
- Analyze (actions): PASS
- cargo test: PASS (required)
- Analyze (rust): PASS
- cargo fmt: PASS (required)
- vitest: PASS

crates/trusted-server-core/src/publisher.rs
prk-Jr: Addressed in cd9ccd4ef. Extracted `finalize_test_publisher_response` and routed both the shared publisher fixture and the EC withdrawal test through it. Both streaming and buffered finalizers remain covered, including warm-cache withdrawal. The full Fastly test suite passes.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Addressed in cd9ccd4ef. Added the static debug log at the request cookie gate when ESI mode is active, using the renamed `CookiePolicy` reason. It includes no cookie names or values and runs even when bypass prevents reservation and store-side checks.

crates/trusted-server-core/src/creative_opportunities.rs
prk-Jr: Addressed in cd9ccd4ef. Corrected the sentence to “Unset or empty means no bypass cookies.”

crates/trusted-server-core/src/cookies/template_cache_policy.rs
prk-Jr: Addressed in cd9ccd4ef. Bound the byte-string fixture to `every_octet`, allowing rustfmt to format the loop and assertion across lines. Cookie-policy regressions and Rust formatting pass.

docs/guide/configuration.md
prk-Jr: Addressed in cd9ccd4ef. Kept the safe `false` default and expanded the example comment to explain that any unlisted cookie, including the TS identity cookie, still forces bypass. It now states when enabling the independence assertion is appropriate. Documentation formatting passes.

crates/trusted-server-core/src/cookies/template_cache_policy.rs
prk-Jr: Addressed in cd9ccd4ef. Restricted duplicate detection to key-cookie names. Unlisted duplicates are eligible only with independence asserted, and every occurrence still passes the existing framing checks; bypass names always bypass. Updated the specification, guide, evaluator matrices, and publisher coverage for duplicate unlisted cookies through both finalizers. Duplicate key names still bypass for equal or different values, within one field or across fields.

crates/trusted-server-core/src/cookies/template_cache_policy.rs
prk-Jr: Addressed in cd9ccd4ef. Configuration validation now rejects `COOKIE_TS_EC`, `COOKIE_TS_EIDS`, and `COOKIE_SHAREDID` in the key list while allowing them in the bypass list. Added regression coverage for all three names and documented the restriction. The error identifies the configuration field and forbidden cookie.

crates/trusted-server-core/src/publisher.rs
prk-Jr: Addressed in cd9ccd4ef. Renamed the variant to `CookiePolicy`, including the request-gate log, store-side return, and existing bypass test. Local validation passes: all four adapter test aliases, all six Clippy aliases, 13 parity tests, 901 JavaScript tests, the JS build, and Rust/JS/docs formatting.

crates/trusted-server-core/src/cookies/template_cache_policy.rs
```diff
@@ -0,0 +1,525 @@
...
+pub(crate) enum TemplateCookieDecision {
+    Bypass,
+    Eligible(Vec<TemplateCookieValue>),
+}
+
+/// Validates exact cookie names across both configured policies.
+///
+/// # Errors
+///
+/// Returns an error naming the field and cookie when a name is invalid or repeated.
```
aram356: ⛏ **nitpick** — The `# Errors` section predates the identity-cookie rejection added in cd9ccd4ef.

```suggestion
/// Returns an error naming the field and cookie when a name is invalid or repeated,
/// or when the key list names a Trusted Server identity cookie.
```

crates/trusted-server-core/src/creative_opportunities.rs
```diff
@@ -455,11 +478,16 @@ impl CreativeOpportunitiesConfig {
     /// Returns an error string when [`gam_network_id`](Self::gam_network_id) is
     /// blank but consumed by a default path or `{network_id}` template; when a
     /// slot has an invalid identifier, page pattern set, format list, or
-    /// dimensions; when `template_cache_max_age_seconds` falls outside 1–86,400;
+    /// dimensions; when cookie policy names are invalid, duplicated, or overlapping;
```
aram356: ⛏ **nitpick** — Same gap here: `validate_runtime` now also fails when the key list names a TS identity cookie.

```suggestion
    /// dimensions; when cookie policy names are invalid, duplicated, or overlapping,
    /// or when the key list names a Trusted Server identity cookie;
```

trusted-server.example.toml
```diff
@@ -355,10 +355,23 @@ auction_timeout_ms = 500
...
+# template_cache_key_cookies = ["ab_bucket"]
+# Presence, including session=, bypasses template lookup and storage.
+# template_cache_bypass_cookies = ["session"]
+# With either list nonempty, false bypasses unlisted cookies; true asserts those
+# cookies do not change origin HTML. No identity or consent cookie is exempt.
+# With both lists empty, false bypasses all Cookie headers, as before.
+# An origin `Vary: Cookie` is always refused, regardless of these settings.
 # origin_is_cookie_independent = false
+# Invalid names, duplicates, and cross-list overlap reject configuration.
+# Named policies bypass ambiguous/nonconforming Cookie fields, including duplicates.
```
aram356: ⛏ **nitpick** — Two lines went stale with cd9ccd4ef: only duplicate *key* cookies bypass now, and the identity-cookie rejection is missing from the configuration-rejection list.

```suggestion
# Invalid names, duplicates, cross-list overlap, and TS identity cookies in the
# key list reject configuration. Named policies bypass ambiguous/nonconforming
# Cookie fields, including duplicate key cookies.
```

_(scratch-verified: `cargo fmt --check`, `cargo clippy-fastly`, `cargo check-axum`, the `template_cookie` tests, and a TOML parse of the example file, alone and together with the other two suggestions)_

aram356 (approved):

PR #1157, opened 2026-09-09
Reduce EID KV write conflicts during page loads

### LINKED ISSUE #993: Reduce EID KV write conflicts during page loads
## Description

Trusted Server saves IDs from the `ts-eids` and `sharedId` cookies while finishing every returning-user response. A normal publisher page creates many first-party requests at once, including scripts, integration requests, analytics calls, and auctions. Each response may try to update the same EC record in KV.

A matched production browser run recorded 45 warnings for one EC identity:

```text
CAS conflict after 5 retries upserting 3 partner IDs
```

The page continued because this sync is best-effort, but some partner IDs may not have been saved.

Relevant code:

- `crates/trusted-server-core/src/ec/finalize.rs`
- `crates/trusted-server-core/src/ec/prebid_eids.rs`
- `crates/trusted-server-core/src/ec/kv.rs`

## Current behavior

For every returning-user response:

1. Read partner IDs from the request cookies.
2. Read the EC record and its generation number from KV.
3. Merge the partner IDs.
4. Write only if the generation number has not changed.
5. Retry immediately up to five times when another request writes first.

Requests may contain cookies captured at different points during the page load. They can also finish in a different order from the one in which they started.

CAS prevents two requests from writing the same version of the record. It does not tell Trusted Server which request has the newest cookie values. An older request can reread the record and try to replace a newer partner ID with an older value.

## Impact

- Some partner IDs may not be saved.
- Older cookie values may replace newer values.
- Page loads create unnecessary KV reads and writes.
- Immediate retries use edge compute while the record remains busy.
- Repeated warnings make Fastly logs harder to read.
- Later auctions and partner syncs may use an incomplete identity record.

The browser request still succeeds, so this problem is easy to miss without Fastly logs.

### Limit where partner IDs are saved

Keep consent checks, withdrawal handling, and EC cookie handling on every response where they are currently required. Limit only the partner ID sync.

Save EID cookie values during:

- New EC creation
- Top-level document navigation
- `POST /auction`

Do not save them while serving static assets, integration requests, analytics calls, or other page resources.

If navigation and auction requests miss IDs discovered late in the page load, add a dedicated browser sync endpoint instead of saving IDs on every response.

### Replace repeated writes with one conflict check

Replace the five immediate write retries with:

1. Read the record and attempt one conditional write.
2. If another request writes first, read the record one more time.
3. Return success if the desired IDs are already present.
4. Otherwise defer the update to a later navigation or auction request.

Do not make another immediate write from the same request after the conflict.

### Prevent older values from replacing newer ones

Before allowing one partner ID value to replace another, define a clear rule for deciding which value is newer. This may require storing the time when Trusted Server observed each partner ID.

Until that rule exists, a request that loses a CAS conflict should not overwrite a different value found during the follow-up read.

### Add measurements

Add counters for:

- Partner ID sync attempts
- Requests where the values already matched
- Successful KV writes
- Conflicts where another request had already saved the same values
- Updates deferred after a conflict
- Whether the sync came from navigation, auction, or new EC creation

Logs and measurements must not contain EC IDs or partner ID values.

## Completion criteria

- Static files and integration requests do not write partner IDs to KV.
- New EC creation, document navigation, and `/auction` still save partner IDs.
- A conflict returns success when another request already saved the same values.
- An older request cannot replace a newer partner ID after losing a conflict.
- Consent withdrawal continues to take priority on every applicable route.
- Concurrent request tests no longer exhaust five immediate retries.
- Measurements distinguish successful, duplicate, and deferred updates without exposing identity data.

## Out of scope

- Increasing the retry limit
- Sleeping inside an edge request
- Logging EC IDs or partner ID values
- Changing consent withdrawal behavior

ChristianPavilonis:
## Summary

- Limit returning-user EID cookie persistence to publisher document navigations and `POST /auction`; new EC creation remains eligible.
- Replace immediate five-write CAS retry loops with one conditional write and one follow-up read.
- Preserve consent withdrawal and defer differing browser-cookie values when trustworthy value freshness is unavailable.

This PR is stacked on #903.

## Changes

| Area | Change |
| --- | --- |
| Fastly routing | Classify navigation and auction requests as allowed EID sync sources; static, analytics, integration, and other subresources remain ineligible. |
| EC finalization | Keep consent and cookie handling before the EID sync boundary, while collecting cookie updates only for eligible sources. |
| KV persistence | Add missing IDs with one CAS write; after conflict, reread once and report matched, deferred, withdrawn, missing, or failed outcomes without retrying. Different existing values are conservatively preserved because the cookies have no value-owned version. |
| Observability | Emit bounded, identity-free fields for attempts, matches, writes, duplicate conflicts, deferrals, outcomes, and source. |
| Tests | Cover route classification, new/returning EC behavior, one-write conflict convergence, concurrent withdrawal, conservative freshness, and measurement dimensions. |

## Closes

Closes #993

## Test plan

- [x] `cargo test-fastly`
- [x] `cargo test-axum`
- [x] `cargo test-cloudflare`
- [x] `cargo test-spin`
- [x] Cross-adapter parity tests
- [x] Fastly, Axum, Cloudflare, Cloudflare Wasm, Spin native, and Spin Wasm Clippy targets
- [x] `cd crates/trusted-server-js/lib && npm run format && npm run lint && npm test`
- [x] `cd docs && npm run format && npm run lint && npm run build`
- [x] Fastly and Spin release Wasm builds

## Checklist

- [x] Changes follow `CLAUDE.md` conventions
- [x] Uses `log` macros rather than `println!`
- [x] New behavior has regression coverage
- [x] No secrets, EC IDs, or partner ID values are emitted by new measurements


crates/trusted-server-adapter-fastly/src/app.rs
```diff
@@ -817,8 +818,12 @@ async fn dispatch_fallback(
...
         // Only for document navigations by recognised browsers; subresource
         // requests may lack consent signals such as Sec-GPC.
-        let is_publisher_navigation = ec.is_real_browser && is_navigation_request(&req);
-        if is_publisher_navigation
+        let is_publisher_navigation = is_navigation_request(&req);
+        if is_publisher_navigation {
+            ec.ec_context.set_eid_sync_source(EidSyncSource::Navigation);
+        }
+        if ec.is_real_browser
+            && is_publisher_navigation
```
prk-Jr: ♻️ **refactor** — `is_publisher_navigation` lost its `ec.is_real_browser` term so it can serve sync-source classification, but it is still the value passed to `set_recovery_eligible` at line 865, whose comment there still reads *"Authorize orphan recovery only for real-browser document navigations"*. Bot navigations are now recovery-eligible.

This is inert today, but only through two invariants a long way from here: non-browsers get `kv_graph = None` (line 450), and the withdrawn-consent case that *does* keep a finalize KV graph returns from `ec_finalize_response` before the returning-user branch that reads `recovery_eligible()`. Nothing in the suite pins it either — no test asserts a non-browser navigation is recovery-ineligible.

Splitting the two concepts keeps the gate where the comment says it is, and the new classification test still passes because it deliberately does not depend on the browser gate.

```suggestion
        let is_publisher_navigation = is_navigation_request(&req);
        if is_publisher_navigation {
            ec.ec_context.set_eid_sync_source(EidSyncSource::Navigation);
        }
        let is_browser_navigation = ec.is_real_browser && is_publisher_navigation;
        if is_browser_navigation
```

Line 865 then becomes `ec.ec_context.set_recovery_eligible(is_browser_navigation);` — outside this diff, so it needs a manual edit alongside the suggestion.

Verified on the PR head in a scratch worktree: `cargo fmt --all -- --check` clean, `cargo clippy-fastly` clean, `app::tests` 41/41 pass.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -87,6 +87,96 @@ impl PartnerIdUpdate {
...
+    for (partner_id, uid) in latest_updates {
+        match entry.ids.get(partner_id) {
+            Some(existing) if existing.uid == uid => continue,
+            // Browser EID cookies carry no value-owned sequence or timestamp.
+            // A different value therefore has unknown freshness and must not
+            // replace the stored value.
+            Some(_) => {
+                deferred = true;
+                continue;
+            }
```
prk-Jr, line 152: 🤔 **thinking** — deferring a different value is the right call for concurrency, but the deferral is permanent rather than conflict-scoped, and that has a consequence worth stating explicitly.

Partner-owned paths still overwrite: pull sync goes through `upsert_partner_ids_from_snapshot` and partner push through `upsert_partner_id_if_exists`, both of which replace differing values. So any pull- or push-enabled partner self-heals after a rotation.

A **cookie-only** partner does not. `_sharedid`, or any Prebid userId module matched by the registry but not configured for pull/push, pins to the first UID observed for that row and keeps it for the row's whole TTL — and that stale UID keeps being attached to bid requests. Before this PR, the next request with a changed cookie corrected it.

Not asking for a fix here: inventing a staleness threshold in this PR would be a magic constant with no grounding. But a tracked follow-up for "cookie values need a freshness source" would be worth opening, and the trade-off is worth a line in the module docs so the next reader does not treat the stored UID as current.

crates/trusted-server-core/src/ec/finalize.rs
```diff
@@ -74,13 +77,13 @@ pub fn ec_finalize_response(
...
-                ec_context.kv_snapshot().clone(),
-            );
-            ec_context.set_kv_snapshot(snapshot);
+            let source = ec_context.eid_sync_source();
+            let updates = source
+                .map(|_| collect_eid_cookie_updates(eids_cookie, sharedid_cookie, registry))
+                .unwrap_or_default();
+            if let Some(source) = source {
+                sync_eid_cookie_updates(graph, ec_context, &ec_id, &updates, source);
+            }
```
prk-Jr, line 86: 🤔 **thinking** — a side effect worth confirming as intended: this branch is where the returning-user KV read used to happen unconditionally, so it is also what fed `reconcile_pull_sync_marker` a real snapshot.

With subresources now skipping the sync entirely, their `kv_snapshot()` stays `NotRead` at finalization where it previously became `Present`/`Missing`. `reconcile_marker` treats `NotRead` as "no information" and only expires an already-`Invalid` marker, so KV-driven marker reconciliation is now navigation/auction-only.

Consent withdrawal is unaffected — it expires the marker on the independent early path in this function, on any request. The residual gap is a row that becomes pull-incomplete or missing: a `Valid` marker now survives on subresource responses until the next navigation or `POST /auction`. Given `/auction` fires on most page loads the practical window is small, but this is a disclosure-window signal, so worth an explicit yes/no rather than an inference.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -479,68 +568,156 @@ impl KvIdentityGraph {
...
                 }
+                let outcome = if partner_id_updates_match(refreshed_entry, updates) {
+                    EidCookieSyncOutcome::ConflictMatched
+                } else {
+                    EidCookieSyncOutcome::DeferredConflict
+                };
+                (refreshed, outcome)
+            }
+            Err(_err) => {
+                log::warn!("EID cookie sync write failed");
```
prk-Jr: 🔧 **wrench** — the store error is bound to `_err` and dropped, so a failed EID write is now undiagnosable: the log line is a fixed string with no cause, no store name, and no ec-id prefix. The path this replaced logged `{err:?}` alongside `log_id(ec_id)`, so including the report is an established pattern here and does not widen identity exposure.

```suggestion
            Err(err) => {
                log::warn!("EID cookie sync write failed: {err:?}");
```

Verified on the PR head: `cargo fmt --all -- --check` and `cargo clippy-fastly` both clean with this applied.

crates/trusted-server-core/src/ec/finalize.rs
```diff
@@ -124,6 +122,71 @@ pub fn ec_finalize_response(
...
+        "EID sync measurement: source={} outcome={} attempted=1 already_matched={} written={} \
+         conflict_duplicate={} deferred={}",
+        measurement.source,
+        measurement.outcome,
+        measurement.already_matched,
+        measurement.written,
+        measurement.conflict_duplicate,
+        measurement.deferred,
+    );
+}
```
prk-Jr, line 189: 🤔 **thinking** — three things about this measurement, all pointing the same direction.

**It cannot show the reduction it exists for.** `sync_eid_cookie_updates` returns before `record_eid_sync_terminal` whenever `updates.is_empty()`, and the unclassified-source case never calls it at all. So the suppressed subresource syncs — the actual behavior this PR ships — emit nothing, and there is no denominator. A `source=none` / suppressed count, or a single line at the classification boundary, would make the write reduction observable in production instead of inferred.

**`EidSyncMeasurement` is a one-hot re-encoding of `outcome`.** All four `u8` fields are pure functions of the single `outcome` value already being logged, and `eid_sync_measurement_dimensions_are_bounded_and_identity_free` mostly re-asserts that mapping rather than any behavior. If a metrics pipeline needs the fan-out it belongs there; if not, logging `source` and `outcome` alone carries the same information with no struct and no test to keep in sync. The `Display` bounds assertion in that test is genuinely valuable and worth keeping either way.

**Level.** This is `info!` on every navigation and every `/auction` that carries EID cookies, i.e. roughly per page load per user. `debug!` for `AlreadyMatched`/`Written` with `info!`/`warn!` reserved for the deferred and failed outcomes would keep the anomalies visible without the steady-state volume.

crates/trusted-server-core/src/ec/finalize.rs
```diff
@@ -244,8 +307,8 @@ fn confirm_then_recover_orphaned_ec(
         EcKvSnapshot::Present { .. } => {
             // The row became visible after the origin round trip: adopt it and
             // merge any pending updates rather than rotating a valid identity.
-            let merged = graph.upsert_partner_ids_from_snapshot(ec_id, updates, confirmed);
-            ec_context.set_kv_snapshot(merged);
+            ec_context.set_kv_snapshot(confirmed);
+            sync_eid_cookie_updates(graph, ec_context, ec_id, updates, EidSyncSource::Navigation);
```
prk-Jr, line 312: ⛏ **nitpick** — `EidSyncSource::Navigation` is hardcoded here rather than read from the context. Correct today, since `recovery_eligible` is only ever granted on the publisher navigation path, but it encodes that coupling in a second place: if another source ever becomes recovery-eligible, this call site silently mislabels its measurements. `ec_context.eid_sync_source().unwrap_or(EidSyncSource::Navigation)` keeps the same behavior and keeps the fact in one place.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -479,68 +568,156 @@ impl KvIdentityGraph {
...
+    /// pre-write snapshot as proof and defers the update.
+    pub(crate) fn sync_eid_cookie_updates_from_snapshot(
+        &self,
+        ec_id: &str,
+        updates: &[PartnerIdUpdate],
+        snapshot: EcKvSnapshot,
+    ) -> (EcKvSnapshot, EidCookieSyncOutcome) {
+        if updates.is_empty() {
+            return (snapshot, EidCookieSyncOutcome::AlreadyMatched);
+        }
```
prk-Jr, line 570: ⛏ **nitpick** — unreachable, and misreporting if it were reached. Both callers pre-check for empty updates (`upsert_partner_ids` at line 568, `sync_eid_cookie_updates` in `finalize.rs`), and `AlreadyMatched` claims the stored values were verified to match when nothing was compared. Either drop the branch or `debug_assert!(!updates.is_empty())` so the invariant is stated rather than papered over.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -87,6 +87,96 @@ impl PartnerIdUpdate {
...
+}
+
+fn partner_id_updates_match(entry: &KvEntry, updates: &[PartnerIdUpdate]) -> bool {
+    updates.iter().all(|update| {
+        entry
+            .ids
+            .get(&update.partner_id)
+            .is_some_and(|existing| existing.uid == update.uid)
+    })
+}
```
prk-Jr, line 186: ⛏ **nitpick** — asymmetry with `apply_cookie_partner_id_updates` directly above: that function collapses duplicate partner ids through a `BTreeMap` (last value wins), this one iterates the raw slice and requires *every* entry to match. Given `("p","a"), ("p","b")` the write applies `b` and this check can never return true, turning a genuine `ConflictMatched` into `DeferredConflict`.

Inert in practice — `collect_eid_cookie_updates` runs `dedupe_partner_updates` — but `sync_eid_cookie_updates_from_snapshot` is `pub(crate)` and takes an arbitrary slice, so the dedupe is a caller-side invariant that is not stated anywhere. Collapsing here too (or documenting the precondition) makes the two halves agree.

crates/trusted-server-core/src/ec/mod.rs
```diff
@@ -86,6 +86,23 @@ use self::kv::{CreateIfAbsentOutcome, KvIdentityGraph};
 use self::kv_types::KvEntry;
 use self::pull_sync_marker::{PullSyncMarkerState, validate_marker_state};
 
+/// Bounded request classifications that may persist browser EID cookies.
+///
+/// Adapters assign a source only after pre-route filters allow dispatch.
```
prk-Jr: ⛏ **nitpick** — "Adapters assign a source" is true for `Navigation` and `Auction` but not `NewEc`, which is assigned inside `ec_finalize_response`, not by any adapter. Worth naming since `NewEc` is a public variant that adapters specifically should *not* set.

```suggestion
/// A source is assigned only after pre-route filters allow dispatch: adapters
/// classify navigation and auction requests, and `NewEc` is assigned during EC
/// finalization.
```

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -459,17 +549,16 @@ impl KvIdentityGraph {
...
+    /// most one conditional write and one follow-up read on conflict.
     ///
     /// # Errors
     ///
-    /// Returns [`TrustedServerError::KvStore`] on store error, missing root
-    /// entry, withdrawn root entry, or CAS exhaustion after
-    /// [`MAX_CAS_RETRIES`] attempts.
+    /// Returns [`TrustedServerError::KvStore`] on store failure, a missing root,
+    /// or a withdrawn root.
     pub(crate) fn upsert_partner_ids(
```
prk-Jr: 🌱 **seedling** — worth a follow-up rather than this PR: after the rewrite this "compatibility entry point" has no production callers left.

Its only caller is `ingest_eid_cookies_with_writer`, reached from `ingest_prebid_eids` and `ingest_sharedid_cookie`, and neither of those is called anywhere outside `prebid_eids.rs` itself. `upsert_partner_id` (singular) is likewise tests-only now. So `upsert_partner_ids`, the `PartnerIdBulkWriter` trait, and both `pub fn ingest_*` entry points exist to keep a `Result`-shaped API alive for tests, while carrying real cost: they are a second way to reach the cookie-sync semantics whose error mapping has to be maintained in parallel.

If it is genuinely reachable through something I missed, a doc line naming that caller would help. Otherwise deleting the chain and porting its tests onto `sync_eid_cookie_updates_from_snapshot` would remove the duplicate surface.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -1723,6 +1899,123 @@ mod tests {
         }
     }
 
+    /// Store that replaces the row during the first EID CAS write and records
+    /// the request's reads and conditional writes.
+    struct EidConflictEcKv {
```
prk-Jr, line 1886: 👍 **praise** — this is the right test double for the claim being made. A PR whose headline is "one conditional write, one follow-up read" needs the read and write *counts* asserted, not just the final row state, and counting them in the store is the only place that can prove it. `with_follow_up_miss` covering the conflict-then-vanish case, and `KvEntry::tombstone` proving a concurrent withdrawal survives a conflict, are both cases that would have been easy to leave untested.

Two other things worth calling out while I am here: dropping `log_id(ec_id)` from the `upsert_partner_ids` error strings tightens identity exposure in logs, and gating `collect_eid_cookie_updates` behind the source check means subresources skip the base64 decode entirely rather than decoding and discarding.

prk-Jr (approved):
## Summary

Reviewed at `a5aad79` against `refactor/remove-legacy-consent-store` (merge-base `aee5bf6`, no drift) in an isolated worktree. Read all six changed files end to end, and scratch-verified the two code suggestions below against the PR head.

The core design holds up. Replacing the five-attempt CAS loop with one conditional write plus one conflict read removes an unbounded write amplifier from the hot path, and the source classification is assigned after pre-route filters — the challenged-request tests prove a DataDome short circuit leaves the request unclassified. The `/auction` and navigation sources both writing the *same* cookie-derived updates means the loser of a concurrent page-load conflict lands on `ConflictMatched` rather than dropping data, which is the property that makes "never retry" safe.

**No blocking findings.** Approving. Ten comments below, ranked: one refactor I would land before merge, one dropped error, two behavioral changes worth an explicit confirmation, and the rest small.

## Cross-cutting

**Adapter scope is correct, and I checked.** `EcFinalizeState` exists only in the Fastly adapter and `ec_finalize_response` has exactly one production call site (`main.rs:308`, via the popped extension), so there is no second router and no Cloudflare/Spin/Axum path silently losing returning-user EID sync. Worth stating because a `set_eid_sync_source` that only one adapter calls is the shape a parity gap usually takes.

**`POST /_ts/page-bids` is no longer an EID sync source.** The test this PR renamed used to name page-bids explicitly alongside `/auction` as a route that saves its first lookup with no later retry. Dropping it is consistent with the PR's goal and navigation covers the same page load, but the old comment made it a deliberate inclusion, so the removal reads as intentional-or-oversight from the diff alone. A line in the PR description would settle it.

**Recovery path still receives its updates.** The `updates` slice is now empty whenever the source is unclassified, and it is also what `confirm_then_recover_orphaned_ec` forwards into a recovered row. That is safe only because `recovery_eligible` implies a navigation, which always classifies — the same coupling the first comment is about. Worth keeping in mind if either gate moves.

## CI Status

All 19 checks **PASS**, including every gate `CLAUDE.md` treats as a PR gate: `cargo fmt`, `cargo test`, `cargo test (axum native)`, `cargo test (cross-adapter parity)`, `cargo test (ts CLI, native)`, spin native + `wasm32-wasip1`, cloudflare native + `wasm32-unknown-unknown`, integration tests, Fastly EC lifecycle, browser integration tests, `vitest`, `format-typescript`, `format-docs`, `Analyze (rust)`, and CodeQL.

Independently re-ran on the PR head: `cargo test -p trusted-server-core --lib ec::` → 382 passed, and `app::tests` → 41 passed.

crates/trusted-server-adapter-fastly/src/app.rs
```diff
@@ -645,6 +645,7 @@ async fn run_named_route(
         NamedRouteHandler::SetTester => handle_set_tester(&state.settings),
         NamedRouteHandler::ClearTester => handle_clear_tester(&state.settings),
         NamedRouteHandler::Auction => {
+            ec.ec_context.set_eid_sync_source(EidSyncSource::Auction);
```
aram356, line 648: ❓ **question** — `/_ts/page-bids` no longer persists returning-user EID cookies.

Before this PR every finalized response that had a KV graph ran the returning-user upsert, and the test comment this PR deletes named `/auction` and `/_ts/page-bids` together as the named routes that "save their first lookup into the context". The `PageBids` arm below (line 666) sets no source, so for a returning user `ec_finalize_response` now skips `sync_eid_cookie_updates` there entirely.

On SPA publishers (the SSR + page-bids setup) a route change never navigates and never calls `POST /auction`; page-bids is the only per-view request that carries the browser's current `ts-eids` / `sharedId`. With this PR those users' IDs sync only on hard reloads.

Is the narrowing to navigation + `/auction` deliberate (it matches #993's list verbatim), or was page-bids overlooked? If deliberate, the PR body and the edge-cookies guide should say so. If not:

```rust
        NamedRouteHandler::PageBids => {
            ec.ec_context.set_eid_sync_source(EidSyncSource::Auction);
```

or a dedicated `EidSyncSource::PageBids` variant so the measurement can tell the two apart. Apply manually — the PageBids arm is outside the diff hunks, so this cannot be a suggestion.

crates/trusted-server-adapter-fastly/src/app.rs
aram356: ♻️ **refactor** — Orphan-recovery eligibility lost its browser gate.

`is_publisher_navigation` used to mean *real browser and document navigation*. It still feeds `set_recovery_eligible` at line 851, under the comment "Authorize orphan recovery only for real-browser document navigations", but it is now just `is_navigation_request(&req)`, so a non-browser navigation whose origin start succeeds becomes recovery-eligible.

Unreachable today: `build_ec_request_state` withholds the finalize KV graph from non-browser requests (except withdrawals, which return before the returning-user branch), so recovery cannot run. It is defense in depth, and the comment is now wrong. Keeping the source classification browser-independent (as your test asserts) while restoring the gate on the recovery flag:

```suggestion
        let is_navigation = is_navigation_request(&req);
        if is_navigation {
            ec.ec_context.set_eid_sync_source(EidSyncSource::Navigation);
        }
        let is_publisher_navigation = ec.is_real_browser && is_navigation;
        if is_publisher_navigation
```

No unit test can pin this: under Viceroy the publisher origin never connects, so `set_recovery_eligible` is never reached in tests (a probe asserting the flag on a browser-less navigation passes on this head for that reason alone). Verified in isolation on a scratch worktree at this head: `cargo fmt --all -- --check` clean, the dispatch and recovery-eligibility tests pass under Viceroy, no post-verification drift.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -479,68 +568,156 @@ impl KvIdentityGraph {
...
+                } else {
+                    EidCookieSyncOutcome::Written
+                },
+            ),
+            Ok(EcKvWriteOutcome::PreconditionFailed) => {
+                let refreshed = self.load_snapshot(ec_id);
+                let Some(refreshed_entry) = refreshed.entry_for(ec_id) else {
+                    let kept = Self::keep_proven(ec_id, refreshed, Some(&current));
+                    return (kept, EidCookieSyncOutcome::DeferredConflict);
+                };
```
aram356, line 680: ♻️ **refactor** — A conflict whose follow-up read misses returns a snapshot carrying the generation the CAS just rejected.

`Self::keep_proven(ec_id, refreshed, Some(&current))` hands back `current` unchanged, and `current` is `Present { generation: Some(g) }` where `g` is exactly the generation the conditional write failed on. Runtime-confirmed: adding

```rust
assert_eq!(snapshot.generation_for(&ec_id), None);
```

to `eid_cookie_sync_keeps_live_proof_when_conflict_follow_up_misses` fails on this head.

Harmless today, since pull sync re-reads the live row before its bulk CAS and withdrawal runs earlier in finalize. But `upsert_partner_ids_from_snapshot` treats `Present { generation: Some }` as CAS input it may use *without a read*, so a snapshot that leaves this function should never advertise a generation known to be stale. Keep the existence proof, drop the generation:

```suggestion
                let Some(refreshed_entry) = refreshed.entry_for(ec_id) else {
                    // The failed CAS proved `current`'s generation stale; keep
                    // only its existence proof so no later write reuses it.
                    let proof = match current {
                        EcKvSnapshot::Present {
                            ec_id: proven_id,
                            entry,
                            ..
                        } => EcKvSnapshot::Present {
                            ec_id: proven_id,
                            entry,
                            generation: None,
                        },
                        other => other,
                    };
                    let kept = Self::keep_proven(ec_id, refreshed, Some(&proof));
                    return (kept, EidCookieSyncOutcome::DeferredConflict);
                };
```

Worth adding the assertion above to that test as well (it depends on this change, so it is not posted as its own suggestion). Verified in isolation on a scratch worktree at this head: `cargo fmt --all -- --check` clean, the 120 `ec::kv` tests pass and the probe assertion passes with the change, no post-verification drift.

crates/trusted-server-core/src/ec/kv.rs
aram356: ♻️ **refactor** — The write-failure log drops the error.

Every other store failure in this module logs the report (`load_snapshot`, the snapshot upsert, pull sync), and `a_store_error_never_carries_the_whole_identifier` already pins that a report never contains the full EC ID, so the identity-free logging goal does not require dropping the cause. As written, a KV outage on this path shows up as a bare "write failed" with no status or reason to correlate with the platform.

```suggestion
            Err(err) => {
                log::warn!("EID cookie sync write failed: {err:?}");
```

Verified in isolation on a scratch worktree at this head: `cargo fmt --all -- --check` clean, the 120 `ec::kv` tests pass, no post-verification drift.

crates/trusted-server-core/src/ec/kv.rs
aram356: 🏕 **camp site** — `upsert_partner_ids` and the `ingest_*` wrappers are dead code re-implemented with new semantics.

`ingest_eid_cookies`, `ingest_prebid_eids` and `ingest_sharedid_cookie` in `prebid_eids.rs` have no callers anywhere in the workspace (finalize uses `collect_eid_cookie_updates` plus the snapshot path; the only other mentions are a comment in `admin.rs` and the old SSC spec). This method exists only for them and for the `PartnerIdBulkWriter` test seam. The PR rebuilt it on `sync_eid_cookie_updates_from_snapshot` and now returns `Ok(())` for `DeferredConflict` and `DeferredFreshness`, reporting success when nothing was written, and its `Failed` arm rebuilds a `kv_error` without the underlying report.

Deleting the three wrappers, the trait, this method and their tests is smaller than maintaining a compat mapping whose semantics now differ from the name. Apply manually — spans `kv.rs` and `prebid_eids.rs`.

crates/trusted-server-core/src/ec/kv.rs
```diff
@@ -479,68 +568,156 @@ impl KvIdentityGraph {
...
+
+        let (mut entry, generation) = match current {
+            EcKvSnapshot::Present {
+                ec_id: ref snapshot_id,
+                ref entry,
+                generation: Some(generation),
+            } if snapshot_id == ec_id => (entry.as_ref().clone(), generation),
+            EcKvSnapshot::Missing { .. } => {
+                let kept = Self::keep_proven(ec_id, current, proven.as_ref());
+                return (kept, EidCookieSyncOutcome::Missing);
```
aram356: ⛏ **nitpick** — `outcome=missing` also counts rows the request already proved exist.

When the request-start snapshot was `Present` for this EC ID but the refresh here misses (edge KV point reads are eventually consistent), `keep_proven` returns the proof and the row is not treated as missing anywhere else, yet the outcome reported to the measurement is `Missing`. The "missing" counter #993 asks for therefore mixes genuinely absent rows with stale reads on live rows, and for `NewEc` the read right after `create_if_absent` is exactly where a stale miss is most likely.

A distinct outcome (say `DeferredStaleRead`), or at least mapping this case to `deferred=1` in `EidSyncMeasurement`, would keep the read-out honest. Prose only — it needs a new variant, its `Display`, and the measurement mapping in `finalize.rs`.

aram356 (changes requested):
## Summary

Reviewed head `c225900` against base `2c5ba81` (`refactor/remove-legacy-consent-store`; the merge base is the base tip, so the PR is current) in an isolated worktree, with runtime probes against the head. The one-write-one-read conflict path, the source gating, and the identity-free measurement are well built and well tested. Requesting changes on one question about scope: `/_ts/page-bids` silently stops persisting returning-user EIDs, which matters for SPA publishers. Everything else is non-blocking. No repository files were modified by this review.

> 3 of the inline comments below carry a one-click `suggestion`. Each was applied alone to a scratch worktree at this head and cleared `cargo fmt --all -- --check`, the targeted tests (the kv module for the two `kv.rs` suggestions, the Fastly dispatch and recovery tests under Viceroy for the `app.rs` one), and a byte-exact pre/post-verification patch check. All three together also cleared `clippy-fastly`, `clippy-cloudflare`, `clippy-spin-native`, `test-fastly` (2704 core tests plus the adapter suites), `test-cloudflare` and `test-spin`. The axum, parity, and wasm clippy lanes could not run on the review machine (local `aws-lc-sys` / `cc` toolchain failures unrelated to the code), so those lanes are not locally verified for the suggestions.

## Blocking

### ❓ question

- **`/_ts/page-bids` no longer persists returning-user EID cookies** — see inline at `crates/trusted-server-adapter-fastly/src/app.rs:648`

## Non-blocking

### ♻️ refactor

- **Orphan-recovery eligibility lost its browser gate** _(suggestion)_ — see inline at `crates/trusted-server-adapter-fastly/src/app.rs:821`
- **A conflict follow-up miss returns the generation the CAS just rejected** _(suggestion)_ — see inline at `crates/trusted-server-core/src/ec/kv.rs:697`
- **The write-failure log drops the error** _(suggestion)_ — see inline at `crates/trusted-server-core/src/ec/kv.rs:711`

### 🏕 camp site / ⛏ nitpick

- **`upsert_partner_ids` and the `ingest_*` wrappers are dead code re-implemented with new semantics** — see inline at `crates/trusted-server-core/src/ec/kv.rs:562`
- **`outcome=missing` also counts rows the request proved exist** — see inline at `crates/trusted-server-core/src/ec/kv.rs:637`

- 🤔 **Deferral is permanent for browser-owned values.** #993 scopes "do not overwrite a different value" to a request that *lost a CAS conflict*. This PR applies it to every write: `apply_cookie_partner_id_updates` defers any different value even when no conflict occurred (`DeferredFreshness`, `WrittenWithDeferredFreshness`). A partner cookie that legitimately changes (a regenerated `sharedId`, a user-ID module re-issuing an ID) can therefore never replace the stored value through this path, and the stale stored value keeps being forwarded as `user.ext.eids` next to the new one from the auction payload. "Deferred to a later navigation or auction" means "never" until a freshness rule exists, because every later request sees the same different value. The PR body states the conservative choice, which is fine, but it is worth saying explicitly that deferred values do not converge on their own.
- 📌 **Follow-up: the freshness rule.** #993's "Prevent older values from replacing newer ones" already sketches storing the time each partner ID was observed. That follow-up is now what turns "deferred" back into "eventually persisted"; suggest filing it before this merges so the permanent-deferral window is tracked.
- 📝 **Docs and spec drift.** `docs/superpowers/specs/2026-07-10-kv-eid-request-snapshot-ec-recovery-design.md` still describes the cookie path as "rereads only on CAS conflict … an exhausted CAS retry returns a failed snapshot" and lists the invariant "CAS conflicts re-merge rather than overwrite concurrent data"; neither holds for `sync_eid_cookie_updates_from_snapshot` any more (one write, no re-merge, different values deferred). `docs/guide/edge-cookies.md`'s sequence ("Next request with ts-eids → Decode cookie and upsert matched partner UIDs") now holds only for document navigations and `POST /auction`. A short addendum, or a spec for #993, would keep the authority the code cites accurate. `docs/guide/integrations/prebid.md` item 5 ("still ingested after the response") stays true for `/auction`.
- 🌱 **Measurement channel.** The measurement is one `log::info!` line per navigation, auction, and new EC. Fine for the first read-out, but the counters #993 asks for (attempts, matches, writes, duplicate conflicts, deferrals, by source) are aggregate questions; the auction telemetry pipeline already carries per-request dimensions to the warehouse, so consider emitting there once the log-based read-out has confirmed the shape.

## CI Status

All 19 reported checks PASS at this head. Required checks are marked.

- Analyze (actions): PASS
- Analyze (rust): PASS
- CodeQL: PASS
- cargo fmt: PASS (required)
- cargo test: PASS (required)
- integration tests: PASS
- vitest: PASS

crates/trusted-server-adapter-fastly/src/app.rs
ChristianPavilonis: Addressed in `1cc78b3`. We split navigation classification from the browser-only recovery gate. EID sync now records any admitted document navigation, while `set_recovery_eligible` still requires `ec.is_real_browser`.

crates/trusted-server-core/src/ec/kv.rs
ChristianPavilonis: Documented in `1cc78b3` in `docs/guide/edge-cookies.md` and the recovery spec. This is intentional for this PR: browser cookies have no trustworthy freshness marker, so a different stored UID is deferred on every browser-cookie sync. Authoritative pull/push can replace it; cookie-only partners wait for a freshness rule or row expiry.

crates/trusted-server-core/src/ec/finalize.rs
ChristianPavilonis: Yes, this narrowing is intentional. Subresources leave the request snapshot `NotRead`, so marker reconciliation has no KV information to act on. Navigation, `/auction`, and admitted `/_ts/page-bids` handle returning-user sync and reconciliation. Consent withdrawal still expires the marker independently on any request. The updated recovery spec records this route scope.

crates/trusted-server-core/src/ec/kv.rs
ChristianPavilonis: Fixed in `1cc78b3`: the write failure now logs `{err:?}` while keeping the message identity-free.

crates/trusted-server-core/src/ec/finalize.rs
ChristianPavilonis: Agreed on the scope. The PR keeps the bounded, identity-free `source`/`outcome` measurement for eligible sync attempts. Suppressed subresources intentionally stop before cookie decode and emit no terminal sync event, so this log is not a denominator for suppressed traffic. The docs and spec describe the route boundary, and aggregate suppression counters remain a follow-up for the telemetry pipeline. We kept the current `info!` level for the initial readout.

crates/trusted-server-core/src/ec/finalize.rs
ChristianPavilonis: Kept `Navigation` explicit here. This helper is only called from the recovery branch gated by a publisher navigation, and the explicit value makes that invariant visible. If recovery eligibility expands to another source, this call site will need to pass that source. No behavior change is needed for this PR.

crates/trusted-server-core/src/ec/kv.rs
ChristianPavilonis: The old compatibility caller was removed in `1cc78b3`. The shared helper still has an empty-update guard as a safe boundary for its internal callers, but finalization returns before calling it with an empty slice. I left that guard as a defensive fast path rather than add an assertion.

crates/trusted-server-core/src/ec/kv.rs
ChristianPavilonis: Fixed in `1cc78b3`: `partner_id_updates_match` now uses the same last-value-wins `BTreeMap` normalization as the write path, with regression coverage.

crates/trusted-server-core/src/ec/mod.rs
ChristianPavilonis: Updated the `EidSyncSource` docs in `1cc78b3`: adapters classify navigation and auction, the page-bids handler classifies admitted SPA requests, and finalization assigns `NewEc`.

crates/trusted-server-core/src/ec/kv.rs
ChristianPavilonis: Removed the unused `PartnerIdBulkWriter`, `upsert_partner_ids`, and `ingest_*` compatibility chain in `1cc78b3`. Tests now exercise collection and the snapshot sync path directly.

crates/trusted-server-core/src/ec/kv.rs
ChristianPavilonis: Thanks. The count-based conflict tests and identity-free error assertions are part of the final change.

crates/trusted-server-adapter-fastly/src/app.rs
ChristianPavilonis: Addressed in `1cc78b3`. `PageBids` now sets its own `EidSyncSource::PageBids` after the route gate, with tests for admitted same-origin and denied cross-site requests. The guide now calls out admitted page-bids as an eligible sync route.

crates/trusted-server-adapter-fastly/src/app.rs
ChristianPavilonis: Addressed in `1cc78b3` by separating `is_navigation` from `is_publisher_navigation`. Source classification remains browser-independent, while recovery eligibility remains browser-gated.

crates/trusted-server-core/src/ec/kv.rs
ChristianPavilonis: Addressed in `1cc78b3`. On a conflict follow-up miss, we preserve only the existence proof and clear the rejected generation. The regression test now asserts `generation_for(...) == None`.

crates/trusted-server-core/src/ec/kv.rs
ChristianPavilonis: Addressed in `1cc78b3`: the write-failure log includes `{err:?}` while remaining identity-free.

crates/trusted-server-core/src/ec/kv.rs
ChristianPavilonis: Addressed in `1cc78b3`: removed the dead wrappers and trait. Tests now validate update collection rather than a compatibility writer.

crates/trusted-server-core/src/ec/kv.rs
ChristianPavilonis: Addressed in `1cc78b3`: a refresh miss after a proven row now returns `DeferredStaleRead`, and `EidSyncMeasurement` counts it as deferred. Regression coverage was added.

aram356:
Pushed `bd0b83a98` to this branch to unblock it.

After the rebase onto `main`, `EcKvStore::list_keys_with_prefix` was a new trait method, but the `EidConflictEcKv` test double in `crates/trusted-server-core/src/ec/kv.rs` was not updated alongside its six sibling doubles. That failed `trusted-server-core` under `cfg(test)` with E0046, which is what took `cargo test` red, and `cargo fmt` red as a cascade.

The fix delegates to the inner store, matching every other double in the module. Locally `cargo fmt --all -- --check` is clean and `cargo test-fastly` passes 2939 tests.

Flagging it here since this is being integrated into the September RC ([#1112](https://github.com/IABTechLab/trusted-server/pull/1112)) - fold it into your next rebase however you prefer.

aram356 (approved):
## Summary

Reviewed head `bd0b83a` against base `e68427e` (`refactor/remove-legacy-consent-store`; merge base is the base tip) in an isolated worktree. Every finding from the two earlier rounds is resolved in this head: `/_ts/page-bids` now classifies itself as `EidSyncSource::PageBids` after the same-origin gate with admitted and denied cases tested, the browser gate on orphan-recovery eligibility is restored, a conflict follow-up miss no longer returns the rejected generation, the write-failure log carries the error, the dead `ingest_*` wrappers and `upsert_partner_ids` are gone, `deferred_stale_read` separates stale reads from real misses, and the edge-cookies guide and the 2026-07-10 spec now state the one-write rule and that deferred browser values do not converge on their own. Locally at this head: the 170 `ec::kv` and `ec::finalize` tests and the Fastly dispatch tests under Viceroy pass. No repository files were modified by this review.

## Non-blocking

### 📝 note

- **PR description is stale.** The summary and the "Fastly routing" row still describe navigation and `POST /auction` as the only returning-user sync sources; the head also classifies admitted `GET /_ts/page-bids` and reports `deferred_stale_read`. A one-line edit keeps the merge commit's description matching the behaviour.

## CI Status

All 20 reported checks PASS at this head. Required checks are marked.

- Analyze (actions): PASS
- Analyze (rust): PASS
- CodeQL: PASS
- cargo fmt: PASS (required)
- cargo test: PASS (required)
- integration tests: PASS
- vitest: PASS

PR #1159, opened 2026-09-10
Prevent accidental Prebid Server stored requests

### LINKED ISSUE #1086: Distinguish empty PBS demand from stored-request intent
## Description

TSJS injects a synthetic `trustedServer` bidder into every ad unit so providers configured with `routing = "all_eligible"`, such as APS, can participate even when the publisher did not configure inline server-side bidders for that slot.

When such a slot has no plan-owned bidder parameters, TSJS sends this envelope:

```json
{
  "bidderParams": {}
}
```

The auction router currently treats missing, null, and empty `bidderParams` as intentional Prebid Server stored-request demand. It routes the slot to every configured `prebid-server` provider and builds:

```json
{
  "ext": {
    "prebid": {
      "storedrequest": {
        "id": "<dynamic-slot-code>"
      }
    }
  }
}
```

For a dynamic slot with no matching stored impression, PBS returns HTTP 400. That rejects the provider's complete OpenRTB request, including otherwise valid impressions with inline bidder parameters. Multiple PBS provider instances repeat the same failing stored-request lookup. Other `all_eligible` providers can still succeed independently.

An empty bidder map is ambiguous. It can mean either "invoke all-eligible providers with no PBS demand" or "load a PBS stored impression." The router should not infer both cases from the same shape.

This was observed while validating #1016, but it is unrelated to per-provider OpenRTB `test: 1` behavior.

## Suggested fix

Add explicit stored-request intent to the `trustedServer` envelope, for example:

```json
{
  "bidderParams": {},
  "storedRequest": false
}
```

TSJS-generated envelopes used only to invoke all-eligible providers should set `storedRequest: false`. Routing should then skip every `prebid-server` provider that has neither routed inline parameters nor stored-request intent, while still admitting eligible non-PBS `all_eligible` providers such as APS. A PBS provider configured as `all_eligible` still has no useful request to make when both inline and stored demand are absent.

Intentional PBS stored demand should use `storedRequest: true`. Omitted values can preserve the legacy stored-request behavior during migration if compatibility requires it. Direct server-generated opportunities should continue to use trusted server-owned routing rather than allowing the browser to select a provider.

### Wire compatibility and internal model

Keep `storedRequest` optional on the wire so the server can distinguish all three states:

| Wire value | Meaning |
| --- | --- |
| `false` | No stored demand. Skip PBS providers without routed inline parameters. |
| `true` | Explicit stored-request intent. |
| omitted | Legacy inference during migration. |

Represent this internally as an enum such as `Disabled`, `Explicit`, and `Legacy`, rather than collapsing omitted and `false` into a Rust boolean. Deploy server support before TSJS starts emitting `storedRequest: false`; an older server will otherwise continue treating the empty envelope as stored demand.

### Stored request identity

The minimal boolean fix can continue deriving the stored impression ID from the slot code. A later design should consider an explicit shape such as `storedRequest: { "id": "homepage-header" }` or `storedRequestId: "homepage-header"`, because dynamic display codes are poor stored-request identifiers. Multiple PBS provider instances also need server-owned provider selection for intentional stored demand instead of automatic fanout. Those changes do not need to block the minimal ambiguity fix.

Regression coverage should include:

- an APS all-eligible slot with no PBS bidder parameters, which reaches APS and skips every PBS provider without demand;
- explicit stored-request intent, which retains the existing PBS path;
- inline bidders split across multiple PBS provider instances;
- one no-PBS slot alongside valid inline PBS impressions, proving it cannot cause a provider-wide HTTP 400;
- `storedRequest: false` with empty bidder parameter objects;
- `storedRequest: false` with only unconfigured bidders after route filtering;
- a malformed non-boolean `storedRequest`, which must fail validation atomically and never trigger legacy stored demand;
- an omitted value from a legacy direct `/auction` caller, which preserves the existing stored-request path;
- TSJS serialization of `storedRequest: false` for an empty all-eligible envelope; and
- a `prebid-server` provider configured as `all_eligible`, which is still skipped when the slot has neither inline nor stored demand.

ChristianPavilonis:
## Summary

- Prevent generated browser auction slots from accidentally requesting Prebid Server (PBS) stored impressions. Previously, an empty bidder map could trigger a lookup using a dynamic slot code; a missing stored impression could reject the whole provider request, including valid sibling impressions.
- Add optional `trustedServer.params.storedRequest` intent: `false` disables stored fallback, `true` permits it, and omission preserves legacy behavior. Filter demandless PBS impressions after server-side parameter overrides without suppressing eligible Amazon Publisher Services (APS) or other providers.
- Preserve publisher intent across repeated and refresh auctions, with regression tests for the serialized browser payload and actual outbound provider request bytes.

## PBS stored-request context

Stored requests are an existing optional Prebid Server feature. A caller normally opts in by referencing configuration already stored in PBS by ID. This PR does not introduce that feature or require new PBS configuration.

The defect was in Trusted Server's fallback behavior: when a slot had no usable inline bidder parameters, it could automatically use the dynamic slot code as a stored-impression ID. That accidentally opted generated slots into stored lookup. This fix makes newly generated envelopes explicitly opt out with `storedRequest: false`, while preserving intentional `true` and the legacy behavior for callers that omit the field.

## Changes

<details>
<summary>Changed files</summary>

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/auction/routing.rs` | Validates stored intent atomically, retains legacy admission facts, and carries intent into provider-local routing. |
| `crates/trusted-server-core/src/auction/openrtb.rs` | Honors intent after bidder overrides, filters paired impressions safely, and skips empty requests before signing or transport. |
| `crates/trusted-server-core/src/auction/openrtb/tests.rs` | Covers explicit, disabled, and legacy fallback, override-populated params, and retained impression identity. |
| `crates/trusted-server-core/src/auction/orchestrator.rs` | Adds transport tests for zero, one, and two active PBS providers alongside APS, including valid bid preservation. |
| `crates/trusted-server-core/src/auction/formats.rs` | Documents the wire field and tests conversion through strict intent admission. |
| `crates/trusted-server-js/lib/src/integrations/prebid/index.ts` | Defaults newly generated envelopes to `false` and preserves authored intent in live ad units and immutable refresh snapshots. |
| `crates/trusted-server-js/lib/test/integrations/prebid/index.test.ts` | Tests initial, repeated, and refresh serialization, including omitted and malformed intent and container-ID lookup. |
| `crates/trusted-server-js/lib/test/core/auction.test.ts` | Confirms the shared serializer retains boolean values, omission, and invalid `null` for server validation. |
| `crates/trusted-server-js/lib/test/prebid-artifact-integration.test.mjs` | Verifies generated and publisher-authored intent through the built adapter. |
| `docs/guide/api-reference.md` | Defines the field, validation behavior, legacy compatibility, and request examples. |
| `docs/guide/integrations/prebid.md` | Explains refresh behavior, server-first deployment, and rollback constraints. |
| `docs/guide/auction-orchestration.md` | Describes provider routing and post-override impression filtering. |
| `CHANGELOG.md` | Records the fix and deployment requirement. |
| `docs/superpowers/plans/2026-09-10-pbs-stored-request-intent.md` | Records the approved plan, scope correction, regression evidence, and verification results. |

</details>

## Scope

This touches 14 files because the intent must survive browser generation, refresh reconstruction, server routing, and final request construction. Most additions are regression tests and documentation; changing only the router would leave a second stored-request fallback in the request builder.

No dependencies, adapter implementations, provider configuration, browser provider selectors, or explicit stored IDs are changed. Legacy inference and intentional stored-demand fanout remain supported. PBS `all_eligible` configuration remains rejected by the existing compiler, so its boundary test is retained rather than enabling a new configuration mode.

## Deployment

**Deploy server support everywhere before serving the new JavaScript.** The old router rejects unknown envelope fields and can drop valid inline demand when it receives `storedRequest`.

Rust artifacts embed the JavaScript bundles. First prepare a server-support-only build retaining old JS emission, then distribute the full build. Keep compatible server admission during rollback while browsers may retain the new JS. Existing legacy callers and intentional stored requests can still reference missing slot-code IDs; this change does not eliminate unrelated PBS validation errors.

## Closes

Closes #1086

## Test plan

Meaningful failing-before assertions were captured for explicit intent admission, generated JS serialization, and the remaining post-override fallback. They pass after the fix. Independent correctness review found no issues.

- [x] `./scripts/test-cli.sh`
- [x] `cargo clippy-cloudflare && cargo clippy-cloudflare-wasm`
- [x] `cargo clippy-spin-native && cargo clippy-spin-wasm`
- [x] JS tests: `cd crates/trusted-server-js/lib && npx vitest run`, 923 tests passed
- [x] JS lint/format/build: `npm run lint`, `npm run format`, and `node build-all.mjs`
- [x] Docs format/build: `cd docs && npm run format && npm run build`
- [x] `git diff --check`
- [ ] Standalone release WASM build: `cargo build --package trusted-server-adapter-fastly --release --target wasm32-wasip1`

Fastly tests ran through Viceroy; Cloudflare and Spin production WASM paths passed Clippy. The full Rust suites passed with 10 pre-existing Fastly tests/doctests ignored. After a mechanical Clippy fix and a test-only example-domain correction, all six Clippy checks and focused routing/OpenRTB tests were rerun successfully; the full suites were not repeated after those two edits. No live PBS service was contacted.

## Checklist

- [x] Changes follow the project coding conventions
- [x] No new `unwrap()` in production code; use `expect("should ...")`
- [x] Existing `log` instrumentation conventions preserved; no `println!` added
- [x] New behavior has regression tests


crates/trusted-server-core/src/auction/openrtb.rs
```diff
@@ -482,35 +485,43 @@ fn apply_prebid(
...
+                })
+                .collect::<Map<_, _>>();
+            let mut prebid = Map::new();
+            if !bidder.is_empty() {
+                prebid.insert("bidder".to_string(), Value::Object(bidder));
+            } else if slot.allows_stored_fallback() {
+                prebid.insert("storedrequest".to_string(), json!({"id": slot.slot().id}));
+            }
+            if prebid.is_empty() {
+                return None;
```
prk-Jr, line 516: 🔧 **[P2] Restrict response admission to impressions actually sent**

Dropping an impression here does not remove it from the response-admission input. `GenericOpenRtbProvider::request_bids_routed` still clones the original `ProviderAuctionInput` into PBS parse state (`provider.rs:401-405`), and PBS builds its allowed impression/dimension index from that input (`integrations/prebid.rs:2066`). With one disabled empty candidate beside a valid inline sibling, the wire request contains only the sibling, but a PBS response naming the dropped candidate is accepted and can win the auction. This breaks the existing rejection of unrequested impressions precisely for the new post-override filtering path.

Carry the emitted impression membership into the provider's response state, and exclude dropped slots before building its dimension index. Add a mixed-request regression that returns both a valid sibling bid and a bid for the omitted impression; only the sibling should survive and the omitted impression should count as `unrequested_impression`.

Proposed implementation shape (manual, illustrative API):

```rust
let sent_impression_ids = request.imp.iter()
    .filter_map(|imp| imp.id.as_deref())
    .collect::<HashSet<_>>();
let response_input = input.filtered_slots(|slot| {
    sent_impression_ids.contains(slot.slot().id.as_str())
});
// Store response_input, rather than input.clone(), in the PBS parse state.
```

Apply manually — this needs a filtered-input helper or equivalent parse-state change across routing/provider code plus a regression test, outside the changed OpenRTB hunk. No one-click suggestion is eligible; the illustrative API above is not an implemented or verified replacement.

prk-Jr (changes requested):
## Summary

Reviewed commit `1e89a976a205bde3feaa9a599f725d99454afecf`.

The explicit intent contract, provider-local filtering, and browser refresh preservation are well covered. One response-admission regression remains: PBS parse state still admits impressions removed from the outgoing request, allowing an unsolicited bid for an omitted slot to win.

## Blocking

- 🔧 **Restrict response admission to impressions actually sent** — see inline at `crates/trusted-server-core/src/auction/openrtb.rs:515–516`.

## Validation

Local checks passed: 222 JS tests, 16 routing tests, 29 OpenRTB tests, one orchestrator transport test, and the 13-module bundle build. An additional scratch transport regression returned a bid for the omitted `fictional-slot`; the assertion that it cannot win failed, confirming the finding. Scratch changes were restored. No live PBS service was contacted; general adapter/lint gates rely on remote CI.

## CI Status

- [integration tests (Fastly EC lifecycle)](https://github.com/IABTechLab/trusted-server/actions/runs/35113296798/job/104855606972): PASS
- [browser integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/35113296798/job/104855606903): PASS
- [integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/35113296798/job/104855606901): PASS
- [CodeQL](https://github.com/IABTechLab/trusted-server/runs/104852468572): PASS
- [cargo test (ts CLI, native)](https://github.com/IABTechLab/trusted-server/actions/runs/35113296973/job/104852242566): PASS
- [Analyze (rust)](https://github.com/IABTechLab/trusted-server/actions/runs/35113297116/job/104852243556): PASS
- [Analyze (javascript-typescript)](https://github.com/IABTechLab/trusted-server/actions/runs/35113297116/job/104852243226): PASS
- [format-docs](https://github.com/IABTechLab/trusted-server/actions/runs/35113297004/job/104852243064): PASS (required)
- [Analyze (actions)](https://github.com/IABTechLab/trusted-server/actions/runs/35113297116/job/104852243536): PASS
- [cargo test (axum native)](https://github.com/IABTechLab/trusted-server/actions/runs/35113296973/job/104852242701): PASS
- [vitest](https://github.com/IABTechLab/trusted-server/actions/runs/35113296973/job/104852242751): PASS
- [CLAUDE.md symlink guard](https://github.com/IABTechLab/trusted-server/actions/runs/35113297004/job/104852242880): PASS
- [cargo check (cloudflare native + wasm32-unknown-unknown)](https://github.com/IABTechLab/trusted-server/actions/runs/35113296973/job/104852242903): PASS
- [Analyze (javascript-typescript)](https://github.com/IABTechLab/trusted-server/actions/runs/35113292938/job/104852236228): PASS
- [format-typescript](https://github.com/IABTechLab/trusted-server/actions/runs/35113297004/job/104852242589): PASS (required)
- [cargo check/build/test (spin native + wasm32-wasip1)](https://github.com/IABTechLab/trusted-server/actions/runs/35113296973/job/104852242804): PASS
- [cargo test (cross-adapter parity)](https://github.com/IABTechLab/trusted-server/actions/runs/35113296973/job/104852242739): PASS
- [prepare integration artifacts](https://github.com/IABTechLab/trusted-server/actions/runs/35113296798/job/104852242153): PASS
- [cargo test](https://github.com/IABTechLab/trusted-server/actions/runs/35113296973/job/104852242713): PASS (required)
- [cargo fmt](https://github.com/IABTechLab/trusted-server/actions/runs/35113297004/job/104852242791): PASS (required)

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -816,6 +827,21 @@ function serverSideBidderParamsForRefresh(
...
+  const match = findRefreshAdUnit(candidateCodes);
+  if (match) {
+    const bid = Array.isArray(match.bids)
+      ? match.bids.find((bid) => bid?.bidder === ADAPTER_CODE)
+      : undefined;
+    return storedRequestParams(bid);
+  }
+  const snapshot = findRefreshSnapshot(candidateCodes);
+  return snapshot ? storedRequestParams({ params: snapshot }) : { storedRequest: false };
+}
```
aram356, line 846: ♻️ **refactor** — This launders a `PublisherAdUnitSnapshot` through a synthetic `{ params: ... }` bid to reuse `storedRequestParams`. It is correct today, but only because of an invisible coupling.

It works because this PR hoisted `storedRequest` to the snapshot's **top level** (`index.ts:385`, as a sibling of `bidderParams` / `clientSideBids` / `zone`), so `hasOwnProperty.call(bid.params ?? {}, STORED_REQUEST_KEY)` at `index.ts:706` happens to read the right field. TypeScript cannot catch a violation, because `PublisherAdUnitSnapshot` is structurally assignable to `Record<string, unknown>` and the value type is `unknown`.

The failure mode is silent and lands exactly on this PR's purpose. If someone later restructures the snapshot to `{ params: {...}, clientSideBids: [...] }` — a natural refactor — this line starts returning `{}` for every snapshot-backed refresh, flipping those slots from "stored fallback disabled" to legacy inference and re-enabling the accidental stored requests this PR exists to prevent. No type error, and no existing test fails: I checked, and every current test moves the write site (`index.ts:743`) and this read site together, so they only break on an *inconsistent* change.

Reading the field directly at the point of use makes the dependency explicit:

```suggestion
  const snapshot = findRefreshSnapshot(candidateCodes);
  if (!snapshot) return { storedRequest: false };
  return Object.prototype.hasOwnProperty.call(snapshot, STORED_REQUEST_KEY)
    ? { storedRequest: copyParamValue(snapshot[STORED_REQUEST_KEY]) }
    : {};
}
```

Scratch-verified on the PR head: `npx vitest run` 981 passed / 45 files, `npm run lint` (eslint `--max-warnings=0`) clean, `prettier --check` clean, and `tsc --noEmit` reports nothing for this file. Byte-exact pre/post patch comparison showed no drift.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -698,7 +700,15 @@ function foldedBidderParams(
   );
 }
 
-/** Capture immutable request-scoped bidder and zone data before the shim mutates an ad unit. */
+/** Preserve authored presence and values, including invalid values for server validation. */
+function storedRequestParams(bid: TrustedServerBid | undefined): { storedRequest?: unknown } {
+  if (!bid) return { storedRequest: false };
+  return Object.prototype.hasOwnProperty.call(bid.params ?? {}, STORED_REQUEST_KEY)
+    ? { storedRequest: copyParamValue(bid.params?.[STORED_REQUEST_KEY]) }
```
aram356, line 707: 🤔 **thinking** — `undefined` and `null` are both "the publisher wrote nonsense here", but they take completely different paths, and the difference is invisible at this line.

I traced both through the real serialization path (`buildRequests` → `buildAdRequest` → `JSON.stringify`) and confirmed by running it:

- `storedRequest: undefined` — `hasOwnProperty` is `true` for an explicitly-assigned `undefined`, so this returns `{ storedRequest: undefined }`. `JSON.stringify` then **drops** the key, the server sees omission, and the slot gets legacy inference.
- `storedRequest: null` — serializes as `"storedRequest": null`, which hits `Some(_) => return None` in `normalize_envelope`. The envelope is rejected atomically, so that slot loses its `bidderParams` **and** its `zone` for PBS.

So a publisher writing `storedRequest: config.enableStored` with a missing config key silently gets legacy behavior, while `storedRequest: config.enableStored ?? null` silently loses all of that slot's PBS demand. The only evidence of the second case is a server-side `record_malformed_envelope` counter the publisher cannot see.

Forwarding invalid values verbatim for server validation is a defensible design and the docstring says so. Being silent about it in the browser is the part worth reconsidering — a one-time `log.warn` when the authored value is not a boolean would turn a silent demand drop into something diagnosable:

```ts
const authored = bid.params?.[STORED_REQUEST_KEY];
if (typeof authored !== 'boolean') {
  log.warn('[tsjs-prebid] non-boolean storedRequest will be rejected by the server', authored);
}
```

Separately, `undefined` is not covered by either `it.each` list (`index.test.ts:1253`, `index.test.ts:2793`) or by `auction.test.ts:93`. Given it is the one non-boolean that behaves differently from all the others, it is the most valuable missing case. Note that asserting it needs `not.toHaveProperty` on the parsed wire object rather than `toEqual`, since `toEqual` treats `{a:1, b:undefined}` and `{a:1}` as equal.

Apply manually — the warn helper and the added test case live outside this hunk.

crates/trusted-server-js/lib/test/core/auction.test.ts
```diff
@@ -90,6 +90,20 @@ describe('auction/buildAdRequest', () => {
     expect(unit2!.bids[0].bidder).toBe('openx');
   });
 
+  it.each([{}, { storedRequest: false }, { storedRequest: true }, { storedRequest: null }])(
+    'retains stored intent presence in serialized shared requests: %j',
```
aram356: ⛏ **nitpick** — This case passes with the entire PR reverted, so the title credits it with coverage it does not provide.

`buildAdRequest` (`src/core/auction.ts:92-104`) copies `params` by reference and neither filters nor rewrites keys, and `src/core/auction.ts` is not in this PR's diff. The test therefore pins a pre-existing pass-through property rather than anything this change introduces.

That is still reasonable regression insurance — it proves the shared serializer will not strip or mangle the new key — but "retains stored intent presence in serialized shared requests" reads like coverage of the new feature, and a reviewer scanning test names would count it as such. Renaming to something like `buildAdRequest passes bid params through verbatim` keeps the value and drops the implication.

Also note the `toEqual` assertion here cannot distinguish "key absent" from "key present with value `undefined`", which is the distinction that matters for the omission case. `prebid-artifact-integration.test.mjs:229` gets this right with `not.toHaveProperty` on the post-`JSON.parse` wire object.

Apply manually — rename only, outside the suggestion-eligible range.

crates/trusted-server-core/src/auction/routing.rs
```diff
@@ -178,8 +179,14 @@ impl ProviderSlotInput {
         self.prebid_zone.as_deref()
     }
 
+    #[cfg(test)]
     pub(crate) fn has_trusted_stored_request(&self) -> bool {
-        self.trusted_stored_request
+        self.bidder_params.is_empty() && self.allows_stored_fallback()
+    }
+
+    pub(crate) fn allows_stored_fallback(&self) -> bool {
```
aram356, line 194: 📝 **note** — The `has_candidates` parameter is dead at the only production call site.

`allows_stored_fallback` passes `!self.bidder_params.is_empty()`, but the one non-test caller is the `else if slot.allows_stored_fallback()` arm in `apply_prebid` (`openrtb.rs:512`), which is reached only when the computed `bidder` map came out empty. So `has_candidates == true` is only reachable there when the slot has routed bidder params that *all* failed the post-override usability filter.

That is not a bug — it is exactly how base's `|| !slot.bidder_params().is_empty()` behaved, and preserving it is the point. Worth a short comment on `StoredRequestIntent::allows_fallback` recording *why* `Legacy` consults `has_candidates`, since the parameter otherwise looks vestigial to a future reader and is an easy thing to "simplify" away — doing so would change the legacy fallback semantics that `openrtb/tests.rs:724` depends on.

Apply manually — a doc comment above `allows_fallback` in `routing.rs:256-264`, outside this hunk.

aram356 (changes requested):
## Summary

The mechanism is correct and the tests are load-bearing rather than decorative. I verified locally against the PR head: 2675 core tests and 981 JS tests pass, `cargo fmt --all -- --check` clean, `cargo clippy-axum` clean, eslint/prettier clean. All 20 remote checks pass.

I checked the two hunks that *read* like regressions and both are safe. The `AllEligible` branch disappearing for prebid providers (`routing.rs:376-382`) is dead code, because `plan.rs:307` rejects `all_eligible` + `prebid-server` at compile time. The `zip` over a `filter_map`-built impression list cannot misalign, because `eligible_banner_slot` guarantees `build_imp` never returns `None` for a routed slot. Neither is obvious from the diff, so both deserve a sentence in the PR description.

The deployment-ordering warning in the docs is accurate and load-bearing: I confirmed the base `normalize_envelope` rejects `storedRequest` as an unknown key, which drops `bidderParams` **and** `zone` for that slot. Old server plus new JS is a real demand-loss hazard, correctly identified.

One blocking finding: replacing a `debug_assert!` invariant with a routine production path that drops demand and emits no metadata, no log, and no counter.

> 1 of the inline comments below carries a one-click GitHub `suggestion` — use **Commit suggestion** to apply it as a commit on the PR branch. The remaining comments describe the fix in prose because the change spans multiple files or lines outside the diff and can't be auto-applied.

## Blocking

### 🔧 wrench

- **`NoImpressions` drops demand with no metadata, log, or counter** — see Cross-cutting below

## Non-blocking

### 🤔 thinking / ♻️ refactor / 📝 note / ⛏ nitpick

- **The original defect is still live for un-upgraded publishers** — see Cross-cutting below
- **`storedRequestParams({ params: snapshot })` launders a snapshot through a fake bid** — see inline at `crates/trusted-server-js/lib/src/integrations/prebid/index.ts:842`
- **`undefined` and `null` authored intent diverge silently** — see inline at `crates/trusted-server-js/lib/src/integrations/prebid/index.ts:707`
- **`auction.test.ts` case passes with the entire PR reverted** — see inline at `crates/trusted-server-js/lib/test/core/auction.test.ts:94`
- **`allows_stored_fallback`'s `has_candidates` argument is dead in production** — see inline at `crates/trusted-server-core/src/auction/routing.rs:187`

- 🔧 **`NoImpressions` drops demand with no metadata, log, or counter** — This PR deletes `debug_assert!(!prebid.is_empty(), "should never route a demandless slot to prebid-server")` and turns that same condition into a routine production path via the new filter in `apply_prebid` plus the early return at `openrtb.rs:266-268`.

  That path lands on an unchanged call site, `crates/trusted-server-core/src/auction/provider.rs:296-301`, which returns a bare `AuctionResponse::no_bid(self.provider_name(), 0)` with empty metadata. To an operator that is indistinguishable from PBS legitimately returning no bids. Partial drops are invisible too: if 3 of 4 impressions are filtered but 1 survives, the request goes out as `Ready` and nothing records that 3 were dropped.

  The reachability change is what makes this matter. On base, `NoImpressions` could only fire when a provider had zero eligible slots from the start, and that case is already annotated upstream by `provider_skipped_response` (`orchestrator.rs:218-223`). This PR adds a second, semantically different trigger — demand filtered out *after* provider-local overrides — and routes it into the same metadata-free response. Once new TSJS ships `storedRequest: false` on every generated envelope, a misconfigured `[auction.bidders]` route produces no PBS request, no metadata, no log line, and no counter.

  The repo already has both idioms to fix this, so it is a small in-pattern change rather than a redesign. Minimum fix at `provider.rs:296` (apply manually — `provider.rs` is outside this PR's diff, so it can't be a suggestion):

  ```rust
  OpenRtbBuildOutcome::NoImpressions => {
      return Ok(ProviderRequestOutcome::Immediate(
          AuctionResponse::no_bid(self.provider_name(), 0).with_metadata(
              "routing",
              serde_json::json!({"skipped_no_usable_demand": true}),
          ),
      ));
  }
  ```

  Better still, count the partial drops in `apply_prebid` using the existing `RoutingDiagnostics` saturating-counter pattern (`routing.rs:70-105`), which would restore the signal the `debug_assert` used to provide. `pbs_disabled_empty_candidate_does_not_become_stored_demand` and the new orchestrator transport test are the natural places to pin it.

- 🤔 **The original defect is still live for un-upgraded publishers** — I verified this with a scratch test rather than inferring it. Routing the envelope `{"bidderParams":{"exampleBidder":{}}}` with no `storedRequest` key through `build_request` still produces:

  ```
  imp[0].ext.prebid = {"storedrequest":{"id":"fictional-slot"}}
  ```

  That is the reported defect — a stored-impression lookup keyed by the dynamic slot code — for any publisher who has not yet received the new JS.

  This is deliberate, not an oversight: base had `|| !slot.bidder_params().is_empty()` in `apply_prebid`, and the PR faithfully preserves it as the `has_candidates` arm of `StoredRequestIntent::Legacy`. The pre-existing test `openrtb/tests.rs:724` (`pbs_empty_params_without_matching_override_fall_back_to_stored_request`) pins exactly this shape, and `creative_opportunities.rs:891` depends on legacy inference, so the compatibility choice is sound.

  The concern is only how it is described. The fix is effective once new JS reaches browsers, which the deployment section correctly says is gated on cached clients. "Closes #1086" reads as fully closed, and the escape hatch — any envelope with non-empty `bidderParams` whose entries are all empty objects — is not called out in the PR body or the api-reference table. Worth one sentence so operators know the server-side hole persists until client rollout completes.

## CI Status

- cargo fmt: PASS (required)
- cargo test: PASS (required)
- cargo test (cloudflare native + wasm32-unknown-unknown): PASS
- vitest: PASS
- integration tests: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS

crates/trusted-server-core/src/auction/openrtb.rs
ChristianPavilonis: Fixed in `f222033f`. PBS parse state now derives its slots from the finalized outbound request impression IDs instead of cloning the unfiltered routed input. The transport regression returns both a valid sibling bid and a higher bid for the omitted impression, then verifies that only the sibling survives and `unrequested_impression` increments. The full Rust test and target-matched Clippy matrix passed.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
ChristianPavilonis: Applied in `99743c09`. Snapshot refresh now reads the top-level `storedRequest` field directly and preserves authored presence with `hasOwnProperty`, without constructing a synthetic bid. The full 984-test JavaScript suite, lint, formatting, type checking, and the 13-module build passed.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
ChristianPavilonis: Added explicit `undefined` wire coverage in `99743c09` for repeated requests, snapshot refresh, live refresh, the shared serializer, and the built adapter. Each assertion inspects parsed JSON with `not.toHaveProperty`, and the API guide now states that JavaScript `undefined` serializes as omission and therefore receives legacy inference. We did not add the browser warning: invalid serializable values intentionally remain server-validated, while logging arbitrary authored values could expose publisher data to console collectors. The full 984-test JavaScript suite passed with no type errors.

crates/trusted-server-js/lib/test/core/auction.test.ts
ChristianPavilonis: Renamed the test in `99743c09` to state the existing pass-through contract rather than crediting the serializer with stored-intent behavior. I also added a separate post-serialization `undefined` case using `not.toHaveProperty`.

crates/trusted-server-core/src/auction/routing.rs
ChristianPavilonis: Documented this compatibility invariant in `99743c09`. The comment now explains that usable overrides become inline demand, while routed candidates that remain empty preserve the pre-intent stored fallback behavior.

ChristianPavilonis:
Addressed the body-level review findings in `99743c09` and the PR description:

- A PBS provider whose impressions are all removed after override processing now returns `routing.skipped_no_usable_demand: true`. The zero, one, and two-provider transport regression pins that metadata. Per the recorded decision, partial-drop counting is deferred until its field name and consumer are defined.
- The PR description now explains the compiled rejection of PBS `all_eligible`, the routed-slot and `build_imp` pairing invariant, and the legacy empty-candidate fallback that remains until upgraded JavaScript reaches publishers.
- The API guide documents that explicit JavaScript `undefined` serializes as omission and receives legacy inference.

Validation passed: all four adapter test aliases, `./scripts/test-cli.sh`, all six target-matched Clippy aliases, Rust formatting, 984 JavaScript tests with no type errors, JS lint/format/build, docs format/build, and `git diff --check`.

crates/trusted-server-core/src/auction/openrtb.rs
```diff
@@ -482,35 +485,43 @@ fn apply_prebid(
...
-        );
-        imp.ext = Some(Map::from_iter([(
-            "prebid".to_string(),
-            Value::Object(prebid),
-        )]));
-    }
+    // Filter paired impressions and slots together so later demand keeps its slot ID.
+    request.imp = std::mem::take(&mut request.imp)
+        .into_iter()
+        .zip(input.slots())
```
prk-Jr, line 491: 🤔 **thinking** — `build_common_request` builds impressions with `filter_map(build_imp)` (line 319), so `request.imp` can in principle be shorter than `input.slots()`, and this `zip` would then pair each impression with the wrong slot.

Before this PR a misalignment only wrote the wrong `imp.ext`. Now it would also delete impressions here, and it would feed the wrong `sent_impression_ids` set in `provider.rs:405` — so valid bids would come back and be rejected as `unrequested_impression`. The `debug_assert_eq!` above is a no-op in release builds.

I checked and this is currently **unreachable**: `eligible_banner_slot` in `routing.rs` admits only formats whose width and height are positive and fit in `i32`, while `build_imp` drops formats only when they exceed `i32::MAX`, so every routed slot yields `Some`. Raising it because the invariant now lives two modules away from the code that depends on it, and the failure mode changed from "wrong extension object" to "silently dropped demand plus rejected bids".

A self-enforcing version of the pairing:

```rust
// inside the filter_map closure, before building `bidder`
if imp.id.as_deref() != Some(slot.slot().id.as_str()) {
    log::error!("openrtb: impression/slot pairing drift; dropping imp");
    return None;
}
```

Apply manually — this changes runtime behaviour, and hardening `build_common_request` so the lengths cannot diverge in the first place may be the better of the two options. Worth a deliberate choice rather than a one-click commit.

crates/trusted-server-core/src/auction/routing.rs
```diff
@@ -422,16 +461,26 @@ fn normalize_slot_demand(
...
+        },
+        ..Default::default()
+    };
     if let Some(envelope) = bidders.get(TRUSTED_SERVER_ENVELOPE) {
         match normalize_envelope(envelope) {
             Some(normalized) => demand = normalized,
-            None => diagnostics.record_malformed_envelope(),
+            None => {
+                diagnostics.record_malformed_envelope();
+                demand = NormalizedSlotDemand::default();
```
prk-Jr, line 482: 📝 **note** — `NormalizedSlotDemand::default()` resolves to `StoredRequestIntent::Disabled`, which suppresses stored fallback unconditionally.

Before this PR, a malformed envelope left `stored_request: false` but `openrtb.rs` still permitted fallback through `|| !slot.bidder_params().is_empty()`. So a slot with a malformed `trustedServer` envelope **and** an independent direct bidder whose params were emptied by provider-local overrides did fall back to `storedrequest.id = <slot code>`. It no longer does.

This is covered by `malformed_stored_intent_rejects_envelope_atomically_but_preserves_direct_demand`, and fail-closed is the right direction. The reason I am flagging it: the PR body, the CHANGELOG entry, and the docs all frame omission as "legacy inference preserved", and this is the one legacy path whose behaviour actually changed. A clause on the CHANGELOG entry noting that a malformed envelope now also disables fallback for that slot's direct demand would keep the record accurate.

crates/trusted-server-core/src/auction/routing.rs
```diff
@@ -178,8 +186,14 @@ impl ProviderSlotInput {
         self.prebid_zone.as_deref()
     }
 
+    #[cfg(test)]
     pub(crate) fn has_trusted_stored_request(&self) -> bool {
```
prk-Jr, line 190: ♻️ **refactor** — This used to return the `trusted_stored_request` field recorded at routing time. It is now `#[cfg(test)]` and derives a different predicate: `bidder_params.is_empty() && allows_stored_fallback()`.

Five call sites still read as "was this slot routed as a trusted stored request" — here in `routing.rs`, plus `formats.rs:786` and `creative_opportunities.rs:2010`. The assertions still catch regressions in `allows_stored_fallback`, so this is naming and indirection rather than a coverage hole, but the name now overstates what the helper knows.

Either rename it to something like `stored_fallback_without_inline_demand()`, or delete it and assert the two production predicates directly at the call sites:

```rust
assert!(slot.bidder_params().is_empty() && slot.allows_stored_fallback());
```

Apply manually — the rename touches `routing.rs`, `formats.rs`, and `creative_opportunities.rs` together, so it cannot be expressed as a single-file suggestion.

crates/trusted-server-js/lib/src/integrations/prebid/index.ts
```diff
@@ -698,7 +700,15 @@ function foldedBidderParams(
   );
 }
 
-/** Capture immutable request-scoped bidder and zone data before the shim mutates an ad unit. */
+/** Preserve authored presence and values, including invalid values for server validation. */
+function storedRequestParams(bid: TrustedServerBid | undefined): { storedRequest?: unknown } {
```
prk-Jr, line 704: 🤔 **thinking** — This deliberately forwards an authored non-boolean (`null`, `"false"`, `0`, and so on) verbatim, per the doc comment, so the server can validate it.

The server-side consequence is atomic rejection: `normalize_envelope` returns `None` for a non-boolean `storedRequest`, which discards the entire envelope — `bidderParams` and `zone` included. So a single publisher typo silently removes every inline PBS bid on that slot, and the only signal is the `malformed_envelope_count` diagnostic. There is no console warning and nothing visible in `window.tsjs`.

The shim already normalizes and filters other params before they reach the envelope, so an alternative is to treat a non-boolean as omission (legacy inference) and emit a console warning, which degrades to the pre-PR behaviour instead of dropping the publisher's demand:

```ts
function storedRequestParams(bid: TrustedServerBid | undefined): { storedRequest?: unknown } {
  if (!bid) return { storedRequest: false };
  if (!Object.prototype.hasOwnProperty.call(bid.params ?? {}, STORED_REQUEST_KEY)) return {};
  const value = bid.params?.[STORED_REQUEST_KEY];
  if (typeof value !== 'boolean') {
    console.warn('[tsjs] ignoring non-boolean trustedServer.params.storedRequest', value);
    return {};
  }
  return { storedRequest: value };
}
```

That trades the server-side diagnostic for client-side resilience, which is a genuine tradeoff rather than a clear win — hence a question, not a change request. Apply manually if you want it; it changes behaviour and the existing tests assert the current pass-through.

prk-Jr (approved):
## Summary

Solid, well-scoped fix. The `StoredRequestIntent` enum preserves every pre-existing legacy admission path while adding explicit opt-in/opt-out, and the post-override impression filter closes the second stored-request fallback in the request builder that a router-only change would have left behind. No blocking findings.

What I checked rather than assumed:

- Compared the old and new admission truth tables path by path. They match on every legacy input (empty `bidders` map; valid envelope with missing/`null`/empty `bidderParams`; non-empty params; envelope absent with direct bidders). Only the two new explicit intents and one malformed-envelope case differ, and that difference is noted below.
- Dropping `routing == AllEligible` from the PBS include arm is a dead path: `plan.rs:307` rejects `all_eligible` for the `prebid-server` profile at compile time.
- Server-generated opportunities are unaffected. `creative_opportunities.rs:891` emits `{"bidderParams":{}}` with no `storedRequest`, which resolves to `Legacy { empty_admission: true }` and retains stored fallback.
- `filtered_slots` feeds `build_bid_dimension_index`, so a bid for an impression that was filtered out resolves to `BidRejectionReason::UnrequestedImpression` rather than being silently admitted.

All four findings are non-blocking and prose-only; none is expressed as a one-click `suggestion`, because each either spans multiple files, changes behaviour, or is a design question rather than a defect.

## Non-blocking

### 🤔 thinking

- **`zip` pairing guarded only by `debug_assert`, and the blast radius is now larger** — see inline at `crates/trusted-server-core/src/auction/openrtb.rs:491`
- **Shim forwards a malformed authored `storedRequest`, which rejects the whole envelope server-side** — see inline at `crates/trusted-server-js/lib/src/integrations/prebid/index.ts:704`

### ♻️ refactor

- **`has_trusted_stored_request()` no longer means what its name says** — see inline at `crates/trusted-server-core/src/auction/routing.rs:190`

### 📝 note

- **Malformed envelope now hard-disables stored fallback even with independent direct demand** — see inline at `crates/trusted-server-core/src/auction/routing.rs:482`

- 👍 **Test coverage is unusually strong** — the matrices cover routing intent x params x direct demand, the orchestrator across zero, one, and two active PBS providers with APS participation preserved, and the `fictional-slot` case proves a bid for an unsent impression is rejected rather than admitted. Routing that rejection through the existing dimension index (`UnrequestedImpression`) instead of an ad-hoc check is the right mechanism.
- 📝 **Deploy ordering is already covered** — the old-server-rejects-unknown-field hazard, and the fact that it drops valid inline demand rather than merely failing to fix the stored lookup, is documented in the PR body, the CHANGELOG entry, and `docs/guide/integrations/prebid.md#stored-intent-deployment`. Noting only that I verified it, not raising it as a finding.
- 📝 **`TrustedProviderRoutes` has no production caller** — the new "server-owned routes do not authorize demandless PBS wire impressions" semantics, covered by `pbs_filtering_keeps_slot_pairs_and_drops_demandless_trusted_routes`, are exercised only by tests today. Pre-existing, not introduced here.

## CI Status

- cargo fmt: PASS (required)
- cargo test: PASS (required)
- vitest: PASS
- Analyze (actions): PASS
- CodeQL: SKIPPED
- Analyze (rust): PENDING

No failing checks at the time of review.

crates/trusted-server-core/src/auction/provider.rs
```diff
@@ -294,10 +294,10 @@ impl GenericOpenRtbProvider {
         )? {
             OpenRtbBuildOutcome::Ready(request) => request,
             OpenRtbBuildOutcome::NoImpressions => {
-                return Ok(ProviderRequestOutcome::Immediate(AuctionResponse::no_bid(
-                    self.provider_name(),
-                    0,
-                )));
+                return Ok(ProviderRequestOutcome::Immediate(
+                    AuctionResponse::no_bid(self.provider_name(), 0)
+                        .with_metadata("routing", json!({"skipped_no_usable_demand": true})),
```
aram356: 📝 **note** — Worth a look, but safe to resolve without changing anything if you disagree.

`materialize_planned_response` (`orchestrator.rs:270-294`) documents itself as *"Skipped providers are routed separately and retain their exclusive skipped diagnostic"*, but the short-circuit at `orchestrator.rs:277-282` only tests for `skipped_no_eligible_slots`. A response carrying the new `skipped_no_usable_demand` key falls through and picks up a sibling `unused_bidder_params_count: 0`.

Behavior is fine: `as_object_mut().insert(...)` merges rather than replaces, so the skip key survives, and `orchestrator.rs:6046-6049` passes through this exact path and asserts it. The `0` is also accurate, since `unused_bidder_params_count` early-returns `0` for PBS.

It is only that the code now contradicts its stated contract — the new skip does not in fact retain an *exclusive* diagnostic. Either adding the new key to the line-277 check, or softening the comment to say only `skipped_no_eligible_slots` is exclusive, would resolve the mismatch. Pure cosmetics; no behavioral difference either way.

crates/trusted-server-core/src/auction/provider.rs
```diff
@@ -399,11 +399,20 @@ impl GenericOpenRtbProvider {
...
-            CompiledOpenRtbProfile::PrebidServer(_) => GenericOpenRtbParseState::Prebid {
-                provider_id: self.provider_name().to_string(),
-                auction_id: input.common_request().id.clone(),
-                input: input.clone(),
-            },
+            CompiledOpenRtbProfile::PrebidServer(_) => {
+                let sent_impression_ids = request
+                    .imp
+                    .iter()
+                    .filter_map(|impression| impression.id.as_deref())
```
aram356: 📝 **note** — Not reachable today; flagging because this commit changes how much weight the invariant carries.

`Imp::id` is `Option<String>` (from the generated OpenRTB schema), so `filter_map` silently skips any impression whose `id` is `None`. Such an impression would be dropped from `sent_impression_ids`, and its slot would then be filtered out of the parse input even though it *was* sent — meaning a legitimate bid for it would be rejected as `unrequested_impression`.

That cannot happen right now. `build_imp` (`openrtb.rs:446-447`) is the only `Imp` construction site and sets `id: Some(slot.slot().id.clone())` unconditionally, and neither `apply_prebid` nor `finalize_request` touches `id`. I verified this rather than assuming it.

The reason it seems worth a note: before this commit, a missing impression ID would have been inert, whereas `sent_impression_ids` is now the response-admission gate, so the invariant became load-bearing while remaining guaranteed only by convention. The generated type cannot be tightened locally, but a cheap assertion before the collect would pin it and mirror the existing style at `openrtb.rs:483-487`:

```rust
debug_assert!(
    request.imp.iter().all(|impression| impression.id.is_some()),
    "should populate every PBS impression ID"
);
```

Entirely optional — resolve if you would rather not add an assertion for an unreachable case.

aram356 (approved):
## Summary

All six findings from the previous pass are resolved, and the blocking one was fixed more thoroughly than requested. Approving.

I re-verified each fix against the code rather than from the replies:

- **`NoImpressions` silent drop** — now carries `routing.skipped_no_usable_demand`. I traced the chain (`types.rs:331` → `ProviderSummary.metadata` → `provider_details`) to confirm an operator actually sees it in the auction response, and the key matches the existing `skipped_no_eligible_slots` shape.
- **Response admission** — `f222033fd` goes further than the review asked, deriving admission from the sent `request.imp` rather than the routed slot list. I confirmed `request` is not mutated between serialization at `provider.rs:339` and the ID collection at `provider.rs:403`, so the gate is built from the same value that produced the wire bytes. I also checked the narrowing is safe for the one consumer that counts over `slots()`: `unused_bidder_params_count` early-returns `0` for PBS at `openrtb.rs:895`, so it never sees the shortened list.
- **The regression test is a real guard.** I reverted `filtered_slots` to `input.clone()` in a scratch worktree and confirmed the test fails loudly at `orchestrator.rs:6031` — `winning_bids.len()` becomes 2 instead of 1, because the 3.0 bid for the never-sent impression wins. It proves rejection, not merely losing.
- **JS snapshot read** — applied as suggested. **`undefined` coverage** — added with `not.toHaveProperty` on parsed wire objects, which is the assertion form that can actually distinguish an absent key. **Test rename** and the **`allows_fallback` comment** are both in.

On the browser warning for non-boolean authored intent: the reasoning for declining it — that logging arbitrary authored values could expose publisher data to console collectors — is a stronger argument than the diagnosability concern that motivated it. Not re-raising.

Verification on this head: `cargo test-axum -p trusted-server-core --lib` 2675 passed, `cargo test-cloudflare` 22, `cargo test-spin` 49/37, parity 13, `npx vitest run` 984 passed / 45 files, `cargo fmt --all -- --check` and `cargo clippy-axum` clean. All 20 GitHub checks pass.

Two optional notes below. Neither affects behavior and neither needs to block merge — resolve them as you see fit.

## Non-blocking

### 📝 note

- **`materialize_planned_response` no longer matches its own doc comment** — see inline at `crates/trusted-server-core/src/auction/provider.rs:299`
- **`Imp::id` is `Option<String>`, and the invariant is now load-bearing** — see inline at `crates/trusted-server-core/src/auction/provider.rs:406`

## CI Status

- cargo fmt: PASS (required)
- cargo test: PASS (required)
- cargo test (cloudflare native + wasm32-unknown-unknown): PASS
- vitest: PASS
- integration tests: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS

PR #1049, opened 2026-08-20
Add full-surface documentation refresh spec

aram356:
## Summary

- Single PR for the full documentation refresh: the design spec plus all eight work packages land here, one commit (or small series) per package, reviewable commit-by-commit.
- The spec was synthesized from a four-track audit of `main` at `2e85a1cdc` (VitePress site, root/crate markdown, in-code docs, config templates), every finding cited at `file:line`, with ground-truth inventories (routes with per-adapter availability, all 15 `Settings` sections, 14-integration capability matrix, `ts` CLI tree) as appendices.

## Changes

| File | Change |
| ---- | ------ |
| `docs/superpowers/specs/2026-08-19-documentation-refresh-design.md` | Design spec: audit findings, source-of-truth map, eight work packages, acceptance criteria |
| WP1 (upcoming commits) | Publishing/policy hygiene: `srcExclude`, CNAME, `fastly.toml` sensitive values, empty `/guide/` page, gate alignment |
| WP2 | Remove fabricated/dead content: dead endpoints, Equativ, `RequestWrapper`, `.with_asset`, GAM/Kargo pages, auction README rot, `FAQ_POC.md`, CHANGELOG/`.env.example` repairs |
| WP3 | Config reference parity with all 15 `Settings` sections and 14 integration configs |
| WP4 | API reference rebuilt from adapter route tables with availability matrix |
| WP5 | New pages: deployment guides, EdgeZero, telemetry/Tinybird, tsjs, GPT handoff, `adserver_mock`, script guards; nav repair; `TESTING.md` rewrite |
| WP6 | Root markdown and crate READMEs |
| WP7 | In-code docs: core module index, `platform/` docs, crate headers, tsjs JSDoc |
| WP8 | Enforcement: docs build on PRs, `cargo doc -D warnings`, doctests in CI, dependabot gaps |

## Closes

Closes #1038
Closes #277
Closes #341

## Test plan

- [x] Docs format: `cd docs && npm run format` (passes on the spec)
- [ ] Docs build: `cd docs && npm run lint && npm run build` (after WP commits land)
- [ ] `cargo fmt --all -- --check` and target-matched clippy/tests for crates touched by WP7
- [ ] `cargo doc --no-deps` warning-free for core and adapters (WP7/WP8)
- [ ] Acceptance greps from WP2-WP4 included in this description when work lands
[...]

PR #1165, opened 2026-09-14
Add a new-engineer setup page alongside the onboarding guide

aram356:
## Summary

Draft, stacked on `spec-docs-refresh` (#1049), rebased onto `5c043a324`. Adds one published VitePress page, `docs/guide/onboarding-new-engineer.md` ("New Engineer Setup"), and a sidebar entry beside the restored Onboarding page.

It is the practical companion to `/guide/onboarding`: day-one access requests, how to run the docs site, a local end-to-end proxy check, and the build traps that cost the most time. It links out to the onboarding guide rather than repeating it.

## Relationship to #1049

#1049 restores `docs/guide/onboarding.md` as a published, sidebar-linked page covering the mental model, request trace, vocabulary, code map, first development loop, and triage map. That is the right home for all of it, and this page no longer duplicates any of it.

This page was originally broader. After the restore landed, it was cut back to only what the onboarding guide does not cover. Its request trace is also more accurate than the one this page previously carried — it names `dispatch_fallback` and the EdgeZero router, where this page had promoted the `run_pre_route_filters` helper to a top-level step — so that section was dropped rather than reconciled.

## What the page covers

**Request access** — Google account for calendar and shared documents, GitHub account with write access, the project board, and a Fastly account with an API token. These are specific to a new hire rather than to working in the codebase, so they do not belong in the general onboarding guide. The section notes that none of them block getting started, since the Axum adapter runs locally with no edge account.

**Reading these docs locally** — `cd docs && npm ci && npm run dev`, served at `http://localhost:5173/trusted-server/`. The base-path suffix is called out because the bare URL redirects rather than serving the home page. Also the pre-pull-request checks, and two authoring conventions: use the `.tool-versions` placeholders rather than literal version numbers, and keep Mermaid node labels short because a box sizes to its explicit line breaks but not to its own text wrapping.

**Prove the first-party proxy locally** — a local origin plus a signed fetch, which exercises signing and proxying end to end.

**Build traps** — a table of environment failures whose error messages do not name their own cause: bare `cargo test` linking against `fastly` symbols (and bare `cargo check` succeeding, giving a false green), a stale `ts` binary rejecting valid configuration, the Rust build requiring Node, `ts` needing the repository root, and the Viceroy pin.

## Test plan

- [x] `cd docs && npm run format` — clean
- [x] `cd docs && npm run lint` — clean
- [x] `cd docs && npm run build` — succeeds; every internal link resolves, and the build fails on dead links
- [x] Verified in a browser against the dev server, not only the static build: the Mermaid `div` is empty until client-side hydration, so a passing build does not prove a diagram renders
- [x] Documented URL requested directly: `http://localhost:5173/trusted-server/` returns 200, bare `http://localhost:5173/` returns 302
- [x] Version placeholders resolve in the built HTML, with none left unresolved

Docs-only: no Rust, JavaScript, workflow, or configuration changes, so the adapter and test gates are unaffected.

## Notes for review

- **Please confirm the project board URL.** `https://github.com/orgs/IABTechLab/projects/3` is carried over from the onboarding guide added in #183; it could not be verified here because the available token lacks the `read:project` scope, and other references in that document had gone stale.
- The Google-account row deliberately says "ask your manager or onboarding contact" rather than naming a person, since contact lists rot and #1049 is moving that material to `internal/`.
- If the maintainers would rather fold these four sections into `/guide/onboarding` as extra sections than carry a second sidebar entry, that is a reasonable call and this PR can be closed in favour of a suggestion on #1049.


aram356:
Revised in `ea597ab86` to reuse the existing onboarding material rather than writing fresh prose.

The guide added in #183 already covered much of this well, and the current 160-line page is what is left after three years of erosion. Recovered from the original:

- the problem statement and what edge position buys
- the architecture diagram
- the first-party proxy before/after contrast
- the key-concepts framing (proxying, identity, integrations, signing)
- the local origin stub smoke test
- the common-issues table

Corrected where it had drifted, rather than copying it forward:

| Original | Now |
| --- | --- |
| `crates/js/lib` | `crates/trusted-server-js/lib` |
| Hardcoded Rust 1.91.1 | `{{RUST_VERSION}}`, read from `.tool-versions` |
| "Synthetic ID" with a `{{ip}}` template | Edge Cookie ID; that config no longer exists |
| `impl Integration for ...` example | Description of the registration hooks — the trait shape changed |
| Single-runtime diagram | All four adapters, plus the ad-stack gate |
| A real origin hostname | `example.com`-style hosts, per the repository guideline |

Common issues keeps only entries that still reproduce, and adds the three that cost a newcomer the most time today: bare `cargo` failing at the workspace root, `ts` needing reinstall after configuration-type changes, and Node being required for the Rust build.

Verified: `npm run format`, `npm run lint`, and `npm run build` pass. Both `{{RUST_VERSION}}` and `{{VICEROY_VERSION}}` resolve in the built HTML to `Rust 1.95.0` and `viceroy --version 0.17.0`, with no unresolved placeholders — so the versions cannot drift from `.tool-versions`.

aram356:
Incorporated the public onboarding content into #1049 in commit 5c043a324, using the existing /guide/onboarding URL. Closing this stacked draft so the work remains in one PR.

aram356:
Closing as superseded by #1049: 5c043a324 restored the public onboarding page at /guide/onboarding (mental model, Fastly request trace, vocabulary, code map, first development loop, triage map). This draft's remaining unique material (access-request checklist, local docs preview, first-party proxy walkthrough, build traps) stays recoverable on the docs/onboarding-new-engineer branch if a follow-up setup page is wanted.

PR #1167, opened 2026-09-15
Show help for a bare ts dev proxy invocation

### LINKED ISSUE #1063: Show help for a bare ts dev proxy invocation
Running `ts dev proxy` with no arguments does not print help. It runs part of proxy startup, then fails with a debug-formatted `error-stack` report that exposes internal source locations.

## Reproducing

On macOS (`ts dev proxy` is macOS-only):

```
$ cargo run -q -p trusted-server-cli --target aarch64-apple-darwin -- dev proxy
sudo: a password is required
warning: Safari: could not auto-restore the system proxy for 'Wi-Fi' (needs admin). ...
[ts] invalid rule configuration
├╴at crates/trusted-server-cli/src/commands/dev/proxy/mod.rs:256:41
│
╰─▶ no rewrite rule: pass --map FROM=TO (or -f/--from with -t/--to)
    ╰╴at crates/trusted-server-cli/src/commands/dev/proxy/config.rs:269:20
```

Two problems, and the second is not in the original report:

1. **Internal paths leak.** `dev/mod.rs:50` renders the report with `format!("{report:?}")`, so every user-facing proxy failure prints `error-stack` debug output with file/line attachments. The underlying message ("no rewrite rule: pass --map FROM=TO") is already good; only the framing is wrong.

2. **Side effects run before the error.** `run` calls `browser::restore_system_proxy_if_pending` at `mod.rs:254` *before* `config::resolve` at `mod.rs:256`. So a bare invocation touches system proxy state and attempts `sudo` before discovering it has no rule to serve. A user who typed `ts dev proxy` to see what it does gets a sudo prompt.

## Suggested approach

Set `arg_required_else_help = true` on `ProxyArgs` (`proxy/mod.rs:81`) so Clap prints help and exits before `run` is ever called. That fixes both problems at once, since nothing in `run` executes.

Watch the interaction with the nested subcommand: `ProxyArgs` carries `command: Option<ProxySub>` (`mod.rs:150`), which is how `ts dev proxy ca install` is reached. `ts dev proxy ca …` takes no rewrite rules and **must keep working**. Verify this by hand, not just by reading — the flag's behavior with an optional subcommand plus all-optional args is worth confirming empirically. If `arg_required_else_help` turns out to break the `ca` path, fall back to an explicit check at the top of `run` that prints help for the no-args, no-subcommand case.

Separately, consider whether `dev/mod.rs:50` should render `{report}` rather than `{report:?}` for user-facing output. That is a slightly wider change — it affects every `ts dev proxy` error, not just this one — so it is acceptable to leave out and file separately if it grows.

## Acceptance criteria

- A bare `ts dev proxy` prints Clap help and exits without touching system proxy state, attempting `sudo`, or printing `error-stack` debug output.
- `ts dev proxy ca path`, `ca install`, `ca uninstall`, and `ca regenerate` still work.
- An incomplete-but-explicit invocation (for example `ts dev proxy --from a.example.com` with no `--to`) still produces the concise no-rule error, not help text.
- A regression test covers the bare invocation. `no_rule_passed_is_a_no_rule_error` in `proxy/config.rs:669` is the neighboring test; a Clap-level test likely belongs with the other `Args::try_parse_from` tests in `run.rs`.
- `cargo clippy` and the CLI test suite pass.

## Getting set up

`ts dev proxy` is macOS-only: its dependencies are cfg-scoped in `Cargo.toml` and the `Proxy` subcommand does not exist on other hosts. The workspace default target is `wasm32-wasip1`, so the CLI needs an explicit host target:

```
cargo run -p trusted-server-cli --target aarch64-apple-darwin -- dev proxy
./scripts/test-cli.sh
```

## Pointers

- `crates/trusted-server-cli/src/commands/dev/proxy/mod.rs:81` — `ProxyArgs`
- `crates/trusted-server-cli/src/commands/dev/proxy/mod.rs:150` — the optional `ca` subcommand field
- `crates/trusted-server-cli/src/commands/dev/proxy/mod.rs:184` — `run`, including the `ca` early return and the ordering problem
- `crates/trusted-server-cli/src/commands/dev/proxy/config.rs:266` — `resolve` and `ConfigError::NoRule`
- `crates/trusted-server-cli/src/commands/dev/mod.rs:50` — the `{report:?}` rendering

dhruv8sh:
## Summary

- A bare `ts dev proxy` now prints Clap help and exits immediately, instead of touching Safari's system proxy state (prompting for `sudo`) and then failing with a debug-formatted `error-stack` report that leaked internal `file:line` paths.
- `ts dev proxy ca …` and an explicit-but-incomplete invocation (`--from` without `--to`) keep working exactly as before — verified this empirically against the pinned `clap` version, not just by reading the derive macro.
- Also switched proxy error rendering from `{report:?}` to `{report:#}` (not plain `{report}`): error-stack's non-alternate `Display` stops after the first context frame, which would have silently dropped the actual cause (e.g. "no rewrite rule: pass --map FROM=TO …") behind the generic wrapper message.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-cli/src/commands/dev/proxy/mod.rs` | Add `#[command(arg_required_else_help = true)]` to `ProxyArgs` so Clap shows help before `run` ever executes on a bare invocation |
| `crates/trusted-server-cli/src/commands/dev/mod.rs` | Render proxy errors with `{report:#}` (chains every `error-stack` context frame) instead of `{report:?}` (leaked source locations) |
| `crates/trusted-server-cli/src/commands/dev/proxy/config.rs` | Fix `base_args()`/`parse_args(&["ts"])` test helpers, which broke from the new `arg_required_else_help` (it applies to any `Command` `ProxyArgs` is flattened into, including these); now restate the `--listen` default explicitly so parsing doesn't hit the same help short-circuit and abort the test binary |
| `crates/trusted-server-cli/src/run.rs` | Add regression tests: bare invocation shows help (and `ca path` still parses under it), and a `--from`-only partial rule still parses instead of showing help |

## Closes

Closes #1063

## Test plan

- [ ] `cargo test-fastly && cargo test-axum` — not applicable, no fastly/axum/core files touched
- [ ] `cargo clippy-fastly && cargo clippy-axum` — not applicable
- [ ] JS tests: `cd crates/trusted-server-js/lib && npx vitest run` — not applicable
- [ ] JS format: `cd crates/trusted-server-js/lib && npm run format` — not applicable
- [ ] Docs format: `cd docs && npm run format` — not applicable
- [ ] WASM build: `cargo build --package trusted-server-adapter-fastly --release --target wasm32-wasip1` — not applicable
- [ ] Manual testing via `fastly compute serve` — not applicable (CLI-only, not a Fastly adapter change)
- [x] Other:
  - `cargo clippy -p trusted-server-cli --target aarch64-apple-darwin --all-targets --all-features -- -D warnings` — clean
  - `./scripts/test-cli.sh` (native `x86_64-unknown-linux-gnu`, macOS `cfg` gates temporarily removed locally to actually execute the macOS-only code on this Linux box) — 175 passed, 1 unrelated pre-existing failure (`restore_system_proxy_if_pending_removes_file_with_empty_service`, a genuinely macOS-only test hitting its own no-op branch off-macOS; untouched by this change)
  - Live-ran `ts dev proxy`, `ts dev proxy ca path`, and `ts dev proxy --from a.example.com` under the same native build to confirm real output matches the acceptance criteria

## Checklist

- [x] Uses `tracing` macros (not `println!`) — no new logging added
- [x] New code has tests


crates/trusted-server-cli/src/commands/dev/mod.rs
```diff
@@ -31,6 +31,6 @@ pub enum DevCommand {
 pub fn run(command: DevCommand) -> Result<(), String> {
     match command {
         #[cfg(target_os = "macos")]
-        DevCommand::Proxy(args) => proxy::run(&args).map_err(|report| format!("{report:?}")),
+        DevCommand::Proxy(args) => proxy::run(&args).map_err(|report| format!("{report:#}")),
```
prk-Jr: **Blocking.** `{report:#}` chains *context* frames, but `error-stack` 0.6's `Display` impl explicitly discards attachments:

```rust
// error-stack-0.6.0/src/fmt/mod.rs:1164
FrameKind::Context(context) => Some(context.to_string()),
FrameKind::Attachment(_) => None,
```

Two error paths in `proxy/mod.rs` carry their entire user-facing message in an attachment rather than a context:

- `proxy/mod.rs:218` — "could not revoke the previously-installed CA from the keychain; aborting regenerate... Remove the old CA manually (Keychain Access), then retry."
- `proxy/mod.rs:235` — "could not remove old CA file {path} during regenerate ({err}); aborting so the stale key is not silently reused"

Before this change `{report:?}` printed those. After it, a user who hits either abort on `ts dev proxy ca regenerate` sees exactly:

```
certificate authority error
```

Confirmed empirically rather than by reading the impl:

```
ALT       = [certificate authority error]     # Report::new(CertAuthority).attach("could not revoke ...")
PLAIN     = [certificate authority error]
CFG_ALT   = [invalid rule configuration: no rewrite rule: pass --map FROM=TO (or -f/--from with -t/--to)]
CFG_PLAIN = [invalid rule configuration]
```

So the reasoning in the PR description holds for `ConfigError` and fails for the CA paths. `attach_printable` is not the fix — `Display` drops that too. I checked the other `.attach` sites (`upstream/mod.rs`, `upstream/connect.rs`); those are per-request and terminate in a 502 plus a log, so they never reach this formatter. The blast radius is precisely the two `ca regenerate` aborts.

**Preferred fix:** drop this hunk from the PR and file it separately, per #1063's own guidance. The bare-invocation AC is already satisfied without it, since `run` never executes.

**If you would rather keep it here,** promote both messages to contexts so `{report:#}` renders them, and add a test asserting the rendered string (nothing currently covers error rendering):

```rust
// proxy/mod.rs — ProxyError
/// The previously-installed CA could not be revoked from the OS trust store.
#[display(
    "could not revoke the previously-installed CA from the keychain; aborting regenerate so \
     on-disk key material still matches OS trust. Remove the old CA manually (Keychain \
     Access), then retry."
)]
CaRevokeFailed,
/// An old CA file could not be removed during regenerate.
#[display("could not remove old CA file {path} during regenerate ({source}); aborting so the stale key is not silently reused")]
CaFileRemoval { path: String, source: String },
```

then `Report::new(ProxyError::CaRevokeFailed)` / `Report::new(ProxyError::CaFileRemoval { .. })` at `proxy/mod.rs:218` and `:235`.

**Third option,** if the goal is only to stop the `at file:line` leak that #1063 actually reported: keep `{report:?}` and install `Report::install_debug_hook::<core::panic::Location>(|_, _| {})` at CLI startup.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
```diff
@@ -359,14 +359,7 @@ mod tests {
...
-        #[derive(clap::Parser)]
-        struct W {
-            #[command(flatten)]
-            a: crate::commands::dev::proxy::ProxyArgs,
-        }
-        W::parse_from(["ts"]).a
+        parse_args(&["ts", "--listen", "127.0.0.1:18080"])
     }
 
     fn parse_args(argv: &[&str]) -> crate::commands::dev::proxy::ProxyArgs {
```
prk-Jr, line 370: Non-blocking, but worth fixing while you are here. `parse_args` uses `Parser::parse_from`, and `arg_required_else_help` now applies to every `Command` that `ProxyArgs` is flattened into — including this test wrapper. On the help short-circuit, `parse_from` calls `Error::exit()`, i.e. `std::process::exit(2)`. That does not fail one test; it aborts the whole test binary with no test name and no attribution:

```
error: test failed ... process didn't exit successfully: ... (exit status: 2)
note: test exited abnormally
```

Reproduced by adding a defaults-only `parse_args(&["ts"])` call. This PR works around the two existing call sites by passing `--listen`, but the trap stays armed for whoever writes the next defaults-only test. Switching to the fallible parser turns it back into an ordinary test failure:

```rust
        W::try_parse_from(argv).expect("should parse proxy args").a
```

Verified: compiles, `cargo fmt --check` clean, all 17 `config` tests pass.

`crates/trusted-server-cli/tests/support/mod.rs:207` has the identical `Wrapper::parse_from` shape. All its current call sites pass arguments, so it is not live today, but it is the same latent trap.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
```diff
@@ -359,14 +359,7 @@ mod tests {
...
     fn base_args() -> crate::commands::dev::proxy::ProxyArgs {
-        // Construct via clap so defaults match the real surface.
-        use clap::Parser;
-        #[derive(clap::Parser)]
-        struct W {
-            #[command(flatten)]
-            a: crate::commands::dev::proxy::ProxyArgs,
-        }
-        W::parse_from(["ts"]).a
+        parse_args(&["ts", "--listen", "127.0.0.1:18080"])
```
prk-Jr: Minor. The deleted comment ("Construct via clap so defaults match the real surface") was carrying a real guarantee, and it is gone now: `127.0.0.1:18080` is duplicated between `proxy/mod.rs:98`'s `default_value` and two spots here, with no test asserting they agree. A future change to the real default leaves `base_args()` silently pinned to the stale value.

Cheapest fix is a shared const referenced from both sides — clap 4 accepts a `&'static str` expression in `#[arg(default_value = ...)]`. For the `clap_parses_rewrite_host_as_a_bool` call site below, any neutral flag sidesteps the duplication entirely, e.g. `parse_args(&["ts", "--insecure"]).rewrite_host`.

crates/trusted-server-cli/src/run.rs
```diff
@@ -678,4 +678,25 @@ mod tests {
...
+    fn dev_proxy_bare_invocation_shows_help_before_running() {
+        let error = Args::try_parse_from(["ts", "dev", "proxy"])
+            .expect_err("a bare `ts dev proxy` should short-circuit to help, not run");
+        assert_eq!(
+            error.kind(),
+            clap::error::ErrorKind::DisplayHelpOnMissingArgumentOrSubcommand,
+            "should print help instead of touching system proxy state or attempting sudo"
+        );
+        parse(&["ts", "dev", "proxy", "ca", "path"]);
+    }
```
prk-Jr: Minor. #1063's acceptance criteria list all four `ca` actions ("`ts dev proxy ca path`, `ca install`, `ca uninstall`, and `ca regenerate` still work"), but only `path` is exercised. This also splits the bundled second assertion out of a test whose name is solely about the bare invocation.

Verified: `cargo fmt --check` clean, 3/3 tests pass.

```suggestion
    }

    #[test]
    #[cfg(target_os = "macos")]
    fn dev_proxy_ca_subcommands_still_parse_under_arg_required_else_help() {
        for action in ["path", "install", "uninstall", "regenerate"] {
            parse(&["ts", "dev", "proxy", "ca", action]);
        }
    }
```

crates/trusted-server-cli/src/commands/dev/proxy/mod.rs
```diff
@@ -78,6 +78,7 @@ async fn finish_interrupted_run<Restore, Stop, Drain>(
 
 /// `ts dev proxy [OPTIONS]` — see the design spec §4.
 #[derive(Debug, clap::Args)]
+#[command(arg_required_else_help = true)]
```
prk-Jr, line 85: Nice. This is the load-bearing line, and verifying the `arg_required_else_help` interaction with an optional subcommand empirically — rather than trusting the derive macro — is exactly what #1063 asked for. Pairing it with `dev_proxy_partial_rule_parses_instead_of_showing_help` pins the non-obvious half too: an explicit-but-incomplete invocation must *not* short-circuit to help.

prk-Jr (changes requested):
Reviewed at `1476137e` in a clean worktree. The `arg_required_else_help` change is correct, well-targeted, and closes every acceptance criterion in #1063 on its own — and confirming the optional-subcommand interaction empirically rather than trusting the derive macro is exactly what the issue asked for.

One blocking ask, on the `{report:#}` hunk that rides along. Three non-blocking nits follow as suggestions.

**Blocking:** `{report:#}` drops every `.attach(...)`, which silently guts the two `ca regenerate` abort messages. Details inline on `dev/mod.rs`. My recommendation is to **drop that hunk from this PR and file it separately**, exactly as #1063 invited ("a slightly wider change — it affects every `ts dev proxy` error, not just this one — so it is acceptable to leave out and file separately if it grows"). It needs its own tests (there are none for rendered error output) and a design call on attachments-vs-contexts. The `arg_required_else_help` half is ready to land without it.

**Test coverage against the issue's acceptance criteria:**

| AC | Covered |
| --- | --- |
| Bare `ts dev proxy` prints help, no sudo / no side effects | yes — `dev_proxy_bare_invocation_shows_help_before_running` |
| `ca path`, `install`, `uninstall`, `regenerate` still work | partial — only `path` (suggestion on `run.rs`) |
| Incomplete-but-explicit invocation still gives the concise no-rule error | yes — new parse test plus the pre-existing `no_rule_passed_is_a_no_rule_error` |
| Regression test covers the bare invocation | yes |
| `cargo clippy` and the CLI suite pass | yes — verified below |

**Verified locally**, against `1476137e` on `aarch64-apple-darwin`:

- `cargo clippy -p trusted-server-cli --target aarch64-apple-darwin --all-targets --all-features -- -D warnings` — clean
- `cargo test --package trusted-server-cli --target aarch64-apple-darwin --lib` — 177 passed, 0 failed (both new tests execute on macOS)
- Each suggestion below applied in isolation and re-verified (fmt clean, tests pass)
- Fastly/axum/cloudflare/spin gates not run — no files outside `trusted-server-cli` are touched

Note that `gh pr checks 1167` currently reports no checks on this branch. Both new tests are `#[cfg(target_os = "macos")]`, so only the `test-cli` job (`runs-on: macos-latest`) actually executes them; the Linux `test-axum` job's `./scripts/test-cli.sh` compiles them out.

crates/trusted-server-cli/src/commands/dev/mod.rs
dhruv8sh: Good catch, and confirmed against the fmt source — I hadn't accounted for `error-stack` 0.6 dropping attachments in `Display`. Went with your third option's spirit but simpler: reverted `dev/mod.rs` to `{report:?}` in ae1d7c6a, since the bare-invocation AC this PR targets never reaches `run()` at all (clap short-circuits to help first), so the formatter change wasn't load-bearing for this PR's actual scope. Filing the CA-attachment→context promotion separately as you suggested rather than scope-creeping it in here.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
dhruv8sh: Fixed in ae1d7c6a — switched `parse_args` to `try_parse_from(...).expect("should parse proxy args")`. Verified the exit trap is real: a defaults-only `parse_args(&["ts"])` call aborts the whole test binary with `parse_from`, and becomes an ordinary panic/failure with `try_parse_from`. Left `tests/support/mod.rs:207` alone since it's genuinely out of scope here (no live call site), but noted it for whoever touches that next.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
dhruv8sh: Done in ae1d7c6a: added `pub const DEFAULT_LISTEN` in `proxy/mod.rs`, used it both in the arg's `default_value` and in `base_args()`, and switched the `clap_parses_rewrite_host_as_a_bool` false-case to your suggested neutral flag (`--insecure`) so it no longer needs the listen literal at all.

crates/trusted-server-cli/src/run.rs
dhruv8sh: Applied your suggested diff verbatim in ae1d7c6a — split out `dev_proxy_ca_subcommands_still_parse_under_arg_required_else_help` covering all four actions (path/install/uninstall/regenerate), and the bare-invocation test now only asserts the help short-circuit.

crates/trusted-server-cli/src/commands/dev/proxy/mod.rs
dhruv8sh: Thanks!

dhruv8sh:
Follow-up self-review after ae1d7c6a caught two more small issues in the same code: a local `use clap::Parser;` inside `parse_args` (violates the repo's no-local-imports rule) and `tests/support/mod.rs`'s `resolve()` still using `Wrapper::parse_from` — the identical process-exit trap this PR just armed via `arg_required_else_help`, now fixed the same way as `config.rs`. Both addressed in 85ac1ff3.

dhruv8sh:
Closing in place of #1176 since this came from a forked repository.

PR #1171, opened 2026-09-16
Support Linux in ts dev proxy

### LINKED ISSUE #1170: Support Linux in ts dev proxy
## Problem

`ts dev proxy` is compiled only on macOS, so Linux developers cannot use it to test production hostnames against local or staging upstreams.

The proxy engine already uses portable native networking libraries. Linux support needs command/dependency gates, browser discovery, and certificate trust management rather than a separate proxy engine.

## Scope

- Enable the proxy and its tests on Linux.
- Support native Chrome/Chromium and Firefox with temporary profiles.
- Manage user-level browser certificate trust safely, including uninstall and rotation.
- Keep Safari macOS-only and preserve macOS behavior.
- Add Linux CI coverage and setup documentation.

System-wide proxy/trust changes, Windows, and Snap/Flatpak browser automation are outside this scope.

## Verification

Run native CLI tests and lint on Linux and macOS. Verify browser certificate trust and revocation with disposable profiles and certificate stores, without disabling TLS verification or the browser sandbox.

ChristianPavilonis:
## Summary

- Make `ts dev proxy` available on Linux. It was previously excluded at compile time even though the proxy engine can run natively on Linux.
- Launch native Chrome/Chromium and Firefox, and manage the development certificate authority through the user's NSS certificate database on Linux. Certificate query or removal failures stop rotation before the old key is changed.
- Add Linux CI coverage, regression tests, an isolated browser check, and setup documentation while retaining macOS support.

## Scope

This is a CLI-only platform extension, not a proxy-engine rewrite. The changes span command registration, browser launch, certificate trust, tests, CI, and documentation because removing the compile gates alone would leave Linux browsers without working trust management.

Most of the new code and tests cover safe certificate installation and removal. A reproduced NSS behavior required an extra check: importing a different certificate with the same subject can make the original certificate's nickname export ambiguous. Installation now rejects that conflict before import, and shared-database locks serialize `ts` trust changes.

Supported automation is limited to native Chrome/Chromium and Firefox. Safari stays macOS-only. Windows, Snap/Flatpak automation, root operations, and desktop-wide proxy or certificate-store changes are excluded. Known packaging paths are skipped, but arbitrary shell wrappers are not classified.

## Changes

CLI paths below are relative to `crates/trusted-server-cli/`; other paths are repository-relative.

| File | Change |
| ---- | ------ |
| `.github/workflows/test.yml` | Run native CLI tests and Clippy on both Ubuntu and macOS. |
| CLI `Cargo.toml` | Enable Linux proxy dependencies and scope the existing certificate parser to Linux trust checks. |
| CLI `src/lib.rs` | Enable CLI output support on Linux. |
| CLI `src/run.rs` | Update developer-command help to include Linux. |
| CLI `src/commands/dev/mod.rs` | Register and dispatch the proxy on Linux. |
| CLI `src/commands/dev/proxy/mod.rs` | Wire typed trust operations and locking into CA commands and initial loading. |
| CLI `src/commands/dev/proxy/browser.rs` | Discover Linux browser launchers, retain temporary-profile settings, and skip Firefox launch if certificate import fails. |
| CLI `src/commands/dev/proxy/config.rs` | Make browser selection platform-aware and reject Safari on Linux. |
| CLI `src/commands/dev/proxy/trust.rs` | Manage recorded certificate destinations, identity checks, locks, installation, removal, and macOS keychain errors. |
| CLI `tests/proxy_cli.rs` | Prove the Linux proxy command is available. |
| CLI `tests/proxy_e2e.rs` | Enable the existing proxy end-to-end tests on Linux. |
| CLI `tests/proxy_perf.rs` | Make the existing opt-in performance workloads available on Linux. |
| CLI `tests/proxy_trust_linux.rs` | Cover failed trust operations, retries, shared stores, unrelated certificates, and same-subject conflicts. |
| CLI `tests/proxy_trust_macos.rs` | Test keychain-query failures using a fake command, without touching a real keychain. |
| `scripts/test-linux-dev-proxy-browser.py` | Provide a repeatable Chrome/Chromium trust and revocation check using disposable directories. |
| `docs/guide/ts-dev-proxy.md` | Document Linux setup, trust scope, browser restrictions, and certificate rotation. |
| `docs/superpowers/plans/linux-dev-proxy.md` | Record the approved scope, implementation decisions, and verification evidence. |

## Closes

Closes #1170

## Test plan

- [x] Linux command regression failed before the change and passed afterward.
- [x] Same-subject certificate regression failed before the fix and passed afterward.
- [x] `./scripts/test-cli.sh`: 182 unit tests and 42 integration/configuration tests passed.
- [x] `cargo test_cli_linux --test proxy_trust_linux -- --include-ignored`: all 11 trust tests passed, including real NSS checks.
- [x] `cargo clippy --package trusted-server-cli --target x86_64-unknown-linux-gnu --all-targets -- -D warnings`
- [x] Docs format: `cd docs && npm run format`, with the existing JS dependency directory added to PATH because docs dependencies were not installed separately.
- [x] Chrome 153 and Chromium 151 passed XDG, default, and legacy certificate-database layouts. Each check proved missing trust fails, installed trust loads, revoked trust fails after restart, and HTTP bypasses a stopped proxy.
- [x] Independent correctness review completed with no remaining blockers under the native-install-only scope.
- [ ] macOS CI and real keychain execution: not run locally; the CI matrix includes macOS.
- [ ] Native Firefox runtime smoke: Firefox was not installed on the test host; launch arguments and preferences have unit coverage.

All real certificate/browser tests used disposable HOME, XDG, CA, profile, and NSS directories. They did not modify the developer's real trust stores or disable TLS verification or the browser sandbox. Adapter, JavaScript, and WASM suites were not rerun because those implementations are unchanged. Manual performance workloads remain opt-in.

## Checklist

- [x] Changes follow project conventions and remain within the CLI platform boundary.
- [x] No new `unwrap()` in production code.
- [x] Existing CLI output and error-reporting conventions are preserved.
- [x] New behavior has regression tests.
- [x] No secrets or credentials committed.


crates/trusted-server-cli/src/commands/dev/proxy/trust.rs
```diff
@@ -0,0 +1,547 @@
...
+        })?;
+        if !remaining.is_empty() {
+            return Err(Report::new(TrustError::State));
+        }
+        Ok(certificate.subject().as_raw().to_vec())
+    }
+
+    /// NSS named exports include every certificate with the same subject.
+    /// Reject a different certificate with that subject before importing anything.
+    fn check_subject_conflicts(entry: &Destination) -> Result<()> {
```
prk-Jr, line 283: 🤔 **thinking** — `ca install` scans and exports *every* nickname in the shared user NSS database, and hard-fails on any one of them.

`check_subject_conflicts` walks all nicknames from `nicknames()` and runs `certutil -L -n <nick> -a` per entry. Any failure — a duplicate nickname, an orphaned entry, a cert whose named export isn't parseable PEM — propagates as `TrustError::Command` / `TrustError::State` and aborts the install.

On the fixture databases in `tests/proxy_trust_linux.rs` that is fine. On a real developer's `~/.pki/nssdb` (corporate roots, leftovers from `mkcert` or earlier `nss-tools` experiments) a single bad entry makes `ca install` permanently unusable, and there is no documented recovery path other than editing the NSS store by hand — which `docs/guide/ts-dev-proxy.md` tells the user not to do.

Fail-closed is right for an *identity* check, but a per-nickname export failure isn't an identity conflict. Two options that keep the guarantee:

- Only hard-fail when a successfully-parsed certificate has a matching subject and different bytes; treat an unparseable/failed export as "not a conflict for this subject" and warn.
- Keep the strict behaviour and document the escape hatch (which nickname failed, and what to do about it) — the `checked()` attach already carries the certutil stderr, so the diagnostic exists, it just isn't actionable in the guide.

crates/trusted-server-cli/src/commands/dev/proxy/trust.rs
```diff
@@ -0,0 +1,547 @@
...
+fn lock_file(path: &Path) -> Result<File> {
+    let file = OpenOptions::new()
+        .create(true)
+        .truncate(false)
+        .read(true)
+        .write(true)
+        .mode(0o600)
+        .open(path)
+        .change_context(TrustError::Io)?;
+    file.try_lock().change_context(TrustError::Busy)?;
```
prk-Jr, line 59: 🤔 **thinking** — every `try_lock` failure is reported as "another operation is running".

`TryLockError` covers `WouldBlock` *and* genuine I/O failures — `ENOLCK`, or a filesystem that doesn't support advisory locks. NFS-mounted home directories are a live case on Linux, which is the platform this PR is adding.

Both map to `TrustError::Busy`, so the user is told `another CA or NSS trust operation is running; retry after it exits` and retries forever against a condition that will never clear. Splitting the two keeps the diagnostic honest:

```rust
match file.try_lock() {
    Ok(()) => {}
    Err(std::fs::TryLockError::WouldBlock) => return Err(Report::new(TrustError::Busy)),
    Err(err) => return Err(Report::new(err).change_context(TrustError::Io)),
}
```

Apply manually — worth confirming the exact `TryLockError` variant names against the pinned 1.95 toolchain before committing to this shape.

crates/trusted-server-cli/src/commands/dev/proxy/trust.rs
```diff
@@ -0,0 +1,547 @@
...
+                entry.database.display(),
+                entry.nickname
+            )));
+        }
+        Ok(true)
+    }
+
+    fn install_into(ca_dir: &Path, cert_path: &Path, database: &Path) -> Result<()> {
+        let mut entries = read_record(ca_dir)?;
+        fs::create_dir_all(database).change_context(TrustError::Io)?;
```
prk-Jr, line 332: 🤔 **thinking** — a rejected install still leaves a freshly created NSS store behind.

`install_into` runs `fs::create_dir_all(database)` here and `initialize()` (which runs `certutil -N`) on the next line, both *before* the `contains()` / `check_subject_conflicts()` preflight further down. So an install rejected for a same-subject conflict creates `~/.local/share/pki/nssdb` with a new `cert9.db` / `key4.db` that did not exist beforehand.

It's benign — browsers create that store themselves — but it's a side effect of a command that otherwise advertises "rejected before mutation", and `real_nss_same_subject_install_conflict_is_rejected_before_mutation` doesn't cover it because the database already exists in that fixture. If the ordering can't move (the preflight needs a queryable database), it may be worth saying so in a comment here.

crates/trusted-server-cli/src/commands/dev/proxy/trust.rs
```diff
@@ -0,0 +1,547 @@
...
+        let database = fs::canonicalize(database).change_context(TrustError::Io)?;
+        // All ts CA directories serialize changes to this shared NSS destination.
+        let _database_lock = lock_file(&database.join("ts-dev-proxy-trust.lock"))?;
+        initialize(&database)?;
+        let certificate = certificate(cert_path)?;
+        // The hash only names the entry. Full DER equality authorizes all mutations.
+        let mut hash = DefaultHasher::new();
```
prk-Jr, line 339: 🌱 **seedling** — the persisted nickname is derived from `DefaultHasher`, whose output isn't stable across Rust releases.

The resulting `ts-dev-proxy-{:016x}` string is written to `managed-nss-trust.json` *and* into the NSS database. Today that's safe, and the comment right above says why: the entry lookup keys on `(database, certificate)` and `uninstall` replays the recorded nickname, so cross-version stability is never required on the happy path.

The one case it bites is recovery: journal lost, or a different `--ca-dir` used, while the NSS import survives. The orphan can then only be re-derived under the exact toolchain that wrote it, so `ca uninstall` can never reach it. A stable digest over the DER — SHA-256 is already reachable through the rustls/ring stack this crate pulls in — would make that recoverable.

Not for this PR; the 16-hex-digit shape `read_record` validates would be unchanged either way.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
```diff
@@ -46,8 +46,8 @@ pub enum ConfigError {
          can reach the proxy. Bind a loopback address, or drop --basic-auth."
     )]
     BasicAuthNonLoopback { value: String },
-    /// An unknown browser name was passed to `--launch`.
-    #[display("unknown browser `{value}` (expected chrome|firefox|safari|all)")]
+    /// An unknown or unsupported browser was passed to `--launch`.
+    #[display("unsupported browser `{value}` (use chrome|firefox|all; safari is macOS-only)")]
```
prk-Jr, line 50: ⛏ **nitpick** — the message drops `safari` from the accepted list on macOS, where it *is* accepted.

A macOS user who typos `--launch chrom` is told to "use chrome|firefox|all", which reads as though safari isn't an option on the platform they're actually on. The trailing clause is doing double duty as both a restriction and an implicit list member.

Something that reads correctly on both targets without needing a `cfg`:

```rust
#[display("unsupported browser `{value}` (use chrome|firefox|all, plus safari on macOS)")]
```

Apply manually — left as prose rather than a one-click suggestion so the wording stays yours.

crates/trusted-server-cli/tests/proxy_trust_linux.rs
```diff
@@ -0,0 +1,575 @@
...
+        let bin = self.root.path().join("bin");
+        fs::create_dir_all(&bin).expect("should create tools directory");
+        let path = bin.join("certutil");
+        fs::write(&path, format!("#!/bin/sh\n{body}\n")).expect("should write tool");
+        fs::set_permissions(&path, fs::Permissions::from_mode(0o700))
+            .expect("should make executable");
+        bin
+    }
+
+    fn assert_rotation_fails_unchanged(&self, path: &Path) {
```
prk-Jr, line 76: 👍 **praise** — `assert_rotation_fails_unchanged` re-asserting the key, the certificate *and* the journal bytes on every failure path is the right shape for testing a trust-store mutator: it turns "rotation never proceeds on an unconfirmed revoke" into a property rather than a comment.

Same for the design it's testing — record-before-mutate journal, full DER equality (not the nickname) as the authorization check, a CA-directory lock plus a per-NSS-directory lock so distinct `--ca-dir`s serialize against a shared store, and the same-subject preflight that stops NSS's named-export ambiguity from silently rebinding someone else's certificate. That last one is a genuinely non-obvious NSS behaviour to have found and closed before shipping.

prk-Jr (approved):
## Summary

Extends `ts dev proxy` from macOS-only to macOS + Linux: dependency and `cfg` gates widened, Chrome/Chromium and Firefox launcher discovery added, and a new `trust.rs` that manages the dev CA through the user's NSS database. The CA-trust work is the substantial part, and it is careful — imports are journalled before NSS is touched, full DER equality authorizes every mutation, and two lock layers serialize CA commands and shared-store writes.

No blocking findings. The comments below are design observations, one seedling, and documentation drift; none of them gate the merge.

All comments are prose — nothing is offered as a one-click `suggestion`, because the two documentation lines worth changing fall outside the diff hunks, the script fix is a file-mode change, and the remainder are judgement calls that should stay the author's.

## Non-blocking

### 🤔 thinking

- **Whole-store NSS export scan hard-fails on any single bad entry** — see inline at `crates/trusted-server-cli/src/commands/dev/proxy/trust.rs:283`
- **Every `try_lock` failure reports as "another operation is running"** — see inline at `crates/trusted-server-cli/src/commands/dev/proxy/trust.rs:59`
- **A rejected install still creates a new NSS store** — see inline at `crates/trusted-server-cli/src/commands/dev/proxy/trust.rs:332`

### 🌱 seedling

- **Persisted nickname derives from `DefaultHasher`** — see inline at `crates/trusted-server-cli/src/commands/dev/proxy/trust.rs:339`

### ⛏ nitpick

- **`ConfigError::Browser` omits `safari` on macOS** — see inline at `crates/trusted-server-cli/src/commands/dev/proxy/config.rs:50`
- **Firefox trust flags drifted from the documented manual command** — below
- **The guide's `--help` transcript is now wrong on Linux** — below
- **New script is not executable** — below

### 👍 praise

- **Trust-mutation design and its failure-path assertions** — see inline at `crates/trusted-server-cli/tests/proxy_trust_linux.rs:76`

- ⛏ **Firefox trust flags drifted from the documented manual command** — `import_firefox` (`trust.rs:94`) narrowed the NSS trust flags from `CT,,` (the previous `browser.rs` code) to `C,,`. Fine on the merits: server-authentication CA trust is all a proxy needs, and dropping the client-auth bit is tighter. But the manual import command in `docs/guide/ts-dev-proxy.md:200` still says `-t "CT,,"`, so a developer following the documented path now grants different trust bits than `--launch firefox` does. Either align the doc block to `C,,` or note why they differ. (Body-level: line 200 is outside the diff hunks, so it can't carry an inline comment.)

- ⛏ **The guide's `--help` transcript is now wrong on Linux** — `docs/guide/ts-dev-proxy.md:347` advertises `--launch <LIST>  Browsers to launch (chrome,firefox,safari or all)`, but `safari` is rejected on Linux by the new `cfg` in `Browser::parse_list`. Line 355's `--ca-dir` default reads `~/Library/Application Support/trusted-server/dev-proxy on macOS` with no Linux equivalent, while the prose added above it documents `$XDG_DATA_HOME/trusted-server/dev-proxy`. Separately, and pre-existing on `main`: neither line matches clap's real output — the actual `--launch` help string is *"Browsers to launch + configure (comma list or `all`)"*, and `ca_dir` is an `Option<String>` with no `default_value`, so clap prints no default at all. The transcript is hand-maintained and already drifting; this PR is a reasonable moment to regenerate or trim it.

- ⛏ **`scripts/test-linux-dev-proxy-browser.py` is mode `644`** — every other file in `scripts/` is `755`. It carries a `#!/usr/bin/env python3` shebang but cannot be executed directly. Nothing breaks, since the guide invokes it as `python3 scripts/...`, but `git update-index --chmod=+x` would make it consistent with `test-cli.sh` and friends.

## CI Status

- cargo test (ts CLI, native) (ubuntu-latest): PASS
- cargo test (ts CLI, native) (macos-latest): PASS
- cargo fmt: PASS (required)
- cargo test: PASS (required)
- vitest: PASS
- Analyze (actions): PASS
- CodeQL: SKIPPED
- Analyze (rust): PENDING
- integration tests: PENDING
- integration tests (Fastly EC lifecycle): PENDING

No failed checks. Four checks were still in progress at review time; the Linux and macOS CLI jobs this PR actually exercises have both passed.

crates/trusted-server-cli/src/commands/dev/proxy/trust.rs
```diff
@@ -0,0 +1,547 @@
...
+                Report::new(TrustError::Command).attach("invalid NSS nickname listing")
+            })?;
+            if trust.split(',').count() != 3
+                || !trust
+                    .bytes()
+                    .all(|byte| byte == b',' || byte.is_ascii_alphabetic())
+            {
+                return Err(Report::new(TrustError::Command).attach("invalid NSS trust attributes"));
+            }
+            names.push(name.trim_end().to_owned());
```
aram356: 🔧 **wrench** — NSS accepts and preserves trailing whitespace in a nickname, but `certutil -L`'s listing renders it indistinguishably from the column padding. `trim_end()` here therefore produces a name that `certutil -L -n` cannot resolve, and `checked()` turns that non-zero exit into a hard error.

Reproduced against a real `certutil` (NSS 3.x):

```text
$ certutil -A -n "Some Corp CA " -t "C,," -i c.pem -d sql:$D
IMPORT OK
$ certutil -L -d sql:$D
Some Corp CA                                                 C,,
$ certutil -L -n "Some Corp CA" -a -d sql:$D     # the name this parser yields
exit 255   certutil: Could not find cert: Some Corp CA
$ certutil -L -n "Some Corp CA " -a -d sql:$D    # the real nickname
exit 0
```

Feeding that captured listing through this function's exact logic returns `Ok(["Some Corp CA"])` — the trailing space is gone.

**Why the blast radius is larger than `install`.** `nicknames()` also backs `contains()` (line 313), which `uninstall` calls at lines 409 and 412, and which `regenerate` reaches through `uninstall`. So a single *unrelated* third-party certificate with a trailing-space nickname in the user's Chrome/Chromium NSS database makes the dev CA impossible to install, impossible to remove, and impossible to rotate — all reported as a generic `browser trust command failed` with nothing pointing at the foreign entry as the cause.

**Second-order effect on the conflict check.** Two distinct certificates under `"Corp CA"` and `"Corp CA "` render identically in the listing and both parse to `"Corp CA"`; the exact-match export then returns only one of them. `check_subject_conflicts` consequently scans the same certificate twice and never inspects the other, so a same-subject conflicting CA hiding behind a trailing-space nickname defeats the check that lines 281-283 exist to enforce. That is a soundness gap in a security control, not only an availability bug.

**Proposed fix** (apply manually — a parsing-strategy change plus a regression test; cannot be expressed as a single-range `suggestion`):

```rust
// Split on the fixed column boundary that certutil emits rather than
// rsplit_once + trim_end, so a nickname's own trailing spaces survive.
// Additionally, treat a nickname that fails to re-export as a
// skip-with-warning rather than a hard failure of the whole operation,
// so one foreign entry cannot block install/uninstall/regenerate.
```

A regression test can be written without NSS installed by feeding the captured listing bytes above straight into `nicknames()`.

crates/trusted-server-cli/src/commands/dev/proxy/trust.rs
```diff
@@ -0,0 +1,547 @@
...
+        let list = certutil(&["-L", "-d", db])?;
+        let text = String::from_utf8(list.stdout).change_context(TrustError::Command)?;
+        let mut names = Vec::new();
+        for line in text.lines().map(str::trim).filter(|line| !line.is_empty()) {
+            if line == "SSL,S/MIME,JAR/XPI"
+                || (line.starts_with("Certificate Nickname") && line.ends_with(" Trust Attributes"))
+            {
+                continue;
+            }
+            let (name, trust) = line.rsplit_once(char::is_whitespace).ok_or_else(|| {
```
aram356: 🔧 **wrench** — NSS also accepts newlines inside a nickname. The listing then wraps across lines and `rsplit_once` treats each fragment as its own row, so the first fragment fails the trust-attribute validation below.

Reproduced against a real `certutil`:

```text
$ certutil -A -n "$(printf 'Line One\nLine Two')" -t "C,," -i c.pem -d sql:$D
IMPORT OK
$ certutil -L -d sql:$D
Line One
Line Two                                            C,,
```

Through this function's exact logic: `Err("invalid NSS trust attributes: \"One\" (line \"Line One\")")` — `"Line One"` splits into `("Line", "One")` and `"One"` fails the three-comma check.

This shares the root cause and the blast radius of the trailing-whitespace finding on line 264: it propagates as `TrustError::Command` out of `install`, `uninstall`, and `regenerate` alike. Severity is lower because it fails **closed** — an error, never a wrong trust decision — and a newline nickname is rarer than a trailing space. The same column-boundary parse fixes both.

**Proposed fix** (apply manually — same change as the line 264 finding):

```rust
// Parse against certutil's fixed column layout so a wrapped nickname is
// reassembled rather than being read as two malformed rows.
```

For context, the header-skip logic immediately above this is correct — I confirmed `SSL,S/MIME,JAR/XPI` is a genuine second header line that equals the literal exactly after `str::trim`. Only the row parse needs changing.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
```diff
@@ -107,14 +107,20 @@ impl Browser {
     /// Returns [`ConfigError::Browser`] on an unknown name.
     pub fn parse_list(raw: &str) -> Result<Vec<Self>, ConfigError> {
         if raw.trim() == "all" {
-            return Ok(vec![Self::Chrome, Self::Firefox, Self::Safari]);
+            return Ok(vec![
+                Self::Chrome,
+                Self::Firefox,
+                #[cfg(target_os = "macos")]
```
aram356, line 117: ♻️ **refactor** — `Browser::Safari` is declared unconditionally at line 99, but both of its construction sites are `#[cfg(target_os = "macos")]` (this one and the `"safari"` match arm at 123-124). On Linux the type therefore admits a state the parser can never produce, while `browser.rs:80` still dispatches `Browser::Safari => launch_safari(cfg)` into a function compiled on Linux with an entirely macOS body (`networksetup`, `route`, `sudo`, `open -a Safari`).

I verified this degrades safely today — `detect_network_service()` returns `None` on non-macOS, so `launch_safari` warns and returns — so this is a **latent trap rather than a live bug**. It bites when a second constructor appears: a clap `ValueEnum`, a config-file default, or a test fixture would all compile happily and then take an unreachable-by-design path at runtime instead of failing the build.

Cfg'ing the variant itself makes the state unrepresentable on Linux and gets the dispatch compiler-checked:

```rust
pub enum Browser {
    Chrome,
    Firefox,
    #[cfg(target_os = "macos")]
    Safari,
}
```

The three existing cfgs in `parse_list` then stay exactly as they are and become *required* rather than conventional; the compiler will point at `browser.rs:80`, which needs a matching `#[cfg(target_os = "macos")]`; and `launch_safari` plus its macOS support cast (`detect_network_service`, `get_auto_proxy_state`, `restore_auto_proxy`, `manual_restore_command`) can then be cfg'd out of the Linux build rather than compiled and never called. The test at 640-659 already spells the cfg'd-element form, so it needs no change.

Apply manually — the change spans `config.rs` and `browser.rs` together, so it cannot be a single-file `suggestion`.

For the record, the cfg-inside-`vec![]` and cfg-on-a-match-arm technique used here is perfectly legitimate stable Rust and reads fine; the objection is only to the variant's unconditional declaration, not to how `parse_list` is written.

crates/trusted-server-cli/src/commands/dev/proxy/browser.rs
```diff
@@ -317,21 +228,58 @@ fn launch_chrome(cfg: &ResolvedConfig) {
...
-        // Linux: the `google-chrome` launcher (unreached on the macOS-only build).
-        Command::new("google-chrome")
-    }
-    #[cfg(not(any(target_os = "macos", target_os = "linux")))]
-    {
-        Command::new("chrome")
+        find_native_browser(CHROME_LAUNCHERS, |name| {
+            which::which(name)
+                .ok()
+                .and_then(|path| path.canonicalize().ok())
```
aram356: ♻️ **refactor** — this closure is byte-identical to the one in `firefox_command()` at lines 379-383:

```rust
which::which(name).ok().and_then(|path| path.canonicalize().ok())
```

The `canonicalize()` is the security-relevant half: without it, a symlink from `/usr/bin` into `/snap` or `/flatpak` would slip past the component check in `find_native_browser` at lines 261-266, which is the mechanism the PR relies on to refuse packaged browsers. Two copies means a future hardening of the lookup has to remember both sites, and a fix applied to one is silently absent from the other.

**Proposed fix** (apply manually — the change touches two separate hunks in this file, so it cannot be a single-range `suggestion`):

```rust
#[cfg(target_os = "linux")]
fn lookup_on_path(name: &str) -> Option<std::path::PathBuf> {
    which::which(name)
        .ok()
        .and_then(|path| path.canonicalize().ok())
}
```

Both call sites then collapse to `find_native_browser(CHROME_LAUNCHERS, lookup_on_path)` and `find_native_browser(&["firefox"], lookup_on_path)`.

Injecting `lookup` as a parameter on `find_native_browser` was the right call — it is what makes the test at 748-766 possible without touching the real `PATH`, and that test covers discovery order, total absence, and the snap rejection. This finding is only about the two duplicate closure bodies.

Minor, while you are here: `firefox_command()` carries a doc comment but `chrome_command()` has none.

crates/trusted-server-cli/src/commands/dev/proxy/trust.rs
```diff
@@ -0,0 +1,547 @@
...
+                entries.push(Destination {
+                    database,
+                    nickname,
+                    certificate,
+                });
+                entries.len() - 1
+            }
+        };
+        let entry = &entries[index];
+        contains(entry)?;
```
aram356, line 360: 🤔 **thinking** — the `bool` is discarded here, but the identical call at line 372 uses its value, which makes this line read as a dropped result rather than a deliberate choice.

Having traced it, the behaviour is correct: `contains()` returns `Ok(true)` only when the entry is present *and* DER-identical (line 320 has already raised `TrustError::Identity` on a mismatch), `Ok(false)` when absent, and in both cases the right next action is to import. The `?` is doing the load-bearing work of surfacing the identity conflict; the presence answer genuinely is not needed. I also confirmed against a real `certutil` that re-importing an identical certificate under the same nickname is idempotent in NSS — the listing still shows one row and `-r` returns a single DER — so the re-import is neither incorrect nor meaningfully wasteful.

The suggestion is just to make that intent legible to the next reader, either by binding the value or by stating it above the call:

```rust
// Called for its identity-conflict check only; an already-present
// identical certificate is re-imported idempotently either way.
contains(entry)?;
```

This file otherwise comments exactly this class of subtlety well — see the notes at lines 307, 334, and 338 — so this line stands out as the outlier.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
aram356: ⛏ **nitpick** — the parenthetical is unconditional, so on macOS, where `safari` is a perfectly valid value, a user who typos `netscape` is told that `safari is macOS-only` on the one platform where it works.

Either cfg the hint so each platform lists what it actually accepts, or drop the platform note from the shared message and let the Linux-specific rejection carry it.

aram356 (changes requested):
## Summary

Extends `ts dev proxy` from macOS-only to macOS + Linux: NSS-based CA trust, native browser discovery, a Linux CI matrix, and substantial regression coverage. The platform split is well constructed and the shared/cfg'd division is right — `certutil`, `checked`, `lock`, `lock_file`, and `import_firefox` are genuinely shared, and only `install`/`uninstall`/`ensure_can_generate` are conditional, which is correct since the keychain and NSS have no common substrate.

Two blocking defects, both in the `certutil -L` listing parser and both reproduced against a real `certutil`. They share one root cause: scraping a fixed-width human-readable table cannot losslessly recover an NSS nickname. Their blast radius is wider than it first appears — `nicknames()` feeds `contains()`, which `uninstall` and `regenerate` also depend on, so an unrelated pre-existing entry in the user's NSS database can make the dev CA impossible to install, remove, *or* rotate.

All inline comments describe fixes in prose; none is expressible as a one-click `suggestion` (findings 1 and 2 need a parsing-strategy change plus tests, and the rest either span two call sites or touch lines outside the diff hunks).

## Blocking

### 🔧 wrench

- **`nicknames()` strips real trailing whitespace, blocking install, uninstall, and regenerate** — see inline at `crates/trusted-server-cli/src/commands/dev/proxy/trust.rs:264`
- **A newline in any nickname hard-fails the same three commands** — see inline at `crates/trusted-server-cli/src/commands/dev/proxy/trust.rs:254`

## Non-blocking

### ♻️ refactor

- **`Browser::Safari` is representable but unconstructible on Linux** — see inline at `crates/trusted-server-cli/src/commands/dev/proxy/config.rs:113`
- **Duplicated `which` + `canonicalize` lookup closure** — see inline at `crates/trusted-server-cli/src/commands/dev/proxy/browser.rs:278-281`

### 🤔 thinking

- **`contains(entry)?;` discards its `bool` beside an identical call that uses it** — see inline at `crates/trusted-server-cli/src/commands/dev/proxy/trust.rs:357`

### ⛏ nitpick

- **`safari is macOS-only` hint is emitted on macOS too** — see inline at `crates/trusted-server-cli/src/commands/dev/proxy/config.rs:50`

- 📝 **Checked and found correct, recorded so these are not re-raised** — several plausible-looking concerns were investigated against the running system and are *not* defects: `DefaultHasher` nicknames are deterministic across processes (and the journal is authoritative regardless, so cross-release hash drift is survivable); `File::try_lock` is stable on the pinned 1.95.0 and is OS-released on `SIGKILL`; the macOS `errSecItemNotFound` → exit-code-44 assumption is correct (verified against `security find-certificate` on a real host); switching the macOS path from `.status()` to `.output()` is safe because `security add-trusted-cert` authenticates via a GUI dialog rather than a TTY prompt; `write_record` persists at mode `0600`; `uninstall`'s ordering is genuinely fail-safe (the record only shrinks after a removal is confirmed); and the absence of a `#[cfg(not(any(macos, linux)))]` arm in `chrome_command`/`firefox_command` is fine because the whole `proxy` module and its dependency set are gated at `dev/mod.rs:1` and `Cargo.toml:35`.

- 📝 **On the platform-abstraction question: do not introduce a `TrustStore` trait.** The `#[cfg]` resolves at compile time, so there is never more than one implementation in a given build — a trait would add a vtable or a generic parameter for zero dispatch, and object-safety would force `&self` onto what are genuinely free functions over a `&Path`. Cfg'd free functions behind a re-export is the correct idiom here and matches what `std::sys` does. The asymmetry between the inline `mod linux` (9 helpers, a `Destination` type, its own tests) and the bare macOS items (one helper) tracks a real size asymmetry and is justified. Two small follow-ups worth considering, neither blocking: a one-line doc comment above the `pub(super) use linux::{...}` re-export naming the three-function platform contract (nothing currently makes the compiler check that the two arms agree in signature — they would diverge silently until someone builds the other platform); and if `mod linux` grows past its current ~375 lines, promoting it to its own file.

- 🌱 **Consider obtaining nicknames in a machine-readable form.** Both blocking findings dissolve if the tool stops parsing the human-readable table. Worth evaluating whether the certificate set can be enumerated without relying on `certutil -L`'s column layout, which is a presentation format with no stability guarantee across NSS releases.

## CI Status

- integration tests: PASS
- CodeQL: PASS
- cargo test (ts CLI, native) (macos-latest): PASS
- cargo test (ts CLI, native) (ubuntu-latest): PASS
- Analyze (actions): PASS
- Analyze (rust): PASS
- cargo fmt: PASS
- cargo test: PASS
- format-docs: PASS
- format-typescript: PASS
- vitest: PASS

Reviewer-side verification on the PR head: `cargo clippy --package trusted-server-cli --target aarch64-apple-darwin --all-targets -- -D warnings` clean, and the macOS CLI test suite passes (29 unit + 7 perf + 1 keychain).

crates/trusted-server-cli/src/commands/dev/proxy/browser.rs
```diff
@@ -317,21 +228,58 @@ fn launch_chrome(cfg: &ResolvedConfig) {
...
+    candidates
+        .iter()
+        .filter_map(|name| lookup(name))
+        .find(|path| {
+            !path
+                .components()
+                .any(|part| matches!(part.as_os_str().to_str(), Some("snap" | "flatpak")))
+        })
+        .map(Command::new)
+}
```
dhruv8sh, line 269: 🤔 **thinking** — The snap/flatpak filter checks path *components* literally, but Ubuntu's default Firefox/Chromium packaging never puts `snap` in the resolved path at all.

```rust
/// Discovers launchers in stable order, skipping known Snap/Flatpak paths.
/// Shell wrappers are not classified; callers must use native installations.
fn find_native_browser(
    candidates: &[&str],
    lookup: impl Fn(&str) -> Option<std::path::PathBuf>,
) -> Option<Command> {
    candidates
        .iter()
        .filter_map(|name| lookup(name))
        .find(|path| {
            !path
                .components()
                .any(|part| matches!(part.as_os_str().to_str(), Some("snap" | "flatpak")))
        })
        .map(Command::new)
}
```

On stock Ubuntu (20.04+), the `firefox` transitional `.deb` package (and `chromium-browser`) installs a real, non-symlink file at `/usr/bin/firefox` that internally `exec`s `snap run firefox`. `which::which("firefox")` resolves to that same `/usr/bin/firefox` path, and `canonicalize()` doesn't change it — there's no `snap` path component anywhere, because the redirection happens *inside* the script/binary, not via a filesystem symlink. `find_native_browser` classifies this as native and launches it directly with `-profile <tmpdir> --no-remote`.

On the exact platform and audience this PR targets (Ubuntu, which is also what CI runs on), `ts dev proxy --launch firefox` can therefore hand a snap-confined Firefox a profile path it may not have permission to use, silently defeating the CA-trust flow the whole tool exists to support — while the tool believes it launched a native browser and prints no warning. The doc comment's own caveat ("Shell wrappers are not classified") is exactly the gap, but it undersells it: this isn't a rare wrapper, it's the default install on the most common Linux desktop.

For contrast: I checked this on an Arch machine, where `/usr/bin/firefox` is also a shell-script wrapper (`exec /usr/lib/firefox/firefox "$@"`), but a harmless one with no sandboxing — so this specific failure mode is Ubuntu/Debian-snap-specific, not universal to "firefox is a wrapper" in general. Flagging as reasoned-through for the Ubuntu case rather than empirically reproduced there — I did not install snap Firefox to confirm the confinement failure mode itself.

crates/trusted-server-cli/src/commands/dev/proxy/trust.rs
ChristianPavilonis: Replaced the whole-store nickname/export scan in 3e081339. The preflight now uses `certutil -L -n <absolute CA certificate path> -a` to obtain the matching-subject certificate set, without reconstructing foreign nicknames from the display table. It validates the returned certificates and requires the expected CA.

I kept query/export/parse failures blocking rather than skipping them: an unreadable result cannot establish that there is no conflicting certificate. The guide now explains recovery, including preserving the CA files and journal. All 18 trust tests pass, including real-NSS whitespace/newline cases and failed, empty, and truncated subject-query output.

crates/trusted-server-cli/src/commands/dev/proxy/trust.rs
ChristianPavilonis: Fixed in 3e081339. Only `TryLockError::WouldBlock` maps to `TrustError::Busy`; other lock failures retain their cause under `TrustError::Io`. The contention test now checks the error context explicitly. Compiled, tested, and linted on the pinned Rust 1.95.0 toolchain.

crates/trusted-server-cli/src/commands/dev/proxy/trust.rs
ChristianPavilonis: Clarified this boundary in the code and guide in 3e081339. Initialization can leave a fresh empty database after a later failure. Certificate trust and the managed journal are not changed until preflight succeeds; the existing same-subject conflict regression still verifies that the populated certificate database is unchanged. Keeping initialization before the query rather than attempting rollback of NSS files.

crates/trusted-server-cli/src/commands/dev/proxy/trust.rs
ChristianPavilonis: Deferred as suggested. The journal remains authoritative for the stored nickname, and mutations still require full DER equality. A stable digest alone would not implement recovery after losing the journal, since uninstall currently follows recorded destinations and nicknames rather than reconstructing them. Keeping journal-loss recovery outside this Linux platform-extension PR; the guide continues to require preserving the journal and using the original CA directory.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
ChristianPavilonis: Applied the suggested wording in 3e081339: `use chrome|firefox|all, plus safari on macOS`. It now includes Safari without presenting it as an accepted Linux option.

crates/trusted-server-cli/tests/proxy_trust_linux.rs
ChristianPavilonis: Thanks. Those key/certificate/journal preservation assertions remain intact. The follow-up also tests that ambiguous managed-name lookalikes fail closed without changing CA material or NSS trust, while ordinary whitespace/newline nicknames no longer block the lifecycle.

crates/trusted-server-cli/src/commands/dev/proxy/trust.rs
ChristianPavilonis: Reproduced both the trailing-space install failure and the hidden same-subject conflict, then fixed them in 3e081339. A fixed column boundary cannot recover trailing spaces that are indistinguishable from padding, so the conflict preflight now queries the certificate's subject using `certutil -L -n <absolute CA certificate path> -a`. It validates the returned DER instead of exporting reconstructed foreign nicknames.

Presence checks only match the complete column for the known managed ASCII nickname and still require an exact named DER export. Failed queries/exports are not skipped. The real-NSS regressions prove repeat install, uninstall, rotation, unrelated-certificate preservation, and rejection of the hidden conflict before database or journal mutation. `cargo test_cli_linux --test proxy_trust_linux -- --include-ignored` passes all 18 tests.

One pre-existing limit is explicit in the guide and tests: a foreign nickname equal to the exact generated managed nickname plus trailing spaces remains visually ambiguous. That case still fails closed rather than treating a failed named lookup as absence.

crates/trusted-server-cli/src/commands/dev/proxy/trust.rs
ChristianPavilonis: Fixed in 3e081339 without trying to reassemble arbitrary nickname rows. The conflict check queries the CA's matching-subject set directly; presence checks no longer parse or validate every foreign row fragment.

`real_nss_newline_nickname_does_not_block_lifecycle` failed before the fix and now passes repeat install, uninstall, and rotation while preserving the foreign certificate. The same-subject conflict test also covers a newline nickname and verifies rejection before mutation. Deliberately imitated managed rows remain a documented fail-closed ambiguity, covered separately by `real_nss_ambiguous_managed_nickname_preserves_material_and_trust`.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
ChristianPavilonis: Applied in 3e081339. The Safari variant, dispatch arm, launch function, and macOS system-command helpers are now macOS-only. Pure quoting/parsing helpers remain available to unit tests, but are excluded from Linux production builds. Linux browser-selection tests and CLI Clippy pass; macOS verification is left to the new CI run.

crates/trusted-server-cli/src/commands/dev/proxy/browser.rs
ChristianPavilonis: Applied in 3e081339. Chrome and Firefox now share `lookup_on_path`, including canonicalization before the packaging-path check. Added the missing `chrome_command` documentation as well. Existing native-discovery and launch-argument tests pass.

crates/trusted-server-cli/src/commands/dev/proxy/trust.rs
ChristianPavilonis: Added the explanation above this call in 3e081339: it is an identity check, and an identical existing certificate is re-imported safely. Repeat-install behavior remains covered by the real-NSS tests.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
ChristianPavilonis: Changed the shared message in 3e081339 to `use chrome|firefox|all, plus safari on macOS`, which lists Safari as an available macOS choice rather than phrasing it as a restriction. Safari itself and its launch path are now also compiled only on macOS.

crates/trusted-server-cli/src/commands/dev/proxy/browser.rs
ChristianPavilonis: Agreed that this is a practical Ubuntu limitation, not just a theoretical wrapper case. In 3e081339 the guide now explicitly names Ubuntu's transitional Firefox/Chromium packages and `/usr/bin/firefox` / `/usr/bin/chromium-browser`, explains why canonicalization cannot detect their Snap redirection, and requires a native installation.

Keeping this PR's approved native-only scope. Wrapper classification and Snap/Flatpak automation remain deferred; this change does not claim to detect those wrappers or make their profile/trust configuration work. Known packaging paths are still skipped. Resolving this thread as a documented scope limitation, not as a claim that wrapper detection has been implemented.

crates/trusted-server-cli/tests/proxy_trust_linux.rs
```diff
@@ -0,0 +1,797 @@
...
+        assert!(
+            fs::read(fixture.db().join("cert9.db")).expect("should read unchanged NSS database")
+                == before
+        );
+    }
+}
+
+#[test]
+#[ignore = "requires certutil; all stores and HOME are disposable"]
+fn real_nss_filename_nickname_collision_fails_closed() {
```
aram356: 🔧 **wrench** — this is the strongest test in the PR, and it never runs.

All 11 `real_nss_*` tests here are `#[ignore]`d, and the `test-cli` job runs plain `cargo test` with no `certutil` installed and no `--include-ignored`. The `--include-ignored` occurrences in `integration-tests.yml` belong to a different crate (`trusted-server-integration-tests`), so they do not reach this file.

**This is not hypothetical.** A run on real Linux NSS reverted `trust.rs` to its pre-fix state while keeping these tests:

```text
test result: FAILED. 12 passed; 6 failed; finished in 1.05s
```

Five of those six failures are `#[ignore]`d tests. The only CI-visible failure is an incidental `!fixture.journal().exists()` assertion that does not name the bug. Under the reverted code, *this* test fails with:

```text
status: ExitStatus(0), stdout: "CA trusted for native Chrome/Chromium in .../pki/nssdb"
```

That is a **silent successful install** — a trust-store mutation involving an unrelated certificate — where the fixed code fails closed. That regression ships green under today's CI.

Separately, I classified every test in this file: **all 7 that CI does run drive a fake `certutil` shell script** via a `PATH` override. Those correctly gate the CLI's own control flow, but they cannot gate the assumption the whole redesign rests on, namely what real `certutil -L -n <file> -a` actually returns. I verified that assumption is correct and genuinely non-obvious (a stored nickname matching the filename wins over the file; a clean database temp-loads the file and returns one certificate; a stored same-subject certificate makes it two). None of that is mockable.

Closing the gap is cheap and measured, not estimated: `libnss3-tools` installs in about 7 seconds on `ubuntu-latest`, and all 18 tests pass in 1.31 seconds. The fixtures are strictly hermetic (every one redirects `HOME`, `XDG_DATA_HOME`, `XDG_CONFIG_HOME` and `XDG_CACHE_HOME` into its own tempdir, and the cross-fixture overrides point into another fixture's tempdir, never a global path), and no flakiness appeared at default parallelism.

**Proposed fix** (apply manually — the `cargo test` step is at lines 233-235 of `.github/workflows/test.yml`, outside this PR's diff hunks, so it cannot be a one-click suggestion):

```yaml
      - name: Install NSS tools for real-NSS trust tests
        if: runner.os == 'Linux'
        run: sudo apt-get update && sudo apt-get install -y libnss3-tools

      - name: cargo test
        run: |
          cargo test --manifest-path crates/trusted-server-cli/Cargo.toml \
            --target "$(rustc -vV | sed -n 's/host: //p')" -- --include-ignored
```

These tests are `#![cfg(target_os = "linux")]`, so the `macos-latest` leg compiles an empty file and is unaffected. `real_nss_preflight_preserves_leaf_intermediate_and_multicert_exports` also needs `openssl`, which is already present on `ubuntu-latest`. If you would rather keep the ignore attributes meaningful for local macOS runs, gate `--include-ignored` behind an env var and set it only on the Linux leg.

.github/workflows/test.yml
```diff
@@ -226,7 +225,7 @@ jobs:
       # No separate `cargo fmt` here: `trusted-server-cli` is a workspace member,
       # so the main `cargo fmt --all` job already formats it. Clippy still needs a
       # native run — the workspace clippy jobs are wasm/adapter-scoped and never
-      # lint the CLI's host-only code (macOS dev proxy vs Linux audit differ by cfg).
+      # lint the CLI's host-only code, including platform-specific trust operations.
```
aram356: 🔧 **wrench** — no CI ran against this revision.

The previous head (`097d2aa29`) ran five workflows: Run Tests, Run Format, Integration Tests, CodeQL Advanced, and Code Quality. This head (`3e081339f`) has exactly one workflow run recorded, Code Quality. No tests, no clippy, no fmt, no CodeQL, on either operating system.

The triggers in this file are correct (`pull_request`, and `synchronize` for the integration workflow), so this reads as a GitHub dispatch miss rather than a configuration bug, and nothing in the diff needs changing to fix it.

It matters more than a normal missed run would, because of what this revision changed. The entire rewritten trust implementation lives in a `#[cfg(target_os = "linux")]` module spanning roughly 373 lines of `trust.rs`. A macOS build never compiles it. I ran clippy, `fmt --check` and the full CLI suite locally on macOS and all are clean, but none of that touched the rewritten code, and I have no Linux cross-linker available to close the gap here. As things stand, **no build anywhere has type-checked the change this PR exists to make**.

One practical note for re-running: `test.yml` and `format.yml` declare no `workflow_dispatch`, so they cannot be re-triggered from the Actions UI. An empty commit or a close/reopen of the PR is needed to fire `synchronize` again.

Worth pairing with the CI change proposed on `tests/proxy_trust_linux.rs:702` — if the workflows are going to be re-triggered anyway, that is the moment to also make the real-NSS tests run.

crates/trusted-server-cli/tests/proxy_trust_linux.rs
```diff
@@ -0,0 +1,797 @@
...
+        "{out:?}"
+    );
+    assert!(!second.journal().exists());
+    assert_eq!(first.export_certificate(nickname), unrelated);
+    assert_eq!(first.export_certificate(conflicting_nickname), conflicting);
+}
+
+#[test]
+#[ignore = "requires certutil; all stores and HOME are disposable"]
+fn real_nss_ambiguous_managed_nickname_preserves_material_and_trust() {
```
aram356: ♻️ **refactor** — this test passes even with the fix reverted, so it does not gate the behavior its name describes.

In the mutation run (base `trust.rs`, new tests kept) this was among the 12 that passed. It asserts only that every operation fails and that `cert9.db` is unchanged, and the pre-fix code also failed everything here — but for an unrelated reason (`invalid NSS trust attributes`, raised by the since-deleted `nicknames()` parser) rather than the intended one (a padded row matches, then the exact `-L -n` lookup fails and refuses to authorize deletion).

I confirmed the intended mechanism against real NSS: a trailing-space alias renders identically to the managed row in the listing, and the exact named lookup then fails with `PR_FILE_NOT_FOUND_ERROR`. That is the behavior worth pinning.

The fix is to assert on stderr the way the sibling tests already do, so the test fails if the guard is ever replaced by some unrelated hard failure:

```rust
// Pin the intended failure rather than accepting any error.
let out = fixture.run(action);
assert!(!out.status.success(), "{foreign:?}; {action}: {out:?}");
assert!(
    String::from_utf8_lossy(&out.stderr).contains("managed certificate identity conflict"),
    "{foreign:?}; {action}: {out:?}"
);
```

Apply manually — the exact expected string depends on which guard you intend to pin, so it is worth choosing deliberately rather than accepting a generated suggestion.

Related but fine as-is: `real_nss_managed_prefix_in_foreign_nickname_does_not_block_lifecycle` also passes under revert, because `"<nickname> foreign-alias"` was never ambiguous under either implementation. It is a reasonable boundary test; it just should not be counted as gating this fix.

crates/trusted-server-cli/tests/proxy_trust_linux.rs
```diff
@@ -0,0 +1,797 @@
...
+
+    fn export_certificate(&self, nickname: &str) -> Vec<u8> {
+        let out = Command::new("certutil")
+            .args(["-L", "-n", nickname, "-r", "-d"])
+            .arg(format!("sql:{}", self.db().display()))
+            .output()
+            .expect("should export fixture certificate");
+        assert!(out.status.success(), "{out:?}");
+        out.stdout
+    }
```
aram356: ♻️ **refactor** — `certutil -L -n … -r` returns raw DER only, which carries no trust information, so this helper cannot detect a trust-flag change.

Verified directly:

```text
$ certutil -L -d sql:$D | grep FlagTest
FlagTest                                                     C,,
$ certutil -M -n FlagTest -t ",," -d sql:$D
$ certutil -L -d sql:$D | grep FlagTest
FlagTest                                                     ,,
DER sha before = d0e757eced16e27025160d0178fefdc17a2cddb2f74b59413e999edab574d96b
DER sha after  = d0e757eced16e27025160d0178fefdc17a2cddb2f74b59413e999edab574d96b
```

The DER is byte-identical across a `C,,` → `,,` transition.

The consequence is in `assert_foreign_nickname_does_not_block_lifecycle`, where the repeated `assert_eq!(fixture.export_certificate(nickname), unrelated)` at lines 585 and 597 proves the foreign **certificate material** survives the lifecycle but not that its **trust flags** do. For the two tests named `..._does_not_block_lifecycle` that claim is arguably fine. It is `real_nss_ambiguous_managed_nickname_preserves_material_and_trust` that puts "and trust" in its name — that one happens to be covered anyway, because it compares the whole `cert9.db` byte-for-byte.

Either assert the trust column explicitly, or drop the trust claim from the name:

```rust
// Trust flags live in the listing, not in the -r DER export.
fn export_trust(&self, nickname: &str) -> String {
    let out = Command::new("certutil")
        .args(["-L", "-d"])
        .arg(format!("sql:{}", self.db().display()))
        .output()
        .expect("should list fixture trust attributes");
    String::from_utf8_lossy(&out.stdout)
        .lines()
        .find(|line| line.starts_with(nickname))
        .unwrap_or_default()
        .to_owned()
}
```

Apply manually — this adds a helper and new call sites, so it spans more than one contiguous range.

crates/trusted-server-cli/tests/proxy_trust_linux.rs
```diff
@@ -0,0 +1,797 @@
...
+    assert_eq!(fixture.export_certificate(nickname), unrelated);
+    assert_eq!(
+        fs::read_to_string(fixture.journal()).expect("should read cleared journal"),
+        "[]"
+    );
+}
+
+#[test]
+#[ignore = "requires certutil; all stores and HOME are disposable"]
+fn real_nss_trailing_space_nickname_does_not_block_lifecycle() {
```
aram356: 🌱 **seedling** — two documentation gaps worth closing, neither blocking.

**The opt-in real-NSS tests are undocumented.** Nothing in `docs/`, `CONTRIBUTING.md` or `AGENTS.md` mentions that these 11 tests exist, that they need `certutil`, or how to run them:

```bash
cargo test_cli_linux --test proxy_trust_linux -- --include-ignored
```

The PR description cites that exact command, and the Python browser proof is documented in the guide, which makes this omission look like an oversight rather than a decision. If the tests stay `#[ignore]`d, documenting them is essential; if they get enabled in CI per the comment on line 702, a one-line note still helps local contributors on macOS, where the file compiles to nothing.

**The filename-nickname collision failure mode is undocumented.** It has a bespoke user-facing error (`NSS subject query returned an unrelated certificate`), it is reproducible, and an operator who hits it has no guidance. The guide's paragraph on padded-listing ambiguity covers the trailing-space case but not this one. A sentence in the same section explaining that a manual import whose nickname equals the CA file's path will be rejected, and that removing or renaming that entry is the remedy, would close it.

aram356 (changes requested):
## Summary

Second pass, against `3e081339f`. **All six findings from the previous review are fixed**, and the two blocking ones were fixed at the root cause rather than patched: nickname reconstruction is gone entirely. I re-verified each fix independently against a real `certutil` rather than relying on the replies.

The headline evidence for the fix working: I built an NSS database containing **both** hostile nicknames (`"Trailing Space CA "` and `"Line One\nLine Two"`) and ran the new subject query against it. Exit 0, exactly one certificate returned. The same database made `install`, `uninstall`, **and** `regenerate` fail under the previous head.

The rewrite opened a new attack surface, and it is already closed. Passing a filename to `-n` is hijackable: I imported an unrelated certificate under a nickname equal to the CA's own path, and NSS returned **the nickname entry, not the file**. `check_subject_conflicts` checks the subject mismatch first and aborts before any mutation, and `real_nss_filename_nickname_collision_fails_closed` pins exactly that.

No new correctness defects were found in the rewrite. Everything blocking this round is about **verification**, not code, and it is not a process complaint: a separate run on real Linux NSS reverted `trust.rs` to its pre-fix state while keeping the new tests, and **6 tests failed, 5 of them `#[ignore]`d**. The regression this PR exists to prevent currently ships green.

## Blocking

### 🔧 wrench

- **The regression tests for this PR's fix never run anywhere** — see inline at `crates/trusted-server-cli/tests/proxy_trust_linux.rs:702`
- **CI never validated this revision** — see inline at `.github/workflows/test.yml:228`

## Non-blocking

### ♻️ refactor

- **`real_nss_ambiguous_managed_nickname_preserves_material_and_trust` is not a regression test** — see inline at `crates/trusted-server-cli/tests/proxy_trust_linux.rs:674`
- **`export_certificate` cannot observe trust-flag changes** — see inline at `crates/trusted-server-cli/tests/proxy_trust_linux.rs:53-61`

### 🌱 seedling

- **Document the opt-in real-NSS tests and the filename-collision error** — see inline at `crates/trusted-server-cli/tests/proxy_trust_linux.rs:606`

- 🌱 **Pre-existing, self-resolving edge case in the post-delete path.** If a foreign alias `"<managed-nickname> "` exists alongside a genuine managed entry, I reproduced this full sequence against real NSS: `certutil -D` correctly deletes only the exact nickname and preserves the alias; the post-delete listing still contains a row that `contains()` matches; the exact named export then fails with `PR_FILE_NOT_FOUND_ERROR` (exit 255); and `uninstall` therefore errors **before** `entries.pop()`, leaving a journal entry for a certificate that is already gone. This is fail-closed, it resolves as soon as the foreign alias is removed, and the deleted `nicknames()` parser behaved identically, so the rewrite neither introduced nor worsened it. Recording it so it is not mistaken for a new defect later; no change requested.

- 📝 **Verified as correct, recorded so they are not re-litigated.** Checked against real `certutil` and found sound: `-n <path>` on an empty or freshly created database still returns the supplied certificate, so `found_expected` does not spuriously fail a first install; the `-a` export contains only PEM across every database shape tried, and degenerate `cert_path` inputs exit non-zero before `rustls_pemfile` sees anything; the new raw-byte row matcher in `contains()` is *stricter* than the deleted parser (it rejects a vertical-tab variant the old one accepted) and cannot produce a false negative on the managed nickname, which contains no whitespace; deletion still requires an exact DER match on both the install and uninstall paths, so a forged listing row cannot authorize a delete; `fs::canonicalize(cert_path)` cannot fail on a fresh machine because `mod.rs` runs `load_or_generate` before `trust::install`; and the `lock_file` change correctly separates `TryLockError::WouldBlock` (`Busy`) from other I/O errors (`Io`), which I confirmed against the real API including release-on-drop.

- 📝 **Local verification performed on the PR head.** On macOS: `cargo clippy --package trusted-server-cli --target aarch64-apple-darwin --all-targets -- -D warnings` clean, `cargo fmt --all -- --check` clean, and the CLI suite passes (29 unit, 7 perf, 1 keychain). This does **not** cover the change under review: the entire rewritten implementation lives in a `#[cfg(target_os = "linux")]` module of roughly 373 lines, which a macOS build never compiles.

## CI Status



- Code Quality: PASS
- Run Tests: **not run** (ran on the previous head `097d2aa29`)
- Run Format: **not run** (ran on the previous head)
- Integration Tests: **not run** (ran on the previous head)
- CodeQL Advanced: **not run** (ran on the previous head)

Workflow triggers are configured correctly (`pull_request` / `synchronize`), so this looks like a dispatch miss rather than a configuration problem. Note that `test.yml` and `format.yml` declare no `workflow_dispatch`, so they cannot be re-run from the Actions UI; an empty commit or a close/reopen is needed to re-trigger them.

PR #1173, opened 2026-09-16
Reject placeholder partner pull tokens on startup

### LINKED ISSUE #1144: Reject partner API and pull-token template placeholders
`Settings::reject_placeholder_secrets` fails startup when a secret still holds a known template value. It covers `ec.passphrase`, `publisher.proxy_secret`, handler passwords, and `ec.partners[].api_token` — but not `ec.partners[].ts_pull_token`, and its `api_token` list has drifted away from the values the template actually ships.

## What's wrong

Two separate gaps, both in `crates/trusted-server-core/src/settings.rs`:

1. `reject_placeholder_secrets` (~line 3084) loops over `ec.partners` and checks only `api_token`. `ts_pull_token` is never checked at all. Its only validation is a secret-key-reference shape check in `config.rs:402` (store mode only) and a present/non-empty check in `ec/registry.rs:336` (only when `pull_sync_enabled`). An unmodified template value therefore passes startup and is sent as the outbound bearer token in `ec/pull_sync.rs:201`.

2. `EcPartner::API_TOKEN_PLACEHOLDERS` (~line 447) lists `partner-api-token-32-bytes-minimum` and `replace-with-partner-api-token-32-bytes-minimum`. Neither appears in `trusted-server.example.toml`, which ships `partner_api_token` (line 99) and `ts_pull_token = "partner_ts_pull_token"` (line 101). So the check that does exist misses the literal an operator is most likely to leave behind.

**Note on an earlier draft:** this issue previously named the string `replace-with-partner-ts-pull-token`. That value does not exist anywhere in the repo — don't go looking for it. The real values are the two `partner_*` literals above.

## Acceptance criteria

- `reject_placeholder_secrets` rejects `ec.partners[].ts_pull_token` when it holds a template placeholder, reporting the field as `ec.partners[<source_domain>].ts_pull_token` to match the existing `api_token` branch.
- The placeholder lists cover the literals `trusted-server.example.toml` actually ships (`partner_api_token`, `partner_ts_pull_token`), keeping the existing entries.
- Regression tests: one asserting a placeholder pull token fails, one asserting a realistic non-placeholder token still passes. Follow the shape of `reject_placeholder_secrets_includes_handler_passwords` (~line 5391).

## Guardrail

`reject_placeholder_secrets` runs at startup via `config.rs:271`, so an over-broad list hard-fails live deployments. Only add literals that genuinely appear in the template — nothing that merely looks placeholder-ish. `partner-api-token-32-bytes-minimum` is unused by the template but harmless to keep; removing entries is out of scope.

## Pointers

- `crates/trusted-server-core/src/settings.rs:447` — `API_TOKEN_PLACEHOLDERS`, `is_placeholder_api_token`
- `crates/trusted-server-core/src/settings.rs:3084` — the partner loop to extend
- `crates/trusted-server-core/src/settings.rs:5391` — test pattern to copy
- `trusted-server.example.toml:99-101` — the template values in question
- `crates/trusted-server-core/src/ec/pull_sync.rs:201` — where the token is consumed

Scope is one constant, one `if let` block, and two tests. `reject_placeholder_secrets` has a single production caller, and no existing test fixture uses the affected literals.

dhruv8sh:
## Summary

- `reject_placeholder_secrets` checked `ec.partners[].api_token` for known placeholder values but never checked `ec.partners[].ts_pull_token`, so an unmodified template value for that field passed startup and was sent as the outbound pull-sync bearer token.
- The placeholder list also didn't include `partner_api_token` / `partner_ts_pull_token`, the literal values `trusted-server.example.toml` actually ships — so even the field that was checked would have missed the value an operator is most likely to leave behind.

## Changes

| File | Change |
| ---- | ---- |
| `crates/trusted-server-core/src/settings.rs` | Added `partner_api_token` / `partner_ts_pull_token` to `EcPartner::API_TOKEN_PLACEHOLDERS`; added a `ts_pull_token` branch to the partner loop in `reject_placeholder_secrets`, reporting it as `ec.partners[<source_domain>].ts_pull_token`; added 3 tests: placeholder pull token rejected, realistic pull token accepted, and a regex-based check that every `api_token`/`ts_pull_token` example value in `trusted-server.example.toml` is a recognized placeholder |

## Closes

Closes #1144 

## Test plan

- [ ] Other:

## Checklist

- [x] New code has tests


crates/trusted-server-core/src/settings.rs
```diff
@@ -5406,6 +5418,85 @@ source_domain = "partner.example.com"
...
+            assert!(
+                EcPartner::is_placeholder_api_token(value),
+                "example config partner secret '{value}' should be a recognized placeholder \
+                 — add it to EcPartner::API_TOKEN_PLACEHOLDERS"
+            );
+        }
+        assert!(
+            checked_any,
+            "should have found at least one partner secret example value in the template"
+        );
```
prk-Jr, line 5497: ♻️ **refactor (non-blocking)** — the drift guard passes when a partner secret key is renamed or added.

`checked_any` is a single global flag across the whole `api_token|ts_pull_token` alternation. If `ts_pull_token` is renamed in the template, or a *new* partner secret key is introduced there, the loop still matches the `api_token` line, sets `checked_any = true`, and the test passes green — while the new key's example value is never checked against `API_TOKEN_PLACEHOLDERS`. That is exactly the class of bug this PR fixes (#1144 was a partner secret field the placeholder check never saw), so the guard currently has a hole shaped like the bug it guards.

Verified rather than asserted: renaming `# ts_pull_token = "partner_ts_pull_token"` to `# pull_sync_bearer = "partner_pull_bearer"` in the template leaves this test **passing** on the PR as submitted. With the suggestion below applied, the same mutation fails with a precise message:

```
trusted-server.example.toml has no example value for partner secret keys ["ts_pull_token"] — the template key was renamed or removed
```

The per-key `format!` supersedes the `EXAMPLE_PARTNER_KEYS.join("|")` alternation, so `pattern` moves inside the loop.

```suggestion
        let mut unchecked_keys: Vec<&str> = Vec::new();
        for &key in EXAMPLE_PARTNER_KEYS {
            let pattern = format!(r#"(?m)^\s*#?\s*{key}\s*=\s*"([^"]+)""#);
            let example_value =
                Regex::new(&pattern).expect("should compile example partner secret regex");

            let mut checked_any = false;
            for capture in example_value.captures_iter(EXAMPLE_TOML) {
                checked_any = true;
                let value = &capture[1];
                assert!(
                    EcPartner::is_placeholder_api_token(value),
                    "example config partner secret '{value}' should be a recognized placeholder \
                     — add it to EcPartner::API_TOKEN_PLACEHOLDERS"
                );
            }
            if !checked_any {
                unchecked_keys.push(key);
            }
        }
        assert!(
            unchecked_keys.is_empty(),
            "trusted-server.example.toml has no example value for partner secret keys \
             {unchecked_keys:?} — the template key was renamed or removed"
        );
```

<sub>Suggestion verified in a scratch worktree: `cargo fmt --all -- --check` clean, `cargo clippy-fastly` and `cargo clippy-axum` clean, `cargo test -p trusted-server-core --target aarch64-apple-darwin --lib` 2670 passed / 0 failed, `cargo test-axum` all suites ok. Not run against it: `test-fastly` / `test-cloudflare` / `test-spin` / parity — the change is confined to one `#[cfg(test)]` fn in core with no cfg or feature interaction.</sub>

prk-Jr (approved):
## Review: approve

Correct, minimal, and on the right layer. `API_TOKEN_PLACEHOLDERS` gains the two literals the template actually ships, and the new `ts_pull_token` branch in `reject_placeholder_secrets` is symmetric with the `api_token` branch that already shipped. One non-blocking suggestion inline; nothing here should hold the merge.

### Verification

Mutation-tested both halves of the fix in a scratch worktree at `a9715d55` — neither new test is vacuous:

| Mutation | Result |
|---|---|
| Baseline (unmodified head) | `2670 passed; 0 failed` |
| Remove the two new `API_TOKEN_PLACEHOLDERS` entries | `example_toml_...` **and** `..._includes_partner_pull_tokens` FAIL |
| Remove the `ts_pull_token` branch, keep the placeholders | `..._includes_partner_pull_tokens` FAILS |

Remote CI is 20/20 green, including all four required checks (`cargo fmt`, `cargo test`, `format-typescript`, `format-docs`) plus parity, spin, cloudflare, integration, browser, vitest and CodeQL.

### Notes (no action required)

**📝 The doc comment on `API_TOKEN_PLACEHOLDERS` no longer matches its contents.** It reads *"Known partner API token placeholders"*, and the predicate is `is_placeholder_api_token`, but the list now also holds `partner_ts_pull_token` and is consulted for `ts_pull_token`. Sharing one list across both fields is the right call; only the prose under-describes it. Renaming the `pub const` / `pub fn` would be a breaking API change and isn't worth it. Suggested wording, if you touch the file again:

```rust
    /// Known partner secret placeholders (`api_token` and `ts_pull_token`) that
    /// must not be used in deployments.
    pub const API_TOKEN_PLACEHOLDERS: &[&str] = &[
```

**🤔 The description's threat model is a little wider than the code path.** Tracing it: `config_payload.rs:59` is the only production caller of `validate_settings_for_runtime` → `reject_placeholder_secrets`, and it runs *after* `resolve_secret_references`. `secret_resolution.rs:169-181` (`resolve_leaf`) hard-errors on a missing key and never falls back to the config literal, so the runtime `ts_pull_token` is always a resolved secret-store **value**, never the template string. The docs (`docs/guide/ec-setup-guide.md:34,63,81`, `docs/guide/configuration.md:607`) present `partner_api_token` / `partner_ts_pull_token` as intended secret-store *key names*.

So the new check fires only when an operator provisioned the secret-store *value* as the literal template string — still worth rejecting as defense in depth, and I confirmed it **cannot** false-positive on the documented key-reference setup, since `reject_placeholder_secrets` never sees unresolved key names. Flagging only so the coverage isn't over-credited.

Related, also checked: the new branch validates `ts_pull_token` unconditionally, ignoring `pull_sync_enabled`. No false positive — `config_payload.rs:76-88` (`remove_inactive_secret_references`) strips `ts_pull_token` from the blob whenever `pull_sync_enabled` is not true, before validation runs.

**🌱 The drift guard covers the TOML template only, not the docs.** `docs/guide/ec-setup-guide.md:34,81` and `docs/guide/configuration.md:606-607` ship the same literals and aren't scanned. They agree with the template today, so nothing to fix; a follow-up could widen `EXAMPLE_TOML` into a slice of `include_str!`'d sources.

### 👍

The new tests mirror the existing `reject_placeholder_secrets_includes_handler_passwords` shape exactly, use `expect` / `expect_err` with `"should ..."` messages, and assert on the field *name* rather than the token value, so no secret can leak into a failure message. `example_toml_partner_secret_examples_are_recognized_placeholders` is the right instinct — a self-maintaining guard beats a hand-copied constant.

### What I did not verify

Runtime behaviour end to end. The claim that `reject_placeholder_secrets` gates the live pull-sync path comes from static call-graph tracing, not from executing a request with a placeholder token in a secret store.

crates/trusted-server-core/src/settings.rs
dhruv8sh: Good catch, but this hardens the bonus drift-guard test against future field renames — #1144 scoped the fix to `api_token`/`ts_pull_token` only. Filing as a follow-up rather than folding into this PR.

ChristianPavilonis (approved):
## Review summary

Approved. Reviewed `a9715d551ea69f4e37269b0bdcd494d52da9d4fb` against `6cae7f5da8911c746cf873581885f90c3820dd96`. The change correctly rejects resolved partner pull-token template values before they can reach outbound pull sync, while retaining realistic tokens. No actionable issues found.

Validation passed: focused Fastly-target tests for placeholder rejection, realistic-token acceptance, template drift coverage, runtime placeholder validation, and config-blob secret resolution; `cargo clippy-fastly`; `cargo fmt --all -- --check`; and `git diff --check`. All 20 reported CI checks are green.

prk-Jr (approved):

crates/trusted-server-core/src/settings.rs
```diff
@@ -450,6 +450,8 @@ impl EcPartner {
         "sharedid-internal-token-32-bytes",
         "inttest-api-key-1-32-bytes-minimum",
         "inttest2-api-key-2-32-bytes-minimum",
+        "partner_api_token",
+        "partner_ts_pull_token",
```
aram356: ❓ **question** — These two entries are secret-store **key names**, but `API_TOKEN_PLACEHOLDERS` is compared against **resolved secret values**. Is there a concrete case where a key name reaches a live `Settings`? If so this is a resolver bug and my read is wrong; if not, these two lines are aimed at the wrong layer.

**Why the layers differ.** Secret handling is two-stage. `settings_from_config_blob` (`config_payload.rs:52-59`) resolves first, then validates:

```
L51  remove_inactive_secret_references(&mut data)
L53  resolve_secret_references::<TrustedServerAppConfig>(...)   // key name -> value
L58  Settings::from_json_value(data)
L59  validate_settings_for_runtime(&settings)                   // reject_placeholder_secrets
```

`resolve_leaf` (`secret_resolution.rs:183`) does `object.insert(key.to_owned(), Value::String(resolved))` — the key name is replaced by the stored value. `reject_placeholder_secrets` has exactly one production caller (`config.rs:271`), six lines after resolution, and all four adapters funnel through it. The deploy path (`ts config push` / `validate`) uses `validate_settings_for_deploy`, which never calls this function; its doc comment states the split outright: *"Secret fields contain secret-store key names at this stage, so this function deliberately excludes checks that require resolved values."*

**The convention this departs from.** No sibling list contains its template key name:

| Template ships | In its placeholder list? |
| --- | --- |
| `password = "handler_password"` | No |
| `proxy_secret = "publisher_proxy_secret"` | No |
| `passphrase = "ec_passphrase"` | No |
| `shared_secret = "trusted_client_ip_shared_secret"` | No |
| `api_token = "partner_api_token"` | **this PR adds it** |

`config_payload.rs:817 placeholder_rejection_happens_after_secret_resolution` encodes the rule directly: key `"placeholder_proxy"` resolves to value `"change-me-proxy-secret"`, and it is the *value* that appears in `PROXY_SECRET_PLACEHOLDERS`. `docs/guide/ec-setup-guide.md:38-42` says the same for these exact fields: they "contain keys in the Trusted Server secret store, not the credential values," and instruct operators to provision a real secret under `partner_api_token`.

**Practical effect.** An operator whose stored secret *value* happens to equal the string `partner_api_token` now hard-fails startup. Narrow, but a real behavior change. The threat in the commit message — an operator who copied the template and never provisioned the key — is already caught: `resolve_leaf` errors when the key is missing from the store (`secret_resolution.rs:169-176`) and rejects an empty resolved value (`:177-181`).

**Suggested resolution.** Drop these two lines and retarget `reject_placeholder_secrets_includes_partner_pull_tokens` to a pre-existing placeholder value, e.g. `partner-api-token-32-bytes-minimum`. I verified that test then passes with the entries removed, which keeps the real `ts_pull_token` fix and gives it a test that does not depend on the key-name entries. The drift-guard test would need to be dropped or re-aimed at the deploy path (`validate_secret_key_references`, `config.rs:388`), since as written it asserts a rule the other four secret fields deliberately do not follow.

Not expressed as a `suggestion` block — resolving it also touches the two tests below in the same file, which is more than one contiguous hunk.

crates/trusted-server-core/src/settings.rs
```diff
@@ -5406,6 +5418,85 @@ source_domain = "partner.example.com"
...
+            Regex::new(&pattern).expect("should compile example partner secret regex");
+
+        let mut checked_any = false;
+        for capture in example_value.captures_iter(EXAMPLE_TOML) {
+            checked_any = true;
+            let value = &capture[1];
+            assert!(
+                EcPartner::is_placeholder_api_token(value),
+                "example config partner secret '{value}' should be a recognized placeholder \
+                 — add it to EcPartner::API_TOKEN_PLACEHOLDERS"
```
aram356: ⛏ **nitpick** — Em dash (U+2014, verified as bytes `e2 80 94`) in the assert message. AGENTS.md calls for a hyphen, comma, colon, or a restructure instead.

```suggestion
                "example config partner secret '{value}' should be a recognized \
                 placeholder; add it to EcPartner::API_TOKEN_PLACEHOLDERS"
```

Scratch-verified at the PR head: `cargo fmt --all -- --check` passes, `cargo clippy-axum` is clean, the target test passes, and a byte-exact before/after patch comparison showed no drift.

crates/trusted-server-core/src/settings.rs
```diff
@@ -5406,6 +5418,85 @@ source_domain = "partner.example.com"
...
+        )];
+
+        settings
+            .reject_placeholder_secrets()
+            .expect("should accept a realistic partner pull token");
+    }
+
+    /// Guards against the placeholder lists drifting away from the example config
+    /// Example value shipped in `trusted-server.example.toml` must be recognized
+    /// as a placeholder
```
aram356: ⛏ **nitpick** — Two independent sentences run together with no period after `config`, and `Example value` is singular while the test iterates every match (there are two). The project doc convention is a single-line summary, a blank `///` line, then details.

```suggestion
    /// Guards against the placeholder list drifting away from the example config.
    ///
    /// Every partner secret value shipped in `trusted-server.example.toml` must
    /// be recognized by [`EcPartner::is_placeholder_api_token`].
```

Scratch-verified at the PR head alongside the em-dash fix: `cargo fmt --all -- --check` passes, `cargo clippy-axum` is clean, the target test passes, and the byte-exact drift check showed no deviation. Note this suggestion replaces 3 lines with 4.

aram356 (changes requested):
## Summary

The `ts_pull_token` branch added to `reject_placeholder_secrets` is a genuine gap-fill: `api_token` was checked and `ts_pull_token` was not, and both are declared secret leaves in `secret_fields()`. That part is correct and should land.

The two new `API_TOKEN_PLACEHOLDERS` entries are a different matter. `partner_api_token` / `partner_ts_pull_token` are secret-store **key names**, but the list is compared against **resolved secret values**. Details and the evidence are in the inline comment.

> 2 of the inline comments below carry a one-click GitHub `suggestion` — use **Commit suggestion** (or **Add suggestion to batch**) to apply them as commits on the PR branch. The remaining comment describes the concern in prose because it is a question about the approach rather than a mechanical edit.

### How this was verified

Rather than reason from reading alone, I probed the real blob path (`settings_from_config_blob`) end to end:

| Case | Result |
| --- | --- |
| Template key name + secret provisioned under it (the normal operator path) | Resolves to the real secret value. The new entries never fire. |
| Secret store returns the key name *as the stored value* | `Insecure default … ts_pull_token` — the new entries fire |

I also removed only the two new list entries and re-ran the suite:

- `reject_placeholder_secrets_allows_realistic_partner_pull_token` — still passes
- `reject_placeholder_secrets_includes_partner_pull_tokens` — **fails**
- `example_toml_partner_secret_examples_are_recognized_placeholders` — **fails**

Both failing tests depend entirely on the new entries. Retargeting the first to the pre-existing placeholder `partner-api-token-32-bytes-minimum` makes it pass *without* them, which shows the real fix stands on its own and currently has no test independent of the key-name entries.

## Blocking

### ❓ question

- **New placeholder entries are secret-store key names, not resolved values** — see inline at `crates/trusted-server-core/src/settings.rs:453`

## Non-blocking

### ⛏ nitpick

- **Em dash in assert message** — see inline at `crates/trusted-server-core/src/settings.rs:5491`
- **Doc comment grammar and format** — see inline at `crates/trusted-server-core/src/settings.rs:5470`

- 📝 **The regex itself is sound.** Tested empirically against the template: `^\s*#?\s*(?:api_token|ts_pull_token)\s*=` matches exactly the two intended lines. Because `\s*` matches whitespace only, `# partner_api_token = "x"`, `tinybird_api_token = "x"` and `x_ts_pull_token = "z"` all correctly fail to match, so the absent leading word boundary is harmless. The `checked_any` backstop is the right call. Minor blind spot: `## api_token = "x"` (two `#`) and single-quoted or multiline TOML values do not match, so a partial reformat of one key could silently drop that key's coverage while `checked_any` stays true.

- 📌 **Pre-existing: a padded `ts_pull_token` reaches the outbound `Bearer` header untrimmed.** `is_placeholder_api_token` trims for comparison, but `ec/pull_sync.rs:237` sends `format!("Bearer {}", token.expose())` with the raw value, so `" real-token "` passes validation and produces a malformed header. `RequestSigning::is_unusable_store_id` (`settings.rs:772`) handles exactly this shape with `store_id != store_id.trim()`, and the comment at `settings.rs:3115-3123` documents the reasoning. I confirmed this behavior is unchanged from `main`, so it is out of scope for this PR — worth a follow-up issue rather than a change here.

- 📝 **Two concerns considered and dismissed, recorded so they are not re-raised.** (1) The unconditional check does *not* break deployments with a disabled partner: `remove_inactive_secret_references` (`config_payload.rs:75-86`) strips `ts_pull_token` before resolution whenever `pull_sync_enabled` is falsy, so it is always `None` by the time `reject_placeholder_secrets` runs; `string_false_pull_sync_flag_removes_a_stale_token` and `inactive_optional_features_do_not_resolve_stale_secret_references` both pin this. (2) An unprovisioned secret key already fails startup at `secret_resolution.rs:169-181`, so the "operator never provisioned the key" case was never reachable as a live placeholder token.

- 🌱 **Optional test hardening.** No test asserts that a partner with *both* `api_token` and `ts_pull_token` set to placeholders reports both fields — `test_partner_with_pull_token` hardcodes `api_token: None`, so nothing pins that the loop accumulates rather than short-circuiting. Separately, `test_partner_with_pull_token` is roughly the twelfth full-struct `EcPartner { … }` literal in the crate; `config_payload.rs:217 partner_with_pull_sync` uses a lighter `serde_json::from_value` form that survives new fields being added. Both are non-blocking.

## CI Status

All checks passing at `a9715d55`. Locally at the PR head: `cargo fmt --all -- --check` passes, `cargo clippy-axum` is clean, and the full `trusted-server-core` lib suite passes (2670 tests).

- integration tests: PASS
- CodeQL: PASS
- vitest: PASS
- cargo fmt: PASS (required)
- cargo test: PASS (required)
- Analyze (rust): PASS
- Analyze (actions): PASS

crates/trusted-server-core/src/settings.rs
dhruv8sh: You're right — verified the pipeline order myself (config_payload.rs:52-59, secret_resolution.rs's resolve_leaf overwriting the key with the resolved value, and the deploy-path doc comment). Applied your suggested resolution: removed the two key-name entries from API_TOKEN_PLACEHOLDERS, retargeted reject_placeholder_secrets_includes_partner_pull_tokens to the pre-existing partner-api-token-32-bytes-minimum placeholder, and dropped example_toml_partner_secret_examples_are_recognized_placeholders since it asserted an invariant none of the sibling secret fields follow (confirmed: proxy_secret, passphrase, and shared_secret example values in trusted-server.example.toml are all key names too, none in their placeholder lists). Not re-aiming it at validate_secret_key_reference — that function only checks non-empty, so it wouldn't express a placeholder-drift guarantee.

crates/trusted-server-core/src/settings.rs
dhruv8sh: Moot — the whole test this assert lived in is dropped as part of resolving the blocking comment below (it asserted an invariant that doesn't hold once the two key-name placeholder entries are removed).

crates/trusted-server-core/src/settings.rs
dhruv8sh: Moot — this test is dropped as part of resolving the blocking comment below.

crates/trusted-server-core/src/settings.rs
dhruv8sh: Superseded — the test this hole was in (example_toml_partner_secret_examples_are_recognized_placeholders) has been removed while resolving aram356's blocking finding below: it asserted that example-config key names should appear in API_TOKEN_PLACEHOLDERS, which turned out to be the wrong invariant (those are secret-store key names, not resolved values). The drift-hardening suggestion no longer applies since there's no drift guard left to harden.

PR #1175, opened 2026-09-16
Align Fastly staging store selectors with EdgeZero PR 381

### LINKED ISSUE #1174: Align Fastly staging store selectors with EdgeZero PR 381
## Description

Fastly staging cannot build Trusted Server application state when the app config references secrets such as `publisher.proxy_secret`. The runtime falls back to the logical secret-store name `trusted_server_secrets` instead of opening the physical store selected by the staging GitHub Environment, such as `ts_secrets`.

The resulting startup error is:

```text
failed to resolve secret reference at `publisher.proxy_secret` from secret store `trusted_server_secrets`
```

EdgeZero `v0.0.8` uses service-scoped runtime selector keys. That model does not apply the canonical store selectors supplied by the selected deployment environment to each production or staging deployment. [EdgeZero PR 381](https://github.com/stackpop/edgezero/pull/381) replaces it: a managed deployment reads canonical `EDGEZERO__STORES__<KIND>__<ID>__NAME` selectors from the deployment environment and links each selected physical store to the target service version under its logical ID. The runtime opens stores by logical ID and reads the config entry under the logical ID for every target. The `edgezero_runtime_env` Config Store and the runtime selector lookup are removed.

Trusted Server should consume that upstream branch while PR 381 is under review and move its Fastly entry point, local Viceroy configuration, and documentation to the logical-ID model.

The expected staging selectors are:

```text
EDGEZERO__STORES__CONFIG__TRUSTED_SERVER_CONFIG__NAME=<physical config store>
EDGEZERO__STORES__KV__TRUSTED_SERVER_KV__NAME=<physical kv store>
EDGEZERO__STORES__SECRETS__TRUSTED_SERVER_SECRETS__NAME=ts_secrets
```

Every logical store declared in `edgezero.toml` needs a selector or a same-named physical store, because the managed deploy resolves all of them before publishing.

## Scope added during implementation

The `trusted-server-js` build script asked Cargo to rerun on every file under `lib/`, including `node_modules`, which emitted 33,932 rerun directives and invalidated the crate after any npm install. PR #1175 also narrows that to the bundle inputs.

## Acceptance criteria

- All EdgeZero workspace dependencies use `fix/fastly-environment-store-selectors` until PR 381 is released.
- The Fastly entry point opens `trusted_server_config` and `trusted_server_secrets` by logical ID and no longer reads `edgezero_runtime_env`.
- Root and integration Viceroy configurations expose the secret store under `trusted_server_secrets` and define no `edgezero_runtime_env` store.
- A regression test verifies both Viceroy configurations expose the store under its logical ID.
- Fastly and CLI documentation describe deploy-time environment selection, logical-ID resource links, the fixed logical config key, and the application release that managed deploys require.
- The `trusted-server-js` build script watches only `lib/src`, the package manifests, and the bundler configuration.
- Target-matched build, test, lint, JavaScript, and documentation checks pass.
- A follow-up replaces the branch dependency with the release tag containing PR 381.

aram356:
## Summary

- Restore Fastly staging secret resolution by consuming the store-selection model from [EdgeZero PR 381](https://github.com/stackpop/edgezero/pull/381): a managed deployment links each physical store selected by a canonical `EDGEZERO__STORES__<KIND>__<ID>__NAME` deployment selector to the target version under its logical ID, and the runtime opens stores by that ID.
- Open `trusted_server_config` and `trusted_server_secrets` by logical ID in the Fastly entry point and read the config entry under `trusted_server_config` for every target. The upstream branch removed `runtime_env_config` and the `edgezero_runtime_env` Config Store it read, and fixed the Fastly config key to the logical ID; staging isolation comes from the physical store the staging environment selects.
- Expose the local Viceroy secret store under `trusted_server_secrets` and drop the `edgezero_runtime_env` selector store, so local runs exercise the same store-opening path as a deployed version.
- Update operator guidance so production and staging select physical stores through their deployment environments and so `ts deploy --adapter fastly` passes a verified application release.
- Watch only bundle inputs in the `trusted-server-js` build script. It previously emitted a rerun directive for every file under `lib/`, including `node_modules`, so any npm install invalidated the crate and Cargo re-checked 33,932 paths on every build.

The EdgeZero dependencies temporarily follow `fix/fastly-environment-store-selectors` and currently lock to `5878eb740897749e4cfab7b410fa28791a371602`. Replace the branch dependency with the release tag after the upstream PR merges.

## Changes

| File | Change |
| ---- | ------ |
| `Cargo.toml`, `Cargo.lock` | Select and lock the EdgeZero PR 381 branch head for every EdgeZero crate. |
| Fastly adapter `main.rs`, `app.rs` | Replace `RuntimeStoreConfig::from_env` with `logical()`: logical store IDs and the logical config key for every target. A unit test covers the bindings. |
| `fastly.toml`, Viceroy integration fixture, template-cache script | Rename the local secret store to `trusted_server_secrets`; remove the `edgezero_runtime_env` mapping. |
| Integration config test | Verify both Viceroy configurations expose the secret store under its logical ID and define no legacy selector store. |
| Fastly and CLI guides | Document deploy-time environment selectors, logical-ID resource links, the fixed logical config key, and the `--application-release` requirement for managed deploys. |
| Design and implementation plan | Record the root cause, the logical-ID runtime model, validation, and the release-tag follow-up. |
| `crates/trusted-server-js/build.rs` | Replace the recursive `lib/` walk with explicit `rerun-if-changed` directives for `lib/src`, the package manifests, and the bundler configuration. No-op `cargo check -p trusted-server-js` drops from 1.1s to 0.15s and the script emits 7 directives instead of 33,932. |

## Closes

Closes #1174

## Test plan

- [x] `cargo clippy-cloudflare && cargo clippy-cloudflare-wasm`
- [x] `cargo clippy-spin-native && cargo clippy-spin-wasm`
- [x] `cargo check-fastly && cargo check-axum && cargo check-cloudflare && cargo check-spin`
- [x] `./scripts/test-cli.sh`
- [x] `cargo test --manifest-path crates/trusted-server-integration-tests/Cargo.toml --test integration local_fastly --target aarch64-apple-darwin`
- [x] `cargo test --manifest-path crates/trusted-server-integration-tests/Cargo.toml --test parity --target aarch64-apple-darwin`
- [x] JS build: `cd crates/trusted-server-js/lib && npm run build`
- [x] JS tests: `cd crates/trusted-server-js/lib && npm test`
- [ ] WASM release build: `cargo build --package trusted-server-adapter-fastly --release --target wasm32-wasip1`

## Checklist

- [x] No production logging changes
- [x] New behavior has a regression test


docs/guide/cli.md
```diff
@@ -118,31 +118,42 @@ deploy. Adapter passthrough arguments must now follow a `--` separator; unknown
...
+adapter flags directly need the `--` added. Trusted Server declares Config, KV,
+and Secret Stores, so EdgeZero treats every Fastly deploy as managed and
+requires a verified application release root via `--application-release`; a
+bare `ts deploy --adapter fastly` without a release is accepted only for
+store-free applications:
 
 ```bash
-ts deploy --adapter fastly --service-id <service-id> --staging
-ts deploy --adapter fastly -- --comment "release"
+ts deploy --adapter fastly --service-id <service-id> --application-release <release-root> --staging
```
ChristianPavilonis, line 128: 🔧 **P2: Select the release manifest in the new deployment examples**

From the repository-root workflow described by this guide, adding `--application-release <release-root>` still loads the checkout's `edgezero.toml`. The flag does not select the release's manifest. EdgeZero rejects that manifest because it lies outside the extracted release root, so both newly documented commands fail before deployment even with a valid extracted release.

The pinned dependency's `run_deploy` loads the manifest before resolving the release root. Running the documented command shape with `/tmp` as an external directory exited 2 with `loaded application manifest …/edgezero.toml is outside application release root /tmp`, before inspecting release contents.

Set `EDGEZERO_MANIFEST` to the extracted release's manifest in both examples, or explicitly change into the extracted release directory. Also link instructions for producing or obtaining the required release. For this command:

```suggestion
EDGEZERO_MANIFEST="<absolute-release-root>/edgezero.toml" \
  ts deploy --adapter fastly --service-id <service-id> --application-release "<absolute-release-root>" --staging
```

ChristianPavilonis (approved):
## Review summary

Reviewed ae47b4350b0297be935f258c9e9d81d22426bf4c against 6cae7f5da8911c746cf873581885f90c3820dd96. The logical-store runtime path passes local validation. No runtime correctness or security findings were confirmed. One P2 deployment-documentation finding is posted inline.

## Validation

- `cargo test-fastly --locked runtime_store_config_opens_logical_store_ids_and_key`: 1 passed.
- `cargo test --locked --manifest-path crates/trusted-server-integration-tests/Cargo.toml --test integration local_fastly --target x86_64-unknown-linux-gnu`: 3 passed.
- `cargo test-fastly --locked config_payload::tests::`: 21 passed.
- `cargo test-fastly --locked secret_resolution::tests::`: 8 passed.
- `cargo test --locked -p trusted-server-cli --target x86_64-unknown-linux-gnu run::tests::`: 29 passed.
- `BID_DELAY=3 ./scripts/template-cache-local-test.sh inline`: 8 assertions passed, exercising CLI-published configuration and logical secret bindings through Viceroy.
- `cd crates/trusted-server-js/lib && npm test -- --run`: 959 passed, no type errors.
- `cargo fmt --all -- --check` and `bash -n scripts/template-cache-local-test.sh`: passed.

CI was green. Existing reviews, inline comments, issue comments, and review threads were empty when inspected.

Residual risk: live Fastly link reconciliation and staging isolation were traced through pinned EdgeZero revision 5878eb740897749e4cfab7b410fa28791a371602 but not exercised against Fastly. JS filesystem-invalidation behavior was inspected but not mutation-tested. The worktree remained clean; no source edits or delegation.

docs/guide/cli.md
```diff
@@ -118,31 +118,42 @@ deploy. Adapter passthrough arguments must now follow a `--` separator; unknown
...
+ts deploy --adapter fastly --service-id <service-id> --application-release <release-root> -- --comment "release"
```
prk-Jr, line 129: 🔧 **[P2] Select the release manifest in both deployment examples**

From the checkout-root workflow, `--application-release` does not select the extracted release's `edgezero.toml`. The pinned EdgeZero CLI first loads the current manifest, then checks that it is inside the supplied release root. Both examples therefore fail before deployment when the release is outside the checkout, even if the release is valid. Select the release manifest explicitly in each command:

```suggestion
EDGEZERO_MANIFEST="<release-root>/edgezero.toml" \
  ts deploy --adapter fastly --service-id <service-id> --application-release "<release-root>" --staging
EDGEZERO_MANIFEST="<release-root>/edgezero.toml" \
  ts deploy --adapter fastly --service-id <service-id> --application-release "<release-root>" -- --comment "release"
```

prk-Jr (changes requested):
## Summary

Reviewed commit `ae47b4350b0297be935f258c9e9d81d22426bf4c`.

The logical store bindings match the pinned EdgeZero deployment model, and the JS watch list retains the inspected bundle inputs. The existing P2 documentation finding remains: both new deployment commands select the checkout manifest unless `EDGEZERO_MANIFEST` is set.

## Blocking

- 🔧 **Select the release manifest in both deployment examples** — see inline at `docs/guide/cli.md:128–129` (one-click suggestion). This confirms the existing unresolved finding and covers both commands.

## Validation

The exact locked dependency's path-resolution helper was independently compiled and exercised: a checkout manifest outside the release root was rejected; the release's own manifest was accepted. The exact documentation suggestion passes full docs formatting with no patch drift. Shell syntax and diff whitespace checks pass. Scratch edits were restored.

Live Fastly staging/resource-link reconciliation and a complete deployment were not exercised. Runtime gates rely on passing remote CI.

## CI Status

- [integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/35413912409/job/105819852822): PASS
- [integration tests (Fastly EC lifecycle)](https://github.com/IABTechLab/trusted-server/actions/runs/35413912409/job/105819852813): PASS
- [browser integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/35413912409/job/105819852793): PASS
- [CodeQL](https://github.com/IABTechLab/trusted-server/runs/105818802042): PASS
- [Analyze (actions)](https://github.com/IABTechLab/trusted-server/actions/runs/35413912527/job/105818687398): PASS
- [cargo test (ts CLI, native)](https://github.com/IABTechLab/trusted-server/actions/runs/35413912458/job/105818687270): PASS
- [cargo test](https://github.com/IABTechLab/trusted-server/actions/runs/35413912458/job/105818687297): PASS (required)
- [Analyze (rust)](https://github.com/IABTechLab/trusted-server/actions/runs/35413912527/job/105818687266): PASS
- [Analyze (javascript-typescript)](https://github.com/IABTechLab/trusted-server/actions/runs/35413912527/job/105818687417): PASS
- [cargo test (axum native)](https://github.com/IABTechLab/trusted-server/actions/runs/35413912458/job/105818687261): PASS
- [cargo fmt](https://github.com/IABTechLab/trusted-server/actions/runs/35413912574/job/105818687511): PASS (required)
- [format-typescript](https://github.com/IABTechLab/trusted-server/actions/runs/35413912574/job/105818687552): PASS (required)
- [cargo check (cloudflare native + wasm32-unknown-unknown)](https://github.com/IABTechLab/trusted-server/actions/runs/35413912458/job/105818687490): PASS
- [format-docs](https://github.com/IABTechLab/trusted-server/actions/runs/35413912574/job/105818687423): PASS (required)
- [cargo check/build/test (spin native + wasm32-wasip1)](https://github.com/IABTechLab/trusted-server/actions/runs/35413912458/job/105818687230): PASS
- [cargo test (cross-adapter parity)](https://github.com/IABTechLab/trusted-server/actions/runs/35413912458/job/105818687138): PASS
- [CLAUDE.md symlink guard](https://github.com/IABTechLab/trusted-server/actions/runs/35413912574/job/105818687655): PASS
- [vitest](https://github.com/IABTechLab/trusted-server/actions/runs/35413912458/job/105818687238): PASS
- [prepare integration artifacts](https://github.com/IABTechLab/trusted-server/actions/runs/35413912409/job/105818686956): PASS
- [Analyze (javascript-typescript)](https://github.com/IABTechLab/trusted-server/actions/runs/35413908282/job/105818677053): PASS

docs/guide/cli.md
aram356: Addressed in 6e9c50ca5. Reproduced the failure from the checkout root against the pinned CLI (exit 2, "loaded application manifest ... is outside application release root") and confirmed `EDGEZERO_MANIFEST=<release-root>/edgezero.toml` moves past that check. Both examples now set it, matching what EdgeZero's `deploy-fastly` action exports, and the section says where the release root comes from.

docs/guide/cli.md
aram356: Addressed in 6e9c50ca5. Reproduced the failure from the checkout root against the pinned CLI (exit 2, "loaded application manifest ... is outside application release root") and confirmed `EDGEZERO_MANIFEST=<release-root>/edgezero.toml` moves past that check. Both examples now set it, matching what EdgeZero's `deploy-fastly` action exports, and the section says where the release root comes from.

crates/trusted-server-integration-tests/tests/common/config.rs
```diff
@@ -74,26 +73,24 @@ mod tests {
...
             );
             assert!(
-                runtime_env
-                    .get("EDGEZERO__STORES__SECRETS__TRUSTED_SERVER_SECRETS__NAME")
+                local_server["config_stores"]
+                    .get("edgezero_runtime_env")
                     .is_none(),
-                "{name} should not define the ignored unscoped secret-store mapping"
+                "{name} should not define the legacy runtime selector store"
             );
```
prk-Jr, line 94: ♻️ **[P3] Also assert the legacy physical store is absent**

This asserts the store exists under the logical ID and that `edgezero_runtime_env` is gone, but nothing asserts `ts_secrets` itself is absent. Given this repo's history of squash merges re-adding deleted blocks, a resurrected `[[local_server.secret_stores.ts_secrets]]` table would pass CI while its entries land in a store nothing opens.

````suggestion
            assert!(
                local_server["secret_stores"].get("ts_secrets").is_none(),
                "{name} should not define the legacy physical secret store"
            );
            assert!(
                local_server["config_stores"]
                    .get("edgezero_runtime_env")
                    .is_none(),
                "{name} should not define the legacy runtime selector store"
            );
````

File: crates/trusted-server-adapter-fastly/src/app.rs [hunk 4]
```diff
@@ -162,11 +160,18 @@ pub(crate) struct RuntimeStoreConfig {
 }
 
 impl RuntimeStoreConfig {
-    pub(crate) fn from_env(env: &EnvConfig) -> Self {
+    /// Store bindings for the Fastly runtime.
+    ///
+    /// Fastly Compute has no process environment. `EdgeZero` links each selected
+    /// physical store to the service version under its logical ID, so the
```
prk-Jr, line 166: ⛏ **[P3] "Fastly Compute has no process environment" isn't accurate**

This crate reads the process environment in six places — `logging.rs:82-83`, `middleware.rs:211,219`, `platform.rs:727-728` (`FASTLY_HOSTNAME`, `FASTLY_REGION`, `FASTLY_SERVICE_VERSION`). The substantive point is right (no operator-settable *store selector* reaches the runtime), but the stated premise is wrong and a future maintainer will believe it.

````suggestion
    /// Store selection is a deploy-time input; no selector reaches the Fastly
    /// runtime. `EdgeZero` links each selected
    /// physical store to the service version under its logical ID, so the
````

File: crates/trusted-server-js/build.rs [hunk 5]
```diff
@@ -15,9 +15,21 @@ use build_print::{info, warn};
 use sha2::{Digest as _, Sha256};
 
 fn main() {
-    // Rebuild if TS sources change (belt-and-suspenders): enumerate every file under lib/
-    println!("cargo:rerun-if-changed=lib");
-    watch_dir_recursively(Path::new("lib"));
+    // Rebuild when the TypeScript sources or anything that shapes the bundle
```
prk-Jr, line 18: 👍 Nice fix. Worth noting *why* the explicit list is the right shape: a **missing** path in `rerun-if-changed` makes Cargo rerun the script on every build — the exact opposite of the intent — so this is only safe because the list is accurate. I verified all seven watched paths exist at this head. Dropping the recursive `lib/` walk that swept `node_modules` (33,932 directives → 7) is well scoped.

prk-Jr (changes requested):
## Summary

Coherent alignment with EdgeZero PR 381: the Fastly entry point now opens stores by logical ID, local Viceroy fixtures expose the secret store under that same ID, and the config-key/selector docs were rewritten to match. The `build.rs` watch-list fix is a clean, well-scoped win. Requesting changes on one item: the EdgeZero dependency pin.

> 2 of the inline comments below carry a one-click GitHub `suggestion` — use **Commit suggestion** (or **Add suggestion to batch**) to apply them. The blocking finding is in prose below because its fix touches both `Cargo.toml` and `Cargo.lock` and can't be auto-applied.

## Blocking

- 🔧 **Mutable branch pin lands on `main`, with no rev and no tracked follow-up** (`Cargo.toml:58-63`) — all six EdgeZero deps move from `tag = "v0.0.8"` to `branch = "fix/fastly-environment-store-selectors"`. The commit `12c3215c…` is recorded only in `Cargo.lock`, so any `cargo update` (including the targeted `-p` updates this repo runs for lockfile dep alignment) or an upstream force-push silently moves `main` onto different upstream bytes; once the upstream PR merges and the branch is deleted, fresh checkouts and `cargo update` on `main` break outright. `git log` on `Cargo.toml` shows `main` has only ever carried `rev =` or `tag =` pins — never an unmerged upstream feature branch.

  I verified locally that `rev = "12c3215c2637d961a27eefa12e235033fadb4c4a"` resolves to the byte-identical commit (`cargo metadata` OK; lockfile source becomes `git+…?rev=12c3215c…#12c3215c…`, same hash fragment). Please:

  1. Swap all six `branch =` to `rev = "12c3215c2637d961a27eefa12e235033fadb4c4a"` and refresh `Cargo.lock`.
  2. Open a tracked issue for replacing the rev with the release tag once stackpop/edgezero#381 merges, and link it here — the note in the PR body and the design doc isn't a work queue.

  Apply manually: the change spans `Cargo.toml` and `Cargo.lock`, so it can't be a one-click suggestion.

## Non-blocking

### 📝 documentation

- **Leftover guidance now contradicts the new model** (`docs/guide/fastly.md:304-305`) — "Do not copy the same app credential store under a second hardcoded `trusted_server_secrets` Fastly link. Configure the mapping instead." was right under the runtime-selector model. This PR makes the deployment link the selected physical store to the service version *under exactly that logical ID* (fastly.md:283-285 as rewritten here), and there is no longer a runtime mapping to configure. An operator reading the section top-to-bottom is now told both to create that link and not to. Delete or rewrite these two lines. (Outside the diff hunk, hence body-level.)

- **The Config Store selector is undocumented for Fastly** (`docs/guide/fastly.md`, Secret Stores / Config Store sections) — `docs/guide/cli.md:194` now makes staging isolation depend entirely on `EDGEZERO__STORES__CONFIG__<ID>__NAME`, but the only concrete example of that variable in the guides is `docs/guide/configuration.md:91`, which is a **Spin** example (`…=default`). `fastly.md` exports only the secrets selector (line 274), and its Config Store section still just says `fastly config-store create --name jwks_store`. An operator following `fastly.md` alone provisions and pushes app config with no Config Store selector set. Add the `EDGEZERO__STORES__CONFIG__TRUSTED_SERVER_CONFIG__NAME=<physical-store>` export alongside the secrets one.

### ♻️ refactor

- **Test doesn't guard against a resurrected `ts_secrets` block** — see inline at `crates/trusted-server-integration-tests/tests/common/config.rs:89`

### ⛏ nitpick

- **"Fastly Compute has no process environment" is inaccurate** — see inline at `crates/trusted-server-adapter-fastly/src/app.rs:165`

### 👍 praise

- **`build.rs` watch list** — see inline at `crates/trusted-server-js/build.rs:18`

## Checked, not flagged

- `edgezero.toml [stores.config] id = "trusted_server_config"` matches the runtime-hardcoded `DEFAULT_CONFIG_STORE_ID`.
- The unscoped `EDGEZERO__STORES__SECRETS__…` export in `fastly.md:274` is a **deploy-time CLI** input, not a runtime lookup, so the "unscoped selectors are ignored from v0.0.7+" rule doesn't apply to it. Correct as written.
- `crates/trusted-server-integration-tests` uses `workspace = true` for EdgeZero deps and carries no separate lockfile on this branch — no dep-alignment risk from the pin change.
- `scripts/template-cache-local-test.sh` secret-store rename is consistent with `fastly.toml`; no stale `ts_secrets` remains in any Viceroy fixture.
- Fastly is the only adapter changed; `settings_data.rs`'s `EnvConfig` helpers stay live for axum/spin. No dead code introduced.

## CI

16 PASS, 1 PENDING (`prepare integration artifacts`), 0 FAIL. All four required checks (`cargo fmt`, `format-typescript`, `cargo test`, `format-docs`) pass.

crates/trusted-server-integration-tests/tests/common/config.rs
aram356: Addressed in a1710a059. Added the `ts_secrets` absence assertion. Verified it catches the case rather than just compiling: reinserting a `[[local_server.secret_stores.ts_secrets]]` block into `fastly.toml` fails the test with "fastly.toml should not define the legacy physical secret store", and it passes once removed.

crates/trusted-server-adapter-fastly/src/app.rs
aram356: Addressed in a1710a059. You are right that the crate reads the process environment in `logging.rs`, `middleware.rs`, and `platform.rs`. The comment now states the real invariant, that no store selector reaches the runtime and selection is a deploy-time input, rather than the false general claim.

crates/trusted-server-js/build.rs
aram356: Thanks. Worth recording that the accuracy point cuts the other way too: a stale path in the list would make Cargo rerun the script on every build, so the list needs revisiting whenever the bundler inputs move.

aram356:
Addressed all findings from the latest review in a1710a059.

**Blocking, mutable branch pin.** All six EdgeZero dependencies now use `rev = "12c3215c2637d961a27eefa12e235033fadb4c4a"`, the commit the lockfile already resolved. `cargo metadata --locked` passes and the lockfile source is now `git+…?rev=12c3215c…#12c3215c…`, byte-identical to what the branch pin resolved to. No unrelated dependency edges moved. Follow-up tracked in #1195, and the PR body now describes the rev pin rather than the branch.

**Non-blocking, leftover guidance contradicts the new model.** The two lines at `fastly.md` are replaced: the deployment creates the `trusted_server_secrets` link from the selected physical store, so operators are told not to create it by hand.

**Non-blocking, Config Store selector undocumented for Fastly.** The Config Store section now exports `EDGEZERO__STORES__CONFIG__TRUSTED_SERVER_CONFIG__NAME` alongside the secrets selector, and says staging isolation comes from selecting a different physical store because the entry key is the logical ID on every target.

The two inline P3 items are answered on their threads.

**Verification.** `cargo metadata --locked`, `check-fastly`, and all eight clippy targets including the new `clippy-cli` and `clippy-codegen` pass. Docs format passes. The new test assertion was checked against a deliberately reinserted `ts_secrets` block, which fails it, then passes once removed. Full test suites are still running locally; CI covers the same gates.

docs/guide/cli.md
```diff
@@ -209,31 +209,53 @@ deploy. Adapter passthrough arguments must now follow a `--` separator; unknown
...
 
-A staged deploy only redirects the staged version's config selector at the
-`<logical-store-id>_staging` key — it does not copy the production config blob
-there. Push the staged config before probing the staged version:
+`<release-root>` is the extracted immutable application release that EdgeZero's
+`package-fastly-application-release` action produces from a `ts build`. The
+EdgeZero `deploy-fastly` action performs both steps, selects the manifest, and
+supplies the release root; see EdgeZero's GitHub Actions deployment guide
+(`docs/guide/deploy-github-actions.md` in the EdgeZero repository) for the
+producer and consumer workflow.
```
prk-Jr: 🔧 **[P2] Correct the application-release instructions**

The pinned EdgeZero revision `12c3215` provides `.github/actions/package-application-release-fastly/action.yml`; the documented `package-fastly-application-release` action does not exist. Also, `deploy-fastly` consumes a prebuilt release archive (requiring `app-release-archive`, `app-release-sha256`, and `expected-source-revision`), rather than building and packaging it. Since an application release is now required for Trusted Server deployments, these instructions obstruct the deployment workflow.

Verified against the pinned upstream action definitions; the replacement passes Prettier.

```suggestion
`<release-root>` is an extracted immutable application release. Build the
application CLI and Fastly package, then use EdgeZero's
`package-application-release-fastly` action to assemble the release archive.
The `deploy-fastly` action consumes that prebuilt archive, verifies and extracts
it, selects the manifest, and supplies the release root. See EdgeZero's GitHub
Actions deployment guide (`docs/guide/deploy-github-actions.md` in the EdgeZero
repository) for the producer and consumer workflow.
```

prk-Jr (changes requested):
## Summary

The logical Fastly store bindings match the pinned EdgeZero implementation. One deployment-documentation correction is required; no confirmed runtime regressions were found.

The inline comment carries one GitHub suggestion that can be applied with **Commit suggestion**.

## Blocking

- 🔧 **Correct the application-release instructions** — see inline at `docs/guide/cli.md:227–232`.

## Non-blocking

None.

## CI Status

All 20 reported checks pass:

- PASS: integration tests
- PASS: browser integration tests
- PASS: integration tests (Fastly EC lifecycle)
- PASS: CodeQL
- PASS: cargo test (ts CLI, native)
- PASS: cargo check (cloudflare native + wasm32-unknown-unknown)
- PASS: Analyze (javascript-typescript) (both reported runs)
- PASS: cargo test (cross-adapter parity)
- PASS: cargo fmt
- PASS: cargo test (axum native)
- PASS: Analyze (actions)
- PASS: format-typescript
- PASS: cargo test
- PASS: CLAUDE.md symlink guard
- PASS: format-docs
- PASS: cargo check/build/test (spin native + wasm32-wasip1)
- PASS: Analyze (rust)
- PASS: prepare integration artifacts
- PASS: vitest

The documentation suggestion was scratch-checked with Prettier; the reviewer worktree was restored clean.

docs/guide/cli.md
```diff
@@ -209,31 +209,53 @@ deploy. Adapter passthrough arguments must now follow a `--` separator; unknown
...
-binary reads `<logical-store-id>_staging` while the active production version
-continues to read the production key.
+The config key is fixed on Fastly: production, staging, and local Viceroy all
+read `<logical-store-id>`. Staging isolation comes from the physical store the
+staging environment selects with `EDGEZERO__STORES__CONFIG__<ID>__NAME`, never
+from a different key. Production and staging may select the same physical
+Config Store or different stores. After `ts config push --staging`, the staged
+binary reads `<logical-store-id>` in the store selected by the staging
+environment, while the active production version continues to read the same
+key in its own store.
```
dhruv8sh: 🔧 **wrench** — Following this paragraph as written can overwrite the production config entry.

At the pinned EdgeZero (`12c3215`), `config push --staging` on Fastly does not select a different store or key. `resolve_config_store_and_key` (`edgezero-cli/src/config.rs:1443`) takes the physical store from whatever `EDGEZERO__STORES__CONFIG__TRUSTED_SERVER_CONFIG__NAME` resolves to in the shell running the push, falling back to `trusted_server_config`, and the key is always the logical ID. `--staging` only changes validation. On `main` it wrote `trusted_server_config_staging`, which kept staging separate regardless of the shell.

Failure scenario: an operator reads "Production and staging may select the same physical Config Store", then runs the example above (`ts config push --adapter fastly --staging`, no selector) from their usual shell. That writes the live production entry, and production instances load the untested config on their next cold start, before the staged version is even probed.

Suggested wording (also makes the selector explicit; the example at lines 240-243 is outside the hunk, so please update it to set the selector as well):

```suggestion
The config key is fixed on Fastly: production, staging, and local Viceroy all
read `<logical-store-id>`. `--staging` does not choose a different store or
key: the physical store is whatever `EDGEZERO__STORES__CONFIG__<ID>__NAME`
resolves to in the shell or deploy job that runs the push, falling back to the
logical ID when it is unset. If production and staging select the same
physical Config Store, a staged push overwrites the production entry and
production instances load it on their next start. Give staging its own
physical Config Store and set its selector explicitly when pushing:

```bash
EDGEZERO__STORES__CONFIG__TRUSTED_SERVER_CONFIG__NAME=<staging-config-store> \
  ts config push --adapter fastly --staging
```
```

Related, outside the diff: `docs/guide/edgezero.md:19` still says `--staging` "writes `LOGICAL_ID_staging` in the same physical store", which contradicts this page.

File: docs/guide/cli.md [hunk 8]
```diff
@@ -209,31 +209,53 @@ deploy. Adapter passthrough arguments must now follow a `--` separator; unknown
...
 
 `--staging` on `config push` / `config diff` writes and compares the
-`<logical-store-id>_staging` key in the same store. It is mutually exclusive
-with `--key`: the staging key is derived from the store's logical id, so an
-explicit key would be written where nothing reads it.
+`<logical-store-id>` key in the physical store selected by the staging
+environment. It is mutually exclusive with `--key`: Fastly accepts only the
+logical store ID as the key for every target, so an explicit key would be
+written where nothing reads it.
```
dhruv8sh: ⛏ **nitpick** — "selected by the staging environment" suggests `--staging` picks the staging selector; it uses whatever the current shell resolves (see the comment above). Also, `--key` is not incompatible only with `--staging`: on Fastly any non-logical `--key` is rejected for every target.

```suggestion
`<logical-store-id>` key in the physical store selected by the current
environment. It is mutually exclusive with `--key`, and on Fastly any `--key`
other than the logical store ID is rejected for every target, staged or not.
```

docs/guide/fastly.md
```diff
@@ -271,43 +271,59 @@ Used for storing public configuration (e.g., public keys, key metadata):
...
 Create the separate request-signing store when that feature is enabled:
 
 ```bash
 fastly secret-store create --name signing_keys
 ```
 
-Do not copy the same app credential store under a second hardcoded
-`trusted_server_secrets` Fastly link. Configure the mapping instead.
+The deployment creates the `trusted_server_secrets` link itself from the
+selected physical store. Do not create that link by hand.
```
dhruv8sh: 🔧 **wrench** — There is no upgrade path here for services already deployed from `main`, and other docs still point at a deploy command that skips the new link.

Services on `main` link the physical `ts_secrets` store (plus `edgezero_runtime_env`). Only EdgeZero's managed deploy creates the `trusted_server_secrets` link; a plain `fastly compute publish` clones the active version's links. `docs/guide/getting-started.md:194` ("Deploy to Fastly: `fastly compute publish`"), `AGENTS.md:60`, and the `deploy =` command in `edgezero.toml:51` all still use it.

Failure scenario: an operator ships this binary with `fastly compute publish` (or activates it without a managed deploy) → the runtime opens `trusted_server_secrets`, which isn't linked → `load_settings_from_config_store` fails and every publisher route returns 500. `/health` keeps returning 200, so health checks can pass during the outage. Separately, staged config previously pushed under `trusted_server_config_staging` is now silently ignored.

Suggested fix (docs only): add a short "Upgrading from EdgeZero v0.0.8" block here:
- deploy only through the managed `ts deploy --application-release` / `deploy-fastly` action so the `trusted_server_secrets` link exists before activation;
- re-push staging config under the logical key into the staging store;
- remove any `EDGEZERO__STORES__CONFIG__*__KEY` selectors (`config push` now rejects them);
- optionally delete the stale `edgezero_runtime_env` link and `*_staging` entries.

And replace `fastly compute publish` in `getting-started.md:194` (and `AGENTS.md` / `edgezero.toml`) with the managed deploy or a pointer to it.

docs/superpowers/specs/2026-09-16-edgezero-fastly-store-selectors-design.md
```diff
@@ -0,0 +1,67 @@
...
+Store; that store is legacy. Trusted Server needs to consume that behavior and
+remove guidance for the superseded service-scoped and runtime-selector models.
+
+## Design
+
+All EdgeZero workspace dependencies will temporarily use the upstream branch
+`fix/fastly-environment-store-selectors`. Keeping the dependencies on one branch
+ensures the adapter, CLI, core, and platform-specific crates resolve to the same
+revision while PR 381 is under review. A follow-up will replace the branch with
+the release tag that contains the merged change.
```
dhruv8sh: ⛏ **nitpick** — The dependencies now use `rev =`, not the branch.

```suggestion
All EdgeZero workspace dependencies are pinned with `rev =` to the immutable
commit `12c3215c2637d961a27eefa12e235033fadb4c4a` on the upstream branch
`fix/fastly-environment-store-selectors`. Pinning every crate to one commit
ensures the adapter, CLI, core, and platform-specific crates resolve to the same
revision while PR 381 is under review. Issue #1195 tracks replacing the rev with
the release tag that contains the merged change.
```

crates/trusted-server-adapter-fastly/src/app.rs
```diff
@@ -162,11 +160,19 @@ pub(crate) struct RuntimeStoreConfig {
...
-    pub(crate) fn from_env(env: &EnvConfig) -> Self {
+    /// Store bindings for the Fastly runtime.
+    ///
+    /// Store selection is a deploy-time input; no store selector reaches the
+    /// Fastly runtime. `EdgeZero` links each selected physical store to the
+    /// service version under its logical ID, so the runtime opens stores by
+    /// logical ID and reads the config entry under that same ID for every
+    /// publication target. Staging isolation comes from linking a different
+    /// physical store, never from a different key.
+    pub(crate) fn logical() -> Self {
```
dhruv8sh: 👍 **praise** — Reducing `RuntimeStoreConfig` to fixed logical bindings instead of keeping a selector fallback is the right call. I checked it against EdgeZero at `12c3215`: it links each selected physical store under its logical ID and `validate_fastly_config_key` rejects any other key, so there is no configuration in which a fallback would be read. A missing store fails closed (publisher routes 500, `/health` still answers).

dhruv8sh (changes requested):
## Summary

Reviewed head `a1710a05` against EdgeZero at the pinned `12c3215`.

The runtime change is correct and small: the Fastly entry point opens `trusted_server_config` / `trusted_server_secrets` by logical ID and reads the config entry under the logical ID for every target, which matches how EdgeZero links stores at that commit and matches `edgezero.toml`. The blocking issues are in the operator docs: one staging instruction can overwrite production config, and there is no upgrade path for services already deployed from `main`.

Findings already raised by earlier reviews are not repeated here (branch pin → `rev =`, `EDGEZERO_MANIFEST`, Config Store selector example, `ts_secrets` assertion, runtime-env doc comment, and the `package-fastly-application-release` action name thread at `cli.md:232`).

## Blocking

### 🔧 wrench
- **Staging config push docs can overwrite production config** — see inline at `docs/guide/cli.md:245`
- **No upgrade path for services on `main`; `fastly compute publish` still documented** — see inline at `docs/guide/fastly.md:325`

## Non-blocking

### ⛏ nitpick / 🏕 camp site / 📝 note
- **`--staging` / `--key` wording** — see inline at `docs/guide/cli.md:255`
- **Design spec still says branch dependency** — see inline at the design spec
- **`scripts/smoke-fastly.sh` still targets `ts_secrets` and fails** — see cross-cutting
- **Pin is one commit behind PR 381 head** — see cross-cutting
- **Stale "derived staging key" test messages** — see cross-cutting

- 🔧 **`docs/guide/edgezero.md:19`** (outside diff) still says `--staging` "writes `LOGICAL_ID_staging` in the same physical store", contradicting the new `cli.md`.
- 🏕 **`scripts/smoke-fastly.sh:93,134-142` and `docs/guide/fastly.md:386`** still use `[local_server.secret_stores.ts_secrets]`, which nothing opens any more. Running the smoke on this head fails (`expected HTTP 500 ... received 200`, exit 1). To be fair: it fails identically on `main` (`a4e01eb5`) — the base `fastly.toml` already contains the three secrets and the awk only strips the appended unindented blocks, so the base values still resolve — and renaming the store alone does not fix it. Fix both (rename + strip/override base entries in the copied `fastly.toml`) or file a follow-up; CI does not run this script.
- 📝 **Pin is one upstream commit behind PR 381.** Upstream head is `61e8444` ("address managed Fastly deployment review"), which changes `config gc` store resolution (manifest defaults, rejects empty selectors) and touches provision and the release verifier. Low impact for Trusted Server, but worth bumping or noting in #1195.
- ⛏ **`crates/trusted-server-cli/src/run.rs:482,494,505,518`** (outside diff) still say "derived staging key" in test messages.

## Verification

Isolated worktree at `a1710a05`:

| Check | Result |
| --- | --- |
| `cargo fmt --all -- --check` | pass |
| `cargo clippy-fastly` | pass |
| `cargo build -p trusted-server-adapter-fastly --release --target wasm32-wasip1` | pass (unchecked in the PR test plan) |
| `cargo test-fastly -p trusted-server-adapter-fastly runtime_store_config` | 1 passed |
| integration `local_fastly` tests | 3 passed |
| `./scripts/smoke-fastly.sh` (PR head / `main`) | fails the same way on both |

EdgeZero source read at `12c3215`: `resolve_config_store_and_key`, `validate_fastly_config_key`, `ConfigPushArgs --staging`, and the custom-entrypoint migration guide.

## CI Status

All 20 checks pass on `a1710a05`, including required `cargo fmt`, `cargo test`, `format-typescript`, and `format-docs`.

PR #1176, opened 2026-09-16
Show help for a bare `ts dev proxy` invocation

### LINKED ISSUE #1063: Show help for a bare ts dev proxy invocation

dhruv8sh:
## Summary

- A bare `ts dev proxy` now prints Clap help and exits immediately, instead of touching Safari's system proxy state (prompting for `sudo`) and then failing with a debug-formatted `error-stack` report that leaked internal `file:line` paths.
- `ts dev proxy ca …` and an explicit-but-incomplete invocation (`--from` without `--to`) keep working exactly as before — verified this empirically against the pinned `clap` version, not just by reading the derive macro.
- Also switched proxy error rendering from `{report:?}` to `{report:#}` (not plain `{report}`): error-stack's non-alternate `Display` stops after the first context frame, which would have silently dropped the actual cause (e.g. "no rewrite rule: pass --map FROM=TO …") behind the generic wrapper message.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-cli/src/commands/dev/proxy/mod.rs` | Add `#[command(arg_required_else_help = true)]` to `ProxyArgs` so Clap shows help before `run` ever executes on a bare invocation |
| `crates/trusted-server-cli/src/commands/dev/mod.rs` | Render proxy errors with `{report:#}` (chains every `error-stack` context frame) instead of `{report:?}` (leaked source locations) |
| `crates/trusted-server-cli/src/commands/dev/proxy/config.rs` | Fix `base_args()`/`parse_args(&["ts"])` test helpers, which broke from the new `arg_required_else_help` (it applies to any `Command` `ProxyArgs` is flattened into, including these); now restate the `--listen` default explicitly so parsing doesn't hit the same help short-circuit and abort the test binary |
| `crates/trusted-server-cli/src/run.rs` | Add regression tests: bare invocation shows help (and `ca path` still parses under it), and a `--from`-only partial rule still parses instead of showing help |

## Closes

Closes #1063

## Test plan

- [ ] `cargo test-fastly && cargo test-axum` — not applicable, no fastly/axum/core files touched
- [ ] `cargo clippy-fastly && cargo clippy-axum` — not applicable
- [ ] JS tests: `cd crates/trusted-server-js/lib && npx vitest run` — not applicable
- [ ] JS format: `cd crates/trusted-server-js/lib && npm run format` — not applicable
- [ ] Docs format: `cd docs && npm run format` — not applicable
- [ ] WASM build: `cargo build --package trusted-server-adapter-fastly --release --target wasm32-wasip1` — not applicable
- [ ] Manual testing via `fastly compute serve` — not applicable (CLI-only, not a Fastly adapter change)
- [x] Other:
  - `cargo clippy -p trusted-server-cli --target aarch64-apple-darwin --all-targets --all-features -- -D warnings` — clean
  - `./scripts/test-cli.sh` (native `x86_64-unknown-linux-gnu`, macOS `cfg` gates temporarily removed locally to actually execute the macOS-only code on this Linux box) — 175 passed, 1 unrelated pre-existing failure (`restore_system_proxy_if_pending_removes_file_with_empty_service`, a genuinely macOS-only test hitting its own no-op branch off-macOS; untouched by this change)
  - Live-ran `ts dev proxy`, `ts dev proxy ca path`, and `ts dev proxy --from a.example.com` under the same native build to confirm real output matches the acceptance criteria

## Checklist

- [x] Uses `tracing` macros (not `println!`) — no new logging added
- [x] New code has tests


crates/trusted-server-cli/src/commands/dev/proxy/mod.rs
```diff
@@ -76,8 +76,13 @@ async fn finish_interrupted_run<Restore, Stop, Drain>(
     let _ = tokio::time::timeout(std::time::Duration::from_secs(2), drain_manager).await;
 }
 
+/// Default `--listen` address, shared with the `config` tests so they cannot
+/// silently drift from the real default.
+pub const DEFAULT_LISTEN: &str = "127.0.0.1:18080";
```
prk-Jr: ⛏ **nitpick** -- `DEFAULT_LISTEN` is exported through a `pub mod` chain, so it lands in the `trusted-server-cli` lib's public API. Its only consumers are the `#[arg(default_value = ...)]` a few lines below and the in-crate `config.rs` `#[cfg(test)]` module -- which is what the doc comment says ("shared with the `config` tests"). `crates/trusted-server-cli/tests/support/mod.rs`, the one integration-test consumer of this module, doesn't reference it. AGENTS.md asks us to "consider visibility carefully -- avoid unnecessary `pub`".

Verified in a scratch worktree with this applied alone: `cargo fmt --all -- --check` clean, CLI clippy `--all-targets --all-features -- -D warnings` clean (no `dead_code` or `unreachable_pub`), tests 178/7/29/1 passed, 0 failed.

```suggestion
pub(crate) const DEFAULT_LISTEN: &str = "127.0.0.1:18080";
```

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
```diff
@@ -359,30 +360,26 @@ mod tests {
...
+        W::try_parse_from(argv).expect("should parse proxy args").a
     }
 
     #[test]
     fn clap_parses_rewrite_host_as_a_bool() {
         assert!(
-            !parse_args(&["ts"]).rewrite_host,
+            !parse_args(&["ts", "--insecure"]).rewrite_host,
             "absent --rewrite-host is false"
         );
```
prk-Jr: ♻ **refactor** -- this assertion is about `rewrite_host` being false when the flag is absent, but it now parses `["ts", "--insecure"]`. `--insecure` has nothing to do with `rewrite_host`; it's there purely so the arg list is non-empty under the new `arg_required_else_help`. A reader has to reconstruct that. `base_args()` is defined ~20 lines above and exists for exactly this -- producing a default-valued `ProxyArgs` without tripping the help short-circuit -- so using it here also removes a second, inconsistent workaround for the same problem.

The replacement spans all four lines because rustfmt collapses the shortened `assert!` onto one line; swapping only the inner line fails `cargo fmt --all -- --check`. Verified in a scratch worktree from a clean reset with this applied alone: fmt clean, clippy `-D warnings` clean, `clap_parses_rewrite_host_as_a_bool ... ok`, tests 178/7/29/1 passed, 0 failed.

```suggestion
        assert!(!base_args().rewrite_host, "absent --rewrite-host is false");
```

prk-Jr (approved):
Reviewed at `afba58bc3` against `origin/main` (`35560c81d`), verified in a scratch worktree on macOS arm64.

The fix does what #1063 asked: `arg_required_else_help = true` short-circuits a bare `ts dev proxy` into clap's help *before* `browser::restore_system_proxy_if_pending` can run, so there is no sudo prompt and no system-proxy side effect on the failure path. I confirmed that behaviourally against a build of this head rather than inferring it from the derive macro:

```
$ ts dev proxy                  -> full clap help, exit 2, no sudo prompt
$ ts dev proxy ca path          -> /.../dev-proxy/ca-cert.pem, exit 0
$ ts dev proxy --from a.example.com -> concise "no rewrite rule" error, not help
```

I also checked the blast radius of the new flag. `grep -rn ProxyArgs crates/ --include=*.rs` turns up only `dev/mod.rs`, `proxy/mod.rs`, the `proxy/config.rs` tests and `tests/support/mod.rs:207`; every `resolve(&[...])` call site in `tests/support/mod.rs` already passes at least one arg, so none of them falls into the help short-circuit. Rules can't arrive from the environment either -- `config.rs` touches `std::env` only for `XDG_DATA_HOME` -- so a bare invocation is always an error and the flag cannot reject an otherwise-valid run.

Local results at head: `cargo fmt --all -- --check` clean, `cargo clippy --manifest-path crates/trusted-server-cli/Cargo.toml --target aarch64-apple-darwin --all-targets --all-features -- -D warnings` clean, `cargo test --manifest-path crates/trusted-server-cli/Cargo.toml --target aarch64-apple-darwin` 178/7/29/1 passed, 0 failed. All 20 remote checks are green. One note on the PR body: the "1 unrelated pre-existing failure" (`restore_system_proxy_if_pending_removes_file_with_empty_service`) passes here -- it looks like an artefact of running on Linux with the `cfg` gates stripped, and the macOS CI job agrees.

Nothing blocking. Three findings below: one stale-description/follow-up note, two readability nits.

---

**🤔 thinking -- the PR body describes an error-rendering change that isn't in the final diff, and #1063's first problem is still live**

The Summary and the `Changes` table both list the `dev/mod.rs` `{report:?}` -> `{report:#}` switch. Commit `1476137e` made that change and commit `ae1d7c6a` reverted it, so `dev/mod.rs` doesn't appear in `origin/main...HEAD` at all. Head is still:

```rust
DevCommand::Proxy(args) => proxy::run(&args).map_err(|report| format!("{report:?}")),
```

#1063 listed two problems and this PR says `Closes #1063`. Problem 2 (side effects before the error) is genuinely fixed. Problem 1 (internal `file:line` leaking into user-facing output) is not -- reproduced against a build of this head:

```
[ts] invalid rule configuration
├╴at crates/trusted-server-cli/src/commands/dev/proxy/mod.rs:261:41
│
╰─▶ no rewrite rule: pass --map FROM=TO (or -f/--from with -t/--to)
    ╰╴at crates/trusted-server-cli/src/commands/dev/proxy/config.rs:269:20
```

The revert itself is right, and worth recording so nobody undoes it: `error-stack` 0.6's `impl Display for Report<C>` filters `FrameKind::Attachment(_) => None` *before* the `alternate()` branch (`fmt/mod.rs:1164`), so `{:#}` drops every `.attach(...)` payload exactly like `{}` does -- `alternate()` only chains additional *contexts*. The `ca regenerate` abort path and the stale-CA-file removal path in `proxy/mod.rs` carry all of their actionable guidance in `.attach(...)` on top of a context that displays as just `"certificate authority error"`, so `{report:#}` would have shown the user that bare string and nothing else. Getting this right needs the CA errors to carry their guidance in a `Display` context (or `attach_printable` plus a custom renderer), which is a bigger change than this PR.

Could you drop those two lines from the body and file the rendering work as a follow-up issue? As written, `Closes #1063` will close an issue whose problem 1 is still open. (The issue text does sanction deferring it -- "acceptable to leave out and file separately" -- so this is a bookkeeping ask, not a code one.)

**👍 praise**

Switching `parse_from` -> `try_parse_from(...).expect("should parse proxy args")` in both `config.rs::parse_args` and `tests/support/mod.rs::resolve` is the right hardening, and it's a subtle one: with `arg_required_else_help` in play, `parse_from` would have had clap call `process::exit(2)` from inside a test, killing the test binary mid-run instead of failing a single test. Making `DEFAULT_LISTEN` the single source for both the clap `default_value` and the test helper is the right call too, and I appreciate that the `ca`-subcommand interaction #1063 flagged as risky was verified empirically rather than assumed -- I re-ran it independently and it holds.

---

Not verified on my side, for the record: `ca install | uninstall | regenerate` were parse-tested but never executed (they mutate the macOS login keychain and the on-disk CA), so the `{report:#}` claim above rests on the `error-stack` source plus the code paths rather than an observed failure. `--launch` browser orchestration and the Safari system-proxy restore path are unchanged by this PR and weren't exercised. Non-macOS hosts compile none of this code.

crates/trusted-server-cli/src/commands/dev/proxy/mod.rs
dhruv8sh: Applied as suggested in 40b65c62 — `pub(crate)`, verified fmt/clippy/tests clean on x86_64-unknown-linux-gnu.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
dhruv8sh: Applied as suggested in 40b65c62 — using `base_args()` instead of the unrelated `--insecure` flag, verified fmt/clippy/tests clean on x86_64-unknown-linux-gnu (proxy module is macOS-only so this particular test doesn't compile here, but the change is a straight substitution of the reviewer's own verified suggestion).

ChristianPavilonis (approved):
## Summary

Reviewed `40b65c62c56207325d3285538c7657a0dfa8ccb3` against `6cae7f5da8911c746cf873581885f90c3820dd96`. The Clap parsing guard prevents a bare `ts dev proxy` from reaching proxy startup, while CA subcommands and explicit partial rules still parse. macOS CI and the focused parser tests pass.

## P3 finding

🔧 **The guide still promises the old result for a bare invocation** (`docs/guide/ts-dev-proxy.md:80`)

The guide says every invocation without `--map`, `-f`, or `-t` produces the `no rewrite rule` error. A completely bare invocation now prints Clap help instead. The old error remains only when another explicit option reaches configuration resolution, so the guide contradicts the central behavior of this PR.

Please state that a bare invocation prints help, while an invocation containing options but no complete rewrite rule reports `no rewrite rule`.

prk-Jr (commented):
## Summary

Reviewed commit `40b65c62c56207325d3285538c7657a0dfa8ccb3`.

The bare invocation is stopped by Clap before proxy startup, and CA subcommands and explicit partial rules still parse. The prior visibility and test-helper suggestions have been applied. No new runtime correctness defect found.

## Non-blocking

🤔 **[P3] Align the description and guide with the actual help-only change**

Body-level: PR description and `docs/guide/ts-dev-proxy.md:80–81` (outside diff).

The PR body still claims a `{report:?}` → `{report:#}` error-rendering change, but `dev/mod.rs` is unchanged and still formats Debug. The guide says all invocations without rewrite flags print `no rewrite rule`, although the newly guarded bare invocation prints Clap help. Remove the inaccurate rendering claim and distinguish bare help from explicit incomplete-rule errors. The wider error-report rendering task may remain separate, as issue #1063 allows.

Proposed guide wording (apply manually because the guide is outside the diff):

```markdown
A bare `ts dev proxy` prints help and exits before proxy startup. An
invocation with explicit options but no complete rewrite rule reports
`no rewrite rule: pass --map FROM=TO (or -f/--from with -t/--to)`.
```

The prior `DEFAULT_LISTEN` visibility and unrelated `--insecure` test-workaround comments are resolved in commit `40b65c62`. The only changes since the previous reviewed commit are those two small source cleanups. CA install/uninstall/regenerate were inspected and parser-tested, not executed against the system keychain.

## Validation

All 32 CLI argument-parser tests and 17 proxy configuration tests passed locally on macOS arm64, including bare help, all four CA actions, and incomplete-rule parsing. System keychain/proxy operations were not executed. General build/lint gates rely on remote CI.

## CI Status

- [integration tests (Fastly EC lifecycle)](https://github.com/IABTechLab/trusted-server/actions/runs/35109243470/job/104841115030): PASS
- [browser integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/35109243470/job/104841115147): PASS
- [integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/35109243470/job/104841115198): PASS
- [CodeQL](https://github.com/IABTechLab/trusted-server/runs/104838609493): PASS
- [Analyze (javascript-typescript)](https://github.com/IABTechLab/trusted-server/actions/runs/35109240510/job/104838328679): PASS
- [cargo test (ts CLI, native)](https://github.com/IABTechLab/trusted-server/actions/runs/35109243484/job/104838331576): PASS
- [vitest](https://github.com/IABTechLab/trusted-server/actions/runs/35109243484/job/104838331574): PASS
- [format-typescript](https://github.com/IABTechLab/trusted-server/actions/runs/35109243468/job/104838331381): PASS (required)
- [Analyze (javascript-typescript)](https://github.com/IABTechLab/trusted-server/actions/runs/35109243758/job/104838333035): PASS
- [cargo test (axum native)](https://github.com/IABTechLab/trusted-server/actions/runs/35109243484/job/104838331605): PASS
- [cargo fmt](https://github.com/IABTechLab/trusted-server/actions/runs/35109243468/job/104838331349): PASS (required)
- [Analyze (rust)](https://github.com/IABTechLab/trusted-server/actions/runs/35109243758/job/104838333146): PASS
- [cargo check (cloudflare native + wasm32-unknown-unknown)](https://github.com/IABTechLab/trusted-server/actions/runs/35109243484/job/104838331248): PASS
- [cargo test](https://github.com/IABTechLab/trusted-server/actions/runs/35109243484/job/104838331583): PASS (required)
- [format-docs](https://github.com/IABTechLab/trusted-server/actions/runs/35109243468/job/104838331136): PASS (required)
- [cargo check/build/test (spin native + wasm32-wasip1)](https://github.com/IABTechLab/trusted-server/actions/runs/35109243484/job/104838331377): PASS
- [CLAUDE.md symlink guard](https://github.com/IABTechLab/trusted-server/actions/runs/35109243468/job/104838331447): PASS
- [prepare integration artifacts](https://github.com/IABTechLab/trusted-server/actions/runs/35109243470/job/104838332162): PASS
- [Analyze (actions)](https://github.com/IABTechLab/trusted-server/actions/runs/35109243758/job/104838333255): PASS
- [cargo test (cross-adapter parity)](https://github.com/IABTechLab/trusted-server/actions/runs/35109243484/job/104838331255): PASS

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
```diff
@@ -359,32 +360,25 @@ mod tests {
...
-            #[command(flatten)]
-            a: crate::commands::dev::proxy::ProxyArgs,
-        }
-        W::parse_from(["ts"]).a
+        parse_args(&[
+            "ts",
+            "--listen",
+            crate::commands::dev::proxy::DEFAULT_LISTEN,
+        ])
     }
```
aram356: 🔧 **wrench** — This rewrite removes all test coverage of the real `--listen` default.

The previous helper was `W::parse_from(["ts"]).a`, with the comment "Construct via clap so defaults match the real surface" — it asserted, implicitly, that clap's `default_value` produced the expected listen address. The new version passes `--listen DEFAULT_LISTEN` explicitly, so the test now feeds itself the same constant that the production attribute reads. It holds regardless of what `#[arg(... default_value = ...)]` actually says.

I verified this by mutation rather than by inspection: changing `default_value = DEFAULT_LISTEN` to `default_value = "1.2.3.4:9999"` in `proxy/mod.rs:102` leaves **all 178 tests passing**. That also makes the `DEFAULT_LISTEN` doc comment's claim ("shared with the `config` tests so they cannot silently drift from the real default") untrue as written.

I tried the two obvious in-place fixes first, and both break existing tests, because `base_args()` semantics are load-bearing:

- `parse_args(&["ts", "--allow-non-loopback"])` → breaks `non_loopback_listen_requires_flag`
- `parse_args(&["ts", "--map", "..."])` → breaks `no_rule_passed_is_a_no_rule_error`

So the clean shape is a separate test that keeps `base_args()` exactly as-is. This passes on the current head, and **fails** under the mutation above:

```rust
#[test]
fn clap_applies_the_real_listen_default() {
    // base_args() passes --listen explicitly to dodge
    // `arg_required_else_help`, so without this test nothing would catch the
    // real `default_value` drifting away from DEFAULT_LISTEN.
    let args = parse_args(&["ts", "--rewrite-host"]);
    assert_eq!(
        args.listen,
        crate::commands::dev::proxy::DEFAULT_LISTEN,
        "should apply the real clap --listen default"
    );
}
```

Apply manually — this adds a new test rather than replacing the approved range, so it can't be expressed as a one-click `suggestion`.

crates/trusted-server-cli/src/commands/dev/proxy/mod.rs
```diff
@@ -76,8 +76,13 @@ async fn finish_interrupted_run<Restore, Stop, Drain>(
     let _ = tokio::time::timeout(std::time::Duration::from_secs(2), drain_manager).await;
 }
 
+/// Default `--listen` address, shared with the `config` tests so they cannot
+/// silently drift from the real default.
```
aram356, line 80: ⛏ **nitpick** — The stated guarantee doesn't hold: because `base_args()` passes `--listen` explicitly, the `config` tests can't detect drift in the real `default_value` (see the mutation result in the `config.rs` comment). Reworded to claim only what the constant actually does.

This is only needed if you decline the separate-test fix above. If that test lands, the original wording becomes accurate and you can skip this one.

```suggestion
/// Default `--listen` address, also referenced by the `config` tests so the
/// expected value is written down once.
```

_(verified in a scratch worktree: `cargo fmt --all -- --check` clean, `cargo clippy --target aarch64-apple-darwin --all-targets -- -D warnings` clean, 178 tests pass, no post-verification drift)_

crates/trusted-server-cli/tests/support/mod.rs
```diff
@@ -206,7 +206,7 @@ fn resolve(argv: &[&str]) -> config::ResolvedConfig {
         #[command(flatten)]
         args: trusted_server_cli::commands::dev::proxy::ProxyArgs,
     }
-    let parsed = Wrapper::parse_from(argv);
+    let parsed = Wrapper::try_parse_from(argv).expect("should parse proxy args");
     config::resolve(&parsed.args).expect("should resolve test config")
 }
```
aram356, line 212: ⛏ **nitpick** — Swapping `parse_from` for `try_parse_from` here is the right call: `parse_from` calls `process::exit` on a parse failure, which under the new `arg_required_else_help` would terminate the test binary instead of failing an assertion.

Minor inconsistency while you're in this file: the sibling change moved `use clap::Parser as _` to module scope in `proxy/config.rs` (matching AGENTS.md's "no local imports within functions" rule), but `resolve()` here still declares it locally at line 204. Pre-existing, so 🏕 boy-scout territory rather than something this PR introduced.

aram356 (changes requested):
## Summary

The core fix works as intended: `ts dev proxy` with no arguments now prints Clap help and exits without touching system proxy state or attempting `sudo`, `ts dev proxy ca <action>` still parses under `arg_required_else_help`, and a partial `--from`-only rule still reaches `run`. Verified on real macOS (`aarch64-apple-darwin`): 178 CLI tests pass, `cargo fmt --all -- --check` is clean, and native clippy with `-D warnings` is clean.

Two blocking issues remain. The `{report:#}` error-rendering fix that the PR description presents as a headline change was reverted mid-review and is not in the head, and the `sudo`/side-effect half of #1063 still reproduces whenever any single flag is passed. Separately, the reworked `base_args()` helper silently dropped all test coverage of the real `--listen` default.

> 1 of the inline comments below carries a one-click GitHub `suggestion`. The remaining comments describe the fix in prose because the change touches files outside the diff, or adds a new test, and can't be auto-applied.

## Blocking

### 🔧 wrench

- **`{report:#}` rendering was reverted but is still claimed in the PR body** — see "Cross-cutting" below
- **#1063's `sudo` / side-effect bug still reproduces with any single flag** — see "Cross-cutting" below
- **New `base_args()` removes all coverage of the real `--listen` default** — see inline at `crates/trusted-server-cli/src/commands/dev/proxy/config.rs:363`

## Non-blocking

### ⛏ nitpick

- **`DEFAULT_LISTEN` doc comment overstates its guarantee** — see inline at `crates/trusted-server-cli/src/commands/dev/proxy/mod.rs:79`
- **Inconsistent `use clap::Parser as _` placement** — see inline at `crates/trusted-server-cli/tests/support/mod.rs:212`

- 🔧 **The `{report:?}` → `{report:#}` fix was reverted, but the PR body still claims it.** The PR description lists this as one of three headline changes, and issue #1063 makes it acceptance criterion #1 ("without ... printing `error-stack` debug output"). Commit `1476137e7` contained it; commit `ae1d7c6a3` reverted it. The head at `crates/trusted-server-cli/src/commands/dev/mod.rs:34` still reads `format!("{report:?}")`.

  Live run against this PR's head:

  ```
  [ts] invalid rule configuration
  ├╴at crates/trusted-server-cli/src/commands/dev/proxy/mod.rs:261:41
  │
  ╰─▶ no rewrite rule: pass --map FROM=TO (or -f/--from with -t/--to)
      ╰╴at crates/trusted-server-cli/src/commands/dev/proxy/config.rs:269:20
  ```

  Internal source locations still leak. I scratch-applied the one-character change and it renders exactly what the PR body describes, with both context frames chained and no `file:line`:

  ```
  [ts] invalid rule configuration: no rewrite rule: pass --map FROM=TO (or -f/--from with -t/--to)
  ```

  Please either restore the change in `dev/mod.rs`:

  ```rust
  DevCommand::Proxy(args) => proxy::run(&args).map_err(|report| format!("{report:#}")),
  ```

  or drop the claim from the PR description so the recorded intent matches the code. Apply manually — `dev/mod.rs` is not part of this diff, so there is no RIGHT-side hunk to anchor a suggestion to.

- 🔧 **Issue #1063's `sudo` / side-effect bug still reproduces with any single flag.** `arg_required_else_help` fires only on a *zero-argument* invocation. One unrelated flag satisfies it, and parsing then falls straight into `run`, which calls `browser::restore_system_proxy_if_pending` at `proxy/mod.rs:259` *before* `config::resolve` at `proxy/mod.rs:261`:

  ```
  $ ts dev proxy --insecure
  sudo: a password is required
  sudo: a password is required
  warning: Safari: could not auto-restore the system proxy for 'Wi-Fi' (needs admin). ...
  [ts] invalid rule configuration
  ├╴at crates/trusted-server-cli/src/commands/dev/proxy/mod.rs:261:41
  ```

  Reproduced identically with `--listen 127.0.0.1:18080` and with `--from a.example.com`.

  This is not a contrived input. `restore_system_proxy_if_pending` early-returns unless the `safari-proxy-restore` file exists, and that file was present on my machine from an earlier hard-killed run — exactly the state described in #1063. The flag-bearing invocation therefore hits the same `sudo` prompt and the same debug-formatted report that the issue was filed about.

  The issue's *suggested approach* was `arg_required_else_help`, but its *acceptance criteria* are behavioural: no touching of system proxy state and no `sudo` attempt before discovering there is no rule. The root cause is the ordering in `run`, so resolving rules first would close the whole class:

  ```rust
  // Resolve rules first: a bare-or-flagged invocation with no rule must not
  // touch system proxy state or attempt sudo before failing.
  let mut cfg = config::resolve(args).change_context(ProxyError::Config)?;

  browser::restore_system_proxy_if_pending(&config::ca_dir(args), false);
  ```

  Worth flagging honestly: the existing comment at `proxy/mod.rs:255-258` argues for the *current* ordering ("a missing/bad rule must not strand the system proxy"), so this is a genuine design tension rather than a simple oversight — swapping the order trades "a bad rule strands the proxy" against "a bad rule prompts for sudo". Either resolution is defensible; what should not stand is closing #1063 while its second stated problem is still reachable. If you would rather scope the ordering change to a follow-up issue, that is reasonable — please say so explicitly on this PR and in the issue so the remaining gap is recorded. Apply manually — the target lines are outside this PR's hunks.

## CI Status

All 20 reported checks pass. The `cargo test (ts CLI, native)` job runs on `macos-latest`, so the three new `#[cfg(target_os = "macos")]` tests do execute in CI rather than being silently skipped.

- integration tests: PASS
- CodeQL: PASS
- vitest: PASS
- cargo fmt: PASS (required)
- Analyze (rust): PASS
- cargo test: PASS (required)
- Analyze (actions): PASS

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
dhruv8sh: Removed `base_args()` entirely rather than patching it — every test now passes its args explicitly through `parse_args`, so there's no shared helper that can mask the real `default_value`. Kept `clap_applies_the_real_listen_default` (your suggested test, applied near-verbatim) as the one place that pins the real clap default independently of any test constant. Also caught two bugs of my own while retesting after this: a pair of invalid `--rewrite-host true/false` assertions (that flag takes no value) and a `no_rule_passed_is_a_no_rule_error` test whose `#[should_panic]` string didn't match reality and had stopped actually exercising `resolve()`'s `NoRule" path once the bare-`ts` case started panicking at parse time — split into `bare_invocation_is_rejected_at_parse_time` (parse-level) and a fixed `no_rule_passed_is_a_no_rule_error` that now uses `--insecure` (a flag, no rule) so it still reaches `resolve()`. 175/176 tests pass (the one failure is the pre-existing, unrelated macOS-only test noted in earlier review rounds); clippy and fmt clean.

crates/trusted-server-cli/src/commands/dev/proxy/mod.rs
dhruv8sh: The separate test landed (`clap_applies_the_real_listen_default`, see the reply below on the config.rs thread), so per your own note here — "if that test lands, the original wording becomes accurate and you can skip this one" — leaving the doc comment as-is.

crates/trusted-server-cli/tests/support/mod.rs
dhruv8sh: Agreed this is inconsistent, but leaving it — as you noted, it's pre-existing and this file has the same local-import pattern in two other places (`use rcgen::{...}` and `use x509_parser::prelude::FromDer as _`), so fixing just this one instance would be inconsistent without a broader pass. Out of scope for this PR's diff.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
```diff
@@ -666,9 +712,19 @@ mod tests {
         );
     }
 
+    #[test]
+    #[should_panic(expected = "DisplayHelpOnMissingArgumentOrSubcommand")]
+    fn bare_invocation_is_rejected_at_parse_time() {
+        // `arg_required_else_help` makes a fully-bare `ts` fail to parse at all,
+        // before `resolve` (and its `NoRule` check) ever runs.
+        parse_args(&["ts"]);
+    }
```
aram356: 🔧 **wrench** — This assertion is pinned to clap's *private* internal representation, not its public API.

`parse_args` surfaces the failure through `expect`, so the panic payload is the `Debug` rendering of `clap::Error`. I dumped the actual text to be sure rather than reasoning about it, and the string this test matches lives inside clap's non-public `ErrorInner` struct:

```
panic message: "should parse proxy args: ErrorInner { kind: DisplayHelpOnMissingArgumentOrSubcommand, context: FlatMap { keys: [], values: [] }, message: Some(Formatted(StyledStr(...
```

Two consequences:

1. `ErrorInner` and its `Debug` shape are not part of clap's public API. A clap patch release that renames the field, reorders it, or changes the derive would make this test pass or fail for reasons unrelated to `ts`. The pinned version is 4.6.1 today, but nothing here fails loudly if that representation drifts.
2. `#[should_panic]` cannot tell *where* the panic came from. Any future panic in this test body that happens to contain the expected substring satisfies it.

Credit where due: the substring is specific enough to discriminate the error *kind* — I verified that mutating the call to `parse_args(&["ts", "--no-such-flag"])` fails the test with `kind: UnknownArgument`, so it is not vacuous. The problem is the coupling, not the strength.

This PR already does the same assertion the right way, with clap's public typed API, in `run.rs:682-690`:

```rust
let error = Args::try_parse_from(["ts", "dev", "proxy"])
    .expect_err("a bare `ts dev proxy` should short-circuit to help, not run");
assert_eq!(
    error.kind(),
    clap::error::ErrorKind::DisplayHelpOnMissingArgumentOrSubcommand,
    "should print help instead of touching system proxy state or attempting sudo"
);
```

Applying that pattern here needs a fallible sibling to `parse_args`, since the current helper swallows the typed error:

```rust
fn parse_args(argv: &[&str]) -> crate::commands::dev::proxy::ProxyArgs {
    try_parse_args(argv).expect("should parse proxy args")
}

fn try_parse_args(
    argv: &[&str],
) -> Result<crate::commands::dev::proxy::ProxyArgs, clap::Error> {
    #[derive(clap::Parser)]
    struct W {
        #[command(flatten)]
        a: crate::commands::dev::proxy::ProxyArgs,
    }
    W::try_parse_from(argv).map(|w| w.a)
}

#[test]
fn bare_invocation_is_rejected_at_parse_time() {
    // `arg_required_else_help` makes a fully-bare `ts` fail to parse at all,
    // before `resolve` (and its `NoRule` check) ever runs.
    let error = try_parse_args(&["ts"])
        .expect_err("a fully-bare invocation should short-circuit to help");
    assert_eq!(
        error.kind(),
        clap::error::ErrorKind::DisplayHelpOnMissingArgumentOrSubcommand,
        "should short-circuit to help rather than reaching resolve"
    );
}
```

Note `try_parse_args` returns the inner `ProxyArgs` rather than the wrapper — returning `W` makes `expect_err` require `W: Debug`, which the local wrapper does not derive.

I verified this in a scratch worktree: `cargo fmt --all -- --check` clean, `cargo clippy-cli` clean, full CLI suite passing, and no post-verification drift. It also retains equivalent mutation coverage — deleting `#[command(arg_required_else_help = true)]` from `ProxyArgs` fails both this test and the `run.rs` one, same as the current `should_panic` version.

Apply manually — the change spans two hunks in this file (the `parse_args` helper and this test), so it cannot be expressed as a single `suggestion`.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
```diff
@@ -391,11 +380,25 @@ mod tests {
         );
     }
 
+    #[test]
+    fn clap_applies_the_real_listen_default() {
+        let args = parse_args(&["ts", "--rewrite-host"]);
+        assert_eq!(
+            args.listen,
+            crate::commands::dev::proxy::DEFAULT_LISTEN,
```
aram356: 📝 **note** — `ffd5bf8ff`'s commit message says it "fixes two invalid `--rewrite-host true/false` assertions (that flag takes no value)", but I can't find anything in the diff that corresponds to that.

I checked every commit on the branch plus the whole tree at head for `--rewrite-host` followed by a value, and there are no matches anywhere — not in the base, not in any intermediate commit, not now. What actually changed in this test is the *absent* case: `!base_args().rewrite_host` became `!parse_args(&["ts", "--from", ..., "--to", ...]).rewrite_host`. Both forms pass `--rewrite-host` as the bare flag it is.

No code problem here, and the test itself is fine. Flagging it only because it is the same class of drift as the `{report:#}` finding: a message claiming a change the diff does not contain. Since this message is what lands in `main`'s history, it is worth amending so a future bisector isn't hunting for a fix that was never made.

docs/guide/ts-dev-proxy.md
```diff
@@ -77,7 +77,8 @@ shorthand, or one or more `--map FROM=TO` rules:
 ts dev proxy -f www.example-publisher.com -t trusted-server-example.edgecompute.app
 ```
 
-With no `--map`/`-f`/`-t`, the proxy exits with
+A bare `ts dev proxy` prints help and exits before proxy startup. An
+invocation with explicit options but no complete rewrite rule reports
 `no rewrite rule: pass --map FROM=TO (or -f/--from with -t/--to)`.
```
aram356: 🤔 **thinking** — Correcting this paragraph was the right call, and the bare-invocation half is now accurate.

The second half is where I'd push slightly. As shipped, an invocation with explicit options but no complete rule does more than "report `no rewrite rule: ...`" — it prompts for `sudo` (twice, on a machine with a leftover `safari-proxy-restore` file), emits a Safari restore warning, and prints the message wrapped in an `error-stack` debug report with two `file:line` attachments. The concise one-line message the guide quotes is what users get only once the `{report:#}` finding is addressed.

So this text describes the intended end state rather than current behaviour. If the two blocking findings land in this PR, it becomes accurate exactly as written and needs no edit. If either is deferred, I'd soften it here so the guide doesn't promise output the binary doesn't produce yet.

No change requested on its own — it just resolves differently depending on how you decide the blocking findings.

aram356 (changes requested):
## Summary

Second review pass, against head `fe35b68ce`. The new test-rework commit (`ffd5bf8ff`) addresses my pass-1 test-coverage finding and goes further than asked: `base_args()` is gone entirely and all 17 existing tests now pass their arguments explicitly through `parse_args`, so no shared helper can mask the real `default_value` any more. I confirmed nothing was lost in the rework — 17 tests became 19, and the only removed symbol is the `base_args` helper itself. The dev-proxy guide correction is welcome and accurate.

Verified locally on macOS (`aarch64-apple-darwin`) at this head: the full CLI suite passes, `cargo fmt --all -- --check` is clean, and `cargo clippy-cli` is clean.

The two blocking findings from pass 1 are unchanged in the code and received no reply, so they are restated below. One new blocking finding concerns the `#[should_panic]` test added in `ffd5bf8ff`.

Note on the diff: `fe35b68ce` is a `main` merge that pulls in five unrelated PRs (#823, #1054, #1090, #1150, #1049). The PR's own delta against its new merge-base `4c6d26a16` is byte-identical to the previous head — the same five files, +156/−66 — so this review applies to the same change set.

> All findings below are prose-only. Each proposed fix either touches a file outside this PR's diff or spans more than one hunk, so none can be offered as a one-click `suggestion`.

## Blocking

### 🔧 wrench

- **`{report:#}` rendering is still reverted while the PR body still claims it** — see "Cross-cutting" below
- **#1063's `sudo` / side-effect bug still reproduces with any single flag** — see "Cross-cutting" below
- **`#[should_panic]` couples the new parse test to clap's private internals** — see inline at `crates/trusted-server-cli/src/commands/dev/proxy/config.rs:716`

## Non-blocking

### 📝 note

- **`ffd5bf8ff`'s message describes a `--rewrite-host true/false` fix with no counterpart in the diff** — see inline at `crates/trusted-server-cli/src/commands/dev/proxy/config.rs:380`

### 🤔 thinking

- **New guide paragraph is accurate but oversells the flag-only path** — see inline at `docs/guide/ts-dev-proxy.md:80`

- 🔧 **The `{report:?}` → `{report:#}` fix is still reverted, but the PR body still advertises it.** Restated from pass 1; unchanged at `fe35b68ce`. `crates/trusted-server-cli/src/commands/dev/mod.rs:34` still reads `format!("{report:?}")`. The PR description lists this as one of three headline changes, and issue #1063 makes it acceptance criterion #1 ("without ... printing `error-stack` debug output"). Commit `1476137e7` contained the fix; `ae1d7c6a3` reverted it.

  Live run at this head:

  ```
  [ts] invalid rule configuration
  ├╴at crates/trusted-server-cli/src/commands/dev/proxy/mod.rs:261:41
  │
  ╰─▶ no rewrite rule: pass --map FROM=TO (or -f/--from with -t/--to)
      ╰╴at crates/trusted-server-cli/src/commands/dev/proxy/config.rs:269:20
  ```

  I re-applied the one-character change in a scratch tree and it renders exactly what the PR body promises, both frames chained and no `file:line`:

  ```
  [ts] invalid rule configuration: no rewrite rule: pass --map FROM=TO (or -f/--from with -t/--to)
  ```

  Either restore it:

  ```rust
  DevCommand::Proxy(args) => proxy::run(&args).map_err(|report| format!("{report:#}")),
  ```

  or drop the claim from the PR description. Right now the recorded intent and the code disagree, and a reader of either one is misled. Apply manually — `dev/mod.rs` is not in this diff, so there is no RIGHT-side hunk to anchor to.

- 🔧 **Issue #1063's `sudo` / side-effect bug still reproduces with any single flag.** Restated from pass 1; unchanged at `fe35b68ce`. `arg_required_else_help` fires only on a *zero-argument* invocation, so one unrelated flag satisfies it and parsing falls through into `run`, which calls `browser::restore_system_proxy_if_pending` at `proxy/mod.rs:259` *before* `config::resolve` at `proxy/mod.rs:261`:

  ```
  $ ts dev proxy --insecure
  sudo: a password is required
  sudo: a password is required
  warning: Safari: could not auto-restore the system proxy for 'Wi-Fi' (needs admin). ...
  [ts] invalid rule configuration
  ├╴at crates/trusted-server-cli/src/commands/dev/proxy/mod.rs:261:41
  ```

  Reproduced identically with `--listen 127.0.0.1:18080` and `--from a.example.com`. This is not a contrived input: `restore_system_proxy_if_pending` early-returns unless the `safari-proxy-restore` file exists, and that file is present on my machine from an earlier hard-killed run — exactly the state #1063 describes. So the flag-bearing path still produces the same `sudo` prompt and the same debug-formatted report the issue was filed about.

  The issue's *suggested approach* was `arg_required_else_help`, but its *acceptance criteria* are behavioural: no touching of system proxy state and no `sudo` attempt before discovering there is no rule. Resolving rules first would close the whole class:

  ```rust
  // Resolve rules first: an invocation with no usable rule must not touch
  // system proxy state or attempt sudo before failing.
  let mut cfg = config::resolve(args).change_context(ProxyError::Config)?;

  browser::restore_system_proxy_if_pending(&config::ca_dir(args), false);
  ```

  In fairness, the comment at `proxy/mod.rs:255-258` argues for the *current* ordering ("a missing/bad rule must not strand the system proxy"), so this is a real tradeoff rather than an oversight: swapping the order trades "a bad rule strands the proxy" for "a bad rule prompts for sudo". Either resolution is defensible. What should not happen is closing #1063 while its second stated problem is still reachable. If you would rather scope this to a follow-up, that is fine — please say so on the PR and file the issue, so the remaining gap is recorded rather than lost. This is the second pass in which the finding has gone unanswered, so an explicit decision either way would help. Apply manually — the target lines are outside this PR's hunks.

## CI Status

CI was still running for `fe35b68ce` when this review was composed: 8 passing, 8 pending, 1 skipped, **0 failing**. Pending checks are recorded as pending, not as failures. The equivalent gates pass locally at this head (full CLI suite, `cargo fmt --all -- --check`, `cargo clippy-cli`). Worth noting for the earlier head `ffd5bf8ff`, all 20 checks passed, and the `cargo test (ts CLI, native)` job runs on `macos-latest`, so the `#[cfg(target_os = "macos")]` tests do execute in CI rather than being silently skipped.

- integration tests: PASS
- vitest: PASS
- Analyze (actions): PASS
- CodeQL: SKIPPED
- cargo fmt: PENDING (required)
- cargo test: PENDING (required)
- cargo test (ts CLI, native): PENDING
- cargo test (cross-adapter parity): PENDING
- cargo check/build/test (spin native + wasm32-wasip1): PENDING
- Analyze (rust): PENDING

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
dhruv8sh: Applied in 8dd9713b. `parse_args` now delegates to a fallible `try_parse_args(argv) -> Result<ProxyArgs, clap::Error>` (returns the inner `ProxyArgs`, so no `Debug` bound on the wrapper), and `bare_invocation_is_rejected_at_parse_time` asserts `error.kind() == clap::error::ErrorKind::DisplayHelpOnMissingArgumentOrSubcommand`, mirroring `run.rs`. The `#[should_panic]` string match on clap's private `ErrorInner` Debug output is gone.

crates/trusted-server-cli/src/commands/dev/proxy/config.rs
dhruv8sh: Correct, that sentence in the `ffd5bf8ff` message is inaccurate: no `--rewrite-host true/false` assertion existed in the base, any intermediate commit, or head. What the commit actually changed in `clap_parses_rewrite_host_as_a_bool` is the absent-flag case, `!base_args().rewrite_host` -> `!parse_args(&["ts", "--from", ..., "--to", ...]).rewrite_host`, as part of removing `base_args()`. I am not rewriting pushed history on a branch under review; the repo squash-merges, so the squash commit message will describe the change accurately and drop that claim.

docs/guide/ts-dev-proxy.md
dhruv8sh: Softened in d100d1b3. With the ordering change in the same commit, a flagged invocation without a rule no longer touches system proxy state or attempts `sudo`, but the rendering change is deferred, so it still prints the `error-stack` Debug report with `file:line`. The guide now says it "fails with a `no rewrite rule` error before touching system proxy state" rather than quoting the one-line message.

dhruv8sh:
@aram356 decisions on the two body-level findings:

**1. `{report:#}` rendering: deferred, PR body corrected.** Keeping the revert from `ae1d7c6a`. As you noted, error-stack 0.6's `Display` drops `.attach(...)` payloads even with `{:#}`, and the CA error paths (`ca regenerate` abort, stale-CA-file removal) carry all their guidance in attachments. Switching would reduce them to a bare "certificate authority error". The description no longer presents the rendering change as part of this PR, and `Closes #1063` is now `Refs #1063`. It states that problem 1 (the `file:line` leak) is deferred to a follow-up and that this PR does not close the issue.

**2. `sudo` / side effect with any flag: ordering changed in d100d1b3.** `run` now calls `config::resolve` before `browser::restore_system_proxy_if_pending`, so `ts dev proxy --insecure` (or `--listen …`, or `--from` alone) fails on the missing rule without touching system proxy state or attempting `sudo`. The tradeoff, which the updated comment in `run` spells out: a proxy stranded by a hard-killed run now stays stranded until the next run with a valid config. That run restores it before loading the CA, binding, or launching browsers. `resolve` only validates arguments (and reads `--basic-auth-file`) and never touches system proxy state, so running it first is safe. Regression test `no_rule_fails_before_touching_pending_system_proxy_restore` seeds a malformed restore file, which the restore path deletes without calling `networksetup`, and asserts the file survives a no-rule `run`. With the old ordering the test fails (checked by mutation).

Verification: `cargo fmt --all -- --check` and `cargo clippy-cli` are clean, and `./scripts/test-cli.sh` passes on x86_64 Linux. The proxy module is macOS-only, so I also ran the CLI suite and clippy with the macOS `cfg` gates temporarily switched to Linux: 511 lib tests and 29 `proxy_e2e` tests passed. The macOS CI job is the authoritative run.

PR #1179, opened 2026-09-17
Adopt EdgeZero reusable-sandbox lifecycle for Fastly

### LINKED ISSUE #856: Spike: opt-in reusable sandbox mode (Serve::with_max_requests)
### Description

From the 2026-07-05 performance & architecture review (wasm section, item 4).

Reusable sandboxes now exist as an opt-in (`Serve::new().with_max_requests()` in Rust). The design doc's rejection of cross-request state ("isolates are not pinned") is no longer absolute. This is a large lever: settings parse, integration registry, signing-key parse, and regex compilation could amortize across requests.

### Spike goal

Stand up the opt-in reusable sandbox mode and measure whether it holds up operationally. If it does, the per-request init items (settings pipeline, signing-key parse) and the registry/orchestrator construction amortize across requests, and the `OnceLock` caches the code already has start earning their keep.

This deserves a measured experiment, not an assumption either way.

### Effort / expected effect

- **Effort:** Spike
- **Expected effect:** Potentially obsoletes most per-request init cost.

_Sequencing item 10 of 10._

prk-Jr:
## Summary

- Fastly Compute normally starts a fresh Wasm sandbox per request, so every request repeats the whole initialization sequence: load and parse settings, resolve secrets, compile the auction plan, build the orchestrator and integration registry, construct the telemetry sink. This adopts EdgeZero's reusable-sandbox lifecycle so one sandbox can serve several requests and pay that cost once.
- **Reuse is off by default and opt-in twice over.** It needs the `reusable-sandbox` Cargo feature (which `fastly.toml` does not pass) *and* explicit bounds in the runtime config store. Absent or partial configuration stays single-request, so the shipped build is unchanged.
- Locally: **6 requests per sandbox with 1 build** instead of 6, and a reused request is ~11x faster than the feature-off baseline on an edge-only route. Deployed behaviour is **not** established — see Limitations.

## What changed, in order

The branch is meant to be read commit by commit.

| Commit | What |
| ------ | ---- |
| `c56ff745e` | Serve loop, no retained state. Feature flag, `main` split, logger guard, measurement counters, `ts dev sandbox-probe`. |
| `00c4e4278` | **Prerequisite:** move script-rewriter accumulation buffers into per-document state. |
| `ca7e7fa7a` | Retain the built application across requests. |
| `f51ba6f51` | vCPU/heap counters; record the local A/B/C results. |
| `0b447f804` | Test the real build decision; tighten evidence claims. |
| `dd74fa202`, `66529f8f1` | Cross-platform fixes (CLI `output` module on Linux). |
| `b6df0b3e9` | Pin EdgeZero to `277544c4` (CLI + Cloudflare fixes). |
| `281c89a98` | **Adopt `edgezero_adapter_fastly::lifecycle`**, deleting our equivalents. Pin `76c59b44`. |
| `52942ba8b` | Fix the final initialization count in retirement logs. |

### A cross-request disclosure fix, worth reading on its own

`GoogleTagManagerIntegration` and `NextJsNextDataRewriter` accumulated inline script fragments in a `Mutex<String>` **on the rewriter**, which the registry holds for its whole lifetime. If a document's stream ended before its final fragment — client disconnect, origin error, truncated body — the partial script stayed in that buffer and the next document prepended it, corrupting the response and potentially disclosing the previous document's content.

This is latent today only because each sandbox serves one request and then dies. Retention is what makes it reachable, so it is fixed in its own commit (`00c4e4278`) ahead of retention. Both regression tests fail against the previous code with the first document's content prepended to the second's response.

### What EdgeZero owns vs what we own

After `281c89a98` the framework owns lazy successful-only retention, the callback count, the initialization-attempt count, the one-time setup guard, and the serving wrappers. Deleted here:

| Removed | Replaced by |
| ------- | ----------- |
| local `struct Sandbox` | `lifecycle::Sandbox<RetainedApp>` |
| `resolve_app` / `retain_app` / `retained_app` | `Sandbox::initialize` |
| `ensure_logger` + `logger_installed` | `Sandbox::setup_once` |
| `begin_request` / `record_build` / … | `requests()` / `initialization_attempts()` |
| `Serve::new()…run_with_context(…)` | `serve_custom` / `run_custom` |

Still ours, deliberately: feature gate, kill switch, limit parsing and fallback; settings retention and refresh policy; health/JA4/metrics routing; fresh per-request metadata, handles, services, extensions, bodies and correlation ids; raw request conversion and router dispatch; response-extension finalization; progressive streaming; duplicate `Set-Cookie`; post-send work; per-document rewrite-buffer isolation.

`serve_app` is **not** adopted: its response conversion buffers streams, which would end progressive delivery and leave nowhere for response-extension finalization.

## Dependency

EdgeZero pinned by SHA to `76c59b440fb35d1317dcb3fa8c1172161e3f5309` on `feat/reusable-app-lifecycle`. `Cargo.lock` moves only the edgezero packages' `source` fields; no other dependency changes. The pin is an unmerged branch revision and should move to a release tag once one contains it.

## Closes

Closes #856

## Test plan

- [x] `cargo test-fastly-reuse` — new alias; `test-fastly` does not pass `--all-features`, so feature-on tests would otherwise never run
- [x] All 8 CI clippy invocations, incl. host-target CLI and openrtb-codegen
- [x] JS tests (959) and JS/docs format
- [x] Parity suite
- [x] Release WASM build, both feature configurations
- [x] Manual runtime validation under Viceroy 0.17.0 — see below
- [x] EdgeZero's own Fastly compatibility suite at the pinned revision: exit 0
- [x] Regressions mutation-checked: reintroducing each bug fails the corresponding test

## Local evidence

| Arm | Requests/sandbox | Builds/sandbox | p50 |
| --- | ---------------- | -------------- | --- |
| feature off | 1 | 1 | 3.546 ms |
| reuse, rebuild per request | 6 | 6 | 3.162 ms |
| reuse, retained | 6 | **1** | **0.308 ms** |

The middle row is the control: the serving loop alone buys almost nothing. The gain is retention.

Verified on observed reused sandboxes: progressive delivery (first byte ~2 s ahead of origin completion), duplicate `Set-Cookie`, request isolation with distinct per-request metadata, probes not constructing the app, failed initialization retried and never retained, and post-commitment stream failures producing exactly one response before a normal request succeeds on the same sandbox.

Full method, raw observations and provenance: `docs/superpowers/specs/2026-09-17-fastly-reusable-sandbox-results.md`.

## Limitations — deployed behaviour is not established

- **Long-lived memory behaviour and deployed eviction: unverified.** Viceroy admits ~6 requests per guest, so this cannot be established locally.
- Six requests per sandbox is a repeated **local observation** under Viceroy 0.17.0, not a demonstrated universal ceiling.
- Same-sandbox initialization recovery (fail → success → reuse) is **unit-level only**: Viceroy's stores are fixed for a guest's lifetime.
- Correctness checks used mock origins; the real publisher origin's bot wall blocks the paths under test.
- vCPU and heap readings are pre-send cumulative samples, not complete per-request costs.
- Reuse is never guaranteed: any request may start a fresh sandbox, so correctness holds for a cold sandbox on every request.
- Latency figures were measured at an earlier revision on this branch and were not re-measured after the lifecycle adoption; build-count and correctness were re-observed at the current pin.

## Rollback

No rung is immediate — limits are read at sandbox startup, so a running sandbox retires on the limits it started with.

```
config key -> 1   later sandboxes stop reusing
feature off       redeploy; restores the original entry path
revert 52942ba8b / 281c89a98 / b6df0b3e9   back to v0.0.8 + local lifecycle
revert ca7e7fa7a  removes retention, keeps the loop and the core fix
revert 00c4e4278  removes per-document buffers (only safe with retention reverted)
revert c56ff745e  removes the lifecycle entirely
```

## Checklist

- [x] New code has tests


crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -74,25 +81,249 @@ fn health_response(req: &FastlyRequest) -> Option<FastlyResponse> {
...
+        })
+}
+
+/// Attaches the sandbox counters to a workload response.
+///
+/// Called before headers are committed, which on the streaming path means
+/// before `stream_to_client`. The counters therefore describe the request up
+/// to commitment and cannot report its eventual outcome; a failure after
+/// commitment is recorded in logs instead and reconciled during analysis.
+fn attach_sandbox_counters(response: &mut HttpResponse, counters: &SandboxCounters) {
```
aram356, line 234: 🔧 **wrench** - Per-request counter headers can be cached and replayed to other users.

`attach_sandbox_counters` runs *after* `apply_terminal_response_effects` (see `send_edgezero_response` at `main.rs:614` then `main.rs:628-630`) and inserts `x-ts-sandbox-request-id`, `x-ts-sandbox-instance`, and `x-ts-sandbox-ordinal` onto **every** response. `main.rs:531` is the universal terminal send path, so publicly cacheable asset-proxy and publisher HTML responses go through here too. There is no `Vary`, no cache-privacy downgrade, and no exclusion for shared-cacheable responses — grepping `x-ts-sandbox` across the tree returns no cache handling for these names anywhere.

This is not confined to the reuse build. `SandboxCounters::capture` (`sandbox.rs:163-176`) is **not** feature-gated; only `debug.sandbox_metrics_enabled` controls it, and `settings.rs:2560-2564` documents that independence as deliberate so the feature-off baseline can be measured on the same channel. So the shipped, feature-off binary emits these headers whenever the config flag is on — which is exactly the configuration the results doc uses.

**Failure scenario.** With `debug.sandbox_metrics_enabled = true`, a response carrying `public, s-maxage=3600` (a proxied asset, or publisher HTML on a cacheable route) is stored in the Fastly cache with request A's unique `x-ts-sandbox-request-id` and `x-ts-sandbox-instance`. Every subsequent client served from that cache entry receives request A's correlation id for the next hour. That is a per-request identifier crossing users, and it silently corrupts the measurement channel these headers exist to provide: the probe in `sandbox_probe.rs` reads ordinals from exactly these headers, and a cached response replays a stale ordinal that `verdict()` would score as a repeated-ordinal anomaly.

**Proposed fix** (apply manually — needs a design decision, so it is not expressed as a `suggestion`). Pick one:

```rust
// Option A - attach only to responses that are already private/no-store.
// Cheapest and matches the "debug channel" intent.
if let Some(counters) = counters.as_ref()
    && trusted_server_core::response_privacy::is_private_no_store(&response)
{
    attach_sandbox_counters(&mut response, counters);
}

// Option B - attach unconditionally, but force the response private first,
// so a measurement run never populates a shared cache entry.
if let Some(counters) = counters.as_ref() {
    trusted_server_core::response_privacy::enforce_private_no_store(&mut response);
    attach_sandbox_counters(&mut response, counters);
}
```

Option B changes cache behaviour while measuring, which is itself a measurement confound worth stating in the results doc. Option A keeps caching intact and simply yields no counters on cacheable routes — the probe already handles absent counters correctly (`sandbox_probe.rs:260-281` reports `unverified` rather than a negative), so A degrades cleanly. Either way, please add a regression test alongside the existing cache-privacy tests in `main.rs` (`late_filter_effects_cannot_make_an_assembled_response_public` and siblings) asserting a cacheable response never carries `x-ts-sandbox-request-id`.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -74,25 +81,249 @@ fn health_response(req: &FastlyRequest) -> Option<FastlyResponse> {
...
         response.send_to_client();
         return;
     }
 
-    logging::init_logger();
-    edgezero_main(req);
+    // Marked complete only once installation succeeds, so a failed install is
+    // retried on a later callback. `setup_once` rolls nothing back, so that is
+    // only correct because `init_logger` is harmless to repeat — see its docs.
+    if sandbox.setup_once(crate::logging::init_logger).is_ok() {
```
aram356: 🤔 **thinking** - The `init_logger` failure is discarded without a trace.

`setup_once` returns `Result<(), String>` and this only tests `.is_ok()`, so the error string is dropped. The PR deliberately converted these two panics into errors (`logging.rs:91-117`, with a careful doc comment on why retry is safe), and then the one caller throws the value away.

Both failure modes are diagnostically interesting. A failed `Logger::builder().build()` means the `tslog` named endpoint is misconfigured; a failed `fern::Dispatch::apply()` means something else already installed a global logger. In either case every subsequent callback in the sandbox retries and fails the same way, invisibly, and `startup.flush()` never runs — so the diagnostics explaining *why reuse was declined* (`sandbox.rs:333-356`, the config-store and service-id messages) are silently stranded for the sandbox's whole life. That is precisely the situation where an operator most needs them.

Logging is unavailable by definition at this point, which is the obvious objection to logging the error. But `init_logger` already sets `echo_stdout(true)`, and the surrounding code path can still reach stdout, so a bare `eprintln!` is a genuine escape hatch here even though `println!`/`eprintln!` are otherwise banned by AGENTS.md — this is the one place where the logger provably does not exist.

**Proposed fix** (apply manually — an `eprintln!` in production code warrants an explicit `#[allow]` with a reason, so this is better as a deliberate edit than a one-click suggestion):

```rust
// The logger is unavailable by definition here, so this is the one place
// stderr is the only channel. A failed install repeats every callback, and
// the startup diagnostics stay stranded until it succeeds.
match sandbox.setup_once(crate::logging::init_logger) {
    Ok(()) => startup.flush(),
    Err(e) => {
        #[allow(clippy::print_stderr, reason = "no logger exists to report this")]
        {
            eprintln!("logger installation failed, retrying next callback: {e}");
        }
    }
}
```

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -74,25 +81,249 @@ fn health_response(req: &FastlyRequest) -> Option<FastlyResponse> {
...
+/// return [`ServeMode::Reuse`] without it.
+#[cfg(feature = "reusable-sandbox")]
+fn serve_loop(limits: crate::sandbox::SandboxLimits, mut startup: StartupDiagnostics) {
+    // `serve_custom` owns the `Sandbox` and drops it when serving ends, so the
+    // retirement line reports snapshots taken while it was still borrowed.
+    // `RetirementCounters` reads the attempt count on the way out, so a build
+    // performed by the final callback is included.
+    let mut counters = RetirementCounters::default();
+
+    let summary = edgezero_adapter_fastly::lifecycle::serve_custom(
```
aram356, line 127: 🌱 **seedling** - `with_max_memory` is the one available bound this loop leaves unset.

`serve_loop` sets `with_max_requests`, `with_max_lifetime`, and `with_timeout`, but not `with_max_memory`. In `fastly-0.12.1/src/http/serve.rs:225-232` that field defaults to `0`, and `0` is documented as unlimited — the loop's memory check at `serve.rs:306-314` is skipped entirely when it is zero. So a reused sandbox has no heap ceiling at all.

This lines up directly with the PR's own leading limitation: *"Long-lived memory behaviour and deployed eviction: unverified. Viceroy admits ~6 requests per guest, so this cannot be established locally."* The SDK ships a purpose-built rung for exactly that risk, and it is the one bound not being used. It also fits the existing policy shape rather than fighting it: `resolve_mode` (`sandbox.rs:223-244`) already refuses reuse unless *all* bounds are present, precisely so an omitted bound cannot silently become `Duration::MAX`. A memory bound omitted here becomes unlimited by the same mechanism the function was written to prevent.

Not blocking, and reasonable to defer — the feature is off by default and cannot engage without explicit configuration. But since the memory question is the headline unknown before any deployment, a fourth `TS__SANDBOX__MAX_MEMORY_MIB` key threaded through `RawLimits` / `SandboxLimits` / `resolve_mode` would let the first deployed experiment start bounded rather than open-ended. Worth deciding before this is switched on in a real service, not after.

Note the heap value is already being read on this path — `heap_mib()` at `main.rs:293-298` calls `heap_memory_snapshot_mib()`, the same accessor the SDK's own limit check uses — so the reporting side is in place and only the bound is missing.

aram356 (changes requested):
## Summary

An opt-in reusable-sandbox lifecycle for the Fastly adapter, gated twice over (Cargo feature plus three runtime bounds) so the shipped build is unchanged. The commit sequence is readable in order, the design and results docs are candid about what was not established, and the cross-request disclosure fix correctly lands in its own commit ahead of the retention that makes it reachable.

Verified independently in a reviewer worktree at `17f80ae0`: `cargo check -p trusted-server-adapter-fastly --target wasm32-wasip1 --features reusable-sandbox` clean, `cargo test-fastly-reuse` 211 passed / 0 failed, `cargo clippy-fastly` (`--all-features --all-targets -D warnings`) clean. The framework contract in the pinned EdgeZero `lifecycle.rs` and the SDK's `run_with_context` were read directly; the ordinal and attempt-count semantics the PR relies on hold as documented.

Two hypotheses were chased and did not hold, so they are not reported as findings: dynamic-backend registration growth under reuse is already mitigated by `canonicalize_transport_timeout_ms` quantization (with a test asserting at most 16 distinct values, explicitly "well under the dynamic backend limit"), and `StartupDiagnostics` is bounded at three messages pushed only in `main`.

One blocking finding, on the counters channel rather than the lifecycle itself.

> No inline comment below carries a one-click `suggestion`. Each fix needs a design decision or targets a file with no RIGHT-side hunk, so all are prose. Scratch verification (applying suggestion bytes in a worktree and re-running the gates) therefore had nothing to verify and was not run; the verification quoted above is of the PR head as-is, not of any proposed patch.

## Blocking

### wrench

- **Per-request counter headers can be cached and replayed to other users** - see inline at `crates/trusted-server-adapter-fastly/src/main.rs:222`

## Non-blocking

- **`init_logger` failure is silently discarded** - see inline at `crates/trusted-server-adapter-fastly/src/main.rs:176`
- **`with_max_memory` is the one available bound left unused** - see inline at `crates/trusted-server-adapter-fastly/src/main.rs:127`

- **♻️ `AppState` doc comment now asserts the opposite of what this PR makes true** - `crates/trusted-server-adapter-fastly/src/app.rs:174-177` still reads: *"In Fastly Compute each request spawns a new Wasm instance, so this struct is effectively per-request."* That is precisely the invariant this PR removes, and `AppState` is the struct now retained across requests by `RetainedApp`. A reader who trusts this comment will assume per-request isolation that no longer holds under the reuse path. The file has no RIGHT-side hunk in this diff, so this cannot be a suggestion; please update the comment to describe the sandbox-lifetime reality and note that retention is opt-in.

- **📝 A test the design doc commits to does not exist** - `docs/superpowers/specs/2026-09-17-fastly-reusable-sandbox-design.md:527-531` states the test plan asserts `IP_CIDR_SOURCE_CACHE`'s "key space is config-derived and not traffic-derived." Grepping `IP_CIDR` across `crates/` returns only `protection_scope.rs` itself, so no such test exists. The cache does appear genuinely config-bounded (keyed by `{config_store, key}` from `DataDomeConfig`), so this reads as a commitment gap rather than a defect. Worth resolving either way, because it is a module-scope `HashMap` with no eviction sweep whose caching becomes real for the first time under reuse: either add the test the plan names, or amend the plan to say why it was dropped.

- **📝 PR body names a stale EdgeZero pin** - the body says the pin is `76c59b440fb35d1317dcb3fa8c1172161e3f5309`, but the tree pins `c4841b609ec366ebabd3489416e2cb8c1359f61d` (`Cargo.toml:58-63`). Commits `385853bb3` and `17f80ae01` moved it after the body was written, and the body's commit table omits both. The body is the reviewer's map for a branch meant to be read commit by commit, so please refresh it. The body's own note that an unmerged branch revision should move to a release tag before merge still stands and is the right call.

## CI Status

All 20 reported checks PASS. No failures, cancellations, or pending checks.

- cargo fmt: PASS (required)
- cargo test: PASS (required)
- cargo test (cloudflare native): PASS
- vitest: PASS
- integration tests: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS

ChristianPavilonis (approved):
## Review summary

Approved on the condition that the outstanding requested changes in [the existing review](https://github.com/IABTechLab/trusted-server/pull/1179#pullrequestreview-5259580855) are addressed before merge. This approval does not waive those requests or resolve their threads.

Reviewed `17f80ae01cb98e5a590a9381d8944f8036b7b7f9` against `6cae7f5da8911c746cf873581885f90c3820dd96`, including the pinned EdgeZero lifecycle implementation and affected callers and consumers. No additional actionable findings beyond the existing feedback.

## Validation

- `cargo test-fastly-reuse --locked`: 211 tests passed.
- `cargo test-fastly --locked an_interrupted_document_leaves_no_residue_for_the_next_document`: both regressions passed.
- `cargo test-fastly --locked sandbox_metrics`: both serialization tests passed.
- `cargo test -p trusted-server-cli --target x86_64-unknown-linux-gnu --locked`: 98 tests passed, including the 11 sandbox-probe tests.
- `cargo build-fastly --locked --features reusable-sandbox`, `cargo fmt --all -- --check`, and `cargo clippy-fastly`: passed.
- Independent Viceroy 0.17.0 runtime check: health and counters probes performed no application build; four subsequent workload requests shared one build, preserved separate cookies and request markers, and delivered initial bytes approximately 0.36 seconds before origin completion.
- All 20 reported CI checks passed.

Deployed eviction and long-lived memory remain unverified. Same-sandbox initialization recovery is unit-tested, not established end to end. Post-commit stream-failure recovery was traced but not independently exercised in this review.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Fixed in ad4e4d2c6 using Option A. `attach_sandbox_counters` now requires both `private` and `no-store` in the final Cache-Control policy, after terminal response effects. Cacheable responses keep their policy and omit all sandbox counters in both feature configurations. The existing quote-aware parser prevents quoted extension values from enabling counters.

Regression coverage checks absent/public/browser-cacheable/qualified-private/quoted policies and terminal-private responses. The regression failed before the fix; the full Fastly and reuse suites pass (215 adapter tests each). Settings and results docs explain that cacheable workload probes are unverified, and historical benchmarks have not been rerun with this restriction.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Fixed in ad4e4d2c6. The caller now matches the setup result and reports the installation error through a narrowly scoped `eprintln!` fallback with an explicit clippy allowance and reason. Failed installation remains eligible for retry, and pending startup diagnostics flush once installation succeeds.

The existing setup retry regression passes in both Fastly suites; clippy passes. The stderr branch was checked by inspection rather than by adding test-only logger indirection.

crates/trusted-server-adapter-fastly/src/main.rs
prk-Jr: Implemented in ad4e4d2c6. `TS__SANDBOX__MAX_MEMORY_MIB` is now required alongside the other three bounds and is passed to `Serve::with_max_memory`. Missing, malformed, zero, overflowing, or `u32::MAX` values fall back to single-request operation. The sentinel is excluded so an unsupported heap snapshot conservatively retires the sandbox.

Tests cover invalid values and valid boundary values, and the full Fastly/reuse suites and all six target-matched clippy checks pass. Docs describe the three-key configuration migration and distinguish between-request retirement from a per-request allocation ceiling. Deployed heap support and long-lived memory behavior remain unverified.

crates/trusted-server-adapter-fastly/src/main.rs
```diff
@@ -74,25 +81,266 @@ fn health_response(req: &FastlyRequest) -> Option<FastlyResponse> {
...
+}
+
+/// Builds the counters snapshot response.
+///
+/// A snapshot only. It reports the sandbox that served *this* probe, which is
+/// not necessarily the sandbox that served any preceding workload request, so
+/// reuse is established from the counters attached to workload responses
+/// rather than from polling this.
+#[cfg(feature = "reusable-sandbox")]
+fn sandbox_metrics_response(sandbox: &Sandbox, ordinal: u64) -> FastlyResponse {
```
dhruv8sh, line 212: ⛏ **nitpick** — `/_ts/debug/sandbox`'s JSON always reports `"ordinal"` and `"requests"` as the same number.

```rust
fn sandbox_metrics_response(sandbox: &Sandbox, ordinal: u64) -> FastlyResponse {
    let body = serde_json::json!({
        "instance": instance_id(),
        "ordinal": ordinal,
        "requests": sandbox.requests(),
        "builds": sandbox.initialization_attempts(),
    });
```

`ordinal` is the current request's ordinal, captured at request entry; `sandbox.requests()` read here, later in the same request with nothing mutating the counter in between, returns that same value. An operator polling this endpoint could reasonably expect two different signals (e.g. "this probe's own ordinal" vs. "total lifetime requests"), but as written they're always identical and one field is redundant.

crates/trusted-server-core/src/integrations/nextjs/script_rewriter.rs
```diff
@@ -95,7 +89,7 @@ impl IntegrationScriptRewriter for NextJsNextDataRewriter {
         // accumulated text via Replace — intermediate fragments were already
         // removed from lol_html's output via RemoveNode.
         buf.push_str(content);
-        let full_content = std::mem::take(&mut *buf);
+        let full_content = std::mem::take::<String>(&mut buf);
```
dhruv8sh: ⛏ **nitpick** — `std::mem::take::<String>(&mut buf)` now relies on deref coercion from `&mut MutexGuard<String>` where the code previously wrote `&mut *buf` explicitly. Functionally identical and clippy-clean, just a small readability step back. Not worth a standalone comment on its own merits — mentioning only because it's the one stylistic wrinkle left after the accumulator migration to `ScriptTextAccumulator`.

dhruv8sh (approved):

crates/trusted-server-cli/src/commands/dev/sandbox_probe.rs
```diff
@@ -0,0 +1,542 @@
...
+/// means the attribution is unreliable — the id is not identifying what it
+/// claims to — so no conclusion is drawn either way. Counting bare repeats
+/// would instead report those runs as reuse.
+fn verdict(observations: &[Observation], ending: &Ending) -> String {
+    if let Ending::TransportError(reason) = ending {
+        return format!("reuse: unverified (transport error: {reason})\n");
+    }
+
+    if observations.iter().all(|o| o.instance.is_none()) {
+        return "reuse: unverified (no counters; enable debug.sandbox_metrics_enabled)\n"
```
aram356: 🔧 **wrench** - The "no counters" hint now misdirects operators, because the finding-1 fix added a second cause it does not mention.

`attach_sandbox_counters` now returns early unless the final policy contains both `private` and `no-store` (`main.rs:232-237`). So there are two distinct reasons a probe sees no counters: the flag is off, or the probed route is cacheable. This diagnostic still names only the first.

The failing case is reachable through this repo's own documented procedure, not a contrived one. `docs/superpowers/specs/2026-09-17-fastly-reusable-sandbox-results.md:60` gives the example invocation as:

```
ts dev sandbox-probe --path /.well-known/trusted-server.json --requests 8
```

That route is very unlikely to be `private, no-store`. An operator who has already set `debug.sandbox_metrics_enabled = true`, follows that command, and gets back "enable debug.sandbox_metrics_enabled" is being pointed at a box they have already ticked, with no hint that the route is the problem. The results doc's new follow-up section does say to "use an already-private/no-store workload", but the tool itself is where that lands at the moment it matters.

The fix is one string. The existing test `absent_counters_report_unverified_not_negative` keeps passing: it asserts the verdict contains `unverified` and does *not* contain `reusable-sandbox`, and the added clause introduces neither.

```suggestion
        return "reuse: unverified (no counters; enable debug.sandbox_metrics_enabled, \
                and probe a route whose response is private and no-store)\n"
            .to_owned();
```

_Scratch-verified in an isolated worktree at this head: `cargo fmt --all -- --check` clean, `cargo clippy-fastly` clean, `./scripts/test-cli.sh` 499 passed / 0 failed, and the post-verification drift check found no difference between the approved bytes and the tested tree._

Worth considering as a follow-up rather than here: `verdict()` could distinguish the two causes outright by checking whether any observed response carried a cacheable policy, and say which one applies. The string fix is the right size for this PR.

aram356 (changes requested):
## Summary

Round 2, against `147655e3`. All six findings from the previous pass are resolved, and each was re-derived from the code and the Fastly SDK rather than taken from the replies. Two new findings, both introduced by the round-1 fixes themselves.

Verified at this head in a reviewer worktree: `cargo test-fastly-reuse` 215 passed / 0 failed (up from 211, the new regressions), `cargo clippy-fastly` (`--all-features --all-targets -D warnings`) clean, `./scripts/test-cli.sh` 499 passed / 0 failed. All 20 CI checks pass.

### Round-1 findings, all confirmed fixed

- **Cacheable counter headers** - fixed with the right mechanism. Requiring *both* `private` and `no-store` through the existing quote-aware `cache_control_headers_have_directive`, evaluated after terminal effects. The two regressions cover the cases a naive `contains()` would fail: field-qualified `private="set-cookie"` and a quoted extension containing `private, no-store`.
- **Discarded `init_logger` error** - fixed, narrowly scoped `eprintln!` with an explicit clippy allowance and reason; diagnostics stay pending until installation succeeds.
- **`with_max_memory`** - implemented beyond what was suggested. Now a required fourth bound, and excluding the `u32::MAX` sentinel is a correct catch I had missed: the SDK does `heap_memory_snapshot_mib().unwrap_or(u32::MAX)` (`fastly-0.12.1/src/http/serve.rs:304-310`), so a bound *of* `u32::MAX` would let an unsupported reading fail to retire the guest. Excluding it makes unavailability retire conservatively.
- **`AppState` doc comment** - fixed, now describes opt-in retention accurately.
- **`IP_CIDR_SOURCE_CACHE` test** - added, and it is a real test: 32 requests varying method, path, query, IP and ASN with the TTL forced to zero, asserting the key set stays at exactly the two configured sources. That is the config-vs-traffic assertion the design doc promised.
- **Stale pin SHA in the body** - fixed, now matches `Cargo.toml`.

Also worth noting: the results doc states plainly that the earlier measurements were not rerun under the new private/no-store restriction, rather than leaving stale numbers implying continued validity.

### This round

Both findings follow from the finding-1 fix. The first is the blocking one: that fix added a second reason counters can be absent, and the diagnostic whose whole job is to explain their absence still names only the first.

> One inline comment below carries a one-click `suggestion`. It was scratch-verified in an isolated worktree: `cargo fmt --all -- --check` clean, `cargo clippy-fastly` clean, all 499 CLI tests pass (including `absent_counters_report_unverified_not_negative`, which asserts on this exact string), and the post-verification drift check found no difference.

## Blocking

### wrench

- **Probe's "no counters" hint is now incomplete and misdirects operators** - see inline at `crates/trusted-server-cli/src/commands/dev/sandbox_probe.rs:261`

- **📝 One forward-looking doc reference still names the superseded revision** - `docs/superpowers/specs/2026-09-17-fastly-reusable-sandbox-results.md:510` reads "Move to a release tag once one contains `76c59b44`." The tree now pins `c4841b60`, two revisions later, so a tag containing `76c59b44` would not necessarily contain what ships. The other `76c59b44` mentions in these docs are correctly historical ("Runtime observations at `76c59b44`", "`service_scoped_runtime_env_key` is private at `76c59b44`") and should stay as they are - this one is guidance about a future action, so it should point at the current pin. Not suggestion-eligible: the docs are outside this diff's RIGHT-side hunks.

## CI Status

All 20 reported checks PASS. No failures, cancellations, or pending checks.

- cargo fmt: PASS (required)
- cargo test: PASS (required)
- cargo test (cloudflare native): PASS
- vitest: PASS
- integration tests: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS

crates/trusted-server-cli/src/commands/dev/sandbox_probe.rs
prk-Jr: Fixed in 47df3d6bd. The no-counters hint now names both prerequisites: enable debug.sandbox_metrics_enabled and probe a response whose policy is private and no-store. Also corrected the forward-looking release-tag reference to c4841b60, preserving the historical revision references.

Merged main separately in 037fd5a84, retaining the document-isolation regressions and adapting the GTM test contexts to the new buffer-limit field. Validation on the merged tree passes: all eight clippy gates; Fastly (3,002), reuse (215), Axum (42), Cloudflare (45), Spin (87), parity (14), CLI (516 unit tests plus integration suites), and JS (1,130); Rust, JS, and docs formatting also pass. The CLI browser fixture helper passed as well.

PR #1181, opened 2026-09-18
Add regression test asserting Settings debug output excludes secret values

### LINKED ISSUE #476: Add regression test asserting Settings debug output excludes secret values
`Redacted<T>` wraps secret-bearing configuration fields so they print `[REDACTED]` instead of their value in `Debug` output. Nothing enforces that every secret field is actually wrapped. A new secret added without the wrapper would print in cleartext anywhere `Settings` is formatted, and no test would fail.

`Settings` currently has 17 `Redacted<…>` fields (`crates/trusted-server-core/src/settings.rs`), and exactly one test checks redaction: `trusted_client_ip_parses_and_redacts_shared_secret_in_debug_output` (line 3785), which covers a single field on a single sub-struct.

## What to add

A test that formats a fully-populated `Settings` and asserts no secret value survives:

1. Build a `Settings` where every secret field holds a distinctive canary value (for example `"CANARY_SECRET_PROXY"`, `"CANARY_SECRET_PASSPHRASE"`, one per field so a failure names the field that leaked).
2. Format it: `format!("{settings:?}")`.
3. Assert the output contains none of the canary strings.
4. Assert the output does contain `[REDACTED]`.

Use per-field canaries rather than one shared string. When this test fails in a year, the assertion message should say which field leaked, not just that something did.

## Why this is worth a test rather than a review habit

`Redacted<T>` lives in `crates/trusted-server-core/src/redacted.rs` and its `Debug` impl is what does the hiding. The failure mode is silent: a new `pub password: String` field is indistinguishable at review time from a correct one unless the reviewer happens to know the convention. This test converts that into a build failure.

## Acceptance criteria

- One test constructs a `Settings` with canary values in every secret-bearing field.
- It asserts the debug output contains no canary and does contain `[REDACTED]`.
- Assertions use the repository's `"should ..."` message convention.
- Deliberately unwrapping one field locally makes the test fail with a message naming that field. Check this by hand before opening the pull request; a test of this shape that cannot fail is worse than no test.

## Scope note

If you find a field that holds a secret and is *not* wrapped in `Redacted<…>`, that is a real finding. Say so in the pull request rather than quietly wrapping it, since changing a field's type touches serialization and deserves its own review.

## Pointers

- `crates/trusted-server-core/src/redacted.rs:30` — the `Redacted<T>` type
- `crates/trusted-server-core/src/redacted.rs:56` — the `Debug` impl that prints `[REDACTED]`
- `crates/trusted-server-core/src/settings.rs:3785` — the existing single-field test to model
- Original review context: https://github.com/IABTechLab/trusted-server/pull/468

dhruv8sh:
## Summary

- Add a test that formats a fully-populated `Settings` with a distinct canary in every `Redacted<String>` secret field and asserts none of them survive `Debug` formatting.
- Locks in the invariant that a new secret-bearing field added to `Settings` without the `Redacted<T>` wrapper will print in cleartext — previously only one field (`trusted_client_ip.shared_secret`) had this coverage.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/settings.rs` | Add `settings_debug_output_redacts_every_secret_field`, covering `publisher.proxy_secret`, `ec.passphrase`, `handlers[].username`/`password`, `ec.partners[].api_token`/`ts_pull_token`, `trusted_client_ip.shared_secret`, `proxy.asset_routes[].auth` (S3 `access_key_id`/`secret_access_key`/`session_token`), and `tinybird.auction_token_secret`/`access_token_secret` |

## Closes

Closes #476

## Test plan

- [x] Other: manually confirmed the test can fail — temporarily made `Redacted<T>`'s `Debug` impl print the inner value, reran the test to confirm it failed, then reverted.

Note: `DataDomeConfig`'s `credential_secret_name`/`server_side_key_secret_name` are not covered here since they live inside `IntegrationSettings`, whose hand-written `Debug` impl already only prints integration IDs, never field values — they're structurally unreachable from `Settings`'s debug output and don't need a canary.

## Checklist

- [x] New code has tests


crates/trusted-server-core/src/settings.rs
```diff
@@ -3806,6 +3806,138 @@ mod tests {
...
+        const CANARY_EC_PARTNER_TS_PULL_TOKEN: &str = "CANARY-EC-PARTNER-TS-PULL-TOKEN-0123456789";
+        const CANARY_TRUSTED_CLIENT_IP_SHARED_SECRET: &str =
+            "CANARY-TRUSTED-CLIENT-IP-SHARED-SECRET-0123456789";
+        const CANARY_S3_ACCESS_KEY_ID: &str = "CANARY-S3-ACCESS-KEY-ID-0123456789";
+        const CANARY_S3_SECRET_ACCESS_KEY: &str = "CANARY-S3-SECRET-ACCESS-KEY-0123456789";
+        const CANARY_S3_SESSION_TOKEN: &str = "CANARY-S3-SESSION-TOKEN-0123456789";
+        const CANARY_TINYBIRD_AUCTION_TOKEN: &str = "CANARY-TINYBIRD-AUCTION-TOKEN-0123456789";
+        const CANARY_TINYBIRD_ACCESS_TOKEN: &str = "CANARY-TINYBIRD-ACCESS-TOKEN-0123456789";
+
+        let mut settings = create_test_settings();
```
ChristianPavilonis, line 3836: 🔧 **P2: The test does not detect newly added unwrapped secret fields**

The test obtains most of `Settings` from `create_test_settings()` and checks only a fixed array of known canaries. A new defaulted secret `String` field on `Settings`, `Publisher`, `Ec`, `Proxy`, or another indirectly constructed type will not require this test to change. Even if the fixture supplies a new canary, the fixed assertions below will ignore it. `ProxyAssetRoute::new()` also initializes omitted fields through `Default`, despite the preceding claim of exhaustive construction.

The regression described by issue #476 can therefore still pass CI: a newly added secret without `Redacted<T>` can appear in derived `Settings` debug output while this test remains green. Only `Handler`, `EcPartner`, `TrustedClientIpConfig`, `S3SigV4AuthConfig`, and `TinybirdSettings` use exhaustive literals here.

Please add a guard that fails whenever the debug-reachable `Settings` schema changes, such as a stable full-debug snapshot with populated containers or exhaustive destructuring/literals for every relevant nested type. Confirm that adding a defaulted unwrapped secret field causes the test to fail without first editing its canary list.

ChristianPavilonis (approved):
## Summary

The current 12 `Redacted<String>` fields reachable through `Settings` are covered, and the focused redaction tests pass against the PR revision. I found one non-blocking coverage gap and noted it inline.

prk-Jr (approved):
## Summary

Reviewed commit `16437172621a48efc1083bb4e92cd962e4ebc8b5`.

The fixture covers all 12 current `Redacted<String>` fields reachable through `Settings` debug formatting with distinct canaries and field-labelled assertions. Integration configuration values are hidden by the existing IDs-only Debug implementation. No present-day redaction defect found.

The focused redaction test passed locally. This guards the enumerated current fields; it does not automatically detect every future secret field added through defaulted helper construction. That previously discussed limitation does not invalidate the current-field canary coverage.

General adapter/build/lint gates rely on remote CI.

## CI Status

- [integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/35336211229/job/105573501644): PASS
- [browser integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/35336211229/job/105573501571): PASS
- [integration tests (Fastly EC lifecycle)](https://github.com/IABTechLab/trusted-server/actions/runs/35336211229/job/105573501565): PASS
- [CodeQL](https://github.com/IABTechLab/trusted-server/runs/105571676374): PASS
- [cargo test (ts CLI, native)](https://github.com/IABTechLab/trusted-server/actions/runs/35336211275/job/105571526118): PASS
- [Analyze (javascript-typescript)](https://github.com/IABTechLab/trusted-server/actions/runs/35336211273/job/105571526258): PASS
- [cargo check (cloudflare native + wasm32-unknown-unknown)](https://github.com/IABTechLab/trusted-server/actions/runs/35336211275/job/105571526177): PASS
- [Analyze (actions)](https://github.com/IABTechLab/trusted-server/actions/runs/35336211273/job/105571526340): PASS
- [vitest](https://github.com/IABTechLab/trusted-server/actions/runs/35336211275/job/105571526243): PASS
- [cargo test (axum native)](https://github.com/IABTechLab/trusted-server/actions/runs/35336211275/job/105571526225): PASS
- [cargo test](https://github.com/IABTechLab/trusted-server/actions/runs/35336211275/job/105571526088): PASS (required)
- [format-typescript](https://github.com/IABTechLab/trusted-server/actions/runs/35336211302/job/105571526524): PASS (required)
- [cargo check/build/test (spin native + wasm32-wasip1)](https://github.com/IABTechLab/trusted-server/actions/runs/35336211275/job/105571526275): PASS
- [format-docs](https://github.com/IABTechLab/trusted-server/actions/runs/35336211302/job/105571526478): PASS (required)
- [CLAUDE.md symlink guard](https://github.com/IABTechLab/trusted-server/actions/runs/35336211302/job/105571526433): PASS
- [Analyze (rust)](https://github.com/IABTechLab/trusted-server/actions/runs/35336211273/job/105571526028): PASS
- [cargo test (cross-adapter parity)](https://github.com/IABTechLab/trusted-server/actions/runs/35336211275/job/105571526191): PASS
- [prepare integration artifacts](https://github.com/IABTechLab/trusted-server/actions/runs/35336211229/job/105571525989): PASS
- [cargo fmt](https://github.com/IABTechLab/trusted-server/actions/runs/35336211302/job/105571526454): PASS (required)
- [Analyze (javascript-typescript)](https://github.com/IABTechLab/trusted-server/actions/runs/35336209022/job/105571524730): PASS

crates/trusted-server-core/src/settings.rs
```diff
@@ -3806,6 +3806,138 @@ mod tests {
...
+    // `Settings`'s derived `Debug` impl. A new secret field added without the
+    // `Redacted` wrapper should be caught here by adding its own canary;
+    // re-run `rg 'Redacted<' crates/trusted-server-core/src` when touching
+    // this test to check the field list is still complete.
+    //
+    // Do not use `..Struct::default()` anywhere in this function. This test's
+    // entire purpose is exhaustive field coverage, and a default spread would
+    // silently swallow any field added to `Handler`, `TinybirdSettings`, or
+    // any other struct built here, defeating that coverage. List every field
+    // explicitly instead.
```
aram356: 🔧 **wrench** — This comment claims an enforcement the test cannot provide, which is worse than no comment: a future maintainer will trust it.

"A new secret field added without the `Redacted` wrapper should be caught here" is not achievable by this design. The test is a **denylist** of canaries it already knows about, checked against a derived `Debug`. A field added without `Redacted` is neither on the canary list nor redacted by anything, so the assertion loop never looks at it.

I confirmed this: adding a bare `String` secret field to `Ec`, populating it, and running the test gives `LEAK_PRESENT=true` with the test still passing.

The `rg 'Redacted<'` recipe is also backwards for the stated purpose — it can only find fields that are **already safe**, never the unwrapped field that is the actual risk. It also returns five hits, three of which are out of scope (`s3_sigv4.rs:42,44` and `ec/registry.rs:49` are runtime types unreachable from `Settings`; `crates/trusted-server-adapter-fastly/src/tinybird.rs` is a different crate), with no stated criterion for excluding them.

Separately, the no-default-spread rationale overstates its own mechanism. Omitting `..Default::default()` forces a **compile break** when a field is added — but the natural fix is to add the field with a non-canary value, which satisfies the compiler and adds zero assertion coverage. The compile break is a prompt for a human to look, not a coverage guarantee. The canary array has no compile-time link to the struct literals.

The rewrite below states what the test actually guarantees and records why integration configs are excluded. Verified in a scratch worktree: `cargo fmt --all -- --check` clean and the test passes.

```suggestion
    // One distinctive canary per `Redacted<String>` field reachable from
    // `Settings`'s derived `Debug` impl. This is a regression guard over the
    // field list below, not a completeness guarantee: a new secret field
    // added without the `Redacted` wrapper has no canary here and will pass
    // this test while leaking. Adding the canary is a manual step.
    //
    // Integration configs are deliberately out of scope. They reach
    // `Settings` as opaque JSON under `IntegrationSettings`, whose
    // hand-written `Debug` impl prints only integration IDs, never values.
    //
    // Do not use `..Struct::default()` anywhere in this function. A default
    // spread would let a new secret field be added to `Handler`,
    // `TinybirdSettings`, or any other struct built here without forcing
    // anyone to consider it. The compile break is the prompt; the canary
    // list below is still maintained by hand. List every field explicitly.
```

If you want the stronger guarantee the original comment implied, that belongs in CI rather than a comment — for example a check that every path in `TrustedServerAppConfig::secret_fields()` (`config.rs:135`) has a corresponding canary here. Happy to open a follow-up issue if that is worth doing.

crates/trusted-server-core/src/settings.rs
```diff
@@ -3806,6 +3806,138 @@ mod tests {
...
+            api_host: String::new(),
+            secret_store: None,
+            auction_dataset: String::new(),
+            access_enabled: false,
+            access_dataset: String::new(),
+            access_sample_rate: 0.0f64,
+            max_body_bytes: 0,
+        };
+
+        let debug = format!("{settings:?}");
```
aram356, line 3913: 🔧 **wrench** — The test skips the one redaction boundary in `Settings` that is actually fragile.

`Settings.integrations` carries **resolved plaintext secrets**, not secret *names*. `crates/trusted-server-core/src/config.rs:185-202` declares `integrations.datadome.server_side_key_secret_name` and `integrations.datadome.protection_test_bypass.credential_secret_name` as secret leaves in the app-config contract, so the runtime resolves them to plaintext **into that JSON tree** before `Settings` is deserialized.

The only thing keeping them out of `format!("{settings:?}")` is the hand-written `Debug` impl at `crates/trusted-server-core/src/settings.rs:224-233`, which prints just the sorted key list. That impl has **no test anywhere** — `rg integration_ids` across the workspace returns only the impl itself. Nothing stops someone from "simplifying" it to `#[derive(Debug)]`.

I confirmed the gap empirically. Replacing that hand-written impl with a derive and populating the DataDome key leaks the plaintext while this test still passes green:

```
DATADOME_LEAK=true
test settings::tests::settings_debug_output_redacts_every_secret_field ... ok
```

With a canary added, the same mutation fails correctly:

```
panicked at crates/trusted-server-core/src/settings.rs:3943:13:
should redact integrations.datadome.server_side_key_secret_name in Settings debug output
```

**Proposed fix** (apply manually — this spans three non-contiguous ranges in the file, so it can't be expressed as a single `suggestion` block):

```rust
// 1. Alongside the other canary consts. rustfmt requires this on one line:
const CANARY_DATADOME_SERVER_SIDE_KEY: &str = "CANARY-DATADOME-SERVER-SIDE-KEY-0123456789";

// 2. Immediately before `let debug = format!("{settings:?}");`
// `IntegrationSettings` stores integration configs as opaque JSON and
// relies on a hand-written `Debug` impl to suppress their values. That
// impl is the only thing keeping resolved DataDome credentials out of
// this output, so pin it here.
settings
    .integrations
    .insert_config(
        "datadome",
        &json!({
            "enabled": true,
            "server_side_key_secret_name": CANARY_DATADOME_SERVER_SIDE_KEY,
        }),
    )
    .expect("should insert datadome integration config");

// 3. As the final entry in the `canaries` array:
(
    "integrations.datadome.server_side_key_secret_name",
    CANARY_DATADOME_SERVER_SIDE_KEY,
),
```

Verified in a scratch worktree: `cargo fmt --all -- --check` clean, `cargo clippy-fastly` clean, `cargo test-fastly` 2,879 passed, `cargo test-axum` passed, and the mutation above fails as intended.

`json!` is already imported in this test module (`use serde_json::json;` at `settings.rs:3684`), and `insert_config` is available to tests via `#[cfg_attr(not(test), allow(dead_code))]` at `settings.rs:262`.

crates/trusted-server-core/src/settings.rs
```diff
@@ -3806,6 +3806,138 @@ mod tests {
...
+        const CANARY_TINYBIRD_ACCESS_TOKEN: &str = "CANARY-TINYBIRD-ACCESS-TOKEN-0123456789";
+
+        let mut settings = create_test_settings();
+
+        settings.publisher.proxy_secret = Redacted::new(CANARY_PROXY_SECRET.to_string());
+        settings.ec.passphrase = Redacted::new(CANARY_EC_PASSPHRASE.to_string());
+
+        settings.handlers = vec![Handler {
+            path: "^/secure".to_string(),
+            username: Redacted::new(CANARY_HANDLER_USERNAME.to_string()),
```
aram356, line 3848: ❓ **question** — `handlers[].username` is `Redacted<String>` at `crates/trusted-server-core/src/settings.rs:678` and gets a canary here, but it is **not** listed in `TrustedServerAppConfig::secret_fields()` (`crates/trusted-server-core/src/config.rs:167-173`), which registers only `handlers[].password`.

So the username is treated as secret by the type system but as non-secret by the deploy and secret-resolution contract. This is not a leak either way, and it does not block the test — but the two sides disagree, and this PR is the first thing to assert on the field.

Which side is intentional?

- If the username is genuinely secret, `secret_fields()` is missing an entry and operators cannot store it in a secret store.
- If it is not secret, the `Redacted` wrapper is over-applied and this canary is asserting a property nobody relies on — and `Redacted` also makes the value harder to use in logs where an operator may legitimately want to see which account was used.

Answering this in the PR is enough; I am not asking for a code change here unless you conclude `secret_fields()` is wrong.

aram356 (changes requested):
## Summary

Adds a single test that plants a distinct canary in each `Redacted<String>` field reachable from `Settings` and asserts none survives `Debug` formatting. The test is correct and every assertion is load-bearing, but its own framing claims more than it delivers: it does not catch a new secret field added without the `Redacted` wrapper, and it skips the one redaction boundary in `Settings` that is genuinely fragile.

> 1 of the inline comments below carries a one-click GitHub `suggestion` — use **Commit suggestion** to apply it as a commit on the PR branch. The remaining comments describe the fix in prose because the change spans several non-contiguous ranges in the file and can't be auto-applied.

### Verification performed

- **Mutation-tested all 12 canaries.** Replaced `Redacted`'s `Debug` with one that still prints `[REDACTED]` but appends the inner value. All 12 assertions fired; none is vacuous.
- **Independently enumerated** every `Redacted<String>` reachable from `Settings`. The 12 covered are complete. `s3_sigv4.rs:42,44` and `ec/registry.rs:49` are runtime types not reachable from `Settings` and are correctly out of scope.
- **Confirmed all five struct literals are exhaustive** and that every struct on the traversal path uses a derived `Debug`.
- Test passes; `cargo fmt --check` and `cargo clippy-fastly` clean.

## Blocking

### 🔧 wrench

- **Test skips the only fragile redaction boundary in `Settings`** — see inline at `crates/trusted-server-core/src/settings.rs:3893`
- **Leading comment claims an enforcement the test cannot provide** — see inline at `crates/trusted-server-core/src/settings.rs:3809`

### ❓ question

- **`handlers[].username` is `Redacted` but absent from `secret_fields()`** — see inline at `crates/trusted-server-core/src/settings.rs:3843`

- 📝 **PR description wording** — The description says "previously only one field (`trusted_client_ip.shared_secret`) had this coverage." That holds when read as "only one field *of `Settings`*", but two other tests already assert `Redacted` debug behaviour on types outside the `Settings` tree: `crates/trusted-server-core/src/s3_sigv4.rs:314-330` (`credentials_debug_redacts_secret_material`) and `crates/trusted-server-core/src/ec/registry.rs:612-634` (`partner_debug_output_redacts_pull_token`). Neither is made redundant by this PR, and neither should be deleted.

  Worth borrowing from `s3_sigv4.rs:314`: it asserts redaction is *selective* — `assert!(debug_output.contains("AKIAIOSFODNN7EXAMPLE"), "should leave non-secret access key ID visible")`. This PR only asserts secrets are absent, so a `Debug` impl that blanket-suppressed every field would pass it. A positive assertion that a known non-secret value (for example `handlers[].path` or `tinybird.api_host`) is still visible would close that hole cheaply.

  Also note `crates/trusted-server-core/src/settings.rs:3784-3807` (`trusted_client_ip_parses_and_redacts_shared_secret_in_debug_output`) covers the TOML deserialization path via `Settings::from_toml`, whereas this test constructs structs directly and never exercises serde. Keep both.

## CI Status

- integration tests: PASS
- CodeQL: PASS
- Analyze (actions): PASS
- vitest: PASS
- cargo test: PASS (required)
- Analyze (rust): PASS
- cargo fmt: PASS (required)

crates/trusted-server-core/src/settings.rs
dhruv8sh: Applied verbatim (the `insert_config` call, canary, and canary-array entry) plus the const. Mutation-verified: swapped `IntegrationSettings`'s hand-written `Debug` for a derive in a scratch copy, test failed as expected, reverted.

dhruv8sh: Applied your suggested rewrite verbatim as the new leading comment.

crates/trusted-server-core/src/settings.rs
dhruv8sh: Applied verbatim (const, insert_config call before the format!, and canary-array entry). Mutation-verified: swapped IntegrationSettings's hand-written Debug for a derive in a scratch copy, test failed as expected, reverted.

crates/trusted-server-core/src/settings.rs
dhruv8sh: Intentional, no code change: `handlers[].username` is a basic-auth login identifier, not credential material — the same category as `proxy.asset_routes[].auth.access_key_id`, which is also wrapped in `Redacted` here despite not being a value operators would keep confidential. Wrapping non-secret identifier-like fields in `Redacted` is an existing pattern in this codebase (defense-in-depth against inadvertent disclosure alongside the password it's paired with in logs), not a signal that `secret_fields()` should provision it through the secret store. `secret_fields()` staying password-only reflects that operators only need out-of-band secret storage for the password. Leaving both as-is.

crates/trusted-server-core/src/settings.rs
dhruv8sh: Addressed via aram356's fix below (applied): the leading comment now states this is a regression guard over the field list, not a completeness guarantee, and explicitly says a new unwrapped secret field has no canary and will pass while leaking. Adding a schema-change guard (stable full-debug snapshot or exhaustive destructuring per nested type) is a larger change than this PR's scope — happy to open a follow-up issue for it if wanted.

crates/trusted-server-core/src/settings.rs
```diff
@@ -3806,6 +3806,166 @@ mod tests {
...
+        // relies on a hand-written `Debug` impl to suppress their values. That
+        // impl is the only thing keeping resolved DataDome credentials out of
+        // this output, so pin it here.
+        settings
+            .integrations
+            .insert_config(
+                "datadome",
+                &json!({
+                    "enabled": true,
+                    "server_side_key_secret_name": CANARY_DATADOME_SERVER_SIDE_KEY,
```
aram356: 📝 **note** — `crates/trusted-server-core/src/config.rs:196-199` declares a second DataDome secret leaf, `integrations.datadome.protection_test_bypass.credential_secret_name`, which has no canary here.

I checked whether that matters and it does not, today: both leaves sit behind the same `IntegrationSettings::fmt` boundary (`settings.rs:224-233`), so the `server_side_key_secret_name` canary already pins the impl that protects both. Adding a nested `protection_test_bypass.credential_secret_name` canary and re-running confirmed it never appears in the debug output for the same reason the first one doesn't.

No change requested. Worth knowing only if that `Debug` impl is ever split per-integration or made field-selective, at which point the two leaves would stop sharing a guard and this canary would no longer cover the bypass credential.

aram356 (approved):
## Summary

All three findings from the previous pass are resolved. Re-reviewed at `b70cb57a` against `main`.

### Verification performed

- **Mutation-tested the new DataDome canary.** Replaced `IntegrationSettings`'s hand-written `Debug` with `#[derive(Debug)]` in a scratch worktree; the test now fails with `should redact integrations.datadome.server_side_key_secret_name`. The boundary is genuinely pinned.
- **Mutation-tested the new selective-redaction assertion.** Gave `Handler` a blanket-suppressing `Debug` impl; the test fails with `should leave non-secret handler path visible`. It catches over-redaction, not just under-redaction.
- **Re-ran the full canary leak mutation.** 12 of the 13 canaries leak when `Redacted`'s `Debug` is made to print its inner value. The DataDome canary correctly does not, because it is raw JSON guarded by `IntegrationSettings::fmt` rather than by `Redacted` — two distinct mechanisms, each pinned by its own canary.
- **Local gate:** `cargo fmt --all -- --check` clean, `cargo clippy-fastly` clean, `cargo test-fastly` 2,879 passed, `cargo test-axum` 41 passed.

### Resolution of prior findings

- **`IntegrationSettings` debug boundary** — fixed. Canary applied and independently mutation-verified.
- **Leading comment overstated its guarantee** — fixed. The comment now states this is a regression guard over the listed fields, not a completeness guarantee, and says plainly that a new unwrapped secret field will pass while leaking.
- **`handlers[].username` vs `secret_fields()`** — answered, no change needed. The username is a login identifier wrapped for defense-in-depth alongside the password, matching the existing treatment of `proxy.asset_routes[].auth.access_key_id` at `settings.rs:883`, which is also `Redacted` without being provisioned through the secret store. `secret_fields()` staying password-only is consistent with that.

## Non-blocking

### 📝 note

- **Second DataDome secret leaf has no canary** — see inline at `crates/trusted-server-core/src/settings.rs:3908`

## CI Status

All 20 checks PASS on `b70cb57a`:

- integration tests: PASS
- CodeQL: PASS
- Analyze (actions): PASS
- vitest: PASS
- cargo test: PASS (required)
- Analyze (rust): PASS
- cargo fmt: PASS (required)

PR #1182, opened 2026-09-18
Reject oversized bid dimensions instead of truncating them

### LINKED ISSUE #420: Integer truncation on bid dimensions in adserver_mock
Bid width and height arrive as `u64` from an external bidder response and are cast to `u32` with `as`, which silently wraps rather than failing. A value above `u32::MAX` becomes an unrelated small number, so a bid that should be rejected can pass the zero-check below and flow onward with corrupted dimensions.

## Scope is smaller than this issue originally said

The issue named two files. Since it was filed, the Prebid path has been handled and now has test coverage for the oversized case (see the `i32::MAX as u32 + 1` tests in `crates/trusted-server-core/src/integrations/prebid.rs`). Only the mock ad-server path is still affected:

`crates/trusted-server-core/src/integrations/adserver_mock.rs:295-296`

```rust
let width = bid["w"].as_u64().unwrap_or(0) as u32;
let height = bid["h"].as_u64().unwrap_or(0) as u32;
if width == 0 || height == 0 {
    // skipped
}
```

The wrap is what makes this worth fixing rather than a style nit: `4_294_967_296_u64 as u32` is `0`, which the zero-check happens to catch, but `4_294_967_396_u64 as u32` is `100`, which passes as a plausible dimension.

Also note the issue text cites `crates/common/`; that crate is now `trusted-server-core`.

## Suggested fix

Use a checked conversion and treat a value that does not fit as invalid input, consistent with how the same file already treats a zero dimension: log at debug and skip the bid. There is a precedent for the conversion style at `crates/trusted-server-core/src/integrations/prebid.rs:3027`, which uses `u32::try_from(v).ok()`.

Decide deliberately between skipping the bid and clamping, and say which you chose in the pull request. Skipping matches the existing zero-dimension behavior in this function and is the safer default; clamping invents a dimension the bidder never sent.

## Acceptance criteria

- A bid whose `w` or `h` exceeds `u32::MAX` is rejected rather than silently truncated.
- A bid with a zero or missing dimension keeps its current behavior.
- A bid with ordinary dimensions is unaffected.
- Tests cover the oversized case and confirm the existing cases still pass. `adserver_mock.rs` already has 16 tests to model.

## Context

This is one finding from the production readiness audit tracked in #396. Treat the audit's file paths as approximate: the crate layout changed after those issues were written.

dhruv8sh:
## Summary

- Bid width and height in the mock ad-server mediation path were cast from `u64` to `u32` with `as`, which silently wraps values above `u32::MAX` into small, plausible-looking dimensions that pass the zero-check.
- Use `u32::try_from` and skip the bid (log at debug) when a dimension doesn't fit, consistent with the existing zero-dimension handling.
- Chose skip over clamp: it matches the existing zero-dimension behavior and doesn't invent a dimension the bidder never sent.

## Changes

| File | Change |
| ---- | ------ |
| `crates/trusted-server-core/src/integrations/adserver_mock.rs` | Replace truncating `as u32` casts on bid `w`/`h` with `u32::try_from`, skipping the bid on overflow; add tests for oversized and zero/missing dimensions |

## Closes

Closes #420

## Test plan

- [ ] Other: 

## Checklist

- [x] New code has tests


crates/trusted-server-core/src/integrations/adserver_mock.rs
```diff
@@ -1323,6 +1333,114 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_parse_mediation_response_skips_oversized_dimensions() {
+        // A dimension above u32::MAX must be rejected rather than silently
+        // wrapped into a small, plausible-looking value.
+        let config = AdServerMockConfig::default();
+        let provider = AdServerMockProvider::new(config);
+
+        let oversized_width = u64::from(u32::MAX) + 1;
```
ChristianPavilonis, line 1343: 🔧 **The oversized regression test also passes with the original truncating cast**

`u64::from(u32::MAX) + 1` equals `2^32`. The old `as u32` conversion produces zero, so the pre-existing zero-dimension guard already rejects both oversized fixtures. Reintroducing the original cast would leave this test green while values such as `2^32 + 100` enter the auction as width or height `100`.

Use `u64::from(u32::MAX) + 101` for both oversized cases, or add equivalent nonzero-wrap fixtures. This makes the test fail with the original implementation.

ChristianPavilonis (approved):
## Summary

The checked conversion correctly prevents oversized mock ad-server dimensions from entering auction consumers. One medium-severity test gap remains: the oversized fixture wraps to zero under the old implementation, so it does not reproduce the original nonzero-truncation defect.

prk-Jr (commented):
## Summary

Reviewed commit `6f95f1d5096240b80fe7c44bb25ddd4c07edb3ea`.

The checked conversion correctly rejects overflowing dimensions and preserves ordinary, zero, and missing-dimension behavior. One non-blocking test improvement remains.

## Non-blocking

- 🤔 **[P2] Exercise a nonzero wrapped value in the overflow regression** — see inline at `crates/trusted-server-core/src/integrations/adserver_mock.rs:1343`.

## Validation

All 18 mock-adserver tests pass at the reviewed head. Mutation checks confirm that the original cast passes the current fixture, fails with a `u32::MAX + 101` fixture, and the checked conversion passes with that stronger fixture. Scratch source was restored byte-for-byte. These checks establish the test gap; they are not a full release-gate validation of a proposed patch.

## CI Status

- [browser integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/35356365217/job/105639217629): PASS
- [integration tests](https://github.com/IABTechLab/trusted-server/actions/runs/35356365217/job/105639217594): PASS
- [integration tests (Fastly EC lifecycle)](https://github.com/IABTechLab/trusted-server/actions/runs/35356365217/job/105639217586): PASS
- [CodeQL](https://github.com/IABTechLab/trusted-server/runs/105636794229): PASS
- [cargo test (ts CLI, native)](https://github.com/IABTechLab/trusted-server/actions/runs/35356365307/job/105636583879): PASS
- [vitest](https://github.com/IABTechLab/trusted-server/actions/runs/35356365307/job/105636583901): PASS
- [Analyze (actions)](https://github.com/IABTechLab/trusted-server/actions/runs/35356365179/job/105636582771): PASS
- [Analyze (javascript-typescript)](https://github.com/IABTechLab/trusted-server/actions/runs/35356365179/job/105636583175): PASS
- [cargo fmt](https://github.com/IABTechLab/trusted-server/actions/runs/35356365216/job/105636582988): PASS (required)
- [cargo test (axum native)](https://github.com/IABTechLab/trusted-server/actions/runs/35356365307/job/105636583724): PASS
- [format-typescript](https://github.com/IABTechLab/trusted-server/actions/runs/35356365216/job/105636582593): PASS (required)
- [cargo check (cloudflare native + wasm32-unknown-unknown)](https://github.com/IABTechLab/trusted-server/actions/runs/35356365307/job/105636583700): PASS
- [cargo check/build/test (spin native + wasm32-wasip1)](https://github.com/IABTechLab/trusted-server/actions/runs/35356365307/job/105636583439): PASS
- [cargo test (cross-adapter parity)](https://github.com/IABTechLab/trusted-server/actions/runs/35356365307/job/105636584129): PASS
- [cargo test](https://github.com/IABTechLab/trusted-server/actions/runs/35356365307/job/105636583778): PASS (required)
- [CLAUDE.md symlink guard](https://github.com/IABTechLab/trusted-server/actions/runs/35356365216/job/105636582887): PASS
- [prepare integration artifacts](https://github.com/IABTechLab/trusted-server/actions/runs/35356365217/job/105636583118): PASS
- [Analyze (rust)](https://github.com/IABTechLab/trusted-server/actions/runs/35356365179/job/105636583148): PASS
- [format-docs](https://github.com/IABTechLab/trusted-server/actions/runs/35356365216/job/105636583047): PASS (required)
- [Analyze (javascript-typescript)](https://github.com/IABTechLab/trusted-server/actions/runs/35356360939/job/105636573236): PASS

crates/trusted-server-core/src/integrations/adserver_mock.rs
prk-Jr, line 1343: 🤔 **[P2] Exercise a nonzero wrapped value in the overflow regression**

The fixture uses `u64::from(u32::MAX) + 1`, which becomes zero under the original `as u32` cast. The pre-existing zero check already rejects it. I restored the old cast temporarily: the new regression still passed. Changing only the fixture to `+ 101` made the old implementation fail (three bids instead of one), and the checked conversion passed with that stronger fixture.

Suggested test value:

```rust
        let oversized_width = u64::from(u32::MAX) + 101;
```

The production fix is sound; strengthening its regression would ensure reintroducing the original bug fails CI. This confirms the previously raised test gap.

aram356: 🔧 **wrench** — This test passes against the unfixed code, so it does not pin the regression.

`u64::from(u32::MAX) + 1` is `4_294_967_296`, which as a `u32` truncates to **exactly 0** — and that is caught by the pre-existing zero-check at line 307. On `main` the old `as u32` code skipped these same bids, just for a different reason, so the assertion holds either way.

Issue #420 called this out directly: "`4_294_967_296_u64 as u32` is `0`, which the zero-check happens to catch, but `4_294_967_396_u64 as u32` is `100`, which passes as a plausible dimension." The value chosen here is the one the issue named as *not* demonstrating the bug.

I grafted these two tests onto the merge-base commit `6cae7f5da` and ran them against the unfixed parser:

```
# base code + this test as written (u32::MAX + 1):
test test_parse_mediation_response_skips_oversized_dimensions ... ok      <- passes on the bug

# base code + same test with u32::MAX + 101:
test test_parse_mediation_response_skips_oversized_dimensions ... FAILED
  assertion `left == right` failed: Bids with oversized w/h should be skipped
  left: 3    right: 1
```

With the `+ 101` offset the two oversized bids truncate to `100x90` and `300x100`, both of which clear the zero-check and flow onward with corrupted dimensions — precisely the corruption this PR fixes. The same test passes on this PR's head.

```suggestion
        // A dimension above u32::MAX must be rejected rather than silently
        // wrapped into a small, plausible-looking value. The offset of 101 is
        // load-bearing: u32::MAX + 1 truncates to 0, which the zero-check
        // below would already have caught, so it would not pin this fix.
        let config = AdServerMockConfig::default();
        let provider = AdServerMockProvider::new(config);

        let oversized_width = u64::from(u32::MAX) + 101;
```

Verified with the full CI gate: `cargo fmt --check`, all six clippy aliases, `cargo test-axum -p trusted-server-core` (2669 passed), `cargo test-cloudflare`, `cargo test-spin`, and the cross-adapter parity suite (13 passed).

crates/trusted-server-core/src/integrations/adserver_mock.rs
```diff
@@ -292,8 +292,18 @@ impl AdServerMockProvider {
                 let restored_bidder =
                     original.map_or_else(|| seat_name.to_string(), |b| b.bidder.clone());
 
-                let width = bid["w"].as_u64().unwrap_or(0) as u32;
-                let height = bid["h"].as_u64().unwrap_or(0) as u32;
+                let Some(width) = bid["w"].as_u64().and_then(|v| u32::try_from(v).ok()) else {
```
aram356, line 295: 🤔 **thinking** — Worth knowing the cited precedent behaves differently here.

The PR body and issue #420 both point at `prebid.rs:3027` as the precedent for this conversion, but that code ends in `.unwrap_or(0)`:

```rust
let width = bid_obj
    .get("w")
    .and_then(serde_json::Value::as_u64)
    .and_then(|v| u32::try_from(v).ok())
    .unwrap_or(0);
```

and `prebid::parse_bid` has no zero-check afterwards — so an oversized bid there is *admitted* with `width: 0`, whereas this PR *rejects* the bid outright.

Both behaviours are defensible, and skipping is what the issue recommended and what fits this function, which already skips zero dimensions two lines below. No change requested. Flagging only so "consistent with prebid" is not taken as literally true of the resulting behaviour — the conversion style matches, the disposition does not.

crates/trusted-server-core/src/integrations/adserver_mock.rs
```diff
@@ -292,8 +292,18 @@ impl AdServerMockProvider {
...
+                    log::debug!(
+                        "adserver_mock: bid for slot '{slot_id}' has invalid width, skipping"
+                    );
+                    continue;
+                };
+                let Some(height) = bid["h"].as_u64().and_then(|v| u32::try_from(v).ok()) else {
```
aram356, line 302: 🤔 **thinking** — This rejects float-encoded dimensions that the shared helper accepts.

`serde_json`'s `as_u64()` returns `None` for any JSON float, including whole-valued ones like `300.0` (I confirmed this empirically rather than assuming). The canonical helper `parse_optional_bid_dimension` at `crates/trusted-server-core/src/auction/openrtb.rs:177-191` deliberately handles that case:

```rust
.or_else(|| {
    let dimension = raw.as_f64()?;
    (dimension.is_finite()
        && dimension > 0.0
        && dimension.fract() == 0.0
        && dimension <= f64::from(u32::MAX))
    .then_some(dimension as u32)
})
```

so a bidder sending `"w": 300.0` is accepted there and skipped here.

To be clear, this is **not a regression** — the old `unwrap_or(0) as u32` also produced `0` and skipped. Same outcome, clearer reason. And for a mock provider fed by mocktioneer it very likely never fires. Raising it only because the fix makes the divergence from the shared helper more visible, and because it is the kind of thing that bites later if this parser is ever pointed at a real bidder.

crates/trusted-server-core/src/integrations/adserver_mock.rs
```diff
@@ -292,8 +292,18 @@ impl AdServerMockProvider {
                 let restored_bidder =
                     original.map_or_else(|| seat_name.to_string(), |b| b.bidder.clone());
 
-                let width = bid["w"].as_u64().unwrap_or(0) as u32;
-                let height = bid["h"].as_u64().unwrap_or(0) as u32;
+                let Some(width) = bid["w"].as_u64().and_then(|v| u32::try_from(v).ok()) else {
+                    log::debug!(
+                        "adserver_mock: bid for slot '{slot_id}' has invalid width, skipping"
```
aram356: ⛏ **nitpick** — Consider including the offending value in the skip logs.

Both new log lines say "has invalid width/height, skipping" without the value that triggered it, while the adjacent zero-dimension log one block below does include the dimensions (`({width}x{height})`). Logging the raw `u64` would make a field report directly diagnosable — you would see `4294967396` rather than having to infer why a bid vanished.

```rust
let Some(width) = bid["w"].as_u64().and_then(|v| u32::try_from(v).ok()) else {
    log::debug!(
        "adserver_mock: bid for slot '{slot_id}' has invalid width {:?}, skipping",
        bid["w"]
    );
    continue;
};
```

Apply manually if you want it — the same treatment should go to the `height` arm, and two separate edits in one hunk cannot be expressed as a single `suggestion` block. Minor either way, since these are `debug!`.

aram356 (changes requested):
## Summary

The production fix is correct and complete. I swept the repository for the JSON-accessor-to-narrowing-`as` idiom this PR removes and found zero remaining instances — `adserver_mock.rs` really was the last one, and Prebid, APS, and the generic OpenRTB path already use checked conversions. Skipping the bid rather than clamping is the right call, and the PR body justifies it well.

The blocking issue is in the test, not the fix: `test_parse_mediation_response_skips_oversized_dimensions` passes against the unfixed code, so it does not pin the regression it was written for.

> 1 of the inline comments below carries a one-click GitHub `suggestion` — use **Commit suggestion** to apply it as a commit on the PR branch. The remaining comments are prose because they describe follow-ups or trade-offs rather than a mechanical edit.

## Blocking

### wrench

- **Oversized-dimension test passes against the unfixed code** — see inline at `crates/trusted-server-core/src/integrations/adserver_mock.rs:1343`

## Non-blocking

### thinking / out of scope / nitpick

- **Skip-vs-`unwrap_or(0)` diverges from the cited prebid precedent** — see inline at `crates/trusted-server-core/src/integrations/adserver_mock.rs:295`
- **Rejects float-encoded dimensions the shared helper accepts** — see inline at `crates/trusted-server-core/src/integrations/adserver_mock.rs:301`
- **Skip logs drop the offending value** — see inline at `crates/trusted-server-core/src/integrations/adserver_mock.rs:297`
- **`adserver_mock` is the only provider not validating dimensions against requested slot formats** — see cross-cutting below

- **Out of scope — `adserver_mock` is the only bid parser that does not validate dimensions against the requested slot formats.** Prebid and the generic OpenRTB path both call `resolve_bid_dimensions` (`crates/trusted-server-core/src/auction/openrtb.rs:195`), which rejects with `DimensionMismatch`; APS checks `dimensions.contains(...)` (`crates/trusted-server-core/src/integrations/aps.rs:671`) and `compatible_dimensions` (`aps.rs:1375`). `adserver_mock.rs:307` only checks `width == 0 || height == 0`, so it still accepts an in-range but never-requested size such as `1x1` or `9999x9999`.

  This is pre-existing and arguably acceptable for a mock provider, so it is not a blocker for this PR. Worth noting that it is the reason this path needed a bespoke dimension check at all: routing it through `parse_optional_bid_dimension` + `resolve_bid_dimensions` like the other three providers would delete the hand-rolled logic and make the truncation fix fall out for free. A reasonable follow-up issue.

- **Completeness check (informational).** Two sweeps over all crates — `as_(u64|i64|f64)\(\).*\bas (u|i)(8|16|32)\b` and `unwrap_or\([0-9]+\) as ` — returned no hits after this change, so this PR closes the last instance of the pattern rather than one of several.

## Verification performed

Scratch-verified in an isolated worktree at `6f95f1d50`:

- The PR's two new tests were grafted onto the merge-base commit `6cae7f5da` (the unfixed code). Both **passed**, confirming the oversized test does not detect the bug.
- Changing only the test input to `u64::from(u32::MAX) + 101` makes that test **fail** on the base commit (`left: 3, right: 1`) and **pass** on this PR's head — a real regression test.
- With the suggested fix applied: `cargo fmt --all -- --check`, all six clippy aliases (`fastly`, `axum`, `cloudflare`, `cloudflare-wasm`, `spin-native`, `spin-wasm`), `cargo test-axum -p trusted-server-core` (2669 passed), `cargo test-cloudflare`, `cargo test-spin`, and the cross-adapter parity suite (13 passed) all pass.

## CI Status

All 20 reported checks PASS.

- cargo test: PASS
- cargo fmt: PASS
- vitest: PASS
- format-typescript: PASS
- format-docs: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS
- integration tests: PASS

crates/trusted-server-core/src/integrations/adserver_mock.rs
dhruv8sh: Fixed in 58963a80 — changed the fixture to `u32::MAX + 101`, which truncates to 100 under the old `as u32` cast and clears the zero-check. Verified by temporarily reverting to the old cast locally: the test now fails (3 bids instead of 1); it passes with the checked conversion restored.

dhruv8sh: Fixed in 58963a80 using the suggested `u32::MAX + 101` value. Confirms your finding — reran the same before/after check you described and got the same result (fails with the old cast, passes with the fix).

dhruv8sh: Fixed in 58963a80: fixture now uses `u32::MAX + 101` and includes the load-bearing comment from your suggestion block. Verified the same way — grafting the old `as u32` cast back in makes the test fail (3 bids instead of 1).

crates/trusted-server-core/src/integrations/adserver_mock.rs
dhruv8sh: Applied in 58963a80 — both the width and height skip-log arms now include the raw `bid["w"]`/`bid["h"]` value.

crates/trusted-server-core/src/integrations/adserver_mock.rs
dhruv8sh: Noted, no change requested here — leaving the skip behavior as-is since it matches the intent of the zero-check two lines below and this is what issue #420 asked for.

crates/trusted-server-core/src/integrations/adserver_mock.rs
dhruv8sh: Noted, no change requested — agreed this is a real divergence from `parse_optional_bid_dimension`, but not worth chasing for a mock provider fed by mocktioneer. Left as-is.

crates/trusted-server-core/src/integrations/adserver_mock.rs
```diff
@@ -1323,6 +1335,116 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_parse_mediation_response_skips_oversized_dimensions() {
+        // A dimension above u32::MAX must be rejected rather than silently
+        // wrapped into a small, plausible-looking value. The offset of 101 is
+        // load-bearing: u32::MAX + 1 truncates to 0, which the zero-check
```
prk-Jr, line 1342: 👍 **praise** — This test was hardened in a follow-up commit specifically so that it fails against the buggy code, and that detail matters more than it looks.

The original fixture used `u32::MAX + 1`, which truncates to `0` under the old `as u32` cast — so the pre-existing zero-dimension check absorbed it and the test passed even without the fix. Switching to `u32::MAX + 101` (which truncates to `100`, a perfectly plausible dimension) makes it a real pin.

I verified this: reverting only the production hunk while keeping these tests fails with `left: 3, right: 1`. Leaving the load-bearing-offset rationale in the comment above is exactly the right thing to do for the next person who tries to "simplify" the constant.

crates/trusted-server-core/src/integrations/adserver_mock.rs
```diff
@@ -1323,6 +1335,116 @@ mod tests {
...
+                            "w": oversized_width,
+                            "h": 90,
+                        },
+                        {
+                            "id": "bid-oversized-height",
+                            "impid": "sidebar",
+                            "price": 1.25,
+                            "adm": "<div>Oversized height</div>",
+                            "w": 300,
+                            "h": oversized_width,
```
prk-Jr, line 1368: ⛏ **nitpick** — `oversized_width` is also used as the `h` value on the `bid-oversized-height` fixture below, so the name reads wrong at that site. `oversized_dimension` covers both.

_(Scratch-verified: applied in an isolated worktree — `cargo fmt --all -- --check` clean, all 18 `adserver_mock` tests pass, `cargo clippy-fastly` exit 0.)_

```suggestion
        let oversized_dimension = u64::from(u32::MAX) + 101;
        let mediation_response = json!({
            "id": "test-auction-123",
            "seatbid": [
                {
                    "seat": "test-bidder",
                    "bid": [
                        {
                            "id": "bid-oversized-width",
                            "impid": "header-banner",
                            "price": 3.50,
                            "adm": "<div>Oversized width</div>",
                            "w": oversized_dimension,
                            "h": 90,
                        },
                        {
                            "id": "bid-oversized-height",
                            "impid": "sidebar",
                            "price": 1.25,
                            "adm": "<div>Oversized height</div>",
                            "w": 300,
                            "h": oversized_dimension,
```

crates/trusted-server-core/src/integrations/adserver_mock.rs
```diff
@@ -1323,6 +1335,116 @@ mod tests {
...
+            provider.parse_mediation_response(&mediation_response, 200, &BidIndex::new());
+
+        assert_eq!(
+            auction_response.bids.len(),
+            1,
+            "Bids with oversized w/h should be skipped, only the valid bid should remain"
+        );
+        assert_eq!(auction_response.bids[0].slot_id, "footer");
+        assert_eq!(auction_response.bids[0].width, 728);
+        assert_eq!(auction_response.bids[0].height, 90);
```
prk-Jr, line 1394: 📝 **note** — Optional, and genuinely marginal — feel free to ignore.

The two new tests pin the rejecting side of the boundary (above `u32::MAX`, and zero/missing). Nothing pins the accepting side: a bid with `w == u32::MAX` exactly should still be kept. Adding that case would close the interval and guard against a future off-by-one in the conversion (e.g. someone swapping `u32::try_from` for a `< u32::MAX` comparison).

Not worth blocking on for a mock provider — raising it only so the choice is deliberate.

prk-Jr (approved):
## Summary

Tight, well-scoped fix that does exactly what #420 asked: the truncating `as u32` casts on bid `w`/`h` in `parse_mediation_response` become checked `u32::try_from` conversions that skip the bid on overflow, so a dimension above `u32::MAX` can no longer wrap into a small, plausible-looking value that sails past the zero-check. The PR picks the safer of the two options the issue offered (skip rather than clamp), states that choice explicitly, and matches the existing zero-dimension behaviour and the `u32::try_from(v).ok()` precedent in `prebid.rs`. No blocking findings.

> 1 of the inline comments below carries a one-click GitHub `suggestion` — use **Commit suggestion** to apply it as a commit on the PR branch. The remaining comments are prose only.

### Verification performed

Reviewed against head `855b49bf7d37b95a1c8d91b0aa0635811bfe8ed9` in an isolated worktree:

- `cargo test -p trusted-server-core --lib integrations::adserver_mock` → **18 passed, 0 failed**.
- **Differential check**: reverted *only* the production hunk back to the `as u32` casts while keeping the new tests → `test_parse_mediation_response_skips_oversized_dimensions` **FAILS** (`left: 3, right: 1`). The regression test genuinely pins the fix.
- `cargo fmt --all -- --check` clean; `cargo clippy-fastly` exit 0.
- Behaviour for missing / negative / float-encoded `w` is unchanged (skip before, skip after) — only the debug log wording differs, so #420's "missing dimension keeps its current behavior" criterion holds.
- `bid["w"]` on a missing key yields `Value::Null` via serde_json's `Index` impl, so the `{:?}` in the new `log::debug!` cannot panic.

## Non-blocking

### ⛏ nitpick

- **`oversized_width` is also used as the `h` value** — see inline at `crates/trusted-server-core/src/integrations/adserver_mock.rs:1347-1368` (one-click suggestion)

### 📝 note

- **No test pins the accepting side of the boundary (`w == u32::MAX`)** — see inline at `crates/trusted-server-core/src/integrations/adserver_mock.rs:1394`

### 👍 praise

- **Regression test was hardened to actually fail without the fix** — see inline at `crates/trusted-server-core/src/integrations/adserver_mock.rs:1342`

- 🌱 **`trusted-server-core` now has three divergent OpenRTB bid-dimension parsers** — after this PR the crate parses a bid `w`/`h` three different ways:

  | Site | Behaviour |
  | --- | --- |
  | `crates/trusted-server-core/src/auction/openrtb.rs:167` `parse_optional_bid_dimension` | `u32::try_from`, **plus** an `f64` fallback for float-encoded dimensions, **plus** a `> 0` filter, returning `BidRejectionReason::InvalidBid`. Already `pub(crate)` and consumed by `prebid.rs`. |
  | `crates/trusted-server-core/src/integrations/prebid.rs:3039-3048` | hand-rolled `as_u64().and_then(u32::try_from).unwrap_or(0)` |
  | `crates/trusted-server-core/src/integrations/adserver_mock.rs:295-314` (this PR) | hand-rolled let-else guards + a separate zero-check |

  `parse_optional_bid_dimension` already subsumes both the overflow guard *and* the zero-check this function now does in two steps, and it additionally accepts a float-encoded `728.0` that `adserver_mock` silently drops.

  **Explicitly not asking for this in this PR.** The return type is `Result<Option<u32>, BidRejectionReason>` while `parse_mediation_response` returns `AuctionResponse`, so the adaptation is non-mechanical, and AGENTS.md's "minimal changes" rule is on your side here. `adserver_mock` is a dev/mock integration, so the divergence is low-risk. Worth a follow-up issue rather than scope creep.

## CI Status

- integration tests: PASS
- cargo test: PASS
- cargo fmt: PASS
- vitest: PASS
- format-typescript: PASS
- format-docs: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS

All 20 reported checks pass; none pending, none failed.

crates/trusted-server-core/src/integrations/adserver_mock.rs
```diff
@@ -1323,6 +1335,116 @@ mod tests {
...
+
+        let auction_response =
+            provider.parse_mediation_response(&mediation_response, 200, &BidIndex::new());
+
+        assert_eq!(
+            auction_response.bids.len(),
+            1,
+            "Bids with zero or missing w/h should be skipped, only the valid bid should remain"
+        );
+        assert_eq!(auction_response.bids[0].slot_id, "footer");
```
aram356: 🔧 **wrench** — This test does not cover the zero-**height** case its name and message claim.

The three bids here are zero *width*, *missing* dimensions, and valid. The missing-dimensions bid is rejected by the width `let-else` at line 295 and `continue`s before the height branch is ever reached, so every rejection path in this test runs through width. Nothing exercises valid-width-with-zero-height.

Verified by mutation — changing line 309 to `if width == 0 {`:

```
test test_parse_mediation_response_skips_zero_dimensions ... ok
test test_parse_mediation_response_skips_oversized_dimensions ... ok
test result: ok. 2763 passed; 0 failed
```

Half of a production guard is unpinned, and not just by these two tests — the entire core suite stays green with `|| height == 0` deleted.

To be fair about scope: the zero-check is pre-existing code, not something this PR introduces. I am raising it here because this PR is what establishes dimension-validation test coverage for this file, and because the test is named `skips_zero_dimensions` and asserts "Bids with zero or missing w/h should be skipped" — it reads as covering this and does not.

The suggestion below adds a zero-height bid, moves the missing-dimensions bid to its own slot so the three rejection reasons stay distinguishable, and asserts on the surviving slot set rather than just the count — so a future regression names the leaked bid instead of printing `left: 2`. It also supplies the descriptive assertion message that `AGENTS.md` requires, which the bare `assert_eq!` on line 1445 is missing.

With this applied, the same mutation fails loudly:

```
assertion `left == right` failed: should drop the zero-width, zero-height, and missing-dimension bids
  left: ["sidebar", "footer"]
 right: ["footer"]
```

```suggestion
                        {
                            "id": "bid-zero-height",
                            "impid": "sidebar",
                            "price": 2.75,
                            "adm": "<div>Zero height</div>",
                            "w": 300,
                            "h": 0,
                        },
                        {
                            "id": "bid-missing-dimensions",
                            "impid": "skyscraper",
                            "price": 1.25,
                            "adm": "<div>Missing dimensions</div>",
                        },
                        {
                            "id": "bid-valid",
                            "impid": "footer",
                            "price": 2.00,
                            "adm": "<div>Valid Ad</div>",
                            "w": 728,
                            "h": 90,
                        }
                    ]
                }
            ],
            "cur": "USD"
        });

        let auction_response =
            provider.parse_mediation_response(&mediation_response, 200, &BidIndex::new());

        let surviving_slots: Vec<&str> = auction_response
            .bids
            .iter()
            .map(|bid| bid.slot_id.as_str())
            .collect();
        assert_eq!(
            surviving_slots,
            ["footer"],
            "should drop the zero-width, zero-height, and missing-dimension bids"
        );
```

Verified with the full CI gate: `cargo fmt --check` (rustfmt-canonical), all eight clippy aliases including `clippy-cli` and `clippy-codegen`, `cargo test-axum -p trusted-server-core` (2763 passed), `cargo test-cloudflare`, `cargo test-spin`, and parity (14 passed). No post-verify drift.

crates/trusted-server-core/src/integrations/adserver_mock.rs
```diff
@@ -292,8 +292,20 @@ impl AdServerMockProvider {
                 let restored_bidder =
                     original.map_or_else(|| seat_name.to_string(), |b| b.bidder.clone());
 
-                let width = bid["w"].as_u64().unwrap_or(0) as u32;
-                let height = bid["h"].as_u64().unwrap_or(0) as u32;
+                let Some(width) = bid["w"].as_u64().and_then(|v| u32::try_from(v).ok()) else {
+                    log::debug!(
+                        "adserver_mock: bid for slot '{slot_id}' has invalid width {:?}, skipping",
+                        bid["w"]
+                    );
```
aram356: 🤔 **thinking** — Safe as written, but this is the first `debug!` in `integrations/` to interpolate a raw upstream value.

Taking the security question seriously first, since it is the one that would matter: this is **not** a log-injection vector. `Debug` for `serde_json::Value` is a distinct impl, not a Display delegate — `Value::String` renders via `write!(f, "String({:?})", s)`, which is `str`'s `Debug`, so a bidder sending `"w": "300\nFATAL: fake log line"` produces a literal backslash-n rather than a real newline. Embedded quotes are escaped too. Size is bounded as well: `UPSTREAM_RTB_MAX_RESPONSE_BYTES` (2 MiB, `integrations/mod.rs:169`) is applied by `collect_response_bounded` at `adserver_mock.rs:378` before the parse, and the whole body is already `trace!`-logged at line 392. So this logs strictly less than the file already does one level down.

Two smaller observations, neither blocking:

1. Every other whole-payload dump in `integrations/` sits at `trace!`, not `debug!` — `adserver_mock.rs:392` and `:434`, `prebid.rs:2219` and `:3170`. This is the first `debug!` of the group.
2. `auction/openrtb.rs:62` documents the house rule for rejected upstream bids as recording them "without retaining upstream payload data", and Prebid follows it at `prebid.rs:2276-2287` by logging `impid` and the seat name rather than the offending value. Adopting `ResponseAdmissionDiagnostics` (see the cross-cutting finding in the review body) would resolve this and the observability gap together.

One thing worth a comment in the code regardless of what you decide: the newline-escaping safety above is load-bearing and currently implicit. A future refactor from `{:?}` to `{}` would silently reintroduce injection, and nothing in the file says so.

crates/trusted-server-core/src/integrations/adserver_mock.rs
```diff
@@ -1323,6 +1335,116 @@ mod tests {
...
+    #[test]
+    fn test_parse_mediation_response_skips_oversized_dimensions() {
+        // A dimension above u32::MAX must be rejected rather than silently
+        // wrapped into a small, plausible-looking value. The offset of 101 is
+        // load-bearing: u32::MAX + 1 truncates to 0, which the zero-check
+        // below would already have caught, so it would not pin this fix.
+        let config = AdServerMockConfig::default();
+        let provider = AdServerMockProvider::new(config);
+
+        let oversized_width = u64::from(u32::MAX) + 101;
```
aram356: 🌱 **seedling** — Three boundaries near this value remain unpinned.

The `+ 101` offset and its comment are right, and the mutation testing confirms this test kills both `try_from` guards. Some adjacent cases that no test covers, for a follow-up rather than this PR:

- **`w: u32::MAX` exactly** — the largest value that must be *accepted*. This is the off-by-one `u32::try_from` could plausibly get wrong, and it is the natural partner to the reject case tested here.
- **Negative values** (`"w": -1`) — `as_u64()` returns `None`, so they are skipped as invalid. Correct, but unpinned.
- **Float-encoded dimensions** (`"w": 300.0`) — also skipped, because `as_u64()` returns `None` for any JSON float.

That last one is a cross-provider divergence worth knowing about: `parse_optional_bid_dimension` at `auction/openrtb.rs:172-190` deliberately accepts integral floats via an `as_f64()` fallback guarded by `is_finite() && > 0.0 && fract() == 0.0 && <= f64::from(u32::MAX)`, so the repo has already decided `300.0` is a legitimate wire encoding for a dimension. `adserver_mock` rejects it.

This pre-dates the PR — `as_u64().unwrap_or(0)` also yielded 0 and skipped — so nothing regresses here. Flagging it because the `as_u64()` / `try_from` chain added in this PR duplicates the first two lines of that existing helper, which makes reusing `parse_optional_bid_dimension` outright a plausible cleanup. That would pick up the float handling and the slot-format validation the other three providers get, and delete the bespoke logic.

aram356 (changes requested):
## Summary

Both blocking findings from the previous pass are resolved, and I confirmed that by re-running the original repro against the new merge-base rather than taking the diff at face value: the strengthened test now **fails** on the unfixed parser (`left: 3, right: 1`) and passes on the fix. It is a genuine regression test. The two merges from `main` are mechanical and touch nothing related to this code.

I then mutation-tested each production guard independently:

| Mutation | Result |
| --- | --- |
| Revert the width guard to `as u32` | caught (`left: 2`) |
| Revert the height guard to `as u32` | caught (`left: 2`) |
| Delete `\|\| height == 0` from the zero-check | **not caught** - all 2763 core tests still pass |

That last row is the one new finding. The rest are non-blocking observations and follow-ups.

> 1 of the inline comments below carries a one-click GitHub `suggestion` - use **Commit suggestion** to apply it as a commit on the PR branch. The others are prose because they propose follow-ups rather than a mechanical edit.

## Blocking

### wrench

- **`skips_zero_dimensions` does not cover zero height** - see inline at `crates/trusted-server-core/src/integrations/adserver_mock.rs:1417`

## Non-blocking

### thinking / seedling

- **First `debug!` in `integrations/` to interpolate a raw upstream value** - see inline at `crates/trusted-server-core/src/integrations/adserver_mock.rs:297`
- **Untested boundaries: `u32::MAX` exactly, negatives, float-encoded dimensions** - see inline at `crates/trusted-server-core/src/integrations/adserver_mock.rs:1347`
- **Silent degrade to `no_bid` is a monitoring blind spot** - see cross-cutting below

- **Seedling - an all-rejected mediation response is indistinguishable from genuine no-demand.** When every bid is skipped, `parse_mediation_response` returns `AuctionResponse::no_bid(...)` at `adserver_mock.rs:352`, which is byte-identical to the mediator having had no demand: same `NoBid` status, same empty `bids`, same empty `metadata`. The `log::info!` at `adserver_mock.rs:396` reports the count *after* filtering, so it prints 0 in both cases, and the per-bid `log::debug!` lines are filtered out at the default `Info` level. A mediator that starts emitting oversized dimensions on every bid is therefore invisible in a default deployment.

  The crate already ships the tool for this: `ResponseAdmissionDiagnostics` and `BidRejectionReason` at `crates/trusted-server-core/src/auction/openrtb.rs:35-87`, wired up by Prebid at `crates/trusted-server-core/src/integrations/prebid.rs:2259-2302`, which attaches bounded rejection counts to `response.metadata`. Both types are `pub(crate)` in the same crate, so the change is roughly three `diagnostics.record(BidRejectionReason::InvalidBid)` calls at the `continue` sites plus one `attach_to` before returning.

  This is a pre-existing property rather than something this PR introduces, so it is not a blocker. Raising it because this PR widens the set of silently-dropped bids. One caveat on how much it would buy you: I found the metadata written by Prebid but could not locate a reader for `metadata["response_admission"]`, so I cannot confirm it currently reaches a dashboard. Worth checking before investing in it.

## Verification performed

Scratch-verified in an isolated worktree at `965449577` (merge-base `a4e01eb55`):

- Grafted the current tests onto the unfixed merge-base: `skips_oversized_dimensions` **FAILED** (`left: 3, right: 1`), `skips_zero_dimensions` passed. The strengthened test pins the fix.
- Mutation tests on the production guards, as tabulated above. Both `u32::try_from` guards are independently covered; the `height == 0` half of the zero-check is not covered by any test in the repository.
- With the suggested fix applied, the zero-check mutation fails loudly with `left: ["sidebar", "footer"], right: ["footer"]`.
- Full CI gate with the suggestion applied: `cargo fmt --all -- --check` (the suggestion is rustfmt-canonical), all eight clippy aliases including the newly added `clippy-cli` and `clippy-codegen`, `cargo test-axum -p trusted-server-core` (2763 passed), `cargo test-cloudflare`, `cargo test-spin`, and the cross-adapter parity suite (14 passed). Post-verify drift check: no drift.

## CI Status

All 20 reported checks PASS. They were mid-flight when this pass started and were re-checked at the end.

- cargo test: PASS
- cargo fmt: PASS
- vitest: PASS
- format-typescript: PASS
- format-docs: PASS
- CodeQL: PASS
- Analyze (rust): PASS
- Analyze (actions): PASS
- integration tests: PASS

crates/trusted-server-core/src/integrations/adserver_mock.rs
dhruv8sh: Applied in bf96305b: renamed to `oversized_dimension`.

crates/trusted-server-core/src/integrations/adserver_mock.rs
dhruv8sh: Added in bf96305b: `test_parse_mediation_response_accepts_u32_max_dimensions` (w/h == u32::MAX exactly is accepted, values preserved) and `test_parse_mediation_response_skips_negative_dimensions` (`"w": -1` and `"h": -1` are dropped, only `footer` survives). Float-encoded dimensions (`300.0`) are also rejected today because of `as_u64()`, which differs from `parse_optional_bid_dimension` in `auction/openrtb.rs`. I left that out of the tests on purpose so the divergence is not locked in. Moving this parser onto `parse_optional_bid_dimension` is a follow-up.

crates/trusted-server-core/src/integrations/adserver_mock.rs
dhruv8sh: Added in bf96305b: `test_parse_mediation_response_accepts_u32_max_dimensions` (w/h == u32::MAX exactly is accepted, values preserved) and `test_parse_mediation_response_skips_negative_dimensions` (`"w": -1` and `"h": -1` are dropped, only `footer` survives). Float-encoded dimensions (`300.0`) are also rejected today because of `as_u64()`, which differs from `parse_optional_bid_dimension` in `auction/openrtb.rs`. I left that out of the tests on purpose so the divergence is not locked in. Moving this parser onto `parse_optional_bid_dimension` is a follow-up.

crates/trusted-server-core/src/integrations/adserver_mock.rs
dhruv8sh: Applied in bf96305b. Added `bid-zero-height` (w 300, h 0) on `sidebar`, moved the missing-dimensions bid to `skyscraper`, left the valid bid on `footer`, and the test now asserts the surviving slots equal `["footer"]`. Mutation check: with `if width == 0 || height == 0` changed to `if width == 0`, the test fails with `left: ["sidebar", "footer"]`, `right: ["footer"]`.

crates/trusted-server-core/src/integrations/adserver_mock.rs
dhruv8sh: Added a comment above the width/height parsing in bf96305b. It says `{:?}` must stay on the raw upstream values because Debug escaping of newlines and quotes is what prevents log injection. Still logged at debug level so the offending value remains visible, as asked earlier.

